Issue #136, opened 2014-01-12
SDL2: Does not take screenshots


TheCycoONE:
Screenshot (snapshot) functionality is not currently implemented in SDL2.

PR #137, opened 2014-01-12
Add screenshots to SDL2 renderer

TheCycoONE:
Fixes issue #136


lewri:
Pulled this out as was browsing the wiki and noticed it in the maybe-TODO list.

There's a gist that appears to now use SDL to get PNG screenshots. Maybe it is possible now?

https://gist.github.com/malja/2193bd656fe50c203f264ce554919976

TheCycoONE:
Best not to necro old pull requests. That gist is using SDL_Image which was always an option - as would be using libpng directly.

PR #328, opened 2014-10-14
[WIP] Rats

benckx:
For now it only detects if the hospital is getting dirty and spawns rat holes accordingly. It doesn't create rat holes in the corridors and it creates some of them on the outside of the building, which is probably wrong (this happens when the walls of a room touch the walls of the hospital). I would need to get the coordinates of the hospital walls to fix these, but so far I didn't find how to get that.

Also, if I remember correctly, in TH, rat holes don't appear behind radiators (or drinking machines) and placing such an object in front of a hole was even a way to get rid of it. Can somebody confirm?

When I fire all handyman, it evolves like this (litter rate checked every month):

litter rate: 0.17 %
litter rate: 0.47 %
litter rate: 4.13 %
rat_hole_spawn_rate: 8
rat holes to make: 2
litter rate: 4.78 %
rat_hole_spawn_rate: 9
rat holes to make: 4
litter rate: 5.07 %
rat_hole_spawn_rate: 10
rat holes to make: 2
litter rate: 10.09 %
rat_hole_spawn_rate: 20
rat holes to make: 8

Here is a test saved game (an hospital with few or no handyman):

https://www.dropbox.com/sh/zfb57sosk92nbc9/AACgFreEA-lhsgpvFNToeOpQa/rats?dl=0


TheCycoONE:
I don't think there's anything gained by renaming Entity to Creature, lets drop that change. Spawning holes needs to take more into consideration as mentioned above - picking a random wall is probably still the way to go, but if the chosen wall is not suitable for a hole we should either try again or give up (if there are no suitable walls we don't want to be in a non-terminating loop).

I am not sure why rat spawning was set to require a distance > 6, in the original rats would go between much closer holes than that.

As I mentioned above, 58fe774c makes a mess of this PR and needs to be reset out.

@benckx do you plan on taking this up again?

TheCycoONE:
Replaced by #1117

Issue #147, opened 2014-01-16
vomit prevents a room from being blueprinted, and blueprint has to be moved for clean up to be detected


mattmill30:
When a person vomits, it is not possible to place a blueprint over it.
It would be much more useful for the blueprint to be queued (such as whilst people get out of the area), but for a janitor to be allowed entry to clean up the vomit and then leave again.

In addition, once vomit has been cleaned up, the blueprint does not automatically update and the room build dialog doesn't enable to tick-box, until the blueprint is edited (e.g. moved  away - and optionally back - or resized).

Alberth289346:
I just found that #708 is caused by vomit in the room. Would that issue get resolved by your patch?

J-Shep:
@Alberth289346 My patch would make a vomit tile blue instead of red in a room's blueprint and then it would delete the vomit when a player clicked OK to enter the "doors and windows" stage of editing/making a room.

I think this bug's issue is actually: how did the vomit get into this room? when patients shouldn't be able to vomit in rooms and the existing room editing code shouldn't allow a player to build room's over vomit tiles.

You could try deleting this room and then seeing if you can build a new room over it and if you can then you could debug the "edit room" code's blueprint checking code while making a new room to find out why it ignores the vomit tile.

J-Shep:
@Alberth289346 I've not looked at my patch's code for a long time but if it won't fix the issue mentioned in #66, which allows room to be built over vomit by not doing a blueprint check just before confirm is clicked, then I will make sure it does.

Alberth289346:
#708 is a crash when the game is trying to get from a built room back to blue-print by collecting the objects in the room (including some litter), and then crashing on the litter. In other words, it never reaches blue-print stage from the built room due to litter.

To me, that is different from clearing litter when placing a new blue-print. I mostly wonder what the relation is with this problem.
Does #708 happen due to a litter-accident while the room was built, or is there another way to get litter in the room? If you fix the litter-accidents while building, would problems like #708 still happen?

As for fixing stuff for 0.50, you seem to have many irons in the fire and time is running short. May I suggest you take out one iron at a time and deal with it completely, preferably selecting an almost ready iron first?

J-Shep:
[...]

PR #381, opened 2014-11-12
[WIP] Allowed rooms to be built over litter objects (all types)

J-Shep:
Dealing with issue #147 this commit's gameplay change is based
on what happens in TH when rooms are built over litter.

This commit makes tiles with litter objects on them buildable
while a blueprint is over them and then deletes litter objects
which are in the blueprint when it is confirmed.

New functions added because this commit requires them:
 World.lua:
 +findObjects()
 +findObjectsInArea()
 Utilty.lua:
 +set_get_missing_from()

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -34,6 +34,7 @@ function UIEditRoom:UIEditRoom(ui, room_type)
   self:addKeyHandler("Enter", self.confirm) -- UIPlaceObjects does not need this
 
   local app = ui.app
+  self.buildable_litter = {}
   -- Set alt palette on wall blueprint to make it red
   self.anims:setAnimationGhostPalette(124, app.gfx:loadGhost("QData", "Ghost1.dat", 6))
   -- Set on door and window blueprints too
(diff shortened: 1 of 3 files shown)


Glimmer-Man:
Tested on #388, and it works in ideal moment to build new room = no patients around and no fresh vomit on target place.

But when I try to build a room on vomit, which cannot be cleaned and other patient will place a new vomit in that area I try to build that room on, game will crush or stuck.

Anyway, if I wait for the right moment, it's a working solution to #388 . :+1:

MarkL1961:
@Glimmer-Man  The odd thing is there really shouldn't be a situation where there is more than one litter on a tile at a time.  But I have to say that I have seen it happen with vomit, for some reason it is sometimes ignoring the check that the tile is free of other objects.   If that was to be resolved then this would resolve the problem here and I think there is another issue open about multiple vomit not being cleaned in one go it would fix too.

J-Shep:
@Glimmer-Man Thanks for testing my commit, I will fix the problem you found.

<< ...for some reason it is sometimes ignoring the check that the tile is free of other objects.
@MarkL1961 I will fix this.

TheCycoONE:
J-Shep have you made any progress on this. I see it marked as a 0.5 item. If not let's remove the milestone.

J-Shep:
@TheCycoONE I've not done any work on this for a long time, I will try to get this and few other unfinished tasks I have done before the freeze deadline. I will remove the milestone though because this isn't an important gameplay change.

TheCycoONE:
As far as I can tell this and #922 are mutually exclusive pull requests, though at the moment both seem to be on hold. Could you guys sort out which of these is the more promising direction?

Alberth289346:
This PR saves the litter in a temp storage (and removes them from the world) until the final confirm, which seems like the better solution.
However, saving them just like this means the callback administration of work for handymen breaks, as a handyman may try to clean the litter in the mean time.

Either the handyman work dispatch should be changed to on-the-fly finding new work (which would also fix all the problems with broken administration of such work due to whatever reason), or this patch should handle removal of the litter from the work list of the handy men until it is clear what the fate of the litter is.

TheCycoONE:
Obsolete by #1119

Issue #342, opened 2014-10-23
I think the trunk should include a CorsixTH Eclipse project/workspace


J-Shep:
This would make setting up [Eclipse](https://www.eclipse.org/downloads/) on a computer for CorsixTH Lua & C++ programming easier especially for contributors who have never used Eclipse before.

On my PC I give CorsixTH its own Eclipse workspace for its Lua & C++ files so that I don't have to close the tabbed views for CorsixTH files I'm working with, in order to work on another unrelated project in this workspace.

If this enhancement idea is accepted then I will make a pull request for the CorsixTH workspace I use with its Lua code and C++ code projects. When a configuration I've provided is accepted and merged then I will create a setup guide in the wiki for Windows 8.1 and Ubuntu 14.04.

PR #399, opened 2014-11-25
Eclipse: Added Trunk Support For CorsixTH Eclipse Configurations

J-Shep:
#342 
Luna Workspace Configuration Screenshot
***In screenshot: the detected problems are not listed in the problems view because I've set it to only list problems for selected files/folders. This setting can't be included in the uploaded project configuration but instructions for using it are included in the setup readme file.***

I've created an Eclipse directory in the root CorsixTH directory which contains a folder for an Eclipse Luna workspace configuration containing a CorsixTH project.

The workspace and project have paths relative to the CorsixTH root directory, so they are ready to be used.

I've included a workspace configuration despite there only being one project because the benefit of giving this project its own CorisxTH workspace is that you can leave the files your working on open in the text editor view and then switch to another workspace to work with its projects files. Then you can just switch back to have all your CorsixTH files ready for you to continue using in the text editor view.

When this project is first opened a setup readme file will be displayed in the text editor view, telling users that they will need to install the Lua Development Tools plugin and a C++ development tools plugin.


TheCycoONE:
Ok, so the language files should be added back.  There isn't really sufficient instructions in the README or any links on how to use this project or what it's for.  It isn't suitable for building CorsixTH without a lot of steps.

J-Shep:
>> Ok, so the language files should be added back.

@TheCycoONE I've now done this.

>> There isn't really sufficient instructions in the README or any links on how to use this project or what it's for.

@TheCycoONE I've now replaced the README's setup instructions with a link to the WIP wiki "Eclipse Project Setup" tutorial, I will change the anchor link. I will add the other information you've asked for to the README & wiki tutorial.

>> It isn't suitable for building CorsixTH without a lot of steps.

@TheCycoONE When I want to compile CorsixTH in Windows with Visual Studio Express I simply have to open its .sln and then tell it to build the "CorsixTH" solution. I could make a compile script for Linux which could be executed by clicking a button in Eclipse.

J-Shep:
>> @TheCycoONE When I want to compile CorsixTH in Windows with Visual Studio Express 2013 I simply have to open its .sln and then tell it to build the "CorsixTH" solution. In the tutorial I've made only Step 9 needs to be done every time CorsixTH is compiled since the settings configured in the other steps are remembered.

I will find out how to setup Eclipse to compile CorsixTH's C++ source code in Windows.

Lego3:
Keep up the good work. :-) We could add C++ support in the future. This is a good first step.

PR #413, opened 2014-12-06
Added base for using a Unit test framework: Busted

Lego3:
This is a proposal how we could start creating unit tests for the Lua part of the code base. What do you guys think?

If someone could try it out on Linux that would be great. :-)


TheCycoONE:
Works well for me and I think it's a good idea.  I don't know that we need to strictly switch to test driven development immediately but if people can start writing tests that would be great.

Issue #346, opened 2014-10-26
Lua Debugger: A Eclipse Lua debugger could debug CorsixTH's Lua code at runtime


J-Shep:
I would be willing to have a go at implementing this in the near future, though I currently don't know how to implement this. I know of two Eclipse plugins which add support for Lua scripting and the "Lua Development Tools" plugin I use implements a Lua debugger.

I've scripted a gameplay mode for an [open source](https://github.com/goldeneye-source/ges-code) game called ["Goldeneye:Source"](http://www.moddb.com/mods/goldeneye-source) which allows the execution of its Python mode scripts to be controlled by the Eclipse Python debugger provided by Eclipse's PyDev plugin. 

I could find out how this game's developers have implemented runtime support for debugging scripts with Eclipse.

PR #439, opened 2014-12-20
Added support for Lua DBGp Debugging such as Eclipse debugging

J-Shep:
For issue: #346

To support this CorsixTH simply needs to have "debugger.lua" and a
"dofile('debugger')()" call to it.

This is a 3rd party Lua script and therefore won't be distributed
with CorsixTH, so developers will have to download it into their
Lua directory when setting up a computer for CorsixTH Lua DBGp
debugging.

I've made a new function for the call "dofile('debugger')()":
App:connectDebugger().

I've created a new main menu button to call connectDebugger() which
will only be in the main menu when config.debug = true. I've also
added a connectDebugger() command to the debug menu.

https://wiki.eclipse.org/LDT/User_Area/User_Guides/User_Guide_1.2#Attach_session

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -25,6 +25,7 @@ local TH = require "TH"
 local SDL = require "sdl"
 local assert, io, type, dofile, loadfile, pcall, tonumber, print, setmetatable
     = assert, io, type, dofile, loadfile, pcall, tonumber, print, setmetatable
+local runDebugger = dofile "run_debugger"
 
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
@@ -63,6 +64,12 @@ function App:App()
   self.strings = {}
   self.savegame_version = SAVEGAME_VERSION
   self.check_for_updates = true
+end
+
+--! Starts a Lua DBGp client & connects it to a DBGp server.
+--!return error_message (String) Returns an error message or nil.
+function App:connectDebugger()
+  return runDebugger()
 end
 
 function App:setCommandLine(...)
@@ -1324,6 +1331,8 @@ function App:checkForUpdates()
     -- LuaSocket is not available, just return
     print "Cannot check for updates since LuaSocket is not available."
     return
+  else
+    self.lua_socket_available = true
   end
   local http = require "socket.http"
   local url = require "socket.url"
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -129,6 +129,7 @@ menu_charts = {
 
 menu_debug = {
   jump_to_level               = "  JUMP TO LEVEL  ",
+  connect_debugger            = "  (CTRL + C) CONNECT LUA DBGp SERVER  ",
   transparent_walls           = "  (X) TRANSPARENT WALLS  ",
   limit_camera                = "  LIMIT CAMERA  ",
   disable_salary_raise        = "  DISABLE SALARY RAISE  ",
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -235,9 +235,13 @@ function UI:setupGlobalKeyHandlers()
   self:addKeyHandler({"alt", "f4"}, self, self.exitApplication)
   self:addKeyHandler({"shift", "f10"}, self, self.resetApp)
 
-  if self.app.config.debug then
-    self:addKeyHandler("f12", self, self.showLuaConsole)
-    self:addKeyHandler({"shift", "d"}, self, self.runDebugScript)
+  self:addOrRemoveDebugModeKeyHandlers()
+end
+
+function UI:connectDebugger()
+  local error_message = TheApp:connectDebugger()
+  if error_message then
+    self.ui:addWindow(UIInformation(self.ui, {error_message}))
   end
 end
 
@@ -759,9 +763,11 @@ function UI:getCursorPosition(window)
 end
 
 function UI:addOrRemoveDebugModeKeyHandlers()
+  self:removeKeyHandler({"ctrl", "c"}, self)
   self:removeKeyHandler("f12", self)
   self:removeKeyHandler({"shift", "d"}, self)
   if self.app.config.debug then
+    self:addKeyHandler({"ctrl", "c"}, self, self.connectDebugger)
     self:addKeyHandler("f12", self, self.showLuaConsole)
     self:addKeyHandler({"shift", "d"}, self, self.runDebugScript)
   end
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,4 @@ CorsixTH/*.suo
 LevelEdit/bin/
 
 CorsixTH/Lua/debug_script.lua
+CorsixTH/Lua/debugger.lua
--- a/CorsixTH/CorsixTH.lua
+++ b/CorsixTH/CorsixTH.lua
@@ -9,50 +9,13 @@ if (package and package.preload and package.preload.TH) == nil then
   error "This file must be invoked by the CorsixTH executable"
 end
 
--- Check Lua version
-if _VERSION ~= "Lua 5.1" then
-  if _VERSION == "Lua 5.2" then
-    print "Notice: Lua 5.2 is not officially supported at the moment"
-    -- Compatibility: Keep the global unpack function
-    unpack = table.unpack
-    -- Compatibility: Provide a replacement for deprecated ipairs()
-    -- NB: It might be wiser to migrate away from ipairs entirely, but the
-    -- following works as an immediate band-aid
-    local rawget, error, type = rawget, error, type
-    if not pcall(ipairs, {}) then
-      local function next_int(t, i)
-        i = i + 1
-        local v = rawget(t, i)
-        if v ~= nil then
-          return i, v
-        end
-      end
-      function ipairs(t)
-        if type(t) ~= "table" then
-          error("table expected, got " .. type(t))
-        end
-        return next_int, t, 0
-      end
-    end
-  else
-    error "Please recompile CorsixTH and link against Lua version 5.1"
+-- Parse script parameters:
+local run_debugger = false
+for _, arg in ipairs({...}) do
+  if arg:match("^%-%-connect%-lua%-dbgp") then
+    run_debugger = true
   end
 end
-
--- If being debugged in Decoda, turn off JIT compilation (as it cannot debug
--- machine code). Note that this file cannot be debugged, but all other files
--- can be. See http://www.unknownworlds.com/decoda/ for Decoda info.
-if decoda_output then
-  _DECODA = true
-  if jit then
-    jit.off()
-    decoda_output "JIT compilation disabled"
-  end
-else
-  _DECODA = false
-end
-
-_MAP_EDITOR = _MAP_EDITOR or false
 
 -- Redefine dofile such that it adds the direction name and file extension, and
 -- won't redo a file which it has previously done.
@@ -95,6 +58,60 @@ end
 
 -- Load standard library extensions
 dofile "utility"
+
+-- If requested run a Lua DBGp Debugger Client:
+if run_debugger then
+  dofile("run_debugger")()
+  done_files["run_debugger"] = nil
+end
+
+-- Check Lua version
+if _VERSION ~= "Lua 5.1" then
+  if _VERSION == "Lua 5.2" then
+    print "Notice: Lua 5.2 is not officially supported at the moment"
+    -- Compatibility: Keep the global unpack function
+    unpack = table.unpack
+    -- Compatibility: Provide a replacement for deprecated ipairs()
+    -- NB: It might be wiser to migrate away from ipairs entirely, but the
+    -- following works as an immediate band-aid
+    local rawget, error, type = rawget, error, type
+    if not pcall(ipairs, {}) then
+      local function next_int(t, i)
+        i = i + 1
+        local v = rawget(t, i)
+        if v ~= nil then
+          return i, v
+        end
+      end
+      function ipairs(t)
+        if type(t) ~= "table" then
+          error("table expected, got " .. type(t))
+        end
+        return next_int, t, 0
+      end
+    end
+  else
+    error "Please recompile CorsixTH and link against Lua version 5.1"
+  end
+end
+
+-- If being debugged in Decoda, turn off JIT compilation (as it cannot debug
+-- machine code). Note that Decoda can't debug this file but it can debug all
+-- other files. See http://www.unknownworlds.com/decoda/ for Decoda info.
+--
+-- A DBGp debugger can debug this file if you start a CorsixTH DBGp client & connect
+-- it to a running server, using this CorsixTH startup arg: -debugger
+if decoda_output then
+  _DECODA = true
+  if jit then
+    jit.off()
+    decoda_output "JIT compilation disabled"
+  end
+else
+  _DECODA = false
+end
+
+_MAP_EDITOR = _MAP_EDITOR or false
 
 -- Enable strict mode
 dofile "strict"
--- a/CorsixTH/Lua/config_finder.lua
+++ b/CorsixTH/Lua/config_finder.lua
@@ -121,6 +121,12 @@ local config_defaults = {
   audio_buffer_size = 2048,
   theme_hospital_install = [[F:\ThemeHospital\hospital]],
   debug = false,
+  DBGp_client_idehost = nil,
+  DBGp_client_ideport = nil,
+  DBGp_client_idekey = nil,
+  DBGp_client_transport = nil,
+  DBGp_client_platform = nil,
+  DBGp_client_workingdir = nil,
   track_fps = false,
   zoom_speed = 80,
   scroll_speed = 2,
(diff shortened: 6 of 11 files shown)


Lego3:
I really like this, it works great for me. :-)

CorsixTH/Lua/app.lua
```diff
@@ -313,6 +313,39 @@ function App:init()
...
 end
 
+---
+-- Initialises a Lua DBGp client and
+-- connects it to a DBGp server so that
+-- it can debug lua scripts at runtime.
+-- @return error_message
+---
+function App:connectDebugger()
+  local error_message = is_file_readable("../Lua/debugger.lua")
```
Lego3: I would prefer to find the absolute path using something like debug.getinfo(1, "S").source:sub(2, -8) (gets this file's path, see other places in the code) as the current approach relies on a specific current working directory. When running through Visual Studio the cwd is different from when you run the executable itself.

CorsixTH/Lua/app.lua
```diff
@@ -313,6 +313,39 @@ function App:init()
   return true
 end
 
+---
+-- Initialises a Lua DBGp client and
+-- connects it to a DBGp server so that
+-- it can debug lua scripts at runtime.
+-- @return error_message
+---
+function App:connectDebugger()
```
Lego3: You don't need to do this now, just a thought:

It would be really nice if this function was callable as soon as humanly possible. It could still be available for calls from the main menu etc, but at some point you might need to debug from the very beginning of the application. Then this would need to be done either in the App constructor, or at the beginning of App:init. It requires the config to be optional and the check for lua socket to lie earlier though.

CorsixTH/Lua/languages/english.lua
```diff
@@ -129,6 +129,7 @@ menu_charts = {
 
 menu_debug = {
   jump_to_level               = "  JUMP TO LEVEL  ",
+  connect_debugger            = "  CONNECT LUA DEBUGGER",
```
Lego3: Sorry if I'm too detailed now, but for consistency there should be two spaces after DEBUGGER. ;-)

CorsixTH/Lua/languages/english.lua
```diff
@@ -255,6 +257,7 @@ tooltip.main_menu = {
   options = "Tweak various settings",
   exit = "No, no, please don't leave!",
   quit = "You are about to quit from CorsixTH.   Are you sure this is what you want to do?",
+  connect_debugger = "Connect CorsixTH to a DBGp Lua debugger server. There's a CorsixTH Wiki debugger tutorial. All CorsixTH programmers should use a Lua debugger to help them quickly & easily locate bugs."
```
Lego3: Not that it hurts, but the last sentence is a little unnecessary?

CorsixTH/Lua/app.lua
```diff
@@ -313,6 +313,39 @@ function App:init()
   return true
 end
 
+---
+-- Initialises a Lua DBGp client and
+-- connects it to a DBGp server so that
+-- it can debug lua scripts at runtime.
+-- @return error_message
```
Lego3: Is this type of function comments recognized by Eclipse? We do have commenting standards tailored towards LDocGen (not entirely consistently used throughout the code unfortunately), but if another standard would be better for us we might consider changing?

CorsixTH/Lua/languages/english.lua
TheCycoONE: I find that last sentence patronizing.  A link to the debugger tutorial would be appreciated though.

J-Shep:
@Lego3 Thanks for reviewing my changes, I've made all of the changes you requested but my change for allowing the debugger to be started asap is a WIP. 

I've moved the connectDebugger() code into its own Lua script called "run_debugger" so that a) this code isn't coupled to any Lua script it can debug and b) Lua & C++ code can start the debugger by just executing this file.

I'm working on allowing run_debugger.lua to be executed before the CorsixTH.lua boot strapping script is executed by making srcUnshared/main.cpp:main() execute this script with a luaL_dofile(L, filename) call when CorsixTH is started with this parameter: -debugger

CorsixTH/Lua/app.lua
J-Shep: @Lego3 I was thinking of LuaDoc conventions when I chose this type of function comment, from now on I will use the LDocGen conventions. With the LDT plugin installed Eclipse can recognize any Lua comment type.

Eclipse does have a good DoxyGen plugin with support for Graphviz comment diagrams:
http://mcuoneclipse.com/2012/06/25/5-best-eclipse-plugins-1-eclox-with-doxygen-graphviz-and-mscgen/

J-Shep: I could make a regular expression script to make all functions comments abide by the LDocGen function comment conventions.

Lego3: Yes, that would be a good idea. :-)
Den 5 jan 2015 14:58 skrev "J-Shep" <notifications@github.com>:

J-Shep: I personally think the LDocGen function comment convention should be changed to be the same as the LuaDoc convention I've used here because a) Eclipse makes these multi-line comments blue where as all other Lua comments are green and b) a CorsixTH programmer could start a new comment line by pressing return without having to type: --!

Lego3: As long as LDocGen also recognizes and parses function comments 
correctly (to get http://www.armedpineapple.co.uk/cth/docs/) I'm fine 
with either way.

/Edvin

TheCycoONE: LDocGen was custom made by Corsix.  I'm quite ok with moving to some more standard format as long as all the tooling is updated at the same time.  Such a project is out of scope for this pull request though.

J-Shep:
@Lego3 I've not updated this pull request's commit to include all the changes you asked for. 

CorsixTH can now be started with a: -debugger parameter to make CorsixTH.lua start a Lua DBGp client before most of this bootstrap script's code has been executed.

CorsixTH/Src/main.cpp
```diff
@@ -181,10 +181,10 @@ int CorsixTH_lua_main_no_eval(lua_State *L)
         lua_getglobal(L, "assert");
         luaL_loadbuffer(L, sLuaCorsixTHLua, strlen(sLuaCorsixTHLua),
             "@main.cpp (l_main bootstrap)");
-        if(lua_gettop(L) == 2)
-            lua_pushnil(L);
-        else
-            lua_pushvalue(L, 1);
+		if (lua_gettop(L) == 3)
```
Lego3: We accidentally got some tabs here... ;-)

J-Shep:
@Lego3 I've removed the tabs you found.

Lego3:
Now it looks good to me. :-) When I run it using -debugger I am able to e.g. set a breakpoint on the first line of App:App. However, if I click to skip the introduction movie I get a "CorsixTH has stopped working" Windows message. Do you get this too? Is it because of some timeout?

J-Shep:
>> ...if I click to skip the introduction movie I get a "CorsixTH has stopped working" Windows message. Do you get this too? Is it because of some timeout?

@Lego3 I will try this later. Thanks for reviewing & merging my other pull requests.

J-Shep:
>> ...if I click to skip the introduction movie I get a "CorsixTH has stopped working" Windows message. Do you get this too? Is it because of some timeout?

@Lego3 If I can get CorsixTH to compile today with support for movies then I will test this.

J-Shep:
This is the compile problem I'm having:
https://groups.google.com/forum/#!topic/corsix-th-dev/1gXQURGYNaA

TheCycoONE:
For some reason the debugger picks the wrong location for my scripts which makes breakpoints and viewing code not work.  I can pause the debugger and see the code trace, but I get this stack:
```
dispatch  [ e:/github/corsixth/build_sdl2/corsixth/corsixth/lua/app.lua:850 ]	
```
app.lua is actually in `e:\github\corsixth\corsixth\lua\app.lua` so it seems to be appending the path from the root of the repository to my build location.

J-Shep:
@TheCycoONE The debugger isn't working for me now as well, I will investigate this problem.

J-Shep:
>> For some reason the debugger picks the wrong location for my scripts ...

@TheCycoONE I think the problem your experiencing might be being caused by your Eclipse project setup and the "Source Mapping" option your using for your "Lua attach to application" configuration for CorsixTH:

1. Local resolution (What I use):

 >> Recommended if the source files loaded by your Lua VM are physically the same than the files in your workspace (in your project buildpath)

2. Module resolution
3. Replace path resolution

J-Shep:
@TheCycoONE I was wrong, your problem is the same as my problem: the "Lua attach to application" configuration window has an option to enable DBGp logging and Eclipse's debugging perspective has a "Script Debug Log" view.

TheCycoONE:
@J-Shep did you figure out a solution / do I need to enable both of those logs?

J-Shep:
>> @J-Shep did you figure out a solution / do I need to enable both of those logs?

@TheCycoONE No I mentioned that logging could be enabled because that's how I found out that I have the same problem. The quote below is from the log message which occurs after I connect CorsixTH's DGBp client to Eclipse's DBGp server:

>> init fileuri="file:///c:/users/j_she_000/programming/github/corsixth/build/corsixth/corsixth/lua/app.lua"

J-Shep:
@Lego3 Do you have a working DBGp_Support commit branch from before whatever change to this commit/CorsixTH's master branch  prevented this commit from making Eclipse's debugger work?

J-Shep:
>> The quote below is from the log message which occurs after I connect CorsixTH's DGBp client to Eclipse's DBGp server:

This quoted log message is actually from after the Eclipse DBGp server has been started so this problem has nothing to do with CorsixTH's DBGp client.

Lego3:
I did have one. I tried it now but I currently get unrelated errors when trying to open it at all (probably wrong dlls compared to the libs during build). I might have some time tomorrow to look at it. Otherwise probably not until next week.

TheCycoONE:
I noticed you added a Startup Args.txt for documenting command line arguments.

Two notes: first, there are already more command line arguments than `-debugger`.  I know for example that `--config-file` is supported; secondly I believe we were trying to follow the getopt long name standard as much as we can, so the flag should be `--debugger`, or I suppose more precisely `--connect-lua-dbgp` (as oppose to any other remote debuggers we might theoretically support)

TheCycoONE:
Also I think '''README.txt''' might be a suitable place for those arguments instead of a new text file.

TheCycoONE:
@J-Shep I reverted your changes to main.cpp back to:
```
if(lua_gettop(L) == 2)
            lua_pushnil(L);
        else
            lua_pushvalue(L, 1); //L,1 = exe path
```
Which fixed the path problem.  My guess is that passing the exe path in is doing more than you think.

J-Shep:
@TheCycoONE Thanks for finding the solution to this problem, I will still need to modify this code to make it push the --connect-lua-dbgp parameter into the Lua stack before the dofile(CorsixTH.lua) call. I  will add this start up parameter to the README file instead of having a "Startup Args" text file.

J-Shep:
I've changed this commit:
* I've changed its commit message.
* I've fixed the path problem.
* I've renamed the -debugger variable to : --connect-lua-dbgp
* The --connect-lua-dbgp argument is now searched for in the same loop which searches for the --intepretter command.
* I've added: --connect-lua-dbgp and the other startup arguments I've found to a list of startup arguments in the README file and I've deleted "Startup Args.txt".
* I've reverted the main menu because I now don't think a "connect debugger" button is worth having in it now that the debugger can be started before the main menu and App class have been loaded, using the --connect-lua-dbgp startup argument.

CorsixTH/Src/main.cpp
```diff
@@ -166,19 +170,29 @@ int CorsixTH_lua_main_no_eval(lua_State *L)
...
-            "@main.cpp (l_main bootstrap)");
+                        "@main.cpp (l_main bootstrap)");
+        //Having pushed the functions for this statement now push its parameters:
         if(lua_gettop(L) == 2)
             lua_pushnil(L);
         else
-            lua_pushvalue(L, 1);
+            lua_pushvalue(L, 1); //The executable's path.
+
+        	if(bRunDBGpClient)
```
TheCycoONE: A couple things about this.

a.  The way this is written that code is not in the else block, but it's indented as if it were.
b.  I think this block could be pulled outside of the if(!bGotScriptFile) block so that it would be pushed regardless of whether the interpreter was specified on the command line?  I'm not 100% what's happening with the lua_State here though.

J-Shep:
>> I'm not 100% what's happening with the lua_State here though.

@TheCycoONE What I know: the protocol for calling a Lua Stack function is to first push it on to  the stack and then its parameters in order.  When its been executed by a LuaL_call(), it and  its parameters will be removed from the stack and the result will be pushed on to the stack.

The command line arguments are pushed onto the top of the stack after the Lua State has been declared in: SrcUnshared/main.cpp . They are still the top stack values when the argument search loop in Src/main.cpp is executed. 

Src/main.cpp:CorsixTH_lua_main_no_eval() finishes by either calling:
* assert(LoadFile(--interpeter=value))

 ````
lua_getglobal(L, "assert"); 
lua_getglobal(L, "loadfile");
lua_pushlstring(pushes the interpreter command's value);
 ````

* or assert(LoadString("Chunk string for executing CorsixTH.lua"))

```
lua_getglobal(L, "assert");
luaL_loadbuffer(L, sLuaCorsixTHLua, strlen(sLuaCorsixTHLua),
                          "@main.cpp (l_main bootstrap)");
Parameters then pushed...
```

>> b. I think this block could be pulled outside of the if(!bGotScriptFile) block so that it would be pushed regardless of whether the interpreter was specified on the command line?

@TheCycoONE @Lego3 If the purpose of the **--interpreter=** command is to change which Lua interpreter executes CorsixTH's scripts then why does the use of this command need to prevent the CorsixTH.lua boot strapping script from being called by: Src/main.cpp:CorsixTH_lua_main_no_eval()? Making bGotScriptFile false.

J-Shep:
>> When I run it using -debugger I am able to e.g. set a breakpoint on the first line of App:App. However, if I click to skip the introduction movie I get a "CorsixTH has stopped working" Windows message. Do you get this too? Is it because of some timeout?

@Lego3 I've still not tried this myself because I've not yet compiled CorsixTH with support for movies. But I've noticed that CorsixTH crashes when its running a Lua DBGp client connected to a DBGp server and its audio is turned on in the main menu's options. The error message is about a memory access problem. So maybe the movie crash you experienced was caused by the same problem.

J-Shep:
I've changed Utility:is_file_unreadable(...) .

TheCycoONE:
As far as I've been able to tell, the purpose of --interpreter= is to specify an alternate lua script to launch instead of CorsixTH.lua.  (Maybe one that supported a different interpreter?)  I have doubts that anyone uses it, or anyone other than @corsix knows exactly what it's intended purpose is.

What I meant above is that while I know more or less how lua_State works, I haven't followed closely enough to understand why many things are done the way they are in this function.  Like why there is a check for `if(lua_gettop(L) == 2)` that pushes a `nil` and skips adding the program path.

My interpretation (which needs more investigation for any degree of certainty) is that the program path `lua_pushvalue(L, 1)` is pushed to the top of the stack so that it would be included in the ... parameter of `@main.cpp (l_main bootstrap)` which specifies the root path to start looking for CorsixTH.lua from.  I would have thought that `lua_pushstring(L, "--connect-lua-dbgp");` could be done outside of that if block, maybe in the same if block you execute the `lua_call` with two arguments.

Anyway I think what I really need to do is write a small function that dumps the entire stack and print it at various points throughout that function while calling CorsixTH.exe with different arguments.

TheCycoONE:
@J-Shep I've done some investigation on this now, and there's no reason to modify main.cpp.  All command line arguments are passed, which is how they could end up in App:setCommandLine(...).

In my CorsixTH.lua
```
local code_dir = base_dir .. "Lua" .. pathsep
for _, arg in ipairs{...} do
  local dir = arg:match"^%-%-lua%-dir=(.*)$"
  local debug = arg:match"^%-%-connect%-lua%-dbgp"
  if debug then
    print 'happy'
  else
    print 'sad'
  end
  if dir then
    code_dir = dir .. pathsep
  end
end
```

That should simplify things.

TheCycoONE:
As you can see above, `--lua-dir=` is another supported command line parameter too.

TheCycoONE:
Finally the readme.txt changes can be a separate commit.

task 1. Add lua debugger support via dbgp
task 2. Document command line options

J-Shep:
@TheCycoONE I've made all the changes you asked for thanks for helping me improve this pull request.

J-Shep:
I need to correct one mistake.

J-Shep:
>> I need to correct one mistake.

Corrected: I had committed debug script test code.

TheCycoONE:
Debugger is working great for me, no issue with either movies or music.

J-Shep:
I've made the README changes @TheCycoONE asked for and I've also now created a CTRL + C global key handler for calling App:connectDebugger().

J-Shep:
As well as adding the new CTRL + C global key handler, the DBGp implementing commit now also makes the debug console and debug script executing commands usable in CorsixTH's pre-game menus when debug mode is enabled.

TheCycoONE:
Very close to merging this, I'll just want to play with it a little more first.  In the mean time, have you looked at the Zerobrane Lua IDE (http://studio.zerobrane.com/).  It appears to support a somewhat similar remote debugging mechanism, as well as live code updating, and support for debugging any lua interpreter luasocket works with (lua 5.1/5.2/5.3, luajit 2.0+)

Maybe just review to make sure it would be easy to add support for this as well later, on the surface it looks as simple as calling modebug.lua instead of debugger.lua.

TheCycoONE:
Alright, I'm confident from another read over of the changes that it will be easy to add zerobrane support in the future if we want.

J-Shep:
@TheCycoONE Thanks for reviewing my pull request.

J-Shep:
>> ... have you looked at the Zerobrane Lua IDE (http://studio.zerobrane.com/). It appears to support a somewhat similar remote debugging mechanism, as well as live code updating, and support for debugging any lua interpreter luasocket works with (lua 5.1/5.2/5.3, luajit 2.0+)

@TheCycoONE I've not tried this IDE before but I can't remember my experience of using it. Live code updating does sound useful.

CorsixTH/Lua/ui.lua
```diff
@@ -235,9 +235,13 @@ function UI:setupGlobalKeyHandlers()
...
-  if self.app.config.debug then
-    self:addKeyHandler("f12", self, self.showLuaConsole)
-    self:addKeyHandler({"shift", "d"}, self, self.runDebugScript)
+  self:addOrRemoveDebugModeKeyHandlers()
+end
+
+function UI:connectDebugger()
+  local error_message = TheApp:connectDebugger()
+  if error_message then
+    self.ui:addWindow(UIInformation(self.ui, {error_message}))
```
sadger: I tried running the debugger without eclipse set up and got a nil pointer here. These references to "self.ui" should just read "self" as you are in the UI class. I made these changes myself and it seemed to show the correct error message then.

sadger:
I've worked out how zerobrane works now and will be trying to add as an option to the debug config.

CorsixTH/Lua/ui.lua
J-Shep: @sadger Thanks for posting your fix for this mistake, I will submit a pull request for it.

J-Shep: I see this mistake has already been fixed.

sadger: Yeah was just a simple PR, no problem :+1:

PR #485, opened 2014-12-23
Fix 32bpp pixel access

Alberth289346:
32bpp image data is not "width * height * 4bytes", you have to decode it.
(function is sort-of a copy of a draw-sprite function)

This probably never failed because the legacy 8bpp -> 32bpp conversion is kind of stupid, so you always have enough data to pretend pData[y*width + x] works.


TheCycoONE:
Merged.

PR #486, opened 2014-12-23
Sdl2 add animations

Alberth289346:
Adding CorsixTH side of loading the animation files produced by the animation encoder of the graphics repository.


TheCycoONE:
Merged.  Thanks @Alberth289346

PR #585, opened 2014-12-28
French translations

wolfy1339:
Using the Dump Strings feature on CorsixTH, i've translated some missing strings


wolfy1339:
Moved to #601

PR #595, opened 2015-01-03
Added a shell script to help get started on a fresh Ubuntu machine

Lego3:
I haven't done much shell scripting in my life (almost none) but I thought I'd share a small script to help Ubuntu users get started at least. What do you guys think?

Where should it lie, what should it be called?

PS: I couldn't get the Anim Viewer to compile at the moment, has someone else tried that on Linux recently?

### DIFF
--- a/run_to_initialize_on_ubuntu.sh
+++ b/run_to_initialize_on_ubuntu.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+MOVIES="-D WITH_MOVIES=OFF"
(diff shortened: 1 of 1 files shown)


TheCycoONE:
I think this script is mostly obsolete for master.  #621 makes the movie section unnecessary (at least for Ubuntu 14.10 and higher), and sdl2 means the map editor flag isn't currently useful.

TheCycoONE:
Perhaps it should go in the wiki as a recommendation for 0.40?

Lego3:
Indeed. I might add something to the wiki or skip it alltogether.

Issue #588, opened 2014-12-29
Lava hole freeze when using French sounds


Sigzegv:
Built 0.40 release under linux ( archlinux ). 

Game seems to freeze every time a lava hole pops when a patient dies.

J-Shep:
I will try to fix this bug today.

J-Shep:
>> The lava sounds do look correct in all the files, but only 5 crashes for me.

This crash occurs for SOUND-5.dat when LAVA004.WAV is played.

Sigzegv:
Did you got the crash with SOUND-1.dat like me ?

J-Shep:
>> Did you got the crash with SOUND-1.dat like me ?

@Sigzegv No, like @MarkL1961 only SOUND-5.dat is crashing my CorsixTH. My TH sound data files are from a gog.com TH install. Is your SOUND-5.dat file from a CD install/demo?

J-Shep:
>> This crash occurs for SOUND-5.dat when LAVA004.WAV is played.

LAVA004.WAV is actually played but the crash occurs when CorsixTH tries to play the next Lava sound. I will try to fix this tonight.

Lego3:
It might be worth including some fallback if possible. There might be some versions of Theme Hospital that lack certain sounds. We have had that issue before (I think it was the Italian original release or something that lacked a certain "idle" sound)

J-Shep:
>> It might be worth including some fallback if possible. There might be some versions of Theme Hospital that lack certain sounds. We have had that issue before (I think it was the Italian original release or something that lacked a certain "idle" sound)

@Lego3 I will create a fallback response and error print statement for when a sound is missing from a releases's active SOUND-X.dat. LAVA004.WAV isn't missing from my SOUND-5.dat.

There's no response for [SDL.Mix_PlayChannelTimed()](http://sdl.beuc.net/sdl.wiki/Mix_PlayChannelTimed) returning an error, so I will add a print call for Mix_GetError():

```
void THSoundEffects::_playRaw(size_t iIndex, int iVolume)
{
    int iChannel = reserveChannel();
    Mix_Volume(iChannel, iVolume);
    Mix_PlayChannelTimed(iChannel, m_ppSounds[iIndex], 0, -1) == -1;
}
```

J-Shep:
SDL.Mix_PlayChannelTimed() is returning a channel number instead of -1 (critical error) when LAVA004.wav is played.

MarkL1961:
[...]

PR #600, opened 2015-01-04
Fixed th_sound:getSoundDuration(index) crash error:#588

J-Shep:
This function's loop for parsing a WAV file would crash CorsixTH by
getting stuck in an infinite loop because this loop has no end condition
and none of its break statements were executed.

Each language TH supports has its own sound effects stored in
SOUND-X.dat. This error would occur for LAVA004.WAV (182) in:
*SOUND-1.dat (CD distribution)
*SOUND-5.dat (gog.com TH distribution)

I've simply inserted a break statement after this loop's variable
assignment statement for the last duration calculation variable it
parses.

I've tested this with the SOUND-5.dat from the gog.com distribution but
I've not tested this with the SOUND-1.dat from @Sigzegv 's TH CD.

### DIFF
--- a/CorsixTH/Src/th_sound.cpp
+++ b/CorsixTH/Src/th_sound.cpp
@@ -125,9 +125,11 @@ size_t THSoundArchive::getSoundDuration(size_t iIndex)
                 break;
             iChunkLength -= 16;
         }
+        //Finally:
         if(iFourCC == FOURCC('d','a','t','a'))
         {
             iWaveDataLength = iChunkLength;
+            break;
         }
         if(SDL_RWseek(pFile, iChunkLength + (iChunkLength & 1), SEEK_CUR) == -1)
             break;


J-Shep:
This fix should be included in a release/released patch asap because this error will always crash the game when hell deaths occur for players who are not using the English sound effects.

TheCycoONE:
I'm a little confused how this gets into an infinite loop unless iChunkLength is negative.  It should keep reading until it hits EOF and break.

Sigzegv:
I can share my Sound-1.dat if needed.

J-Shep:
>> I'm a little confused how this gets into an infinite loop unless iChunkLength is negative. It should keep reading until it hits EOF and break.

Well whatever causes this loop to get stuck, the break statement I've added ends this WAV duration data parsing immediately after its parsed all the required data and it prevents this game crash error from occurring.

>> I can share my Sound-1.dat if needed.

@Sigzegv If your using Windows 8.1 64bit I could give you a CorsixTH.exe to test and if it fixes the crash error you've experienced you could carry on using it with your language set to French.

Sigzegv:
Sorry I'am under linux :) I will be able to rebuild when the patch is commited

TheCycoONE:
@J-Shep I agree that exiting the loop after reading the data size is adequate.  I would just like to know what condition causes it to be a continuous loop and what this means for the actual result and the wav.

The final seek is for padded out `fmt ` headers and to skip unprocessed subchunks like `info`, but it should have also skipped to the end of the file on `data` and caused a break on the next iteration.  I was thinking maybe if the `data` chunk length didn't fill the file it might cause issues, but at the beginning of the loop we read 8 bytes forward unconditionally so eventually we would hit the end.  The only thing I can think of is that the data length is negative, but that would cause incorrect results even with your fix.

TheCycoONE:
My negative chunk size theory doesn't pan out.  It's a uint32_t, and while RWSeek takes a signed int, it's 64bit so the chunk size would only ever be positive.

J-Shep:
>> I would just like to know what condition causes it to be a continuous loop and what this means for the actual result and the wav.

@TheCycoONE I will investigate this when I have time.

PR #603, opened 2015-01-10
Sdl2

TheCycoONE:
The big switch to SDL2.  There are a couple known issues at the moment, the biggest being that the map editor doesn't work at all but SDL2 is intended to be our focus going forward.

Note that the pull request can't be as clean as we regularly like because the SDL2 branch was developed in public.

Cheers.


Lego3:
Looks to be working good to me. I'll give it a day to let someone more try it out briefly before we merge.

PR #607, opened 2015-01-13
Updated Windows installer to only take into account one executable

Lego3:
With SDL2 we should move to one executable imo. Then we would update the installer like this.


admdly:
Nice to see things moving along on this front; the multiple executables never made much sense imho. I'll be glad to see this merged and naturally support this.

TheCycoONE:
Can the unicode changes be a separate commit?

Lego3:
Done. :-)

PR #610, opened 2015-01-17
Drop included source for LFS and LPEG

TheCycoONE:
Newer versions of both libraries exist and they should be managed
externally, e.g. through luarocks.

LFS had been extended with a volumes method, this commit moves that method
to TH.LfsExt.

Note: After this commit, on windows you will need to have luafilesytem and lpeg dlls included along with the other dlls in your build folder.


TheCycoONE:
libsdl.org is down at the moment which is causing the travis CI failure.

Lego3:
Have you been able to produce 64-bit dlls for Windows? I haven't found any precompiled ones, and haven't managed to compile any using Visual Studio.

Lego3:
Works fine on Ubuntu though

TheCycoONE:
Yes, actually I developed this against Windows x64, using LuaRocks and Visual Studio 2013 CE.  IIRC I just installed the packages in luarocks from the Visual Studio x64 terminal and it compiled them correctly.

Lego3:
I see now that I ran the wrong command prompt. Apparently the VS2013 Pro installer adds certain shortcuts to the start menu which are not the correct ones...

CorsixTH/Src/th_lua_lfs_ext.cpp
```diff
@@ -0,0 +1,104 @@
+/*
+Copyright (c) 2010 Peter "Corsix" Cawley
+Copyright (c) 2014 Stephen E. Baker
+
```
Lego3: This whole file contains just tabs instead of spaces.

TheCycoONE:
Uck, sorry about that.  Must not have my settings configured in that IDE.

mrtnptrs:
Well, this seemed to broke Travis, as it now produces an error when trying to compile a new build: The command "sudo luarocks install lpeg" failed and exited with 1 during .

Lego3:
From what I can see it looks like luarocks.org is down at the moment.

mrtnptrs:
Then very sorry! :) Hopefully this will be fixed when luarocks comes back online.

TheCycoONE:
It might be worth noting that newer luarocks versions use other mirrors for packages so they continue to work.  It might be worth getting lua rocks from some other ppa?

jorgenpt:
This change is a little frustrating -- there's no CMake step that checks for them, and I don't even see how we are bundling these libraries with the game. The first time this will fail with an error is when I try to start the game and Lua scripts look for the `lfs` / `lpeg` requires. There's no indication on how to resolve this problem. In addition, we require a specific version of these libraries, so why is it not versioned in sync with the tree? (At least for LFS, we could use a git submodule to manage it, if need be.)

If you make a build now, will it rely on the users to install these packages?

TheCycoONE:
@jorgenpt We don't require specific versions of these libraries, in fact the versions we were bundling were archaic.  To ship the libraries with the game the dlls will be included in the CorsixTH folder.

TheCycoONE:
They are run time dependencies, not compile time dependencies so it doesn't make sense for CMake to check for them, it would be the responsibility of the packager to ensure they were included.

jorgenpt:
I don't think the distinction between "run time" and "compile time" makes sense here -- they are required parts of the build. CMake produces a non-working product (.app bundle, in the context of Mac.)

I think CMake should both check for the existence of these libraries, *and* copy them into the bundle, so that our builds produce a fully functioning package.

jorgenpt:
The danger of **not** enforcing this compile-time is that the packager will have these things installed in a system-accessible location, the .exe will work fine for them, they will ship an installer / zip / whatever to customers, and then the game won't work on their machine (because they don't have it in a system-accessible location.)

MarkL1961:
Is this why I am having issues compiling today?  If the process is now different then perhaps the "how to compile" instructions should be updated.  On a side not there is a r missing from occurred in the message as well.

Notice: LuaJIT not being used.
Consider replacing Lua with LuaJIT to improve performance.
An error has occured in CorsixTH:
...uild\CorsixTH\Debug\..\..\..\CorsixTH\Lua\strict.lua:66: module 'lfs' not fou
nd:
        no field package.preload['lfs']
        no file '.\lfs.lua'
        no file 'C:\Users\Mark\Documents\GitHub\MarkL\CorsixTH\build\CorsixTH\De
bug\lua\lfs.lua'
        no file 'C:\Users\Mark\Documents\GitHub\MarkL\CorsixTH\build\CorsixTH\De
bug\lua\lfs\init.lua'
        no file 'C:\Users\Mark\Documents\GitHub\MarkL\CorsixTH\build\CorsixTH\De
bug\lfs.lua'
        no file 'C:\Users\Mark\Documents\GitHub\MarkL\CorsixTH\build\CorsixTH\De
bug\lfs\init.lua'
        no file 'C:\Program Files (x86)\Lua\5.1\lua\lfs.luac'
        no file '.\lfs.dll'
        no file '.\lfs51.dll'
        no file 'C:\Users\Mark\Documents\GitHub\MarkL\CorsixTH\build\CorsixTH\De
bug\lfs.dll'
        no file 'C:\Users\Mark\Documents\GitHub\MarkL\CorsixTH\build\CorsixTH\De
bug\lfs51.dll'
        no file 'C:\Users\Mark\Documents\GitHub\MarkL\CorsixTH\build\CorsixTH\De
bug\clibs\lfs.dll'
        no file 'C:\Users\Mark\Documents\GitHub\MarkL\CorsixTH\build\CorsixTH\De
bug\clibs\lfs51.dll'
        no file 'C:\Users\Mark\Documents\GitHub\MarkL\CorsixTH\build\CorsixTH\De
bug\loadall.dll'
        no file 'C:\Users\Mark\Documents\GitHub\MarkL\CorsixTH\build\CorsixTH\De
bug\clibs\loadall.dll'
stack traceback:
        [C]: in function 'fn'
        ...uild\CorsixTH\Debug\..\..\..\CorsixTH\Lua\strict.lua:66: in function
'require'
        ...H\build\CorsixTH\Debug\..\..\..\CorsixTH\Lua\app.lua:23: in main chun
k
        [C]: in function 'dofile'
        ...\build\CorsixTH\Debug\..\..\..\CorsixTH\CorsixTH.lua:93: in function
'dofile'
        ...\build\CorsixTH\Debug\..\..\..\CorsixTH\CorsixTH.lua:107: in main chu
nk
        [C]: ?

Lego3:
I updated the how to compile and Windows dev pages when i merged this. I do agree with jorgenpt though. It gets harder and harder to "do it yourself". Thats the most annoying thing with open source in general imo. When you have to struggle for two days before it works.

TheCycoONE:
I don't mind a CMake check for OSX to include it in the bundle.

On Windows, CMake doesn't handle any dlls at all, which is why the distinction between run time and compile time dependencies is important.  A packager cannot rely on CMake or having the program run on their system already because of system dlls like msvcr100.dll and msvcp100.dll, and other lua modules like lua socket which are not part of CMake.

On Linux the distribution's copies of these libraries should be used.  It is possible for the package maintainer not to have a copy of the lua modules but to specify them as required in their packages.

If we want to help people who aren't reading the wiki we could make a more informative failure message when the libraries are missing and/or a message in CMake informing people they will need them.

MarkL1961:
In the updated wiki the path is http://luarocks.org/en/Download., but when I click on it, I am taken to https://rocks.moonscript.org/.  Although I can see a link there for LuaFileSystem I cannot see how to install this Lua Rocks thing on windows!

Lego3:
I suspect that they have temporarily changed to pointer for luarocks.org 
to moonscript. That site has been down for a while, which is why the 
Travis builds fail at the moment. Let's hope that it's temporary at 
least... There was a fairly good walkthrough on the old site.

/Edvin

MarkL1961:
Thanks Edvin, I'll keep a lookout - thought I was going crazy with not seeing some obvious download link

MarkL1961:
Would this lfs change stop the daily builds from working as I get an error message about lfs?

TheCycoONE:
Ah yes, lfs.dll, lpeg.dll, and probably mime/core.dll socket/core.dll, /lua/socket.lua and /lua/socket/*.lua should be included in the zip sent to alan's site, thanks for the heads up MarkL.  Maybe open an issue about it so I don't forget.

TheCycoONE:
Created #686 to address the missing `lfs.dll` and `lpeg.dll`.  Ran out of time tonight to include all the luasocket files, but they are not as necessary yet.

TheCycoONE:
And... added luasocket as well.  All the dependencies should be there now. :-)

hishamhm:
Hi, the MoonRocks website indeed took over the role as the LuaRocks.org website. The download URLs still work as always, but we still need to convert the documentation (raw files currently at https://github.com/hishamhm/luarocks-docs ). Note that even during the luarocks.org downtime, the default repository for rocks was already rocks.moonscript.org and it stayed up all the time — if you got build errors during the luarocks.org downtime, maybe you were using an old version that still fetched rocks via luarocks.org? This can be fixed via configuration.

Issue #605, opened 2015-01-12
Failed to index field black_font 


sadger:
Arch linux 64 bit. Loaded from the most recent master ref 6316114

Loaded a new-ish save from 0.40 save index 86 created using save index 80.

Staff dialog was open when the save was made so would appear immediately on loading. 
Crashes with error.

Error in frame handler: 
./Lua/dialogs/staff_rise.lua:140: attempt to index field 'black_font' (a nil value)
stack traceback:
	./Lua/dialogs/staff_rise.lua:140: in function 'draw'
	./Lua/window.lua:1250: in function 'draw'
	./Lua/game_ui.lua:150: in function 'draw'
	./Lua/app.lua:876: in function <./Lua/app.lua:869>
	(tail call): ?
	./Lua/app.lua:758: in function <./Lua/app.lua:753>
Warning: No event handler for frame

TheCycoONE:
I made the change that caused this issue, sorry. (https://github.com/CorsixTH/CorsixTH/commit/9e962a022715c24b0eab8b51454f4db4019fd6e3)  You wouldn't be able to load the save in 0.40 final either.

sadger:
Ah ok. Removed the SDL2 bit from the title then.

Lego3:
Hmm. What do we do with this type of thing? We create a backwards compatibility fix and include it in a 0.41 release maybe?

TheCycoONE:
@sadger would you mind to attach the problematic save to the ticket.

MarkL1961:
@sadger Can you load your save in an earlier version, save it when you closed the open menu (was it staff wage request) and then load in current version?

sadger:
Save file: https://www.dropbox.com/s/tiyi123les6pdfu/epidemicpatients.sav?dl=0

Can confirm it works pre 9e962a0.

Loading in an older version and closing the dialog and saving again fixes the save and new pop up windows seem to work. Looks like it's affecting saves where the dialog is present when the file was saved. Hope that helps..

MarkL1961:
I would say that is going to be rare, so does it need a fix?

MarkL1961:
I can't think of any games that remember what menus you had open when the save was created.
Would a solution be for the future for us to stop our saves from remembering?
That would also solve the issue where you cannot create a save if the jukebox menu is open and if we had done this earlier would have stopped this issue.

sadger:
[...]

PR #617, opened 2015-01-21
SDL 2: Renamed "enter" key character references to: "return"

J-Shep:
SDL 2 changed the name of this key from "enter" to "return" so
this commit updates references to this key's name.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -29,7 +29,7 @@ local assert, io, type, dofile, loadfile, pcall, tonumber, print, setmetatable
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 100
+local SAVEGAME_VERSION = 101
 
 class "App"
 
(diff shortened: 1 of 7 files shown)


sgtbigman:
Guessing we'll want to eventually update Corsix's keyhandler to use SDL_Keycode Values so that this kind of issue is automatically handled.

TheCycoONE:
This pull request doesn't handle updating save games.  I think it should be fine as long as no dialogs are open though?

J-Shep:
>> This pull request doesn't handle updating save games.

@TheCycoONE I will add code for this.

J-Shep:
I've pushed my latest code for this commit I will test it now.

J-Shep:
I've tested the afterLoad() changes with each affected dialog with 0.40 saves and they all work except for the "edit room" dialog which crashes the game when return is pressed but this crash isn't being caused by my simple change because it also occurs in 0.40 when I try to load this 0.40 saved game (with its room blueprint & edit room dialog) and then press return to confirm the blueprint's shape phase.

TheCycoONE:
Yes, the edit room bug is mine, which is why I thought to mention it. #605

PR #620, opened 2015-01-22
Load black_font if it doesn't exist in staff_rise.lua

### LINKED ISSUE #605: Failed to index field black_font

TheCycoONE:
Fixes #605. Old saves did not include that font.

### DIFF
--- a/CorsixTH/Lua/dialogs/staff_rise.lua
+++ b/CorsixTH/Lua/dialogs/staff_rise.lua
@@ -190,3 +190,9 @@ function UIStaffRise:increaseSalary()
     world:setSpeed(world.prev_speed)
   end
 end
+
+function UIStaffRise:afterLoad(old, new)
+  if not self.black_font then
+    self.black_font = self.ui.app.gfx:loadFont("QData", "Font00V")
+  end
+end


Lego3:
If we decide to create a 0.41 at some point we can always cherry-pick onto 0.40 instead. It's impossible to know in advance exactly what we will want in a 0.41 release imo.

Issue #594, opened 2015-01-02
Ubuntu 14.04: compile error


legluondunet:
Hello,

I tried to compile the last git code on Ubuntu 14.04 and I obtain this error:

$  cmake .
Note: Using SDL as renderer (default)
Note: SDL audio is enabled (default)
Note: FFMPEG video is enabled (default)
Note: FreeType2 is enabled (default)

Building CorsixTH
  SDL found
-- Found LuaJIT library: /usr/lib/x86_64-linux-gnu/libluajit-5.1.so
  LuaJIT found
  SDL_mixer found
-- CMAKE_MODULE_PATH=/home/legluondunet/Applications_persos/Jeux/DOS/Theme_Hospital/Corsix-th/CorsixTH-master03012015/CMake
CMake Error at /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:108 (message):
  Could NOT find FFmpeg (missing: SWRESAMPLE_LIBRARIES
  SWRESAMPLE_INCLUDE_DIRS)
Call Stack (most recent call first):
  /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:315 (_FPHSA_FAILURE_MESSAGE)
  CMake/FindFFmpeg.cmake:145 (find_package_handle_standard_args)
  CorsixTH/CMakeLists.txt:142 (FIND_PACKAGE)

-- Configuring incomplete, errors occurred!
See also "/home/legluondunet/Applications_persos/Jeux/DOS/Theme_Hospital/Corsix-th/CorsixTH-master03012015/CMakeFiles/CMakeOutput.log".
See also "/home/legluondunet/Applications_persos/Jeux/DOS/Theme_Hospital/Corsix-th/CorsixTH-master03012015/CMakeFiles/CMakeError.log".

Could you help me?

LGDN

wolfy1339:
Do you have the ffmpeg libraries installed? That's what that error points to; it can't find the FFMPEG development libraries.

legluondunet:
Ubuntu does not use ffmpeg but libav fork...FFMPEG is no more in the Ubuntu official depot.

Lego3:
Indeed. Therefore you need to add Severinsson's PPA if you want movie support. If you don't need it you can just do -D WITH_MOVIES=OFF. Otherwise the compile instructions link you to his site: https://launchpad.net/~jon-severinsson/+archive/ubuntu/ffmpeg
[...]

PR #621, opened 2015-01-23
Support playing movies with libav and fix ffmpeg audio stutter

TheCycoONE:
Makes libav an optional alternative to ffmpeg, selected in cmake.

The intro movie actually plays better with this libav code, but the frame
rate had to be determined experimentally as time_base for these videos in
libav seems to be giving a value of 0/1.

This addresses the issue of building on Ubuntu (e.g. #594)

The second commit fixes the audio stutter in ffmpeg, making it use the same path for calculating the out samples as libav is using.

The third commit fixes deprecated function warnings while falling back to earlier versions when those functions are not available.  Most importantly it uses av_frame_free instead of av_free for frames which is needed to free extended data associated with frames.


TheCycoONE:
Actually I'd like to look into the framerate issue a little more before this is merged.  The value I came up with is perfect for the intro video (17.5fps) but the lose videos have a drastically different intended framerate (8.333fps).

TheCycoONE:
Actually I was pretty close to the answer.  It seems I should be using the time_base from the AVStream not the AVCodecContext.  I'll fix this tonight or tomorrow night.

TheCycoONE:
I've corrected the framerate issue, and now I'm quite happy with the state of playing videos with libav.  (So happy that unless I can improve the ffmpeg version I'm going to ship 0.50 with libav)

TheCycoONE:
Added a second commit to this pull request which fixes the stutter in ffmpeg.  Incidentally I had caused it trying to take advantage of newer features.  That they weren't available in libav and audio was working fine lead me to the solution.

TheCycoONE:
Added explicit include for libavutil/mathematics.h for libav-9.16 as reported by @legluondunet

TheCycoONE:
Note: I have verified that this works with libav-11.1 for windows and Ubuntu 14.10.  This code does not work on the earlier libav-9.16 included in Ubuntu 14.04 LTS.  (Of note: avplay in ubuntu 14.04 is also unable to properly play intro.sm4 so it *may* not be possible to support our movies with that version of libav.)

Lego3:
I didn't manage to create 64-bit binaries for Visual Studio here either... I think I'll scrap Visual Studio soon. :-P On Ubuntu on the other hand everything worked as a charm as usual.

Lego3:
While the sound doesn't stutter anymore I get some image stuttering back and forth when running on a virtual Ubuntu machine (always had this)

TheCycoONE:
Again you have to run the libav compile instructions from the Visual Studio x64 console.  Enhanced edition can't, but community edition and pro have it.

TheCycoONE:
@Lego3 when you say 'always had this', you mean you had an issue with video stuttering in ffmepg as well and libav just doesn't fix it?  Is there anything holding this back from being merged?

Lego3:
I don't think there is anything holding it back, no. I managed to compile once I found the correct command prompt. :-P CMake still asks for postproc, which I couldn't see is a part of libav. But the game works fine without it...

Lego3:
And by the way, has SDL_mixer inclusion always been under the advanced tab in CMake?

TheCycoONE:
You are right about postproc. I'll fix FindLibAV to match all the right libraries before it's merged. Not sure about sdl2mixer - not a change I made.

TheCycoONE:
Removed postproc and add libavfilter to FindLibAV.cmake for correctness.  Neither are used by CorsixTH.

PR #626, opened 2015-01-23
Fix crash on screenshot

pwaller:
I pressed CTRL-S expecting to save my game, and it rudely crashed instead.

The reason was that pWindowSurface was NULL.

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -456,6 +456,12 @@ bool THRenderTarget::takeScreenshot(const char* sFile)
     //cannot directly save it.  Instead we have to create a new surface based
     //on the pixel data in the renderer and save that.
     SDL_Surface* pWindowSurface = SDL_GetWindowSurface(m_pWindow);
+    if (pWindowSurface == NULL)
+    {
+        fprintf(stderr, "Unable to take screenshot (pWindowSurface == NULL)\n");
+        return false;
+    }
+
     SDL_Surface* pRgbSurface = NULL;
     int iPitch = pWindowSurface->w * pWindowSurface->format->BitsPerPixel;
     unsigned char* pPixels = new unsigned char[pWindowSurface->h * iPitch];


pwaller:
Done @TheCycoONE

TheCycoONE:
Spiffy, thanks!

pwaller:
FWIW, the error I get is now:

```
Unable to take screenshot (pWindowSurface == NULL): No hardware accelerated renderers available
```

jorgenpt:
From SDL_video.h, SDL_GetWindowSurface(): `You may not combine this with 3D or the rendering API on this window.`

jorgenpt:
This is probably legacy code predating SDL_Render. We should use `SDL_RenderReadPixels` instead:

    /** 
     *  \brief Read pixels from the current rendering target.
     *
     *  \param renderer The renderer from which pixels should be read.
     *  \param rect   A pointer to the rectangle to read, or NULL for the entire
     *                render target.
     *  \param format The desired format of the pixel data, or 0 to use the format
     *                of the rendering target
     *  \param pixels A pointer to be filled in with the pixel data
     *  \param pitch  The pitch of the pixels parameter.
     *
     *  \return 0 on success, or -1 if pixel reading is not supported.
     *
     *  \warning This is a very slow operation, and should not be used frequently.
     */
    extern DECLSPEC int SDLCALL SDL_RenderReadPixels(SDL_Renderer * renderer,
                                                     const SDL_Rect * rect,
                                                     Uint32 format,
                                                     void *pixels, int pitch);

pwaller:
Could we merge this even if it doesn't do the ideal thing? It seems switching to RenderReadPixels is a bigger task that deserves its own issue/pull request.

I don't have time to take a crack at this any time soon and it's cluttering my pull request queue :(

TheCycoONE:
Sorry @pwaller I don't see much value in merging this as is. If there was an associated issue we could copy jorgenpt's comments there and close this PR so someone else could take it.

jorgenpt:
Sorry for the long turnaround. I submitted a pull request now. We were already using SDL_RenderReadPixels, but we were trying to retrieve the window surface to get the format to pass to RenderReadPixels. This was not necessary -- we can just specify our own pixel format.

Issue #629, opened 2015-01-23
setClipRect on the return value of getClipRect not always a no-op.


jorgenpt:
Looking at a bug reported by @pwaller, I noticed an interesting fact:

    THClipRect rc = {};
    pCanvas->getClipRect(&rc) 
    pCanvas->setClipRect(&rc);

Is not guaranteed to return the clip rect to the original state. I don't know if we ever render something without a clip rect (we might not), but if we have clipping disabled, then getClipRect returns an empty clip rect (0x0 size.) setClipRect() has special casing for empty cliprects, converting them to out-of-bounds clipping, so that nothing is ever rendered.

This is because SDL and CTH has different notions of what an empty clip-rect means: We say it's clip *every* pixel, SDL says it's clip *none* of the pixels. The right solution might be to have THClipRect  contain a "clip_enabled" field, and use that to convert rects to & from SDL convention.

pwaller:
So my attempt at fixing this doesn't solve the bug I was complaining about (graphics cease updating after some minutes of gameplay).

However, disabling the clipping entirely does fix the bug!

pwaller:
@jorgenpt so we have to solve this mystery, why doesn't #629 fix the problem, but disabling clipping entirely does?

Andy51:
@pwaller after some research i know why your fix is not correct:
it only handles
```c++
THClipRect rc = {};
pCanvas->getClipRect(&rc);
pCanvas->setClipRect(&rc);
```
case, but not the
```lua
local x_, y_, w, h = canvas:getClip()
canvas:setClip(x_, y_, w, h)
```
so in order to have correct mapping of SDL behavior to TH expected behavior, we have to return nonempty clip rect when clipping disabled, like canvas size.

jorgenpt:
@Andy51, nice catch!

PR #631, opened 2015-01-23
Fix issue #629 (first attempt)

pwaller:
This is an attempt at fixing #629. I don't have very high confidence in it, but it doesn't seem to obviously break anything.

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -397,7 +397,10 @@ bool THRenderTarget::fillRect(uint32_t iColour, int iX, int iY, int iW, int iH)
 void THRenderTarget::getClipRect(THClipRect* pRect) const
 {
     SDL_RenderGetClipRect(m_pRenderer, reinterpret_cast<SDL_Rect*>(pRect));
-    if (pRect->w == 0 || pRect->h == 0) {
+    
+    const SDL_Rect *pSDLRect = reinterpret_cast<const SDL_Rect*>(pRect);
+    
+    if (SDL_RectEmpty(pSDLRect)) {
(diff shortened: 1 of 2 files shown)


TheCycoONE:
This seems pretty much what @jorgenpt was suggesting on irc.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -397,17 +397,26 @@ bool THRenderTarget::fillRect(uint32_t iColour, int iX, int iY, int iW, int iH)
 void THRenderTarget::getClipRect(THClipRect* pRect) const
 {
     SDL_RenderGetClipRect(m_pRenderer, reinterpret_cast<SDL_Rect*>(pRect));
+    if (pRect->w == 0 || pRect->h == 0) {
```
jorgenpt: I would create a temp `pSDLRect`, pass that to `SDL_RenderGetClipRect`, and then check  `SDL_RectEmpty(pSDLRect)`. That way you're consistent with `THRenderTarget::setClipRect`. It should be exactly the same check under the hood, though.

jorgenpt:
Other than that one in-line nitpick, LGTM (Looks Good To Me).

pwaller:
I've made the change whilst from my mobile connection through the github UI, so I haven't even tested if it compiles, but I think it's straightforward enough. PTAL @jorgenpt

TheCycoONE:
Looks alright to me.  All these reinterpret_casts make me a bit nervous but I know they predate this patch.  :+1:

TheCycoONE:
I believe this pr is superseded by #656.  Thanks @pwaller.

pwaller:
Great to see it fixed! Good work @Andy51

Issue #370, opened 2014-11-09
Zoom related graphical glitches [SDL2]


sadger:
I have been testing some custom graphics and both with and without the custom graphics loaded the whole SDL2 version is very glitchy. Picking up objects or zooming in can make the rendering mess up. Often it seems it is displaying sprites that have been rendered in the past so you get duplicates all over the screen.

Tested on ArchLinux 64bit. 

An screenshot to give you an idea of what happens quite often:
![glitchy](http://i.imgur.com/APm32sv.png)

Notice the repeated receptionist and the filing cabinet in the GP.

sadger:
This is still an issue in the latest master (6316114) since sdl2 has been merged. Makes the game unplayable for me. I'll see if I can track down what is causing it. 

Video of the offending glitches:
https://www.youtube.com/watch?v=ny7aVIUxBno

Lego3:
So far I haven't seen this type of issue when running on Windows. What I have seen though is that it sometimes can show a black frame.

pwaller:
I've hit it too. Linux Ubuntu 14.04 x64 d61335421a5c0a9e3121c868e033d5f33279296b.

sadger:
I can confirm that it is the zooming for me causing these glitches. The offending commit for me is commit 476df8b. adding zoom support to sdl2. I tried the fix by @pwaller  but didn't fix any of the glitches for me. But checking out the commit prior to 476df8b I have no graphical issues.

Andy51:
Some additional info on various zoom glitches.

1. Do some zooming
2. Start placing a room -> Bottom GUI panel disappears
3. Place the door and press "Confirm" -> The whole screen blinks black
4. Press "Cancel" -> The room still displayed in furnishing mode, ground cursor gets duplicated
![screenshot8](https://cloud.githubusercontent.com/assets/4835002/5989424/cdca693a-a998-11e4-84ac-fb0c1c1af406.jpg)

And in fullscreen mode additional to listed above:
1. Do some zooming -> bottom GUI panel jumps to the center of the screen
![screenshot6](https://cloud.githubusercontent.com/assets/4835002/5989443/6ad1e7d0-a999-11e4-8fb4-17c9d97d5790.jpg)
2. Start placing a room -> notice the ground cursor position vs mouse cursor:
![screenshot5](https://cloud.githubusercontent.com/assets/4835002/5989449/93f5daea-a999-11e4-8c90-67f445ec5ebe.jpg)

[...]

PR #632, opened 2015-01-23
Fix issue #370

pwaller:
This causes the most of (but not all, because I think I didn't find all of the cases where this happens yet) the symptoms of #370 to go away.

It's at the expense of making the scale of some objects wrong, but at least rendering doesn't completely break.

This is not a complete fix, since it introduces another bug.
--

I'm recording this here quickly so that what I learned after a small debugging session isn't lost.

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -933,12 +933,12 @@ function UIEditRoom:draw(canvas, ...)
     local ui = self.ui
     local x, y = ui:WorldToScreen(self.mouse_cell_x, self.mouse_cell_y)
     local zoom = self.ui.zoom_factor
-    if canvas:scale(zoom) then
-      x = x / zoom
-      y = y / zoom
-    end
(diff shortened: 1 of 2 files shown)


pwaller:
@sadger, you said it doesn't work for you on #370, let's bring that conversation over here.

This change doesn't actually fix all cases. Does the patch improve things for you even a little bit? Under what circumstances do you hit the zoom issue?

The main remaining problem that I'm aware of is when building staff.

Andy51:
I can confirm that most of the bugs went away with you patch applied, but not with fullscreen mode.
And it introduces one more bug not mentioned before - there is no ground cursor when starting to place a room

pwaller:
Interesting.

sadger:
I checked out your branch and I still get most of them. Mostly picking staff/objects up after zooming makes them glitch all over the screen. And graphics are tearing in some places too.

TheCycoONE:
@pwaller  Thanks for identifying that scaling the cursor sprites is (at least part of) the problem.  I don't think we need to keep this pull request open.  It will still be linked to the ticket for illustration.

PR #647, opened 2015-01-31
2 enhancements: dynamic info during pause and spacebar to center view

xuancong84:
1.
a small change: when paused, the bottom panel will still display object info if mouse is moved over an object, previously it only displays the
tag "PAUSED" so that you cannot actual play the game when paused

2.
added a new function: when there is an announcement such as "doctor required in GP office", the player can press <tab> to center view or press <spacebar> to scroll to view, so that he does not need to check one-by-one which GP office is missing a doctor when there are many GP offices


Lego3:
For the pause thing: The game is not meant to be playable while paused (a kind of difficulty mechanism). However, there is an advanced setting in the options menu that let's you tick "play while paused" to on. When on we should indeed show info in the dynamic bar.

I would suggest to create two pull requests, one for each change.

xuancong84:
I don't think that is efficient as I have made further commits. Maybe when @CycoOne merging my branch, you add an additional condition that the allow-edit-during-pause option must also be enabled as well. 

Cheers,
Xuancong

TheCycoONE:
@xuancong84 this is why it's important to make your changes in different branches.

xuancong84:
I see. I think things are a bit messed up because the CorsixTH/CorsixTH
master branch is now full of dirt:
1. the clip bug
2. background music speed sometimes faster
3. screen flashed black and white if you zoom during right-click picked
4. ghost images: so many humanoids while right-click picked

I have created branch 0.40 and 0.45 in my fork. When the time comes, you
just watch the difference between 0.40 and 0.45 while patching up the
master.

xc

xuancong84:
hi Stephen,

It seems that both the "scale" function and "setClip" function in SDL has
problems, they cause blackout and flicking during rendering. Or should we
fall back to the old SDL?

xc

TheCycoONE:
@xuancong84 I think you responded to the wrong thread?

TheCycoONE:
@xuancong84 there are too many different unrelated commits in this pull request.  Some of them look interesting, but nothing can be merged as is.

I'm going to close this pr, and you can open new ones for any individual fixes that are still relevant.

PR #652, opened 2015-01-31
update issue tracker URL

ledowong:


Lego3:
Good catch. :-) Unfortunately the main reason that we didn't link directly to the issue tracker in the first place is that the original in-game font doesn't contain the forward slash or colon characters. I did however update the th-issues.corsix.org link to point to github instead. Thank you for finding it!

PR #661, opened 2015-02-02
Fix busted --lpath

TheCycoONE:


TheCycoONE:
Still appears to be failing... I'm going to be running some tests in this pr so don't merge it until I say I'm done

TheCycoONE:
@Lego3 Busted is working with this change, seems like the version in TravisCI doesn't like --lpath?  Should probably be merged very soon.

Andy51:
I can only suspect it stopped working because of different busted andor luafilesystem compared to initial build in PR:
luafilesystem 1.6.2-2 -> 1.6.3-1
busted 2.0.rc3-0 -> 2.0.rc5-0

Issue #672, opened 2015-02-06
Typo in .busted file


TheCycoONE:
The busted devs looked at our code when I filed the bug about --lpath not working.  It appears that there is a typo in the path in one of our .busted files.

See https://github.com/Olivine-Labs/busted/issues/327

PR #675, opened 2015-02-07
Fixed path separator in Luatest/.busted file.

Alberth289346:
Fixes issue #672


Lego3:
Luarocks still down apparently... This change is easy enough to test anyway. :-)

PR #676, opened 2015-02-07
[RFC|RDY] Custom campaigns

Lego3:
I am aware that this pull request is not ready for merging. I just wanted to start getting feedback on the code, as parts of it has been lying around for over a year now.

Especially the C++ part of the code probably needs some improvements, please advice on what it should look like.


Lego3:
Started fixing comments from TheCycoONE. Still todo is m_ and most languages following the update of how introduction texts are parsed.

TheCycoONE:
Checked the TravisCI error.  Its probably related to dangling white space.

Lego3:
The skylink download mirror on SourceForge appears to be down or similar.

TheCycoONE:
Yes that mirror seems to have been down for a few days.  I haven't gotten around to changing it yet

Lego3:
What does RFC stand for? I'm thinking of trying to finish this in the coming month or so. What would be most important in your opinion? Should I split it into many pull requests?

TheCycoONE:
It stands for request for comments, though I guess this could have been wip (work in progress) as well.
The reason for the prefixes is because only owners and contributors can set labels. With the prefixes people that make the pull requests can indicate their state.
[WIP] - work in progress i.e. don't review this, I'm still working on it.
[RFC] - request for comment i.e. this isn't ready to merge, but it would be good if people could give feedback
[RDY] - ready i.e. I feel this is ready to merge. Review it and merge if you agree.

TheCycoONE:
Probably no need to split it into separate pull requests.

Lego3:
Okay, so I rebased onto latest master now. What do you guys think? It is still the C++ side I'm most unsure about whether it's good enough since I'm not really a C++ programmer...

TheCycoONE:
Alright, I've skimmed it all. There are a few problems with your last commit as mentioned above (am I right to guess you have some Java background?). The rest of it reads well.

It would be good to make a note about the LevelFile -> MapFile change. I agree with it but it will break all the other custom levels out there, maybe we could support both for legacy? I'll let you make the call on that.

Lego3:
This looks much neater indeed. :-) Does it look like it will work?

Lego3:
Excellent comments! Since I suspect that they will disappear when I push a new version I'll answer them here first. :-)

Stephen: I changed to "= {}" and fixed the checkinteger in the correct commit. When it comes to reinterpret_cast I basically just copied from another place it was used, so what would the line look like with a static_cast? I don't really know what really happens there... :-P

Alberth: Added doxygen and Lua comments in (I hope) all places where I added the thing, used std::max (needed to include <algorithm>, that's ok right?) Also did the various small changes you mention except "colors" and "to -> at": for colors I just did like it has been done before (with that said I really would like to refactor the whole color definition mess...) so I'll leave it for now, and I believe it should be "to"? "To the right of something?" It's not at the right inside the list, but rather to the right of the list.

Lego3:
Hm, I'm not entirely sure what is the cause of the failed Travis build?

TheCycoONE:
Looks like LDocGen choked on something, but I'm not sure what.

As for ```reinterpret_cast```, just replace that text with ```static_cast``` directly. A ```reinterpret_cast``` ignores compile and run time checks and basically just points a new type at the old type's memory. This is dangerous because it will never fail and could quite likely produce garbage, e.g. ```reinterpret_cast<int>(double_value)```. A ```static_cast``` is checked at compile time and converts one type to another.  There's also dynamic_cast, which does a run time check on the type for conversion (useful in some cases of polymorphism when the conversion cannot be checked at compile time), and ```const_cast```, which is also compile time checked and only adds or removes a 'const' constraint.

```static_cast``` is almost always what you want when converting one type to another.

Alberth289346:
````patch
diff --git a/CorsixTH/Lua/world.lua b/CorsixTH/Lua/world.lua
index ef0304f..9839c59 100644
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1497,7 +1497,7 @@ function World:checkWinningConditions(player_no)
 end
 
 --! Process that the given player number won the game.
---!param (integer) player_no Number of the player who just won.
+--!param player_no (integer) Number of the player who just won.
 function World:winGame(player_no)
   if player_no == 1 then -- Player won. TODO: Needs to be changed for multiplayer
     local text = {}
````

No time for other things now

Alberth289346:
just small things, looking great!

Lego3:
Fixed the things Alberth mentioned. For some reason the diff --check command still didn't show anything for me, but searching for tabs I found some I probably introduced. I hope I found them all anyway.

Lego3:
Ok, so the only thing left is that reinterpret_cast thing. But I can't find out what to write instead...

TheCycoONE:
@Lego3 Just replace the word ```reinterpret``` with the word ```static```

Lego3:
That doesn't compile for me.

TheCycoONE:
Ok, I'll take a closer look.

TheCycoONE:
I believe what you want here is ```iNextChar = unicodeToCodepage437(static_cast<unsigned char>(*s));```

This whole iNextChar is a little hacky, because it only interprets the next 8 bits, not the entire next character. In reality this works because they are checking the result against ascii characters so any interesting value will always be 0-127 here (because that's how UTF-8 encoding works).

TheCycoONE:
Also I see there are a ton of unnecessary reinterpret_casts throughout this file in particular and the rest of the source in general. I'll work on cleaning them up at some point.

Lego3:
Yes, that is what I copied... I'll try that out tomorrow.

Lego3:
Seems to work fine. :-) Is this getting ready?

TheCycoONE:
Yes, @Alberth289346 has given the :+1: and I just want to review the font and iNextChar stuff a little more closely. I've been a little busy with my Travis CI changes (trying to get better C++ x11 support), but should be able to finish my review within the next couple weeks.

Lego3:
That whole file really needs some refactoring, and I didn't really manage to make it much better... I was thinking that might be the next thing I do to learn more C++ at the same time.

TheCycoONE:
The business with iNextChar in particular is a bit troubling. It's converting the first byte from utf-8 to cp437 then comparing the char code against the value of certain characters in compiler's code page (most likely cp1257 or utf-8, possibly ascii)

It seems to me that this section on the coincidence that utf-8 and most code pages are (at least mostly) a superset of ASCII. The conversion to cp437 here is not helpful.

Issue #450, opened 2014-12-23
Changing to High Resolution Leaves Blank Screen (in DirectX)


TheCycoONE:
_From [chris_lo...@hotmail.com](https://code.google.com/u/101094553588000836480/) on July 01, 2010 09:47:02_

What steps will reproduce the problem? 1. Change the resolution to 1920x1200 What is the expected output? What do you see instead? The resolution changes to 1920x1200 however all that is visible is the mouse cursor on a completly black screen.

If you were to alt+F4 out of the game, and re-launch it works fine in the new resolution.

What version of CorsixTH are you using (e.g. "Beta 2", "SVN r123 ")?
Beta3 What operating system / compile settings are you using? Windows 7 64-bit

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=350_

Lego3:
I cannot reproduce this error now using sdl2. Please try it out. I do however get a message in the log each time I switch: "THMovie overlay should be deallocated before being allocated!"

MarkL1961:
Ok for me

Lego3:
Do you get the message in the log though?

MarkL1961:
I see nothing at all in the console, which is a little odd, there is usually something there - even if its only about not checking for updates.

I did notice that changing to full screen put the menu/button bar a third of the way up the screen when it normally stays at the bottom.

MarkL1961:
![menu bar](https://cloud.githubusercontent.com/assets/4965487/5747853/45b67538-9c33-11e4-8ace-f74f642531cd.png)

TheCycoONE:
I also get the "THMovie overlay should be deallocated before being allocated!", ever since an earlier change @jorgenpt made to fix changing resolution / full screen.  Forgot about it.

jorgenpt:
@TheCycoONE, @MarkL1961: This seems resolved, since @MarkL1961 used to have the problem. Deallocated / allocated issue is benign and unrelated (and fixed in #678.)

Close this?

TheCycoONE:
I think so, probably specific to the old directx renderer and not the sdl2 directx path.

PR #678, opened 2015-02-07
Fix THMovie overlay spew

jorgenpt:
Addresses issue mentioned in the discussion of #450.


TheCycoONE:
Good catch.  Works well for me.

PR #686, opened 2015-02-10
Add lpeg.dll, lfs.dll and all luasocket files to travis builds

TheCycoONE:
Sadly the makefiles for luafilesytem and lpeg didn't consider cross compiling with mingw so I had to do the compiling myself.  luasockets did have mingw support in it's makefile so I took advantage of it (which is helpful because it has a much more involved compile process).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,24 @@ before_install:
   - tar xJf lua-5.1.4-4-mingw32-dll-51.tar.xz -C lua
   - wget http://netcologne.dl.sourceforge.net/project/mingw/MinGW/Extension/lua/lua-5.1.4-4/lua-5.1.4-4-mingw32-dev.tar.xz
   - tar xJf lua-5.1.4-4-mingw32-dev.tar.xz -C lua
+  # Get LuaFileSystem
+  - wget http://github.com/keplerproject/luafilesystem/archive/v_1_6_3.tar.gz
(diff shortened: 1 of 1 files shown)


MarkL1961:
The latest build I can see is CTH-20150208-master-fa0a7775b5.zip     from   08-Feb-2015 20:39  
Is this because this needs merging first?  I had thought we got a build from travis as soon as the testing was done.
I'd jump in and click on merge, but as I can't test it until I have a build I don't like to interfere...

TheCycoONE:
@alanwoolley do pull request builds end up somewhere else, or are they not saved.  I noticed this as well as I had planned to download the build when it was finished and make sure it was all working correctly.

alanwoolley:
Only pull requests from the project. It's for security reasons really, so whilst I could probably work around it, I don't particularly want to.

MarkL1961:
If for some odd reason it does not work it is easy enough to reverse?  @TheCycoONE

TheCycoONE:
It would be relatively easy (if a little messy) to reverse or repair after merging yes.  Another possibility would be temporarily modifying the pull request to push the file somewhere else, though I don't have any servers set up for that at the moment.

MarkL1961:
I'll keep my fingers crossed then :)

MarkL1961:
:+1:

TheCycoONE:
Spiffy

PR #687, opened 2015-02-11
Fix #370 (zoom glitches on some platforms)

### LINKED ISSUE #370: Zoom related graphical glitches [SDL2]

pwaller:
The `m_pZoomTexture` was being used as a render target even though
it was uninitialized. This sometimes worked out, depending on your
platform, and other times not.

The fix is to initialize the zoom buffer to transparent and then use
SDL_BLENDMODE_BLEND when it is time to flush the zoom buffer.

Fixes #370

I've played the game briefly and this resolves all of the graphical glitching-making-it-unplayable issues I know of right now.

The only potential downside is that due to the alpha blending, some of the cursors may not look quite pixel perfect. Noticably the red square used to highlight tiles doesn't have quite the right perceived colour (compare it with the red build blueprint):
 
screenshot from 2015-02-11 02 12 00

compared with the usual unzoomed:

screenshot from 2015-02-11 02 14 32

.. but this is preferable to the graphics completely breaking :)

--

After spending too much time staring at an OpenGL API trace with [vogl](https://github.com/ValveSoftware/vogl), it's apparent that an uninitialized texture ends up being the render target. Sometimes this coincidentally contains "okay-ish" textures, and other times it contains unaligned junk (whatever happens to be in texture memory).

[I suspect offending code is this `SDL_CreateTexture`](https://github.com/CorsixTH/CorsixTH/blob/c384754bad1fdcb1d9a258e924ebb13fcc8ab61b/CorsixTH/Src/th_gfx_sdl.cpp#L300-L307), where a texture is created and then immediately used as render target.

I've been struggling to think of a decent fix, my first thought was to blank the texture with `fillBlack()`, which works fine until you zoom whilst constructing and then it breaks (everything is black except the cursor). I then tried copying various buffers around but obviously this was slow (and difficult to get correctly aligned). So I went with the alpha-blend trick.

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -303,6 +303,7 @@ bool THRenderTarget::setScaleFactor(float fScale, THScaledItems eWhatToScale)
                                            virtWidth,
                                            virtHeight
                                           );
+
         SDL_RenderSetLogicalSize(m_pRenderer, virtWidth, virtHeight);
         if(SDL_SetRenderTarget(m_pRenderer, m_pZoomTexture) != 0)
         {
@@ -310,6 +311,10 @@ bool THRenderTarget::setScaleFactor(float fScale, THScaledItems eWhatToScale)
             m_pZoomTexture = NULL;
             return false;
         }
+
+        // Clear the new texture to transparent/black.
+        SDL_SetRenderDrawColor(m_pRenderer, 0, 0, 0, SDL_ALPHA_TRANSPARENT);
+        SDL_RenderClear(m_pRenderer);
 
         return true;
     }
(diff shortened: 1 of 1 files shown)


TheCycoONE:
Excellent deduction, thanks @pwaller.  I will try it myself later, but maybe you can answer now - is the reason for the darker red because of the scaling and pixel bleed from the black background, or because the colour is being blended with something on the RenderCopy?  Or as a consequence, is the whole doctor miscoloured on pickup or just the edges?  (Or does this only affect editing rooms for some reason?)

pwaller:
I don't know. I figured it out and then called it a day pretty quickly.

I observe that a lifted doctor seems to look (perception-wise) pixel perfect when you drop him. It's just the red square where I notice any difference.

sadger:
For me, it fixes the issue which is great! Nice work @pwaller

TheCycoONE:
Works great for me too.  Excellent work!

Andy51:
Fixes the issues for me too, gj @pwaller !

Color issues are definitely because of blending, you can check it by zooming in maximum, but i believe we would find how to workaround this eventually.

PR #688, opened 2015-02-12
[WIP] New debug menu & global key command: Save & Restart CorsixTH

J-Shep:
If there's a running game its saved as RestartSave.sav and then the
existing App:resetApp() function is called. RestartSave.sav will be
deleted when it has been successfully loaded.

I've removed the old SHIFT + F10 global key command for calling
resetApp() directly.

New functions:
- ui.lua:saveAndRestart() : key command : CTRL + R (When debug mode is enabled)

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -29,7 +29,7 @@ local assert, io, type, dofile, loadfile, pcall, tonumber, print, setmetatable
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 101
+local SAVEGAME_VERSION = 102
 
 class "App"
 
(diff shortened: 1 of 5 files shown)


J-Shep:
I will improve this commit later:
1) I will make RestartSave.sav being auto loaded after CorsixTH has been restarted.
2) When CorsixTH is restarted to view the new layout of a pre-game dialog I will make it so that this dialog will be auto displayed after CorsixTH has been restarted.

Lego3:
I personally like to have the Shift + F10 possibility, I use it quite a lot. Maybe we could have both?

TheCycoONE:
I don't think the function deserves two shortcuts, but I'm fine with either option.

Lego3:
It depends how much functionality goes into the new shortcut. You dont always want to reload an open game etc.

J-Shep:
>> It depends how much functionality goes into the new shortcut. You dont always want to reload an open game etc.

@Lego3 I make SHIFT + F10 still call UI:resetApp() directly if you want this key command to be kept.

TheCycoONE:
No update in over 2 years. If there is no update in 2 weeks I will close this PR.

PR #689, opened 2015-02-12
contributing.txt: Asked for commit messages to name modified things

J-Shep:
So that commit message searches can find commits which introduce and
change functions & classes.

I didn't ask for new and changed variables to be mentioned because
having to mention/list all the changed and new variables in a big commit
could be too tedious.

### DIFF
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -1,4 +1,4 @@
-TELL OTHER DEVS WHAT YOU ARE WORKING ON
+TELL OTHER DEVS WHAT YOUR WORKING ON
 
 When you start working on a fix/enhancement for an open issue please always
 post a comment in this issue's discussion to tell other devs that you have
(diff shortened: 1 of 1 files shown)


MarkL1961:
YOUR is grammatically incorrect.  Should either be YOU ARE or contracted to YOU'RE

TheCycoONE:
Between git blame, git log -G (or -S), and git bisect I haven't really encountered a problem what commit made any particular change.

Of course the commit messages should be informative, and if it changes the behavior of a public function that should definitely be noted.  In cases where the change is to a private function or changes to a public function where the behaviour isn't changed, often the change and the reason for the change is more important than where - particularly if it's broad, like the whitespaces fixes were, or if someone were to clean up all the signed to unsigned comparisons.

J-Shep:
@MarkL1961 @TheCycoONE Thanks for reviewing my pull request. I've decided to close it because I now think the changes I've made should not be pulled.

PR #693, opened 2015-02-17
Disable recolouring if the image has no palette to prevent a crash

Alberth289346:
Fixes recolouring crashes in 32bpp due to trying to do palette recolouring (which makes no sense in 32bpp).

Added new recolour flags for greyscale conversion (non-placable objects), and red/blue colour swapping (bluprints turning red). The latter doesn't really work well, and needs further work, but it needs proper graphics first.

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -37,11 +37,12 @@ function UIEditRoom:UIEditRoom(ui, room_type)
   self:addKeyHandler("return", self.confirm) -- UIPlaceObjects does not need this
 
   local app = ui.app
+  local blue_red_swap = self.anims.Alt32_BlueRedSwap
   -- Set alt palette on wall blueprint to make it red
-  self.anims:setAnimationGhostPalette(124, app.gfx:loadGhost("QData", "Ghost1.dat", 6))
+  self.anims:setAnimationGhostPalette(124, app.gfx:loadGhost("QData", "Ghost1.dat", 6), blue_red_swap)
   -- Set on door and window blueprints too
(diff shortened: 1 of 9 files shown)


TheCycoONE:
Commit message for first commit isn't encompassing of what that commit does.  It really adds an alternative mechanism for recolouring 32bpp graphics (rather than simply disabling attempted pallet swaps)

Alberth289346:
fixed (both uint32_t and the commit message)

I saw more types that can be changed, but resisted the temptation to change them now :p

 :palm_tree: just for fun

Alberth289346:
fixed both issues

TheCycoONE:
I would prefer i the THDF_Alt32_* parameters were only defined once in lua.  The convention seems to be to make a table of properties and use lowercase naming, something like:

```
alternate_32bit_pallets = {
  GreyScale = 1 * 32,
  RedBlueSwap = 2 * 32,
}
```

Ideally I think that table would be attached to TH.anims so that it would be available everywhere you might want it, but it would likely be easier and clean enough to put it in ui.lua.  (I'm not aware of anywhere we currently store a table of constants to access from lua with a luaT_class, so the infrastructure likely isn't in place to do it simply.)

Alberth289346:
Added a new 'luaT_setconstant' macro the the lua/C interface code, and added the two values as constants in the 'anims' class.

TheCycoONE:
:+1: Code looks good to me.

TheCycoONE:
Close/reopen to trigger travis rebuild.

TheCycoONE:
Ah right, the rebuild won't succeed until it's rebased on the latest master.

Alberth289346:
Rebase :sparkles: seems to build :ok_hand:

sadger:
Tested it and confirmed as working.

Issue #582, opened 2014-12-26
Windows Installer: Remember last install location


mpmc:
Hey folks,
This is pretty much as the title but..

I've noticed that when I install a new release the installer uses the default program files location, would it be possible to have the installer remember the previous install location?

Thanks.

SimplyRyan:
This has been added #698

SimplyRyan:
@TheCycoONE this can be closed as #698 resolved this.

PR #698, opened 2015-02-20
Windows Installer: Check if already installed

SimplyRyan:
Installer update:

Also resolves issue #582

- Updated welcome.bmp image, simple screenshot from the game looks a
little cleaner (see attached preview).

- Added a check to see if the game is already installed, this is done
via a registry read which pulls out the uninstall path, parses it and
uses that as the currently installed path. The installer then proceeds
with this path for the install. If this data isn't found in the registry
the default directory is used. Tested on Windows 8.0 x64 and Windows 7
x64(TheCycoONE), compiled clean on MakeNSIS v3.0b1.

- Small cleanup of the .nsi file for readability and to remove some old
unused code.


WindowsInstaller/Win32Script.nsi
```diff
@@ -206,9 +225,11 @@ Section "MainSection" SEC01
   ${If} $CONFIGAPPDATA == 1
     SetOutPath "$APPDATA\CorsixTH"
     IfFileExists "$APPDATA\CorsixTH\Saves" saves
+	
     ; Don't overwrite previous installations though!
     CreateDirectory "$APPDATA\CorsixTH\Saves"
   ${Else}
+
```
TheCycoONE: Can get rid of this one too?

Otherwise perfect.

TheCycoONE:
Merged.

PR #703, opened 2015-02-25
Lua 5.2 and 5.3 compatibility

TheCycoONE:
With this pull request the game is playable in Lua 5.2, and is able to load all of my save games.  Save games should be fully compatible back and forth between Lua 5.1 and Lua 5.2.

The risk of impacting Lua 5.1 is pretty small.

### DIFF
--- a/CMake/FindLua.cmake
+++ b/CMake/FindLua.cmake
@@ -22,12 +22,12 @@ SET (LUA_INTERPRETER_TYPE "")
  
 IF (WITH_LUAJIT)
     SET (LUA_INTERPRETER_TYPE "LuaJIT")
-    SET (LUA_LIBRARY_NAME luajit-5.1)
+    SET (LUA_LIBRARY_NAME luajit-5.1 lua51)
     SET (LUA_INCLUDE_DIRS include/luajit-2.0 include)
 ELSE (WITH_LUAJIT)
     SET (LUA_INTERPRETER_TYPE "Lua")
-    SET (LUA_LIBRARY_NAME lua5.1 lua51 lua lua-5.1)
-    SET (LUA_INCLUDE_DIRS include/lua5.1 include/lua51 include/lua include/lua-5.1 include)
+    SET (LUA_LIBRARY_NAME lua53 lua52 lua5.1 lua51 lua lua-5.1 liblua)
+    SET (LUA_INCLUDE_DIRS include include/lua include/lua5.1 include/lua51 include/lua-5.1)
 ENDIF(WITH_LUAJIT)
  
 FIND_PATH (LUA_INCLUDE_DIR lua.h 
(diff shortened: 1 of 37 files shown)


TheCycoONE:
1. Command line arguments are actually functioning, I was simply lacking luasocket in that build folder.
2. I believe that th_getfenv is actually functioning.  I printed the keys to the tables returned in lua 5.2 and lua 5.1 and they are identical.  That suggests the problem is somewhere else in persistence.

TheCycoONE:
I managed to get some help from #lua on freenode.  At least in the case of music it's because we're trying to set the metatable to the environindex, which in lua 5.2 we mapped to the first upvalue, but functions created with lua_setfuncs(state, reg, 0) have no upvalues.

TheCycoONE:
Re-based on the latest master, and added casts to avoid undoing the work done by @Alberth289346

TheCycoONE:
Travis failure was due to sourceforge downloads again.  Will close and open to trigger rebuild.

alanwoolley:
@TheCycoONE You don't need to close/open. If you navigate to the failing build on Travis CI, then you should find a 'restart build' button on the right. It'll pull the commit in again and retry the build. You might need to login with your GitHub account, I'm not sure.

TheCycoONE:
@alanwoolley Noted, sorry for the noise.

Alberth289346:
Ran a small test by building a GP office, and noticed the tile cursor jumping around in a weird way. See the attached image. The needle cursor move parallel to the wall of the GP office.

jumping tile cursor

TheCycoONE:
Last commit fixes the issue @Alberth289346 mentioned above.  When we are ready to merge it should be rebased into b42445b

Alberth289346:
Found another issue, if you change zoom with the mouse wheel, the world moves a tile toward the bottom right. It does not matter where the cursor is in the screen, or whether you zoom in or out.

For best effect, turn the wheel slowly so the program can keep up and sees it as a sequence of small changes.

TheCycoONE:
Removes a few more useless math.floor statements and fixes the zoom issue @Alberth289346 found.  I had really gone overboard with the math.floor for lua 5.3 compatibility.

The next two commits fix a conversion warning introduced in the new debugging functions.

Alberth289346:
Crash when clicking the 'research' button:
./CorsixTH/Lua/dialogs/fullscreen/research_policy.lua:113: attempt to call field 'gfind' (a nil value)

Doesn't exist any more in lua 5.2 see http://lua-users.org/lists/lua-l/2013-04/msg00117.html
Seems it's called 'gmatch' now (also in that link). 'ack' says there is only 1 gfind in the program.

TheCycoONE:
Fixed research_policy, replacing gfind with gmatch.  Thanks @Alberth289346

Alberth289346:
Line 405 of CorsixTH/Src/th_lua.cpp, below the line
printf("boolean %s\n", lua_toboolean(L, i) ? "true" : "false");

has a TAB character in the line.
Otherwise, it looks fine to me (looked a the combined diff from the master branching point)

TheCycoONE:
Fixed whitespace error pointed out by @Alberth289346 and rebased on lasted master, squashing all fix commits.

PR #705, opened 2015-02-27
Do not use pkg-config to find FFmpeg/Libav on Unix

ezeenova:
This is necessary for CMake to find FFmpeg/Libav in a cloned CorsixTH/deps repo on GNU/Linux. As a bonus, it now uses the same method to find includes/libraries on all platforms.

I have only tested this on GNU/Linux. I haven't touched any of the Windows stuff so that should work normally, and it will probably work on OSX too. Could definitely use some testing on various platforms though.

### DIFF
--- a/CMake/CMakeFFmpegLibavMacros.cmake
+++ b/CMake/CMakeFFmpegLibavMacros.cmake
@@ -1,0 +1,65 @@
+#-------------------------------------------------------------------
+# FindFFmpeg.cmake and FindLibAV.cmake are dependent on this file.
+#
+# Redistribution and use is allowed according to the terms of the BSD license.
+# For details see the accompanying COPYING-CMAKE-SCRIPTS file
+#
+#-------------------------------------------------------------------
+
+### Macro: set_component_found
+#
+# Marks the given component as found if both *_LIBRARIES AND *_INCLUDE_DIRS is present.
+#
+macro(set_component_found _component )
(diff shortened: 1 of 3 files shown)


TheCycoONE:
Will close/reopen to trigger travis rebuild.

TheCycoONE:
I like it.

Online you were talking about duplication between FindFFMPEG and FindLibAV and I said that I wanted to keep them as separate modules.  Having said that if you want to move the find_component macro into it's own file and include it in both of those Find* files that would be alright.  (Maybe with a less generic name e.g. CMakeFfmpegLib and find_ffmpeg_component).  If you don't feel like doing that I'm good with merging as is.

ezeenova:
Caveat: If the user has Libav installed on their system (in /usr/lib/ and usr/include), and additionally downloads CorsixTH/deps and runs

   cmake -DCMAKE_PREFIX_PATH=pathtodeps -DWITH_LIBAV

CMake will use ffmpeg from deps instead of libav from /usr (except for libavrescale, obviously). Is that acceptable?

TheCycoONE:
@ezeenova as far as I know it's unavoidable.  Despite the libraries being incompatible both projects continued to use the same naming and version scheme. It won't compile likely.  Maybe we can print a warning if the include directories are not all the same or something - otherwise I think there's some things the compiler will just have to watch out for.

PR #710, opened 2015-03-01
Improve expected size of c++ data, to avoid surprises at some platforms.

Alberth289346:
Basically fixing warnings about signed/unsigned comparisons, followed by changing the type of things that looked like they were wrong. It compiles without warnings for -Wall at my system, not sure how to check all sizes make sense everywhere.

If you like to have fun with complicated bit shuffles, try th_lua_map.cpp :)


Alberth289346:
Fixed the 4 problems in the first commit, and added a second commit with moar type changes

Alberth289346:
commit:f49343e  breaks strings

Alberth289346:
Solved pretty much all warnings. Path finder is coded too silly to even try fixing, lots of warnings from video library, and one or two warnings or so left, where I don't understand what is actually wrong.

I played level 2, and it seems to work.

TheCycoONE:
On MSVC 2013 this brings us down to 14 warnings (from over 40).  Most of those are sign mismatch warnings in th_gfx_font.cpp, e.g. line 791.  If you're not getting them in gcc I have to assume that the freetype types differ by platform or you're building without freetype?

TheCycoONE:
Below is the output from msvc 2013 and freetype 2.5.5 (specifically referring to the signed/unsigned mismatch warnings at the top).

I checked the freetype2 api guide and those fields are indeed unsigned int: http://www.freetype.org/freetype2/docs/reference/ft2-basic_types.html#FT_Bitmap.  This was a change from earlier versions of freetype, e.g. 2.4.10: http://freetype.sourceforge.net/freetype2/docs/reference/ft2-basic_types.html#FT_Bitmap

```
Warning	5	warning C4018: '<' : signed/unsigned mismatch	E:\GitHub\CorsixTH\CorsixTH\Src\th_gfx_font.cpp	756	1	CorsixTH
Warning	6	warning C4018: '<' : signed/unsigned mismatch	E:\GitHub\CorsixTH\CorsixTH\Src\th_gfx_font.cpp	767	1	CorsixTH
Warning	7	warning C4018: '<' : signed/unsigned mismatch	E:\GitHub\CorsixTH\CorsixTH\Src\th_gfx_font.cpp	791	1	CorsixTH
Warning	8	warning C4018: '<' : signed/unsigned mismatch	E:\GitHub\CorsixTH\CorsixTH\Src\th_gfx_font.cpp	801	1	CorsixTH
Warning	4	warning C4267: 'argument' : conversion from 'size_t' to 'FT_Long', possible loss of data	E:\GitHub\CorsixTH\CorsixTH\Src\th_gfx_font.cpp	396	1	CorsixTH
Warning	11	warning C4267: 'argument' : conversion from 'size_t' to 'Uint32', possible loss of data	E:\GitHub\CorsixTH\CorsixTH\Src\th_lua_sound.cpp	242	1	CorsixTH
Warning	10	warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning)	E:\GitHub\CorsixTH\CorsixTH\Src\th_lua_gfx.cpp	659	1	CorsixTH
Warning	9	warning C4800: 'Uint32' : forcing value to bool 'true' or 'false' (performance warning)	E:\GitHub\CorsixTH\CorsixTH\Src\th_gfx_sdl.cpp	293	1	CorsixTH
Warning	12	warning C4996: 'sprintf': This function or variable may be unsafe. Consider using sprintf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	E:\GitHub\CorsixTH\CorsixTH\Src\th_map_overlays.cpp	99	1	CorsixTH
Warning	2	warning C4996: 'std::_Copy_impl': Function call with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct. To disable this warning, use -D_SCL_SECURE_NO_WARNINGS. See documentation on how to use Visual C++ 'Checked Iterators'	C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\xutility	2132	1	CorsixTH
Warning	3	warning C4996: 'std::_Copy_impl': Function call with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct. To disable this warning, use -D_SCL_SECURE_NO_WARNINGS. See documentation on how to use Visual C++ 'Checked Iterators'	C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\xutility	2132	1	CorsixTH
Warning	1	warning C4996: 'std::_Transform1': Function call with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct. To disable this warning, use -D_SCL_SECURE_NO_WARNINGS. See documentation on how to use Visual C++ 'Checked Iterators'	C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\algorithm	1026	1	CorsixTH
Warning	13	warning C4996: 'vsprintf': This function or variable may be unsafe. Consider using vsprintf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.	E:\GitHub\CorsixTH\CorsixTH\Src\th_map_overlays.cpp	184	1	CorsixTH
Warning	14	warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library	E:\GitHub\CorsixTH\build_sdl2\CorsixTH\LINK	CorsixTH
```

TheCycoONE:
Given that freetype changed their api I think we can address those outside of this issue.

Alberth289346:
I added a patch for warnings 4, 11, 10, and 9. This concludes my efforts for this patch, as far as I am concerned.

TheCycoONE:
I notice that if you compile with libav instead of ffmpeg there are some sign compare warnings in th_movies as well, but these can be dealt with at a later time.

Merged.

PR #718, opened 2015-03-14
Fix race condition decoding audio while movie is terminating

TheCycoONE:
The race condition would sometimes cause a segfault or 0xc0000005 error when
closing a movie with audio (such as the intro) early.


TheCycoONE:
Removed: I had put some locking up there and then moved it further down so it probably came from that.

TheCycoONE:
Ops... messed this up.  I'll fix it.

TheCycoONE:
There, fixed.

PR #719, opened 2015-03-15
Fix bad error code -1 from function with size_t (unsigned) return type.

TheCycoONE:
Also enables c++11 mode in gcc/clang which is required for SIZE_MAX.


TheCycoONE:
I see the gcc in travis is pretty old and needs -std=c++0x to compile this.

PR #210, opened 2014-03-04
Generate reference doc for c++ (and java) code

timdiels:
Usage:
cmake .
make doc

Result is in: ./doc

Requirements: DoxyGen

PR #721, opened 2015-03-16
Pr 210 add doc generation (take 2)

Alberth289346:
Adding doc generation seems useful, and the OP timdiels hasn't touched PR #210 for 10 months, so it seems safe to breath some life in it again.

I rebased PR #210 to current master, fixing the conflict in .gitignore. I added a patch to

* Add a doc/index.html for easier browsing
* Stopped doxygen from generating absolute paths
* Reduced the amount of output (before about 980 lines output, now about 20)
* Stopped a few 'obsolete XML_SCHEMA and XML_DTD' warning messages.

Luadoc output doesn't seem to handle the 'files' tab correctly, but that's a different problem imho


TheCycoONE:
The travis ci is responsible for generating the latest master docs and uploading them to @alanwoolley 's site at http://www.armedpineapple.co.uk/cth/docs/. `.travis.yml` will need to be updated accordingly.

Alberth289346:
Travis doesn't actually build the documents (no doxygen?). Also, I changed the scp command to perform a recursive copy, that @alanwoolley may want to check.

TheCycoONE:
I'm thinking no doxygen, it could be added to the sudo apt-get list.

Alberth289346:
Travis refuses to build the documentation, claiming "nothing to do for target "doc" or something like that, currently out of ideas about a cause

TheCycoONE:
Not sure what happened to my line comment... if it shows up later I blame github.  Anyway the folder is ${TRAVIS_BUILD_DIR}\fresh, ${HOME} is not suppose to be used in travis-ci.

Alberth289346:
Thanks for your help TheCycoONE.

Stuff seems to be working now, merged my temporary test fixes into the right place, but kept the improved error reporting that I added while diagnosing the problems as it also fixes a bug in the write call (only lua 5.2 promises that it will return the file handle).

I haven't heard whether the scp command was correct with the 'recursive' flag, and the different target directory (nor was I able to test that).

TheCycoONE:
Generating the docs for lua appears to assume that lua is in the system path, which may not always be the case.  FindLua should be extended to also locate the binary so we can use it instead of hard coding *lua* in the CMakeList.

Error:

```
4>  Generating API documentation for corsixth_lua
4>  'lua' is not recognized as an internal or external command,
4>  operable program or batch file.
```

Alberth289346:
And Travis fails again, as I added 'have lua interpreter' into 'lua found', and apparently, 'lua51' install doesn't add lua in a way that I expect :(

Alberth289346:
Works, use 'cmake -DLUA_PROGRAM_PATH=/path/to/your/lua' to build docs for lua.

TheCycoONE:
With regards to copying the index file to the build dir when make doc is called.  http://www.cmake.org/pipermail/cmake/2009-March/027892.html has an outline of how to do it.

Basically a command is added to the target, which calls cmake with -E copy and the source and destination paths.  cmake's executable is defined as ${CMAKE_COMMAND}.

http://www.cmake.org/cmake/help/v3.2/manual/cmake.1.html#command-line-tool-mode is another handy link.

Alberth289346:
- Added the 'add_dependencies' as part of adding the sub-targets, which is indeed much better.
- I added the duplicate 'if (DOXYGEN_FOUND)' conditions, so each target was independent, but removed them now (now 1 block for adding all doxygen targets).
- doc/index.html is indeed written during 'cmake .' run. Initially based on the idea of having a static-like file although this file is not retrievable from 'git' or by 'make'. I attempted to find a solution for it, but found that it's impossible to output literal text to a file as part of the build itself (ie in a target). As writing a program to generate that text seems a bit overkill, I left it as-is.

alanwoolley:
@Alberth289346 Yes, the recursive flag should work fine.

TheCycoONE:
On windows the LDocGen target still isn't working.

```
E:\Development\LuaJIT-2.0.3\x64\luajit.exe: ../CorsixTH/Lua/strict.lua:66: module 'helpers' not found:
3>  	no field package.preload['helpers']
3>  	no file '.\helpers.lua'
3>  	no file 'E:\Development\LuaJIT-2.0.3\x64\lua\helpers.lua'
3>  	no file 'E:\Development\LuaJIT-2.0.3\x64\lua\helpers\init.lua'
3>  	no file '.\helpers.dll'
3>  	no file 'E:\Development\LuaJIT-2.0.3\x64\helpers.dll'
3>  	no file 'E:\Development\LuaJIT-2.0.3\x64\loadall.dll'
3>  stack traceback:
3>  	[C]: in function 'fn'
3>  	../CorsixTH/Lua/strict.lua:66: in function 'require'
3>  	E:/GitHub/CorsixTH/LDocGen/main.lua:46: in main chunk
3>  	[C]: at 0x7ff71ab22000
```

TheCycoONE:
Also the doc folder that's created in the source folder instead of the build folder.  I assume that's unintentional.

Alberth289346:
It's not ldocgen, it's luajit
<pre>
$ cd doc
$ /usr/bin/lua ../LDocGen/main.lua
$ /usr/bin/luajit ../LDocGen/main.lua 
/usr/bin/luajit: ../CorsixTH/Lua/strict.lua:66: module 'lpeg' not found:
        no field package.preload['lpeg']
        no file './lpeg.lua'
        no file '/usr/share/luajit-2.0.3/lpeg.lua'
        no file '/usr/local/share/lua/5.1/lpeg.lua'
        no file '/usr/local/share/lua/5.1/lpeg/init.lua'
        no file '/usr/share/lua/5.1/lpeg.lua'
        no file '/usr/share/lua/5.1/lpeg/init.lua'
        no file './lpeg.so'
        no file '/usr/local/lib/lua/5.1/lpeg.so'
        no file '/usr/lib64/lua/5.1/lpeg.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
stack traceback:
        [C]: in function 'fn'
        ../CorsixTH/Lua/strict.lua:66: in function 'require'
        ./c_tokenise.lua:21: in main chunk
        [C]: in function 'fn'
        ../CorsixTH/Lua/strict.lua:66: in function 'require'
        ../LDocGen/main.lua:47: in main chunk
        [C]: at 0x004044d0
</pre>
Note that luajit is lying here, if I add ./lpeg.so, it still claims it doesn't exist.

TheCycoONE:
I've solved the LDocGen problems (minimally but a larger fix is probably out of scope for this).

'''
our_dir = our_dir:match("@(.*)"..package.config:sub(1, 1)) or "."
'''
The path seperators given are not the platform path separators.  In case they happen to be in some other version of lua I set the line to:
'''
our_dir = our_dir:match("@(.*)[/"..package.config:sub(1,1).."]") or "."
'''

With that fixed, the only concern I have is the target directory for the output of the documentation. I feel it should be inside the build folder not the source as I mentioned previously.

Alberth289346:
Fixed the LDocGen script as suggested (thanks TheCycoONE), docs are now generated out of source. Reverted movement of the LDocGen/output data files by timdiels, and added copying of them in the CMakeLists file.

Alberth289346:
'ls' lists the doc data, so that seems to work too

TheCycoONE:
Merged.

Issue #724, opened 2015-03-18
Allow setting non-shifted scroll speed


leaveboss:
I am using CorsixTH(v0.40 RC3) with a full version of ThemeHospital. The scroll speed is too slow when I move the mouse cursor to the edge of the screen. It is much slower compared to the original game version. I set the scroll_speed from 2 to 9 in 'config.txt' (%Users%\AppData\Roaming\CorsixTH\config.txt). Back to game, nothing changed.
Can anyone help me with this? Thanks

TheCycoONE:
It's not very clear on this point, but the scroll_speed in the config file is just the multiplier when fast scrolling is enabled (while you are holding down the shift key).  2 is 2x normal speed.

leaveboss:
Got this. Is there a way to accelerate the scroll speed by mouse?

TheCycoONE:
Holding shift will accelerate both mouse and keyboard scrolling, but there is no way (without modifying the game's lua) to speed up the scrolling with just the mouse at this time.

It would be pretty easy to allow setting both the regular scroll speed and the modified scroll speed in the config file, which would allow people to set regular scroll to fast and shift scroll to fine if they'd rather work that way.

MarkL1961:
`  It is much slower compared to the original game version.   `
For me they seem to be about the same.  Maybe Theme hospital seemed faster as it was a much small resolution?
Perhaps it is something that is specific to some computers compared to others - different graphics cards for instance may slow things down if they have a higher definition! 
~ just throwing ideas out there ~

TheCycoONE:
Didn't seem much different to me either (running dosbox and CorsixTH side by side) but if we fix up the configuration options everyone should be happy.

sadger:
This is solved now after pr #726

PR #726, opened 2015-03-20
Add shift_scroll_speed and make scroll_speed apply to unshifted scrolling

TheCycoONE:
Solves #724 

~~Warning, since the old shift scroll speed setting was called scroll_speed and this is now the unshifted scroll speed, existing players will have their regular scroll speed increased to twice normal.  Adjust your config.~~

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -688,6 +688,15 @@ function App:fixConfig()
     if key == "height" and type(value) == "number" and value < 480 then
       self.config[key] = 480
     end
+
+    if (key == "scroll_speed" or key == "shift_scroll_speed") and
+        type(value) == "number" then
+      if value > 10 then
+        self.config[key] = 10
+      elseif value < 1 then
+        self.config[key] = 1
+      end
+    end
   end
 end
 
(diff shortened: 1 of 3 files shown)


TheCycoONE:
Updated the pull request to multiply scroll speed by 0.5 so old default config values match the new one, and bound the scroll_speeds between 1 and 10 as mentioned in the config file description.

TheCycoONE:
Fixed a tab/space issue, and added a comment explaining the 0.5 multiplication in code.

CorsixTH/Lua/app.lua
```diff
@@ -688,6 +688,15 @@ function App:fixConfig()
     if key == "height" and type(value) == "number" and value < 480 then
       self.config[key] = 480
     end
+
+    if (key == "scroll_speed" or key == "shift_scroll_speed") and
+      type(value) == "number" and value > 10 then
+      self.config[key] = 10
+    end
+    if (key == "scroll_speed" or key == "shift_scroll_speed") and
```
sadger: This check apart from the "value" value is identical to the one above. Can probably be written more concisely. Switching on the "value" value only.

TheCycoONE:
Like that @sadger or did you have something else in mind?

sadger:
Looks good to me :+1:

PR #728, opened 2015-03-28
Cleanup pathfinder

Alberth289346:
Move the #define magic of the 4 path finders to 4 derived classes, cleaning up compile warnings along the way. Focus was on making the code readable, and remove the compiler warnings.

Intentionally left the patch as many small changes (except for 1 or 2), so it's at least somewhat trackable what I did, until it is decided this is what is wanted.

Some things remain to be handled somewhen in the future, as far as I can see
- _openHeapPromote method in BasePathing::pathingTryNode looks wrong. The path finder could be broken here. It does get called as well.
- The 'return 0;' in MakeGuess reduces A* to Dijkstra algorithm, a much slower algorithm. By reversing the search direction, it may be possible to make it an A* search
- C++ has a priority queue nowadays in std::

Throwing out path finding uses in the lua code (and switch to entity map) may make some of the path finders completely obsolete.

### DIFF
--- a/CorsixTH/Src/th_pathfind.cpp
+++ b/CorsixTH/Src/th_pathfind.cpp
@@ -28,7 +28,407 @@ SOFTWARE.
 #include <queue>
 #include <math.h>
 
-THPathfinder::THPathfinder()
+BasePathing::BasePathing(THPathfinder *pf) : m_pPf(pf)
+{
+}
+
+node_t *BasePathing::pathingInit(const THMap *pMap, int iStartX, int iStartY)
+{
+    int iWidth = pMap->getWidth();
+    m_pPf->m_pDestination = NULL;
+    m_pPf->_allocNodeCache(iWidth, pMap->getHeight());
+    node_t *pNode = m_pPf->m_pNodes + iStartY * iWidth + iStartX;
+    pNode->prev = NULL;
+    pNode->distance = 0;
+    pNode->guess = makeGuess(pNode);
+    m_pPf->m_ppDirtyList[0] = pNode;
+    m_pPf->m_iDirtyCount = 1;
+    m_pPf->m_iOpenCount = 0;
+    return pNode;
+}
+
+/*! No need to check for the node being on the map edge, as the N/E/S/W
+    flags are set as to prevent travelling off the map (as well as to
+    prevent walking through walls).
+ */
+bool BasePathing::pathingNeighbours(node_t *pNode, uint32_t iFlags, int iWidth)
+{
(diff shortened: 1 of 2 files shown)


TheCycoONE:
The macros in the pathfinder started with caps, but the replacement methods should follow our convention of camelCase.

CorsixTH/Src/th_pathfind.cpp
```diff
@@ -28,166 +28,154 @@ SOFTWARE.
...
-    { \
-        m_pDestination = NULL; \
-        break; \
-    } \
-    else \
-        pNode = _openHeapPop()
+bool PathFinder::TryNode(node_t *pNode, uint32_t iFlags,
+                         node_t *pNeighbour, int direction)
+{
+    uint32_t iPassable = iFlags & THMN_Passable;
```
TheCycoONE: Unless I missed something, iPassable and iNFlags are defined the same for all the pathfinder types?  So a derived TryNode isn't needed?

TheCycoONE: Nevermind, I see that objectVisitor and idlePathFinder are a little more tricky.

Alberth289346:
Fixed name coding style (renaming method names only), travis fails on apt-get, but it compiles locally, as did travis the previous time.

mrtnptrs:
Testing if it now can build, as the site of the file it wants to download is just online.

TheCycoONE:
@mrtnptrs As was told to me, you can rebuild from the TravisCI site, no need to close and open the PR here.

mrtnptrs:
@TheCycoONE At your service!

TheCycoONE:
I've played with this a bit and it's running well.  It's also somewhat easier to follow than the earlier path finder and obviously doesn't rely on macros.  I'm in favour of squashing the commits and merging as is. :+1:

Alberth289346:
Done!

CorsixTH/Src/th_pathfind.h
```diff
@@ -26,6 +26,174 @@ SOFTWARE.
...
+    BasePathing(THPathfinder *pf);
+
+    //! Initialize the path finder.
+    /*!
+        @param pMap Map to search on.
+        @param iStartX X coordinate of the start position.
+        @param iStarty Y coordinate of the start position.
+        @return The initial node to expand.
+     */
+    node_t *pathingInit(const THMap *pMap, int iStartX, int iStarty);
```
ryunp: iStartY? Just the definition, but caught my OCD

TheCycoONE:
@ryunp good catch.

PR #736, opened 2015-05-05
Update German translation

Wuzzy2:
This PR will update the German translation. Things done:

- Translate all missing strings (which were added into 0.40 and possibly later, too)
- Use “Sie” instead of “du” everywhere (was inconsistent)
- Copy-editing (many minor typo and grammar fixes)


Alberth289346:
It looks to me like a few commas are in a weird place, but much better now.
Thanks!

Issue #739, opened 2015-05-08
Pressing CTRL-S crashes the game


pwaller:
Platform: Ubuntu 14.04, amd64, clang.

See also this issue, especially @jorgenpt's comments.

https://github.com/CorsixTH/CorsixTH/pull/626

PR #741, opened 2015-05-11
Reimplement THRenderTarget::takeScreenshot.

jorgenpt:
This creates a surface based on the rendertarget dimensions and a
standard RGB24 format (which SDL_SaveBMP should be able to use directly
without any conversions), rather than trying to retrieve the window
surface of our (potentially hardware-accelerated, which means it doesn't
have a backing surface) window.

Fixes CorsixTH/CorsixTH#739.


Alberth289346:
Code looks fine to me, and it works at my Linux system.

PR #743, opened 2015-05-23
Move math.floor down to draw command after step is multiplied.

### LINKED ISSUE #742: Reputation bar is always set at 0.
This is a regression from 4.0. in Lua\dialogs\bottom_panel.lua:

local step = math.floor(width / (self.ui.hospital.reputation_max - self.ui.hospital.reputation_min))

math.floor results in step always being 0

TheCycoONE:
Fixes #742.  Thanks AgainstTheCurrent.


Alberth289346:
Looks good

PR #747, opened 2015-06-06
Reduce render target creation params magic

Alberth289346:
Adding some doxymentation, and removing a #define.


Alberth289346:
Removed the bReuseContent, and squashed all commits into one

PR #763, opened 2015-06-25
Use c++ style standard includes

Alberth289346:
Testing with a few other compilers would be useful, imho


TheCycoONE:
Ug. This actually still breaks on MSVC. I tested this pull and it compiled successfully but not subsequently. :(  I will be trying to come up with an emergency patch - but the basic problem is that SDL_main_win32.c is C, but includes "config.h" which is now C++

TheCycoONE:
I can't imagine running a very old version of SDL_main.c on Windows in SDL2 is a good idea, so I'm testing removing that file altogether.

PR #764, opened 2015-06-26
[RDY] Remove sdl main c

TheCycoONE:
This is an attempt at removing SDL_main_win32.c, which is not needed when SDL2main is linked.

Once this is merged the `<cstdint>` change can be reapplied to config.h.  See 4916f155d08485


jorgenpt:
That looks good to me.

Issue #749, opened 2015-06-09
Level editor should not allow non-treatable diseass


bobcanthelpyou:
When a patient is diagnosed as pregnant and try to leave the GP's office an error occurs and the patient get stuck at the rooms door.

![bildschirmfoto von 2015-06-09 20 46 16](https://cloud.githubusercontent.com/assets/4523264/8066610/d9c2b30e-0ee8-11e5-8469-599fd7ee0102.png)

![bildschirmfoto von 2015-06-09 20 16 33](https://cloud.githubusercontent.com/assets/4523264/8066284/16dfda02-0ee7-11e5-9d12-2a5c443c7287.png)

If i try to send the patient home the patient is stuck outside the office at the door.

![bildschirmfoto von 2015-06-09 20 16 42](https://cloud.githubusercontent.com/assets/4523264/8066291/1ff1a454-0ee7-11e5-885f-05dab33bcfb7.png)

```
Will not check for updates since this is the Trunk version.
An error has occured!
Almost anything can be the cause, but the detailed information below can help the developers find the source of the error.
Running: The timer handler.
A stack trace is included below, and the handler has been disconnected.
./Lua/humanoid_actions/seek_room.lua:104: room operating_theatre not available
stack traceback:
    [C]: in function 'assert'
    ./Lua/humanoid_actions/seek_room.lua:104: in function 'action_seek_room_no_treatment_room_found'
    ./Lua/humanoid_actions/seek_room.lua:284: in function <./Lua/humanoid_actions/seek_room.lua:201>
    ./Lua/entities/humanoid.lua:575: in function 'Humanoid_startAction'
    ./Lua/entities/humanoid.lua:667: in function 'finishAction'
    ./Lua/humanoid_actions/walk.lua:152: in function 'timer_function'
    ./Lua/entity.lua:218: in function 'tick'
    ./Lua/world.lua:1108: in function 'onTick'
    ./Lua/app.lua:876: in function 'dispatch'
    ./Lua/app.lua:777: in function <./Lua/app.lua:772>

Warning: No event handler for timer
```

Single Scenario: Avatar 
CorsixTH version: 0.40 and master branch

TheCycoONE:
Could you post a save game?

bobcanthelpyou:
Now it is the patient in the ward: 
http://filebin.ca/24gmB2gYGKZc

MarkL1961:
[...]

PR #766, opened 2015-06-26
Disable diseases that need unavailable treatment rooms. Addresses #749.

Alberth289346:
Also split hospital room cost and discovery state initialization away from room initialization.


Alberth289346:
Untangled things a bit further. 'initRooms' decides about rooms, and delivers info about all rooms. 'initLevel' derives diseases (through world), hospitals use room info and diseases to initlialize themselves.

Also changed unused loop variables by _ here and there.

PR #787, opened 2015-07-09
Fix font cache segfault with unicode fonts when no canvas is given

### LINKED ISSUE #785: Unicode languages do not work
Although simplified_chinese.lua and traditional_chinese.lua is in "/usr/share/CorsixTH/Lua/languages" directory, I still can't select these two languages.

I installed the requested libs, especially freetype2 mentioned by the wiki, before compiling, but still not work.

Any clue?

Thanks.

TheCycoONE:
Fixes #785. Also removes unnecessary _setNullTexture function, and sets the
texture to NULL in _freeTexture as is conventional in C++.

### DIFF
--- a/CorsixTH/Src/th_gfx_font.cpp
+++ b/CorsixTH/Src/th_gfx_font.cpp
@@ -322,7 +322,7 @@ THFreeTypeFont::THFreeTypeFont()
         pEntry->iLastX = 0;
         pEntry->pData = NULL;
         pEntry->bIsValid = false;
-        _setNullTexture(pEntry);
+        pEntry->pTexture = NULL;
     }
 }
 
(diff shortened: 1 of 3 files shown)


TheCycoONE:
Thanks for noticing that, iTexture shouldn't exist at all anymore and possibly the entire union replaced with SDL_Texture. It was originally written the way it was to facilitate multiple rendering back ends, where some would return an index (because you can't have pointers to GPU memory), and others would return a pointer; but now we only need to support SDL2 and this pr could break on another back end because it compares against NULL. If there is still interest in supporting those abstractions I could restore _setNullTexture and add _isNullTexture. (subclasses or templates would probably be a better choice if we were to go down that road again though)

I could certainly document several of the fields in cached_text_t.

TheCycoONE:
Removed iTexture and all references to it, and documented cached_text_t.

Issue #796, opened 2015-07-12
Drug casebook: new condition discovered change selected disease


bobcanthelpyou:
When a disease is selected in the drug casebook and a new disease is discovered (and the name of the new disease is in the alphabet before the selected disease) the selection change to a different disease. This is an unexpected behavior.

For example: The disease 'The Squids' is selected and 'Heaped Piles' is discovered, now the disease 'Invisibility' is selected instead of 'The Squids'

![bildschirmfoto von 2015-07-12 08 25 21](https://cloud.githubusercontent.com/assets/4523264/8636724/cf6bb526-2870-11e5-819d-d4d77d5a56a8.png)

![bildschirmfoto von 2015-07-12 08 31 31](https://cloud.githubusercontent.com/assets/4523264/8636725/d1ee561e-2870-11e5-8cc0-a89bc161de9f.png)

PR #797, opened 2015-07-12
Fix: Instead of changing the disease name currently pointed to, select the current disease again.

Alberth289346:
That prevents a change of the selected disease in the open drug casebook window when a new disease is discovered.
See also issue #796


bobcanthelpyou:
Thanks for the fast fix, works for me.

PR #808, opened 2015-07-18
[RDY] Eclipse: Created Mars Lua Project & Made Luna Project Lua Only

J-Shep:
I now think there should be two CorsixTH projects, a Lua project and a
separate C++ project because Eclipse's debugger can only be setup to
debug Lua scripts from a Lua project & vice versa for making Eclipse
debug C++ files.

An update to the Lua project's configuration was required for Mars to
make it recognise that this is a Lua project.  This is why anyone trying
to setup the debugger in Mars for the CorsixTH project couldn't because
of the "invalid project" error message.

The Mars configuration for the C++ project is just a duplicate of its
Luna counterpart.


TheCycoONE:
I'm going to suggest not including the C++ project.  Your current one is tied to windows (with mingw), and cmake is able to generate eclipse projects fairly easily that work for the c++ (but not the lua).

J-Shep:
@TheCycoONE I've removed the C++ project.

TheCycoONE:
Bundled eclipse project has been discontinued. Eclipse users can target eclipse from cmake.

Issue #561, opened 2014-12-24
Error when following a person going home


TheCycoONE:
_From [w.draa...@gmail.com](https://code.google.com/u/102444415587886192605/) on July 13, 2013 05:04:03_

When following a person leaving the hospital going home an error
appears. What is the expected output? What do you see instead? Programm stops with an error

What version of CorsixTH are you using (Version 0.21 - Windows 64-bit Archive)

What operating system: Windows 7

What level was this on: (Full game level 2)

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1665_

TheCycoONE:
Fixed.

TheCycoONE:
Not fixed, but duplicate of #124

PR #813, opened 2015-07-19
Fix crash when window of cured patient is opened while he walks off the map

Alberth289346:
Don't know if this should go in 0.50.

I found issue #561, which doesn't give much information, but I interpreted it as having a window on a cured patient open while he walks home. Sure enough, when he walks off the map, his px/py coordinates become floats which the C++ function doesn't want, and the drawing code crashes.

I traced the code to game_ui, where a "limit to diamond" function performs floating point arithmetic to compute the x/y position in that case. The fix was then a simple 'floor' around the returned coordinates.
I didn't change the other returns, as no computations are being done on the coordinates in those cases.


TheCycoONE:
It wouldn't be the same error, lua wasn't picky about floats vs ints until 5.3.

TheCycoONE:
The issue is fixed though, it use to be 100% reproducible.

TheCycoONE:
I was premature in saying the crash was fixed, reproduced.

Issue #133, opened 2014-01-11
Editing room causes an error


MarkL1961:
Load the save and edit the slack tongue clinic and you get this error

Error in buttonup handler: 
...pidemics\CorsixTH\Lua\humanoid_actions\seek_room.lua:110: attempt to index field 'hospital' (a nil value)
stack traceback:
	...pidemics\CorsixTH\Lua\humanoid_actions\seek_room.lua:110: in function 'action_seek_room_no_treatment_room_found'
	...pidemics\CorsixTH\Lua\humanoid_actions\seek_room.lua:287: in function '?'
	...imental-epidemics\CorsixTH\Lua\entities\humanoid.lua:551: in function 'Humanoid_startAction'
	...imental-epidemics\CorsixTH\Lua\entities\humanoid.lua:618: in function 'setNextAction'
	...orsixTH-experimental-epidemics\CorsixTH\Lua\room.lua:95: in function 'callback'
	...orsixTH-experimental-epidemics\CorsixTH\Lua\room.lua:808: in function 'deactivate'
	...orsixTH-experimental-epidemics\CorsixTH\Lua\room.lua:815: in function 'tryToEdit'
	...ixTH-experimental-epidemics\CorsixTH\Lua\game_ui.lua:520: in function <...ixTH-experimental-epidemics\CorsixTH\Lua\game_ui.lua:481>
	(tail call): ?
	(tail call): ?
	...CorsixTH-experimental-epidemics\CorsixTH\Lua\app.lua:740: in function <...CorsixTH-experimental-epidemics\CorsixTH\Lua\app.lua:735>

https://www.dropbox.com/s/2umrgixkwf81hkg/Autosave7.sav

there may be a connection to http://code.google.com/p/corsix-th/issues/detail?id=794, although with that issue it is usually diag rooms

J-Shep:
The problem was that a destroyed patient wasn't removed from this room's "humanoids_enroute" set because enroute patients aren't removed from this set when they are sent home or die. I will fix this later.

TheCycoONE:
Originally reported against v0.10 as seen in #497

J-Shep:
I will submit pull request with a fix by Saturday, I'm sorry didn't fix this 2 weeks ago when I had identified the cause of this problem.

J-Shep:
>> The problem was that a destroyed patient wasn't removed from this room's "humanoids_enroute" set because enroute patients aren't removed from this set when they are sent home or die.

[...]

PR #822, opened 2015-07-22
Fixed Lua Error #133

J-Shep:
#133: Patients who ceased to exist after using the autopsy machine could still be referenced in the humanoids_enroute table for a treatment room they required when this researched treatment room was created while a patient requiring it was using an autopsy machine. This would queue an action for them to go to the created room after using the autopsy machine, adding them to its humanoids_enroute table.

This reference to a deceased patient in a room's humanoids_enroute table would cause a lua error to occur when the room was edited because the room's deactivate function would reference the patient's hospital field which was nil because the autoposy machine had made it nil.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/seek_room.lua
+++ b/CorsixTH/Lua/humanoid_actions/seek_room.lua
@@ -257,7 +257,12 @@ local function action_seek_room_start(action, humanoid)
           end
         end
         if found then
-          action_seek_room_goto_room(room, humanoid, action.diagnosis_room)
+          -- Don't add a "go to room" action to the patient's queue if the
+          -- autopsy machine is about to kill them:
+          local current_room = humanoid:getRoom()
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -257,7 +257,12 @@ local function action_seek_room_start(action, humanoid)
           end
         end
         if found then
-          action_seek_room_goto_room(room, humanoid, action.diagnosis_room)
+          -- Don't add a "go to room" action to the patient's queue if the
+          -- autopsy machine is about to kill them:
+          local current_room = humanoid:getRoom()
+          if not current_room or not (current_room.room_info.id == "research" and current_room:getStaffMember() and current_room:getStaffMember().action_queue[1].name == "multi_use_object") then
```
TheCycoONE: Could you explain why
```if not current_room or current_room.room_info.id ~= "research" then ``` isn't sufficient?

J-Shep: A patient can leave the research room to go to the created treatment room up until the staff member has started their multi_use_object action.

Issue #819, opened 2015-07-21
0.50 beta 2 - room.lua:699: attempt to perform arithmetic on a boolean value


zsolmanz:
**What steps will reproduce the problem?**
1. Load provided autosave game file of Level 10 Eggsenham
2. Purchase south/eastern plot for $6750
3. Build Operating Theatre within that plot
1. Alternatively, load "Can't build" save game file of Level 10 Eggsenham and repeat test.

**What is the expected output? What do you see instead?**
Operating Theatre should build without issue, as it does in other plots (limited testing) and as other rooms do within this plot (limited testing).
Instead, game throws "No event handler for button up" issue.
With the "Can't build" save game file, it also tends to throw a "No event handler for timer" issue, and ask whether a recovery attempt should be made.  Obviously you can't click the tick to make a recovery with a button up error, but it's a nice thought.

**What version of CorsixTH are you using (e.g. "0.30", "4c66e39985")?**
0.50 beta 2

**What operating system / compile settings are you using?**
Windows 8.1 64bit, based on the GOG game files.

**What level was this on (e.g. "Demo level", "Full game level 12")?**
Full game level 10, Eggsenham

**Upload a saved game file (preferably an autosave from before the error) if you
have one and it is relevant... If possible, also upload the gamelog. Look at HOW TO UPLOAD GAMELOG AND/OR SAVES for how to do this:**
You need to update your CONTRIBUTING.txt to explain where to find this stuff is - I wouldn't have known to look in AppData\Roaming for it.
The files are [**here**](https://drive.google.com/open?id=0BzMXwT_4ohHBfkNTcjlYQzQxdlZNcTBsdDRPOWpoOUUxMTlnWkFHci1Dc2NObHFGdVl6a3M).

**Please provide any additional information below:**
There's a stuck handyman in decontamination from an earlier (recovered) error, but to my knowledge he hasn't caused any issues besides making it impossible to build upon his square if you delete the room.
It's worth highlighting that this issue is unrecoverable simply by nature of preventing all mouse-clicks.

Alberth289346:
Good point about the CONTRIBUTING.TXT

Copied the stack dump for easier access, can't upload the savegame unfortunately.
Thanks for the report.

````
Warning: Trying to remove nonexistant room build callback (function: 00000084F8448530) from humanoid (table: 00000084FD43C260).
[...]

PR #825, opened 2015-07-24
Don't crash for unreachable room

TheCycoONE:
This fixes the immediate crashes caused by building a room in an unreachable
area as per issue #819 

It should be noted that this is a very small step towards supporting unreachable areas. If you actually put any staff in this area the game will still quickly crash.


MarkL1961:
Just to clarify in TH you could not place staff in blocked off areas, but you could trap them by blocking off an area they were in.  If you tried to place one they remained on the cursor in grey the same as they do outside the hospital.
If I have the time I'll check what the situation was for rooms and objects.

Alberth289346:
Don't think trapping is an actual problem. If they can't go out, patients can't go in, and you're paying their salary.

PR #831, opened 2015-07-25
Add the same functionally to the keypad enter key as the return key (enter key)

bobcanthelpyou:


CorsixTH/Lua/ui.lua
```diff
@@ -803,7 +804,7 @@ function UI:afterLoad(old, new)
 
   if old < 100 then
     self:removeKeyHandler({"alt", "enter"}, self)
-    self:addKeyHandler({"alt", "Return"}, self, self.toggleFullscreen)
+    self:addKeyHandler({"alt", "return"}, self, self.toggleFullscreen)
```
TheCycoONE: I'm not sure why you changed this line, or the other place you substituted Return for return. It's not documented in your commit.

TheCycoONE:
Keyboard bindings also need to be added for old save games, at the very least the one for ui.lua.  Bump the savegame number, and add an afterLoad condition to add the new one.

CorsixTH/Lua/ui.lua
bobcanthelpyou: It is just 'consistency', cause every other addKeyHandler is using lowercase key-names, i know addKeyHandler will lowercase all key-names, too. So this is not necessary, but i did it. I hope it does not make problems with the afterLoad stuff for old savegames

bobcanthelpyou:
I just have to add something like:
```
   if old-savegame-version < new-savegame-version then
     self:addKeyHandler("keypad enter", self, ...)
```

TheCycoONE:
yep, and bump the new savegame version (in app.lua)

bobcanthelpyou:
added the new keyhandler to the afterload functions and bump the savegame version: #833 

Did i missed something? Tested most of them with some old savegames, should work hopefully.

PR #834, opened 2015-07-26
Use the 'resize_room' cursor when editing a room like TH does

bobcanthelpyou:
Had a look at the origin TH, today and recognized that there is a different cursor used when a room is in blueprint mode. Want to see it in CorsixTH, too :)

bildschirmfoto vom 2015-07-26 02 59 56

In TH the cursor changes back to the default cursor when mouseover the dialog or the menu bar at the top of the screen, not sure if that is needed in CorsixTH, too.

Tested several scenarios starting at 'wall phase' up to 'objects phase' and backwards. Building a new room and editing another room.

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -187,6 +187,7 @@ function UIEditRoom:cancel()
     else
       self:close()
     end
+    self.ui:setCursor(self.ui.default_cursor)
   elseif self.phase == "objects" then
     self:stopPickupItems()
     self:returnToDoorPhase()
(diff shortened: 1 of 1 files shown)


TheCycoONE:
Going to hold off on this until after the release, thanks for the contribution.

bobcanthelpyou:
Okay until then i have some time to improve the pr a little bit.

Maybe it is better to use setDefaultCursor instead of setCursor while the room is blueprinted. Also i could try to improve the behavior with dialogs and the menu bar.

I also think the name of the cursor 'resize_room' is confusing and maybe should rename to something different, cause the cursors for resizing are different cursors. It looks like a tool for mortar, but i am not sure which could be a good name for the cursor.

Issue #835, opened 2015-07-26
Incorrect response from VIP visit


MarkL1961:
Taken this from the forum
http://i.imgur.com/iNZPJBD.png
http://forums.corsixth.com/index.php/topic,5608.msg8097.html?PHPSESSID=a70ecbc699a25878d9fc65155360ae57#new

Alberth289346:
In Lua/entites/vip.lua, around line 151, the mediocre rating (3 of 1..5) fax is constructed, which has a 'mediocre' string, a boost, and a cash award. That looks like our guy.
I cross-checked with the message text, and indeed S[44][203] is a mediocre remark in the original strings. In other words, it's a mediocre VIP rating.

Line 473 defines the cash_reward, and a little below the rating is defines, which is between 45 and 64 for mediocre (probably of 100).

I agree the message is double.

The lesser levels (rating 1 and 2) have a reputation loss and no money, the higher levels (rating 4 and 5) are like the mediocre level at the moment, except for rating 5, which also seems to have a "thank you" action of some kind.

Something "in-between" would be nice, but how to achieve that?
* Drop the boost
* Drop the reward
* Reduce the reward

I don't know what to do here.

MarkL1961:
For such a comment I would think no reward and the reputation in community should go down and the fax should reflect that the reputation is reduced.

Alberth289346:
that's what "bad" and "very bad" do currently, so instead of 3 good-ish ones and 2 bad-ish ones, we'd have 2 good-ish ones and 3 bad-ish ones. I think mediocre deserves something in-between good and bad

Maybe the string itself is wrong?

MarkL1961:
I'll take a look at the strings then and get back to you as I was thinking that one was bad

MarkL1961:
The good ones should get a cash reward and a boost to the rep, but it only tells you about the cash.
Medium get a boost to the rep only and the bad get nothing

MarkL1961:
When john added the VIP I think he did some scheme to work from, I'll see if I can track it down

Alberth289346:
FYI, the string at the fax is https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/languages/original_strings.lua#L2065 Around it you can see the string numbers for the other ratings

MarkL1961:
https://www.dropbox.com/s/p7c40gdjsr7lpet/vip-algorithm.pdf?dl=0

MarkL1961:
[...]

PR #837, opened 2015-07-26
[RFC] VIP stops giving money and reputation to mediocre hospitals

Alberth289346:
No reputation boost or money if you are not at least 'good' for the VIP. See also discussion at #835
I expect this to go in after 0.50, but who knows :)


TheCycoONE:
Agree this can wait for 0.60; that release should have a number of balance changes if the drug price effect lands.

PR #845, opened 2015-07-30
update the tips

kkguo:
remove the old tips, as either their info is obsoleted, or they are just advertisement.
Translate according to English


bobcanthelpyou:
That looks much nicer, thank you very much :)

TheCycoONE:
Merged.

PR #850, opened 2015-08-03
Add missing strings to dutch.lua by Spanjard (Forum)

bobcanthelpyou:
Spanjard from the CorsixTH Forum provided these translations.

I provided him the missing strings from the debug-strings-diff-nederlands.txt file and insert them into the dutch.lua file. I did some checks if the strings are used ingame correctly, but not for all. 

I only do the pull request as requested by Spanjard.


spanjard:
I agree, on both changes maybe weergeven should be 'weergave'

bobcanthelpyou:
@Alberth289346 @spanjard Updated the commit with your recommendations, could you check again if it is okay?

spanjard:
I think it's okay, thnx Bob

Alberth289346:
And also thanks for forwarding the changes into the repository

PR #855, opened 2015-08-06
Update french.lua

Masteroshi:
Same string I modified yesterday, better translation.


TheCycoONE:
Is that your final answer? :)

Masteroshi:
Yes, Regis, That's My Final Answer. (tested in game : perfect)

PR #857, opened 2015-08-08
Fix extraneous .0 in progress tooltips under lua 5.3

### LINKED ISSUE #830: Decimal for hospital value
Greetings! I'm running CorsixTH Version 0.50 beta 2 for Windows and I encountered the following problem. When checking the hospital value with the mouseover the corresponding graph, the hospital value has a point and decimal rather than being an integer.

TheCycoONE:
Fixes #830 

I don't think it's right that %d ever formats to include a decimal, but I haven't been able to produce a simplified test case to report upstream to lua. This workaround solves the problem.


Alberth289346:
Given that there may be string proxies involved (app.lua, around line 406), which end up in th_lua_strings.cpp line 730, I am quite not sure we actually have normal lua strings here.

PR #863, opened 2015-08-09
Fix #862 mismatch of some German and English strings with full game

### LINKED ISSUE #862: Wrong adviser.warnings.pay_back_loan in German translation
When running the game in German i see the following message which sounds like the English equivalent of `adviser.level_progress.financial_criteria_met`:

bildschirmfoto von 2015-08-07 15 17 39

When running the game in English (and in some other languages i tested, too) i see the following message which is the `adviser.warnings.pay_back_loan` message

bildschirmfoto von 2015-08-08 18 08 26

What could be wrong here? Messages at different places in origin language files?

Should the `adviser.warnings.pay_back_loan` be overridden in the german.lua file to fix that or somewhere else?

Savegame where the message appear shortly after loading the savegame http://filebin.ca/2BRxAtTFSOIn/Autosave8.sav

bobcanthelpyou:
This should fix #862 a mismatch of some German strings compared to the English stings of the full game language files. This does not affect the demo game language files.

### DIFF
--- a/CorsixTH/Lua/languages/original_strings.lua
+++ b/CorsixTH/Lua/languages/original_strings.lua
@@ -44,6 +44,16 @@ local function fixGermanStrings(lang_num)
     S[44][168] = S[44][168]:sub(1, 45)
   end
 
+  -- Strings [28][32] and [28][33] are equal in the German translation of the full
+  -- version, this causes a mismatch with the English strings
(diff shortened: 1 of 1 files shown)


TheCycoONE:
Thanks for the fix, at this point the PR will have to wait for after the release.

PR #872, opened 2015-08-13
[RDY] Fix #800 don't play sound effects in new scenarios if play_sounds is disabled

### LINKED ISSUE #800: Sound is turned off but played on level startup
The sound is turned off (ALT + S) in an other level or scenario before, but on some maps (at the start of a scenario) the sound is played (like the typing of the doctors, ..)

bildschirmfoto von 2015-07-12 22 28 17

To turn the sound off it is needed to turn it on and off again.

If the game is saved and loaded the sound is off like it should be.

Expected behavior: the sound should be off

Debian Jessie 64 bit
CorsixTH trunk

bobcanthelpyou:
Added audio:playSoundEffects() in App:loadLevel() to disable / enable sound effects like it is done in LoadGame() for savegames

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -493,6 +493,9 @@ function App:loadLevel(level, ...)
   self.world = World(self)
   self.world:createMapObjects(map_objects)
 
+  -- Enable / disable SoundEffects
+  self.audio:playSoundEffects(self.config.play_sounds)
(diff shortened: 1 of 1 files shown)


bobcanthelpyou:
Updated the comment in the Lua code

TheCycoONE:
Good for merge following 0.50 :+1:

PR #875, opened 2015-08-17
Re-Organise french.lua, Translation fixes & formatting/organisation fixes

wolfy1339:
This PR re-organises french.lua to match the layout of english.lua

### DIFF
--- a/CorsixTH/Lua/languages/french.lua
+++ b/CorsixTH/Lua/languages/french.lua
@@ -23,15 +23,15 @@ Inherit("english")
 Inherit("original_strings", 1)
 
 ----------------------------------------------------------- Override -----------------------------------------------------------
-adviser.information.promotion_to_specialist = "L'un de vos INTERNES est devenu MÉDECIN." -- Fix the famous "Level 5 bug"
 misc.save_failed = "ERREUR : partie non sauvegardée." -- Much more french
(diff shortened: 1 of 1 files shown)


bobcanthelpyou:
Is it useful to order the translation files similar as the english.lua file?

Other translation files don't have an original language file in the background.

Is the order of the output of the debug-strings-new-grouped.txt file (created by 'dump strings') always the same? And could that file be a better reference for reorganize the language files in a better way?

wolfy1339:
There were some other fixes too mixed in there.

> Is the order of the output of the debug-strings-new-grouped.txt file (created by 'dump strings') always the same?

I'm not sure about that, but I think so.
> And could that file be a better reference for reorganize the language files in a better way?

Well, I would agree with you. the language files should be prganized the same

TheCycoONE:
@wolfy1339 is it possible for you to separate the fixes from the reordering. It's pretty hard to review the fixes when everything is moved.

wolfy1339:
I can try

wolfy1339:
I just pushed the fixes and will push the reordering later

TheCycoONE:
Thanks @wolfy1339

wolfy1339:
@TheCycoONE All done

TheCycoONE:
@bobcanthelpyou there is no 'order' for a lua table, unless you index it 1, 2, 3..., as such the dump doesn't have an order and it's not something that could easily be fixed.

PR #876, opened 2015-08-18
[RDY] Use C++ includes, namespaces, and keywords

TheCycoONE:
The first commit replaces #include "stdint.h" and #include "stddef.h" with C++ equivalents and
properly references the std namespace for functions and types imported from C
libraries. This process revealed several missing includes as the std:: namespaced functions were not implied in gcc without an explicit include.

The second commit replaces all instances of the C macro NULL with the C++11 keyword nullptr, which has the advantage of being an actual pointer type.


Alberth289346:
Perhaps add some typedefs around std::uint8_t  and friends? Having to type the std:: prefix for such common types is not so nice.

TheCycoONE:
It would be '''using namespace std::uint8_t''' rather than a typedef, but the practice seems controversial even for stdint types. I'm open to the change but I'd like to see what other large projects do.

Alberth289346:
using namespace  would be fine too, I wasn't aware of that option.

TheCycoONE:
I've looked around for some examples in the wild:
Boost always supplies fully qualified names.
Hedgewars uses the C library (stdint.h)
OpenTTD uses the C library
Webkit uses <cstdint> and splatters e.g. 'using std::size_t' etc. throughout the code (in both headers and cpp files) seemingly at random.
Qt typedef's it's own sized types e.g. quint32, and tries to avoid contamination by stdint types when using 3rd party code, going as far as:
```
#define uint64_t QT_PREPEND_NAMESPACE(quint64)
#define uint32_t QT_PREPEND_NAMESPACE(quint32)
#define uint8_t QT_PREPEND_NAMESPACE(quint8)
#define int_least16_t QT_PREPEND_NAMESPACE(qint16)

// Header from rfc6234 with 1 modification:
// sha1.h - commented out '#include <stdint.h>' on line 74
#include "../../3rdparty/rfc6234/sha.h"

#undef uint64_t
#undef uint32_t
#undef uint68_t
#undef int_least16_t
```
(the typo in the above code suggests they don't quite succeed)

I guess it's up to us what we want to do - especially considering we are not developing a library to be consumed by anyone else. Presumably Boost would be the most technically correct given their close affiliation with the C++ standards committee; but they are also a library and reference implementation for future C++ additions - we don't have the same requirement for technical cleanliness vs verbosity.

TheCycoONE:
Used ```using``` in config.h.in for all the cstdint and cstddef types that we use commonly as per @Alberth289346 's recommendation

TheCycoONE:
Sadly travis CI's build environment doesn't seem to be new enough to include C++11's 'to_string' method. I hope we can improve the environment, but if necessary it could be done with an ostringstream like the others.

TheCycoONE:
For reference: http://en.cppreference.com/w/cpp/string/basic_string/to_string

TheCycoONE:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52015 is the gcc / mingw bug. Resolved fixed in 4.8

TheCycoONE:
https://github.com/travis-ci/travis-ci/issues/1379 is relevant, we should be able to specify a newer gcc version.

TheCycoONE:
I've removed my last commit from this pull request so that it can be merged now. Switching to C++ string functions and updating travis to handle more C++11 features is something I'll do in the future.

PR #890, opened 2015-09-04
[RDY] Use travis containers and g++ 5 in TravisCI

TheCycoONE:
Travis is moving to new container system with faster start times. You can read about it here: http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

As a restriction we can't run sudo.

Additionally we are moving towards C++11, which is not while supported by Travis' built in g++-4.6. This pull request abandons the mingw builds (Windows builds are now handled natively by appveyor) and moves to native g++-5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,38 @@
 language: cpp
+compiler: gcc
+sudo: false
+addons:
+  apt:
+    packages:
+      - cmake
+      - p7zip-full
+      - build-essential
+      - sshpass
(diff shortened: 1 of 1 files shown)


TheCycoONE:
I'm not sure how to go from here to including mingw-w64-g++ 4.8. I like that the travis ci containers are much faster, but we might still need sudo so I'm hesitant to have this merged yet.

PR #894, opened 2015-09-06
Fix: French language could not be loaded, due to indexing into non-existing intermediate tables.

Alberth289346:
Double indexing for fields in a table does not work.

I fixed it enough to make it load again, whether all entries got preserved, I don't know.
Alternatively, patch by @wolfy1339 could be reverted.


TheCycoONE:
@wolfy1339 can you take a look at this.

wolfy1339:
Sorry for not replying. I will take a look

wolfy1339:
I am fine with the patch. Haven't seen anything abnormal
:+1:  from me

PR #898, opened 2015-09-14
Updating Spanish translation

IlDucci:
Removing a superfluous line, updating the copyright year.


Alberth289346:
Usually, you make such changes as part of other translation updates, but fair enough :)

PR #899, opened 2015-09-15
[WIP] Use FindLua included in CMake

TheCycoONE:
The FindLua module included in recent versions of CMake does a better job of
locating lua and using consistent versions for the include directories and
libraries.

In using this module I sacrafice the WITH_LUAJIT flag (not needed), and support
for building on x64 OSX with LuaJIT. It is recommended to build against Lua 5.3.


TheCycoONE:
I see the version of cmake included in travis CI is too old to contain the FindLua module. Given that I will modify the pull request to include the latest version of the upstream FindLua.cmake module

PR #900, opened 2015-09-21
LDocGen should not crash if it cannot get the name of the parameter.

Alberth289346:


Lego3:
Looks good. :-) Will make my pull request run through at least. :-P

PR #901, opened 2015-09-22
CMakeLists.txt: Cached CMAKE_INCLUDE_PATH & CMAKE_LIBRARY_PATH

J-Shep:
I've cached these CMake variables so that Windows devs who don't have an OS libraries directory can create their own "Libraries" directory with "include" & "lib" sub directories for their libraries to share, then they would only need to give the CMake GUI the two paths for these directories.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,8 @@ IF(APPLE)
 ENDIF()
 
 # Define our options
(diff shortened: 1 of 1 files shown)


J-Shep:
On second thoughts I guess caching these CMake variables wouldn't be worthwhile when the CorsixTH wiki tutorial for compiling it could just be modified to recommend using these variables, mentioning that their paths can be specified in the CMake GUI by using the "Add Entry" button.

Alberth289346:
Yep, a compile guide should explain how to set things up with cmake.

In addition, adding such a thing to our CMake file would mean you have to handle Linux and MacOS as well, which quickly becomes a much more complicated task.

PR #904, opened 2015-10-01
[RDY] Reduce macro usage

TheCycoONE:
Replaces many macros with consts or inline functions.
In a couple cases I changed types to match where they are used and avoid casts.


TheCycoONE:
A number of problems, const char[] cannot take numbers over 127, static enum doesn't make sense (should be enum).

TheCycoONE:
Implemented changes recommended by @Alberth289346. Use C++ vector in bootstrap so that the data can be passed to a function and maintain their size.

TheCycoONE:
Based on the conventions in the wiki I think camel case is right. I used underscore in rnc.cpp just to match the surrounding code which I think was copied from another project.

TheCycoONE:
Ok I've made the rest of the changes mentioned by @Alberth289346 that were appropriate. In the case of the lua.hpp functions there is already a comment in that file explaining why those functions are being defined. I've additionally changed rangeScale to range_scale to match other functions in that file.

PR #910, opened 2015-10-12
[WIP|RFC] Cpp11 font

TheCycoONE:
The second commit is the more ambitious and problematic one. I switched from our own utf8next function which is a bit of macro soup to the C++11 standard library unicode functionality.

Unfortunately there are a few problems with this approach.
1. In my tests (using visual studio 2013) it's very slow. It needs to convert/copy the characters to a u32string, which at least in that implementation led to extreme lag even on a modern gaming PC, when viewing the debug overlays.
2. It's currently broken in the latest Visual Studio 2015 (https://connect.microsoft.com/VisualStudio/feedback/details/1403302/unresolved-external-when-using-codecvt-utf8)
3. As with our current solution, this does not handle decomposed unicode characters (NFD form). (See https://en.wikipedia.org/wiki/Unicode_equivalence)

I have done some research and found that a lot of people use a library called ICU (http://site.icu-project.org/) for character set conversion. It appears to be well maintained, extensive, and capable of converting not just utf-8, but also the ibm-437 code page used for sprites. Unfortunately it requires that all strings be stored as unicodeStrings, so it may not avoid the performance issues I mentioned in point 1.

Does anyone have any thoughts on where to go with this?

To reiterate, ideally we want:
1. Correct results
2. Performance (enough to not slow the game down noticeably)
3. Clean macro-free code (that doesn't reinvent the wheel).

### DIFF
--- a/CorsixTH/Src/th_gfx_font.cpp
+++ b/CorsixTH/Src/th_gfx_font.cpp
@@ -209,7 +209,7 @@ void THBitmapFont::drawText(THRenderTarget* pCanvas, const char* sMessage, size_
 
 THFontDrawArea THBitmapFont::drawTextWrapped(THRenderTarget* pCanvas, const char* sMessage,
                         size_t iMessageLength, int iX, int iY, int iWidth,
-                        int iMaxRows, int iSkipRows, eTHAlign eAlign) const
+                        int iMaxRows, int iSkipRows, text_align text_align) const
 {
     THFontDrawArea oDrawArea = {};
     int iSkippedRows = 0;
@@ -274,7 +274,7 @@ THFontDrawArea THBitmapFont::drawTextWrapped(THRenderTarget* pCanvas, const char
                 {
                     int iXOffset = 0;
                     if(iMsgBreakWidth < iWidth)
-                        iXOffset = (iWidth - iMsgBreakWidth) * static_cast<int>(eAlign) / 2;
+                        iXOffset = (iWidth - iMsgBreakWidth) * static_cast<int>(text_align) / 2;
                     drawText(pCanvas, sMessage, sBreakPosition - sMessage, iX + iXOffset, iY);
                 }
                 iY += static_cast<int>(iTallest) + m_iLineSep;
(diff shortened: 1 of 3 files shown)


Alberth289346:
having values with the same name as the enum type looks so horribly wrong :p

anyway, if the problem is only utf8next, I am sure freerct has some utf8 decoder somewhere without macros. Since I wrote it, I can give it a new license :)
If the problem you're solving is bigger, as in satisfying the wish to go C++11, I'd suggest to convert everything to unicode strings on startup. Assuming that's not feasible in a few simple commits, the first step is probably adding support for old and new string formats, and then slowly moving strings over to the new format.
Another option is to cache converted strings, so you only have to convert them once for the duration of running the game.

libicu is also used by openttd, it's indeed powerful, but also big afaik (yep, very, by the looks of it at my system, although no idea what you actually need). I also remember problems with building for windows and mac, but no idea whether that is still the case.

TheCycoONE:
Heh, we disagree on the enum type name thing - I think it's the enum class name is usually the most elegant name for variables/properties of that type - after all they represent the same thing. Perhaps because I've been exposed to the .NET framework for so long and they almost always use that convention e.g. [CommandType](https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlcommand.commandtype%28v=vs.110%29.aspx)

A replacement for utf8next would be interesting, particularly if we could turn it into an iterator for a new lightweight string type that stores a const char pointer and a length without doing a copy on construction. It wouldn't handle normalization, but I don't think there's anyway to properly do normalization and character set conversions and case conversions without large tables like the ones that bloat ICU. We can address those issues in the translation files themselves though, so we probably don't need to be doing that processing at run time.

The biggest issue with what we have is that we can only handle very simple left to right languages. If there is ever any interest in an Arabic translation work will be needed.

The problem I see with using unicode strings everywhere is that almost all the strings in CorsixTH come from and are manipulated in lua, and lua is only aware of c strings. It would be possible to provide a framework for making lua handle unicodeString but I think it would take a lot of work.

My searches have also turned up the [UTF-CPP](http://utfcpp.sourceforge.net/) library which looks promising as far as what we're looking for, though I'm concerned that it may not be widely used or well maintained.

TheCycoONE:
I don't think this was the right direction to take.

PR #912, opened 2015-10-18
[RDY] Machines smoke when needing repair

JoshKeegan:
Machines will now smoke when they are need of repair.
Shows 1-3 plumes of smoke depending on how badly it needs repairing.
3 plumes: use again & it will blow up
2 plumes: 1 use left
1 plume: 2 uses left

Smoke will still be displayed whilst the machine is in use, and whilst the repair icon is showing, disappearing once the machine is repaired or blows up.

Ideally, different animations should be displayed for each machine (& maybe orientation?). The smoke animations seem to occupy the range 3424-3472. 
As working out which smoke animation goes with which machine could end up being quite laborious I've just used 3424 for every machine, but if I (or anyone else) have the time to work out which animation goes with which machine, it will be straight forward to change this to use an id specified on the class rather than the hard-coded one.

### DIFF
--- a/CorsixTH/Lua/dialogs/place_objects.lua
+++ b/CorsixTH/Lua/dialogs/place_objects.lua
@@ -491,6 +491,7 @@ function UIPlaceObjects:placeObject(dont_close_if_empty)
     real_obj = object.existing_objects[1]
     table.remove(object.existing_objects, 1)
   end
+  local room = self.room or self.world:getRoom(self.object_cell_x, self.object_cell_y)
   if real_obj then
     -- If there is such an object then we don't want to make a new one, but move this one instead.
     if real_obj.orientation_before and real_obj.orientation_before ~= self.object_orientation then
@@ -503,11 +504,14 @@ function UIPlaceObjects:placeObject(dont_close_if_empty)
     end
     -- Some objects (e.g. the plant) uses this flag to avoid doing stupid things when picked up.
     real_obj.picked_up = false
+    -- Machines may have smoke, recalculate it to ensure the animation is in the correct state
+    if real_obj.strength then
+      real_obj:calculateSmoke(room)
+    end
   else
     real_obj = self.world:newObject(object.object.id, self.object_cell_x,
     self.object_cell_y, self.object_orientation)
   end
-  local room = self.room or self.world:getRoom(self.object_cell_x, self.object_cell_y)
   if room then
     room.objects[real_obj] = true
   end
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -39,7 +39,6 @@ function Machine:Machine(world, object_type, x, y, direction, etc)
 
   -- We actually don't want any dynamic info just yet
   self:clearDynamicInfo()
-  -- TODO: Smoke, 3424
   -- Change hover cursor once the room has been finished.
   local callback
   callback = --[[persistable:machine_build_callback]] function(room)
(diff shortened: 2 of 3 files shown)


JoshKeegan:
Made a small change to make smoke still show if the room was edited & the machine moved (without having to wait for the machine to be used again).

Alberth289346:
Prefixed the title to denote it's not ready to be merged; please change it to "RDY" if you think it is ready. Add "RFC" if you want to discuss something.
(and also post a remark here, title updates are not broadcasted to others by githb :( )

TheCycoONE:
The effect of moving the machine impacts your other pr #913 since presumably a proper fix wouldn't call onDestroy. Leaving it open too.

JoshKeegan:
Method for recalculating smoke now extracted from machineUsed & added comments, so comments on that are welcome.

@MarkL1961 I haven't looked at the machine usage being reset yet, but I intend to keep that as a separate PR request anyway (and as @TheCycoONE says it will probably tie in with #913), so if everyone's happy I think this should be almost there.

MarkL1961:
Take a look in \lua\dialogs\fullscreen\bank manager as there are examples there.  line 161 for instance.

Where function has function: function name () local is written differently as it is only called from within the same file - others can be called from other places e.g. world: tick()

JoshKeegan:
Thanks :)

So going back to the original point, the fn currently gets called from outside of the Machine class. Are saying that you'd like that call to be removed (as it might not be needed once the behaviour of moving a machine has changed)?

JoshKeegan:
I had a quick play around with local whilst renaming calculateUsesLeft() to getRemainingUses() and by making the function local, I was unable to access the object (with self.). Will recalculateSmoke not have the same problem if made local, or is there some way of accessing self from within a local function?

I suppose you could pass self as a parameter, but that seems more clunky than leaving it as is. I'm guessing I've missed something here!

TheCycoONE:
@JoshKeegan ```obj:foo(args)``` in lua is syntactic sugar for ```obj.foo(obj, args)```. ```function obj:myfunct(args)``` is syntactic sugar for ```function obj.myfunct(self, args)```. So you are already passing self as a parameter, it's just hidden. In the case of a local function you will need to pass ```self``` explicitly, but this is no different than what's already being done except that this time you see it.

It's a small price to pay for encapsulating functions that shouldn't be called by the outside world.

I think ```calculateUsesLeft``` could be left exposed though. That seems like harmless and possibly useful information for another object to ask for.

JoshKeegan:
@TheCycoONE Fair enough, I just thought I'd missed something because I expected there to be a way of not having to pass self.

I'm still unclear on what you wanted though, as recalculateSmoke is currently called from outside of the Machine class. 
That call might not be necessary in future once the behaviour of moving a machine is changed, but for now that call is necessary to make sure smoke shows again after a machine is moved.

TheCycoONE:
I see, it's setSmoke(bool) that's the helper method not used elsewhere. place_object unfortunately needs to be aware of recalculateSmoke (no way around that without refactoring so that objects are notified when they are placed - which is a good idea, but out of scope for this PR). I'm ok with it then.

JoshKeegan:
Sounds good, I'll make the change but it might have to wait until next week now.

Are you happy with the method name recalculateSmoke? I'm not a fan, but couldn't think of something more descriptive.

TheCycoONE:
Yeah, I'm not a big fan either. Maybe ```updateSmokeLevel``` or just ```placeSmoke```? Also I think in the case of ```room.crashed``` it should setSmoke(false) instead of just returning, though the condition won't come up. When you do that you can squash the commits together too.

Works well in my tests.

Lego3:
I really like that this is being added. :-) It's one of the bigger things left after AI and Multiplayer.

TheCycoONE:
Yes, good work

JoshKeegan:
Thanks guys, apologies for the delay.

I've made setSmoke local, fixed a typo, rebased & squashed the commits.
I've only tested this very quickly (after rebasing) but appears to work, so I think this is now ready to be merged.

JoshKeegan:
Made changes requested in comments on https://github.com/JoshKeegan/CorsixTH/commit/bc03e734a3a52305d7b3b20ecf679bfc06845997 & squashed commits

TheCycoONE:
:+1

PR #913, opened 2015-10-18
[WIP] Fix machine repair icon showing after machine is moved

JoshKeegan:
Fixes machine repair icon from still being shown when the machine has been picked up and moved.
corsixth move object mood icon

When picked up, the machine is destroyed which will remove any queued interactions with it (so the handyman will no longer be coming to repair it anyway).

Note: since all entities can have moods (not just machines), I've cleared any mood whenever any entity is destroyed. Can someone who knows the code think if this could cause any issues?

### DIFF
--- a/CorsixTH/Lua/entity.lua
+++ b/CorsixTH/Lua/entity.lua
@@ -275,6 +275,8 @@ end
 -- Function which is called when the entity is to be permanently removed from
 -- the world.
 function Entity:onDestroy()
+  -- Clear any mood
+  self:setMoodInfo()
   self:setTile(nil)
   self.world.dispatcher:dropFromQueue(self)
   -- Debug aid to check that there are no hanging references after the entity


TheCycoONE:
As mentioned in #912 the current behavior when machines is moved is wrong. I think correcting that behavior will completely change how this is managed.

TheCycoONE:
No update in over 2 years. If there is no update in 2 weeks I will close this PR.

JoshKeegan:
Sorry, I doubt I'll be able to look at re-implementing moving machines any time soon, which is why this is still marked as WIP.  

If nobody else is going to be working on it either though, this change did fix the bug so might be worth merging.  
If so though, it would be technical debt & should to be noted in the bug report for re-implementing moving machines as it's otherwise an unnecessary change.

TheCycoONE:
I actually think this change is worth merging anyway. Thanks @JoshKeegan

PR #914, opened 2015-10-18
Staff management screen scrolling fixes.

JoshKeegan:
- If staff fired leading to the current page being empty, scroll up a page
 - Recalculate whether the scroll dot should be visible after hiring/firing staff


TheCycoONE:
We are striving to have all new functions documented, please add documentation comments describing this function.

JoshKeegan:
Done, although I wasn't sure if documentation comments should start with --! or just -- as there seems to be a mixture used.

TheCycoONE:
:+1:

PR #919, opened 2015-10-25
Remove th_map #define macros

Alberth289346:
Rewrote copying of "Passable if not blueprint" bit to "passable" bit, to make it more clear and free from assuming the passable bit is bit 0.


Alberth289346:
Improved the comment message above the code block, and dropped passable as it wasn't doing much. While at it, found and fixed another case, which is perhaps also the final one, as the shiftdelta is now not needed any more.

PR #920, opened 2015-10-25
Remove 'Flag' macro for setting tile node flag values.

Alberth289346:


TheCycoONE:
Actually, this looks like a perfect candidate for std::map. Something like:
```
std::map<const char *, THMapNodeFlags> lua_node_flag_map = {
    {"passable", THMN_Passable},
    {"hospital", THMN_Hospital},
...
auto iter = lua_node_flag_map.find(field);
if(iter != lua_node_flag_map.end())
{
  if (lua_toboolean(L, 6) == 0)
    pNode->iFlags &= ~((*iter).second);
...
```

* Above was written by hand in this comment section, untested, I can't vouch for it but I think it should work.

TheCycoONE:
Actually you'll need to either specify an alternate compare function or use std::string for the key.

Alberth289346:
Suggestion implemented

Alberth289346:
Good suggestion.

Moved lua_node_flag_map up, and re-used it for getcellflags.

Alberth289346:
Updated to "val"

TheCycoONE:
TravisCI failure caused by https://github.com/Olivine-Labs/busted/issues/459, we can ignore it.

TheCycoONE:
Lets merge these commits, the second rewrites a lot of the first anyway.

TheCycoONE:
Merged with squash.

PR #921, opened 2015-10-26
prevent several handyman repair calls for a machine

bobcanthelpyou:
It is a little bit annoying when i call a handyman to repair a machine and the machine will be repaired several times. e.g. when the handyman was called automatically or several times via the callHandyman button, before.

### DIFF
--- a/CorsixTH/Lua/dialogs/machine_dialog.lua
+++ b/CorsixTH/Lua/dialogs/machine_dialog.lua
@@ -89,7 +89,7 @@ end
 
 function UIMachine:callHandyman()
   if self.machine.times_used ~= 0 then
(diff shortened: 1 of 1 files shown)


TheCycoONE:
I don't think this is the right approach for this issue. You are suppose to be able to call that function after a handyman has already been assigned to the machine, and it bumps the task to the top of the queue in that case. If you are getting multiple calls in the queue then getIndexOfTask is where I'd look.

bobcanthelpyou:
Oh thanks for the hint, there was a ```machine_tile_y``` instead of ```machine.tile_y``` that fixed this issue

bobcanthelpyou:
Oh thanks for the hint, there was a ```machine_tile_y``` instead of ```machine.tile_y```, this caused ```y``` of ```getIndexOfTash``` to be ```nil``` when using the callHandyman button. Using the correct value fixes the issue.

PR #922, opened 2015-10-26
[WIP] Remove any remaining litter after removing humanoids from the blue-print area

### LINKED ISSUE #66: Vomit in a room (editing it causes an error)
I think this is already reported here: http://code.google.com/p/corsix-th/issues/detail?id=1338&q=vomit%20in%20room But I now also know how to reproduce this:

1.Make a blueprint of a room and make sure a patient which is going to vomit is already standing at the blueprint or will stand (walk to the blueprint) at the blueprint after you made the selection of the blueprint, but before you confirmed it.
2. Confirm the blueprint.
3. The patient will vomit at the blueprint and the room can be builded over it.

And when I want to edit the room, I get this nice error:
Savegame version is 78 (0.30), originally it was 78 (0.30)
Error in buttonup handler: 
C:\Program Files\CorsixTH\Lua\dialogs\place_objects.lua:227: attempt to index field 'orientations' (a nil value)
stack traceback:
    [C]: in function 'sort'
    C:\Program Files\CorsixTH\Lua\dialogs\place_objects.lua:226: in function 'addObjects'
    C:\Program Files\CorsixTH\Lua\dialogs\edit_room.lua:603: in function 'returnToDoorPhase'
    C:\Program Files\CorsixTH\Lua\dialogs\edit_room.lua:187: in function 'callback'
    C:\Program Files\CorsixTH\Lua\window.lua:645: in function 'handleClick'
    C:\Program Files\CorsixTH\Lua\window.lua:1406: in function 'onMouseUp'
    C:\Program Files\CorsixTH\Lua\dialogs\place_objects.lua:449: in function 'onMouseUp'
    C:\Program Files\CorsixTH\Lua\window.lua:1390: in function 'onMouseUp'
    C:\Program Files\CorsixTH\Lua\ui.lua:703: in function 'dispatch'
    C:\Program Files\CorsixTH\Lua\app.lua:739: in function <C:\Program Files\CorsixTH\Lua\app.lua:734>

Attached save: https://drive.google.com/file/d/0Bwc5O5IR0pMpUkE1NExSTW9DQWc/edit?usp=sharing

This bug can be reproduced especially since version 0.30, because of the vomit tweaking, which causes much more vomit.

Alberth289346:
Fixes #066 at least partly.

It solves the "vomit between confirm and actual construction of room" problem, but if you test with the save game, you will notice red/blue colours of the blueprint are not properly updated until you move or resize the blueprint.

Not sure if that is supposed to be fixed here too.

I pondered about allowing overbuilding on litter, but that would mostly defeat the purpose of litter I think, so I didn't change that.

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -255,6 +255,13 @@ function UIEditRoom:confirm(force)
   end
 end
 
+--! Return whether humanoid entity is sticking around or walking into an area.
+--!param entity (Humanoid) Humanoid to check.
+--!param x1 (int) Smallest x value of the room.
+--!param x2 (int) Biggest x value of the room.
+--!param y1 (int) Smallest y value of the room.
+--!param y2 (int) Biggest y value of the room.
+--!return (bool) Whether the entity is blocking or going to block the area.
 function UIEditRoom:isHumanoidObscuringArea(entity, x1, x2, y1, y2)
   if entity.tile_x then
     if x1 <= entity.tile_x and entity.tile_x <= x2
@@ -335,6 +342,12 @@ function UIEditRoom:clearArea()
   end
 
   if next(humanoids_to_watch) == nil then
+    -- There might be litter on the floor (most likely a humanoid vomiting after
+    -- establishing the room blueprint, but before confirming).
+    -- Clear it all (ie also 'normal' litter).
+    local rect = self.blueprint_rect
+    self.world:clearLitterFromArea(rect.x, rect.y, rect.x + rect.w - 1, rect.y + rect.h - 1)
+
     -- No humanoids within the area, so continue with the room placement
     self:confirm(true)
     return
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -2257,6 +2257,11 @@ function World:getObjects(x, y)
   return self.objects[index]
 end
 
+--! Get an object from a node in the world.
+--!param x (int) X position of the node to retrieve from.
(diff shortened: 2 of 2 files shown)


TheCycoONE:
Pinging @MarkL1961 oracle of all things original theme hospital :)

MarkL1961:
I'd have to check that, I can remember being able to place objects over litter and it clearing.  It came in handy when there was a vomit wave as you could help the handymen to clean it up.  As for rooms, I don't think you could, but will check it for you later.

MarkL1961:
You can build a room over litter.   I did this with the original setup, so cannot do a screenshot.
If someone can remind me how to run TH in dos box I will do screenshots for you.  I thought all had to do was type in hospital or cd\hospital and enter, but it is returning illegal command

MarkL1961:
sorted
build over litter

TheCycoONE:
Thanks @MarkL1961. @Alberth289346 I think we should replicate this.

Alberth289346:
Yep, will see how the litter currently works, thanks MarkL

sadger:
Looking at the edit room code we could optimise the detection of entities in the build area by using the entity map instead of looping through all world entities. It's directly related to this change or should I do a fresh PR?

TheCycoONE:
I don't mind if it's tacked on as a separate commit in this PR, on the other hand I see no reason why it has to be, and this has been in purgatory for a long time so it might be best to make a new one.

Alberth289346:
There was something wrong about the patch, as in, not working correctly in combination with the handy men. As such, addressing performance is best done separately.

sadger:
OK, thanks for the input guys.

Alberth289346:
I dea was too simplistic, #381 looks like a better approach

PR #926, opened 2015-11-04
[RDY|RFC] Cleanup THMapNode

TheCycoONE:
Changes THMapNode's iFlags (a uint32_t bit field) into a custom class where
each field is directly accessable.

Also converts the uint64_t * structure used to list THObjectType into a standard std::list, greatly simplifying access and manipulation of that structure.


TheCycoONE:
Fails in gcc with "th_lua_map.cpp:(.text+0x47a3): undefined reference to `th_map_node_flags::passable_mask'" etc.

I haven't yet looked into why this is, or whether gcc or msvc is more 'correct'.

TheCycoONE:
Swapped static const uint32_t values for an enum, which provides more consistent support across msvc and g++.

TheCycoONE:
Updated to satisfy @Alberth289346's personal hatred of ```== false``` ;-)

TheCycoONE:
Further cleaned up THMapNode, removing the object_type value from flags and making a single objects list.

TheCycoONE:
Fixed the unused ```found``` variable as per @Alberth289346 's comment, and changed th_map_node_flags to use array indexing instead of bit operations to modify values by key.

PR #927, opened 2015-11-04
[WIP] Added animation loader

sadger:
Just the start for loading animations, works with plant idle animations and nothing else currently.


TheCycoONE:
No update in over a year. If there is no update in 2 weeks I will close this PR.

PR #930, opened 2015-11-06
Reduce ownerships of level config, cleanup

Alberth289346:


TheCycoONE:
Looks good, I think you could join some of these commits - then you won't need a commit with the message "Other Improvements".

Alberth289346:
Changes:

- Reverted free_build_mode code change in hospital.lua (was in commit 91be0b34de6a1e8acee9972416236ab949e775f8 around line 1457)
- Added ResearchDepartment:afterLoad in commit e49db0605bf7b0e4fbef9bf8f86dc3ed9215546b and bumped savegame (place in list above is wrong for some reason)
- Added Epidemic:afterLoad in commit 067ff83017d5ad037ed16831f9a6a6b8ce72c7fa (also wrong place in list)

As for merging, just merge everything into one commit?

TheCycoONE:
The order of commits on the github site is confusingly just by date.
Have you verified those afterLoad methods are called? I know world.lua is suppose to call it on all hospitals, entities and rooms that it's aware of which probably covers the ResearchDepartment. I don't see where epidemic would be called from. (I would guess it should be chained in the afterLoad from Hospital and isn't?)

Alberth289346:
Updated hospital with calls to afterload in commit be876e17fa92a9b42a4147dc6f640942c0cf0145 and commit e9ab153d5045a0b4e6b5f29a7291205d4a76c172

TheCycoONE:
Ok, I think it's good. Can be rebased into one commit.

Alberth289346:
squashed & reworded commit message

PR #935, opened 2015-11-13
Shorten code for changing handyman task priorities

Alberth289346:


TheCycoONE:
Reads well to me :+1:

PR #937, opened 2015-11-14
Use RAII techniques with THStringList

TheCycoONE:
Move loading of THStringList to the constructor. Use exceptions for failures.
Improve the internal representation in THStringList. Notably avoids the naked new and delete statements.

### DIFF
--- a/CorsixTH/Src/th.cpp
+++ b/CorsixTH/Src/th.cpp
@@ -23,7 +23,7 @@ SOFTWARE.
 #include "config.h"
 #include "th.h"
 #include <cstring>
-#include <new>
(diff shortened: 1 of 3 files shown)


TheCycoONE:
I looked into moving this class but I'm not happy with it. By convention th_lua* just contains the lua bindings. The class also requires the utf8encode and the two charset functions which are not normally exposed anywhere else, as well as the charset lookup tables. To me it feels cleaner to leave it in th.cpp.

TheCycoONE:
Improved the documentation of THStringList. Explicitly delete unimplemented move and copy operations.

PR #938, opened 2015-11-17
[RDY] Various small steps in reducing code complexity

Alberth289346:
May need a little testing, in particular Hospital:onPlaceObject was quite modified


Alberth289346:
Changed the staffed desk description, and dropped handling new things in the hospital.

PR #940, opened 2015-11-19
Fix parameters to sws_scale

### LINKED ISSUE #939: On FFmepg 2.8+ every animation cause crashes
My data files come from Origin (french) and on Windows everything is perfect.

But on Archlinux (https://aur.archlinux.org/packages/corsix-th) every animation cause crash with the error: 
`[1]    10302 segmentation fault (core dumped)  CorsixTH`

I have to delete or rename the folders `ANIMS` and `INTRO` for the game to work.

TheCycoONE:
Was passing destination height instead of source height, leading to segfaults

Fixes #939


TheCycoONE:
I tacked on a second commit that replaces some deprecated ffmpeg/libav with the newer alternatives because they are dropped in the current development versions and it needed to be done.

PR #942, opened 2015-11-21
[RDY] Do not use shared opengl context for movies

TheCycoONE:
*Updated*
Avoids the use of GL shared contexts which are buggy on some hardware. This PR also moves the scaling of movies to SDL where it may be done in hardware, and improves the documentation of thmovie.h. A few smaller housekeeping items were taken care of as well.

### DIFF
--- a/CorsixTH/Src/th_lua_movie.cpp
+++ b/CorsixTH/Src/th_lua_movie.cpp
@@ -71,11 +71,13 @@ static int l_movie_play(lua_State *L)
     const char* warning;
     THMovie *pMovie = luaT_testuserdata<THMovie>(L);
     pMovie->clearLastError();
-    pMovie->play(static_cast<int>(luaL_checkinteger(L, 2)),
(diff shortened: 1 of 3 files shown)


Alberth289346:
Intro, and level board movies work flawlessly with this patch. I also tried a different resolution, that works too. Looks like you're on the right track!

TheCycoONE:
Excellent, I'll clean this up for a pr then.

TheCycoONE:
I've updated this PR. I now consider it ready to be merged (pending review).

Alberth289346:
Except for some minor things in the added docs, looks fine to me.
It's also still running at may computer.

TheCycoONE:
*my computer.  Alright, I need spell check then. :) Thanks.

TheCycoONE:
Updated with the above corrections. For some reason Doxygen is only generating help for config.h so I wasn't able to test.

Issue #889, opened 2015-09-04
reset MP3 folder


legluondunet:
Hello, I choose a folder with mp3 tracks. Now I would like to reset this option to "no folder, use original midi files from game". But this doesn't seem possible from the interface. Could you add this functionality from the interface?

Lego3:
This will come in the soon to be released version 0.60

PR #947, opened 2015-12-02
[RDY] Mp3 folder reset functionality

Lego3:
Solves #889.
I did a few commits, should I squash them?


TheCycoONE:
7789566 doesn't remove the usage of music_warning from folder_settings.lua, it should be removed from code as well as the language files right?

1e93923 and fce138f should be squashed and 7789566 should remain on its own I think.

Lego3:
Indeed, 7789566 does not remove any usage. I only removed a now incorrect part of the string music_warning, not the whole thing.

PR #948, opened 2015-12-03
[RDY] Fix luaT_register

TheCycoONE:
luaL_newlibtable expected a C array to set the size of the table based on the
number of elements, but since luaT_register is a function that array is being
auto cast to a pointer where the size wasn't available. By using vectors instead
we preserve the size.


Alberth289346:
Looks fine to me

PR #954, opened 2015-12-12
Broadcast new rooms to entities instead of keeping a callback administration.

Alberth289346:
there is loads more cb queues, not sure if it should be added here

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -30,7 +30,7 @@ local runDebugger = dofile "run_debugger"
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 107
+local SAVEGAME_VERSION = 108
 
 class "App"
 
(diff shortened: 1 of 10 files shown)


TheCycoONE:
This will currently break any save game where any of those callbacks are currently registered. There's not enough changes yet that I think that would be worth it, maybe add e.g.
```
-- callbacks for savegame compatibility
----[[persistable:patient_toilet_build_callback]] function(room)
if room.room_info.id == "toilets" then
  self.going_to_toilet = "no"
  self:unregisterRoomBuildCallback(callback)
end
```

I believe you could put that anywhere, and conventionally we put these at the bottom of the files they were originally defined in. Actually you could probably just make the definition blank and set callback to nil in your afterLoad?

Alberth289346:
Added dummy callback function to patient, machine, and staff files.

Alberth289346:
Last part of the patient constructor is now
```` lua

  -- Is the patient trying to get to the toilet? ("yes", "no", or "no-toilets")
  self.going_to_toilet = "no"
end
````

PR #955, opened 2015-12-12
Making more use of the entity_map

Alberth289346:


Alberth289346:
"underway" -> "on her way" (which works until we get male receptionists)
added "notify_distance" variable (bench_notify_distance was a bit too long for the line)
"get_entities_by_id" -> "get_objects_by_id" function rename

PR #956, opened 2015-12-16
Plant refactorings

Alberth289346:
Plants needed some love (and water)

### DIFF
--- a/CorsixTH/Lua/objects/plant.lua
+++ b/CorsixTH/Lua/objects/plant.lua
@@ -166,13 +166,6 @@ end
 
 --! When the plant needs water it periodically calls for a nearby handyman.
 function Plant:callForWatering()
-
-  if self.unreachable then
-    local ux, uy = self:getBestUsageTileXY(handyman.tile_x, handyman.tile_y)
-    if ux and uy then
-      self.unreachable = nil
-    end
-  end
   -- If self.ticks is true it means that a handyman is currently watering the plant.
   -- If there are no tiles to water from, just die.
   if not self.ticks then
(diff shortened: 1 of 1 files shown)


Alberth289346:
Commit 0929e39 is best seen with -w:
```` diff
diff --git a/CorsixTH/Lua/objects/plant.lua b/CorsixTH/Lua/objects/plant.lua
index 63f70fa..816dc61 100644
--- a/CorsixTH/Lua/objects/plant.lua
+++ b/CorsixTH/Lua/objects/plant.lua
@@ -175,22 +175,22 @@ function Plant:callForWatering()
   end
   -- If self.ticks is true it means that a handyman is currently watering the plant.
   -- If there are no tiles to water from, just die.
-  if not self.ticks and not self.unreachable then
+  if not self.ticks then
+    if not self.unreachable then
       local index = self.hospital:getIndexOfTask(self.tile_x, self.tile_y, "watering")
       if index == -1 then
         local call = self.world.dispatcher:callForWatering(self)
-      if self.current_state > 1 and not self.plant_announced then
-        self.world.ui.adviser:say(_A.warnings.plants_thirsty)
-        self.plant_announced = true
-      end
         self.hospital:addHandymanTask(self, "watering", self.current_state + 1, self.tile_x, self.tile_y, call)
       else
+        self.hospital:modifyHandymanTaskPriority(index, self.current_state + 1, "watering")
+      end
+    end
+
+    -- If very thirsty, make user aware of it.
     if self.current_state > 1 and not self.plant_announced then
       self.world.ui.adviser:say(_A.warnings.plants_thirsty)
       self.plant_announced = true
     end
-      self.hospital:modifyHandymanTaskPriority(index, self.current_state + 1, "watering")
-    end
   end
 end
````

I think the removed code is dead. There is only one call to this function, and it has "not self.unreachable" as pre-condition. In addition, there is a "handyman" variable mentioned in the removed code, which does not exist at that point. Last but not least, self.unreachable gets cleared in tickDay, so at worst, removing this code just delays trying again.

PR #962, opened 2016-01-02
Fix Lua CMake To Find Lua On OS X systems

msmollin:
This update adds additional search directories valid for homebrew installed versions of Lua on OS X platforms.
Also included are two additional failure errors for more specific messaging about what fails to be found.


TheCycoONE:
If someone needs lua 5.2 specified that way we can let them make a PR for it.

Issue #961, opened 2016-01-02
CMake on OS X not producing Xcodeproj file possibly due to <malloc.h> not being available?


msmollin:
I'm attempting to build the OS X version of CorsixTH from source (see #959 & #960 for other issues) and assuming I satisfy 959 and 960, I can successfully configure the project, and click the generate button.

However generation goes very quickly (even for the computer I'm working on) and no xcodeproj bundle is produced. Looking in CMakeError.log reveals what I believe is the culprit:

```
Determining if files malloc.h exist failed with the following output:
Change Dir: /Users/msmollin/corsix_build/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/make" "cmTC_bd4de/fast"
/Applications/Xcode.app/Contents/Developer/usr/bin/make -f CMakeFiles/cmTC_bd4de.dir/build.make CMakeFiles/cmTC_bd4de.dir/build
Building C object CMakeFiles/cmTC_bd4de.dir/CheckIncludeFiles.c.o
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc    -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk -mmacosx-version-min=10.11   -o CMakeFiles/cmTC_bd4de.dir/CheckIncludeFiles.c.o   -c /Users/msmollin/corsix_build/CMakeFiles/CMakeTmp/CheckIncludeFiles.c
/Users/msmollin/corsix_build/CMakeFiles/CMakeTmp/CheckIncludeFiles.c:2:10: fatal error: 'malloc.h' file not found
#include <malloc.h>
         ^
1 error generated.
make[1]: *** [CMakeFiles/cmTC_bd4de.dir/CheckIncludeFiles.c.o] Error 1
make: *** [cmTC_bd4de/fast] Error 2

Source:
/* */
#include <malloc.h>


int main(void){return 0;}
```

This is an OS X specific problem: since OS X based on FreeBSD 7 or 8, malloc.h is actually defined in stdlib.h. Other projects have run into this problem and ifdef'd around it (see - https://github.com/facebook/folly/issues/116), but I am loathe to make this change by myself due to my unfamiliarity with this codebase. Any thoughts on how to proceed? Is this actually the error that I'm running into, or is it possibly because I solved #959 & #960 incorrectly?

[...]

PR #963, opened 2016-01-04
Replace all usage of non-standard malloc.h / alloc.h

TheCycoONE:
Calls to alloc have been replaced with C++ vectors. malloc.h is not available
on some OSX machines as per #961

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,8 +99,6 @@ ENDIF(MSVC)
 
 # Environment handling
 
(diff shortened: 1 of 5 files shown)


msmollin:
+1 for modern C++ code :+1:

TheCycoONE:
Changed to use C++ casts instead of C style casts. It should be possible to remove these mallocs entirely by replacing m_pData and m_pStringBuffer with stl containers, but that is complicated by the fact that these variables are dual purposed and out of scope for this PR.

PR #966, opened 2016-01-21
Simplify and speed-up: Patient:changeDisease(new_disease)

Alberth289346:


TheCycoONE:
To recap, it's a good patch set. I think that visited_or_unavailable might actually only be visited but I'm not 100% sure on that. If it's the case the commit message should reflect that.

Alberth289346:
Thanks for the comments. While considering the remarks, I realized it's still way too slow.
Rewrote the entire thing now, much better now :)

PR #970, opened 2016-01-31
Map editor cell flags and door objects

Alberth289346:
Added different floor types leading to different combinations of passable, buildable, and hospital flags.
Added door objects

Remaining issue is a weird shadow sprite that gets added somewhere.

I also have a number of sanity checks for parcels (more can/should be added), but it needs a way to tell the user what is wrong.
A message is probably possible, but how to indicate what tile(s) is/are involved?

Can also be added later if you like. It sounds like a nice job for a starting lua programmer.


TheCycoONE:
:+1: from me. You can merge this directly, we don't require devs to make pull requests on feature branches.

Alberth289346:
I know I can just push it, but this is easier for you to find on a sun(ny?) day.

PR #313, opened 2014-09-16
[WIP] Drug casebook price impact [2]

benckx:
Issue: So far prices have no impact whatsoever. So player can increase all prices up to 200% without any consequence. It provides an unbalanced and undeserved advantage (it's almost a cheat).

The idea behind this patch is to compute the gap between the expected price level (given staff skill, fatigue, treatment reputation, cure effectiveness, etc.) and the price level in the casebook. Patient happiness is adjusted proportionally. Hospital reputation can only be impacted if the price distortion reaches thresholds (e.g. a 30% gap) and randomly (1 out of 10); this way it's not too impacting.

One patch had already been discussed earlier. This change is partially based on this conversation: https://github.com/CorsixTH/CorsixTH/pull/266

It's quite difficult to know exactly how to adjust all the variables (weight/influence of each variable, randomness, reputation changes, etc.) After testing with a few different settings, when I finally got something nice and more or less balanced, I run the following simulations. Let me know if you find a nice setting. I would be interesting to test it with poorly skilled staff.

Initial:
- Reputation: 582
- Balance: +44,000
- Cured: 131
- Highly skilled staff
- Wage requests granted

If I let it run for one year without any manual intervention and with different price levels.

All prices at 50%:
- Reputation: 704
- Balance: -26,000
- Cured: 233
- Price distortion: -0.70 to -0.50

All prices at 100%:
- Reputation: 652
- Balance: +44,000
- Cured: 225
- Price distortion: -0.45 to -0.25

All prices at 200%:
- Reputation: 617
- Balance: +171,000
- Cured: 231
- Price distortion: 0.25 to 0.45

I didn't look how it works exactly but it seems like reputation does not influence the number of patients coming to the hospital. That could be for another patch.

<b>Testing saved games:</b>
https://www.dropbox.com/sh/zfb57sosk92nbc9/AABcO6Q2Q70WDp1BORASEB9Sa/prices_impact?dl=0

PR #971, opened 2016-01-31
[WIP] Drug casebook price impact [2.2]

benckx:
More info [on the previous PR](https://github.com/CorsixTH/CorsixTH/pull/313)

### DIFF
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -46,6 +46,7 @@ function Patient:Patient(...)
   self.vaccination_candidate = false
   -- Has the patient passed reception?
   self.has_passed_reception = false
+  self.go_home_reasons = {CURED = "cured", KICKED = "kicked", OVER_PRICED = "overpriced"}
 
   -- Is the patient trying to get to the toilet? ("yes", "no", or "no-toilets")
   self.going_to_toilet = "no"
@@ -256,7 +257,7 @@ function Patient:treated() -- If a drug was used we also need to pay for this
       self:changeAttribute("happiness", 0.8)
       hospital:changeReputation("cured", self.disease)
       self.treatment_history[#self.treatment_history + 1] = _S.dynamic_info.patient.actions.cured
-      self:goHome("cured")
+      self:goHome(self.go_home_reasons.CURED)
       self:updateDynamicInfo(_S.dynamic_info.patient.actions.cured)
     end
   end
@@ -504,7 +505,7 @@ function Patient:goHome(reason)
     self:setHospital(nil)
     return
   end
-  if reason == "cured" then
+  if reason == self.go_home_reasons.CURED then
     self:setMood("cured", "activate")
     if not self.is_debug then
       hosp:changeReputation("cured", self.disease)
@@ -516,7 +517,7 @@ function Patient:goHome(reason)
         hosp.emergency.cured_emergency_patients = hosp.emergency.cured_emergency_patients + 1
       end
     end
-  elseif reason == "kicked" then
+  elseif reason == self.go_home_reasons.KICKED then
     self:setMood("exit", "activate")
     if not self.is_debug then
       hosp:changeReputation("kicked", self.disease)
@@ -525,7 +526,7 @@ function Patient:goHome(reason)
       local casebook = hosp.disease_casebook[self.disease.id]
       casebook.turned_away = casebook.turned_away + 1
     end
-  elseif reason == "over_priced" then
+  elseif self.go_home_reasons.OVER_PRICED then
     self:setMood("sad_money", "activate")
     if not self.is_debug then
       hosp:changeReputation("over_priced", self.disease)
@@ -575,7 +576,7 @@ function Patient:tickDay()
   if self.waiting then
     self.waiting = self.waiting - 1
     if self.waiting == 0 then
-      self:goHome("kicked")
+      self:goHome(self.go_home_reasons.KICKED)
       if self.diagnosed then
         -- No treatment rooms
         self:updateDynamicInfo(_S.dynamic_info.patient.actions.no_treatment_available)
@@ -622,7 +623,7 @@ function Patient:tickDay()
     if math.random(1,30) == 1 then
       self:updateDynamicInfo(_S.dynamic_info.patient.actions.fed_up)
       self:setMood("sad2", "deactivate")
-      self:goHome("kicked")
+      self:goHome(self.go_home_reasons.KICKED)
     end
   elseif self.attributes["health"] >= 0.14 and self.attributes["health"] < 0.18 then
     self:setMood("sad2", "deactivate")
@@ -1126,6 +1127,9 @@ function Patient:afterLoad(old, new)
     else
       self.going_to_toilet = "no"
     end
+  end
+  if old < 102 then
+    self.go_home_reasons = {CURED = "cured", KICKED = "kicked", OVER_PRICED = "overpriced"}
   end
   Humanoid.afterLoad(self, old, new)
 end
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -162,11 +162,9 @@ end
 -- Modifies the diagnosis progress of a patient.
 -- incrementValue can be either positive or negative.
 function Patient:modifyDiagnosisProgress(incrementValue)
-  if self.hospital ~= nil then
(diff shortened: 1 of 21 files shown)


benckx:
Next changes I'm planning to do:

- `SAVEGAME_VERSION` flag will have to be updated (the current value is 108, and our checks are at 102).
- I think it would be nice to be able to shift all prices to 50% or to 200% from the cheat UI or something, so it can be tested more easily. Doing so via the normal UI is bit a tedious.
- Move the `go_home_reasons` enum to class level.

I'd also like to run a few simulations and post the results.

CorsixTH/Lua/entities/patient.lua
```diff
@@ -46,12 +46,15 @@ function Patient:Patient(...)
...
   -- Has the patient passed reception?
   self.has_passed_reception = false
+  self.go_home_reasons = {CURED = "cured", KICKED = "kicked", OVER_PRICED = "overpriced"}
 
   -- Is the patient trying to get to the toilet? ("yes", "no", or "no-toilets")
   self.going_to_toilet = "no"
 end
 
 function Patient:onClick(ui, button)
+  self.over_priced = true
```
benckx: @sadger: Can we drop this variable now? It looks like it's only there for test/log purpose?

sadger: Oh yeah, oops, you can, didn't mean to put that in my commits, hope it didnt ruin your testing.

benckx: Haha, no it didn't. I'm just reviewing the code right now.

CorsixTH/Lua/entities/patient.lua
```diff
@@ -245,16 +249,17 @@ function Patient:treated() -- If a drug was used we also need to pay for this
...
       self.attributes["health"] = 1
-      self:changeAttribute("happiness", 0.8)
-      hospital:changeReputation("cured", self.disease)
-      self.treatment_history[#self.treatment_history + 1] = _S.dynamic_info.patient.actions.cured
-      self:goHome(self.cured)
       self:updateDynamicInfo(_S.dynamic_info.patient.actions.cured)
+      self.hospital:msgCured()
+
+      -- If the patient refused to pay they are on the way home already
+      if not self.going_home then
```
sadger: This seems like a bit of a hacky solution to me. The patient should be sent home in this function rather than before ideally. Just an idea, maybe a parameter to treated might work  treated(boolean : happyToPay) or maybe that just adds extra complexity. Not sure, open to ideas.

benckx: It's a good point. That would be more consistent.
Maybe `Hospital:receiveMoneyForTreatment(patient)` should return a boolean, so we know whether the patient paid or not and hence for which reason he's sent home.
Also, I'm wondering if statistic handling (`hospital.num_cured + 1`) shouldn't be managed in one place only (in `goHome()` only and not in `treated()` anymore).

CorsixTH/Lua/entities/patient.lua
```diff
@@ -23,6 +23,7 @@ class "Patient" (Humanoid)
 
 ---@type Patient
 local Patient = _G["Patient"]
+local go_home_reasons = {CURED = "cured", KICKED = "kicked", OVER_PRICED = "overpriced"}
```
TheCycoONE: There isn't any reason for this table to exist in lua, unlike C it doesn't add any type safety since you can refer to unset keys, and also unlike C it's not a free call. Ultimately you're saying go_home_reasons["CURED"] = "cured". You could achieve a little bit of type safety with something like go_home_reasons.CURED = {}, but it's probably best to leave the reason as a string and document the possible values on goHome().

CorsixTH/Lua/entities/patient.lua
```diff
@@ -161,9 +163,9 @@ end
 -- Modifies the diagnosis progress of a patient.
 -- incrementValue can be either positive or negative.
 function Patient:modifyDiagnosisProgress(incrementValue)
-  self.diagnosis_progress = math.min(self.hospital.policies["stop_procedure"],
-    self.diagnosis_progress + incrementValue)
-  self.diagnosis_progress = math.max(0.000, self.diagnosis_progress)
+    self.diagnosis_progress = math.min(self.hospital.policies["stop_procedure"],
```
TheCycoONE: This indentation change looks wrong?

CorsixTH/Lua/entities/patient.lua
sadger: I guess this makes sense, the "enum" only serves to self-document the reasons for going home, annotating the method would be fine then.

CorsixTH/Lua/entities/patient.lua
sadger: I can see both of those things working, go for it.

CorsixTH/Lua/entities/patient.lua
benckx: I see. I'll drop this then.

CorsixTH/Lua/entities/patient.lua
benckx: Looks like `dealtWithPatient()` would be impacted as well (it calls `receiveMoneyForTreatment()`); so we could also possibly send the patients home from there?
It's actually quite a big change, I'll have a look at it later and see if we can clarify the workflow and the encapsulation.

sadger: I tried to make this change myself before I think, needs a lot of thinking and restructure I'd agree, reluctant to allow this change into the code base as it is however. I'm sure we can find an agreeable solution.

benckx:
- Moved all payments logic out of `treated()`, so everything is handled consistently within `dealtWithPatient()`.
- `dealtWithPatient()` now decides whether patient goes home (either because he's cured, either because he refuses to pay).
- Handle all statistics in `goHome()` rather than in `treated()` for the sake of consistency.

I think this way the role of each function is a bit more clear.

sadger:
Moving forward with this. It's looking great and I think its very nearly ready! Can we try and get the commits squashed down, to one or two commits for merging and also its easier to review in this state. I know you wanted to do more simulations, but I mentioned it to the others in IRC and we can always balanced it after merging and after its had some play testing. What do you think?

benckx:
I was considering one last commit:

- The name `Patient:getDiseaseId()` bothers me a bit. I was thinking `getCasebookId()` or something (sometimes it returns "Invisiblity", sometimes "General Practice").
- Move `Patient:computePriceDistortion()` as a local method.
- Review the documentation.

Otherwise it looks good to me.
Regarding the balance I think it's already quite consistent.

benckx:
This is the last commit I wanted to do. In my opinion, after all feedback I got in the previous PR and more recently from @sadger, it's now ready for merging. However I remain open to further suggestions.

sadger:
The code logic and comments looks good. There is a general consensus that you have overused local functions however, don't really need to make a function local if you are only going to call it once, probably just pull that code out into the member function its nested into.  If you do feel that a function HAS to be local that's probably fine, can either keep it nested or make it a top-level local function.  Will ask the other guys to suggest any things that need changing too so we can move forward with this if you are happy to do the changes.

TheCycoONE:
@benckx I've cloned your pr and rebased it on the latest master (https://github.com/TheCycoONE/CorsixTH/commit/price_impact). We want the commits in this pull request rewritten in such a way that each one introduces a small self contained feature or refactor. Since this is a difficult thing to do and you are new to git I have taken it upon myself to make the changes.
When I am done you can either reset your branch on my changes, or we can close this PR and I'll open a new one.
Your authorship will be maintained throughout.

TheCycoONE:
My progress can be monitored at https://github.com/TheCycoONE/CorsixTH/commits/price_impact2

At the moment I still have a ways to go. Some of the later fix up commits are particularly broad and will need to be broken up and sorted into their respective features.

benckx:
@sadger Maybe I should have used top-level local functions indeed. My goal was to make them "private" for encapsulation purpose.

@TheCycoONE Thanks for your support with that. We can proceed any way your prefer.

So if I understand correctly, the 26 commits need to be regrouped into a smaller number of little features (for example: "French translation", "Cheats", "treat() refactoring", "dealtWithPatient() integration")?

 Is there any guideline page somewhere that specifies the commit policy?

TheCycoONE:
@benckx this policy is mentioned at https://github.com/CorsixTH/CorsixTH/blob/master/CONTRIBUTING.txt#L75, though perhaps it could be more explicitly stated.

Policy regarding nested functions is not explicitly stated anywhere at this time, I will be adding some guidelines on that to the wiki now.

PR #972, opened 2016-02-01
[WIP] Small fixes and improvments of the Lua code & use of Luacheck

bobcanthelpyou:
I was looking for a Lua lint checker or something different and found luacheck https://github.com/mpeterv/luacheck/ 

I thing it could be useful to improve the Lua code a little bit.

I don't know if the changes are correct at all, i couldn't test them at all, yet.

I also have a local config file for Luacheck `.luacheckrc` i am not sure if it could be useful to add it to the git repo of CorsixTH.


CorsixTH/Lua/room.lua
```diff
@@ -259,7 +259,7 @@ function Room:testStaffCriteria(criteria, extra_humanoid)
     end
     return false
   else
-    for attribute, count in pairs(missing) do
+    if next(missing) ~= nil then
```
sadger: ``if next(missing) then`` be clearer here?

TheCycoONE:
@bobcanthelpyou if all the issues are cleaned up or ignored could you add this to the travis script? I think lualint would be most useful as part of the continuous integration. If there are still lots of open issues then we'll hold off.

bobcanthelpyou:
@TheCycoONE do you think about a special lualint, maybe https://github.com/philips/lualint ? Or could it be luacheck? Something else? 

I could try to add it to the travis script, if i am sure which one to use :)

Is there a common way to test my changes to the travis script somehow?

TheCycoONE:
@bobcanthelpyou sadly I only know two ways of testing travis changes, either spinning up an Ubuntu 12.04 server vm, or committing changes to a branch that's part of a pull request like this one and reading the output until you get it right. I use both methods.

I was thinking luacheck, since that's what you're familiar with.

TheCycoONE:
The travis script is failing due to https://github.com/travis-ci/travis-ci/issues/5548 we are awaiting some upstream direction.

bobcanthelpyou:
@TheCycoONE 
There is a suggestion to fix the travis script issue https://github.com/travis-ci/travis-ci/issues/5548#issuecomment-179244964
> Solution is to somehow install a newer version of luarocks.
Could you check if that fixes the issue with luasec?

TheCycoONE:
Rather frustrated with the ancient software stack travis is running, but yes I should be able to replace luarocks.

wolfy1339:
You can use Ubuntu Trusty containers on TravisCI
Docs: https://docs.travis-ci.com/user/trusty-ci-environment/
Other project using Trusty: [ZNC](https://github.com/znc/znc),

TheCycoONE:
Thanks @wolfy1339 #977 switches to Trusty containers.
Sadly at least for the last run it was really slow to spin up, losing all the advantage I gained by disabling sudo earlier, but hopefully that's something that will improve with time.

Anyway once that pr is merged you should be able to continue @bobcanthelpyou

houqp:
@TheCycoONE , if you still want to use container build, try https://github.com/torch/luajit-rocks. That's what KOReader uses to install the new luarocks.

TheCycoONE:
Thanks for the suggestion @houqp for now we'll move to Trusty, but we won't use sudo so once they have containers on Trusty we can use them.

CorsixTH/Lua/room.lua
TheCycoONE: I think it's good as is.

bobcanthelpyou:
i added luacheck to travis ci, i am not sure if it is okay like it is now

There are a lot of warnings which also could be ignored. The complete list of warnings is available here: http://luacheck.readthedocs.org/en/stable/warnings.html

Maybe the following warnings could be ignored but fixing them could also make the code shorter and easier to read
* 211 - 213 Unused variables / arguments
* 421 - 433 Shadowing

What do you think?

TheCycoONE:
I like it a lot. Obviously we can't merge this PR as is since it breaks the travis build, but I think all those unused and shadow warnings are worth correcting if you're willing to do it.

bobcanthelpyou:
I could do that, but it could take some time ^^

Also i am not sure, if i possible start to replace several other stuff, e.g. i already found some other loops which could be easily replaced by conditions. (but of cause that could be done later)

How should i commit the changes? One commit per function, per file, per directory or per warning type? Should i put the changes in this pull request, in a different pull requests or all in one? Is there a common way to do a lot of small changes all over the code?

TheCycoONE:
Lets keep it in this PR as much as possible. If you would like assistance with a particular warning then comment on it here and when someone else agrees to help they should link back here in their PR.

As for commits in this case, it's up to you what you consider a logical grouping that makes review straight forward - there's no hard fast rules for this scenario; in general I would suggest that any larger refactors be isolated in their own commit, and many smaller similar refactors (e.g. all the cases where you remove ```local``` to avoid shadowing) be grouped in a single commit.

Any remaining references to _DECODA should go, I believe the same with _MAP_EDITOR.

TheCycoONE:
One last note: I would actually avoid doing larger refactors to clean up things in this PR. Just do the bare minimum to clean up those warnings - that will make merging simpler if other people make large changes to the lua code at the same time. (I know that there are other significant refactoring projects underway).

TheCycoONE:
@bobcanthelpyou are there any of the remaining warnings you would like a hand on?

TheCycoONE:
@bobcanthelpyou have you given up on this? I'd really like to get luacheck support merged early after 0.60, even if it means we have to disable a lot of errors to start with. (There can be an ongoing issue to enable them again).

bobcanthelpyou:
@TheCycoONE I will continue asap, sorry if i am a little inactive atm. There are some commits which should fix 'critical' problems like undefined global variables, i am not sure if these fixes should be moved into single pull request to commit them into the 0.60 release (but maybe they where already fixed, sorry didn't had a look at all the code changes lately). I could try to rebase that stuff to the latest code changes at the weekend. Maybe i could then say what might be critical and should be fixed in the 0.60 release.

Alberth289346:
Doing last minute fixes is a big risk, as there is little time to fix any errors, before 0.60 gets spread to everywhere. Doing this after 0.60 is much easier.
I'd also say that your fixes are useful, but the really bigger thing is that we get regular checking by a lint program with the automagic builds, which is much more valuable and longer lasting.

Alberth289346:
Hello @bobcanthelpyou ,

With 0.60 released, the road seems clear to move this patch towards merging. I think it's useful to have this check in the project for our merges.

Are you still interested in doing that, or shall I have a go?

CorsixTH/Lua/humanoid_actions/seek_reception.lua
```diff
@@ -63,7 +63,7 @@ local function action_seek_reception_start(action, humanoid)
 
     -- We don't want patients which have just spawned to be joining the queue
     -- immediately, so walk them closer to the desk before joining the queue
-    if can_join_queue_at(humanoid, humanoid.tile_x, humanoid.tile_y, x, y) then
```
Alberth289346: You can't just remove arguments, also delete the function parameters at line 21

CorsixTH/Lua/utility.lua
```diff
@@ -279,7 +279,7 @@ function compare_tables(t1, t2)
     if t2[k] ~= v then return false end
   end
   local count2 = 0
-  for k, v in pairs(t2) do
+  for _, _ in pairs(t2) do
```
Alberth289346: This looks weird, simpler would be  local count2 = #t2  and removal of the loop, as far as I can see

.luacheckrc
```diff
@@ -0,0 +1,36 @@
+stds.lpeg = require "lpeg"
+globals = { "_A", "_S", "action_queue_leave_bench", "class", "compare_tables",
```
Alberth289346: would be better to add this in the first commit, so new files are all added at the same time

CorsixTH/Lua/world.lua
```diff
@@ -534,7 +534,6 @@ function World:spawnVIP(name)
 
   vip.enter_explosions = hospital.num_explosions
 
-  local spawn_point = self.spawn_points[math.random(1, #self.spawn_points)]
```
Alberth289346: Removing this one is wrong, it's used at the line below
It's also not defined above, it seems

(there was one such line duplicate, but that has been removed already, maybe your patch rebase broke made the wrong guess here?)

CorsixTH/Lua/dialogs/fullscreen.lua
```diff
@@ -75,28 +75,28 @@ function UIFullscreen:onChangeResolution()
   end
 end
 
-function UIFullscreen:draw(canvas, x, y)
+function UIFullscreen:draw(canvas, dx, dy)
```
Alberth289346: dx and dy typically express "difference x" or "difference y", rather use xpos, or x_pos, or something like that, as it expresses "position", much what x or y does.

CorsixTH/Lua/dialogs/resizable.lua
```diff
@@ -83,32 +83,32 @@ function UIResizable:setColour(colour)
   self.background_panel:setColour(colour)
 end
 
-function UIResizable:draw(canvas, x, y)
+function UIResizable:draw(canvas, dx, dy)
```
Alberth289346: Here, use something that expresses "position" too

CorsixTH/Lua/dialogs/resizables/lua_console.lua
```diff
@@ -115,10 +115,10 @@ function UILuaConsole:buttonExecute()
     print(err)
   else
     print("Executing UserFunction...")
-    local s, err = pcall(func)
+    local s, err_func = pcall(func)
```
Alberth289346: "func" expresses function, so the code here tries to print a function, it seems
better make it "err_message"  or "err_msg" or so.

CorsixTH/Lua/entities/patient.lua
```diff
@@ -199,7 +199,7 @@ function Patient:completeDiagnosticStep(room)
     local divisor = math.random(1, 3)
     local attn_detail = room.staff_member.profile.attention_to_detail / divisor
     local skill = room.staff_member.profile.skill / divisor
-    local diagnosis_bonus = (attn_detail + 0.4) *  skill
+    diagnosis_bonus = (attn_detail + 0.4) *  skill
```
Alberth289346: this is actually a bug fix :)

CorsixTH/Lua/graphics.lua
```diff
@@ -336,10 +336,10 @@ function Graphics:onChangeLanguage()
   self.load_info = {} -- Any newly made objects are temporary, and shouldn't
                       -- remember reload information (also avoids insertions
                       -- into a table being iterated over).
-  for object, load_info in pairs(load_info) do
+  for object, load_i in pairs(load_info) do
```
Alberth289346: Abbreviating the name throws away information, in this case "pairs(load_info)" looks wrong (iterating over a non-plural value looks weird.
I would suggest you rename that variable to "load_infos" instead to express it's a container with many values (and the self.load_info is thus also wrong, but better not touch that, it could be used in many places)

CorsixTH/Lua/map.lua
```diff
@@ -228,11 +228,11 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
     if level == "" then
       i, objects = self.th:loadBlank()
     else
-      local data, errors = self:getRawData(level)
+      local data, errors_map_editor = self:getRawData(level)
```
Alberth289346: perhaps errors_level ?

CorsixTH/Lua/map.lua
```diff
@@ -244,11 +244,11 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
     self.level_intro = level_intro
     self.level_number = level
     self.map_file = map_file
-    local data, errors = self:getRawData(map_file)
+    local data, errors_other = self:getRawData(map_file)
```
Alberth289346: as this looks more map-ish, eg "errors_map" for example?

CorsixTH/Lua/audio.lua
```diff
@@ -358,14 +358,13 @@ function Audio:playSoundsAtEntityInRandomSequenceRecursionHandler(sounds, entity
...
     if self:canSoundsBePlayed() then
-      local _, warning
       local x, y = Map:WorldToScreen(entity.tile_x, entity.tile_y)
       local dx, dy = entity.th:getPosition()
       x = x + dx - self.app.ui.screen_offset_x
       y = y + dy - self.app.ui.screen_offset_y
 
       self.played_sound_callbacks[tostring(self.unused_played_callback_id)] = sound_played_callback
-      _, warning = self.sound_fx:play(sounds[math.random(1,#sounds)],
+      _, _ = self.sound_fx:play(sounds[math.random(1,#sounds)],
```
Alberth289346: Just delete the assignment completely, ie only call the function "self.sound_fx:play(...)"

CorsixTH/Lua/persistance.lua
```diff
@@ -38,13 +37,8 @@ if _G._VERSION == "Lua 5.2" or _G._VERSION == "Lua 5.3" then
     end
     return val
   end
-
-  th_getupvalue = function(f, n)
-    return debug.getupvalue(f, n + 1)
-  end
 else
   th_getfenv = debug.getfenv;
-  th_getupvalue = debug.getupvalue;
```
Alberth289346: this one is tricky to delete; it looks like it is used for debugging, so the function may get called from something outside this program (which luacheck probably doesn't understand), better keep this

CorsixTH/Lua/app.lua
```diff
@@ -1084,7 +1084,7 @@ end
 
 function App:checkInstallFolder()
   self.fs = FileSystem()
-  local status
+  local status, _
```
Alberth289346: we need to declare _  ?  seems a bit silly, imho. Why?

CorsixTH/Lua/dialogs/patient.lua
```diff
@@ -184,7 +184,7 @@ function UIPatient:draw(canvas, x_, y_)
 
     local dx_h = hor_length / size
     local line = nil -- Make a line the first time we find a non-nil value.
-    for i = 1, size do
+    for _, size do
```
Alberth289346: should be  for _ = 1, size do

CorsixTH/Lua/utility.lua
bobcanthelpyou: Would it be better to remove count1, too and use
```
if #t1 ~= #t2 then return false end
```
instead of
```
if count1 ~= count2 then return false end
```

CorsixTH/Lua/app.lua
bobcanthelpyou: Without:
```
./Lua/app.lua:1089: assign to undeclared variable '_'
stack traceback:
	[C]: in function 'error'
	./Lua/strict.lua:35: in function <./Lua/strict.lua:31>
	./Lua/app.lua:1089: in function 'checkInstallFolder'
	./Lua/app.lua:114: in function 'init'
	./CorsixTH.lua:130: in main chunk
	[C]: ?
```

CorsixTH/Lua/rooms/staff_room.lua
```diff
@@ -56,8 +56,6 @@ function StaffRoom:onHumanoidEnter(humanoid)
       humanoid:setNextAction({name = "use_staffroom"})
       self.door.queue.visitor_count = self.door.queue.visitor_count + 1
     end
-  else
```
Alberth289346: Alternatively, add the 'else' in comment too:
```` lua
-- else: Other humanoids shouldn't be entering, so don't worry about them
````

File: CorsixTH/Lua/graphics.lua [hunk 18]
```diff
@@ -336,10 +335,10 @@ function Graphics:onChangeLanguage()
   self.load_info = {} -- Any newly made objects are temporary, and shouldn't
```
Alberth289346: This table should be renamed "load_infos"

CorsixTH/Lua/utility.lua
Alberth289346: Definitely better, I agree

CorsixTH/Lua/graphics.lua
TheCycoONE: info is the plural of info

Alberth289346: there is also a singular load_info variable there. Perhaps use load_info_table here instead?

CorsixTH/Lua/persistance.lua
TheCycoONE: It can't be called outside the program, it's local.

PR #976, opened 2016-02-04
[RDY] Fix #745 and enhance numpad / keypad support

### LINKED ISSUE #745: Save name can change game speed
I have noticed that if you were to use a number in your saves name, as you type it your game can speed up i.e. save the game as level 5 and the game speeds up to fastest speed.

bobcanthelpyou:
This should fix #745

### DIFF
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -211,6 +211,10 @@ local scroll_keys = {
   right = {x =  10, y =   0},
   down  = {x =   0, y =  10},
   left  = {x = -10, y =   0},
+  ["keypad 8"] = {x =   0, y = -10},
+  ["keypad 6"] = {x =  10, y =   0},
+  ["keypad 2"] = {x =   0, y =  10},
+  ["keypad 4"] = {x = -10, y =   0},
 }
 
 function GameUI:updateKeyScroll()
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -87,6 +87,24 @@ function UI:initKeyAndButtonCodes()
     end
   end
 
+  local keypad = {
+    ["Keypad 0"] = "insert",
+    ["Keypad 1"] = "end",
+    ["Keypad 2"] = "down",
+    ["Keypad 3"] = "pagedown",
+    ["Keypad 4"] = "left",
+    ["Keypad 6"] = "right",
+    ["Keypad 7"] = "home",
+    ["Keypad 8"] = "up",
+    ["Keypad 9"] = "pageup",
+    ["Keypad ."] = "delete",
+  }
+
+  -- Apply keypad remapping
+  for k, v in pairs(keypad) do
+    key_remaps[key_norms[k]] = key_norms[v]
+  end
+
   self.key_remaps = key_remaps
   self.key_to_button_remaps = key_to_button_remaps
 
@@ -498,6 +516,8 @@ end
 function UI:onKeyDown(rawchar, modifiers, is_repeat)
   local handled = false
   -- Apply key-remapping and normalisation
+  rawchar = string.sub(rawchar,1,6) == "Keypad" and
+            modifiers["numlockactive"] and string.sub(rawchar,8) or rawchar
   local key = rawchar:lower()
   do
     local mapped_button = self.key_to_button_remaps[key]
@@ -507,6 +527,9 @@ function UI:onKeyDown(rawchar, modifiers, is_repeat)
     end
     key = self.key_remaps[key] or key
   end
+
+  -- Remove numlock modifier
+  modifiers["numlockactive"] = nil
 
   -- If there is one, the current textbox gets the key.
   -- It will not process any text at this point though.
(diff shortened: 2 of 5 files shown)


TheCycoONE:
Could you clarify what additional is needed.

bobcanthelpyou:
Oh sorry, there was no (additional) code to detect the numlock state, but SDL should return a value now.

I thought about enhance the keypad support at other parts of the game, too (moving with the arrow keys on the map when numlock is inactive and change the speed of the game only when numlock is active, only). But now it is only for keypad interaction and text input boxes.

Should i fix the other stuff in this PR, too?

TheCycoONE:
I think the scope of this PR is fine as is, thanks.

TheCycoONE:
Please rebase on the latest master so that this PR can run against the new working Travis (as well as the comment fix I mentioned above).

TheCycoONE:
I just tested this and it doesn't appear to fix #745 

Pressing the number keys during the save dialog still changes the game speed for me.

TheCycoONE:
Alright, it appears to depend on the initial state of numlock. If I start the game with numlock off and turn it on in game then everything works, but if it's on when the game starts then ti doesn't, and numlockactive doesn't turn on until I turn off numlock.

TheCycoONE:
http://forums.libsdl.org/viewtopic.php?p=43816 explains it. An SDL bug.

TheCycoONE:
I tested with SDL 2.0.4 and I'm getting some very strange behavior. When I turn numlock on I can't change the game speed with any of the number keys even when the dialog isn't showing. When numlock is off I can change the game speed with the top numbers only.

bobcanthelpyou:
As i said the keypad support need some changes at other parts of the game. But i will do that in a different PR.

TheCycoONE:
I think it needs to be part of this PR as the current commit causes a regression not just in keypad support but regular number support. I believe a quick workaround would be to remove numlockactive before https://github.com/CorsixTH/CorsixTH/blob/80836f6e24bdd22b9772b93ab1d7003e1639de55/CorsixTH/Lua/ui.lua#L579

bobcanthelpyou:
Did a quick workaround 

@TheCycoONE could you test it with SDL 2.0.4?

Would it be better to save the numlock state in a different table? Or is there a better way to handle the difference of 'permanent' modifiers (Numlock, Capslock, ..) compared to the 'only active while pressing a key' modifiers (Shift, Ctrl, Alt, ..)?

wolfy1339:
Please rebase against the latest revision of the master branch.

TheCycoONE:
I spent some time thinking about this, but I'm not sure the best way to handle numlock. I am leaning towards it being in that table but making the key bindings check more sophisticated to ignore particular modifiers, alternatively I thought of removing numlock from the table early in onKeyDown, and replacing Keypad 1-9 with the actual numbers in 'char' when numlock is on, and the arrow keys when it is not, kinda like we do with ```esc```.

CorsixTH/Lua/ui.lua
```diff
@@ -505,8 +505,17 @@ function UI:onKeyDown(rawchar, modifiers, is_repeat)
...
     end
-    key = self.key_remaps[key] or key
+    if string.sub(key,1,6) ~= "keypad" or modifiers["numlockactive"] == nil then
+      key = self.key_remaps[key] or key
+    end
+  end
+
+  -- If numlock is active and the keypad is used remove the Keypad prefix
+  if string.sub(rawchar,1,6) == "Keypad" and modifiers["numlockactive"] then
+    rawchar = string.sub(rawchar,8)
```
TheCycoONE: What happens if you put a condition here for ```numlockactive``` on and off, setting to numbers like you do when on, or arrows like you do in the file right now when off, then remove the condition you added above.

I think it would be more straight forward.

TheCycoONE:
@bobcanthelpyou How is this going? I would like it in 0.60 (even if that means dropping the movement support for now).

bobcanthelpyou:
@TheCycoONE please check the last commit. If possible also the behavior when numlock is active before the game is started and the input to text boxes (using the latest SDL version). With SDL 2.0.2 and numlock active before game start the keypad acts as arrow keys and number keys at the same time within text boxes.

TheCycoONE:
:+1: works well in SDL 2.0.4

CorsixTH/Lua/game_ui.lua
```diff
@@ -211,6 +211,10 @@ local scroll_keys = {
   right = {x =  10, y =   0},
   down  = {x =   0, y =  10},
   left  = {x = -10, y =   0},
+  ["keypad 8"] = {x =   0, y = -10},
```
TheCycoONE: Is this still necessary? Looks to me like it would be fixed by the key remap in ui.lua

TheCycoONE:
The version in `CorsixTH/Lua/api_version.lua` needs to be bumped (whenever the Lua depends on a new version of the C++).

bobcanthelpyou:
@TheCycoONE api version bumped, do i need to change something else? Is afterLoad required?

TheCycoONE:
Nothing else, I think this can be marked [RDY] if you do.

bobcanthelpyou:
Fixed the issue @Alberth289346 mentioned.

PR #980, opened 2016-02-10
[RDY] Cleanup the C++ object oriented code

TheCycoONE:
- Reduce scope of members from protected to private as allowed by current usage
- Use override keyword, not virtual, to mark overridden virtual functions on
derived classes.
- Use default virtual destructors on base classes when there were no resources
to free on the base.
- Mark derived classes as final where it is obvious that they should not be
subclassed.

### DIFF
--- a/CorsixTH/Src/iso_fs.h
+++ b/CorsixTH/Src/iso_fs.h
@@ -95,7 +95,7 @@ public:
     */
     bool getFileData(file_handle_t iFile, uint8_t *pBuffer);
 
-protected:
+private:
     struct _file_t
     {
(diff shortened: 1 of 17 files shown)


TheCycoONE:
Precisely I'm enforcing:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c133-avoid-protected-data

PR #981, opened 2016-02-11
[RDY] Add missing parentheses

Alberth289346:


CorsixTH/Lua/entities/object.lua
```diff
@@ -817,7 +817,8 @@ function Object.processTypeDefinition(object_type)
             solid_points[point[1] * 100 + point[2]] = point
           end
         end
-        for _, key in ipairs{"use_position_secondary", "finish_use_position", "finish_use_position_secondary"} do
+        for _, key in ipairs({"use_position_secondary", "finish_use_position",
+                              "finish_use_position_secondary"}) do
```
sadger: Indent me please

TheCycoONE:
sadly I can't think of anything nicer than matching to the parenthesis there, except that I would put all the values on separate lines like:

```lua
for _, key in ipairs({"use_position_secondary",
                               "finish_use_position",
                               "finish_use_position_secondary"}) do
```

Alberth289346:
fixed, in the way as suggested by TheCycoONE

PR #985, opened 2016-02-18
[RDY] Add github issue template

TheCycoONE:
GitHub recently introduced support for issue templates: https://github.com/blog/2111-issue-and-pull-request-templates

Try it out at: https://github.com/TheCycoONE/issue-template-test/issues/new

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,40 @@
+#### Describe the issue
+
+#### Steps to Reproduce
+1. *Load save game attached to this ticket*
+
+2.
+
+3.
+
+#### Save Game
+*It is often useful for us if you have a save game from shortly before the
+issue occurs that can be used to recreate the problem. Sometimes a save that
+shows the issue happening may be useful as well. As GitHub currently only
+supports uploading of images, you will have to upload your your savegames to
+an another source such as [Google Drive](https://drive.google.com),
+[Dropbox](https://dropbox.com), or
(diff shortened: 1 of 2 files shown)


Alberth289346:
looking good, maybe link to a wiki page where to find the gamelog ?
perhaps "try to find a savegame just before the issue happens?"
maybe add that to "steps to reproduce"

TheCycoONE:
Alright I've tried to use @Alberth289346's suggestions. Any other comments?

Alberth289346:
:+1:

TheCycoONE:
Pinging @Lego3

Lego3:
Ah, sorry. I follow updates in the ForkHub app for Android, and there the template doesn't work... so I forgot about it.

The only thing would be whether this same pull request should move the other file into the .github folder as well? (CONTRIBUTING) Or do we want README and CONTRIBUTING to stay in the root?

TheCycoONE:
@Lego3 My feeling is that ISSUE_TRACKER.md is a clear case of a file that has no meaning outside of GitHub so it definitely belongs in .github, README.md is probably in that category too so if it works under .github I'll move it there.

CONTRIBUTING.txt isn't as clear cut, I think it has just as much value read locally as it does in GitHub, so I'd prefer to leave it where it is.

Sensible?

Lego3:
Yep.

TheCycoONE:
I did a test in my repo with a readme.txt at the root and readme.md in .github and github ended up using the readme.txt, so I think this has to stay as is. Marking [RDY].

PR #989, opened 2016-02-28
fix copyright header of AnimView/rnc.cpp

a-detiste:
Hi,

This made the Debian FTP Masters angry :sweat: 
Hopefully the .zip archive does include a licens :smile:

### DIFF
--- a/AnimView/rnc.cpp
+++ b/AnimView/rnc.cpp
@@ -1,8 +1,32 @@
 /*
-    RNC decompression library. Copyright and ownership status unknown.
+    RNC decompression library.
 
-    Original code was from http://www.yoda.arachsys.com/dk/utils.html, but that
-    site nor the code itself states any ownership or copyright claims.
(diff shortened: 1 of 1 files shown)


a-detiste:
P.S.: that makes sense, the LICENSE.TXT is dated 2014; so years after inclusion of this code into Corsix-TH.

corsix:
Nice sleuthing, and nice to see licensing/copyright status resolved. Merging.

corsix:
Although it seems that possibly past-me possibly considered this in 2009 (see https://github.com/CorsixTH/CorsixTH/commit/7af98d5e1df814d3cafa276ff55ecd1e12ab6c01). Good past-me. It would also seem that past-me duplicated the RNC code to two places, and only updated the licensing notes in one of those places. Bad past-me.

TheCycoONE:
I fixed an invalid read in rnc.cpp a few years ago, I bet it's also still in the AnimView version. Consolidation will go in the TODO list.

PR #993, opened 2016-03-06
[RDY] Drug casebook price impact

TheCycoONE:
A cleanup of the patch queue from #971
There are few changes from the code in #971 rebased onto the latest master.

The pull request does a few things:
1. It introduces a response to drug casebook prices, patients will go home without paying if prices are too high and reputation is affected.
2. Makes hospital reputation impact the number of patients spawned
3. Fixes the death rate for under diagnosed patients
4. Refactors patient treatment, of particular note patient.hospital is now never nil.

### DIFF
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -214,6 +214,24 @@ function Patient:setHospital(hospital)
   end
 end
 
+--! Decide the ID of the disease or treatment that the patient is paying for.
+--!return (string or nil) Id of the disease or treatment, or nil if the Id could
+--! not be decided.
+function Patient:getTreatmentDiseaseId()
+  if self.diagnosed then
+    return self.disease.id
+  else
+    local room_info = self:getRoom()
+    if not room_info then
+      print("Warning: Trying to receive money for treated patient who is "..
+          "not in a room")
+      return nil
+    end
+    room_info = room_info.room_info
+    return "diag_" .. room_info.id
+  end
+end
+
 function Patient:treated() -- If a drug was used we also need to pay for this
   local hospital = self.hospital
   local amount = self.hospital.disease_casebook[self.disease.id].drug_cost or 0
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -487,11 +487,11 @@ end
 
 function Patient:goHome(cured)
   local hosp = self.hospital
-  if not hosp and self.going_home then
+  if self.going_home then
     -- The patient should be going home already! Anything related to the hospital
     -- will not be updated correctly, but we still want to try to get the patient to go home.
     TheApp.world:gameLog("Warning: goHome called when the patient is already going home")
-    self:setHospital(nil)
+    self:despawn()
     return
   end
   if not cured then
@@ -525,7 +525,7 @@ function Patient:goHome(cured)
   if room then
     room:makeHumanoidLeave(self)
   end
-  self:setHospital(nil)
+  self:despawn()
 end
 
 -- This function handles changing of the different attributes of the patient.
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -516,6 +516,25 @@ function Patient:tapFoot()
   end
 end
 
+-- Increments "cures" statistics
+function Patient:incrementCuredCounts()
+  local hosp = self.hospital
+  hosp.num_cured = hosp.num_cured + 1
+  hosp.num_cured_ty = hosp.num_cured_ty + 1
+  local casebook = hosp.disease_casebook[self.disease.id]
+  casebook.recoveries = casebook.recoveries + 1
+  if self.is_emergency then
+    hosp.emergency.cured_emergency_patients = hosp.emergency.cured_emergency_patients + 1
+  end
+end
+
+-- Increment "not_cured" statistics
+function Patient:incrementNotCuredCounts()
+  local hosp = self.hospital
+  hosp.not_cured = hosp.not_cured + 1
+  hosp.not_cured_ty = hosp.not_cured_ty + 1
+end
+
 --! Make the patient leave the hospital. This function also handles some
 --! statistics (number of cured/kicked out patients, etc.)
 --! The mood icon is updated accordingly. Reputation is impacted accordingly.
@@ -544,21 +563,14 @@ function Patient:goHome(reason)
     if hosp.num_cured < 1 then
       self.world.ui.adviser:say(_A.information.first_cure)
     end
-    hosp.num_cured = hosp.num_cured + 1
-    hosp.num_cured_ty = hosp.num_cured_ty + 1
-    local casebook = hosp.disease_casebook[self.disease.id]
-    casebook.recoveries = casebook.recoveries + 1
-    if self.is_emergency then
-      hosp.emergency.cured_emergency_patients = hosp.emergency.cured_emergency_patients + 1
-    end
+    self:incrementCuredCounts()
     self:updateDynamicInfo(_S.dynamic_info.patient.actions.cured)
     self.hospital:msgCured()
   elseif reason == "kicked" then
     self:setMood("exit", "activate")
     if not self.is_debug then
       hosp:changeReputation("kicked", self.disease)
-      hosp.not_cured = hosp.not_cured + 1
-      hosp.not_cured_ty = hosp.not_cured_ty + 1
+      self:incrementNotCuredCounts()
       local casebook = hosp.disease_casebook[self.disease.id]
       casebook.turned_away = casebook.turned_away + 1
     end
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -269,14 +269,7 @@ function Patient:treated() -- If a drug was used we also need to pay for this
   hospital:updatePercentages()
 
   if self.is_emergency then
-    local killed = hospital.emergency.killed_emergency_patients
-    local cured = hospital.emergency.cured_emergency_patients
-    if killed + cured >= hospital.emergency.victims then
-      local window = hospital.world.ui:getWindow(UIWatch)
-      if window then
-        window:onCountdownEnd()
-      end
-    end
+    hospital:checkEmergencyOver()
   end
 end
 
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -263,12 +263,7 @@ end
 
 function Patient:treated() -- If a drug was used we also need to pay for this
   local hospital = self.hospital
-  local amount = self.hospital.disease_casebook[self.disease.id].drug_cost or 0
   hospital:receiveMoneyForTreatment(self)
-  if amount ~= 0 then
-    local str = _S.drug_companies[math.random(1 , 5)]
-    hospital:spendMoney(amount, _S.transactions.drug_cost .. ": " .. str)
-  end
 
   -- Either the patient is no longer sick, or he/she dies.
   if self:isTreatmentEffective() then
@@ -280,7 +275,7 @@ function Patient:treated() -- If a drug was used we also need to pay for this
   end
 
   hospital:updatePercentages()
-
+  hospital:paySupplierForDrug(self)
   if self.is_emergency then
     hospital:checkEmergencyOver()
   end
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -242,38 +242,28 @@ function Patient:treated() -- If a drug was used we also need to pay for this
   end
 
   -- Either the patient is no longer sick, or he/she dies.
-
-  local cure_chance = hospital.disease_casebook[self.disease.id].cure_effectiveness
-  cure_chance = cure_chance * self.diagnosis_progress
-  if self.die_anims and math.random(1, 100) > cure_chance then
-    self:die()
-  else
-    -- to guess the cure is risky and the patient could die
-    if self.die_anims and math.random(1, 100) > (self.diagnosis_progress * 100) then
-      self:die()
-    else
-      if hospital.num_cured < 1 then
-        self.world.ui.adviser:say(_A.information.first_cure)
-      end
-      self.hospital.num_cured = hospital.num_cured + 1
-      self.hospital.num_cured_ty = hospital.num_cured_ty + 1
-      self.hospital:msgCured()
-      local casebook = hospital.disease_casebook[self.disease.id]
-      casebook.recoveries = casebook.recoveries + 1
-      if self.is_emergency then
-        self.hospital.emergency.cured_emergency_patients = hospital.emergency.cured_emergency_patients + 1
-      end
-      self.cured = true
-      self.infected = false
-      self:setMood("cured", "activate")
-      self.world.ui:playSound("cheer.wav") -- This sound is always heard
-      self.attributes["health"] = 1
-      self:changeAttribute("happiness", 0.8)
-      hospital:changeReputation("cured", self.disease)
-      self.treatment_history[#self.treatment_history + 1] = _S.dynamic_info.patient.actions.cured
-      self:goHome(self.cured)
-      self:updateDynamicInfo(_S.dynamic_info.patient.actions.cured)
+  if self:isTreatmentEffective() then
+    if hospital.num_cured < 1 then
+      self.world.ui.adviser:say(_A.information.first_cure)
     end
+    self.hospital.num_cured = hospital.num_cured + 1
+    self.hospital.num_cured_ty = hospital.num_cured_ty + 1
+    self.hospital:msgCured()
+    local casebook = hospital.disease_casebook[self.disease.id]
+    casebook.recoveries = casebook.recoveries + 1
+    if self.is_emergency then
+      self.hospital.emergency.cured_emergency_patients = hospital.emergency.cured_emergency_patients + 1
+    end
+    self:cure()
+    self:setMood("cured", "activate")
+    self.world.ui:playSound("cheer.wav") -- This sound is always heard
+    self:changeAttribute("happiness", 0.8)
+    hospital:changeReputation("cured", self.disease)
+    self.treatment_history[#self.treatment_history + 1] = _S.dynamic_info.patient.actions.cured
+    self:goHome(self.cured)
+    self:updateDynamicInfo(_S.dynamic_info.patient.actions.cured)
+  else
+    self:die()
   end
 
   hospital:updatePercentages()
@@ -288,6 +278,23 @@ function Patient:treated() -- If a drug was used we also need to pay for this
       end
     end
   end
+end
+
+--! Either the patient is cured, or he/she dies.
+--!return (boolean) True if cured, false if died.
+function Patient:isTreatmentEffective()
+  local cure_chance = self.hospital.disease_casebook[self.disease.id].cure_effectiveness
+  cure_chance = cure_chance * self.diagnosis_progress
+
+  local die = self.die_anims and math.random(1, 100) > cure_chance
+  return not die
+end
+
+--! Change patient internal state to "cured".
+function Patient:cure()
+  self.cured = true
+  self.infected = false
+  self.attributes["health"] = 1
 end
 
 function Patient:die()
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -268,6 +268,7 @@ end
 --! depending on whether they consider the price reasonable.
 function Patient:treatDisease()
   local hospital = self.hospital
+
   hospital:receiveMoneyForTreatment(self)
 
   -- Either the patient is no longer sick, or he/she dies.
@@ -284,6 +285,16 @@ function Patient:treatDisease()
   if self.is_emergency then
     hospital:checkEmergencyOver()
   end
+end
+
+--! Returns true if patient agrees to pay for the given treatment.
+--!param disease_id (string): The id of the disease to test
+function Patient:agreesToPay(disease_id)
+  local casebook = self.hospital.disease_casebook[disease_id]
+  local price_distortion = self:getPriceDistortion(casebook)
+  local is_over_priced = price_distortion > self.hospital.over_priced_threshold
+
+  return not (is_over_priced and math.random(1, 5) == 1)
 end
 
 --! Either the patient is cured, or he/she dies.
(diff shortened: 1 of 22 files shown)


Alberth289346:
Not sure my previous comments make a lot of sense at this stage, you may want to ignore them.

I would like to keep the actions as-is, reformatting is not part of the patch, and it will break my patch for no good reason since even in the complete patch I can already see lines that were not addressed.
The patch: (apply to 2c3956c3):
````
diff --git a/CorsixTH/Lua/room.lua b/CorsixTH/Lua/room.lua
index 553112d..a162753 100644
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -173,7 +173,7 @@ function Room:dealtWithPatient(patient)
       -- Increment diagnosis_progress, and send patient back to GP.
 
       patient:completeDiagnosticStep(self)
-      patient:queueAction { name = "seek_room", room_type = "gp" }
+      patient:queueAction{name = "seek_room", room_type = "gp"}
       if patient:agreesToPay() then
         self.hospital:receiveMoneyForTreatment(patient)
       else
@@ -186,7 +186,7 @@ function Room:dealtWithPatient(patient)
       local next_cure_room = patient.disease.treatment_rooms[patient.cure_rooms_visited + 1]
       if next_cure_room then
         -- Do not say that it is a treatment room here, since that check should already have been made.
-        patient:queueAction { name = "seek_room", room_type = next_cure_room }
+        patient:queueAction{name = "seek_room", room_type = next_cure_room}
       else
         -- Patient is "done" at the hospital
         if patient:isTreatmentEffective() then
@@ -206,8 +206,8 @@ function Room:dealtWithPatient(patient)
       end
     end
   else
-    patient:queueAction { name = "meander", count = 2 }
-    patient:queueAction { name = "idle" }
+    patient:queueAction{name = "meander", count = 2}
+    patient:queueAction{name = "idle"}
   end
 
   if self.dealt_patient_callback then
diff --git a/CorsixTH/Lua/rooms/gp.lua b/CorsixTH/Lua/rooms/gp.lua
index 7d7896f..2feaea3 100644
--- a/CorsixTH/Lua/rooms/gp.lua
+++ b/CorsixTH/Lua/rooms/gp.lua
@@ -136,7 +136,7 @@ function GPRoom:dealtWithPatient(patient)
       patient:completeDiagnosticStep(self)
       if patient.diagnosis_progress >= self.hospital.policies["stop_procedure"] then
         patient:setDiagnosed(true)
-        patient:queueAction { name = "seek_room", room_type = patient.disease.treatment_rooms[1], treatment_r
oom = true }
+        patient:queueAction{name = "seek_room", room_type = patient.disease.treatment_rooms[1], treat
ment_room = true}
 
         self.staff_member:setMood("idea3", "activate") -- Show the light bulb over the doctor
         -- Check if this disease has just been discovered
````

**Comment by TheCycoONE** [comment]
Thanks @Alberth289346 

Big ugly is sorta what was left after I sorted out what I wanted as I didn't want to change the final result much without consulting the community first. Thanks for your patch for it. I was also planning to unlocal computePriceDistortion and make it return a value again (I don't see any reason for making it work via side effect?)

**Comment by Alberth289346** [comment]
+1 on making it return a value, "compute" doesn't cover its current functionality imho.

I also saw some points where I wanted to change things, but I think it's currently more important to get the patch in a useful shape

**Comment by TheCycoONE** [comment]
I've applied your action queue patch. Thanks for sorting through some of the individual changes. I too have questions about getCasebookId. @benckx could you comment on that, and maybe some of the other changes in that commit?

**Comment by TheCycoONE** [comment]
Reset on @Alberth289346 's hacking_ugly branch

**Comment by TheCycoONE** [comment]
Made all the early commit changes proposed by @Alberth289346

**Comment by benckx** [comment]
I used the name `casebookId` instead of `diseaseId` as it returns a casebook ref., not always a disease (it can be a disease id if patient is diagnosed, or a diagnostic room id).

I moved the logic out of `hospital` to `patient` because I needed the same logic within patient:
- To compute the distortion (needs the actual price level of the last treatment)
- From `goHome()`, to build the Adviser message in case patient leaves without paying

The printed message in the middle of `getCasebookId()` is not consistent anymore, it should have been removed.

I hope it clarifies.

**Comment by TheCycoONE** [comment]
Thanks @benckx 
I notice that in the various diagnosis are called diseases in the existing code (e.g. https://github.com/CorsixTH/CorsixTH/blob/c595120db48773fbfa123489bff5be80fa4bf39b/CorsixTH/Lua/diagnosis/diag_gp.lua#L22 ) too, and use the diag_*roomname* format which is what you're replicating here. Maybe the getDiseaseId name you didn't like was right after all. Anyway this is something I'll spend some time looking at, along with the 15257cd commit that Albert says is broken.

If you have anything else helpful to add looking over our patch set I'm happy to hear it.

**Comment by benckx** [comment]
I don't have anything to add since my last commit. I'll try to remain available in the coming days, in case you need more input from my side.

It looked kind of odd to me that the term "disease" was used for two different things (diseases and diag rooms). I actually had mentioned it in a comment before committing it and had had a quick word with @sadger on the IRC about it. Anyway I don't have such a complete view of the codebase as you do, so I'd trust your judgement on that.

I took note of your feedback regarding the documentation, local functions, commit conventions, etc. I'll take that into for my next contributions.

**Comment by TheCycoONE** [comment]
The problems with Hospital:recevieMoneyForTreatment were not in benchx's original PR (or my original commit set). I will resolve them later.

```Big ugly refactor``` commit is getting smaller, though there are some bug fixes in it that need to be moved back to the appropriate commits, and a couple other things to pull out of it.

There are also some outstanding white space issues.

I don't think there's anything left I don't know what to do with, so thanks everyone.

**Comment by TheCycoONE** [comment]
@MarkL1961 do you recall if you're suppose to receive money from a patient when they die from the cure? Without this patch you do and with it you do not.

**Comment by TheCycoONE** [comment]
I have yet to test this successfully, so it's very likely to crash spectacularly but all of the substantial changes I had wanted to make are done. I will be testing this later myself, but if anyone else wants to get a head start and review the patches or test the pull request it would be helpful at this stage.

**Comment by TheCycoONE** [comment]
When reviewing in GitHub remember you can add &w=1 to ignore white space changes. This is particularly helpful for ```Refactor common "if patient.disease" condition out```

**Comment by TheCycoONE** [comment]
Just pushed a new copy as I fixed a crash on curing a patient.

**Comment by Alberth289346** [comment]
https://github.com/Alberth289346/CorsixTH/tree/impact_fixes   contains a bunch of doc fixes + commit hash to apply them to.

Otherwise looks quite sane, although it's still difficult to follow

**Comment by MarkL1961** [comment]
@TheCycoONE   I don't remember that and who would pay for a fatal treatment?!

**Comment by Lego3** [comment]
From what I remember you did get paid, but I'm not sure. I do remember that we added the charge for everyone in CorsixTH on purpose though.

**Comment by MarkL1961** [comment]
Thinking about it the payment is made before the patient dies. not sure if the same happens with insurance, how would you know.

**Comment by TheCycoONE** [comment]
Alright, with these changes the check for the treatment being effective is done before payment so they can either go home paying, go home refusing to pay, or die from an ineffective cure. I'm not sure how much difference it makes, but I can change it.

**Comment by TheCycoONE** [comment]
@Alberth289346 sadly the commit comments were lost. In short:
I removed the addition of the local variables in the goHome commit. I'm ambivalent on them, so the smaller patch won out.
I reduced the likelihood of the adviser message by 10x. In my testing it did come up very often.
The initial value of quality is used, so I didn't make any changes there.
I also rebased your documentation fixes in. I actually had a more detailed comment for getTreatmentDiseaseId but I seem to have lost it.

**Comment by TheCycoONE** [comment]
I just made one more change. I didn't think Room was a good place for the treatment logic so I moved it back into Patient in a new function Patient:treatDisease().

**Comment by MarkL1961** [comment]
As I recall patients leaving without paying would happen between rooms or it always seemed that way.
Visit GP be told what the treatment was and on the way to treatment realise how much it was going to cost and it was then they left without paying.
In the case where the GP is pricey they make the decision after going to reception - so again before using the room.

**Comment by TheCycoONE** [comment]
@MarkL1961 that makes sense, though it's a pretty serious deviation from what @benckx did.

How do people feel about leaving this the way benchx implemented it, vs MarkL's suggestion that patients should leave because of pricing once they know what the next room is, not after treatment. To be honest I think what MarkL is saying makes a lot more sense.

**Comment by Lego3** [comment]
It makes more sense, but as I recall it it was after the treatment in TH. Isn't the advicer message "someone just left without paying for their treatment"?

**Comment by MarkL1961** [comment]
Tried it out on TH and it is before the room use not afterwards.

**Comment by TheCycoONE** [comment]
The prompt is: "A patient left without paying for X because it's too expensive." I suppose that could go either way but it does imply they received service and refused to pay.

**Comment by TheCycoONE** [comment]
@MarkL1961 you mean to say as soon as they leave a room they refuse to pay for the next room. E.g like you said above, as soon as they leave the reception desk they might go home because GP is too expensive?

**Comment by MarkL1961** [comment]
@TheCycoONE  Yes, they make that decision before getting to the room.

If just arriving and you have increased the cost of GP they will decide or not it is too expensive after the reception desk.  If on the way back to GP it will happen as the leave the diagnosis room.  With treatment it will happen as they have been diagnosed and are leaving the GP office.  Some walk out and some pay it.  It did seem that those that were insured were more likely to pay, but the sample was small, so that may not have always been the case.  
It was as if they found out how much it was going to cost and decided then it was too much.

**Comment by TheCycoONE** [comment]
Alright, thanks @MarkL1961 
I'll take some time to look at how substantial the modifications will need to be to implement it that way. My feeling is that it will be fairly simple to implement.

goHome('over_priced') gets simplified because the patient would never be in a cured state when that happened. Likewise treatDisease()/treated() is simplified because the patient won't refuse to pay at that step. room.lua / gp.lua / reception_desk will be complicated slightly where the next room is set, and patient will need a getNextTreatementDiseaseId() or similar.

Locking the price from the time they leave the last room is something I'm more reluctant to do - It's such an edge case that I don't think it's worth preserving state for. All that means is that a micromanaging player early on could lower the price so the patient accepts it and then quickly raise it before they have to pay - the patient will still be unhappy (and it will harm reputation) about it but won't refuse to pay. I think that's acceptable - later in the game it wouldn't be a viable strategy anyway.

Thoughts?

**Comment by Lego3** [comment]
I agree to all points.

**Comment by MarkL1961** [comment]
Agreed
With reception first room is always the GP.  Does that make it any simpler to do?

**Comment by TheCycoONE** [comment]
@MarkL1961 One more question I'm hoping you might know the answer to as our resident Theme Hospital expert. Does a patient leaving do to over_priced affect the reputation of the patients disease, the specific diagnosis that's over priced, or just the overall hospital?

The way it's implemented in this patch at the moment, it's the patient's disease.

**Comment by TheCycoONE** [comment]
I've run into an issue with switching to the way the original does it. @benckx based a lot of (half the weight) of the expected price on the quality of the staff working in the treatment room. Now we won't know which staff will treat the patient at the point we have to decide on whether the price is appropriate.

I plan to use the patient happiness as part of the calculation instead and possibly and weight more towards disease reputation. Any thoughts? Particularly from @benckx as I know you did quite a bit of analysis on the old calculation.

**Comment by Lego3** [comment]
Is the final price currently based on the quality of the actual staff? I
was wondering whether we should remove that dependency in that case. As a
patient it would feel a little strange to just happen to get a much more
expensive treatment just because I went into the "wrong" room? Of course
it's better for the player if he/she trains the staff if your income
increases later on, but the speed of treatment also increases, so it could
be "overpowered"?

**Comment by TheCycoONE** [comment]
@Lego3 the price doesn't depend on the staff, the impact of the price on the patient does, but I've had to remove that anyway.

On my local system it is now a factor of the patient's happiness, the drug effectiveness, and the reputation of the treatment (which appears tied to the reputation of the hospital), where I've made drug reputation the most important factor.

benchx had:
50% staff quality
40% drug reputation
10% drug effectiveness

I currently have:
50% drug reputation
30% drug effectiveness 
20% patient happiness

Those ratios are open to adjustment - when I'm done other changes we can work with them. The problem with valuing happiness higher is that it's too prone to extremes (even starting at 100%) and we want values that tend to the center by default. I'm leaning to reducing it to 10% and pushing drug reputation higher, like 70%.

**Comment by MarkL1961** [comment]
@TheCycoONE   Like what you have done, the staff factor can affect reputation if you wanted.
Hospitals that employ a lot of juniors that are not training or on those lines.
Who actually treats you in he end is not something you would be aware of at the time or before the event.  Your happiness when you leave the hospital would or could affect the reputation if you were sent to multiple rooms for diagnosis because the Dr was unsure (a junior) and you would be happier if you had seen a consultant who knew what he was doing and you were treated quickly.

Some of those factors are already in as your happiness goes down the longer you are in the hospital, although it does improve when you are cured

All something for another time

**Comment by TheCycoONE** [comment]
Ok, I've pushed the changes I've made for the new behavior. I haven't yet merged them back into the earlier commits yet. If anyone wants to try it out go ahead.

**Comment by TheCycoONE** [comment]
I undid the change that caused treatment to only be payed for on success. I'm fairly confident they are meant to pay, then leave the room and die.

I also fixed what I think was a bug where the cure was no longer added to the transaction log.

**Comment by TheCycoONE** [comment]
I've also significantly reduced the noise this added to the adviser.

**Comment by TheCycoONE** [comment]
@Lego3 would you mind to review this. The rest of the active contributors have code in here.

**Comment by Lego3** [comment]
I am on a business trip, but I will check it out during easter.

File: CorsixTH/Lua/rooms/research.lua [hunk 1]
```diff
@@ -185,10 +185,7 @@ function ResearchRoom:commandEnteringPatient(patient)
         self.hospital.discover_autopsy_risk = self.hospital.discover_autopsy_risk + 10
       end
       if patient.hospital then
-        -- NB: Do not use patient:setHospital(nil) as that also causes despawning, which
-        -- has some bad effects if the autopsy machine is directly next to the room's door
         hosp:removePatient(patient)
-        patient.hospital = nil
```
Lego3: Do we know of a specific reason to set the hospital to nil, that is no longer there?

TheCycoONE:
We use to set nil for the hospital for all patients that were no longer active. This patch eliminates that entirely. despawn does not set nil either.

Lego3:
Yes, so as long as we are sure there are no places that rely on hospital being set to nil we're fine.

CorsixTH/Lua/entities/patient.lua
```diff
@@ -214,64 +214,106 @@ function Patient:setHospital(hospital)
...
+
+  return price_level - expected_price_level
+end
+
+--! Handle attempting to treat this patient
+--!
+--! If the treatment is effective the patient will be sent home, otherwise they
+--! will die. The patient may or may not agree to pay for the treatment
+--! depending on whether they consider the price reasonable.
+function Patient:treatDisease()
```
Lego3: I like the refactorings. The code is getting cleaner and easier to read.

CorsixTH/Lua/entities/patient.lua
```diff
@@ -485,24 +527,74 @@ function Patient:tapFoot()
   end
 end
 
-function Patient:goHome(cured)
+-- Increments "cures" statistics
+function Patient:countCured()
```
Lego3: The name implies that you just count something, but nothing is returned and it's an increment. Maybe something like incrementCuredCount?

CorsixTH/Lua/entities/patient.lua
```diff
@@ -485,24 +527,74 @@ function Patient:tapFoot()
...
+--! -"cured": When the patient is cured.
+--! -"kicked": When the patient is kicked anyway, either manually,
+--! either when no treatment can be found for her/him, etc.
+--! -"over_priced": When the patient decided to leave because he/she believes
+--! the last treatment is over-priced.
+--param disease_id (string): When the reason is "over_priced" this is the
+--! id of the disease/diagnosis that the patient considered over_priced
+function Patient:goHome(reason, disease_id)
   local hosp = self.hospital
   if not hosp and self.going_home then
```
Lego3: If we no longer set hospital to nil, is this a plausible error check?

CorsixTH/Lua/entities/patient.lua
```diff
@@ -485,24 +527,74 @@ function Patient:tapFoot()
...
+    self:updateDynamicInfo(_S.dynamic_info.patient.actions.cured)
+    self.hospital:msgCured()
+  elseif reason == "kicked" then
     self:setMood("exit", "activate")
     if not self.is_debug then
       hosp:changeReputation("kicked", self.disease)
-      self.hospital.not_cured = hosp.not_cured + 1
-      self.hospital.not_cured_ty = hosp.not_cured_ty + 1
-      local casebook = self.hospital.disease_casebook[self.disease.id]
+      hosp.not_cured = hosp.not_cured + 1
```
Lego3: As a further refactoring this could be put in something like incrementNotCuredCount? Like the countCured function

CorsixTH/Lua/entities/patient.lua
```diff
@@ -485,24 +527,74 @@ function Patient:tapFoot()
...
+    self.world.ui:playSound("cheer.wav") -- This sound is always heard
+    if not self.is_debug then
+      hosp:changeReputation("cured", self.disease)
+    end
+    if hosp.num_cured < 1 then
+      self.world.ui.adviser:say(_A.information.first_cure)
+    end
+    self:countCured()
+    self:updateDynamicInfo(_S.dynamic_info.patient.actions.cured)
+    self.hospital:msgCured()
```
Lego3: Future refactoring: In the spirit of having spelled out names, is messageCured too long? Or maybe even sendMessageCured?

CorsixTH/Lua/world.lua
```diff
@@ -1242,7 +1242,7 @@ end
 -- during the coming month.
 function World:updateSpawnDates()
   -- Set dates when people arrive
-  local no_of_spawns = math.n_random(self.spawn_rate, 2)
+  local no_of_spawns = math.n_random(self.spawn_rate * self:getReputationImpact(self:getLocalPlayerHospital()), 2)
```
Lego3: Hm, the previous design thought has been that the population variable of the hospital class should be used to vary how many patients a hospital gets. Since we have no AI it is at first 25 % and then always 50 %. A map maker knows that (with some variance of course) the spawn rate for his or her "World" is a certain number, and then it's up to the player to get as many of those people as possible. Changing the total number of spawns based on the local player's reputation like this gets us even further painted into a corner?

It's just a matter of semantics, but I would prefer to move the local player hospital dependency down to the next row just below here, to have them in the same place, and to let the map maker dictate more closely the spawn rate.

Lego3:
I've had a look at the code, see comments.

CorsixTH/Lua/entities/patient.lua
TheCycoONE: Should be removed, good catch.

CorsixTH/Lua/entities/patient.lua
TheCycoONE: Or neither one, countCured use to be called more than once, but not now that they can't be cured and over_priced. I'm fine with changing it either way.

CorsixTH/Lua/entities/patient.lua
TheCycoONE: Not sure what you mean here? Are you talking about renaming the goHome reason?

CorsixTH/Lua/world.lua
TheCycoONE: Agreed, there is reference in the original theme hospital to a world spawn rate and a pie chart of what is allocated to each hospital, we should do our hospital's spawn there.

CorsixTH/Lua/entities/patient.lua
Lego3: No, I was thinking about "msgCured", but it's not a part of this pull request anyway.

TheCycoONE:
Made the changes pointed out by @Lego3

TheCycoONE:
Rebased on the latest master to fix build checks.

CorsixTH/Lua/world.lua
```diff
@@ -1205,13 +1205,8 @@ function World:onEndMonth()
...
   end
 
-  -- Change population share for the hospitals, TODO according to reputation.
-  -- Since there are no competitors yet the player's hospital can be considered
-  -- to be fairly good no matter what it looks like, so after gbv.AllocDelay
-  -- months, change the share to half of the new people.
-  if self.month >= self.map.level_config.gbv.AllocDelay then
-    self:getLocalPlayerHospital().population = 0.5
-  end
+  local local_hospital = self:getLocalPlayerHospital();
```
Lego3: We want a calculation like this, yes, but I still believe we want it after the AllocDelay time. It is used to let the map maker decide when the hospitals no longer have 25 % each. For instance if on a certain map you start with a very low reputation, but still want the player to get some patients the first year to build reputation fast.

CorsixTH/Lua/entities/patient.lua
Lego3: Should we remove it in this pull request?

CorsixTH/Lua/world.lua
TheCycoONE: My understanding from the comment was that AllocDelay was simply simulating a reputation impact - it was just assumed after that period that you were doing so much better than the AI hospitals that you would receive half of the patients instead of the fair share (25%). Since the reputation impact means you can end up with up to 75% of the patients I don't think that's any value any more.

I feel that the map maker should increase the world spawn rate instead if that's the effect they are going for.

TheCycoONE: Unless AllocDelay means always give fair population until that date? I'm not actually familiar with the variable. If that's the case I'll change it.

CorsixTH/Lua/entities/patient.lua
TheCycoONE: Yes, sorry missed this - I was only looking on the conversation page.

CorsixTH/Lua/world.lua
TheCycoONE: After reading the description of AllocDelay I agree.

Should be 0.25 until AllocDelay, and this after.

TheCycoONE:
Changed to account for AllocDelay and remove nil hospital check from goHome.

CorsixTH/Lua/entities/patient.lua
TheCycoONE: I removed it in the one where we changes hospital = nil to despawn();

Lego3:
Looks good to me now. Is it okay nowadays to merge even if the branch is not up-to-date given that I know what happened in the meantime and the checks passed?

TheCycoONE:
In this case I've just done the rebase.

Requiring the pr to be up to date with the latest master is new, I haven't completely sorted out how I feel about it yet. I've gone both ways on it.

Lego3:
Then it's time to get this into master. :-)

PR #996, opened 2016-03-10
[RDY] Drop stricmp

Alberth289346:
Replace stricmp uses with strcmp for own data comparison, or with own ignorecase_cmp


Alberth289346:
Deleted some more lines (#pragma stuff), appveyor reports no new problems

Alberth289346:
Thanks for the review, good points.

Added "std::" prefix to tolower (I did get your second to last comment too, weird enough).
Deleted the #pragma that you indicated as well, which again did not change anything in appveyor output. If you look for "pragma" though, there are a lot more. No idea what they do.

Alberth289346:
Added the std::strcmp cases too, and removed a now obsolete #include <cctype> that I added

PR #1004, opened 2016-03-22
Always print integer numbers in the bank manager window. Fixes issue 974

Alberth289346:
With many thanks @TheCycoONE


TheCycoONE:
The output of sum() is in the 3 print lines around line 285. Without anything specifying that sum() should only be used for ints, I would move the floors down there.

Alberth289346:
Fixed

PR #1006, opened 2016-03-23
[RDY] Update Rsync filter file and use

tobylane:
This adds the newer folders to the rsync list and changes the OS X target of the rsync to the MacOS folder where the CorsixTH executable is. It's done this way in the 0.40 release and I'm unsure why others haven't seen this problem (it's hidden if you move the app to the top folder of the repository). Only the Icon.icns is left in Resources because that sets the Finder icon, which may not be in control of CMake, my testing didn't make that clear.

### DIFF
--- a/CorsixTH/CMakeLists.txt
+++ b/CorsixTH/CMakeLists.txt
@@ -76,7 +76,8 @@ IF(APPLE)
   #Add an extra step at the end of the build process to copy the resources into the bundle.
   add_custom_command(TARGET CorsixTH
   					 POST_BUILD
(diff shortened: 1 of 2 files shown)


TheCycoONE:
Is the RequiredResources.txt change related to this?

tobylane:
That change just adds the newer folders, it will help any platform that uses the filter file. It's not necessary for the OS X app to work.

TheCycoONE:
Alright, could you move it to a separate commit.
```
git reset HEAD^
git add  CorsixTH/CMakeLists.txt
git commit --amend
git add  CorsixTH/RequiredResources.txt
git commit
git push -f
```

tobylane:
Done.

PR #1009, opened 2016-03-26
Add newer languages to config_finder and config_template

tobylane:
These are up to date, and I've added a note in the wiki for future language contributors to update these files.


Lego3:
Thank you :-)

PR #1016, opened 2016-03-31
[RDY] OS X: Package.cpath and luarocks

tobylane:
First commit adds the relevant folders of a local luarocks installation to the package.cpath var. The working directory of a process launched as an app is $HOME so no relative path will ever work.

Second commit adds a cmake option to turn on a post build rule, another always on post build rule and some comments edits. The first includes dylibs (dependancies) by copying them in and linking CorsixTH  to them and each other. It's assumed the user installed them by homebrew, others can be added when users emerge. The second script creates a local luarocks installation and patches Luasocket.

Output of osx_dylibs https://travis-ci.org/tobylane/CorsixTH/jobs/119871660#L1255 and osx_luarocks https://travis-ci.org/tobylane/CorsixTH/jobs/119871660#L978

### DIFF
--- a/CorsixTH/CMakeLists.txt
+++ b/CorsixTH/CMakeLists.txt
@@ -277,7 +277,12 @@ if(NOT USE_SOURCE_DATADIRS)
     install(CODE "
       INCLUDE(BundleUtilities)
       SET(BU_CHMOD_BUNDLE_ITEMS ON)
-      FIXUP_BUNDLE(${CMAKE_INSTALL_PREFIX}/CorsixTH.app \"\" \"\")
+      FIXUP_BUNDLE(\"${CMAKE_INSTALL_PREFIX}/CorsixTH.app\" \"\" \"\")
       ")
+    if(WITH_LUAROCKS)
+      install(CODE "execute_process(
+        COMMAND bash \"${CMAKE_SOURCE_DIR}/scripts/macos_luarocks\" \"${CMAKE_INSTALL_PREFIX}\")
+      ")
+    endif()
   endif()
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@
 #   - WITH_VLD    	: Build with Visual Leak Detector (requires Visual Studio) (no)
 #   - USE_PRECOMPILED_DEPS : Use precompiled dependencies on *nix (no)
 #   - USE_VCPKG_DEPS : Build vcpkg dependencies locally (requires Visual Studio) (no)
+#   - WITH_LUAROCKS : Install required luarocks in the macOS app (requires luarocks)
 
 #   - ENABLE_UNIT_TESTS : Enable Unit Testing Target (requires Catch2) (no)
 #   - BUILD_ANIMVIEWER : Generate AnimViewer build files (no)
@@ -51,6 +52,10 @@ endif()
 # Dependency management
 if(UNIX AND CMAKE_COMPILER_IS_GNU)
   option(USE_PRECOMPILED_DEPS "Use Precompiled Dependencies" OFF) # Make *nix systems opt in
+endif()
+
+if(APPLE)
+  option(WITH_LUAROCKS "Install required luarocks in the app" OFF)
 endif()
 
 option(ENABLE_UNIT_TESTS "Enables Unit Testing Targets" OFF)
--- a/scripts/macos_luarocks
+++ b/scripts/macos_luarocks
@@ -1,0 +1,74 @@
+#!/bin/bash
+: ' Copyright (c) 2016- Toby "tobylane"
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+of the Software, and to permit persons to whom the Software is furnished to do
+so, subject to the following conditions:
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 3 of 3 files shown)


TheCycoONE:
Maybe make your second commit a make target (like docs) instead of a cmake flag? Or would that not work?

tobylane:
I believe it would have to be a duplicated CorsixTH sub project definition as CMake projects can't be cloned/extended, or the user would run CMake a second time after Xcode.

tobylane:
Added a third commit that does the dylib change for AnimViewer. Output https://travis-ci.org/tobylane/CorsixTH/jobs/120023001#L1488

TheCycoONE:
I meant `ADD_CUSTOM_TARGET`, or better yet `INSTAL`' and then after building `make foo` or `make install`.

All of the rsync steps should be in the install target. 

I believe `INSTALL` can already do most of the dylib  changes for you too.

For xcode I have seen it run as `xcodebuild --target install`

tobylane:
Changes per request. Install as a target is already taken and can't be reliably reclaimed so they are named IncludeDylibs and IncludeAnimViewDylibs. They can be chained in a command with the app build like `xcodebuild -configuration Release -target CorsixTH -target IncludeDylibs`

TheCycoONE:
Ok, I'm fine with this, or maybe just a ```redistributable``` target, as includeDylibs is a bit wordy and specific.

CorsixTH/CorsixTH.lua
```diff
@@ -17,6 +17,14 @@ for _, arg in ipairs({...}) do
   end
 end
 
+-- On Mac OS X add the local luarocks folder to package.cpath.
+if (io.popen"uname":read'*l' == "Darwin") then
```
TheCycoONE: Should use brackets for popen and read. Could potentially be used everywhere not just OSX, so maybe this outer condition isn't necessary.

scripts/osx_luarocks
```diff
@@ -0,0 +1,48 @@
...
+elif [ "$lua" = "" ]; then lua=5.2
+fi
+
+if [ -d luarocks/share/lua/"$lua" ]
+  then echo "Luarocks for Lua"$lua" in place."
+  exit
+elif [ -d "$dir/share/lua/$lua/socket/" ] && [ -z "$1" ]
+  then echo "Copying luarocks from global installation."
+  mkdir -p luarocks/share/lua
+  cp -R "$dir/share/lua/$lua" luarocks/share/lua/
```
maki-chan: Missing copy command for libraries here, should also have something like:

`cp -R "$dir/lib/lua/$lua" luarocks/lib/lua/`

as the path `luarocks/lib/lua` was defined for the various needed Lua packages in the `package.cpath` variable.

tobylane:
@TheCycoONE: Will do.

@maki-chan: I started off doing that but found it wasn't necessary. I renamed the share folder in your build, removed my local copy of luafilesystem and the game works fine. We can investigate this further, would you like to use IRC? We're at FreeNode #corsix-th.

TheCycoONE:
If lfs.so doesn't exist in your lua path there is no way CorsixTH would work, so it must be somewhere on your system.

Actually I was thinking about it, and those lua so files may have been what was giving the bad instruction errors.

clobber:
@tobylane Still working on getting this merged? This looked promising.

CorsixTH/CorsixTH.lua
```diff
@@ -17,6 +17,12 @@ for _, arg in ipairs({...}) do
   end
 end
 
+-- Add the local luarocks folder to package.cpath.
+local luarocks_path = debug.getinfo(1, "S").source:sub(2):match("(.*/)") ..
+    "luarocks/lib/lua/" .. string.sub(_VERSION,5)
+package.cpath = luarocks_path .. "/?.so;" .. luarocks_path .. "/mime/?.so;" ..
```
TheCycoONE: A little late now, but /mime/?.so and socket/?.so shouldn't have been part of this. ? is replaced by the module name, which is mime.core and socket.core.

tobylane:
Redone to be a part of the install process, simplified the lua part.

TheCycoONE:
Sorry @tobylane I didn't realise that you had updated this. It needs a CMAKE flag I think - we don't want to unconditionally brew install anything. I'll review it closer soon.

TheCycoONE:
I still feel this needs a toggle in cmake. Everywhere else we download external tools we opt in e.g. `USE_VCPKG`

tobylane:
Fixed, and removed the reference to homebrew.

CorsixTH/CMakeLists.txt
```diff
@@ -279,5 +279,10 @@ if(NOT USE_SOURCE_DATADIRS)
       SET(BU_CHMOD_BUNDLE_ITEMS ON)
       FIXUP_BUNDLE(${CMAKE_INSTALL_PREFIX}/CorsixTH.app \"\" \"\")
       ")
+    if(WITH_LUAROCKS)
+      install(CODE "execute_process(
+        COMMAND bash -c ${CMAKE_SOURCE_DIR}/scripts/macos_luarocks ${CMAKE_INSTALL_PREFIX})
```
colinjmatt, line 284: In order for bash to handle spaces in the file path, this would be better as
`COMMAND bash \"${CMAKE_SOURCE_DIR}/scripts/macos_luarocks\" \"${CMAKE_INSTALL_PREFIX}\")`

tobylane:
Fixed, thanks.

Ready.

CorsixTH/CMakeLists.txt
```diff
@@ -277,7 +277,12 @@ if(NOT USE_SOURCE_DATADIRS)
     install(CODE "
       INCLUDE(BundleUtilities)
       SET(BU_CHMOD_BUNDLE_ITEMS ON)
-      FIXUP_BUNDLE(${CMAKE_INSTALL_PREFIX}/CorsixTH.app \"\" \"\")
+      FIXUP_BUNDLE(\"${CMAKE_INSTALL_PREFIX}/CorsixTH.app\" \"\" \"\")
```
colinjmatt, line 280: Is adding quotes here needed as that's CMake code and should be handled fine? They were needed on line 284 because bash doesn't handle unquoted filepaths with spaces.

Someone who knows more abotu CMake would need to comment on that.

tobylane: > -- Installing: /Applications/test path/CorsixTH.app/Contents/Resources/LICENSE.txt
> -- fixup_bundle
> --   app='/Applications/test'
> --   libs='path/CorsixTH.app'
> --   dirs=''
> --   ignoreItems=''
> -- warning: *NOT* handled - directory/file does not exist...
> CMake Error at /usr/local/Cellar/cmake/3.17.3/share/cmake/Modules/BundleUtilities.cmake:988 (message):
>   error: fixup_bundle: not a valid bundle
> Call Stack (most recent call first):
>   cmake_install.cmake:62 (FIXUP_BUNDLE)
> -- fixup_bundle: done

Without that, I get this error where CMake has the same problem.

colinjmatt: Ah yes that makes sense. It might be worthwhile as a separate PR to go through and quote all filepath variables to prevent this anywhere else in CMake.

Issue #451, opened 2014-12-23
Disappearing wall when buying a new plot of land


TheCycoONE:
_From [mark.l1...@hotmail.co.uk](https://code.google.com/u/117099726060447259318/) on July 09, 2010 15:27:02_

What steps will reproduce the problem? 1.start game
2.build rooms
3.buy another plot of land What is the expected output? What do you see instead? At the moment you can't build across what will become corridors if/when you buy the adjacent plot.   You can build next to these plots if you are not blocking off an area of the hospital. To a certain extent the same could happen in the original, but what I did find odd was building against a plot that I did not own for a wall of a clinic to disappear when I did buy the plot. see screen shot

What version of CorsixTH are you using (e.g. "Beta 2", "SVN r123 ")? r942 What operating system / compile settings are you using? win XP
What level was this on (e.g. "Demo level", "Full game level 12")?
full version - L9 Please provide any additional information below. I think that in the original you could block off areas and when you did you could not place an object in the blocked off space.  You could not block off an area if someone was walking in the area. This was usefull at the start of a game to maximise the area you could build in.  Once you purchased another plot you would have to move or delete the objects doing the blocking before you could add objects to the new space.

**Attachment:** [screen.bmp](http://code.google.com/p/corsix-th/issues/detail?id=384)

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=384_

TheCycoONE:
Confirmed that this bug still affects 0.40 from #681

TheCycoONE:
As seen in #1011 this can lead to crashes. Also confirmed in 0.50.

Lego3:
Time to have a look at this.

Lego3:
We need #1018 as well to fix this completely.

Lego3:
I put the test map I used here: https://www.dropbox.com/sh/1urn1hfvq76e61b/AAAPVCfWrz63QQlyV2jVfsgGa?dl=0
#1015 fixes issues when you build rooms next to expansion walls, while #1018 fixes issues when you later on buy the expansion.

TheCycoONE:
It was closed automatically by merging the pull request because it had fix #451 in the message.

Lego3:
Yes, I know. Would have looked strange to just open it though. :-P

TheCycoONE:
#1018 was merged.

PR #1018, opened 2016-04-01
[RDY] Fix disappearing room walls next to new plots

Lego3:
One proposed solution to #451.

The actual buying of a plot needs to be done in C++ as it would be way too slow in Lua (I gave it a go). At the same time C++ doesn't know about wall types for different rooms etc, so the logic for which type of wall to add when a player buys a new plot must be done in Lua. Therefore this proposed solution makes the C++ "buying" code return which tiles inside you current hospital are adjacent to the plot you are buying, and then Lua checks whether some extra room wall needs to be placed at these now removed external expansion wall positions.

### DIFF
--- a/CorsixTH/Lua/map.lua
+++ b/CorsixTH/Lua/map.lua
@@ -265,11 +265,69 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
     self.parcelTileCounts[plot] = self.th:getParcelTileCount(plot)
     if plot > 1 and not map_editor then
       -- TODO: On multiplayer maps, assign plots 2-N to players 2-N
-      self.th:setPlotOwner(plot, 0)
+      self:setPlotOwner(plot, 0)
     end
   end
 
   return objects
+end
+
+--[[! Sets the plot owner of the given plot number to the given new owner. Makes sure
+      that any room adjacent to the new plot have walls in all directions after the purchase.
+!param plot_number (int) Number of the plot to change owner of. Plot 0 is the outside and
+       should never change owner.
+!param new_owner (int) The player number that should own plot_number. 0 means no owner.
+]]
+function Map:setPlotOwner(plot_number, new_owner)
+  local split_tiles = self.th:setPlotOwner(plot_number, new_owner)
+  for _, coordinates in ipairs(split_tiles) do
+    local x = coordinates[1]
+    local y = coordinates[2]
+
+    local _, north_wall, west_wall = self.th:getCell(x, y)
+    local cell_flags = self.th:getCellFlags(x, y)
+    local north_cell_flags = self.th:getCellFlags(x, y - 1)
+    local west_cell_flags = self.th:getCellFlags(x - 1, y)
+    local wall_dirs = {
+      south = { layer = north_wall,
+                block_id = 2,
+                room_cell_flags = cell_flags,
+                adj_cell_flags = north_cell_flags,
+                tile_cat = 'inside_tiles',
+                wall_dir = 'north'
+              },
+      north = { layer = north_wall,
+                block_id = 2,
+                room_cell_flags = north_cell_flags,
+                adj_cell_flags = cell_flags,
+                tile_cat = 'outside_tiles',
+                wall_dir = 'north'
+              },
+      east = { layer = west_wall,
+               block_id = 3,
+               room_cell_flags = cell_flags,
+               adj_cell_flags = west_cell_flags,
+               tile_cat = 'inside_tiles',
+               wall_dir = 'west'
+             },
+      west = { layer = west_wall,
+               block_id = 3,
+               room_cell_flags = west_cell_flags,
+               adj_cell_flags = cell_flags,
+               tile_cat = 'outside_tiles',
+               wall_dir = 'west'
+             },
+    }
+    for _, dir in pairs(wall_dirs) do
+      if dir.layer == 0 and dir.room_cell_flags.roomId ~= 0 and
+          dir.room_cell_flags.parcelId ~= dir.adj_cell_flags.parcelId then
+        local room = self.app.world.rooms[dir.room_cell_flags.roomId]
+        local wall_type = self.app.walls[room.room_info.wall_type][dir.tile_cat][dir.wall_dir]
+        self.th:setCell(x, y, dir.block_id, wall_type)
+      end
+    end
+  end
+  self.th:updatePathfinding()
 end
 
 --[[! Saves the map to a .map file
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -589,8 +589,8 @@ void THMap::save(std::string filename)
    * A divider wall is removed if the owners are the same and 'iParcelId' is involved.
    \return True if a border was removed, false otherwise
 */
-static void addRemoveDividerWalls(THMap* pMap, THMapNode* pNode, const THMapNode* pOriginalNode,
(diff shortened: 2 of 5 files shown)


CorsixTH/Src/th_map.cpp
```diff
@@ -645,6 +629,26 @@ void THMap::setParcelOwner(int iParcelId, int iOwner)
     _updatePurchaseMatrix();
 }
 
+void THMap::_checkDividers(THMapNode* pNode, const THMapNode* pOriginalNode,
```
TheCycoONE: The only access this needs to private THMap properties is m_pPlotOwner and there is a public method for getting the plot owner.

In conformance with C++ Core Guidelines C.4, this method should not be a part of the THMap class. Since it's not used outside of this file, it can be defined as 
```cpp
static void check_dividers(THMapNode* pNode ...)
{
...
}
```

CorsixTH/Src/th_map.cpp
```diff
@@ -581,7 +581,7 @@ void THMap::save(std::string filename)
     os.close();
 }
 
-void THMap::setParcelOwner(int iParcelId, int iOwner)
+void THMap::setParcelOwner(int iParcelId, int iOwner, std::vector<std::pair<int, int>>* pSplitTiles)
```
TheCycoONE: Use a reference rather than a pointer here. & instead of *.

CorsixTH/Src/th_map.cpp
Lego3: Okay. How should the method get the THMap object to call? Should it also be a part of the argument list?

CorsixTH/Lua/map.lua
```diff
@@ -265,13 +265,45 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
...
   return objects
 end
 
+function Map:setPlotOwner(plot_number, new_owner)
+  local split_tiles = self.th:setPlotOwner(plot_number, new_owner)
+  for _, coordinates in ipairs(split_tiles) do
+    local x = coordinates[1]
+    local y = coordinates[2]
+
+    local __, north_wall, west_wall = self.th:getCell(x, y)
```
TheCycoONE: I think luacheck explicitly ignores redefining _ so it can be reused in cases like this. Nicer than introducing a second ignore name.

CorsixTH/Lua/map.lua
```diff
@@ -265,13 +265,45 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
...
+    local cell_flags = self.th:getCellFlags(x, y)
+    local north_cell_flags = self.th:getCellFlags(x, y - 1)
+    local west_cell_flags = self.th:getCellFlags(x - 1, y)
```
TheCycoONE: west_cell_flags isn't used until 294, so this line could be moved down.

CorsixTH/Src/th_map.cpp
Lego3: Seems to work fine. :-)

CorsixTH/Src/th_map.cpp
TheCycoONE: Yes, it should be the first argument.

CorsixTH/Src/th_map.cpp
```diff
@@ -619,8 +619,16 @@ void THMap::setParcelOwner(int iParcelId, int iOwner)
                     }
                 }
             }
-            _checkDividers(pNode, pOriginalNode, iX, 1, 2, iParcelId);
-            _checkDividers(pNode, pOriginalNode, iY, 128, 1, iParcelId);
+            bool isSplitTile = _checkDividers(pNode, pOriginalNode, iX, 1, 2, iParcelId);
```
Alberth289346: Any particular reason you don't fold the _check..(..) thingie directly in the 'if' ? the boolean is used one time, and "bool isSplitTile = " is longer than "if (".

CorsixTH/Src/th_map.cpp
```diff
@@ -629,7 +637,7 @@ void THMap::setParcelOwner(int iParcelId, int iOwner)
     _updatePurchaseMatrix();
 }
 
-void THMap::_checkDividers(THMapNode* pNode, const THMapNode* pOriginalNode,
+bool THMap::_checkDividers(THMapNode* pNode, const THMapNode* pOriginalNode,
```
Alberth289346: Commit misses documentation what magic return value means (checkDividers == true, does that mean it's checked, or checked good, or needs fixed, or ... ?

CorsixTH/Src/th_map.cpp
Lego3: No, not really. In this case it works just as well.

CorsixTH/Lua/map.lua
Lego3: Yes, I had one go at refactoring it but wasn't satisfied so I thought I'd get some comments on it first. :-) I'll give it another go.

Lego3:
Added all suggested changes. The Lua function gets a little longer, but it feels like it's easier to read now?

CorsixTH/Src/th_map.h
```diff
@@ -423,6 +423,16 @@ class THMap
...
 };
 
+//! Add or remove divider wall for the given node
+/*!
+    If the given 'pNode' has an indoor border to another parcel in the 'delta' direction:
+    * A divider wall is added in the layer specified by 'block' if the owners
+      of the two parcels are not the same, or
+    * A divider wall is removed if the owners are the same and 'iParcelId' is involved.
+*/
+static void addRemoveDividerWalls(THMap* pMap, THMapNode* pNode, const THMapNode* pOriginalNode,
```
TheCycoONE: This should not be here. addRemoveDivderWalls should only be defined in the th_map.c compilation unit. You will have to move the definition in th_map.c higher in the file (before it is used).

CorsixTH/Lua/map.lua
```diff
@@ -265,13 +265,69 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
...
+      west = { layer = west_wall,
+               block_id = 3,
+               room_cell_flags = west_cell_flags,
+               adj_cell_flags = cell_flags,
+               tile_cat = 'outside_tiles',
+               wall_dir = 'west'
+             },
+    }
+    for _, dir in pairs(wall_dirs) do
+      if dir.layer == 0 and dir.room_cell_flags.roomId ~= 0 and dir.room_cell_flags.parcelId ~= dir.adj_cell_flags.parcelId then
```
TheCycoONE: Maybe add some newlines after the conditions here (like @Alberth289346 has done in some of his refactor changes), just to make this take less horizontal space.

Similar with the next line should be split. I'm not sure the best ways to add line breaks to this one.

@Alberth289346 what's your opinion on this function overall now?

Alberth289346: I'd add a \n after the "and" at line 322, and indent the remaining text 4 spaces relative to the "if" (ie double indent on continuation lines).

For line 323, "self.app.walls[self.app.world.rooms[dir.room_cell_flags.roomId]" looks like "room_type" or so to me, so I'd assign that to that variable above line 323 as intermediate step.

Otherwise, the function looks much better with the table.

CorsixTH/Src/th_map.h
Lego3: I see. I was thinking that maybe all documentation for functions should be in .h

Lego3:
Made additional tweaks.

TheCycoONE:
Anything left, or are you RDY?

PR #1019, opened 2016-04-01
[reference] Split callback functions out of the action literals.

Alberth289346:
Seemingly big rewrites, bust mostly just shifting lines. Should be better if you ignore white space.

### DIFF
--- a/CorsixTH/Lua/dialogs/staff_dialog.lua
+++ b/CorsixTH/Lua/dialogs/staff_dialog.lua
@@ -314,12 +314,7 @@ end
 
 function UIStaff:placeStaff()
   self.staff.pickup = true
-  self.staff:setNextAction({
-    name = "pickup",
-    ui = self.ui,
-    todo_close = self,
-    must_happen = true,
-  }, true)
+  self.staff:setNextAction(PickupAction():setUi(self.ui):setTodoClose(self):setMustHappen(), true)
 end
 
 function UIStaff:fireStaff()
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -261,7 +261,10 @@ function App:init()
 
     dofile "room"
     self.rooms = self:loadLuaFolder("rooms")
+
+    dofile "humanoid_action"
     self.humanoid_actions = self:loadLuaFolder("humanoid_actions")
+
     local diseases = self:loadLuaFolder("diseases")
     self.diseases = self:loadLuaFolder("diagnosis", nil, diseases)
 
--- a/CorsixTH/Lua/calls_dispatcher.lua
+++ b/CorsixTH/Lua/calls_dispatcher.lua
@@ -258,7 +258,7 @@ function CallsDispatcher.sendNurseToVaccinate(patient, nurse)
     -- The epidemic may have ended before the call can be executed
     -- so just finish the call immediately
     CallsDispatcher.queueCallCheckpointAction(nurse)
-    nurse:queueAction{name = "answer_call"}
+    nurse:queueAction(AnswerCallAction())
     nurse:finishAction()
     patient.reserved_for = nil
   end
@@ -418,11 +418,8 @@ end
 -- A interrupt handler could be supplied if special handling is needed.
 -- If not, the default would be reinsert the call into the queue
 function CallsDispatcher.queueCallCheckpointAction(humanoid, interrupt_handler)
-  return humanoid:queueAction{
-    name = "call_checkpoint",
-    call = humanoid.on_call,
-    on_remove = interrupt_handler or CallsDispatcher.actionInterruptHandler
-  }
+  return humanoid:queueAction(CallCheckPointAction():setCall(humanoid.on_call)
+      :setOnRemove(interrupt_handler or CallsDispatcher.actionInterruptHandler))
 end
 
 -- Default checkpoint interrupt handler
@@ -488,7 +485,7 @@ function CallsDispatcher.unassignCall(call)
   assert(assigned.on_call == call, "Unassigning call but the staff was not on call or a different call")
   call.assigned = nil
   assigned.on_call = nil
-  assigned:setNextAction{name = "answer_call"}
+  assigned:setNextAction(AnswerCallAction())
 end
 
 function CallsDispatcher.verifyStaffForRoom(room, attribute, staff)
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -309,7 +309,7 @@ function UIEditRoom:clearArea()
         elseif entity.action_queue[1].name == "seek_room" or (meander and meander.name == "seek_room") then
           -- Make sure that the humanoid doesn't stand idle waiting within the blueprint
           if entity.action_queue[1].name == "seek_room" then
-            entity:queueAction({name = "meander", count = 1, must_happen = true}, 0)
+            entity:queueAction(MeanderAction():setCount(1):setMustHappen(), 0)
           else
             meander.done_walk = false
           end
@@ -599,7 +599,7 @@ function UIEditRoom:returnToDoorPhase()
   local room = self.room
   room.built = false
   if room.door and room.door.queue then
-    room.door.queue:rerouteAllPatients({name = "seek_room", room_type = room.room_info.id})
+    room.door.queue:rerouteAllPatients(SeekRoomAction(room.room_info.id))
   end
 
   self.purchase_button:enable(false)
--- a/CorsixTH/Lua/dialogs/fullscreen/fax.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/fax.lua
@@ -157,20 +157,13 @@ function UIFax:choice(choice_number)
     elseif choice == "guess_cure" then
       owner:setDiagnosed()
       if owner:agreesToPay(owner.disease.id) then
-        owner:setNextAction{
-          name = "seek_room",
-          room_type = owner.disease.treatment_rooms[1],
-          treatment_room = true,
-        }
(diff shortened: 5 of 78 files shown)


TheCycoONE:
Looks valid, any particular reason you prefer named local functions to lambdas in the method call?

Alberth289346:
So you can write chains with callback function names without a zillion lines between the start and the end of the statement, mostly

```` lua
humanoid:queueAction(UseObjectAction(obj):setLoopCallback(desk_loop):mustHappen())
````

**Comment by TheCycoONE** [comment]
Ok, lets hold this commit for the PR with the rest of that action API.

**Comment by Alberth289346** [comment]
Constructed the actions and replaced the actions I had in my patch
first commit (commit 1a1c012 has not changed)

File: CorsixTH/Lua/dialogs/staff_dialog.lua [hunk 1]
```diff
@@ -314,12 +314,7 @@ end
 
 function UIStaff:placeStaff()
   self.staff.pickup = true
-  self.staff:setNextAction({
-    name = "pickup",
-    ui = self.ui,
-    todo_close = self,
-    must_happen = true,
-  }, true)
+  self.staff:setNextAction(PickupAction():setUi(self.ui):setTodoClose(self):setMustHappen(), true)
```
TheCycoONE: What do you think about mixing the highPriority flag into the action, i.e. :setHighPriority() instead of ,true?

(out of scope but the queueAction position number bugs me a bit too, seems like leaky abstraction)

Alberth289346: The aim of this patch was to fix a lot of formatting nuisances I had for a long time. For the actions, it was mostly lack of parentheses, but the big literals didn't look nicely either, and a few experiments with a more function-call like syntax proved to be very useful.

I haven't studied the actions in much detail, and I am not sure I want to go there. At the surface it all looks very much duplicated and overly complicated to me. I believe the reason is that the code is trying to decide the next action from the end of the current action. (Would love some feedback about that, but it's not trivial to say the least.)

Some actions (probably the older code, like toilets, and gp) hook that into the "on-exit" call, leading to nested persistent recursive functions. Other actions push things into the action_queue (probably was added after deciding nested persistent recursive functions aren't very nice). Then you get problems like pick_up, or interrupts like getting thirsty or going to the toilets, which break such a queue, leading to more work-arounds. Someone messes up and you get an empty action queue, which then leads to new code to decide what to do in that case.

So, in all, I believe the core problem is the chain of actions. Instead, I think there should be a point in a humanoid or object(?) that tracks what it is aiming to do long and short term, and generate actions from that. An action is then not much more than an animation. All logic of what to do next moves to that tracking point.
One form of such a point is a state machine, but I am not sure it's the best solution.

One thing I haven't decided what to make of it, are the registered callbacks that happen all over the place. I think they complicate matters, but don't know how badly.

As for your questions, technically the ", true" is the replacement priority for already existing actions rather than the priority of the new action. (I agree however you can see it as the priority of the new action as well.)
I haven't looked at the details of that priority flag, but it looks like very bad smelling code to me. In fact, the entire action queue smells, as I just explained. Currently, it looks like a waste of time trying to fix actions, to me, as explained above.

MarkL1961:
It may or may not help, but several years back @corsix  started a new action queue branch which is available on github.  It will be years out of date now and lots of other things have been added since, but it might be the basis for a change to the action queue system - with a lot of work no doubt.

Alberth289346:
I know, I have already checked that code. It adds action classes (more properly than I do here). He also solves the synchronization issue more nicely (so when several humanoids together do something they both start at the same time). For details, read commit 0b15a7d6744084473645009ba9ad210d91180351 only the fully new files are interesting, and they have a lot of text, which is quite nice.

At the same time, he is still squarely in the action system domain, which, as I have just argued above, is one of the causes for much complexity in my view, which makes it unattractive to go in that direction for me.
Obviously, I could be wrong, since I don't have the answer to a better solution yet.

Alberth289346:
Could remove the RFC now, and add a 0.70 label, I think?  @TheCycoONE

TheCycoONE:
This PR can't be cleanly merged now. Could you rebase it on the latest master.

Alberth289346:
rebased

Alberth289346:
Removed a "," that was left behind (research.lua, after  'persistable:research_desk_loop_callback' function

TheCycoONE:
Humanoid.lua walkTo function. The must_happen logic is convoluted for a bool.

TheCycoONE:
Crash:
```
...H\Debug\..\..\..\CorsixTH\Lua\objects\reception_desk.lua:228: attempt to call a nil value (global 'StaffReception')
stack traceback:
        ...H\Debug\..\..\..\CorsixTH\Lua\objects\reception_desk.lua:228: in method 'occupy'
        ...H\Debug\..\..\..\CorsixTH\Lua\objects\reception_desk.lua:169: in method 'checkForNearbyStaff'
        ...H\Debug\..\..\..\CorsixTH\Lua\objects\reception_desk.lua:187: in method 'setTile'
        ...CorsixTH\Debug\..\..\..\CorsixTH\Lua\entities\object.lua:60: in method 'Object'
        ...H\Debug\..\..\..\CorsixTH\Lua\objects\reception_desk.lua:77: in local 'constructor'
        ...s2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\class.lua:76: in field '?'
        ...s2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\world.lua:1940: in method 'newObject'
        ...TH\Debug\..\..\..\CorsixTH\Lua\dialogs\place_objects.lua:514: in method 'placeObject'
        ...TH\Debug\..\..\..\CorsixTH\Lua\dialogs\place_objects.lua:472: in method 'onMouseUp'
        ...2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\window.lua:1462: in field 'onMouseUp'
        ...d-vs2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\ui.lua:649: in function <...d-vs2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\ui.lua:636>
        (...tail calls...)
        ...-vs2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\app.lua:904: in function <...-vs2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\app.lua:899>
```

TheCycoONE:
The problem line should be StaffReceptionAction instead.

TheCycoONE:
Another crash
```
...ug\..\..\..\CorsixTH\Lua\humanoid_actions\knock_door.lua:58: attempt to index a nil value (local 'door')
stack traceback:
        ...ug\..\..\..\CorsixTH\Lua\humanoid_actions\knock_door.lua:58: in field '?'
        ...rsixTH\Debug\..\..\..\CorsixTH\Lua\entities\humanoid.lua:576: in upvalue 'Humanoid_startAction'
        ...rsixTH\Debug\..\..\..\CorsixTH\Lua\entities\humanoid.lua:652: in method 'queueAction'
        ...TH\Debug\..\..\..\CorsixTH\Lua\humanoid_actions\walk.lua:306: in function <...TH\Debug\..\..\..\CorsixTH\Lua\humanoid_actions\walk.lua:254>
        (...tail calls...)
        ...TH\Debug\..\..\..\CorsixTH\Lua\humanoid_actions\walk.lua:232: in local 'timer_function'
        ...2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\entity.lua:219: in method 'tick'
        ...s2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\world.lua:1075: in method 'onTick'
        ...-vs2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\app.lua:1003: in function <...-vs2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\app.lua:1000>
        (...tail calls...)
        ...-vs2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\app.lua:904: in function <...-vs2015-x64\CorsixTH\Debug\..\..\..\CorsixTH\Lua\app.lua:899>
```

TheCycoONE:
For the above, KnockDoorAction should set self.door = door, not self.knock_door = door.

TheCycoONE:
What was the logic used to decide if a parameter should be in the constructor or set via a setter?

Alberth289346:
> Humanoid.lua walkTo function. The must_happen logic is convoluted for a bool.
Good point.
Upon inspection I realized it was even wrong (it was meant to catch 'not nil', but obviously it also fails on 'false'). Checked this for all my new 'set' functions, and for all vulnerable functions, checked all their uses. That resulted in finding 7 more troubled spots, now attached as 4th commit.

5th commit removed the parameter of two 'set' methods, as they always set the property.

> What was the logic used to decide if a parameter should be in the constructor or set via a setter?
Logic was just a check what was always set and what was sometimes set, as that gave the simplest conversion path.

TheCycoONE:
There is a regression in this patch series, people no longer die when they reach 0 health (play chapter 1 with no nurse in the pharmacy). There are no errors in the console.

TheCycoONE:
Why does GetUpAction have self.must_happen =true.

I don't see that in the old code for get_up

MarkL1961:
It is perhaps something I missed when it was added.  The whole of fall down and get up animation for earthquakes was causing problems so it was disabled.  Similar to #1094  where the fallen would sometimes slide on their belly off the map

Alberth289346:
Ah right, thanks, that code was already not making any sense.

I think however it's a separate issue, this PR is about refactoring, just rewriting the code without functional changes. Changing behavior is not part of it.

Alberth289346:
New attempt in #1107, keeping this open for reference

TheCycoONE:
A PR doesn't need to be opened to be referenced, link from top of #1107 will work.

PR #1020, opened 2016-04-02
Add patient graph, fixes #72.

### LINKED ISSUE #72: Health graph for patients missing
When clicking on a patient. it is possible to see his/her profile but even if the health graph background is being shown, there is no value. I guess this is not implemented but it could as well be a graphic bug.

Alberth289346:


TheCycoONE:
Very pretty!

Issue #71, opened 2013-10-26
Hospital never gets dirty


mounirlamouri:
If ones never adds bins to the hospital, it still stays very clean contrary to the original version. I guess this is just not implemented but given that I have not seen that mentioned, I guess it's better to open an issue.

PR #1022, opened 2016-04-03
[RDY] Patient littering

Alberth289346:
See #71 feature request. phillipp took it on, but that didn't materialize,apparently.

Technically, it needs a savegame bump, I can add one if you think it's needed, but we just had 2.

### DIFF
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -29,6 +29,8 @@ function Patient:Patient(...)
   self.hover_cursor = TheApp.gfx:loadMainCursor("patient")
   self.should_knock_on_doors = true
   self.treatment_history = {}
+  self.going_home = false -- If set, the patient is going home.
+  self.litter_countdown = nil -- If set, number of tiles to walk before attempting to drop litter.
   self.has_fallen = 1
   self.has_vomitted = 0
   self.action_string = ""
@@ -905,8 +907,8 @@ function Patient:tickDay()
         end
         -- The patient might also throw the can on the floor, bad patient!
         if math.random() < 0.6 then
-          -- It will be dropped between 1 and 5 tiles away.
-          self.litter_countdown = math.random(1, 5)
+          -- It will be dropped between 1 and 12 tiles away (litter bin catches 8 radius).
+          self.litter_countdown = math.random(1, 12)
         end
       end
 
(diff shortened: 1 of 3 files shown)


Lego3:
While this is a clear step away from the original I think it is an all-good change. :-) It was always strange why you couldnt place bins anywhere, and their purpose was unclear. Maybe we could allow litter inside rooms as well?

Alberth289346:
Handy men don't clean inside the rooms currently, which is somewhat weird, but it makes the corridor have its own special behavior, which I think is good.

TheCycoONE:
The large bin radius might make it too easy to avoid litter. I agree it makes sense to have bins in the hall, and it makes sense for patients not to litter if they encounter a bin, but we may have to nerf this a bit.
Anyone have a feeling on whether to play test this first, or merge it and see what the players think?

MarkL1961:
I think a three tile radius from the bin there is a 90% chance the patient uses it,  five tiles there is a 50% chance of using it.  People are lazy, so any further there is more chance of them just littering.
I also think there should be a chance everyone has litter, I think at the moment only some patients that use a drinks machine can litter, whereas in TH everyone dropped litter

Alberth289346:
The patch gives everybody litter that they drop after a number of tiles (random, from 30 to200 or so), within the hospital. I checked, my level 1 & level 4 hospitals need about 40-60 tiles for patients to walk before they go home. Drinks machine lowers the number of tiles to walk between 1 to 12, I think. I was trying to avoid littering the smaller hospitals (in the earlier levels).

The bin is only searched when they are about to drop something. that means a row of bins at the entrance will not work (which would if you monitor how close they pass a bin constantly, and delete the litter when they pass close enough). To catch all litter you need to place bins everywhere. To avoid needing to have a bin every few tiles (which looks stupid imho), I increased the range to 8, which can be interpreted as "they just passed a bin or will soon pass a bin".

I also added a small chance they drop litter despite having a bin nearby.

I pondered about the range, but the tile count is already random, so it would be equal to reducing the radius of finding the bin. (Adding something random on top of something random doesn't it make more random, the amount of randomness stays mostly the same.)
Things are still pretty much untested for having useful numbers. Maybe chances or ranges should be improved.

MarkL1961:
@Alberth289346  Sounds okay to me, I look forward to trying it out when I have some spare time

CorsixTH/Lua/entities/patient.lua
```diff
@@ -1028,9 +1039,12 @@ function Patient:setTile(x, y)
           end
         end
       end
-      self.litter_countdown = nil
+
+      -- Always give new litter to drop.
+      self.litter_countdown = math.random(30, 150)
```
Alberth289346: The idea was that people staying longer in the hospital and dropping litter have less litter to drop later, but otherwise, not really.

Splitting between the first and next literings make it feasible to hit bigger / less efficient hospitals harder of softer. Not sure what the right values are here, likely they need a little tweaking later.

Alberth289346:
rebased

PR #1023, opened 2016-04-03
[RDY] Refactor patient

Alberth289346:
Moving hospital administration bits out of the patient. I think there is more, but this seems like a good first step.

Code is mostly copy/paste, but the various self/patient/hospital replacements are easy to mess up.
I'd be fine with postponing after 0.60

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1870,6 +1870,13 @@ end
 --! Update the 'cured' counts of the hospital.
 --!param patient Patient that was cured.
 function Hospital:updateCuredCounts(patient)
+  if not patient.is_debug then
+    self:changeReputation("cured", patient.disease)
+  end
+
+  if self.num_cured < 1 then
+    self.world.ui.adviser:say(_A.information.first_cure)
+  end
   self.num_cured = self.num_cured + 1
   self.num_cured_ty = self.num_cured_ty + 1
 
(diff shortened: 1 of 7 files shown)


TheCycoONE:
Just as an update I'm taking your suggestion to postpone this until post 0.60.

CorsixTH/Lua/hospital.lua
```diff
@@ -664,7 +664,7 @@ function Hospital:countPatients()
   -- i.e. calling it from town map to use here
   -- so Town map now takes this information from here.  (If I am wrong, put it back)
   self.patientcount = 0
-  for _, patient in pairs(self.patients) do
+  for _, patient in ipairs(self.patients) do
```
TheCycoONE: Any reason to switch to ipairs here?

TheCycoONE:
924fbaa3c6d3bc82c2b56ab0d542d29b10f67de4 is dubious. A function to remove from a table by value could be useful in utils, repeating the code doesn't seem better to me.

TheCycoONE:
3a7f11f61fa4b2b6c85f65a058b5f3f17a6e00c6 is another one I question the value of. Patient may not be the right place for it, but room is worse imho.

CorsixTH/Lua/hospital.lua
Alberth289346: Other places use it as a list (which is stupid imho, but I am doing patient rather than hospital atm).

TheCycoONE:
With the exception for the two commits I've commented I'm very happy with the direction of this patch.

Alberth289346:
just dropped commit 3a7f11f although I disagree with the patient updating hospital administration.

reverted the local function thing of commit 924fbaa too

PR #1026, opened 2016-04-03
[RDY] Remove initial warmth grace period, reduce impact of too cold/warm. Fixes #74.

### LINKED ISSUE #74: Cold hospital has low consequences
I played the first level of the game with barely no heating and patients such as doctors where quite upset. Nearly every patient had a frowny face and doctors were asking for raises/bonuses very often.

The same scenario applied to CorsixTH does not go that bad. Patients will be cold but not too upset and doctors will barely complain. Also, the heat indicator will be at ~15-25% even though there is not one single heater in the hospital. It was quite close to 0% in the original game.

Alberth289346:
Original patch #225 aiming to fix #74 by MarkL, which seemed mostly ok, but the last fix was never done, instead the patch was closed.

Original comment:
````
To replace issue #81 and issue #121

This will allow the player to turn off the initial grace period in
config.txt.
Which is what I think issue #74 is about. Currently happiness does not
go down if it is too cold in the first year.
```

Updates done: Replaced boolean value by month count, moved grace period handling to the monthly hospital function, tuned down the happiness changes a bit (0.02 -> 0.015 factor).

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -48,7 +48,6 @@ function Hospital:Hospital(world, avail_rooms, name)
   self.acc_overdraft = 0
   self.acc_heating = 0
   self.discover_autopsy_risk = 10
-  self.initial_grace = true
 
   -- The sum of all material values (tiles, rooms, objects).
   -- Initial value: hospital tile count * tile value + 20000
@@ -648,6 +647,10 @@ function Hospital:afterLoad(old, new)
     -- price distortion
     self.under_priced_threshold = -0.4
     self.over_priced_threshold = 0.3
+  end
+
+  if old < 111 then
+    self.initial_grace = nil
   end
 
   -- Update other objects in the hospital (added in version 106).
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -812,19 +812,23 @@ function Humanoid:tickDay()
 
   -- If it is too hot or too cold, start to decrease happiness and
   -- show the corresponding icon. Otherwise we could get happier instead.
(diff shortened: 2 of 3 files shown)


Alberth289346:
More tweaking done. compared to #225, replaced boolean value by month count, default is 3 months instead of 12 (should be sufficient to setup a heating system, but not too long as otherwise #74 isn't fixed). Users can extend the period to 18 months.

Reverted tuning down the happiness change (so it's 0.02 like it was).

Hospital itself handles decrement of the counter, rather than the world doing it on their behalf.

CorsixTH/Lua/hospital.lua
```diff
@@ -650,6 +650,12 @@ function Hospital:afterLoad(old, new)
     self.over_priced_threshold = 0.3
   end
 
+  if old < 111 then
+    -- Warmth grace period in months (12 as that was the old default value).
+    self.warmth_grace_period = self.initial_grace and 12 or 0
```
TheCycoONE: Will be 12 months since save. Need to subtract current month.

TheCycoONE:
The PR looks good (except my one comment, above).

I have strong reservations about putting a setting like this into the config file though. It means if we ever decide to do a balance pass later everyone with an old install will still have their old values. I also don't see the need for a grace period when there is more than enough money and time to put the game on the slowest setting and build a fine starting hospital including radiators before any patients arrive.

Alberth289346:
New attempt.

Remove initial grace period but, reduce negative impact from 0.14 to 0.10
    
Grace period was introduced as it was reported that CorsixTH became too cold too quickly. This change started as a report that it was not sufficiently becoming cold (github issue #74). Rather than make the period configurable, it was decided to remove the period entirely and let the user deal with it (github PR #1026).
    
0.02 / 0.14 = 0.14285714285714285
0.10 is less than that, happiness will decrease less quickly.

PR #1030, opened 2016-04-07
[RDY] Updated LICENCE.txt for 2016 and expand author list

TheCycoONE:
Some effort was made to capture the names of contributors. I did not include
contributors in the list who did not include their full name for their commit.
I also used a very crude process of selecting those who did include their full
name based loosely on commit count. I encourage anyone missing from the list
to update the file with the appropriate details or contact me with your full
name and the name that appears on the author line of your commits.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,7 +1,59 @@
 CorsixTH is free software, distributed under the following terms:
 
-Copyright (C) 2009-2015 Peter "Corsix" Cawley, Edvin "Lego3" Linge,
-  Haico "brainball" van Roeden, Manuel "Roujin" Wolf, et al.
+Copyright (C) 2009-2016 Peter "Corsix" Cawley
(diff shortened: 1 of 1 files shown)


TheCycoONE:
@Alberth289346 you are certainly the most prolific contributor not named. I don't see anywhere where you've used your full name.

TheCycoONE:
I need to extend this list with the names of people who have copyrights in individual files. Some significant contributors never directly committed anything while we were on SVN.

TheCycoONE:
I've updated LICENCE.txt, I believe it now contains the name of everyone that ever claimed copyright in CorsixTH, as well as some other contributors who did not claim copyright in any files.

Lego3:
Nice! :-) Ribeiro is there twice instead of 2011-2012

TheCycoONE:
Thanks @Lego3 I fixed that and made the file encoding consistently utf-8. It had been using latin-1 which may not work for some names.

PR #1034, opened 2016-04-07
[RDY] Fixes regression caused by commit b54c5a8

### LINKED ISSUE #1031: Firing staff does not remove them from payroll and they keep costing money
#### Describe the issue

Firing staff does not remove them from payroll and they keep costing money

#### Steps to Reproduce
1. Hire doctor

2. Fire doctor

3. Check expense log after a month ends

#### Expected Behavior

Staff stop costing money when you fire them

#### System Information
CorsixTH Version: 0.60 beta 1

Operating System: Windows 7

### Additional Info

http://i.imgur.com/pJ0obdx.jpg

sadger:


TheCycoONE:
Perhaps you could clean up the setHospital overrides that this replaces at the same time. The patient one in particular still conditionally checks for the nil hospital case. The staff one can be simplified because it would only be called when adding the staff to the hospital, transferring staff is not an option in Theme Hospital.

MarkL1961:
@TheCycoONE  Regarding transferring staff is not an option in Theme Hospital

I think it was in the multiplayer @Lego3 may remember, I think poaching staff was an option, or I could be mixing this up with something else

TheCycoONE:
I remember litter bombs, auctions on the town map, and emergencies cycling between the hospitals when they were refused. I think the hiring list might have been shared but I don't recall any way to actually poach / hire staff off another hospital. It has been a couple years since I've played multiplayer though, and I've only played a handful of games.

MarkL1961:
I have only ever play one or two myself, but check out the language strings (which may have been used)
"One of your staff has been poached by another Hospital."
 "No way! Someone is already trying to poach that person."
"Somebody is poaching your staff. You're going to have to fight to keep them."

MarkL1961:
I think you would have to click on the other players staff and the other player would see a pay rise request to stay or leave

TheCycoONE:
Thanks for pointing that out, I'll have to test it. I think it should be relatively easy to add setHospital for staff that does the appropriate thing when the time comes (compared to some of the other challenges in implementing multiplayer).

PR #1036, opened 2016-04-14
[RDY] Refactor vip

Alberth289346:


CorsixTH/Lua/entities/vip.lua
```diff
@@ -255,6 +255,22 @@ function Vip:evaluateEmergency(success)
...
+--! the provided value is at or below the upper bound.
+--!return Evaluation value.
+local function evaluateProperty(value, eval_table)
+  for _, boundary in ipairs(eval_table) do
+    if not boundary.upper or boundary.upper >= value then
+      return boundary.value
+    end
+  end
+  -- Should never get here, but just in case
+  return 0
```
TheCycoONE: Should assert or print paranoia debug output here, don't want exception cases to be silent.

CorsixTH/Lua/entities/vip.lua
```diff
@@ -255,6 +255,22 @@ function Vip:evaluateEmergency(success)
   end
 end
 
+--! Evaluate a value based on an evaluation array.
+--!param value (number) Value to evaluate.
+--!param eval_table (table of {upper, value}) Evaluation array with increasing
+--! upper bound (leave out for the final entry), and the value to return if
+--! the provided value is at or below the upper bound.
+--!return Evaluation value.
+local function evaluateProperty(value, eval_table)
```
TheCycoONE: name isn't great. Maybe rangeMapLookup?
Could possibly be general enough (or generalized enough) to move up to utilities.

CorsixTH/Lua/entities/vip.lua
```diff
@@ -265,15 +281,14 @@ function Vip:setVIPRating()
...
-    if deathRatio > 12 then
-      self.vip_rating = self.vip_rating + 10
-    elseif deathRatio >= 8 then
-      self.vip_rating = self.vip_rating + 5
-    elseif deathRatio >= 2 and deathRatio <= 4 then
-      self.vip_rating = self.vip_rating - 10
-    elseif deathRatio < 2 then
-      self.vip_rating = self.vip_rating - 20
-    end
+    local death_ratio_table = {
```
TheCycoONE: If we agree to call this a range map above I'd like to keep that name here - e.g. death_ratio_rangemap.

TheCycoONE:
:+1: from sight inspection of all the averages changes.

Alberth289346:
Generalized the rangemap function, and moved it to utils; merged all the averaging changes into the 2nd commit. Aside from 3 additional comments "default value does not generate a message", it's equivalent.
Added new commit to rename camel case variables to lower_underscore_case to unify the variable names somewhat. Tested with a level 1 save, with a VIP visit, giving a report.

PR #1046, opened 2016-04-20
Add missing comma in README.md

wolfy1339:
I believe there should be a comma there because it reads better


TheCycoONE:
The sentence still feels wrong to me, with or without the comma. Perhaps it should be split?

PR #1050, opened 2016-04-21
Do not kill cured emergency patients after emergency

### LINKED ISSUE #1032: Emergencies
#### Describe the issue
When you do an emergency, the last patient always "dies" even if he does not (boo sound gets played)
If the emergency is 10 people, you will get 10 deaths added to your death count even if you cure them all

#### Steps to Reproduce
1. Start emergency

2. Cure everybody

3. Check death count

#### Expected Behavior

The actual number of deaths get added to the death count

#### System Information

CorsixTH Version: 0.60 beta 1

Operating System: Windows 7

TheCycoONE:
Fixes #1032

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1256,7 +1256,7 @@ function Hospital:resolveEmergency()
   local emer = self.emergency
   local rescued_patients = emer.cured_emergency_patients
   for i, patient in ipairs(self.emergency_patients) do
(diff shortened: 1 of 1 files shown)


Lego3:
Hm, this is the type of thing I mentioned in #993, evidently there are more places that rely on the hospital being set to nil than we anticipated. Would probably be good to do another search for even more places.

TheCycoONE:
Agreed.

Issue #1045, opened 2016-04-19
Custom maps should save under users folder.


TheCycoONE:
Currently custom maps are being saved in the same directory as the maps shipped with CorsixTH which is often under Program Files in Windows and probably /usr/share on Linux. Programs are not suppose to be saving files in these directories (and may not have write access to them).

We need to be saving under %APPDATA%/CorsixTH/Levels on Windows and the equivalent directory on Linux. When showing the list of custom levels to load, we should include both levels in that folder and levels shipped with the game.

Lego3:
On Windows it might be better to start saving stuff that users are supposed to easily find in e.g. My Documents?

TheCycoONE:
@Lego3 not a bad idea but out of scope for this issue. My main concern is that without fixing this people that use a Linux distribution package will not be able to save their maps. That is probably a 0.60 blocker.

PR #1052, opened 2016-04-21
[RDY] Adds a user level directory

TheCycoONE:
Similar to the savegame directory, this commit adds a user level file. By
default this uses a new Levels directory in the same path as the config. When
picking a single scenario both the users level directory and the built in level
directory are checked.

Addresses #1045

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -115,9 +115,8 @@ function App:init()
   self.good_install_folder = good_install_folder
   -- self:checkLanguageFile()
   self.level_dir = debug.getinfo(1, "S").source:sub(2, -12) .. "Levels" .. pathsep
-
+  self:initUserLevelDir()
   self:initSavegameDir()
-
   self:initScreenshotsDir()
 
   -- Create the window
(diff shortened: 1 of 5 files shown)


CorsixTH/Lua/config_finder.lua
```diff
@@ -376,6 +376,14 @@ unicode_font = nil -- [[X:\ThemeHospital\font.ttc]]
 --
 savegames = nil -- [[X:\ThemeHospital\Saves]]
 
+-------------------------------------------------------------------------------------------------------------------------
+-- Levels. By default, the "Levels" directory alongside this config file will
+-- be used for storing new maps / levels in. Should this not be suitable, then
+-- uncomment the following line, and point it to a directory which exists and
+-- is more suitable.
+--
+levels = nil -- [[X:\ThemeHospital\Saves]]
```
Alberth289346: \Levels  instead of  \Saves  ?

Alberth289346:
Looks like the right direction to me

CorsixTH/Lua/config_finder.lua
TheCycoONE: Ah yes, it was a very copy pasta sort of patch.

TheCycoONE:
Would someone mind to check this and merge or let me know of any changes I need to make before Sunday?

CorsixTH/Lua/app.lua
```diff
@@ -324,6 +323,28 @@ function App:init()
...
+  if lfs.attributes(self.user_level_dir, "mode") ~= "directory" then
+    if not lfs.mkdir(self.user_level_dir) then
+       print("Notice: Level directory does not exist and could not be created.")
+       return false
+    end
+  end
+  if self.user_level_dir:sub(-1, -1) ~= pathsep then
+    self.user_level_dir = self.user_level_dir .. pathsep
+  end
+  return true
```
Lego3: Why do we return true or false if it is not used in the call to this function anyway?

TheCycoONE: It's copy / pasted from initSavegameDir below. I didn't want to expand the pr by consolidating them this late into 0.60, but they have identical functionality.

PR #1056, opened 2016-04-24
[RFC] Allow handyman priorities to be adjusted with 'click-and-hold'.

### LINKED ISSUE #1055: Handyman priorities can't be adjusted via click-and-hold
#### Describe the issue

In the original Theme Hospital you can click and hold a handyman priority (cleaning, watering plants, repairing machinery) to increase it. In CorsixTH you need to click, and click, and click

#### Steps to Reproduce
Hire a handyman. Try adjusting his priorities with click and hold.

#### Expected Behavior

Clicking and holding a priority should keep increasing it till it hits max.

#### System Information
**CorsixTH Version: 0.60-beta2

**Operating System: Windows 10

sgtbigman:
- Closes #1055.

An issue I've ran into with implementing this: if I saved a game with a handyman's dialog window open that used the 'old', non-repeat buttons, and then loaded that game using my code changes to the buttons, the buttons are non-repeat ones until I close and reopen the dialog window.

@sadger suggested implementing some type of `afterLoad()` functionality, if we don't want this behavior to persist. Feel free to offer your suggestions/thoughts. :)

Thanks.

### DIFF
--- a/CorsixTH/Lua/dialogs/staff_dialog.lua
+++ b/CorsixTH/Lua/dialogs/staff_dialog.lua
@@ -97,9 +97,9 @@ function UIStaff:UIStaff(ui, staff)
     end
     self:addPanel(302,   5, 205) -- View circle top/Wage
     self:addPanel(313,  15, 189) -- Tasks bottom
-    self:addPanel(314,  37, 145):makeButton(0, 0, 49, 48, 315, self.doMoreCleaning):setTooltip(_S.tooltip.handyman_window.prio_litter)
-    self:addPanel(316,  92, 145):makeButton(0, 0, 49, 48, 317, self.doMoreWatering):setTooltip(_S.tooltip.handyman_window.prio_plants)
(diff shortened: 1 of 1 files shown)


TheCycoONE:
This will land after 0.60 since we're in feature freeze right now. Good work!

sadger:
Thought so @TheCycoONE .  I'll put WIP for this one just because we need to decide if we want after load so the behaviour works for old saves.

TheCycoONE:
If we are deciding something (rather than the author writing something) lets make this RFC. Otherwise no action is going to happen.

TheCycoONE:
An afterload here is non-trivial, I'm in favour of merging as is.

Alberth289346:
Any effort for afterload seems overkill if it only delays introduction of the feature due to an opened window.

TheCycoONE:
Not only that, but any new staff dialog would have the new feature because it's newly constructed, so it would just be for open windows. (Comment based on reading source, not testing)

TheCycoONE:
Verified, works in saves.

PR #1043, opened 2016-04-19
[RDY] FFmpeg 3 deprecation fixes

TheCycoONE:
Replaces functions that are deprecated in ffmpeg 3.0. Compatibility is preserved with all previous FFmpeg and libav versions we support. These issues were mentioned in #1042

PR #1057, opened 2016-04-24
[RFC] Restrict av_image_get_buffer_size to proper libav versions

TheCycoONE:
Fixes a flaw with the handling of libav in #1043

Needs verification from @bobcanthelpyou


TheCycoONE:
These functions weren't added to libav until very recently: https://git.libav.org/?p=libav.git;a=blob;f=doc/APIchanges;h=c3b81faf36d13448896e4eb21820f2de50ecad58;hb=HEAD#l192

I've updated the code to reflect that.

bobcanthelpyou:
Could compile the game on Debian 8, now.

Thank you :)

PR #1070, opened 2016-05-07
Fix #1064 - Machine smoke not disappearing after being replaced

### LINKED ISSUE #1064: Machine smoke freezes on machine replacement
#### Describe the issue
When a machine (in this case, a Slicer) was at a poor state of repair and started smoking, I selected it and chose to replace it (the handymen were too far away and I was in the midst of an emergency). When I did that, the number of times used went down to 0 as normal, but the smoke froze instead of vanishing.

#### Steps to Reproduce
1. Get a machine down far enough that it starts to smoke

2. Click on it and choose to replace it

#### Expected Behavior
Smoke should vanish completely

#### System Information
**CorsixTH Version:** *0.60 Release Candidate 1*

**Operating System:** *Windows 10 64-bit*

**Theme Hospital Version:** *CD*

#### Gamelog.txt

---------------------------------------------------------------

Welcome to CorsixTH v0.60-rc1!

---------------------------------------------------------------

This window will display useful information if an error occurs.

---------------------------------------------------------------

Checking for CorsixTH updates...
You are running the latest version of CorsixTH.

### Additional Info
*Screen shot (2 uses after replacing machine):*
image

JoshKeegan:


CorsixTH/Lua/entities/machine.lua
```diff
@@ -255,6 +255,28 @@ function Machine:createHandymanActions(handyman)
...
 end
 
+--! Replace this machine (make it pretend it's brand new)
+function Machine:machineReplaced()
+  -- Reset usage stats
+  self.total_usage = 0
+  self.times_used = 0
+  
+  -- Update strength to match the current level of research for it
+  local strength = self.hospital.research.research_progress[self.object_type].start_strength
```
TheCycoONE: I don't think you use strength anywhere except the next line. You might as well just have ```self.strength = self.hospital...```

TheCycoONE:
Looks good, one minor nit pick and then it should be ready to merge.

JoshKeegan:
done :)

Issue #732, opened 2015-04-16
Infinite loop causes crash on spawnPatient


crucifigo:
Update: I restarted and opened the hospital immediately with one receptionist and nothing else, and it still crashed.

The timer ran down on my hospital and I turned off pause, and when the first patient would normally arrive, the game freezes up.

What steps will reproduce the problem:
Simply leaving it or turning off pause.

What is the expected output? What do you see instead?
The game freezes instead of spawning patients.

What version of CorsixTH are you using (e.g. "0.30", "4c66e39985")?
0.40

What operating system / compile settings are you using?
Win7

What level was this on (e.g. "Demo level", "Full game level 12")?
Full game lv.11 (Croaking Hospital)

Upload a saved game file (preferably an autosave from before the error) if you
have one and it is relevant:

Dropbox: https://www.dropbox.com/s/0577fqrrn3n9lwb/Main.sav?dl=0
![image](https://cloud.githubusercontent.com/assets/11978903/7181578/be980a72-e443-11e4-80ff-4f8d588a8014.png)

TheCycoONE:
Copying stack trace so that a search pulls up this issue:
world.lua:531 Suspected infinite loop

world.lua: in function 'spawnPatient'
world.lua:1097: in function 'onTick'
app.lua:852: in function 'dispatch'
app.lua:753: in function <app.lua:748>

TheCycoONE:
The code block that is failing is:

```lua
    local disease = self.available_diseases[math.random(1, #self.available_diseases)]
    while disease.only_emergency or not isVisualDiseaseAvailable(disease) do
      disease = self.available_diseases[math.random(1, #self.available_diseases)]
    end
```
which suggests possibly that the disease list being loaded isn't complete.

Would you be able to paste the md5sums of your EASY11.SAM, FULL11.SAM, and HARD11.SAM files.  Of course mention if you have done any modifications yourself to either the level files or the lua.

PR #1071, opened 2016-05-11
[rdy] Avoid infinite looping if there are no viable diseases for new patients. Fixes issue #732

Alberth289346:
If a level file only has visible diseases, and these are postponed until the future by the level file, the unfixed code would loop forever.

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -479,6 +479,10 @@ end
 --!param hospital (Hospital) Hospital that the new patient should visit.
 --!return (Patient entity) The spawned patient, or 'nil' if no patient spawned.
 function World:spawnPatient(hospital)
+  if not hospital then
+    hospital = self:getLocalPlayerHospital()
+  end
+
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/world.lua
```diff
@@ -447,10 +447,42 @@ function World:calculateSpawnTiles()
   end
 end
 
+--! Function to determine whether a given disease is available for new patients.
+--!param disease (disease) Disease to test.
+--!param hospital (Hospital) Hospital that needs a new patient.
+--!return (boolean) Whether the disease is available for new patients.
+function World:isDiseaseAvailable(disease, hospital)
```
TheCycoONE: Should be a local function still (just top level), pass in self as a parameter.

Alberth289346:
More, smaller steps, nicer fix, seems to work, both the lucky shot and the unlucky shot

TheCycoONE:
Looks quite a bit better, thanks @Alberth289346

PR #1074, opened 2016-05-17
norwegian.lua

Grimdoc:
And another try.
This isn't self-explanatory. 
There should be a much easier way to update translations.

### DIFF
--- a/CorsixTH/Lua/languages/norwegian.lua
+++ b/CorsixTH/Lua/languages/norwegian.lua
@@ -104,7 +104,6 @@ tooltip.message.button_dismiss = "Venstreklikk for å åpne faksen, höyreklikk 
 menu_options = {
   lock_windows = "  LÅS VINDUER  ",
   edge_scrolling = "  KANTSCROLLING  ",
-  settings = "  INNSTILLINGER  ",
   adviser_disabled = "  RÅDGIVER  ",
   warmth_colors = "  VARMEFARGER", --Todo: Open for suggestions for a better word for it (warmth colours).
 }
@@ -206,7 +205,6 @@ custom_game_window = {
   load_selected_level = "Start",
 }
 tooltip.custom_game_window = {
-  start_game_with_name = "Last brett %s",
   free_build = "Huk av denne boksen om du vil spille uten penger og uten bestemte forutsetninger for vinne eller tape", --todo: finne en bedre oversettelse eller omskrive
   load_selected_level = "Last og spill valgt brett",
   choose_game = "Klikk på et brett for å lese mer om det",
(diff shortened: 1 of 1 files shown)


TheCycoONE:
@Grimdoc I think @SimplyRyan is working on something easier for you. For now just install the latest release candidate, make your changes, and attach it to an issue with dropbox or copy it in surrounded by ``` tags.

See us on irc about cleaning up your repository, and never make a pull request from master.

Grimdoc:
Sure.

With SVN I actually had control of what I was doing.
To be able to use GitHub effectively you have to use it often. But only programmers have the desire/need to use it often. I think this is the biggest barrier for the translators.

Alberth289346:
With git you also have control, but there are more steps to do, and it's easier to make a mess.
I believe the barrier is not only for translators, also new lua programmers are hit by it.

MarkL1961:
Agreed :+1:

TheCycoONE:
I don't want this to devolve into git vs svn, or github vs googlecode; lets work on solutions instead.

Clearly pull requests are complicated for casual users, and possibly made even more so by github's public user repositories. I assume that creating patch files is not any easier for casual users.

My suggestion is that people create a regular issue (not a pull request). In that issue they should say the commit hash or release version of CorsixTH that they started from and paste the entire contents of the changed file. That way one of the regular contributors can turn it into a pr and that person can manage the patch if anyone else has corrections.

Is there any problem with this?

Grimdoc:
Sounds ok.
Unfortunately this means more work for the coders.

MarkL1961:
For me it is not creating the pull that is the problem it is when you have to make changes as it is so easy to then get in a mess and have a branch you cannot get out of. Or splitting things to then be asked to merge them.  For those used to git it is a great tool, but a complicated one for those more used to a windows interface like svn.
If I choose to contribute in the future I will take you up on the offer as above
Thanks

Grimdoc:
Is it best to use one permanent issue for all languages, one permanent for each language, or one new issue for each single language update?

Alberth289346:
@MarkL1961 if making a branch is not a problem, that could work too perhaps. Make a branch, push it onto your repository, and make an issue with the link (a PR would get messy, as the PR gets linked to your repository, which means no-one else can easily change it).
(At least I hope you do make branches, or an update from master, or the second patch becomes a mess.)

@TheCycoONE  One small addition, please also mention the name of the file, with directory if possible. Just in case there is confusion about which file the issue is.

@Grimdoc  in general, 1 issue for 1 language update (where "language update" may change as many strings as you like) is the least confusing. The person modifying it into a PR should worry about splitting into separate commits and so on.

Not sure how to deal with changes in several files, one comment for each file? Won't happen in translation updates, I guess.

PR #1082, opened 2016-06-01
[RFC] Load hospital and build flags for all tiles when in map editor

### LINKED ISSUE #1044: Inside tiles not buildable when placed before the parcel
#### Describe the issue
I created a map on the map editor, but when I tried it in game it wouldnt let me build on it. I transfered to the outside map editor to check the flags on it and sure enough they weren't marked with the "H", I checked if every corner was closed and they were,after some tinkering I decided to remove and add again the back walls. Got my "H"s by doing so, saved, went back to the game. Couldnt build. Loaded the map again on my map editor. The "H"s were gone. After some failed attemps to fix it and a research with no results about this issue I gave up and decided to report the bug.

#### Steps to Reproduce
1. Load the .map file on map editor
2. Check the flags, no H marked
3. Remove the back wall and replace with a new one
4. Check the flag, got the H marks
5. Save the map

#### Save Game
 Folder with .map file and .level:
https://www.dropbox.com/sh/izyffd8zmfop3df/AABaG4na9sLVtw4xdjK56BZGa?dl=0 

#### Expected Behavior

6. Load it again or try in game
6.1 - if loaded again H marks are gone
6.2 - if tried in game can't build

#### System Information
**CorsixTH Version:0.50

**Operating System: Windows 8.1

#### Gamelog.txt
No game log file
```

### Additional Info
![screenshot 2016-04-19 05 18 23](https://cloud.githubusercontent.com/assets/9095043/14632353/83dded02-05ee-11e6-9fc2-239f444b9df4.png)
![screenshot 2016-04-19 05 18 51](https://cloud.githubusercontent.com/assets/9095043/14632356/8bdbcf10-05ee-11e6-98cd-31f1e32482c8.png)
![screenshot 2016-04-19 05 19 44](https://cloud.githubusercontent.com/assets/9095043/14632366/919af64c-05ee-11e6-9f40-0259a3a1ec66.png)

TheCycoONE:
Fixes #1044 

When we move towards multiplayer this may change again because there will be purchased plots that cannot be built on - we will need to check the owner as well as the flags for all builds and there will be no need to not load flags from parcel 0. In the mean time this fix is low impact.

### DIFF
--- a/CorsixTH/Lua/map.lua
+++ b/CorsixTH/Lua/map.lua
@@ -195,7 +195,7 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
     local data
     data, errors = self:getRawData(map_file)
     if data then
-      i, objects = self.th:load(data)
+      i, objects = self.th:load(data, map_editor)
     else
       return nil, errors
     end
@@ -230,7 +230,7 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
     else
       local data, errors = self:getRawData(level)
       if data then
-        i, objects = self.th:load(data)
+        i, objects = self.th:load(data, map_editor)
       else
         return nil, errors
       end
@@ -246,7 +246,7 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
     self.map_file = map_file
     local data, errors = self:getRawData(map_file)
     if data then
-      i, objects = self.th:load(data)
+      i, objects = self.th:load(data, map_editor)
     else
       return nil, errors
     end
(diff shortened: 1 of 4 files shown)


TheCycoONE:
Before this is merged, we should discuss whether it is worth making a 0.61 patch release for.

The workaround in the mean time is to redraw the inside tiles after loading the map if you had not placed a parcel on it before saving; alternately always place parcels before saving.

MarkL1961:
If you were to do a 0.61 you could make it clear that it is a patch to fix the map editor.  Some people then may not bother as I don't think it is all players who delve into making their own maps

TheCycoONE:
We can see if anything else significant comes up in the next couple weeks. If this is going to be for 0.70 I will do the proper fix (checking for owner everywhere buildable is checked) instead.

TheCycoONE:
Only preserving this in case we want to make a patch and #1086 is too complex.

PR #1086, opened 2016-06-03
[RDY] Load flags regardless of parcel, and ensure player match.

### LINKED ISSUE #1044: Inside tiles not buildable when placed before the parcel

TheCycoONE:
Corrects an issue with multiplayer maps where one player would be able to
build on another player's hospital. Also fixes a bug in the map editor
causing flags to be lost if they were not assigned a parcel before being
saved.

This fixes #1044 and obsoletes #1082 though it is a more complicated patch.

I have tested the cases I could think of, but I may have missed something.

### DIFF
--- a/CorsixTH/Lua/dialogs/place_staff.lua
+++ b/CorsixTH/Lua/dialogs/place_staff.lua
@@ -81,9 +81,11 @@ function UIPlaceStaff:draw(canvas)
   if self.world.user_actions_allowed then
     self.world.map.th:getCellFlags(self.tile_x, self.tile_y, flag_cache)
     local room = self.world:getRoom(self.tile_x, self.tile_y)
+    local player_id = self.ui.hospital:getPlayerIndex()
     local valid = flag_cache.hospital and flag_cache.passable and
-      (self.allow_in_rooms or flag_cache.roomId == 0) and
-      (not room and true or not room.crashed)
+        (self.allow_in_rooms or flag_cache.roomId == 0) and
+        (not room and true or not room.crashed) and
+        flag_cache.owner == player_id
     self.anim:setFlag(valid and 0 or flag_altpal)
     local zoom = self.ui.zoom_factor
     if canvas:scale(zoom) then
@@ -107,8 +109,11 @@ function UIPlaceStaff:onMouseUp(button, x, y)
       self:onMouseMove(x, y)
       self.world.map.th:getCellFlags(self.tile_x, self.tile_y, flag_cache)
       local room = self.world:getRoom(self.tile_x, self.tile_y)
+      local player_id = self.ui.hospital:getPlayerIndex()
       if flag_cache.hospital and flag_cache.passable and
-          (self.allow_in_rooms or flag_cache.roomId == 0) and (not room or not room.crashed) then
+          (self.allow_in_rooms or flag_cache.roomId == 0) and
+          (not room or not room.crashed) and
+          flag_cache.owner == player_id then
         if self.staff then
           self.staff:setTile(self.tile_x, self.tile_y)
         else
--- a/CorsixTH/Lua/humanoid_actions/seek_reception.lua
+++ b/CorsixTH/Lua/humanoid_actions/seek_reception.lua
@@ -20,14 +20,17 @@ SOFTWARE. --]]
 
 local function can_join_queue_at(humanoid, x, y, use_x, use_y)
   local flag_cache = humanoid.world.map.th:getCellFlags(x, y)
-  return flag_cache.hospital and not flag_cache.room
+  return flag_cache.hospital and not flag_cache.room and
+      humanoid.hospital and
+      flag_cache.owner == humanoid.hospital:getPlayerIndex()
 end
 
 local function action_seek_reception_start(action, humanoid)
   local world = humanoid.world
-
   local best_desk
   local score
+
+  assert(humanoid.hospital, "humanoid must be associated with a hospital to seek reception")
 
   -- Go through all receptions desks.
   for _, desk in ipairs(humanoid.hospital:findReceptionDesks()) do
@@ -92,7 +95,7 @@ local function action_seek_reception_start(action, humanoid)
     -- No reception desk found. One will probably be built soon, somewhere in
     -- the hospital, so either walk to the hospital, or walk around the hospital.
     local procrastination
-    if world.map.th:getCellFlags(humanoid.tile_x, humanoid.tile_y).hospital then
+    if humanoid.hospital:isInHospital(humanoid.tile_x, humanoid.tile_y) then
       procrastination = {name = "meander", count = 1}
       if not humanoid.waiting then
         -- Eventually people are going to get bored and leave.
--- a/CorsixTH/Src/th_lua_map.cpp
+++ b/CorsixTH/Src/th_lua_map.cpp
@@ -123,7 +123,7 @@ static int l_map_save(lua_State *L)
 static THAnimation* l_map_updateblueprint_getnextanim(lua_State *L, int& iIndex)
 {
     THAnimation *pAnim;
-    lua_rawgeti(L, 10, iIndex);
+    lua_rawgeti(L, 11, iIndex);
     if(lua_type(L, -1) == LUA_TNIL)
     {
         lua_pop(L, 1);
@@ -133,10 +133,10 @@ static THAnimation* l_map_updateblueprint_getnextanim(lua_State *L, int& iIndex)
         lua_createtable(L, 0, 2);
         lua_pushvalue(L, 1);
         lua_setfield(L, -2, "map");
-        lua_pushvalue(L, 11);
+        lua_pushvalue(L, 12);
         lua_setfield(L, -2, "animator");
         lua_setfenv(L, -2);
-        lua_rawseti(L, 10, iIndex);
+        lua_rawseti(L, 11, iIndex);
     }
     else
     {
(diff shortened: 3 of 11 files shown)


MarkL1961:
@TheCycoONE  I wanted to try this out, I downloaded the test build but the game won't run. (closes on open with no log report)
I know it says I need Microsoft Visual C 2013 Runtime. It says the same for 0.60 and that works so I assume I have that.
What am I missing?

TheCycoONE:
I need to bump the api version. That's not related to MarkL's problems though, the app veyor version runs for me. Unlike the release version it's only 64bit. The run time required is VS 2015, but that's for the release version as well.

MarkL1961:
I think that is what I have

runtime

TheCycoONE:
I believe the problem is that AppVeyor is making debug builds, which depend on the debug runtime, which is not independent of Visual Studio and not redistributable. I'll correct AppVeyor to make release builds.

TheCycoONE:
Alright, I've rebased on the appveyor changes, and bumped the api version. I neglected to test anything related to staff so there may be issues there.

TheCycoONE:
@MarkL1961 not sure what your plans are, but you should be able to test this now.

MarkL1961:
Unless I am missing something I have tested this with #1089 and gave it the thumbs up.
Drag is working as I would expect and is smooth and I can run the exe created by appveyor now
If I missed something please shout up
Thanks

TheCycoONE:
This PR is about not being able to build on other players hospitals (or any tile with the build flag which is not in a parcel owned by the current player). I tested placing objects and building / editing rooms myself somewhat. I didn't deal with placing staff, but it occurs to me that you should only be able to place staff on tiles you own too. So generally things relating to land ownership.

MarkL1961:
The saving worked for me anyway.

At the weekend I will create a multiplayer map and see what restrictions there are.
From my limited (twice I think) knowledge of multiplayer there are some things you are allowed to do.  Placing litter bombs and moving them was one and the other was poaching staff.  I can't remember exactly how that worked, but I am sure if that became an issue later a fix could be found?

So for testing, create a map for more than one player and parcel it into areas for starting for each player with an helipad and some un-owned areas for players to buy?  I can't think how you zone an area for each player from the controls I have seen.  I saw something about different helipads though.

If I have any problems I'll create an issue for each one

TheCycoONE:
Making a multiplayer map won't work because the game doesn't support it. What I did was made a hospital that wasn't part of any parcel. The current map editor gives it the build flag, but it's not owned by the player and it's not removed by the parcel purchase stuff.

MarkL1961:
Okay

TheCycoONE:
I've now also updated the PR to include properly assigning the parcels to other players on multiplayer maps, so your original method would work. Additionally I've rebased so you can have the map editor performance improvements.

CorsixTH/Src/th_lua_map.cpp
```diff
@@ -191,9 +191,12 @@ static int l_map_gettemperature(lua_State *L)
  * @param pNode Node to examine.
  * @return Whether the node position is valid for a new room.
  */
-static inline bool is_valid(bool entire_invalid, const THMapNode *pNode)
+static inline bool is_valid(
+    bool entire_invalid, const THMapNode *pNode,
+    const THMap *pMap, int player_id)
```
Alberth289346: doxygen comments are not adjusted for these new parameters

TheCycoONE: Thanks, I will correct that.

MarkL1961:
I could be doing something wrong.  Built a basic map to start with and set the camera positions in each of the corner plots for each player and set the middle plots as parcels 2, 3 etc.  Corner plots are all 1.
In town map it appears different players own the other corner plots, but I can build on them.
Those parcelled 2 and up can be seen in game and in the town plan, whereas you only normally see an outline of those until you buy them.  It shows in town map that I own all the other plots, which might be why I can see them, but I can't see what you would do other than parcel other than 1 to make them unowned
Map is very basic at the moment and I do need to add walls and paths/roads - so that might cause what I am seeing.   I will work on finishing the map at the weekend.

TheCycoONE:
You're misunderstanding slightly.

Player 1 owns all parcel 1s at the start, Player 2 owns all parcel 2s from the start, Player 3 owns 3s, and 4 owns 4s. Plots above the set player count are not owned from the start. You should be able to build on all the parcel 1 hospitals, but not the parcel 2 ones.

MarkL1961:
Yes, I think I just got that as I have been updating my map.  So far sort of working as planned as you have just said the first four parcels were in different spots to where I placed the camera.   As I get the hang of map making for multi player I'll update the wiki
One point though is I don't think 9 parcels is enough as that is essentially two buildings per player and one over

TheCycoONE:
The 9 parcel restriction comes from the original game. We can see about easing it at some point perhaps though my feeling is that the 128x128 grid would feel tight after that.

The original multiplayer capitalized on the scarcity, e.g. one of the objectives was to buy a plot of land, and you could force your components out of the game by outbidding them on the properties. The games also tended to be short and frantic because of these escalating objectives.

MarkL1961:
How was that determined?
I have just opened L44 from TH and there are 16 parcels in that one

Alberth289346:
My game manual says 2-4 players in multi-player

MarkL1961:
Parcels not players

TheCycoONE:
@MarkL1961 sorry, I thought it said that in Alexander's map details, but it doesn't, it actually says that there are two bytes assigned for parcels, which is enough for more parcels than there are map tiles, so I don't know why such a high number is supported. I guess the 9 is just from the old map editor? So yes, we'll extend that at some point, please open another ticket.

MarkL1961:
Old map editor went up to 31, so that was one short based on L43 it seems

MarkL1961:
Back to testing this update:
I am able to place staff in buildings I don't own.  I am not sure whether you could in TH, but I can't see why you would want to.
You can't place objects or rooms in building you don't own as expected.
Patients wander other building, that is fine and from what I remember is the same as TH.
Town map displays the other player name for their buildings.  I think though their plots should appear in a colour for each player - this will need to be verified in TH.
I can't test the placement of the litter bomb as they are not implemented.

@TheCycoONE  Can you think of any thing else to check for?

TheCycoONE:
I've just tested the staff issues you mentioned in the original theme hospital (playing an IPX game against myself). In the original you could not place staff in other player's hospitals.

I didn't see patients wander in other player's hospitals, but the map wasn't set up that they had any reason to either - maybe there's a map where I could quickly set up such a situation?

In the mean time I will change my PR to prevent placing staff outside of owned plots.

MarkL1961:
I was using L43 which as four plots close together in the centre at the start.  
I did notice later that peeps had been sick in one of the un-owned plots and that a handyman went to clean it up.
If we stop patients wandering into other plots this won't be a problem.  It does occur though that if an handyman gets a call to clean vomit in another players building the same could happen with Dr's and nurses.  Same with plant watering and machine repairs.

Lego3:
Yes, to get multiplayer working the whole call system needs some additions. The same goes for when patients are about to go to the next room. Not  that it's a part of this PR, but at some point. :-)

TheCycoONE:
I've added restrictions on where you can place staff, and enforce patients to wait for a reception desk in their own hospital, as well as epidemic injections in their own hospital (indirectly). There are probably a lot more actions that would need to be filtered, once multiplayer is seriously implemented but I've probably caught the most obvious.

I will be cleaning this up and looking for any more obvious places to check the hospital owner then move it back to RFC. Feel free to test as is, but it's not ready to review the code yet.

TheCycoONE:
I don't think there's any regressions now. Implementing multiplayer may show some need for improvement but that can be addressed then.

CorsixTH/Lua/humanoid_actions/seek_reception.lua
```diff
@@ -20,15 +20,18 @@ SOFTWARE. --]]
 
 local function can_join_queue_at(humanoid, x, y, use_x, use_y)
   local flag_cache = humanoid.world.map.th:getCellFlags(x, y)
-  return flag_cache.hospital and not flag_cache.room
+  return flag_cache.hospital and not flag_cache.room and
+    humanoid.hospital and
```
Alberth289346: Continuation lines have double indent. I'd also merge line 25 after this one, but that's optional :)

CorsixTH/Lua/dialogs/place_staff.lua
```diff
@@ -81,9 +81,11 @@ function UIPlaceStaff:draw(canvas)
   if self.world.user_actions_allowed then
     self.world.map.th:getCellFlags(self.tile_x, self.tile_y, flag_cache)
     local room = self.world:getRoom(self.tile_x, self.tile_y)
+    local player_id = self.ui.hospital:getPlayerIndex()
     local valid = flag_cache.hospital and flag_cache.passable and
       (self.allow_in_rooms or flag_cache.roomId == 0) and
```
Alberth289346: Any chance in getting more indenting in these continuation lines?

TheCycoONE:
Fixed indentation issues.

PR #1089, opened 2016-06-03
[RDY] Fix performance issue with drag draw in mapper

### LINKED ISSUE #1084: Drag pasting in map editor slow
When dragging to paste objects/tiles the cursor stops and then jumps.  In the clip it does not seem bad with walls only floor tiles and parcel placing.  However if I try to build a larger building I get the same problem with walls as well.

see attached clip https://www.dropbox.com/s/3edgcgq8lv3yo7f/map%20edit.wmv?dl=0

Expect smoother mouse movement that allows you to place tiles in the desired place. it gets worse with larger areas and you end up with tiles where you didn't want them

Windows 9 using 0.60 and original cd data for TH

TheCycoONE:
Scale once for the entire drag area instead of once per tile. Greatly
improves performance.

Fixes #1084


MarkL1961:
:+1:

PR #1093, opened 2016-06-17
[RFC] Disables update check

TheCycoONE:
There is a remote code execution vulnerability in the update check. This
commit simply disables the update check until a proper fix can land.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -63,7 +63,9 @@ function App:App()
   }
   self.strings = {}
   self.savegame_version = SAVEGAME_VERSION
-  self.check_for_updates = true
+
+  -- Disable updates until remote exploit can be fixed.
+  self.check_for_updates = false
 end
 
 --! Starts a Lua DBGp client & connects it to a DBGp server.
@@ -1453,6 +1455,7 @@ function App:checkForUpdates()
     return
   end
 
+  -- XXX: Unsafe, we must not execute code fetched over the internet!
   local update_table = loadstring_envcall(update_body, "@updatechecker"){}
   local changelog = update_table["changelog_" .. default_language]
   local new_version = update_table["major"] .. '.' .. update_table["minor"] .. update_table["revision"]


Alberth289346:
I'd prefer to also take out the actual problem; lua seems too flexible to trust a single boolean value set somewhere.

Lego3:
I believe we only run code if it comes from our own domain CorsixTH.com. There is still a risk if that site is harmed though.

alanwoolley:
Don't disagree with the conclusion, but this was discussed by myself & Corsix at the time and deemed acceptable. What's changed? 

https://groups.google.com/d/msg/corsix-th-dev/2P9DjuL4pEk/GsUItLLdq2kJ

TheCycoONE:
Thanks for joining the discussion Alan.

@corsix himself should have known that malicious bytecode passed to loadstring_envcall could potentially escape its sandbox, he created examples himself.

As for what has changed - probably just a few more years of experience and the warning of a 3rd party who saw our bug posting on lua-socket and followed our code a bit.

alanwoolley:
If that's the case and there's a POC that demonstrates this as an exploitable vulnerability then that's clear - this should be disabled until it can be reworked.

Alberth289346:
I would really prefer a more explicit separation; maybe the lua-devs decide to revert the change made by corsix, or create a hole in it by accident. Explicit code that parses things avoid all those problems, and makes it clear to people browsing the code there is no problem.

TheCycoONE:
I think @corsix's https://gist.github.com/corsix/6575486 is a proof of concept, at least for a particular lua version and operating system. There is a workaround in lua 5.2 and higher to disable bytecode being loaded (which we don't implement). Parsing is safer though. The old xml may have been overkill, but a delimited flat file is an option.

TheCycoONE:
@Lego3 the version information is transmitted in plain text, so it's fairly trivial to intercept and replace with a payload of the attacker's choice.

Lego3:
Yes.

PR #1095, opened 2016-06-19
[RDY] Check that the Visual Studio redistributable has been installed

Lego3:
We quite often get questions/comments on e.g. Facebook that CorsixTH does not start, and most of those times this happens because the person in question does not have the correct Visual Studio redistributable installed, as required by the current release's installer.

This pull request tries to address that by checking on install for the redistributable in the registry, as suggested here: https://msdn.microsoft.com/en-us/library/ms235299.aspx

All languages except Swedish currently contain the English text.

I have only had the change to test this on a 64-bit Windows system, could someone test it on a 32-bit version as well to see whether the registry entry is possibly different?


wolfy1339:
I believe that the URL and name should be adjusted per language
e.g. in French the name would be 'Redistribuable Visual C++ pour Visual Studio 2015' and the URL would be https://www.microsoft.com/fr-fr/download/details.aspx?id=48145 (France - French)

Lego3:
I updated the pr to have localizable name and url too. I also changed the dialog to be a "download now?" Yes/No dialog, since the written url couldn't be clicked or selected to copy.

wolfy1339:
@Lego3 Here's the french text: `CorsixTH exige le '${vcredist STRING}' pour fonctionner, mais il n'a pas été trouvé sur cet ordinateur. Voulez-vous le télécharger ?`

Lego3:
Updated French language. Still to do is to test on a 32-bit system.

TheCycoONE:
@Lego3 now that it's past the end of August do you have the resources to test on 32 bit? (Though that's increasingly unimportant now that XP is no longer supported)

Lego3:
As I note in the text I unfortunately don't have a 32-bit system to test
on, and I don't know anyone with it either. I do possibly have some old
Vista version somewhere that I could create a VM from, but maybe we could
even go as far as to say that we only supply a 64-bit installer?

TheCycoONE:
@Lego3 I've been inclined for some time to go the other way and only ship the x86 installer. CorsixTH doesn't benefit from 64bit at all, and all amd64 (x86_64) systems are x86 compatible.

I think we can merge this now and delay that decision. RDY?

PR #1097, opened 2016-06-20
Normalise lua version detection for cross-platform

emorrp1:
This will make it easy to add future versions for all platforms:
* lua5x ??, presumably used by Visual Studio
* lua5.x used by debian-based linux distros
* lua-5.x ??, was there in 5.1
* liblua.5.x.dylib used by homebrew on OS X

### DIFF
--- a/CMake/FindLua.cmake
+++ b/CMake/FindLua.cmake
@@ -26,8 +26,8 @@ IF (WITH_LUAJIT)
     SET (LUA_INCLUDE_DIRS include/luajit-2.0 include)
 ELSE (WITH_LUAJIT)
     SET (LUA_INTERPRETER_TYPE "Lua")
-    SET (LUA_LIBRARY_NAME lua53 lua5.2 lua52 lua5.1 lua51 lua lua-5.1 liblua liblua.5.3.dylib)
-    SET (LUA_INCLUDE_DIRS include include/lua include/lua5.1 include/lua51 include/lua-5.1 include/lua5.2 include/lua5.3 include/lua-5.3)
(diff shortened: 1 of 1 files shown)


TheCycoONE:
While you're tweaking the order, the library order should be the same as the include order.

Kitware's module is better than ours: https://github.com/Kitware/CMake/blob/master/Modules/FindLua.cmake because it does more to ensure include and library versions match. (This is an issue on arch linux, and possibly others).

emorrp1:
Thanks, I wasn't sure if the reverse ordering between the two lines was significant. I've rebased on latest master.

I looked into the alternative FindLua you linked and it does look more thorough. It's also provided in the CMake main Modules directory (in debian /usr/share/cmake-3.0/Modules/FindLua.cmake). I got it to build with the below patch, but I'm not confident in how to make the necessary changes cross-platform:

[0001-global-FindLua.cmake.txt](https://github.com/CorsixTH/CorsixTH/files/326734/0001-global-FindLua.cmake.txt)

TheCycoONE:
The reason we are not yet using the built in module is that it does not come with CMake 2.8.12 and that version is still important to support for TravisCI. I speculate it could be back ported but simply putting newest version first for now would be an improvement.

emorrp1:
Ah, I see Travis uses Ubuntu LTS, but according to [travis-ci #5821](https://github.com/travis-ci/travis-ci/issues/5821#issuecomment-214452987): "At the moment, it is unlikely that we'll be supporting 16.04 as a native environment this year.". So it looks like CMake 3(.5) won't be available until 2017, unless you enable TravisCI's docker integration.

emorrp1:
For future reference, it looks like lua-5.x and lua5x are used in [freebsd](https://svnweb.freebsd.org/ports/head/games/corsix-th/files/patch-CMake_FindLua.cmake?view=markup&pathrev=416180)

Issue #1102, opened 2016-06-23
Building while paused without overlay in bottom panel


Brainsample:
When "allow_user_actions_while_paused" is enabled, almost everything works as advertised. The only exception is the "Pause" overlay in the bottom panel. The buttons in the bottom panel are still accessible with hotkeys, but any mouse-hover information is hidden. I thought it would be more consistent not to use the "Pause" overlay in this case.
It is easy to do:

Original \lua\dialogs\bottom_panel.lua line 206:
```lua
function UIBottomPanel:drawDynamicInfo(canvas, x, y)
  if self.world:isCurrentSpeed("Pause") then
```

Patched \lua\dialogs\bottom_panel.lua line 206:
```lua
function UIBottomPanel:drawDynamicInfo(canvas, x, y)
  if self.world:isCurrentSpeed("Pause") and not self.world.user_actions_allowed then
```

CorsixTH v0.60

TheCycoONE:
:+1: 
I remember you saying you don't do pull requests - maybe we can convince you to reconsider.

Brainsample:
Hi TheCycoONE,

It's true, every time I tried to work in a repository my head got dizzy, and dyslexia made me stop coding all together. I have what you can call a very specific skill set. :) However, I do want to share my work, so I'll try to set up a repo. Until then (or if I don't succeed) I will present my suggestions as comprehensible as possible.

On topic of the Pause overlay: The patch comes in very handy for testing purposes.

Alberth289346:
Fixed in commit 7b58de79a73345c0ba06590d5718c5171d07fee2, thanks

PR #1103, opened 2016-06-24
Allow use of bottom panel during pause

TheCycoONE:
When allow_user_actions_while_paused is enabled we don't want to block the
bottom panel buttons with the pause overlay.

Thanks Brainsample.

Patch taken from #1102


Brainsample:
You're welcome. :)

MarkL1961:
I don't understand the issue as you can still access the buttons if you are allowing build on pause.

PR #1107, opened 2016-07-02
[ready] Rewrite actions to use construction functions

Alberth289346:
New attempt of #1019 hopefully better for reviewing.

Compared to the previous attempt, things have less clutter, the "split function" commits are best reviewed using a whitespace ignore on the diff,
Ran a test on an old level 1 and level 4 save game, and that seems to run fine.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -261,7 +261,10 @@ function App:init()
 
     dofile "room"
     self.rooms = self:loadLuaFolder("rooms")
+
+    dofile "humanoid_action"
     self.humanoid_actions = self:loadLuaFolder("humanoid_actions")
+
     local diseases = self:loadLuaFolder("diseases")
     self.diseases = self:loadLuaFolder("diagnosis", nil, diseases)
 
--- a/CorsixTH/Lua/calls_dispatcher.lua
+++ b/CorsixTH/Lua/calls_dispatcher.lua
@@ -258,7 +258,7 @@ function CallsDispatcher.sendNurseToVaccinate(patient, nurse)
     -- The epidemic may have ended before the call can be executed
     -- so just finish the call immediately
     CallsDispatcher.queueCallCheckpointAction(nurse)
-    nurse:queueAction{name = "answer_call"}
+    nurse:queueAction(AnswerCallAction())
     nurse:finishAction()
     patient.reserved_for = nil
   end
@@ -485,7 +485,7 @@ function CallsDispatcher.unassignCall(call)
   assert(assigned.on_call == call, "Unassigning call but the staff was not on call or a different call")
   call.assigned = nil
   assigned.on_call = nil
-  assigned:setNextAction{name = "answer_call"}
+  assigned:setNextAction(AnswerCallAction())
 end
 
 function CallsDispatcher.verifyStaffForRoom(room, attribute, staff)
--- a/CorsixTH/Lua/calls_dispatcher.lua
+++ b/CorsixTH/Lua/calls_dispatcher.lua
@@ -418,11 +418,8 @@ end
 -- A interrupt handler could be supplied if special handling is needed.
 -- If not, the default would be reinsert the call into the queue
 function CallsDispatcher.queueCallCheckpointAction(humanoid, interrupt_handler)
-  return humanoid:queueAction{
-    name = "call_checkpoint",
-    call = humanoid.on_call,
-    on_remove = interrupt_handler or CallsDispatcher.actionInterruptHandler
-  }
+  interrupt_handler = interrupt_handler or CallsDispatcher.actionInterruptHandler
+  return humanoid:queueAction(CallCheckPointAction(humanoid.on_call, interrupt_handler))
 end
 
 -- Default checkpoint interrupt handler
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -599,7 +599,7 @@ function UIEditRoom:returnToDoorPhase()
   local room = self.room
   room.built = false
   if room.door and room.door.queue then
-    room.door.queue:rerouteAllPatients({name = "seek_room", room_type = room.room_info.id})
+    room.door.queue:rerouteAllPatients(SeekRoomAction(room.room_info.id))
   end
 
   self.purchase_button:enable(false)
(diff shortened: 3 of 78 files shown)


TheCycoONE:
Verified by visual inspection up to ec0b4f025ee

The following commit uses ```enableTreatmentRoo``` is there justification for why you use enable here and set for others e.g. ```setMustHappen```

TheCycoONE:
Maybe ```setIsTreatmentRoom```, taking an optional bool for consistency? The rest of that commit looks good.

TheCycoONE:
For SpawnAction in d1ea6253af0efe8e620f2421d23973cd1f84b19c I think point is required but you include setPoint rather than putting it in the constructor. My understanding is that required fields go in the constructor (which would make it easier to add verification in the future).

TheCycoONE:
Visually reviewed up to 2d4aceaf511c015 inclusive.

TheCycoONE:
In 3ddeedee057 seek_reception.lua setMustHappen is set to true instead of action.must_happen

TheCycoONE:
684cc7e32 and 06e20d9de look ok to me in kdiff3

Alberth289346:
> The following commit uses enableTreatmentRoo is there justification for why you use enable here and set for others e.g. setMustHappen

The rule is: "setXYZ" takes a value, "enableXYZ"/"disableXYZ" does not. Before, I had some "setXYZ" with a value, and some "setXYZ" without value, which was too confusing, so I renamed the ones without value. Taking or not taking a value is mostly determined by whether the property is ever given at least more than a single value, although it possible I missed some case there.

> Maybe setIsTreatmentRoom, taking an optional bool for consistency?

I had that at some point, but then you have these property setters that only ever assign "true", which looks silly too (in particular with properties used in the negative sense "setNoWait(true)"-like calls).

> My understanding is that required fields go in the constructor

That is the intention indeed

Alberth289346:
Updated the branch:

- Added the setPoint() property into the SpawnAction constructor (spawn rewrite)
- Fixed the "action.must_happen" in seek_reception (queue rewrite)
- Fixed two typos found by luacheck diff (use_object rewrite)
- Fixed the doc problem (spawn rewrite)

TheCycoONE:
Right, I've seen other enable/disable flags now so that's alright. enableTreatmentRoom still sounds weird to me for a flag that advises that the target room is for treatment but I can live with it. The flag name predates this pr.

TheCycoONE:
Commit message typo in 5eefb3d bood to blood.

TheCycoONE:
6ebf067 results in incorrect indentation for queuing sink action. Maybe fixed in later commit?

TheCycoONE:
Visual review up to b7fbd264 exclusive. Have not gone back to changes yet.

Alberth289346:
Queueing sink action gets completely rewritten in 'use object rewrite' in commit 141517c, resolving the indent problem

TheCycoONE:
b7fbd2643ac29f1d8f443ef42d73267a693b3eb9 changes the function signature for the loop_callback function to take the action as an argument instead of a closure. A quick grep of the code shows that some actions call loop_callback passing in the action and some do not.

I'm not saying it's wrong, but it is suspicious.

TheCycoONE:
In the use_object commit, calling setWateringPlant with no argument sets it to nil not true. It is called with no argument in plant.lua.

Maybe should use enable style here?

Additionally lots of action properties were not converted to methods. Is than an oversight or justified?

TheCycoONE:
Also use_object. In dna_fixer ```setProlongedUsage``` is called with no argument. Should be called with ```false```

TheCycoONE:
In ward.lua setProlongedUsage is called without argument for true, but the function is not defined that way.

TheCycoONE:
For multi_use rewrite, again setProlongedUsage does not set to true with no argument. Called that way in place of true frequently. Not listing them here.

TheCycoONE:
In cardiogram.lua loop_callback function name does not match in diff

TheCycoONE:
Use_object and multi call setNoTruncate. Function is disableTruncate

TheCycoONE:
With respect to the loop_callback argument above, it looks alright in this case. Two caveats are that in meander the loop_callback is called without providing the action, and in idle_spawn, the callback is copied from another action so the argument would be different than the closure. It doesn't look like either of those can come up as a result of a use_object action.

TheCycoONE:
Sometimes a temporary is introduced to split lines and sometimes the next method in the chain is just placed indented on the next line. I prefer this method if we are not otherwise using the action variable. Can point them out inline if you like.

Alberth289346:
Fixes:
- walk rewrite: ": continuation lines"
- queue rewrite: ": continuation lines"
- split function from cardiogram... : fix cardiogram loop_callback function name
- split function from blood room... : fix commit message typo
- split function from toilets action...: Did not fix incorrect indent, it's not fatal, and gets removed further in the queue anyway
- use_object rewrite:
  * Use enableWatering as property instead of setWatering
  * ": continuation lines"
  * setProlongedUsage() argment fixes
  * use disableTruncate everywhere
- multi_use rewrite:
  * setProlongedUsage() argment fixes
  * use disableTruncate everywhere
  * add setLayer3 property method
- futher enhancements: Small action simplifications with action properties not noticed before

TheCycoONE:
Further enhancements isn't a descriptive commit message, but it looks like it can be squashed into *walk rewrite*

TheCycoONE:
Re-reviewed up to *seek staffroom rewrite*, all looks good so far.

TheCycoONE:
Reviewed up to *walk rewrite* exclusive. Please do the suggested rebase before I continue. sha-1s won't change up to f0456ce037 and I can then only look at 'final' commits.

Alberth289346:
merged the 'further enhancements' into the walk rewrite, thanks for all your efforts here

spinktvis:
Hi Albert, after trying the latest AppVeyor build i discovered that there is a double colon on line 56 of multi_use_object.lua. After removing it it is running fine. 

+function MultiUseObjectAction::setLayer3(layer3)

Alberth289346:
Clearly a case of writing too much C++ :)
Thanks for finding, fixed it.

TheCycoONE:
Typo in walk rewrite, scanner_room. *diableTruncate*

Alberth289346:
Fixed typo "diable" -> "disable", and reverted the accident (looks like git tried to be too smart in applying a patch at first sight)

TheCycoONE:
Reviewed all action rewrite commits. I need to go back and look at the split commits again.

Alberth289346:
Fixed the messed up loop callback, the original inline persistable function line was not replaced with the newly introduced loop_callback_training variable.

Diff between new head and the previous head is empty, ie no changes got introduced in the final rewrite result.

TheCycoONE:
All LGTM. Still going to test.

Alberth289346:
\o/

PR #1109, opened 2016-07-03
Fix wrong translation for epidemic compensation

ghandmann:
The translation was mentioning a fine while the original text (and what really happens in game) is a compensation.


TheCycoONE:
Thanks

Issue #1106, opened 2016-06-30
Reflexion/Idea3 not deactivated after placing second doctor in GP


Brainsample:
Repro:
- Find a GP office where a doctor is treating a patient.
- Pick up and place another doctor in the room.

Bug observed: The original doctor walks out the door with "Reflexion" or "Idea3" still active.
# 

Usually, the moods are deactivated in 

``` lua
function GPRoom:onHumanoidLeave(humanoid)
  -- Reset moods when either the patient or the doctor leaves the room.
  if self.staff_member then
    self.staff_member:setMood("idea3", "deactivate")
    self.staff_member:setMood("reflexion", "deactivate")
  end

```

But the doctor who is leaving is not the staffmember anymore, so it doesn't work.
(Unrelated: A leaving handyman can deactivate the mood too soon)
# 

Here's a possible fix, just deactivate everyone's mood for simplicity:

``` lua
function GPRoom:onHumanoidLeave(humanoid)
  -- Reset moods when either the patient or a doctor leaves the room.
  if humanoid.humanoid_class ~= "Handyman" then
    for staff, _ in pairs(self.humanoids) do
      staff:setMood("idea3", "deactivate")
      staff:setMood("reflexion", "deactivate")
    end
  end

```

CorsixTH v0.60

TheCycoONE:
Couldn't we use humanoid:setMood here instead of looping over the staff, since that is the humanoid that is leaving?

TheCycoONE:
Disregard the previous, I wasn't thinking about patients leaving.

PR #1110, opened 2016-07-05
Fix mood when staff leaves GP

TheCycoONE:
Thanks Brainsample

Copied from #1106

### DIFF
--- a/CorsixTH/Lua/rooms/gp.lua
+++ b/CorsixTH/Lua/rooms/gp.lua
@@ -186,9 +186,11 @@ end
 
 function GPRoom:onHumanoidLeave(humanoid)
   -- Reset moods when either the patient or the doctor leaves the room.
-  if self.staff_member then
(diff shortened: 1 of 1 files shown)


Brainsample:
Hi TheCycoONE, 

And thank you too!
Unfortunately the code in this pull request is not correct. :)

It uses:

```lua
      self.staff_member:setMood("idea3", "deactivate")
      self.staff_member:setMood("reflexion", "deactivate")
```

But that just deactivates the mood of the same entity we did before.

It should be:

```lua
      staff:setMood("idea3", "deactivate")
      staff:setMood("reflexion", "deactivate")
```

So everyone's idea3/reflexion gets deactivated.

TheCycoONE:
/me fails at copy and paste. Fixed.

Brainsample:
That's got it! :)
Thanks again.

PR #1118, opened 2016-08-12
[RDY] Action parameter checking

TheCycoONE:
The primary purpose of this pr is to add type checking to the action queue and enforce consistent types for given properties to make errors easier to trace.

During the process I also found a typo which is corrected in the first commit.

### DIFF
--- a/CorsixTH/Lua/humanoid_action.lua
+++ b/CorsixTH/Lua/humanoid_action.lua
@@ -27,6 +27,8 @@ local HumanoidAction = _G["HumanoidAction"]
 --! Construct a humanoid action (base class constructor).
 --!param name (str) Name of the action.
 function HumanoidAction:HumanoidAction(name)
+  assert(type(name) == "string", "Invalid value for parameter 'name'")
+
   self.name = name
   self.count = nil -- 'nil' means 'forever' (until finished), else the number to perform.
   self.must_happen = false -- If true, action cannot be skipped.
@@ -40,6 +42,8 @@ end
 --!param count (int or nil) Set to 'nil' if 'forever', else integer count.
 --!return (action) Returning self, for daisy-chaining.
 function HumanoidAction:setCount(count)
+  assert(count == nil or type(count) == "number", "Invalid value for parameter 'count'")
+
   self.count = count
   return self
 end
@@ -48,6 +52,8 @@ end
 --!param must_happen (bool) Whether or not the action must happen.
 --!return (action) Returning self, for daisy-chaining.
 function HumanoidAction:setMustHappen(must_happen)
+  assert(type(must_happen) == "boolean", "Invalid value for parameters 'must_happen'")
+
   self.must_happen = must_happen
   return self
 end
@@ -57,6 +63,9 @@ end
 --! termination conditions.
 --!return (action) Returning self, for daisy-chaining.
 function HumanoidAction:setLoopCallback(loop_callback)
+  assert(loop_callback == nil or type(loop_callback) == "function",
+      "Invalid value for parameter 'loop_callback'")
+
   self.loop_callback = loop_callback
   return self
 end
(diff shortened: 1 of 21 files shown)


CorsixTH/Lua/humanoid_actions/idle.lua
```diff
@@ -33,6 +33,11 @@ end
 --!param direction (string) Direction of facing.
 --!return (action) Self, for daisy-chaining.
 function IdleAction:setDirection(direction)
+  assert(direction == nil or
+      direction == "north" or direction == "south" or
+      direction == "east" or direction == "west",
+      "direction must be north, south, east, or west")
```
Alberth289346: Misses 'nil' as possible value

I am not so sure about the usefulness of the messages, as they seem to duplicate the code, but if you like them, why not.

CorsixTH/Lua/humanoid_actions/idle.lua
```diff
@@ -41,6 +46,9 @@ end
 --!param on_interrupt (function) Function to call on interrupt.
 --!return (action) Self, for daisy-chaining.
 function IdleAction:setOnInterrupt(on_interrupt)
+  assert(on_interrupt == nil or type(on_interrupt) == "function",
+      "on_interrupt not a function")
```
Alberth289346: missing nil as value

CorsixTH/Lua/humanoid_actions/knock_door.lua
```diff
@@ -27,6 +27,11 @@ local KnockDoorAction = _G["KnockDoorAction"]
 --!param door (Object) Door to knock on.
 --!param direction (string) Direction of facing.
 function KnockDoorAction:KnockDoorAction(door, direction)
+  assert(class.is(door, Door), "expecting door object")
+  assert(direction == "north" or direction == "south" or
```
Alberth289346: maybe refactor the direction test out into a separate function? it seems to happen a lot.

CorsixTH/Lua/humanoid_action.lua
```diff
@@ -65,6 +74,9 @@ end
 --!param after_use (func) Callback function that is called after the action ends.
 --!return (action) Returning self, for daisy-chaining.
 function HumanoidAction:setAfterUse(after_use)
+  assert(after_use == nil or type(after_use) == "function",
+      "after_use must be a function")
```
Alberth289346: misses nil in the error message

CorsixTH/Lua/humanoid_action.lua
```diff
@@ -57,6 +63,9 @@ end
 --! termination conditions.
 --!return (action) Returning self, for daisy-chaining.
 function HumanoidAction:setLoopCallback(loop_callback)
+  assert(loop_callback == nil or type(loop_callback) == "function",
+      "loop_callback must be a function")
```
Alberth289346: misses nil in the error message

TheCycoONE:
I found the message helpful to know the parameter, otherwise you just get the function and line number in the error message. The actual parameter restriction is unnecessary though, so maybe I'll just change them all to 'Invalid value for x!'

CorsixTH/Lua/humanoid_actions/knock_door.lua
TheCycoONE: I considered it but I don't know where I would put it.

Ideally I have in mind something like class.lua but for simple enums, so we'd write something like ```enum('Direction', { 'north', 'south', 'east', 'west' })``` then everywhere we would write ```enum.is(point.direction, Direction)```

If you don't mind I won't do that for this pr.

Alberth289346:
In C/C++ it also dumps the failed assert, I might be confused with that.

I agree a message is useful. I was mostly worried about the fact that you write the same thing both in code and in words. On the other hand, if you run into this thing in development, it's nice to  get a description of what is expected immediately without having to check those lines.

Both descriptions are fine to me, do what your :heart: tells you, or throw a :game_die: to decide.

TheCycoONE:
Travis failure appears to be due to a change in Busted, unrelated to this PR.

I've updated the PR to change the assert message to ```Invalid value for parameter 'foo'``` everywhere.
I also created a ```HumanoidAction:afterLoad``` function and update the actions to conform to the proper parameter types there instead of `humanoid.lua`

TheCycoONE:
Looks like Busted was fixed.

PR #1119, opened 2016-08-13
[rdy] Litter room fix

Alberth289346:
This should solve the room litter problem once and for all. It ignores litter as cause for non-buildability for rooms, clears litter from new rooms, and removes litter from rooms in old saves.

I am however not so happy with the buildable flag. It gets treated asif there is a single cause for the tile being non-buildable, but we have several possible causes. In case of litter, there is definitely the option to have more than 1 litter object at a tile, which means setting it to true unconditionally is just wrong.

Maybe the flag should have a cause , or it should be derivable from the tile contents (which would be major +1). Ideas?

Edit: https://github.com/CorsixTH/CorsixTH/pull/381#issuecomment-68917824  says we should never have more than one litter on a tile.

_Edit: Ignore the first 2 commits, they are fixes for a more useful work environment._

### DIFF
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -64,6 +64,8 @@ function Room:initRoom(x, y, w, h, door, door2)
   self.objects = {--[[a set rather than a list]]}
   -- the set of humanoids walking to this room
   self.humanoids_enroute = {--[[a set rather than a list]]}
+
+  self.world:removeAllLitterFromRectangle(self.x, self.y, self.width, self.height)
 end
 
 --! Get the tile next to the door.
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1956,7 +1956,7 @@ function World:canNonSideObjectBeSpawnedAt(x, y, objects_id, orientation, spawn_
   for _, tile in ipairs(objects_footprint) do
     local tiles_world_x = x + tile[1]
     local tiles_world_y = y + tile[2]
-    if self:areFootprintTilesCoardinatesInvalid(tiles_world_x, tiles_world_y) then
+    if not self:isOnMap(tiles_world_x, tiles_world_y) then
       return false
     end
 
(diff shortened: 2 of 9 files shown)


CorsixTH/Lua/room.lua
```diff
@@ -947,9 +948,15 @@ end
 
 --! Stub to be extended in subclasses, if needed.
 function Room:afterLoad(old, new)
-  if old and old < 46 then
+  if not old then old = -1 end
```
TheCycoONE: Does this ever happen? I don't think it's caught other places.

CorsixTH/Lua/world.lua
```diff
@@ -2253,14 +2253,35 @@ function World:getObject(x, y, id)
...
+--!param x (int) Start x position of the area.
+--!param y (int) Start y position of the area.
+--!param w (int) Number of tiles in x direction.
+--!param h (int) Number of tiles in y direction.
+function World:removeAllLitterFromArea(x, y, w, h)
+  for xpos = x, x + w - 1 do
+    for ypos = y, y + h - 1 do
+      -- Loop repeatedly over the objects, as there may be more than one litter object in a tile.
+      while true do
+        local litter = self:getObject(xpos, ypos, "litter")
```
TheCycoONE: More efficient to use self:getObjects and loop through them here. Might be useful to extend getObjects to take type filter there instead (then simplify the getObject function).

Not critical.

TheCycoONE:
I don't understand why litter would affect the buildable flag at all, it shouldn't get in the way of building. I think that's a historical accident that you can simplify right out. Still need to fix the existing ones for the save though.

Alberth289346:
Resolved all litter problems I know about

- Litter is stackable now
- Handyman tasks don't remove the wrong task any more (causing the non-cleanable vomits)
- Cleanable (ie non-exploded litter) litter stays buildable, rooms and objects can be built on it, and litter will be removed.

- Improved finding the correct hospital on the map.

First 2 patches need removed, and sane new game version number needs to be assigned

CorsixTH/Lua/world.lua
```diff
@@ -2264,13 +2263,24 @@ function World:removeLitter(obj, x, y)
...
 function World:getRoom(x, y)
   return self.rooms[self.map:getRoomId(x, y)]
 end
 
+--! Get the hospital at a given tile location.
+--!param x (int) X position of the queried tile.
+--!param y (int) Y position of the queried tile.
+--!return (Hospital) Hospital at the given location or 'nil'.
+function World:getHospital(x, y)
+  for _, hosp in ipairs(self.hospitals) do
```
TheCycoONE: This is a little roundabout, hosp:isInHospital calls a C++ function that tells it what hospital is at the current hospital and returns true if that hospital matches the hospital number. Because there are very few hospitals it doesn't matter, but this could be O(1) instead of O(N)

TheCycoONE:
I really like this patch set. Reviewed all of it. :+1:

Alberth289346:
Removed the debug patches from the start of the queue, and improved the getHospital function by using plots, which is indeed much faster, thanks for the suggestion.

Alberth289346:
Even faster getHospital function :)

PR #1125, opened 2016-08-27
Remove litter from non-passable tiles too

Alberth289346:
Reception desk did not remove litter


TheCycoONE:
I know, it looked funny but didn't seem to do any damage because the handymen only had to walk beside it. This is better.

PR #1130, opened 2016-09-17
[messedup] Lua check adding and fixing

Alberth289346:
Besides my 2 convenience fixes, this is a first test in how far we jump
I think I got most things that needed fixing

To consider
- [x] I copied the lucheckrc file from bob, which I didn't thoroughly review
- [x] Near the top of the patch queue it gets messy, that may need cleaning, or even re-shuffling all patches per file or something else that is easy to check
- [x] I skipped "unused argument Foo", which needs disabling, I think
- [x] There are some other remaining messages too, part of the submission is to see which ones, and it needs a decision what to do with them

### DIFF
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -721,7 +721,7 @@ function UIBottomPanel:addDialog(dialog_class, extra_function)
         self.ui:setEditRoom(false)
         local dialog = _G[dialog_class](self.ui)
         if extra_function then
-          _G[dialog_class][extra_function](dialog)
+          _G[dialog_class][extra_function](d)
         end
         self.ui:addWindow(dialog)
         self:updateButtonStates()
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -721,7 +721,7 @@ function UIBottomPanel:addDialog(dialog_class, extra_function)
         self.ui:setEditRoom(false)
         local dialog = _G[dialog_class](self.ui)
         if extra_function then
-          _G[dialog_class][extra_function](d)
+          _G[dialog_class][extra_function](dialog)
         end
         self.ui:addWindow(dialog)
         self:updateButtonStates()
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -199,7 +199,7 @@ function Patient:completeDiagnosticStep(room)
     local divisor = math.random(1, 3)
     local attn_detail = room.staff_member.profile.attention_to_detail / divisor
     local skill = room.staff_member.profile.skill / divisor
-    local diagnosis_bonus = (attn_detail + 0.4) *  skill
+    diagnosis_bonus = (attn_detail + 0.4) * skill
   end
   self:modifyDiagnosisProgress(diagnosis_base + (diagnosis_bonus * multiplier))
 end
@@ -483,7 +483,7 @@ function Patient:pee()
       self:finishAction()
     end
     self:setMood("poo", "deactivate")
-    self:changeAttribute("happiness", -0.02)  -- not being able to find a loo and doing it in the corridor will make you sad too
+    self:changeAttribute("happiness", -0.02) -- not being able to find a loo and doing it in the corridor will make you sad too
     if not self.hospital.did_it_on_floor then
       self.hospital.did_it_on_floor = true
       self.world.ui.adviser:say(_A.warnings.people_did_it_on_the_floor)
@@ -631,12 +631,12 @@ function Patient:tickDay()
   -- patient has been in the hospital for over 6 months and is still not well, so will become sad and will either get fed up and leave
   -- or stay in the hope that you will cure them before they die
   -- strange, but in TH happiness does not go down, even when close to death IMO that is wrong as you would be unhappy if you waited too long.
-  -- TODO death animation for slack female is missing its head.  For now the only option is for her to get fed up and leave
+  -- TODO death animation for slack female is missing its head. For now the only option is for her to get fed up and leave
   -- this can be changed when the animation thing is resolved
   -- TODO clean up this block, nonmagical numbers
   if self.attributes["health"] >= 0.18 and self.attributes["health"] < 0.22 then
     self:setMood("sad2", "activate")
-    self:changeAttribute("happiness", -0.0002)   -- waiting too long will make you sad
+    self:changeAttribute("happiness", -0.0002) -- waiting too long will make you sad
     -- There is a 1/3 chance that the patient will get fed up and leave
     -- note, this is potentially run 10 ((0.22-0.18)/0.004) times, hence the 1/30 chance.
     if math.random(1,30) == 1 then
@@ -703,8 +703,8 @@ function Patient:tickDay()
   if self.vomit_anim and not self:getRoom() and not self.action_queue[1].is_leaving and not self.action_queue[1].is_entering then
     --Nausea level is based on health then proximity to vomit is used as a multiplier.
     --Only a patient with a health value of less than 0.8 can be the initial vomiter, however :)
-    local initialVomitMult = 0.002   --The initial chance of vomiting.
-    local proximityVomitMult = 1.5  --The multiplier used when in proximity to vomit.
+    local initialVomitMult = 0.002 --The initial chance of vomiting.
+    local proximityVomitMult = 1.5 --The multiplier used when in proximity to vomit.
     local nausea = (1.0 - self.attributes["health"]) * initialVomitMult
     local foundVomit = {}
     local numVomit = 0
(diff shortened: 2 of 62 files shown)


TheCycoONE:
Eventually I'd like the commit to add luacheck to be at the end, so that they all pass travis in between.

CorsixTH/Lua/filesystem.lua
```diff
@@ -18,7 +18,8 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
-local LFS = require "lfs"
+require "lfs"
```
TheCycoONE: Similarly, typically in lua if nothing is done with the require it's not necessary. (It's possible for the require to define globals, but that's considered bad practice.)

CorsixTH/Lua/game_ui.lua
```diff
@@ -27,10 +27,8 @@ class "GameUI" (UI)
 local GameUI = _G["GameUI"]
 
 local TH = require "TH"
-local SDL = require "sdl"
-local WM = SDL.wm
-local lfs = require "lfs"
-local pathsep = package.config:sub(1, 1)
+require "sdl"
```
TheCycoONE: Try without them

CorsixTH/Lua/graphics.lua
```diff
@@ -19,7 +19,8 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
 local TH = require "TH"
-local SDL = require "sdl"
+require "sdl"
```
TheCycoONE: broken record

CorsixTH/Lua/room.lua
```diff
@@ -18,7 +18,7 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
-local TH = require "TH"
+require "TH"
```
TheCycoONE: ""

CorsixTH/Lua/sprite_viewer.lua
```diff
@@ -22,8 +22,9 @@ SOFTWARE. --]]
 -- when you don't have a copy of AnimView, or the means to compile it, then a
 -- crude sprite viewer is better than no sprite viewer.
 
+require"lfs"
```
TheCycoONE: ""

CorsixTH/Lua/class.lua
```diff
@@ -65,8 +65,8 @@ local function define_class(name, super, adopts_self)
   local methods = {}
   local methods_mt = {}
 
-  local function new_class(methods, ...)
-    local constructor = methods[name]
+  local function new_class(methods2, ...)
```
TheCycoONE: Not a fan of calling it methods2, as that implies it's different from methods. It's not, but call on a metatable passes the actual table as the first argument. ```t``` is a fairly conventional name for that first argument. ```m``` would be fine as well in our case.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1209,11 +1209,11 @@ function UIEditRoom:setDoorBlueprint(x, y, wall)
     flags = flags + 16 -- Use red palette rather than normal palette
   end
   if self.room_type.swing_doors then
-    for _, anim in ipairs(anim) do
-      anim:setAnimation(self.anims, 126, flags)
+    for _, anim2 in ipairs(anim) do
```
TheCycoONE: This function is huge. Suggest breaking it up, which might avoid using silly variable names.

CorsixTH/Lua/entities/patient.lua
```diff
@@ -199,7 +199,7 @@ function Patient:completeDiagnosticStep(room)
     local divisor = math.random(1, 3)
     local attn_detail = room.staff_member.profile.attention_to_detail / divisor
     local skill = room.staff_member.profile.skill / divisor
-    local diagnosis_bonus = (attn_detail + 0.4) *  skill
+    diagnosis_bonus = (attn_detail + 0.4) *  skill
```
TheCycoONE: Two spaces between * and skill. I know you didn't put them there, but since you touched the line anyway, might as well fix it.

CorsixTH/Lua/entity_map.lua
```diff
@@ -45,8 +45,8 @@ end
 @param y (integer) the y coordinate of the entity
 @param entity (Entity) the entity to be added to the map in the (x,y) position]]
 function EntityMap:addEntity(x,y,entity)
-  local function add_entity_to_table(entity,entity_table)
-    entity_table[#entity_table+1] = entity
+  local function add_entity_to_table(enty, entity_table)
```
TheCycoONE: Promote the local function to the top level. It doesn't use the closure and that avoids renaming the variable.

CorsixTH/Lua/entity_map.lua
```diff
@@ -65,12 +65,12 @@ the assertion that the entity is contained in the map to fail.
 @param x (integer) the x coordinate of the entity
 @param y (integer) the y coordinate of the entity
 @param entity (Entity) the entity to be removed from the map in the (x,y) position ]]
-function EntityMap:removeEntity(x,y,entity)
+function EntityMap:removeEntity(x, y, entity)
   -- Iterates through a table and removes the given entity
-  local function remove_entity_from_table(entity, entity_table)
+  local function remove_entity_from_table(enty, entity_table)
```
TheCycoONE: Promote this local function to top level, same reasons as the previous one.

CorsixTH/Lua/epidemic.lua
```diff
@@ -599,11 +599,11 @@ end
 function Epidemic:getBestVaccinationTile(nurse, patient)
   -- General usage tile finder, used in the other cases
   -- when the patient isn't sitting on a bench
-  local function get_best_usage_no_bench(nurse,patient)
+  local function get_best_usage_no_bench(nurse2, patient2)
```
TheCycoONE: Possibly promote the function instead. It does use the closure for self, so the entity map would have to be passed in another way.

CorsixTH/Lua/game_ui.lua
```diff
@@ -459,25 +459,25 @@ function GameUI:onMouseMove(x, y, dx, dy)
       (x < scroll_region_size or y < scroll_region_size or
        x >= self.app.config.width - scroll_region_size or
        y >= self.app.config.height - scroll_region_size) then
-    local dx = 0
-    local dy = 0
+    local dx2 = 0
```
TheCycoONE: May have noticed I'm not a fan of *2 variables. Here dx, dy are change in the x / y location of the mouse between calls of ```onMouseMove``` dx2 and dy2 are the scroll distances. Silly things are done with ```tick_scroll_amount_mouse``` otherwise I'd suggest dropping the temp variable entirely. Suggesting ```tsam_x``` and ```tsam_y``` instead.

CorsixTH/Lua/graphics.lua
```diff
@@ -359,8 +359,8 @@ function Graphics:loadLanguageFont(name, sprite_table, ...)
       -- TODO: Choose face based on "name" rather than always using same face.
       font:setFace(self.ttf_font_data)
       font:setSheet(sprite_table)
-      local function reloader(font)
-        font:clearCache()
+      local function reloader(fnt)
```
TheCycoONE: Could move this function to top level and rename it ```font_reloader```

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -721,7 +721,7 @@ function UIBottomPanel:addDialog(dialog_class, extra_function)
         self.ui:setEditRoom(false)
         local dialog = _G[dialog_class](self.ui)
         if extra_function then
-          _G[dialog_class][extra_function](d)
+          _G[dialog_class][extra_function](dialog)
```
TheCycoONE: Incidentally the only place this is used is to call ```UIBankManager.showStatistics``` https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/dialogs/fullscreen/bank_manager.lua#L318

The argument to that function is a bool for whether to replace the cursor.

TheCycoONE: dialog is right. With this notation self needs to be explicit.

Alberth289346:
Loooong patch queue, but it addresses all points raised by lua check.
Many thanks to bobcanthelpyou for the initial start of the patch in #972 

Things are split by file, and while it's a lot of patches, they seem quite readable.

CorsixTH/Lua/entities/patient.lua
```diff
@@ -199,7 +199,7 @@ function Patient:completeDiagnosticStep(room)
     local divisor = math.random(1, 3)
     local attn_detail = room.staff_member.profile.attention_to_detail / divisor
     local skill = room.staff_member.profile.skill / divisor
-    local diagnosis_bonus = (attn_detail + 0.4) *  skill
+    diagnosis_bonus = (attn_detail + 0.4) * skill
```
Alberth289346: This is a bugfix

TheCycoONE:
For future consideration splitting by file isn't very helpful, it is pretty easy to see the file boundaries in a commit, and it makes navigation here more cumbersome.

Alberth289346:
This isn't working

TheCycoONE:
I like your tags

Alberth289346:
Merging this in smaller steps instead

PR #1131, opened 2016-09-24
[rdy] Rewrite functions

Alberth289346:
Taking some of the bigger rewrites out of the lua check patch

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -1071,6 +1071,21 @@ local window_floor_blueprint_markers = {
   west = 36,
 }
 
+--! Check whether the given tile can function as a door entry/exit tile.
+--!param xpos (int) X position of the tile.
+--!param ypos (int) Y position of the tile.
+--!param player_id (int) Player id owning the hospital.
+--!param flag_names (array) If set, array with two additional required properties.
+--!return Whether the tile is considered to be valid.
+function UIEditRoom:validDoorTile(xpos, ypos, player_id, flag_names)
+  local th = self.ui.app.map.th
+
+  local tile_flags = th:getCellFlags(xpos, ypos)
+  if not (tile_flags.buildable or tile_flags.passable or tile_flags.owner == player_id) then return false end
+  if not flag_names then return true end
+  return tile_flags[flag_names[1]] and tile_flags[flag_names[2]]
+end
+
 function UIEditRoom:setDoorBlueprint(x, y, wall)
   local orig_x = x
   local orig_y = y
@@ -1181,7 +1196,6 @@ function UIEditRoom:setDoorBlueprint(x, y, wall)
   end
   if self.blueprint_door.valid then
     -- Ensure that the door isn't being built on top of an object
-    local flags = {}
     local flag_names
     if wall == "west" then
       flag_names = {"buildableNorth", "buildableSouth"}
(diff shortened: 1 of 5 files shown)


CorsixTH/Lua/dialogs/patient.lua
```diff
@@ -158,12 +163,8 @@ function UIPatient:draw(canvas, x_, y_)
...
   end
   -- How warm the patient feels
-  local warmth_bar_width = 22
-  local warmth = patient.attributes["warmth"]
-  if warmth then
-    warmth = self.normaliseWarmth(warmth)
-    warmth_bar_width = math_floor(warmth * 40 + 0.5)
-  end
+  local warmth = self.normaliseWarmth(patient.attributes["warmth"])
+  local warmth_bar_width = math_floor(warmth * 40 + 0.5)
```
TheCycoONE: Not identical in the nil case, but I assume the old value was made up anyway.

CorsixTH/Lua/dialogs/patient.lua
```diff
@@ -114,6 +114,20 @@ function UIPatient.normaliseWarmth(warmth)
...
+--! Draw a bar in the patient dialogue window.
+--!param canvas Canvas to draw at.
+--!param xbase Horizontal base position.
+--!param ybase Vertical base position.
+--!param xpos Horizontal offset.
+--!param ypos Vertical offset.
+--!param value Fraction to draw.
+function UIPatient:drawBar(canvas, xbase, ybase, xpos, ypos, value)
+  local width = math.floor(value * 40 + 0.5)
+  for dx = 0, width - 1 do
```
TheCycoONE: I think you still need to handle the 0 case here, otherwise math.floor(0 * 40 + 0.5) == 0, then you subtract 1, so you would be saying from 0 to -1 step 1, which seems like a bad thing to do.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1257,10 +1257,10 @@ function UIEditRoom:placeWindowBlueprint()
   end
 end
 
-function UIEditRoom:setWindowBlueprint(x, y, wall)
-  local orig_x = x
-  local orig_y = y
-  local orig_wall = wall
+function UIEditRoom:setWindowBlueprint(orig_x, orig_y, orig_wall)
+  local x = orig_x
```
TheCycoONE: And yet they get written to on Line 1286 anyway. But that's ok, closer to const parameters.

CorsixTH/Lua/dialogs/patient.lua
TheCycoONE: Disregard, lua handles this.

CorsixTH/Lua/dialogs/patient.lua
Alberth289346: Another comment said a bar would be half-filled if it has an unknown value. I think 20 is close enough to 22 to make that hold.

CorsixTH/Lua/dialogs/patient.lua
Alberth289346: Indeed, tested this too

Alberth289346:
No idea how to retrigger the build, it failed on getting cth-deps from GH.

TheCycoONE:
I get an error trying to run this branch:
```
Error loading dialogs\edit_room.lua:
...\CorsixTH\Lua\dialogs\edit_room.lua:1476: 'end' expected (to close 'function' at line 1120) near <eof>
```

Alberth289346:
fixed, running a quick edit room test caught two other errors in room editing

TheCycoONE:
Plays well now

PR #1133, opened 2016-09-30
[RDY] Improve command for running luac in travis

TheCycoONE:
Switches from a loop in a subshell to using xargs, improving the find command to be more efficient and handling the possibility of spaces in file names. The command is also now agnostic to the current directory. exit logic is unnecessary now.


TheCycoONE:
I had really hoped to do more with this PR, but luac is rather limited (and quiet).

PR #1138, opened 2016-10-09
[rfc] Query custom graphics with the animation number as string.

Alberth289346:
This enables simple replacement of original animations by custom
graphics.

### DIFF
--- a/CorsixTH/Src/th_gfx.cpp
+++ b/CorsixTH/Src/th_gfx.cpp
@@ -704,10 +704,33 @@ size_t THAnimationManager::getFrameCount() const
 
 size_t THAnimationManager::getFirstFrame(size_t iAnimation) const
 {
+    size_t result = getCustomFirstFrame(iAnimation);
+    if (result > 0)
+        return result;
+
+    return getOriginalFirstFrame(iAnimation);
+}
+
+size_t THAnimationManager::getOriginalFirstFrame(size_t iAnimation) const
+{
     if(iAnimation < m_iAnimationCount)
         return m_vFirstFrames[iAnimation];
     else
         return 0;
+}
+
+size_t THAnimationManager::getCustomFirstFrame(size_t iAnimation) const
+{
+    if (iAnimation > 99999) // Number too large to fit, requested animation doesn't exist.
+        return 0;
+
+    std::string name = std::to_string(iAnimation);
+    const AnimationStartFrames &anim_frames = getNamedAnimations(name, 64);
+    if (anim_frames.iNorth > 0) return getOriginalFirstFrame(anim_frames.iNorth);
+    if (anim_frames.iEast > 0)  return getOriginalFirstFrame(anim_frames.iEast);
+    if (anim_frames.iSouth > 0) return getOriginalFirstFrame(anim_frames.iSouth);
+    if (anim_frames.iWest > 0)  return getOriginalFirstFrame(anim_frames.iWest);
+    return 0;
 }
 
 size_t THAnimationManager::getNextFrame(size_t iFrame) const
--- a/CorsixTH/Src/th_gfx.h
+++ b/CorsixTH/Src/th_gfx.h
@@ -297,6 +297,10 @@ public:
     size_t getFrameCount() const;
 
     //! Get the index of the first frame of an animation
+    /**!
+        @param iAnimation Animation to obtain.
+        @return The first frame of the requested animation if it exists, else \c 0.
+     */
     size_t getFirstFrame(size_t iAnimation) const;
 
     //! Get the index of the frame after a given frame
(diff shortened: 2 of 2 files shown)


CorsixTH/Src/th_gfx.cpp
```diff
@@ -701,12 +701,48 @@ size_t THAnimationManager::getFrameCount() const
...
         return 0;
 }
 
+size_t THAnimationManager::getCustomFirstFrame(size_t iAnimation) const
+{
+    if (iAnimation > 99999) // Number too large to fit, requested animation doesn't exist.
+        return 0;
+
+    // Convert number to text.
+    char buffer[8];
```
TheCycoONE: Use ```std::to_string``` http://en.cppreference.com/w/cpp/string/basic_string/to_string

Alberth289346:
Thanks for the review.

The code is fixable, but after making the above patch, I am having doubts that this is the right solution. Sorry for not explaining earlier, but only after talking with @sadger yesterday, the problem became clear enough to be able to express it in words.

The patch will work for the simple stuff I think. Plants, bins, walls, etc. Gui stuff should work too. (Basically, because we probably never want anything else there, I realize now.)

What I am having doubts about are the animations with people in it. Current code is completely tailored for the graphics that we now have. Woman walking north is animation XYZ, with layer flags for head, cloths. Woman walking west is also animation XYZ with some mirror flag and the same cloth/head flags.
So if we replace animation XYZ with new graphics, we're going to be stuck with that one animation and a set of flags that is being used for woman walking west and north with some layer flags for cloths and head. You cannot add a new animation for walking north only, or for the woman with the long hair or.. etc.

Is this something we want or should want?

Alberth289346:
Reading this again after 4 years makes the path more clear.

### Idea

I think what we want is
a) Compatibility with the original graphics, and
b) freedom to construct new graphics as we see fit for current animations.
[c) freedom to add new animations, but that seems quite far away]

To make this happen, an animation should be self-supporting, it should not need any information about the animation to play from the program other than some position on the map.. That means all layer, mirror, and transparency flags should be included in the animation itself.

The game should only need to say "this animation please, at this spot in the map". A name or a unique number of the animation should be enough.
(Note that frame-number as unique identity is likely to fail unless you decouple from the actual frame number, animations with new graphics may be ordered differently, and/or may be missing some animations.)

One way to achieve this is to add a "setupdata" block to each animation, which contains all the layer/mirror/transparency/etc information, as well as a link to the animation frames, colour palette, and what else you need. For the original graphics setupdata blocks cannot be added, so the loader would need to create those blocks (or load them from file?!), and attach the original animations to them.

Likely the setupdata should be extended to the point that we can show all patients, at least the radiation patient is missing, possibly others as well??

Such a setupdata block could have different content for new graphics, allowing a designer to change the internal structure of the animation, eg not share or layer the animation.

In the code you will get more animations, since flags etc are now encoded into different animations. In theory that can be solved by creating a table with flag settings to the animation-name to use.

### Supporting change in graphics mid-game

While experimenting with new graphics, Sadger at one point loaded an old savegame, while having a new plant animation available.
The result was that all existing plants in the hospital were using the old (original) graphics, while newly bought plants used the new graphics.

The reason is that the save system saves and loads the running animations as well as all its settings in the various objects. A running animation doesn't query the animation frames to use from the animation collection on load.

While this makes for great fun, I would say this is undesired. When loading new graphics, the user should see new graphics when loading a game if we support that.
This means that the save system should neither save running animations with its full details nor should it save any data with respect to which animations are currently available (things like the first frame number or number of frames etc). On a next load that data may be invalid, leading to anything from weird graphics to crashes.

Instead, it should query the name of the running animations while saving, as wel ass how far it progressed, so on the next load, a different animation can be loaded and positioned at a compatible point in time.
This implies an animation needs a notion of progress that works between different graphics sets.

### New graphics

The above setupdata block introduction means we need a list of all available animations, with their settings. Likely we want to at least attach a name to make it understandable what animation we're talking about, perhaps even with a one-two line description.

Such information would be great to provide to designers so they have a clue what to create.

For graphics set compatibility, we may also want to state required number of frames (or length in time perhaps) as well as size, people involved, relative start positions, and/or end-positions of objects or people, etc.

Maybe the current system already solves that in some way?

sadger:
Weirdly I found myself drawn back to the project to try and look at the missing patient rendering. I think it's both "serious radiation" and "jellyitus". Managed to find a few youtube videos of the diseases in progress on the original for inspiration.  My idea was to try set a flag on the affected patients then do some rendering/pallete swap of each frame of the animations based on the flag. Of course we don't need custom animations for this but it would work as a generic solution.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -703,13 +703,36 @@ size_t THAnimationManager::getFrameCount() const
...
 {
     if(iAnimation < m_iAnimationCount)
         return m_vFirstFrames[iAnimation];
     else
         return 0;
 }
 
+size_t THAnimationManager::getCustomFirstFrame(size_t iAnimation) const
+{
+    if (iAnimation > 99999) // Number too large to fit, requested animation doesn't exist.
```
sadger, line 724: Too large to fit where? Could be a constant perhaps?

Alberth289346: The number is (or rather, was) converted to a string in a fixed size char buffer (so all animations have a name rather than a number), so this is an overflow protection. As this is used only for the original animations, the upper limit is known.
TheCycoONE didn't like that though and wanted a std::string for the buffer and fixed it I think. I have no problems with std::string fixing or with introducing a constant as you suggested, but as this patch isn't anywhere near covering the problem it's trying to address, I didn't fix it since I consider this a failed attempt (but a very successful one for learning about the true problem).

Alberth289346:
Nice experiment, but this solution is not going to work

Issue #1128, opened 2016-08-30
Level 10 should not have earthquakes


willsv:
In almost every level I've had most of my rooms destroyed by huge earthquakes. It seems that they are too frequent and too strong.

Also, I've read that in Eggsenham (level 10) on the original TH there were no earthquakes but I had the same problem.

Best.

TheCycoONE:
Possibly related to #550 ?

TheCycoONE:
I believe you are right about level 10. If you don't mind I'd like to make this issue focus on that, rather than the perception of being overly violent (possibly related to #550)

TheCycoONE:
Related to #296

willsv:
What he is referring to is to the screen movement, which is not quite the same as I was saying.
When I say "strong" i mean in the way it affects the machines and how it wipes out half my rooms, and not just screen movement.

TheCycoONE:
Okay. Addressing #296 by removing random occurrence earthquakes would make the frequency match the original game and fix the level 10 problem.

The actual strength is difficult to determine exactly - my understanding is that it isn't consistently too strong. In the original game earthquakes were a big problem - once you got an earthquake warning you had to actually click through your machines and make sure they were near max strength or they would get wiped out in later levels. It is possible that our upper range is too high. I still want to make this about the easier issue, then address balance in a new issue afterwards if necessary after.

willsv:
Makes sense, thanks!

Tristesseract:
Hi,
Eggsenham Level 10 should not have earthquakes. I ran 25 years of game time on the gog version to confirm.

Thanks

TheCycoONE:
I believe this is fixed already by #1139

Issue #550, opened 2014-12-24
Earthquakes are more violent at slower game speeds


TheCycoONE:
_From [dugu...@gmail.com](https://code.google.com/u/115875571364606670451/) on May 03, 2013 09:55:42_

What steps will reproduce the problem? 1. Put game speed into slow speed
2. Wait for earthquake
3. See whole screen shake around more violently than at faster game speeds. What is the expected output? What do you see instead? I expect the screen to move around only a few pixels, as it does at higher game speeds. Instead, the slower the game speed is, the more violent the screen movement is.

What version of CorsixTH are you using (e.g. "Beta 2", "SVN r123 ")?
Version 0.20 What operating system / compile settings are you using? Windows 7 64-bit, DirectX 9

What level was this on (e.g. "Demo level", "Full game level 12")?
Full game Level 12 Attach a saved game file (preferably an autosave from before the error) if you have one and it is relevant. Please provide any additional information below. Save is a few days before an earthquake (starts on 20th Sept). Saved at high speed, so change the speed down before this date to see the difference.

Due to a piece of litter blocking an exit, this also causes a crash around the 24th Sept. I've posted this as separate bug 1608

**Attachment:** [debug.sav](http://code.google.com/p/corsix-th/issues/detail?id=1609)

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1609_

tobylane:
Still happens. In this autosave from March there's an earthquake on the 8th. http://cl.ly/0m0h212t403f

TheCycoONE:
Fixed

Issue #296, opened 2014-08-15
User and campaign maps should not have earthquakes unless they opt in


TheCycoONE:
At the moment it is hard coded that there can be no earthquakes in the first 4 levels of the campaign. However this does not help with custom maps as leaving earthquake control blank still generates random earthquakes.

When a level contains only the instruction `#quake_control[0].StartMonth.EndMonth.Severity 0 0 0` no random earthquakes should be added.

TheCycoONE:
I think this behavior from the original (no random earthquakes) would be a good fix.

MarkL1961:
Or a toggle that allows you to turn off random earthquakes.  As I said in the other issue this then gives the player the choice of not having any in a map they create or for that to be ignored by turning on randoms.
It should be a simple enough fix, as I recall when I attempted a solution it was the random that caused me the problems when trying to make it accept 0 severity means none and remove the hard coded bit for levels 1 - 5 or something like that.

TheCycoONE:
Alright, I think it's fine to add a new level flag to enable/disable random earthquakes, defaulting to disabled.

Stefan1200-de:
I support this, also looking for a way to disable earthquakes in custom levels (not playing campaign). Is that already possible?

Alberth289346:
the issue is still open, so no

TheCycoONE:
The current earthquake trigger is wrong. I'm in favor of earthquakes being purely opt-in, and only at the dates defined in the level file. If we want random quakes indefinitely we can add a new level file line for that.

MarkL1961:
I can't see this mentioned here, but it may have been said elsewhere. In TH you always had a warning (tremor or slight screen jolt) to tell you a quake was on the way.  This was when most of us would have slowed the game down and checked all our machines were up to it.  CorsixTH is lacking this tremor.   

Edvin and I did look at this a few years back, but the current code is complex and Edvin was busy at that time.  Was thinking that perhaps a month prior to the specified quake the screen moves quickly and you get the warning announcement etc, but no damage - then get the quake.

In favour of the other suggestions as well

TheCycoONE:
Thanks @MarkL That warning was critical on later earthquakes.

TheCycoONE:
I've run several tests on the original with earthquakes added to the first level.

Unlike ours it seems nothing is random in the original:
Damage is the severity of the earthquake. We do severity plus or minus 1.
[...]

PR #1139, opened 2016-10-11
[RDY] Earthquakes

TheCycoONE:
This pull request addresses is a substantial reworking of Earthquakes, though it doesn't include a planned refactor of Earthquake related code into a new class.

It addresses issues: #1128 #550 and #296 and adds the warning quakes from the original game which cued the player that they need to repair any damaged machines which should solve a lot of users issues with earthquakes.

### DIFF
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -32,6 +32,10 @@ local WM = SDL.wm
 local lfs = require "lfs"
 local pathsep = package.config:sub(1, 1)
 
+-- The maximum distance to shake the screen from the origin during an
+-- earthquake with full intensity.
+local shake_screen_max_movement = 50 --pixels
+
 --! Game UI constructor.
 --!param app (Application) Application object.
 --!param local_hospital Hospital to display
@@ -80,6 +84,11 @@ function GameUI:GameUI(app, local_hospital, map_editor)
   self.current_momentum = {x = 0.0, y = 0.0, z = 0.0}
 
   self.speed_up_key_pressed = false
+
+  -- The currently specified intensity value for earthquakes. To abstract
+  -- the effect from the implementation this value is a number between 0
+  -- and 1.
+  self.shake_screen_intensity = 0
 end
 
 function GameUI:setupGlobalKeyHandlers()
@@ -151,12 +160,16 @@ function GameUI:draw(canvas)
     canvas:fillBlack()
   end
   local zoom = self.zoom_factor
+  local dx = self.screen_offset_x +
+      math.floor((0.5 - math.random()) * self.shake_screen_intensity * shake_screen_max_movement * 2)
+  local dy = self.screen_offset_y +
+      math.floor((0.5 - math.random()) * self.shake_screen_intensity * shake_screen_max_movement * 2)
   if canvas:scale(zoom) then
-    app.map:draw(canvas, self.screen_offset_x, self.screen_offset_y, math.floor(config.width / zoom), math.floor(config.height / zoom), 0, 0)
+    app.map:draw(canvas, dx, dy, math.floor(config.width / zoom), math.floor(config.height / zoom), 0, 0)
     canvas:scale(1)
   else
     self:setZoom(1)
-    app.map:draw(canvas, self.screen_offset_x, self.screen_offset_y, config.width, config.height, 0, 0)
+    app.map:draw(canvas, dx, dy, config.width, config.height, 0, 0)
   end
   Window.draw(self, canvas, 0, 0) -- NB: not calling UI.draw on purpose
   self:drawTooltip(canvas)
@@ -749,6 +762,18 @@ function GameUI:scrollMap(dx, dy)
   self.screen_offset_y = floor(dy + 0.5)
 end
 
+--! Start shaking the screen, e.g. an earthquake effect
+--!param intensity (number) The magnitude of the effect, between 0 for no
+-- movement to 1 for significant shaking.
+function GameUI:beginShakeScreen(intensity)
+  self.shake_screen_intensity = intensity
+end
+
+--! Stop the screen from shaking after beginShakeScreen is called.
+function GameUI:endShakeScreen()
+  self.shake_screen_intensity = 0
+end
+
 function GameUI:limitCamera(mode)
   self.limit_to_visible_diamond = mode
   self:scrollMap(0, 0)
@@ -1045,6 +1070,9 @@ function GameUI:afterLoad(old, new)
     self:removeKeyHandler("x", self, self.toggleWallsTransparent)
     self:addKeyHandler("z", self, self.keySpeedUp)
     self:addKeyHandler("x", self, self.keyTransparent)
+  end
+  if old < 115 then
+    self.shake_screen_intensity = 0
   end
   return UI.afterLoad(self, old, new)
 end
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -51,6 +51,9 @@ local local_criteria_variable = {
   {name = "population",       icon = 11, formats = 1},
 }
 
+-- time between each damage caused by an earthquake
+local earthquake_damage_time = 16 -- hours
+
 function World:World(app)
   self.map = app.map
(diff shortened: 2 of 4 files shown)


CorsixTH/Lua/world.lua
```diff
@@ -1370,6 +1390,7 @@ function World:nextEarthquake()
     self.next_earthquake.size = control.Severity
     self.next_earthquake.remaining_damage = self.next_earthquake.size
     self.next_earthquake.damage_timer = earthquake_damage_time
+    self.next_earthquake.warning_timer = earthquake_warning_period
```
Alberth289346: Do these new values have to be added to the new table in the afterLoad function too?

CorsixTH/Lua/game_ui.lua
```diff
@@ -151,12 +153,16 @@ function GameUI:draw(canvas)
     canvas:fillBlack()
   end
   local zoom = self.zoom_factor
+  local dx = self.screen_offset_x +
+      math.floor((0.5 - math.random()) * self.shake_screen_intensity * 100)
+  local dy = self.screen_offset_y +
+      math.floor((0.5 - math.random()) * self.shake_screen_intensity * 100)
```
Alberth289346: Instead of magic 100, why not make the shake_screen_intensity have tile units, and do the translation from intensity to tiles in the beginShakeScreen?

CorsixTH/Lua/game_ui.lua
```diff
@@ -80,6 +80,8 @@ function GameUI:GameUI(app, local_hospital, map_editor)
   self.current_momentum = {x = 0.0, y = 0.0, z = 0.0}
 
   self.speed_up_key_pressed = false
+
+  self.shake_screen_intensity = 0
```
Alberth289346: Please state meaning of the variable as well

CorsixTH/Lua/world.lua
```diff
@@ -569,32 +563,22 @@ end
...
 
     -- set up the next earthquake date
     self:nextEarthquake()
   else
     -- Multiplier for how much the screen moves around during the quake.
     local multi = 4
-    if (self.day > self.next_earthquake.stop_day) or (self.day < math.round(self.next_earthquake.size / 3)) then
-      -- if we are in the first two days of the earthquake, make it smaller
+    if self.next_earthquake.remaining_damage > 2 or
+        self.next_earthquake.size - self.next_earthquake.remaining_damage > 2 then
```
Alberth289346: It looks to me that you're computing when "big" shifts should happen

CorsixTH/Lua/world.lua
```diff
@@ -612,14 +612,14 @@ function World:tickEarthquake()
   else
     -- Multiplier for how much the screen moves around during the quake.
     local multi = 4
-    if (self.day > self.earthquake_stop_day) or (self.day < math.round(self.earthquake_size / 3)) then
+    if (self.day > self.next_earthquake.stop_day) or (self.day < math.round(self.next_earthquake.size / 3)) then
       -- if we are in the first two days of the earthquake, make it smaller
-      self.randomX = math.random(-(self.earthquake_size/2)*multi, (self.earthquake_size/2)*multi)
-      self.randomY = math.random(-(self.earthquake_size/2)*multi, (self.earthquake_size/2)*multi)
+      self.randomX = math.random(-(self.next_earthquake.size / 2) * multi, (self.next_earthquake.size / 2) * multi)
```
TheCycoONE: This commit just moves variables into the next_earthquake structure. A commit further in the series replaces this whole block.

CorsixTH/Lua/world.lua
TheCycoONE: Yes, suggesting I should restore the comment I cut here, or add a new one?

CorsixTH/Lua/world.lua
TheCycoONE: I set it to 0 in afterLoad because any save game with an active earthquake the actual earthquake is already in progress (warning period is over).

CorsixTH/Lua/world.lua
Alberth289346: Oh, new comment seems fine if the condition also matches that reality :)

TheCycoONE:
Updated:
More comments have been added throughout, including documentation for all newly added variables.
The commit that moved stop_day calculations has been dropped since it was entirely removed in the next commit anyway.
The magic number used in the shake size calculation has been moved into a local variable and commented.

TheCycoONE:
I actually did have the size calculation logic wrong. Must have been tired, anyway fixed now.

CorsixTH/Lua/world.lua
```diff
@@ -1129,15 +1112,11 @@ function World:onEndDay()
   end
 
   -- check if it's time for an earthquake, and the user is at least on level 5
-  if (self.year - 1) * 12 + self.month == self.next_earthquake_month and
-      self.day == self.next_earthquake_day then
+  if self.next_earthquake and self.next_earthquake.size and
+      (self.year - 1) * 12 + self.month == self.next_earthquake.start_month and
```
sgtbigman: Probably not worth doing for this PR, but `(self.year - 1) * 12 + self.month` almost seems worth putting in a function at some point.

TheCycoONE:
Thanks @sgtbigman. I will fix that comment. In an earlier version of this patch I addressed the date as well but now I am leaving it for a refactor purely focused on date handling.

CorsixTH/Lua/world.lua
```diff
@@ -931,7 +915,7 @@ function World:setSpeed(speed)
   local pause_state_changed = nil
   if speed == "Pause" then
     -- stop screen shaking if there was an earthquake in progress
-    if self.active_earthquake then
+    if self.next_earthquake.active then
       self.ui.tick_scroll_amount = {x = 0, y = 0}
```
TheCycoONE: This needs to be updated for new screenShake methods.

CorsixTH/Lua/world.lua
```diff
@@ -74,6 +74,16 @@ function World:World(app)
...
 
+  -- All information relating to the next or current earthquake, nil if
+  -- there is no scheduled earthquake.
+  -- Contains the following fields:
+  -- active (boolean) Whether the earthquake is currently happening.
+  -- start_month (integer) The month the earthquake is triggered.
+  -- start_day (integer) The day of the month the earthquake is triggered.
+  -- stop_day (integer) the day of the month the earthquake ends.
+  -- size (integer) The amount of damage the earthquake causes (1-9).
+  self.next_earthquake = nil
```
Alberth289346: Would it make sense to initialize to {active=false} ? then you can always query the boolean instead of "if world.next_earthquake and world.next_earthquake.active then .."

CorsixTH/Lua/world.lua
```diff
@@ -544,14 +554,14 @@ end
 
 function World:createEarthquake()
```
Alberth289346: Assuming you figured out what this function does, it would be helpful to document it

CorsixTH/Lua/world.lua
```diff
@@ -544,14 +554,14 @@ end
...
   -- Sanity check
-  if not self.earthquake_size then
+  if not self.next_earthquake or not self.next_earthquake.size then
```
Alberth289346: I'd expect a test for being active, and not for having a size

CorsixTH/Lua/world.lua
TheCycoONE: Agreed to always set ```{ active = false }``` but it's not enough to test that because an earthquake scheduled for the future is also ```active = false``` (but with a start_month and start_day).

CorsixTH/Lua/world.lua
TheCycoONE: I kill this function later in the pull request, so imo it's not worth documenting.

CorsixTH/Lua/world.lua
TheCycoONE: Can't test active here, active is expected to be false. start_day and start_month should be nil or 0 if the earthquake shouldn't happen though so the size test is unnecessary.

TheCycoONE:
Addressed multiple no-earthquake states, consolidating on ```{ active = false }```

PR #1140, opened 2016-10-13
[RDY] Allow resizing the main window

ecrips:
This change makes the window resizeable which allows e.g. simply maximizing the window if you don't want to go into full screen mode. I've (lightly) tested this on Debian, I'd be interested to see if others think this is a good idea.

### DIFF
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -126,6 +126,27 @@ function GameUI:makeVisibleDiamond(scr_w, scr_h)
   }
 end
 
+--! Calculate the minimum valid zoom value
+--!
+--! Zooming out too much would cause negative width/height to be returned from
+--! makeVisibleDiamond. This function calculates the minimum zoom_factor that
+--! would be allowed.
+function GameUI:calculateMinimumZoom()
+  local scr_w = self.app.config.width
+  local scr_h = self.app.config.height
+  local map_h = self.app.map.height
+
+  -- Minimum width:  0 = 32 * map_h - (scr_h/factor) - (scr_w/factor) / 2,
+  -- Minimum height: 0 = 16 * map_h - (scr_h/factor) / 2 - (scr_w/factor) / 4
+  -- Both rearrange to:
+  local factor = (scr_w + 2 * scr_h) / (64 * map_h)
+
+  -- Due to precision issues a tolerance is needed otherwise setZoom might fail
+  factor = factor + 0.001
+
+  return factor
+end
+
 function GameUI:setZoom(factor)
   if factor <= 0 then
     return false
@@ -179,6 +200,11 @@ function GameUI:draw(canvas)
 end
 
 function GameUI:onChangeResolution()
+  -- Calculate and enforce minimum zoom
+  local minimum_zoom = self:calculateMinimumZoom()
+  if self.zoom_factor < minimum_zoom then
+    self:setZoom(minimum_zoom)
+  end
   -- Recalculate scrolling bounds
   local scr_w = self.app.config.width
   local scr_h = self.app.config.height
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -57,6 +57,7 @@ function App:App()
     mousewheel = self.onMouseWheel,
     motion = self.onMouseMove,
     active = self.onWindowActive,
+    window_resize = self.onWindowResize,
     music_over = self.onMusicOver,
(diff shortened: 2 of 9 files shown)


TheCycoONE:
Sounds like a good enhancement to me. I'll have to play test it later to see how it feels in practice.

We are striving for better coding practice than we had in the past, so please add comments to the top of the new functions using LDocGen syntax (https://github.com/CorsixTH/CorsixTH/wiki/Coding-Conventions#ldocgen-syntax)

Otherwise on the surface the actual code changes look good to me.

TheCycoONE:
Edge scrolling and resizable windows do not mix super well, but I don't see the harm in adding it anyway.

This patch doesn't trigger the movies to resize but it should be fairly easy to add that. Since around 0.50 the movie is sized as it's drawn in refresh(), so the target size could be passed there instead of play.

TheCycoONE:
What was changed in your last rebase?

ecrips:
I added documentation comments to the two new functions added, and I realised that I'd unnecessarily duplicated the `changeResolution` function in `game_ui.lua` so removed that.

TheCycoONE:
@ecrips are you interested in making resizing during movies work a bit better?

ecrips:
I've made a new commit which resizes the video dynamically per frame.

TheCycoONE:
@ecrips That's what I was thinking yes.
It's working well for movies. An issue I've found play testing - if I zoom all the way out, and then stretch the window bigger, the screen glitches and I can no longer zoom in or move around the map.

ecrips:
I've added another commit which calculates and enforces a minimum zoom on resolution change. Hopefully that will prevent the view being too zoomed out when resizing.

CorsixTH/Lua/game_ui.lua
```diff
@@ -166,6 +184,12 @@ function GameUI:draw(canvas)
 end
 
 function GameUI:onChangeResolution()
+  -- Calculate and enforce minimum zoom
+  local minimum_zoom = self:calculateMinimumZoom()
+  print("Minimum zoom:", minimum_zoom, " current:",self.zoom_factor)
```
TheCycoONE: Works well, please remove this print.

CorsixTH/Lua/game_ui.lua
```diff
@@ -117,6 +117,24 @@ function GameUI:makeVisibleDiamond(scr_w, scr_h)
   }
 end
 
+--! Calculate the minimum valid zoom value
+--!
+--! Zooming out too much would cause negative width/height to be returned from
+--! makeVisibleDiamond. This function calculates the minimum zoom_factor that
+--! would be allowed.
+function GameUI:calculateMinimumZoom()
```
TheCycoONE: world:adjustZoom has to deal with minimum zoom levels too, perhaps the logic can be consolidated.

ecrips: From what I can tell `adjustZoom` relies on `setZoom` failing to limit zooming out (it does provide the maximum zoom as a 'virtual width' of 200). `setZoom` fails when `makeVisibleDiamond` returns a negative width or height. The logic in `calculateMinimumZoom` simply working out when the width/height would be 0.

Unfortunately having testing this a bit more I've found that it is still possible to trigger the behaviour where scrolling/zooming doesn't work in some cases. I believe this is due to floating point precision issues, so I've added a small tolerance in `onChangeResolution` to hopefully avoid this (I haven't been able to reproduce the problem with that in). Let me know if you've got a better idea for how to solve that!

PS. Thanks for spotting the `print` I left in for debugging in the previous commit

TheCycoONE: Add your tolerance directly to the MinimumZoom returned by the function, since it's that value that is off, not the usage of it.

I myself don't see a way around it. Be sure to document the magic number, and please include a space around your operator.

ecrips: I've moved the tolerance to the `calculateMinimumZoom` function as you suggested

TheCycoONE:
Looks good, rebase into one commit and it should be good to merge.
Thanks!

ecrips:
I've squashed the commits together and rebased onto the latest master

wolfy1339:
Just a note, you can squash commits and rebase them on the latest master when you merge pull requests from the GitHib web interface.
Also, @ecrips change the RFC to RDY to mark this as ready to merge

TheCycoONE:
@wolfy1339 I'm aware, but when people are capable of using git properly like ecrips I'd rather they did it themselves. Keeps everyone in sync. :-)

PR #1141, opened 2016-10-15
[WIP] Rename Room's 'room_info' member variable to 'data'.

sgtbigman:
room_info holds room-based data, and since it is already attached to a
Room object, the 'room_' part of the variable is redundant.

- This refactor helps simplify code snippets from `room.room_info` to
`room.data`.

*Note*: I need to test this more on my end; an old save seems to be
throwing errors about `room.data` being `nil` in the `hasRoomOfType`
function called in the Hospital `tick` function.

Also, if this is not worth pursuing, feel free to let me know.

Thanks.

### DIFF
--- a/CorsixTH/Lua/calls_dispatcher.lua
+++ b/CorsixTH/Lua/calls_dispatcher.lua
@@ -59,7 +59,7 @@ function CallsDispatcher:callForStaff(room)
       self:callForStaffEachRoom(room, attribute, attribute .. i)
     end
   end
-  local sound = room.room_info.call_sound
+  local sound = room.data.call_sound
   if anyone_missed and sound and not room.sound_played then
     room.world.ui:playAnnouncement(sound)
     room.sound_played = true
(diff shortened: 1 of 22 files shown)


Alberth289346:
I can see that "room_info" isn't optimal, but at least it somewhat says what it contains. Your proposed "data" is a bit too generic I think, as it can contain anything by that name. Maybe it should be some name with a "static", "fixed", or "properties" or so, notion.

As for old saves dying, yep, they do if you rename a field. "Save" means literally store the current state, as in "save all reachable Lua tables". "Load" means then "re-instate the Lua tables from the file". Function names are saved for all class methods and all --[[persistable:....]] magic functions.
Thus if you change a field in the code, and you load an old save, the changed code is trying to find the new name, which fails, since the loaded data still contains the old field name.
The current solution to this problem is to increment the savegame version in app.lua, and add an 'if' to the afterLoad method(s) to fix the loaded data.

TheCycoONE:
My feeling is that the name data is not worth pursuing because it has no semantic meaning. I understand what is stored in this table is room type information.

sgtbigman:
Thanks to you both for the feedback. I think the repeat `room` bothered me, but I definitely see its value over something generic like `data`.

I will look into more important issues/features to address in the codebase/project.

Thanks again.

PR #1148, opened 2016-10-31
[rdy] Fix code and variables in various dialogs.

Alberth289346:


CorsixTH/Lua/dialogs/queue_dialog.lua
```diff
@@ -360,15 +358,16 @@ function UIQueuePopup:UIQueuePopup(ui, x, y, patient)
...
-  local function send_to_hospital(i)
-    return --[[persistable:queue_dialog_popup_hospital_button]] function()
-      -- TODO: Actually send to another hospital (when they exist)
-      self.patient:goHome("kicked")
-      local str = _S.dynamic_info.patient.actions.sent_to_other_hospital
-      self.patient:updateDynamicInfo(str)
-      self:close()
-    end
-  end
+  -- TODO: Add this when there are other hospitals to send patients to.
```
TheCycoONE: If this function isn't called, drop it. It probably won't actually be done this way later anyway.

Alberth289346:
updated

PR #1151, opened 2016-11-02
Fix Lua code and variables of remaining windows, and a few entities.

Alberth289346:
The repair_loop_callback change in entities/machine.lua is a no-op, as
it was already false due to the "setProlongedUsage(false)" used at the
repair_action variable.


TheCycoONE:
I think it is OK. Will test later

PR #1152, opened 2016-11-04
Improve entities, graphics, filesystem, epidemic, and hospital

Alberth289346:
Also two bug fixes detected by Luacheck.

### DIFF
--- a/CorsixTH/Lua/entity_map.lua
+++ b/CorsixTH/Lua/entity_map.lua
@@ -26,7 +26,7 @@ local EntityMap = _G["EntityMap"]
 --[[ An entity map is a structure is a 2 dimensional structure created from a
 game map, it has the same dimensions as the game map which intitalises it.
 The purpose of the map is store the location of entities in the map in
-real-time. Each cell given by an (x,y) coordinate is a
+real-time. Each cell given by an (x, y) coordinate is a
 table {humanoids ={}, objects = {}} where the tables of humanoids/objects
 may be empty or contain the entity/entities that currently exist in that tile.]]
 function EntityMap:EntityMap(map)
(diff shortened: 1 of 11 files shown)


Alberth289346:
Fixed them now, and one extra for bonus. I don't usually keep notes, as I forget them anyway.
Moved the other 'sheet' variable instead, which looks like it should work too.

CorsixTH/Lua/entity_map.lua
```diff
@@ -59,28 +60,28 @@ function EntityMap:addEntity(x,y,entity)
...
+  for i, h in ipairs(entity_table) do
+    if h == entity then
+      -- If we have found it no need to keep looking
+      index = i
+      break
+    end
+  end
+  -- We haven't found it so we have incorrectly tried to remove it
+  assert(index ~= -1, "Tried to remove entity from entity_map - entity not found")
+  table.remove(entity_table,index)
```
sgtbigman: Minor (missing space): `table.remove(entity_table,index)` -> `table.remove(entity_table, index)`?

CorsixTH/Lua/epidemic.lua
```diff
@@ -614,20 +596,23 @@ end
...
       local direction = object_in_use.direction
       if direction == "north" then
         return px, py - 1
+
       elseif direction == "south" then
         return px, py + 1
+
       elseif direction == "east" then
         return px + 1, py
+
```
sgtbigman: Any reason for the newlines here?

CorsixTH/Lua/hospital.lua
```diff
@@ -1460,24 +1460,24 @@ end
 function Hospital:addToEpidemic(patient)
   --[[ See if there exists a future epidemic with the same
   disease as the contagious patient - if so add them to it
-  @param patient (Patient) the patient to add to the epidemic
+  @param pat (Patient) the patient to add to the epidemic
   @return true if patient was successfully added to a future epidemic
   false otherwise (boolean) ]]
-  local function add_to_existing_future_epidemic(patient)
+  local function add_to_existing_future_epidemic(pat)
```
sgtbigman: Is there a reason for this change?

CorsixTH/Lua/hospital.lua
```diff
@@ -1460,24 +1460,24 @@ end
...
     return false
   end
 
   --[[ Create a new epidemic and add it to the pool of future epidemics
-  @param patient (Patient) the patient who will be the first
+  @param pat (Patient) the patient who will be the first
   contagious patient of the new epidemic]]
-  local function add_new_epidemic_to_pool(patient)
-    local new_epidemic = Epidemic(self,patient)
+  local function add_new_epidemic_to_pool(pat)
```
sgtbigman: Same as above.

CorsixTH/Lua/graphics.lua
```diff
@@ -199,8 +199,8 @@ function Graphics:makeGreyscaleGhost(pal)
     local g_index = 0
     local g_diff = 100000 -- greater than 3*63^2 (TH uses 6 bit colour channels)
     for j = 0, #entries do
-      local entry = entries[j]
-      local diff = (entry[1] - g)^2 + (entry[2] - g)^2  + (entry[3] - g)^2
+      local entry_j = entries[j]
```
sgtbigman: Almost seems like this could be `possible_nearest_entry` or just `nearest_entry`. IMO, if the `j` variable ever changes, this will lose its meaning.

CorsixTH/Lua/graphics.lua
```diff
@@ -258,11 +258,12 @@ function Graphics:loadRaw(name, width, height, dir, paldir, pal)
   end
   bitmap:setPalette(palette)
   assert(bitmap:load(data, width, self.target))
-  local function reloader(bitmap)
-    bitmap:setPalette(palette)
-    local data = self.app:readDataFile(dir, name .. ".dat")
-    data = data:sub(1, width * height)
-    assert(bitmap:load(data, width, self.target))
+
+  local function reloader(bm)
```
sgtbigman: Since you renamed the other (below) `reloader` to `font_reloader`, perhaps this could become `bitmap_reloader`?

CorsixTH/Lua/hospital.lua
TheCycoONE: To not mask the patient argument from the parent function. Though the parameter is redundant: patient is an upvalue here so pat is not needed at all.

CorsixTH/Lua/hospital.lua
TheCycoONE: Ditto

CorsixTH/Lua/graphics.lua
Alberth289346: There is a lot more wrong here (and elsewhere too), my primary aim is to fix all errors reported by luacheck rather than really improving the local code.

EDIT: I rewrote it, but the longer I look at it, the less sense it makes. The magic numbers to convert to grey scale look like the regular conversion values, but corsix uses 6 bit colour values, as it says in the comment, so shouldn't that get compensated somewhere in the greyscale conversion, and then back to the 6 bit value in the diff calculation?

CorsixTH/Lua/graphics.lua
Alberth289346: Sounds good, I changed a few more :p

CorsixTH/Lua/hospital.lua
Alberth289346: It's indeed about fixing the masking, we don't yet agree on the redundancy property :)

Alberth289346:
Further improvements

TheCycoONE:
GitHub isn't sending me any sort of notice when you bump this, so please comment each time.

Issue #799, opened 2015-07-12
Can't create room door cause of drinking machine but the other way round it is possible


ghost:
It is possible to place a drinking machine near a door like this:

![bildschirmfoto von 2015-07-12 21 58 51](https://cloud.githubusercontent.com/assets/4523264/8639748/1e533872-28e3-11e5-8977-0704a4b5add8.png)

But it is not possible to create a door at the same place when the drinking machine is like this:

![bildschirmfoto von 2015-07-12 22 01 02](https://cloud.githubusercontent.com/assets/4523264/8639756/3c0319e6-28e3-11e5-875b-142b87097279.png)

expected:
- both should be possible or
- both shouldn't be possible

TheCycoONE:
I believe this is the same missing check that allows building doors that are blocked by radiators.  Both shouldn't be possible.

bobcanthelpyou:
In this case it could by a missing check by placing the drinking machine as i can't build a door near the drinking machine front?

TheCycoONE:
Agreed.

TheCycoONE:
Duplicate of #221

Issue #221, opened 2014-03-14
Should not be possible to place objects in front of doors


iMiKED:
Hello!

What steps will reproduce the problem?
1. Load game
2. Accept incoming emergency patients with bloaty heads
3. Get numerous errors (Empty action queue)

What is the expected output? What do you see instead?
I expect patients to go to Bloaty Head clinic

What version of CorsixTH are you using (e.g. "0.30", "4c66e39985")?
Version 0.30 - released November 2013

What operating system / compile settings are you using?
WinXP SP3 Pro

What level was this on (e.g. "Demo level", "Full game level 12")?
Full game level 10

Saves and gamelog are available here:
https://mega.co.nz/#F!hs9mRZgY!aqHZDrfLGohXc_GhAnLYRA
 -- or here --
https://drive.google.com/folderview?id=0B5wa6TO0a4ZHc2l4SC1LTmNCa0U

TheCycoONE:
I'm also not able to put a radiator in front of a door, but somehow @Frassle managed to (in the bathroom) as evidenced by the save in that ticket.

@Frassle, was that game started in 0.40 or did you originally start it in an earlier version?

Frassle:
Game was started in 0.40, was the first version of Corsix I've used. Using the downloaded installer not built from scratch.

MarkL1961:
@TheCycoONE   Try placing a radiator in a position where if you were to build a room you would place the door.  Now build the room around the radiator and whilst in blue print mode you can place the door next to the radiator.  Now you have a room with a radiator blocking the door

TheCycoONE:
Ahh, thanks @MarkL1961 so it's the door placement that needs to be modified rather than object placement.

TheCycoONE:
Also I just checked the original, and in it you couldn't place a radiator facing any direction on the tiles adjacent to doors.  In CorsixTH you can as long as the radiator isn't on the door wall, which I feel could lead to other bugs.  Suggesting we use the same restrictions as the original when it comes to doors (parcels must be completely clear on each side).

MarkL1961:
I think there is still an issue with the blue double doors and that it is still possible to place a radiator in front and over lapping them.  Can't check it again at the moment, but it was still possible the last time I checked.
Also bench placement can cause a problem when next to a door - think that is in another issue somewhere

TheCycoONE:
[...]

Issue #826, opened 2015-07-24
[META] Support for blocking off areas of the hospital


TheCycoONE:
This is a meta issue for supporting blocked off areas of the hospital, as in the original Theme Hospital. Any bug affecting blocked off areas should reference this issue. Once all of the critical bugs preventing this feature are implemented, then we can remove the build restrictions.

_Until a debug flag is implemented, to test a blocked off area, build plants or drink machines against the wall separating a plot that hasn't been purchased yet, then purchase that plot._
- [x] Implement debug flag to allow placing objects that would block off a portion of the map
- [ ] Fix bugs that arise because a part of the map is blocked off
- [ ] Remove the debug flag and default to allowing blocked maps

TheCycoONE:
#819

lewri:
#2021

lewri:
#1963

lewri:
Just adding my thoughts here, that blocking off areas shouldn't be used inside rooms as it can break the room.

lewri:
#2072

lewri:
#2211

Alberth289346:
Reporting more glitches isn't really needed. We know they exist in half a zillion variations, and more if you start looking for them,
Cyco looked at the problem already and established that the current way of painting rooms to the screen is simply broken beyond repair. The current solution is never going to fix all render bugs.

As such, it's not going to be a small local fix with a testcase like you made to verify it's indeed fixed. Instead, the entire paint code must be taken out and replaced by something better. Trouble is, we don't yet know what that "something" should be. I think we also established that likely the current paint code doesn't get enough information to solve the problem, so that needs some thinking too.

This is basically the point where we are currently wrt this problem. No easy solution available, I am afraid.

ARGAMX:
I suggest to close every blocking off areas sub-issues (as we do that for Graphical Glitches sub-issues):

1. Each of them cannot be solved individually.
2. If the objects are already placed in an invalid way, then this situation can no longer be resolved. A mechanism is needed that simply will prevent such placement.

An example of such mechanism is described here:

Theme Hospital `Safe Blocked Off Areas` Algorithm:

1. When placing corridor items or new room these checks should be executed:
1.1. This placing not making any already exited room door path-disconnected from the map Entrance. (room_door==world) [“==” - is mean connected]
1.2. This placing does not make any humanoid path-disconnected from the map Entrance. (humanoid==world)

2. When building room also these checks should be executed:
[...]

PR #1153, opened 2016-11-04
Door and Object placement corrections Issues #221, #799

mugmuggy:
Fixes here hopefully resolves the  #221 & #799 issues. I also corrected a number of inconsistencies with the original TH. Like restricting objects to the room when building a room. Persisting the footprint checks across all tiles, rather than each individual tile. Restricting other objects to the wall boundaries of the hospital (I can't recall if litter can be dropped outside, and if this will have an issue).
Thats just a couple lines in place_object.lua
+        -- Check if tile is in the hospital building footprint
+        is_object_allowed = is_object_allowed and flags["hospital"]

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -1105,7 +1105,7 @@ end
 --!param xpos (int) X position of the tile.
 --!param ypos (int) Y position of the tile.
 --!param player_id (int) Player id owning the hospital.
---!param flag_names (array) If set, array with four additional required properties.
+--!param flag_names (array) If set, array with arbitrary additional required properties.
 --!return Whether the tile is considered to be valid.
 local function validDoorTile(xpos, ypos, player_id, flag_names)
   local th = TheApp.map.th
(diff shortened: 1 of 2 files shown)


TheCycoONE:
Thanks for looking into this. Litter can only be dropped inside the hospital outside of any rooms, which is already handled. Checking the hospital flag is not normally enough - ownerid should be tested as well.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1105,15 +1105,19 @@ end
...
+--!param flag_names (array) If set, array with arbitrary additional required properties.
 --!return Whether the tile is considered to be valid.
 local function validDoorTile(xpos, ypos, player_id, flag_names)
   local th = TheApp.map.th
 
   local tile_flags = th:getCellFlags(xpos, ypos)
   if not (tile_flags.buildable or tile_flags.passable or tile_flags.owner == player_id) then return false end
   if not flag_names then return true end
-  return tile_flags[flag_names[1]] and tile_flags[flag_names[2]] and tile_flags[flag_names[3]] and tile_flags[flag_names[4]]
+  local accumulate_flags = true
```
TheCycoONE: You don't need this, if there is any false you can exit early: ```if not tile_flags[flagtype] return false end``` and ```return true``` at the end.

TheCycoONE:
What's the reason you make an exclusion for the invisible tiles in the blueprint?

mugmuggy:
Check 4 for example, Bench has invisible tile, if you place a bench 90 degrees to a tile adjacent to a door, without this check, it prevents you from placing the object, however in TH it does allow you to.

mugmuggy:
Check 5 in  place_objects needs some tweaking though. I'll double check on that one as it prevents placing side objects in front of a bench. Doesn't copy TH faithfully.

TheCycoONE:
Invisible was introduced in 506a03c73 and only intended to affect the appearance of the blueprint. If you want to modify when a tile is acceptable modify the footprint / one of the other flags.

That aside, there are some discrepancies with the original game due to the path finding requirement in CorsixTH as well, but that can't currently be removed without causing crashes: see https://github.com/CorsixTH/CorsixTH/issues/826

I suggest keeping your first PR as small as possible in scope to resolve the door issues and leave other placement issues for your next patch.

mugmuggy:
No probs I'll close this one and raise a new one

PR #1154, opened 2016-11-06
Doors should be passable

### LINKED ISSUE #221: Should not be possible to place objects in front of doors

mugmuggy:
Door logic updated to better reflect issues raised in and preparing to fix #221, #799. This commit doesn't prevent objects being placed in the path of a door, just that a door cannot be placed with an object in the way.

There are some changes to displaying the individual tile on a swing door that is blocked, either 1, 2 or all 3 on swing doors will show as red if blocked.

Swing doors were not correctly being placed into corners, I've replicated TH - which might have been a bug originally, that it has to be two tiles down on the north wall to the east, not 1 like the others.
East (right corner) and South walls (left corner) were not enforcing swing doors out of the corners

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -1105,7 +1105,7 @@ end
 --!param xpos (int) X position of the tile.
 --!param ypos (int) Y position of the tile.
 --!param player_id (int) Player id owning the hospital.
---!param flag_names (array) If set, array with four additional required properties.
+--!param flag_names (array) If set, array with arbitrary additional required properties.
 --!return Whether the tile is considered to be valid.
 local function validDoorTile(xpos, ypos, player_id, flag_names)
   local th = TheApp.map.th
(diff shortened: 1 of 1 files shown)


mugmuggy:
Reopened this for discussion on the changes proposed, with additional commits after more testing. Most changes are for aligning CorsixTH with original TH behaviours mainly with swinging doors. Blocking object detection I feel might be a bit rough (invisible of bench doesn't block doors, but passable of drink machines does). Door positioning on walls could have a bug in original TH on north walls. Added the floor blueprint for swing doors, and corrected the single doors to match TH.

Alberth289346:
Hi, you seem to have trouble with git.

Things work a lot better if you make a new branch at master, before you start adding your changes:
```` sh
git checkout master   # Switch to master
git branch mybranch # Make a new branch
git checkout mybranch # Switch to the new branch
````
In this way, you're not so badly affected when master moves.
In general, NEVER commit anything directly to master, always use a separate branch.

When you want to copy changes in master at the CorsixTH repository into your master branch:
```` sh
git checkout master # Switch to your master
git fetch upstream # Pull new changes in (assumes 'upstream' points to the central CorsixTH repository)
git merge --ff-only upstream/master  # Merge new changes from upstream into your master
````
The '--ff-only' flag ensures, your master will stay a proper mirror of the CorsixTH master, so you keep a clean base version.

When you want to incorporate new local master changes into your work, do a rebase:
```` sh
git checkout mybranch # Switch to your branch
git rebase master  # Move your patch onto the new master.
# This may fail if there are modifications close to each other. In that case, you need to fix them
````

I am not sure why you make new PRs and close them after a while by yourself.
If you prefix the PR title with "[WIP]" we won't touch it. If you prefix with "[RFC]" it means you're asking for feedback (which we'll be happy to provide), finally, prefix "[RDY]" means 'ready for merging', that is, you think everything is ready.
You can update the changes of a PR by pushing the same 'mybranch' again. Github will notice it, and update the PR.

Last but not least, if you want more interaction, please join the #corsixth IRC channel, we may not always be there, but be patient, eventually someone will join.

PR #1161, opened 2016-11-16
[wip] Lots of files with few simple improvements

Alberth289346:


CorsixTH/Lua/humanoid_actions/walk.lua
```diff
@@ -112,7 +112,7 @@ action_walk_interrupt = permanent"action_walk_interrupt"( function(action, human
 end)
 
 local flag_list_bottom = 2048
-local flag_early_list = 1024
+--  flag_early_list = 1024
```
TheCycoONE: Why leave this here?

CorsixTH/Lua/map.lua
```diff
@@ -180,8 +180,9 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
   local result = file(path .. "Lua" .. pathsep .. "base_config.lua")
 
   local base_config = result
-  local errors
+  local _
```
TheCycoONE: Let's not require declaring (or avoiding declaring) ```_```. It is a near universal convention (mentioned in PIL and enforced by several IDEs) in lua to use _ as a dummy variable so the chance of conflicting with anything else is very slim, which makes this a micro-optimization.

CorsixTH/Lua/objects/plant.lua
```diff
@@ -73,9 +73,6 @@ object.orientations = {
 -- For plant: droop down: 1950
 -- back up again: 1952
 
--- The states specify which frame to show
-local states = {"healthy", "drooping1", "drooping2", "dying", "dead"}
```
TheCycoONE: This is currently the only documentation for ```current_state```, best to replace it with a comment indicating what each of the current_state values mean.

CorsixTH/Lua/humanoid_actions/walk.lua
Alberth289346: Documentation purposes, mostly. I don't know another place where those bit numbers and names/explanation are related with each other.
Note that I also don't know if the complicated way of building the screen is still used/useful; using OpenGL blitting opens a lot of other ways of doing things, like the Z-buffer.

CorsixTH/Lua/objects/plant.lua
Alberth289346: Fair point, replaced this by a line of comment

Alberth289346:
Pushed updated branch (plant state list modified as described)

CorsixTH/Lua/map.lua
TheCycoONE: I see, luacheck allows shadowing _, but not referencing it as a global. Fine then, we'll follow their lead.

CorsixTH/Lua/humanoid_actions/walk.lua
TheCycoONE: It's still used.

Alberth289346: Removed the line, it's also at objects/door.lua, so the link doesn't get lost

Alberth289346:
Removed the "--  flag_early_list = 1024" line entirely from walk.lua, as another file still has it.

Alberth289346:
Added list with plant state numbers

CorsixTH/Lua/objects/plant.lua
```diff
@@ -70,11 +70,13 @@ object.orientations = {
 -- For litter: put broom back 356
 -- take broom out: 1874
 -- swoop: 1878
--- For plant: droop down: 1950
--- back up again: 1952
+-- Frames for plant states are
```
TheCycoONE: I didn't mean animation numbers.... ok, maybe you should just document self.current_state (which was an index into the state array)

Alberth289346: No it is not an index, as it starts at 0. It also ends at 5, so there is a missing index somewhere.
I don't want to debug and fix all plant-related stuff now.

PR #1163, opened 2016-11-20
[RDY] Door alignments with TH

mugmuggy:
screenToWall - Swing door placements in corners of walls matches original TH which - north wall, west corner is 2 tiles over for example

door_floor_blueprint_markers - aligns with TH

checkDoorWalls - update to return an invalid_tile flag to identify the door blueprints that are invalid.

validDoorTile - removed the flag name checks. Does further additional checks for passable tiles in the walkable tiles zone of objects, except bench. Stops drinks machine and reception desk being placed in front of an door for example.

setDoorBlueprint - created a local function - doorWallOffsetCalculations as the functionality needs to be called on previous tile, aswell as current tile position as swing door floor blueprints take 3 tiles, mapSetCell had to be inserted intothe for loop, the rest of the updates related to the invalid door tile changes

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -711,7 +711,13 @@ function UIEditRoom:screenToWall(x, y)
     -- top corner
     local x_, _ = self.ui:WorldToScreen(cellx, celly)
     if x >= x_ then
-      return cellx + 1 + modifier, celly, "north"
+      -- correctly reflects (at least origin version) of TH.
+      -- Swing doors in top corner to the east, actually skip another tile
+      if swinging then
+        return cellx + 2 + modifier, celly, "north"
+      else
+        return cellx + 1 + modifier, celly, "north"
+      end
     else
       return cellx, celly + 1 + modifier, "west"
     end
@@ -727,7 +733,11 @@ function UIEditRoom:screenToWall(x, y)
     -- left corner
     local _, y_ = self.ui:WorldToScreen(cellx, celly)
     if y >= y_ + 16 then
-      return cellx + 1, celly, "south"
+      if swinging and cellx <= rect.x + 2 then
+        return cellx + 1 + modifier, celly, "south"
+      else
+        return cellx + 1, celly, "south"
+      end
     else
       return cellx, celly - 1, "west"
     end
@@ -735,7 +745,11 @@ function UIEditRoom:screenToWall(x, y)
     -- right corner
     local _, y_ = self.ui:WorldToScreen(cellx, celly)
     if y >= y_ + 16 then
-      return cellx, celly + 1, "east"
+      if swinging and celly <= rect.y + 2 then
+        return cellx, celly + 1 + modifier, "east"
+      else
+        return cellx, celly + 1, "east"
+      end
     else
       return cellx - 1, celly, "north"
     end
@@ -752,19 +766,22 @@ function UIEditRoom:screenToWall(x, y)
     return rect.x, celly, "west"
   elseif (celly == rect.y - 1 or celly == rect.y) and rect.x <= cellx and cellx < rect.x + rect.w then
     -- north edge
-    if cellx == rect.x then
-      cellx = rect.x + 1
+    -- correctly reflects (at least origin version) of TH.
+    -- Swing doors in top corner to the east, actually skip another tile
+    if swinging and cellx <= rect.x + 2 then
+      cellx = rect.x + 3
+    elseif cellx == rect.x then
+      cellx = rect.x + 1 + modifier
     elseif cellx == rect.x + rect.w - 1 then
       cellx = rect.x + rect.w - 2
-    end
-    if swinging and cellx <= rect.x + 1 then
-      cellx = cellx + 1
     end
     return cellx, rect.y, "north"
   elseif (cellx == rect.x + rect.w or cellx == rect.x + rect.w - 1) and
       rect.y <= celly and celly < rect.y + rect.h then
-    -- east edge
-    if celly == rect.y then
+      -- east edge
+    if swinging and celly <= rect.y + 1 then
+      celly = rect.y + 2
+    elseif celly == rect.y then
       celly = rect.y + 1
     elseif celly == rect.y + rect.h - 1 then
       celly = rect.y + rect.h - 2
@@ -773,7 +790,9 @@ function UIEditRoom:screenToWall(x, y)
   elseif (celly == rect.y + rect.h or celly == rect.y + rect.h - 1) and
       rect.x <= cellx and cellx < rect.x + rect.w then
     -- south edge
-    if cellx == rect.x then
+    if swinging and cellx <= rect.x + 1 then
+      cellx = rect.x + 2
+    elseif cellx == rect.x then
       cellx = rect.x + 1
     elseif cellx == rect.x + rect.w - 1 then
       cellx = rect.x + rect.w - 2
@@ -1053,11 +1072,20 @@ function UIEditRoom:setBlueprintRect(x, y, w, h)
   rect.h = h
 end
 
+--25 north wall, west
+--26 north wall, east
+--27 east wall, south
+--28 east wall, north
+--29 south wall, west
+--30 south wall, east
+--31 west wall, south
+--32 west wall, north
+-- single door blue print values matching TH
 local door_floor_blueprint_markers = {
-  north = 25,
-  east = 28,
-  south = 29,
-  west = 32,
+  north = 26,
+  east = 27,
+  south = 30,
+  west = 31
 }
 
 local window_floor_blueprint_markers = {
@@ -1086,62 +1114,84 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
     dx = 1
     dy = 0
   end
-
+  local invalid_tile = 0
   if th:getCell(x, y, wall_num) % 0x100 ~= 0 then
-    return false
+    invalid_tile = 4
   end
 
   -- If it is a swing door there are two more locations to check.
   if has_swingdoor then
-    if th:getCell(x - dx, y - dy, wall_num) % 0x100 ~= 0 or
-        th:getCell(x + dx, y + dy, wall_num) % 0x100 ~= 0 then
-      return false
+    if th:getCell(x - dx, y - dy, wall_num) % 0x100 ~= 0 then
+      invalid_tile = invalid_tile | 2
+    end
+    if th:getCell(x + dx, y + dy, wall_num) % 0x100 ~= 0 then
+      invalid_tile = invalid_tile | 8
     end
   end
-  return true
+  return invalid_tile
 end
 
 --! Check whether the given tile can function as a door entry/exit tile.
 --!param xpos (int) X position of the tile.
 --!param ypos (int) Y position of the tile.
 --!param player_id (int) Player id owning the hospital.
---!param flag_names (array) If set, array with two additional required properties.
+--!param wall wall facing direction
+--!param wall (string) wall facing direction
+--!param world wall facing direction
 --!return Whether the tile is considered to be valid.
-local function validDoorTile(xpos, ypos, player_id, flag_names)
+local function validDoorTile(xpos, ypos, player_id, wall, world)
   local th = TheApp.map.th
-
   local tile_flags = th:getCellFlags(xpos, ypos)
-  if not (tile_flags.buildable or tile_flags.passable or tile_flags.owner == player_id) then return false end
-  if not flag_names then return true end
-  return tile_flags[flag_names[1]] and tile_flags[flag_names[2]]
+  if not (tile_flags.buildable and tile_flags.owner == player_id) then
+    -- any object will cause it to be blocked (ignore litter)
+    if not(tile_flags.thob == 0 or tile_flags.thob == 62) then return false end
+    if tile_flags.passable then
+      -- check for any object with a passable tile built within the door footprint
+      -- drinks machine and reception desk in corridors for example
+      for o in pairs(world:findAllObjectsNear(xpos, ypos, 1)) do
+        if o.object_type.id ~= "bench" then
+          for _, xy in ipairs(o:getWalkableTiles()) do
+            if xy[1] == xpos and xy[2] == ypos then
+              return false
+            end
+          end
+        end
+      end
+    end
+  end
+  return (tile_flags.thob == 0 or tile_flags.thob == 62)
 end
 
 function UIEditRoom:setDoorBlueprint(orig_x, orig_y, orig_wall)
-  local x = orig_x
-  local y = orig_y
-  local wall = orig_wall
-
-  -- Used to get the adjacent tiles when placing swing doors.
-  local x_mod
-  local y_mod
-  if wall == "south" then
-    y = y + 1
-    wall = "north"
-    x_mod = 2
-  elseif wall == "east" then
-    x = x + 1
-    wall = "west"
-    y_mod = 2
-  elseif wall == "north" then
-    x_mod = 2
-  else
-    y_mod = 2
+  local function doorWallOffsetCalculations(x, y, wall)
+    local x_mod
+    local y_mod
+    if wall == "south" then
+      y = y + 1
+      wall = "north"
+      x_mod = 2
+    elseif wall == "east" then
+      x = x + 1
+      wall = "west"
+      y_mod = 2
+    elseif wall == "north" then
+      x_mod = 2
+    else
+      y_mod = 2
+    end
+    return x, y, x_mod, y_mod, wall
   end
+  
+  local x, y, x_mod, y_mod, wall = doorWallOffsetCalculations(orig_x, orig_y, orig_wall)
   local map = TheApp.map.th
 
   if self.blueprint_door.anim then
     if self.room_type.swing_doors then
       if self.blueprint_door.anim[1] then
+        -- retrieve the old door position details to reset the blue print
+        local oldx, oldy, oldx_mod, oldy_mod, old_wall = doorWallOffsetCalculations(self.blueprint_door.floor_x,
+                                                           self.blueprint_door.floor_y,
+                                                           self.blueprint_door.wall)
         -- If we're dealing with swing doors the anim variable is actually a table with three
         -- identical "doors".
         for i, anim in ipairs(self.blueprint_door.anim) do
@@ -1149,9 +1199,11 @@ function UIEditRoom:setDoorBlueprint(orig_x, orig_y, orig_wall)
             self.blueprint_door.old_flags[i])
           anim:setTag(nil)
           self.blueprint_door.anim[i] = nil
+          oldx = oldx_mod and (self.blueprint_door.floor_x + (i - oldx_mod)) or self.blueprint_door.floor_x
+          oldy = oldy_mod and (self.blueprint_door.floor_y + (i - oldy_mod)) or self.blueprint_door.floor_y
+          map:setCell(oldx, oldy, 4, 24)
         end
-        map:setCell(self.blueprint_door.floor_x, self.blueprint_door.floor_y, 4, 24)
-      end
+       end
     else
       self.blueprint_door.anim:setAnimation(self.anims, self.blueprint_door.old_anim,
         self.blueprint_door.old_flags)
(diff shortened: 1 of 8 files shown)


CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1182,6 +1182,16 @@ function UIEditRoom:setDoorBlueprint(orig_x, orig_y, orig_wall)
     return x, y, x_mod, y_mod, wall
   end
   
+  local function BitAND(a,b)--Bitwise and
```
TheCycoONE: Language extensions like this function belong in utility.lua

Please add spaces after commas, and between operators, matching the style of the surrounding code.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1202,49 +1264,62 @@ function UIEditRoom:setDoorBlueprint(orig_x, orig_y, orig_wall)
...
   end
-  if not self.blueprint_door.valid then
-    flags = flags + 16 -- Use red palette rather than normal palette
-  end
   if self.room_type.swing_doors then
-    for _, animation in ipairs(anim) do
-      animation:setAnimation(self.anims, 126, flags)
+    for i, animation in ipairs(anim) do
+    -- calculation here to flag blocked blueprint tiles
+      animation:setAnimation(self.anims, 126, flags + (BitAND(invalid_tile,2^i)>0 and 1 or 0)*16)
```
Alberth289346: Maybe a "hasBit" function is better, as that's what you're testing here.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,72 +1114,96 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
-      return false
+    if th:getCell(x - dx, y - dy, wall_num) % 0x100 ~= 0 then
+      invalid_tile = invalid_tile | 2
+    end
+    if th:getCell(x + dx, y + dy, wall_num) % 0x100 ~= 0 then
+      invalid_tile = invalid_tile | 8
     end
   end
-  return true
+  return invalid_tile
```
Alberth289346: 'true' and 'false'  are very different from '0' or '4', please adapt (or create) the function documentation, so we can read what the magic numbers mean.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,72 +1114,96 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
+    -- any object will cause it to be blocked (ignore litter)
+    if not(tile_flags.thob == 0 or tile_flags.thob == 62) then return false end
+    if tile_flags.passable then
+      -- check for any object with a passable tile built within the door footprint
+      -- drinks machine and reception desk in corridors for example
+      for o in pairs(world:findAllObjectsNear(xpos, ypos, 1)) do
+        if o.object_type.id ~= "bench" then
+          for _, xy in ipairs(o:getWalkableTiles()) do
+            if xy[1] == xpos and xy[2] == ypos then
+              return false
```
Alberth289346: This looks a lot like your other function, maybe refactor it into its own function, and use that at both places?

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,72 +1114,96 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
-    x_mod = 2
-  elseif wall == "east" then
-    x = x + 1
-    wall = "west"
-    y_mod = 2
-  elseif wall == "north" then
-    x_mod = 2
-  else
-    y_mod = 2
+  local function doorWallOffsetCalculations(x, y, wall)
```
Alberth289346: Document the function

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,72 +1114,96 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
+  
+  local x, y, x_mod, y_mod, wall = doorWallOffsetCalculations(orig_x, orig_y, orig_wall)
   local map = TheApp.map.th
 
   if self.blueprint_door.anim then
     if self.room_type.swing_doors then
       if self.blueprint_door.anim[1] then
+        -- retrieve the old door position details to reset the blue print
+        local oldx, oldy, oldx_mod, oldy_mod, old_wall = doorWallOffsetCalculations(self.blueprint_door.floor_x,
+                                                           self.blueprint_door.floor_y,
```
Alberth289346: Line continuations are done by a double indent on the line below. That avoids that code shifts all the way to the right margin, creating many lines with mostly white space.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1202,49 +1254,62 @@ function UIEditRoom:setDoorBlueprint(orig_x, orig_y, orig_wall)
...
   end
-  if not self.blueprint_door.valid then
-    flags = flags + 16 -- Use red palette rather than normal palette
-  end
   if self.room_type.swing_doors then
-    for _, animation in ipairs(anim) do
-      animation:setAnimation(self.anims, 126, flags)
+    for i, animation in ipairs(anim) do
+    -- calculation here to flag blocked blueprint tiles
+      animation:setAnimation(self.anims, 126, flags + ((invalid_tile & 2^i)>0 and 1 or 0)*16)
```
Alberth289346: more bit tests.
Also, all operators have white space around them, "^", ">", and "*"

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1202,49 +1254,62 @@ function UIEditRoom:setDoorBlueprint(orig_x, orig_y, orig_wall)
...
     end
   else
-    anim:setAnimation(self.anims, 126, flags)
+    anim:setAnimation(self.anims, 126, flags + (invalid_tile ~= 0 and 1 or 0)*16)
```
Alberth289346: "*" needs whitespace

CorsixTH/Lua/utility.lua
```diff
@@ -311,3 +311,14 @@ function rangeMapLookup(number, buckets)
...
   assert(false) -- Should never get here.
 end
+
+-- work around for bitwise operators
+-- see http://lua-users.org/wiki/BitwiseOperators
+function bit(p)
+  return 2 ^ (p - 1)  -- 1-based indexing
+end
+
+-- Typical call:  if hasbit(x, bit(3)) then ...
```
Alberth289346: This pattern is actually the only sane way to call this function, why not fold the "bit" function into this one, and make the second parameter the bit number you want to test?
(oh, and please don't make this 1-based, preserving the 2^p relation is a lot saner imho.)

Also, document the function. See also https://github.com/CorsixTH/CorsixTH/wiki/CodeDocumentationStyle

CorsixTH/Lua/utility.lua
```diff
@@ -311,3 +311,14 @@ function rangeMapLookup(number, buckets)
...
+function hasbit(x, p)
+  return x % (p + p) >= p       
+end
```
Alberth289346: Last line should also have a \n character. For your editor that probably means you should have an empty line below the last code line.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,72 +1114,96 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
-  return true
+  return invalid_tile
 end
 
 --! Check whether the given tile can function as a door entry/exit tile.
 --!param xpos (int) X position of the tile.
 --!param ypos (int) Y position of the tile.
 --!param player_id (int) Player id owning the hospital.
---!param flag_names (array) If set, array with two additional required properties.
+--!param wall wall facing direction
```
sgtbigman: I believe you can remove this documented parameter line since it seems to be a duplicate of the next line (minus the `(string)`).

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,72 +1114,96 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
+--!param wall (string) wall facing direction
+--!param world wall facing direction
```
sgtbigman: This line should be changed from the copied `wall facing direction`.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,72 +1114,96 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
             self.blueprint_door.old_flags[i])
           anim:setTag(nil)
           self.blueprint_door.anim[i] = nil
+          oldx = oldx_mod and (self.blueprint_door.floor_x + (i - oldx_mod)) or self.blueprint_door.floor_x
+          oldy = oldy_mod and (self.blueprint_door.floor_y + (i - oldy_mod)) or self.blueprint_door.floor_y
+          map:setCell(oldx, oldy, 4, 24)
         end
-        map:setCell(self.blueprint_door.floor_x, self.blueprint_door.floor_y, 4, 24)
-      end
+       end
```
sgtbigman: I think an extra space was added here.

mugmuggy:
Seeking more feedback on this as I've combined 1164 changes here as they use the same refactored function now and that bit of code is more less complicated than the greater change

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,71 +1114,90 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
-      return false
+    if th:getCell(x - dx, y - dy, wall_num) % 0x100 ~= 0 then
+      invalid_tile = invalid_tile + 2
+    end
+    if th:getCell(x + dx, y + dy, wall_num) % 0x100 ~= 0 then
+      invalid_tile = invalid_tile + 8
     end
   end
-  return true
+  return invalid_tile
```
TheCycoONE: Now that you are returning a number instead of a boolean, fix the ```!return``` line in the function comment.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,71 +1114,90 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
-local function validDoorTile(xpos, ypos, player_id, flag_names)
+local function validDoorTile(xpos, ypos, player_id, world)
   local th = TheApp.map.th
-
   local tile_flags = th:getCellFlags(xpos, ypos)
-  if not (tile_flags.buildable or tile_flags.passable or tile_flags.owner == player_id) then return false end
-  if not flag_names then return true end
-  return tile_flags[flag_names[1]] and tile_flags[flag_names[2]]
+  -- check builable and own it
+  if not (tile_flags.buildable or tile_flags.owner == player_id) then return false end
```
TheCycoONE: This condition is wrong. It was wrong before your patch too. The tile needs to be buildable *and* owned by the player.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,71 +1114,90 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
+  --! Calculate position offsets and door blueprint wall values
+  --! param x - doors blueprint x value
+  --! param y - doors blueprint y value
+  --! param wall - original wall orientation
+  --! return x - updated x value
+  --! return y - updated y value
+  --! return x_mod - offest value to apply to tile count to determine relative position
+  --! return y_mod - offset value to apply to tile count to determine relatitve position
+  --! return wall - wall orientation style (only 2 styles)
+  local function doorWallOffsetCalculations(x, y, wall)
```
TheCycoONE: Move this function up to file scope (out of the setDoorBlueprint function). We are progressing to a style where we avoid nested functions unless closure is needed.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1086,71 +1114,90 @@ local function checkDoorWalls(x, y, wall, has_swingdoor)
...
+        local oldx, oldy, oldx_mod, oldy_mod, old_wall = doorWallOffsetCalculations(self.blueprint_door.floor_x,
+          self.blueprint_door.floor_y, self.blueprint_door.wall)
         -- If we're dealing with swing doors the anim variable is actually a table with three
         -- identical "doors".
         for i, anim in ipairs(self.blueprint_door.anim) do
           anim:setAnimation(self.anims, self.blueprint_door.old_anim[i],
             self.blueprint_door.old_flags[i])
           anim:setTag(nil)
           self.blueprint_door.anim[i] = nil
+          oldx = oldx_mod and self.blueprint_door.floor_x + (i - oldx_mod) or self.blueprint_door.floor_x
```
TheCycoONE: Why did you put this inside the anim loop? It looks like it would do the same thing every loop iteration unless I'm missing something.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1202,49 +1249,61 @@ function UIEditRoom:setDoorBlueprint(orig_x, orig_y, orig_wall)
...
-      flag_names = {"buildableWest", "buildableEast"}
-    end
-    local player_id = self.ui.hospital:getPlayerIndex()
-
-    if not validDoorTile(x, y, player_id, flag_names) or
-        not validDoorTile(x2, y2, player_id, flag_names) then
+  local world = self.ui.app.world
+  -- invalid_tile used to select the individual blueprint that is blocked
+  local invalid_tile = checkDoorWalls(x, y, wall, self.room_type.swing_doors)
+  self.blueprint_door.valid = invalid_tile == 0
```
TheCycoONE: Add braces around ```invalid_tile == 0``` to avoid confusion

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1202,49 +1249,61 @@ function UIEditRoom:setDoorBlueprint(orig_x, orig_y, orig_wall)
...
+  -- Ensure that the door isn't being built on top of an object
+  local player_id = self.ui.hospital:getPlayerIndex()
+  if not validDoorTile(x, y, player_id, world) or
+      not validDoorTile(x2, y2, player_id, world) then
+    self.blueprint_door.valid = false
+    invalid_tile = 4
```
TheCycoONE: Shouldn't this be ```+4```

TheCycoONE:
You seem to have ended up with one of @Alberth289346's commits in your PR. It shouldn't be there. Do you know enough about git to remove it cleanly?

I'm not a big fan of using a bit mask for invalid_tile. Bit manipulation doesn't play to lua's strengths, a table seems more appropriate. At the very least it should be documented (I didn't see any documentation on it / what the bit positions mean) in your PR.

TheCycoONE:
I am largely happy with the direction of the latest attempt. I think the flag name passthrough could be clearer, I had to read your comment to know what it meant. Maybe sharable or exclusive?

CorsixTH/Lua/dialogs/edit_room.lua
mugmuggy: Just to cover this off - I've updated the code to support the 3 tile blueprint for swing doors.

The doorWallOffsetCalculations function returns nil for oldx_mod/oldy_mod depending on orientation and the (i -oldx(y)_mod) calculation performed within.

The old code would end that loop and then call
map:setCell(self.blueprint_door.floor_x, self.blueprint_door.floor_y, 4, 24) which would be the centre tile of the door.

But that is just one blueprint tile being reset. So upon adding the support for the blueprint for the entire 3 tiles of the swing door, all 3 tiles need to be cleaned up, and the supporting calculation is done on the orientation.

CorsixTH/Lua/dialogs/edit_room.lua
mugmuggy: I have another push for comment, as this actually needed to be a bitwise operation and due to lack of 5.1 support for bitwise operations I've got a bit of a workaround in place to deal with that.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1141,43 +1145,43 @@ local function validDoorTile(xpos, ypos, player_id, world)
   local th = TheApp.map.th
   local tile_flags = th:getCellFlags(xpos, ypos)
   -- check builable and own it
-  if not (tile_flags.buildable or tile_flags.owner == player_id) then return false end
+  if not tile_flags.buildable and tile_flags.owner ~= player_id then return false end
```
TheCycoONE: Still pretty sure you actually want not buildable or not owned by the player here.

TheCycoONE:
That luadoc error isn't very helpful. I'll try to get to the bottom of it.

CorsixTH/Lua/utility.lua
```diff
@@ -311,3 +311,37 @@ function rangeMapLookup(number, buckets)
...
+  end
+else
+  -- this is a pseudo bitwise OR operation
+  -- assumes value2 is always a power of 2 (limits carry errors in the addition)
+  -- mimics the logic of hasBit with the addition if bit not set
+  function bitOr(value1, value2)
+    return value1 % (value2 + value2) >= value2 and value1 or value1 + value2
+  end
+  --! Check bit is set
+  --! param value - value to check set bit of
```
TheCycoONE: Not sure if it's the cause of the error, but there should be no space between the `!` and `param` here. See https://github.com/CorsixTH/CorsixTH/wiki/Coding-Conventions#ldocgen-syntax

CorsixTH/Lua/world.lua
```diff
@@ -2708,3 +2708,32 @@ function World:resetSideObjects()
     end
   end
 end
+
+--[[ When placing doors and objects the passable tiles need to be checked for overlapping
+passable tiles. This presents problems with objects like Bench where the passable tile
+is not for exclusive use of the Bench (another object can share that same tile)
+the footprint.shareable differentiates shareable passable tiles, and exclusive use
+passable tiles (the norm for most objects)]]
+--! param x - x map tile position
```
TheCycoONE: Similarly there should be no space between `!` and `param` or `!` and `returns`. `returns` should be `return` and the type should be in round brackets.

CorsixTH/Lua/utility.lua
mugmuggy: I have 5.3 supported syntax for the bitwise operators in utility.lua - specifically bitOr and the value1 | value2
lua_tokenise needs a slight alteration to include them 
operator = Ct(C(P".." + P"<=" + P">=" + P"==" + P"~=" + P">>" + P"<<" + S"+-*/^%<>#&|~") * Cc"operator"),

Travis still fails as it doesn't use lua 5.3 (yet) - luacheck on utility.lua fails for the same reason, thats not too much more work to build lua in the travis build.

TheCycoONE: Thanks for looking into it! Those changes seem like a prerequisite for this PR, as long as you want to continue using lua 5.3 bitwise operators.

It's unfortunate that we can't continue to automatically build and test lua 5.2, as I'm sure compatibility will degrade if it's not used. That said the situations where the latest CorsixTH was available, but only lua 5.2 are likely very close to 0 these days.

mugmuggy: Ok, I see the issue now. Just my misunderstanding.

I could do something like the following for those functions and as long as 5.3 was compiled with compatability on it would use the bit32 stuff, else fall back to the old stuff or just use the old stuff (which is probably the way to go). Just that bitOr routine isn't a real Or - it Ors a single bit only effectively

bitOr = bit32 and function(value1, value2) return bit32.bor(value1, value2) end or function(value1, value2)
  return value1 % (value2 + value2) >= value2 and value1 or value1 + value2
end

hasBit = bit32 and function(value, bit) return bit32.btest(value, 2^bit) end or function (value, bit)
  local p = 2 ^ bit
  return value % (p + p) >= p
end

CorsixTH/Lua/world.lua
```diff
@@ -2708,3 +2708,32 @@ function World:resetSideObjects()
...
+
+--[[ When placing doors and objects the passable tiles need to be checked for overlapping
+passable tiles. This presents problems with objects like Bench where the passable tile
+is not for exclusive use of the Bench (another object can share that same tile)
+the footprint.shareable differentiates shareable passable tiles, and exclusive use
+passable tiles (the norm for most objects)]]
+--!param x (int) x map tile position
+--!param y (int) y map tile position
+--!param distance (int) searchable distance for nearby objects
+--!returns (boolean) indicating if exclusively passable or not
```
TheCycoONE: s/`returns`/`return`

CorsixTH/Lua/dialogs/edit_room.lua
mugmuggy: I have found a defect with this logic that I will push out.
tile_flags.buildable will be false for every door when a bench footprint 'passable'/shareable tile is in front of the door, it basically only needs to check its owned by the player, as the rest of the code would check the object footprints for built on tiles

TheCycoONE:
I did a bit of testing, and I'm confused by the following:

Why is no door allowed here:
no-door

And a door allowed here?
yes-door

My understanding (though I haven't double checked the original) is that the bench footprint shouldn't block the door in either case.

MarkL1961:
@TheCycoONE   I would have said that it should not be allowed in either case, unless the bench was rotated.  This is purely from memory though, as the legs of those waiting would block access to the door.  If I can remember how to run TH in dosbox I will check this out.

mugmuggy:
Only thing I can think of is fa38fca was missed but then you couldn't build the one next to the radiator either if that was the case.

TheCycoONE:
I built the first bench before loading your branch, and the second after - perhaps existing benches aren't getting the new flag?

CorsixTH/Lua/objects/bench.lua
```diff
@@ -194,4 +194,12 @@ function Bench:onDestroy()
   Object.onDestroy(self)
 end
 
+--! This function is automatically called after loading a game and serves for compatibility.
+function Bench:afterLoad(old, new)
+  if old < 118 then
```
TheCycoONE: You also need to bump the savegame version to 118.

TheCycoONE:
The savegame version must be set. It also must be bumped as 118 was taken by the mouse capture pr. I'd like to land this next if you could focus here @mugmuggy?

TheCycoONE:
Merged

PR #1164, opened 2016-11-20
[RFC] Object placement fixes

### LINKED ISSUE #221: Should not be possible to place objects in front of doors

mugmuggy:
2 additional checks for object placement

Check 4 - check for existing object on tile

Check 5 - if placing anything besides a bench, checks that the walkable tiles of nearby objects do not overlap with any other object. ie can't place drinks machine in front of reception desk, but bench can be in the same position This combined with #1163 appears to fix #221 and #799

### DIFF
--- a/CorsixTH/Lua/dialogs/place_objects.lua
+++ b/CorsixTH/Lua/dialogs/place_objects.lua
@@ -662,6 +662,20 @@ function UIPlaceObjects:setBlueprintCell(x, y)
           is_object_allowed = map:getCellFlags(xpos, ypos, flags)[flag] and (player_id == 0 or flags.owner == player_id)
         end
 
+        -- Check 4: only one object per tile allowed original TH
+        is_object_allowed = is_object_allowed and map:getCellFlags(xpos, ypos, flags)["thob"] == 0 and (player_id == 0 or flags.owner == player_id)
+        
+        -- Check 5: only placeable if not on another objects passable footprint unless that is the invisible tile of a bench
+        if is_object_allowed and object.id ~= "bench" then
+          for o in pairs(world:findAllObjectsNear(xpos, ypos, 10)) do
+            if o.object_type.id ~= "bench" then
+              for _, xy in ipairs(o:getWalkableTiles()) do
+                if xy[1] == xpos and xy[2] == ypos then is_object_allowed = false end
+              end
+            end
+          end
+        end
+
         -- Having checked if the tile is good set its blueprint appearance flag:
         if is_object_allowed then
           if not tile.invisible then


TheCycoONE:
Did you look into modifying the bench's footprint instead? It seems like that passable flag isn't doing much now.

mugmuggy:
I haven't looked into it too much, as early investigations appeared to cause other logic to break, like building the bench facing a wall. It might be possible as I'm completely new to Lua as well as new to CorsixTH. I guess from what I've gathered so far, is that the passable tile (within the footprint), will still be needed within the footprint to prevent objects being placed in front of it anyway. It doesn't completely own it, but it is still required I guess from a navigation point of view too, unless there is a way to ensure passable into the tile in front another way.  #1163 includes the same sort of check for benches near the doors walkable tiles, where it seems passable has an effect.

Alberth289346:
It's weird that a bench has a special case, can you explain why that is? That is, I don't understand why the empty tile of say, a drinks machine is different from an empty tile of a bench.

If there is an explanation we should document that in the code, I think, so it won't get lost. If there is no explanation, something fishy is going on, that needs further investigation. Making special cases without clear understanding why it is the right solution doesn't seem like a good idea to me.

mugmuggy:
For the bench case, I'm aligning with the original TH for placement of objects. For example in the original you could not have 2 or 3 drinks machines sharing the same 'passable' tile. However, you can have 2 benchs and a drinks machine, all facing the one passable tile.
[BenchDrinkMachine](https://drive.google.com/open?id=0B9A8vBjxZZ6CUFlqaW1IaHlBMGc)
These are actually with the PR applied. (door tile in corsix is the centre door, th uses an end tile - lower x/y depending on wall direction) but are duplicating what you would see in TH.

Likewise the similar code in the door PR, A bench passable tile (for a doors walkable tiles) does not block a door, but a drinks machine/reception desk should or any other object should.
These are from Origin TH
[Door1](https://drive.google.com/open?id=0B9A8vBjxZZ6CUWNnMXM5Vmw4Vzg)
[Door2](https://drive.google.com/open?id=0B9A8vBjxZZ6CeGNYR0dPR3BOZGc)

And again with objects, the walkable tiles are checked to prevent the SideObjects being placed on a walkable tile except with the bench. Again Origin TH. [SideObjects](https://drive.google.com/open?id=0B9A8vBjxZZ6CRGZkY3MxUzAxaVU) Hopefully this clears up why bench is different.

Alberth289346:
Ok, fair enough. So the bench is different. Your explanation however suggests it's a systematic difference. That tile of the bench behaves different from other passable tiles.

Why not change or create a new tile type, add that to the bench, and adjust the code for that new tile type?

That gets rid of all the special case "bench" code that is being scattered around in the code now, and replaces it by another type of tile that has to be dealt with. While the latter is a bit more work, I think it's a much better solution. We may find other objects to need such a tile too, we may somewhen in the future move the bench data to a separate data file, or add more types of benches. In all these cases (and probably there are several others as well), the new tile type will continue to work without trouble, while hard-coded references to benches in the code will all die, and need to be replaced by some complicated new system to handle the generalized bench case or something.
Trying to keep object as much as possible separately from each other, and interface through some simple means like tile types, will keep the code much more maintainable towards the future.

PS For the next time, could the drop the pictures in the text instead of linking to your private google storage? It's easier to read, and all information is in one place. To preserver all information now, you need to keep those pictures there forever.

TheCycoONE:
Merged into #1163

PR #1166, opened 2016-11-21
[rdy as it ever will be] More fixes and improvements

Alberth289346:
Not entirely done, but patch was large enough already

### DIFF
--- a/CorsixTH/Lua/audio.lua
+++ b/CorsixTH/Lua/audio.lua
@@ -310,6 +310,32 @@ function Audio:cacheSoundFilenamesAssociatedWithName(name)
 end
 
 --[[
+This function's integer array parameters for the min and max silence lengths should provide lengths
+for this game's different speeds, indexed as follows:
+[1] Slowest [2] Slow [3] Normal [4] Fast [5] Maximum
+
+!param min_silence_lengths (integer array) The desired minimum silence lengths for this game's different speeds.
+!param max_silence_lengths (integer array) The desired maximum silence lengths for this game's different speeds.
+!param num_silences (integer) How many silence lengths should be in the returned table of generated lengths.
+!return (table) A table of randomly ordered integers for the generated silence lengths.
+--]]
+local function getSilenceLengths(min_silence_lengths, max_silence_lengths, num_silences)
+  local min_silence = min_silence_lengths[TheApp.world.tick_rate]
+  local max_silence = max_silence_lengths[TheApp.world.tick_rate]
+
+  local silences = {}
+  if min_silence == max_silence then
+    silences[1] = min_silence
+  else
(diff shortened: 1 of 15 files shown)


CorsixTH/Lua/rooms/operating_theatre.lua
```diff
@@ -283,24 +281,23 @@ function OperatingTheatreRoom:commandEnteringPatient(patient)
   end
   ----- END Save game compatibility -----
 
-  local room = self
```
TheCycoONE: Normally I agree with removing pointless aliasing, but in this case it may save a sneaky gotcha.

If operation_standby gets called with a : instead of a . then ```self``` would be shadowed by whatever table the callback happened to be a part of. This would be a programming error, but a subtle one that could be hard to catch - whereas room is safe.

TheCycoONE: Sorry, that's not true. I clearly need my morning tea.

CorsixTH/Lua/audio.lua
```diff
@@ -310,6 +310,32 @@ function Audio:cacheSoundFilenamesAssociatedWithName(name)
...
+local function getSilenceLengths(min_silence_lengths, max_silence_lengths, num_silences)
+  local min_silence = min_silence_lengths[TheApp.world.tick_rate]
+  local max_silence = max_silence_lengths[TheApp.world.tick_rate]
+
+  local silences = {}
+  if min_silence == max_silence then
+    silences[1] = min_silence
+  else
+    for i = 1, num_silences do
+      silences[i] = math.random(min_silence,max_silence)
```
sgtbigman: Minor: Missing space: `min_silence, max_silence`.

CorsixTH/Lua/rooms/operating_theatre.lua
```diff
@@ -104,6 +104,30 @@ function OperatingTheatreRoom:isOperating()
...
+--! Builds the second operation action (i.e. with the surgeon whose we
+--! see the back). Called either when the operation starts or when the
+--! operation is resumed after interruption caused by the picking up of
+--! the second surgeon.
+--!param multi_use (action): the first operation action (built with via buildTableAction1()).
+--!param operation_table_b (OperatingTable): slave object representing the operation table.
+local function buildTableAction2(multi_use, operation_table_b)
+  local num_loops = math.random(2, 5)
+
+  local loop_callback_use_object = --[[persistable:operatring_theatre_use_callback]] function(action)
```
sgtbigman: Hmm, not sure if worth pursuing, but I wonder if `operatring` could be changed to `operating` in these callback names without much hassle (unless the spelling was intentional).

Alberth289346:
Fixed missing space

Issue #1008, opened 2016-03-24
On graph, labels should not fall below the bottom of the chart. They should be coloured to match lines.


altiereslima:
Steps to Reproduce
1. Open Status Menu

Or
1. Open Charts Menu
2. See screenshots

System Information
CorsixTH Version: 0.50

Operating System: Windows 7

Theme Hospital Version: CD

Additional Info

Cursixth
![cursixth charts](https://cloud.githubusercontent.com/assets/18044879/14016980/91df2516-f1a2-11e5-8a27-569e264fff72.png)
![cursixth](https://cloud.githubusercontent.com/assets/18044879/14016981/91e15368-f1a2-11e5-87fa-6154528dcde4.png)

DOSBox
![otiginal charts](https://cloud.githubusercontent.com/assets/18044879/14016982/91e2eda4-f1a2-11e5-8a17-c7f8f5b915ff.png)
![otiginal status](https://cloud.githubusercontent.com/assets/18044879/14016983/91ed45ec-f1a2-11e5-83da-62fe8ffe158d.png)

TheCycoONE:
We haven't implemented the AI hospitals yet which is why you see the population pie chart in one case but not the other.

Lets focus this issue on the position and color of the legend in the graph.

lewri:
Partly resolved. Rest is a duplicate of #967

PR #1168, opened 2016-12-03
[rdy] Keep labels of the graph within their space, addresses #1008

Alberth289346:
Don't print graph labels below the bottom of the graph, 2 pictures to show below.

The order of the entries in a clustered sequence looks unstable, but it's not a problem imho, as such labels all belong to the same position at the screen, so order has little meaning anyway

screenshot from 2016-12-03 13-41-36

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local runDebugger = dofile "run_debugger"
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 116
+local SAVEGAME_VERSION = 117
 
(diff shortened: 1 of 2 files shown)


Alberth289346:
Replaced // by math.floor, as some luas don't understand integers

Alberth289346:
If you remove lines, the labels don't move, I see no other way than recompute label position when you add or remove lines.

Alberth289346:
Well, this would do it. Not sure the commit sequence makes a lot of sense, but it rewrites pretty much the entire file. The sequence is a compromise between not too many commits and not too many changes at the same time.

Patch only solves the positioning problem of the graph labels. The text-label colour problem is currently blocked (for me) at the magic Lua/C++ bridge for fonts.

PR #1175, opened 2016-12-08
[rdy] Order of the sliders could make the 'guess_cure' slider unusable.

### LINKED ISSUE #1174: diagnosis procedure sliders getting stuck in hospital policy screen (F9)
#### Describe the issue

In the hospital policy screen (F9) under diagnosis procedure when the 'send home' and 'guess at cure' slider are close the 'send home' slider can no longer be moved. As the 'guess at cure' slider can only be moved to the right of the 'send home' button this results in the 'send home' slider getting stuck if both sliders are moved to the right.

#### Expected Behavior

Both sliders should be moveable even if they are in the same spot.

#### System Information
**CorsixTH Version:** 2fe2658da038a7537768c6b5e06cf03991715a52

**Operating System:**  Linux Debian Jessie

Alberth289346:
Fixes #1174

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local runDebugger = dofile "run_debugger"
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 115
+local SAVEGAME_VERSION = 116
 
 class "App"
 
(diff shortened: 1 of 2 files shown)


MarkL1961:
For me it now does not save your choice.
I move the sliders and the next time I open they are back in the default positions

Alberth289346:
Thanks @MarkL1961, fixed that.
I also updated the patch for comments by @TheCycoONE

Alberth289346:
And fixed again, please review

Alberth289346:
Comment in the afterload fixed

MarkL1961:
Works for me and fixes the save in the original issue

Alberth289346:
Thanks for testing!

PR #1196, opened 2017-02-25
Added version tag

RyanJGray:
Added a proper version tag for the latest source build of CorsixTH (Version number 117.)

It was just showing as "Trunk" and now it will show as "v0.61" (Whoever updated the version number to 117 forgot to add a version tag for it.)


MarkL1961:
Version numbers are for released versions and the last release was 0.60.  Between released builds they have always been named as "trunk", we then refer to the build numbers

RyanJGray:
Ah okay, thanks for clarifying.

Issue #573, opened 2014-12-24
Patients should queue for rooms with 0 max queue set


TheCycoONE:
_From [fang.xianfu](https://code.google.com/u/fang.xianfu/) on October 19, 2013 07:06:32_

As discussed here: http://forums.corsixth.com/index.php/topic,5454.0.html This patch removes the exception for rooms with a max queue set to 0 in World:findRoomNear(). This means that there will be functionally no difference between a room with 6 patients queueing and a max queue of 6, and a room with a max queue of 0. See the thread for more info.

I was amazed that this change was so tiny, but it looks like this was all it took. Incidentally this may also fix a bug with the toilet callback in patient.lua that would prevent the callback from being called even though there was a usable toilet. 

findRoomNear() is used in the following places:

seek_room.lua, seek_toilets.lua - used to find nearest sought room. Intended change to allow patients to queue for rooms with 0 max queue, which would be skipped before. Because they use advanced mode, rooms with max queue 0 always receive the "full queue" penalty to their getUsageScore() and will not be returned unless all rooms have full queues.

patient.lua - used to find a nearby toilet to determine if one is built and so it's ok to queue a seek_toilet action. Previously this function couldn't tell the difference between an unbuilt toilet and a toilet with 0 max queue, so the callback would be created if the only available toilets had a max queue of 0. If the max queue was increased after the callback was created, it would never be called despite there now being an available toilet.

This change will cause a seek_toilet to be queued even if all available toilets have 0 max queue, where previously it would have acted as if there were no toilets built. Since patients will now be allowed to queue for toilets with 0 max queue by seek_toilet.lua, this is appropriate. 

staff.lua, seek_staffroom.lua - used to determine if a staffroom exists or if we need to wait for it to build, and to find nearest staff room to rest in. Should be unaffected as staff rooms don't have user-settable queues.

Thanks!

**Attachment:** [zeroqueue.patch](http://code.google.com/p/corsix-th/issues/detail?id=1717)

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1717_

TheCycoONE:
I guess the patch is lost on this, but the idea isn't a bad one. It's clear from the thread that some people want the original theme hospital behavior, and when there is controversy it's good to side with the original game.

fangxianfu:
I reconstructed the patch, I'll make a PR :)

PR #1197, opened 2017-02-26
Patients queue for rooms with max queue size 0

fangxianfu:
Related to issue #573 

This restores the original TH behaviour of having rooms with max queue size 0 be used once all available queues are above their max queue size.

Closed rooms still receive a penalty in findRoomNear so they should never be used before open rooms.

Not 100% sure how this will interact with the usage of findRoomNear in patient.lua, which doesn't use advanced mode.

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1806,7 +1806,7 @@ function World:findRoomNear(humanoid, room_type_id, distance, mode)
     distance = 2^30
   end
   for _, r in pairs(self.rooms) do repeat
-    if r.built and (not room_type_id or r.room_info.id == room_type_id) and r.is_active and r.door.queue.max_size ~= 0 then
+    if r.built and (not room_type_id or r.room_info.id == room_type_id) and r.is_active then
       local x, y = r:getEntranceXY(false)
       local d = self:getPathDistance(humanoid.tile_x, humanoid.tile_y, x, y)
       if not d or d > distance then


fangxianfu:
To get this working properly, it seems like the room.readiness_bonus needs reducing. 5 patients in queue is a *lot* of penalty - it effectively means that if you have one room open (max queue size > 0) and one closed (max queue size 0), the max queue size setting is ignored until the queue of the open room exceeds 5 patients. So this new ability to "soft close" rooms isn't useful if you wanted to set the max queue size of the open rooms to less than 5 - which, personally, I usually do.

It also seems quite weird even without this change, though. The current readiness_bonus means that if you have, say, two Pharmacies, both with max queue size >5, and one has a nurse already and the other doesn't, that patients won't choose the staff-member-less room until the queue of the room with a staff member is 5 or more. The weird part about this for me is that because patients will reject ever going to the room without a staff member, a staff member will never be requested for it. So you get the kind of weird result of two open pharmacies, one with a queue of 5 and one with a queue of 0, and no call being sent for another nurse.

I'm going to experiment with some different values for readiness_bonus; has it been studied before?

TheCycoONE:
The tribal knowledge on this project is largely lost as the original contributors have moved on. I am not aware of any such experiments.

MarkL1961:
The zero thing was added several years ago and the reasons for it where in an issue in google code.  It was that long ago I cannot remember all the details, only it was something we came up with in the absence of any knowledgeable code writers (Edvin was in his last year at university I think, so was not around that much).  Myself and Grimdoc tried to keep things ticking over for quite a while.
It may not have been the best solution though.
I think there was an issue around building rooms to fill the available space and not having the staff to make use of them i.e build two operating theatres and only have access to two surgeons.  So close one off to stop having one surgeon in each and neither of them actually working.

TheCycoONE:
@MarkL thanks for the insight.

Issue #1199, opened 2017-03-12
Game freezes after earthquake


lucasmr:
Using Windows 7 64.

stack traceback:
        C:\Users\Lucas\Downloads\CorsixTH\Lua\world.lua:589: in method 'tickEarthquake'
        C:\Users\Lucas\Downloads\CorsixTH\Lua\world.lua:1003: in method 'onTick'
        C:\Users\Lucas\Downloads\CorsixTH\Lua\app.lua:1005: in function <C:\Users\Lucas\Downloads\CorsixTH\Lua\app.lua:1002>
        (...tail calls...)
        C:\Users\Lucas\Downloads\CorsixTH\Lua\app.lua:906: in function <C:\Users\Lucas\Downloads\CorsixTH\Lua\app.lua:901>

Warning: No event handler for timer

Save file: http://filebin.ca/3FMoNG3xPI0g

TheCycoONE:
If this was 0.60, earthquakes have been largely rewritten since then. Try the latest builds instead. Please include a version number in your tickets.

lucasmr:
Happens on 0.60 and also on the latest dev build.

MarkL1961:
Your save is from 0.60.  Here is the error message running this save after around one minute



---------------------------------------------------------------

Welcome to CorsixTH Trunk!

---------------------------------------------------------------

This window will display useful information if an error occurs.

---------------------------------------------------------------

`  Will not check for updates since this is the Trunk version.
Savegame version changed from 111 (v0.60) to 116 (Trunk). The save was created using 111 (v0.60)
Warning: loaded savegame version 117 (Trunk) in older version 116 (Trunk).
Savegame version changed from 111 (v0.60) to 116 (Trunk). The save was created using 111 (v0.60)
An error has occurred!
Almost anything can be the cause, but the detailed information below can help the developers find the source of the error.
Running: The timer handler.
A stack trace is included below, and the handler has been disconnected.
C:\Users\Mark\Downloads\CorsixTH (c258bbf1 )\Lua\world.lua:589: attempt to compare number with nil
stack traceback:
        C:\Users\Mark\Downloads\CorsixTH (c258bbf1 )\Lua\world.lua:589: in method 'tickEarthquake'
        C:\Users\Mark\Downloads\CorsixTH (c258bbf1 )\Lua\world.lua:1003: in method 'onTick'
[...]

PR #1200, opened 2017-03-13
RDY: Fix loading saves with pending earthquakes

TheCycoONE:
Fixes the immediate problem with the earthquake not loading properly in  #1199 
Unfortunately there are further problems with that save, which cause problems immediately afterwards when trying to crash the problem


TheCycoONE:
Rebased on the latest master

PR #1202, opened 2017-04-02
Updated README.md

Zelgadis87:
Clarifies the requirements for the installation, trivial change.


TheCycoONE:
Yes, the new luacheck is noisy - we'll need to suppress some warnings until they can be cleaned up, but this PR is clear.

PR #1217, opened 2017-05-03
[RDY] Missing polish translations

tiriana:
Adds missing polish strings to:

Corsix main menu
Map Editor
Top menu
'Folders' menu
'Customize' menu

### DIFF
--- a/CorsixTH/Lua/languages/polish.lua
+++ b/CorsixTH/Lua/languages/polish.lua
@@ -386,6 +386,7 @@ tooltip = {
   },
   main_menu = {
     exit = "Nie, nie, proszę nie odchodź!",
+    custom_campaign = "Rozegraj kampanię stworzoną przez społeczność",
     custom_level = "Zbuduj swój szpital na dodatkowych poziomach",
     network = "Rozpocznij grę sieciową",
     quit = "Wyjdź",
@@ -397,6 +398,7 @@ tooltip = {
     },
     new_game = "Zacznij grę od samego początku‏",
     load_game = "Wczytaj zapisaną grę",
+    map_edit = "Stwórz własną mapę",
   },
   patient_window = {
     graph = "Kliknij, aby przełączyć się pomiędzy wykresem zdrowia osób a ich historią leczenia",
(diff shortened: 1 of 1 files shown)


tiriana:
I also wanted to translate the warning about missing/corrupted files: 

image

but I don't see key for it in translation file. Please let me know if there is a way do to it.

MarkL1961:
This warning is in app.lua.

I could be wrong, but I think this is loaded before the routine that runs the translations (I think it is something that has been mentioned before several years ago)

There are different people involved with the programming now, so they may have some thoughts on how this could be achieved.

@Alberth289346 @TheCycoONE

tiriana:
@MarkL1961 @Alberth289346 @TheCycoONE Just a ping. I don't know how long usually code review takes here. 

This one should be pretty straight forward - it only adds some translations.

MarkL1961:
Not long, someone would need to verify that the code is good and the translations are genuine.

TheCycoONE:
For most languages there just isn't the people to verify the accuracy of the translations.

When your changes are ready for review you should mark your PR with [RDY], if you are not ready for merge but would like someone to take a look, use [RFC].

Are you comfortable enough with git to squash your changes yourself?

As for your other question, it actually should be pretty straight forward to have localized the corrupt files message. The text is currently generated before the languages are loaded, but not displayed until after. That said it's not yet possible to just add a translation for it. An issue for that could be opened.

tiriana:
Yes, I'm comfortable with squash, I ues git on a daily basis, I just wanted to wait for fixes etc and than squash. I was playing today a bit and found one or too missing text, I will add them to this PR, squeshed. And I'll add the prefix than.

About the second thing - the translation of warning - I will create an improvement ticket for this as well.

thz

TheCycoONE:
@tiriana yes that would be helpful, thanks!

CorsixTH/Lua/languages/polish.lua
```diff
@@ -2196,9 +2209,67 @@ options_window = {
   cancel = "Anuluj",
   data_location = "Lokalizacja danych",
   font_location = "Lokalizacja czcionki",
-  language = "Game language",
+  language = "Język gry",
```
naugtur: I'd suggest leaving some indication that this is the language menu in English. 

`Język gry (language)`

Everyone who switched anything to Chinese accidentally knows why.

kamilkoterba:
+1 Translations are correct.

CorsixTH/Lua/languages/polish.lua
tiriana: Good idea. 

I'll leave the final decision to @TheCycoONE though, as it is something that would be useful for other languages too.

...

but on the other hand - the button on which this test is displayed might be to narrow for such long text. I can check it later and attach screenshot.

naugtur:
LGTM.

Also, if you ever need to check Polish translation without knowing Polish, just check if the word "dupa" is not lingering anywhere :)

PR #1221, opened 2017-06-01
Updating Spanish translation

IlDucci:


TheCycoONE:
@IlDucci what state is this in, could you annotate it with [WIP] or [RDY]?

IlDucci:
It is ready.

Issue #1210, opened 2017-04-09
Integer-based UI scaling


matejdro:
On modern hi-res monitors, CorsixTH UI is very tiny and text hard to read.

One solution to that is to pick lower resolution, but then menus and the text becomes blurry.

Original DOS version is only VGA, but when using DOSBox's integer scaling (scaling that only duplicates original pixels without doing any interpolation), its UI and text in it is still very sharp and nicely looking while being big enough to even play in a window.

How about an option that makes UI of the game scale in integer steps (2x, 3x, 4x, etc.)?

EDIT: There is a bit better explanation of the issue with pictures attached [in a comment below](https://github.com/CorsixTH/CorsixTH/issues/1210#issuecomment-312422714).

phobos2077:
Or just a checkbox "Double pixels" or something, like OpenRA?

TheCycoONE:
I've begun to implement a fairly naive scaling setting for the game using the same approach as zoom (using SDL_RenderSetLogicalSize)

TheCycoONE:
@matejdro if you're around, can I have your opinion on #1224

matejdro:
Thanks for looking into this. I tried it, but I don't think this is what this issue is about.

This pull request only increases window size, putting windowed experience on par with full screen experience. But it is still using blurry interpolation method for scaling (text is still blurry).

What I was talking about in this issue is scaling mode that just duplicates pixels. For example, if there is one blue pixel, and you have 2x scaling, this blue pixel will turn into 4 exactly the same blue pixels, without any interpolation or math involved.

For example take a look at dos version with integer scaling enabled in dosbox:

![dosbox](https://user-images.githubusercontent.com/507922/27760966-bba82388-5e53-11e7-92be-39578f48381e.png)

And this is your test build of CorsixTH:

![corsix](https://user-images.githubusercontent.com/507922/27760971-d2f86d5e-5e53-11e7-99bf-d85c13dc1c5f.png)

Even though both games are running at 640x480 resolution with 2x scaling, you can clearly see that DosBox version with integer scaling is very crisp, while CorsixTH's text and UI does not look very pleasant, it is too blurry.

TheCycoONE:
Thanks @matejdro 

I was hoping SDL could be a little more naive in it's scaling, but I think we have to do it all ourselves. The result will be better anyway, at least for those not using the built  in bitmap fonts.

theAlevan:
Any progress on scaling on hidpi screens?

[...]

PR #1224, opened 2017-06-16
[RDY] Window scaling for high dpi support

TheCycoONE:
New scaling option in the config allows the window size to be a multiple of the logical resolution.
This addressed #1210


Alberth289346:
While testing, it appears that in the settings.
- A "scale" parameter was added, and
- Size of the screen is interpreted as unscaled size.

Please extend the textual description how to interpret the scale and the screen size here.

TheCycoONE:
I've updated the description of the various window size / resolution settings. I want to properly expose scale in the settings page as well before this is merged.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -232,7 +232,7 @@ bool THRenderTarget::create(const THRenderTargetCreationParams* pParams)
     m_pFormat = SDL_AllocFormat(SDL_PIXELFORMAT_ABGR8888);
     m_pWindow = SDL_CreateWindow("CorsixTH",
                                  SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED,
-                                 pParams->iWidth, pParams->iHeight,
+                                 pParams->iWidth * pParams ->fWindowScale, pParams->iHeight * pParams->fWindowScale,
```
Alberth289346: Spurious space:
```` c
pParams ->fWindowScale
       ^
````

File: CorsixTH/Lua/dialogs/resizables/options.lua [hunk 2]
```diff
@@ -149,6 +164,8 @@ local --[[persistable:options_window_language_button]] function language_button(
 local --[[persistable:options_width_textbox_reset]] function width_textbox_reset() end
 local --[[persistable:options_height_textbox_reset]] function height_textbox_reset() end
 
+
+
```
Alberth289346: Any particular reason for adding empty lines?
While at it, just write a line documentation perhaps?

TheCycoONE: No, an oversight. I was going to write the callback function here, but I needed the closure.

TheCycoONE:
I have doubts about making committing these changes without:
a) An auto mode that is default
b) and HighDPI awareness in Windows

As is I don't think this feature brings any improvement to the situation on Windows and if we release with a scale of 1 and add auto later a lot of people will end up with tiny windows and have to edit their config

Alberth289346:
good point

TheCycoONE:
SDLs scaling is not ideal for our use.

Alberth289346:
In OpenGL you can configure interpolation with GL_LINEAR, see "Texture Filtering" at https://learnopengl.com/#!Getting-started/Textures . If SDL uses opengl for this, it may be configurable too

TheCycoONE:
It's slightly configurable: https://wiki.libsdl.org/SDL_HINT_RENDER_SCALE_QUALITY?highlight=(\bCategoryDefine\b)|(CategoryHints)

Alberth289346:
0 or "nearest" looks like very much "just take the pixel". OpenGL also names it "nearest". Does that do anything useful? (In the sense that the white of the letters should stay white, and not get interpolated or anti-aliased with background (which the OP sees as blurring, I think).)

TheCycoONE:
@reubot are you able to test / take over my PR then?

PR #1229, opened 2017-07-12
Fix compile error on non-existing std::exception class (TheCycoOne)

Alberth289346:


TheCycoONE:
To be fair the class exists, it just doesn't have a constructor that takes a string.

PR #1230, opened 2017-07-16
[RDY] Use C++11 standard threading for th_movie.cpp

TheCycoONE:
Replace SDL threading with C++11 threading.

In making this change I already fixed one possible deadlock where the mutex was
not released in write if the picture didn't have a buffer.

### DIFF
--- a/CorsixTH/Src/th_movie.cpp
+++ b/CorsixTH/Src/th_movie.cpp
@@ -54,20 +54,6 @@ extern "C"
 #define av_frame_free avcodec_free_frame
 #endif
 
-static int th_movie_stream_reader_thread(void* pState)
-{
-    THMovie *pMovie = (THMovie *)pState;
-    pMovie->readStreams();
-    return 0;
-}
-
-static int th_movie_video_thread(void* pState)
-{
-    THMovie *pMovie = (THMovie *)pState;
-    pMovie->runVideo();
-    return 0;
-}
-
 static void th_movie_audio_callback(int iChannel, void *pStream, int iStreamSize, void *pUserData)
 {
     THMovie *pMovie = (THMovie *)pUserData;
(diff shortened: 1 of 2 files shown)


TheCycoONE:
Sorry for the noise, I keep changing my mind on how far I want to take this. There is one more modification I want to make.

CorsixTH/Src/th_movie.cpp
```diff
@@ -166,24 +145,26 @@ void THMoviePictureBuffer::allocate(SDL_Renderer *pRenderer, int iWidth, int iHe
...
-    SDL_UnlockMutex(m_pMutex);
+    {
+        std::lock_guard<std::mutex> lock(m_mutex);
+        m_fAllocated = false;
+    }
+
     for(int i = 0; i < ms_pictureBufferSize; i++)
     {
-        SDL_LockMutex(m_aPictureQueue[i].m_pMutex);
+        std::lock_guard<std::mutex> pictureLock(m_aPictureQueue->m_mutex);
```
Alberth289346: Weird that you don't need a [i] here, while the previous version had one, and line 256 also seems to have one

CorsixTH/Src/th_movie.cpp
```diff
@@ -136,9 +116,8 @@ THMoviePictureBuffer::~THMoviePictureBuffer()
 void THMoviePictureBuffer::abort()
 {
     m_fAborting = true;
-    SDL_LockMutex(m_pMutex);
-    SDL_CondSignal(m_pCond);
-    SDL_UnlockMutex(m_pMutex);
+    std::lock_guard<std::mutex> lock(m_mutex);
+    m_cond.notify_all();
```
Alberth289346: SDL woke up 1 thread, but  I guess this is fine, as everybody will find the m_fAborting flag, and quit?

CorsixTH/Src/th_movie.cpp
TheCycoONE: Oops, yes this looks wrong.

CorsixTH/Src/th_movie.cpp
TheCycoONE: Intentional - releasing should unlock every thread waiting for the queue. It's theoretical though because currently only one other thread would be waiting.

TheCycoONE:
Fixed the picture lock, and added a second commit to make the aborting flag atomic, since it was relying on unsafe / undefined behavior (there is no guarantee without atomic that other threads would ever see the updated value, or that the code wouldn't be reordered.)

PR #1245, opened 2017-09-15
[RFC] Indicate to SDL2 when we are expecting textbox input

TheCycoONE:
This will open a virtual keyboard on appropriate environments and tell SDL2 to
fire SDL_TEXTINPUT events which handles unicode (IME) input. This commit does
not add processing of SDL_TEXTINPUT events.

I don't actually have any devices to test this on, so if anyone has a touch screen with an on demand virtual keyboard please run this and see if it works!

### DIFF
--- a/CorsixTH/Lua/window.lua
+++ b/CorsixTH/Lua/window.lua
@@ -26,6 +26,8 @@ class "Window"
 ---@type Window
 local Window = _G["Window"]
 
+local sdl = require("sdl")
+
 -- NB: pressed mouse buttons are denoted with a "mouse_" prefix in buttons_down,
 -- i.e. mouse_left, mouse_middle, mouse_right
 Window.buttons_down = permanent"Window.buttons_down" {}
@@ -973,9 +975,11 @@ function Textbox:setActive(active)
     self.cursor_state = true
     self.cursor_pos[1] = type(self.text) == "table" and #self.text or 1
     self.cursor_pos[2] = type(self.text) == "table" and string.len(self.text[#self.text]) or string.len(self.text)
+    sdl.startTextInput()
     -- Update text
     self.panel:setLabel(self.text)
   else
+    sdl.stopTextInput()
     self.cursor_state = false
   end
 
--- a/CorsixTH/Src/sdl_core.cpp
+++ b/CorsixTH/Src/sdl_core.cpp
@@ -357,10 +357,24 @@ static int l_get_ticks(lua_State *L)
     return 1;
 }
 
+static int l_start_text_input(lua_State *L)
+{
+    SDL_StartTextInput();
+    return 0;
+}
+
+static int l_stop_text_input(lua_State *L)
+{
+    SDL_StopTextInput();
+    return 0;
+}
+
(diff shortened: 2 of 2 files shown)


wolfy1339:
The appveyor build failed. When the build passes and there's a downloadable build i'll test this

CorsixTH/Src/sdl_core.cpp
```diff
@@ -348,10 +348,22 @@ static int l_get_ticks(lua_State *L)
     return 1;
 }
 
+static int l_start_text_input(lua_State *L)
+{
+    SDL_StartTextInput();
+}
```
Alberth289346: Given that it doesn't push anything, I guess this is missing "return 0;" to fix appveyor error

CorsixTH/Src/sdl_core.cpp
```diff
@@ -348,10 +348,22 @@ static int l_get_ticks(lua_State *L)
...
+
+static int l_stop_text_input(lua_State *L)
+{
+    SDL_StopTextInput();
+}
```
Alberth289346: here too "return 0;"

CorsixTH/Lua/window.lua
```diff
@@ -973,9 +973,11 @@ function Textbox:setActive(active)
     self.cursor_state = true
     self.cursor_pos[1] = type(self.text) == "table" and #self.text or 1
     self.cursor_pos[2] = type(self.text) == "table" and string.len(self.text[#self.text]) or string.len(self.text)
+    require("sdl").startTextInput()
```
Alberth289346: Any particular reason for this inline import? I'd prefer doing imports at the top of the file to avoid endless searches for imports

TheCycoONE: That's not how lua works, require caches the object in packages.loaded so every subsequent call pulls up the same object - it can even store state.

I picked up the convention from https://github.com/pkulchenko/MobDebug, but I concede we don't use it elsewhere and it isn't entirely clear to people who aren't read up on lua - I'll change it if the functions work.

TheCycoONE:
@wolfy1339 I have a fix, but I'm waiting for #1247 to ensure I give you a build with the latest SDL version.

wolfy1339:
@TheCycoONE As #1274 was merged can we start testing this?

TheCycoONE:
@wolfy1339 just had to push my changes and rebase on master. Should be testable now.

TheCycoONE:
Based on wolfy1339's tests, SDL2 doesn't bring up the on screen keyboard for Windows laptops. I have read that it does work on Windows RT (but only for UWP software). I'm not sure what other platforms are or are not supported.

The only other advantage is the potential to support IME character entry, but that will require some other input changes, and I don't have any ability to test that either.

My suggestion is to leave this PR open as long as it still merges cleanly in case someone comes along that can work with it, but not to merge for now.

lewri:
@TheCycoONE Do we still want this one open? It's been 4 years since last activity.

TheCycoONE:
Doesn't merge clean anymore and no one has offered to test with a Surface or something. Not to mention it was trivial to start with. Closing.

PR #1247, opened 2017-09-16
[HOLD] Use vcpkg for handling dependencies

TheCycoONE:
Note: luasocket is not included. There is no suitable release versions of that library.

Builds now take quite a bit longer - it might be nice to see if we can cache the compiled vcpkg packages in the future.


TheCycoONE:
Holding, depending on the direction of #1259

TheCycoONE:
Closing in favour of #1274

PR #1257, opened 2017-10-12
[RDY] Fix cppcheck errors

TheCycoONE:
Fixes all errors reported by cppcheck 1.80 using the default arguments. A few warnings remain in scanner.cpp.

Mostly this involved fixing edge case memory leaks, either by adding deletes or by moving away from using raw memory buffers entirely.

### DIFF
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -1226,6 +1226,43 @@ void THMap::setTemperatureDisplay(THMapTemperatureDisplay eTempDisplay)
     if (eTempDisplay < THMT_Count) m_eTempDisplay = eTempDisplay;
 }
 
+uint32_t THMap::thermalNeighbour(uint32_t &iNeighbourSum, bool canTravel, uint32_t relative_idx, THMapNode* pNode, int prevTemp) const
+{
+    int iNeighbourCount = 0;
+
+    THMapNode* pNeighbour = pNode + relative_idx;
+
+    // Ensure the neighbour is within the map bounds
+    THMapNode* pLastNode = m_pCells + m_iWidth * m_iHeight;
+    if (pNeighbour < m_pCells || pNeighbour >= pLastNode) {
+        return 0;
+    }
+
+    if (canTravel) {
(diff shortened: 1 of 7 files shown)


CorsixTH/Src/th_pathfind.cpp
```diff
@@ -530,14 +524,8 @@ void THPathfinder::pushResult(lua_State *L) const
...
-    if(m_iOpenCount == m_iOpenSize)
-    {
-        m_iOpenSize = (m_iOpenSize + 1) * 2;
-        m_ppOpenHeap = (node_t**)realloc(m_ppOpenHeap, sizeof(node_t*) * m_iOpenSize);
-    }
-    int i = m_iOpenCount++;
-    m_ppOpenHeap[i] = pNode;
-    pNode->open_idx = i;
+    m_openHeap.push_back(pNode);
+    pNode->open_idx = m_openHeap.size() - 1;
```
Alberth289346: move assignment to before the push_back? Gets rid of the -1

CorsixTH/Src/th_map.cpp
```diff
@@ -1226,6 +1226,43 @@ void THMap::setTemperatureDisplay(THMapTemperatureDisplay eTempDisplay)
...
 
+uint32_t THMap::thermalNeighbour(uint32_t &iNeighbourSum, bool canTravel, uint32_t relative_idx, THMapNode* pNode, int prevTemp) const
+{
+    int iNeighbourCount = 0;
+
+    THMapNode* pNeighbour = pNode + relative_idx;
+
+    // Ensure the neighbour is within the map bounds
+    THMapNode* pLastNode = m_pCells + m_iWidth * m_iHeight;
+    if (pNeighbour < m_pCells || pNeighbour > pLastNode) {
```
Alberth289346: for a 1x1 map, ``pNeightbour`` is allowed to be ``m_pCells[0]`` and ``m_pCells[1]``, which seems wrong

TheCycoONE:
Fixed the issues pointed out by @Alberth289346 

Added commit to rename the confusingly named pLastNode variable.

PR #1259, opened 2017-10-15
Use CMake to handle pre-built dependencies automatically 

DavidFair:
This PR adds an option to CMake to automatically pull in and use the pre-compiled dependencies. This is something I've been planning on doing since writing the instruction on how to compile for Windows.

- On MSVC this is on by default, 
- On GCC + Linux the option requires users to opt in and select the arch they intend to compile against.
- We use CMake to checkout a specific commit from the deps repo, determine the appropriate folder to use and finally update the path for find modules to use.

I've tested this against MSVC and it successfully builds as expected. Could somebody on Linux with GCC test this:
- Is it off by default, does the option correctly appear "USE_PRECOMPILED_DEPS"?
- Can you select the correct architecture with "DEPS_ARCH" when "USE_PRECOMPILED_DEPS" is selected?
- Does it work as expected (links against the pre-built libraries rather than system libs)?

Also - I wasn't sure about the projects licensing so went for MIT. Would it be worth mentioning this in the contribution guidelines?

Edit: I've noticed AppyVeyor also has its own logic for handling dependencies which this was inspired by. Ideally if everything works a future issue would be to have AppVeyor use this new option too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ CorsixTH/config.txt
 /Makefile
 /SDL
 /SDL_mixer-1.2.8
+/PrecompiledDeps
 /doc
 
 # For CorsixTH directory and Windows
--- a/CMake/PrecompiledDeps.cmake
+++ b/CMake/PrecompiledDeps.cmake
@@ -1,0 +1,98 @@
+# Clones and sets any dependencies up
+include(ExternalProject)
+    
+# Inform CMake about the external project
+set ( _DEPS_PROJECT_NAME PrecompiledDependencies)
+
+# Place files into ./precompiled_deps folder
+set( PRECOMPILED_DEPS_BASE_DIR ${PROJECT_SOURCE_DIR}/PrecompiledDeps CACHE PATH "Destination for pre-built dependencies")
+set( _DEPS_GIT_URL "https://github.com/CorsixTH/deps.git")
+# Select the optimal dependencies commit regardless where master is.
+set( _DEPS_GIT_SHA "a23eb28bb8998b93215eccf805ee5462d75a57f2" )
+
+
+ExternalProject_Add( ${_DEPS_PROJECT_NAME}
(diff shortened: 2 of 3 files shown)


TheCycoONE:
Hi @DavidFair 

Thank you for the contribution. The license is MIT contributed to the individual author so you guessed correctly. It is documented in a couple places, we could certainly add an instruction in the contributing guidelines about it.

Regarding the CorsixTH/deps package, that was my idea, but I have been the only one maintaining it, and after switching to visual studio 2017 I didn't cherish the thought of recompiling everything by hand again, so I'm looking to vcpkg now. You can see what I've done so far in #1247.

Now vcpkg can generate a redistributable of all the compiled packages, which I'm considering uploading to CorsixTH/deps for VS2017. I won't be maintaining multiple versions of MSVC deps, though I'm happy to accept them from someone else.

See: https://github.com/Microsoft/vcpkg

DavidFair:
I fully agree we shouldn't maintain multiple packages for old compilers. I think its worth dropping VS2010 support and leaving VS2015 up as long as it successfully compiles using the deps repo.

I hadn't come across vcpkg before, but it seems like a much better way of doing things. Could I propose a roadmap of changes, I'm completely open to feedback:

- Could we look at merging this PR before looking at vcpkg, it means for those of us developing on VS2015 it "just works"?

- I propose we make a new repository called something like "CorsixTH/vcpkg_deps"? I'm happy to look at building the dependencies for MSVC 2017 in a separate repo. However this may be the slowest part of the process as I have no previous experience pre-compiling libraries especially on MSVC.

- Once we have our own vcpkg dependencies built, would it be okay to move the changes in #1247 into this build script? This way users and CI servers both use the same code path and the folder is cloned relative to their repository and not to the C:\ drive (which causes me problems at least.).

- Lastly we could then look at enabling VS2017 support with these changes in place afterwards?

DavidFair:
Merged into #1261

Issue #593, opened 2015-01-02
Difficult camera controls on touch devices


Dinnerbone:
On devices with only touch as input (such as surface pro3 , windows 8.1) it's almost impossible to scroll the game around, and more so to zoom in and out. I'd propose that we take advantage of the touch API introduced in SDL2 and allow two finger gestures to move around + pinching to zoom.

Most other things in the game is usable, it's just the lack of camera controls which makes it fall short! I'm willing to investigate implementation if others agree it will be worthwhile to have (and if those gestures are enough).

Lego3:
Yes, it would certainly be worth it for SDL2. :-)

Alberth289346:
@Dinnerbone  is the issue solved now with sdl2?

TheCycoONE:
I don't think so, we didn't add any gesture support explicitly during the SDL2 transition.

wolfy1339:
@TheCycoOne is right here. I can say that on a similar device touch gestures dont work.

lewri:
May be resolved by PR #1260

TheCycoONE:
Yes, touch support is good now.

PR #1260, opened 2017-10-15
[RDY] Touch support for pinch zoom and move map

mugmuggy:
I have access to a convertable device that I have tested quickly at this stage addressing touch controls.

Both pinch zoom and move map which is working mostly okay with this code.

What I don't have is a device specifically like that mentioned in #593 which may have this behave entirely different or interfere generally with use so would be nice to get this tested elsewhere.

### DIFF
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -32,6 +32,12 @@ local TH = require "TH"
 -- earthquake with full intensity.
 local shake_screen_max_movement = 50 --pixels
 
+-- 0.002 is about 5 pixels on a 1920 pixel display
+local multigesture_pinch_sensitivity_factor = 0.002
+-- combined with the above, multiplying by 100 means minimum current_momentum.z for any detected pinch
+-- will result in a call to adjustZoom in the onTick method
+local multigesture_pinch_amplification_factor = 100
+
 --! Game UI constructor.
 --!param app (Application) Application object.
 --!param local_hospital Hospital to display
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/game_ui.lua
```diff
@@ -600,6 +601,38 @@ function GameUI:onMouseUp(code, x, y)
   return UI.onMouseUp(self, code, x, y)
 end
 
+function GameUI:onMultiGesture(numfingers, dTheta, dDist, x, y)
+  -- only deal with 2 finger events for now
+  if numfingers == 2 then
+    -- calculate magnitude of pinch
+    local mag = math.abs(dDist)
+    if mag > 0.002 then
```
TheCycoONE: Make this a variable (constant) defined near the top of the class. Let's avoid magic numbers.

CorsixTH/Lua/game_ui.lua
```diff
@@ -600,6 +601,38 @@ function GameUI:onMouseUp(code, x, y)
...
+      -- pinch action - constant needs to be tweaked
+      self.current_momentum.z = self.current_momentum.z + dDist * 100
```
TheCycoONE: 100 is another magic number, so put it in a variable too.

CorsixTH/Lua/game_ui.lua
```diff
@@ -649,6 +682,10 @@ function GameUI:onTick()
       self.current_momentum.z = self.current_momentum.z * self.momentum
       self.app.world:adjustZoom(self.current_momentum.z)
     end
+    if self.multigesturemove.x ~= 0.0 then
```
TheCycoONE: condition isn't necessary, just set them to 0.

CorsixTH/Lua/game_ui.lua
```diff
@@ -600,6 +601,38 @@ function GameUI:onMouseUp(code, x, y)
   return UI.onMouseUp(self, code, x, y)
 end
 
+function GameUI:onMultiGesture(numfingers, dTheta, dDist, x, y)
```
TheCycoONE: Please add documentation to new functions.

TheCycoONE:
I'd still like to see this one tested more before it lands. I'll see if I can get my hands on any devices. I think @wolfy1339 has one?

wolfy1339:
I don't have my Surface anymore. I now have a laptop with a touch screen (Razer Blade Stealth)
________________________________

TheCycoONE:
I can't seem to generate those events on y linux chromebook, both the touchscreen and the touchpad only make normal mouse events.

@wolfy1339 testing even with the razer blade would be helpful.

wolfy1339:
👍 📦

PR #1261, opened 2017-10-15
[HOLD] Add script to build and package libs in MS vcpkg

DavidFair:
This PR adds a new powershell script in the CMake folder (if there is a more appropriate place please let me know) that automatically updates / packages the required libs in a vcpkg.

It follows on from ideas in #1259 that @TheCycoONE was investigating.

When the script is run it does the following:
- Clones the vcpkg repo (or updates it if it already exists) in the current folder
- Bootstraps and builds vcpkg
- Selects the correct libraries for CorsixTH and builds them locally
- Packages the libraries in 7zip file and moves it to where the script was run
- Gives the packages sane names.

As VS2017 is ABI compatible with 2015 I recommend we build dependencies in 2017 from now on and link both 2015/17 against these libraries. This should help us consolidate the various dependencies we currently have in CorsixTH/deps.


TheCycoONE:
I like having this scripted!

The cmake folder isn't appropriate - that should just be cmake modules. In my opinion 'scripts' is the best folder, an argument could be made for the 'build_files' folder as well.

Your script covers the build time dependencies for CorsixTH. There are a couple runtime dependencies which I have been successful in submitting to upstream vcpkg (luafilesystem and lpeg) and one which I haven't attempted to handle yet. luasocket is going to be tricky - there isn't a release (or release candidate) version that works, so we have to pick a semi-recent git commit instead; which I think that upstream wouldn't accept.

There is also wxwidgets needed for building AnimView

It is better to pull from a specific vcpkg commit or release; or from my fork (or a CorsixTH organization fork), because new releases will break our builds until we can fix compatibility.

The export could be optional, it's not necessary for local builds.

For building the lua documentation, it is useful to additionally copy lfs.dll and lpeg.dll into the vcpkg /tools directories (where lua.exe and luac.exe are placed)

Perhaps the script could write-out the cmake line to use the vcpkg, or a link to the embedded documentation: https://github.com/Microsoft/vcpkg/blob/master/docs/users/integration.md#cmake-toolchain-file-recommended-for-open-source-cmake-projects ?

DavidFair:
> The cmake folder isn't appropriate - that should just be cmake modules. In my opinion 'scripts' is the best folder, an argument could be made for the 'build_files' folder as well.

Okay, I'll put this in scripts as build_files is picked up by .gitignore with build*

> There are a couple runtime dependencies which I have been successful in submitting to upstream vcpkg (luafilesystem and lpeg)

I didn't see them listed in the wiki as dependencies. I'll add them too and update the wiki.

> luasocket is going to be tricky - there isn't a release (or release candidate) version that works, so we have to pick a semi-recent git commit instead; which I think that upstream wouldn't accept.

It looks like development has stopped and nobody has an active fork. Does the vcpkg team have a policy on the state of a library before it is accepted?

> There is also wxwidgets needed for building AnimView

> The export could be optional, it's not necessary for local builds.

Originally my plan was to build all dependencies and distribute them pre-compiled like we currently do with the dependencies. However the final size of the .7z archive is 80MB so I prefer your approach of getting users to build locally rather than forcing them to download the exported libs. 

I'll adapt the script so it integrates with CMake and allows us to specify if AnimView is being built. This should allow us to compile only what users need instead of everything.

>It is better to pull from a specific vcpkg commit or release; or from my fork (or a CorsixTH organization fork)

II'm going to leave it pointing at the official repository and include checking out a specific commit. We can just change the commit we point to instead of having to pull the latest changes to test.

>Perhaps the script could write-out the cmake line to use the vcpkg[...]

If CMake invokes the script we could easily glob the final path and handle everything within the build script. 

My only other concern is CI servers, I'm reluctant to force them to rebuild the same dependencies each time. If AppVeyor/Travis allows us to preserve files between builds we could save the current commit we use for vpckg and export the dependencies. This way we can easily determine if we need to rebuild or just extract and reuse previously built ones?

TheCycoONE:
> > The cmake folder isn't appropriate - that should just be cmake modules. In my opinion 'scripts' is the best folder, an argument could be made for the 'build_files' folder as well. 
>Okay, I'll put this in scripts as build_files is picked up by .gitignore with build*
Good

> > There are a couple runtime dependencies which I have been successful in submitting to upstream vcpkg (luafilesystem and lpeg) 
> I didn't see them listed in the wiki as dependencies. I'll add them too and update the wiki.
Thanks - it's on all of the wiki pages I checked. Which wiki page are you reading?

> > luasocket is going to be tricky - there isn't a release (or release candidate) version that works, so we have to pick a semi-recent git commit instead; which I think that upstream wouldn't accept.
> It looks like development has stopped and nobody has an active fork. Does the vcpkg team have a policy on the state of a library before it is accepted?
The people that work on vcpkg prefer to use upstream releases with the minimal number of patches for compatibility (and they prefer those patches get merged upstream.) I'm not aware of another package on it in a situation like luasocket. As far as I know luasocket is not dead, but has received very little attention over the past number of years. I would like to drop luasocket in favour of implementing the functionality through libcurl, but that is out of scope for now.

> > There is also wxwidgets needed for building AnimView
> The export could be optional, it's not necessary for local builds.
Sounds good.

> Originally my plan was to build all dependencies and distribute them pre-compiled like we currently do with the dependencies. However the final size of the .7z archive is 80MB so I prefer your approach of getting users to build locally rather than forcing them to download the exported libs.
If we maintain our own vcpkg ports we could (optionally) dramatically reduce that by controlling the ffmpeg compile flags, similar to what I did with `CorsixTH/deps`
 
> I'll adapt the script so it integrates with CMake and allows us to specify if AnimView is being built. This should allow us to compile only what users need instead of everything.
Good

> > It is better to pull from a specific vcpkg commit or release; or from my fork (or a CorsixTH organization fork)
> II'm going to leave it pointing at the official repository and include checking out a specific commit. We can just change the commit we point to instead of having to pull the latest changes to test.
I'm feeling fairly strongly that we will be adding a CorsixTH\vcpkg fork so that we can maintain anything we need that deviates from upstream. It will be easy to swap though.

> > Perhaps the script could write-out the cmake line to use the vcpkg[...]
> If CMake invokes the script we could easily glob the final path and handle everything within the build script.
I'm interested in seeing how you do this.
 
> My only other concern is CI servers, I'm reluctant to force them to rebuild the same dependencies each time. If AppVeyor/Travis allows us to preserve files between builds we could save the current commit we use for vpckg and export the dependencies. This way we can easily determine if we need to rebuild or just extract and reuse previously built ones?
Both AppVeyor and Travis have the concept of dependency caching, so we can avoid rebuilding every time.

CMakeLists.txt
```diff
@@ -28,6 +28,14 @@ ENDIF(MINGW)
 INCLUDE(CheckIncludeFiles)
 SET(CORSIX_TH_DONE_TOP_LEVEL_CMAKE ON)
 
+
+# Dependency management
+if ( MSVC )
+OPTION(USE_VCPKG_DEPS "Build vcpkg dependencies locally" ON)
```
TheCycoONE: indent

CMakeLists.txt
```diff
@@ -90,8 +98,16 @@ ELSE()
   SET(CORSIX_TH_USE_VLD OFF)
 ENDIF(MSVC)
 
-# Environment handling
+# Get precompiled dependencies before running the various find modules
+if ( USE_PRECOMPILED_DEPS )
+  message("Note: Using precompiled dependencies.")
+  include( PrecompiledDeps )
+elseif ( USE_VCPKG_DEPS)
+  message("Note: Using locally build vcpkg dependencies.")
```
TheCycoONE: s/build/built

DavidFair:
I've merge the PR from #1259 into this. 

Here is a list of the work I've done since last time:
- The script no longer has an option to export, we just build locally
- I've integrated it into the CMake script fully so this is completed automatically
- I looked into using a toolchain and went with using `CMAKE_PREFIX_PATH` (reasons below)

Regarding the toolchain to copy and paste my comment from the file:

```
    # We cannot use a toolchain file at this point despite it being recommended by MS.
    # The arch is determined by the generator in use on Windows such as 
    # Visual Studio xx Win64 <= 64 bit build. If we use a toolchain this
    # always defaults to a 32 bit build.
    
    # Also we would need to restart the build at this point, delete the 
    # cache and invoke cmake again with the toolchain set if the user
    # specified their own generator as it is too late at this point.
```
I spent several hours looking for a workaround but it rapidly descends into dirty hacks. We can however use the `CMAKE_PREFIX_PATH` which performs the same functionality for finding libraries as we previously had.

> I'm feeling fairly strongly that we will be adding a CorsixTH\vcpkg fork so that we can maintain anything we need that deviates from upstream. It will be easy to swap though.

Completely forgot about advantages of a fork such as patching our own dependencies. I'll change where we point to if you want to set-up a fork?

TheCycoONE:
Ok, there is now a `CorsixTH/vcpkg`. For the moment is a straight fork. I read your changes after writing my last response - CMAKE_PREFIX_PATH sounds like a good way to go.

TheCycoONE:
The other thing that vcpkg toolchain does, which would be nice to have, is copy the dlls to the program output directory: https://github.com/Microsoft/vcpkg/blob/master/scripts/buildsystems/vcpkg.cmake#L126

We could do something more specific to CorsixTH that also copied `lfs.dll` and `lpeg.dll`.

DavidFair:
Okay, I'll add a section to the powershell script (as it has a list of libraries for MSVC) that copies the debug and release binaries to the relevant locations.

Is it worth me also adding a small section to the CMake script which copies the various files / folders we require into the build folder? This will save us having to manually do it each time we have a fresh build folder.

I've also noticed in the `Corsixth/CmakeLists.txt` file we do not specify debug and optimized libraries for `target_link_libraries`. Currently we just link against release libraries regardless of whether we are in debug or release. This is something I'll look into in the future as the change should be relatively easy but beyond the scope of this PR.

TheCycoONE:
For now I suggest end users make CorsixTH in a subdirectory of the checkout path, which is why build* is ignored. The program correctly finds CorsixTH.lua and the lua/ directories in this configuration.

I think it's out of scope for this PR, but config.h.in should be given the destination path to CorsixTH.lua and lua; and cmake install should copy them out to the appropriate path e.g. to the output directory on windows and <prefix>/share/corsix-th on Linux.

DavidFair:
Ahh I do my builds out of source hence my different perspective.

My original plan was to use the library names to guess the DLLs to copy over, however ffmpeg doesn't following this convention. Ultimately I took the kitchen sink approach by copying over all DLLs and the contents of the tools folder. These are placed into the correct Release or Debug build folders.

scripts/build_vcpkg_deps.ps1
```diff
@@ -124,6 +129,32 @@ function run_script {
...
+    # the CMake console
+    if (-Not (Test-Path -Path $debug_build_path)){
+        New-Item -ItemType Directory -Force -Path $debug_build_path | Out-Null
+    }
+    if (-Not (Test-Path -Path $release_build_path)){
+        New-Item -ItemType Directory -Force -Path $release_build_path | Out-Null
+    }
+
+    # Copy all tools to the output folders
+    Copy-Item -Force -Path ".\tools\*" -Destination $debug_build_path
```
TheCycoONE: We don't copy `tools` to the output path. The only thing we need `tools` for is LUA_PROGRAM_PATH to run LDocGen. We also need lfs.dll and lpeg.dll in the `tools` directory for that to work (or we copy everything out to another directory like I did in my appveyor PR)

DavidFair: So just to clarify we don't need top copy any files from the tools directory out? 
The earlier changes copied `lfs.dll` and `lpeg.dll` into the tools directory and I was unsure if they needed copying into the build directory.

TheCycoONE: `lfs.dll` and `lpeg.dll` need to be in the output directory; they should still be in the `bin` folder though. The tools folder only contains `lua.exe` and `luac.exe`, which we don't need to distribute.

TheCycoONE:
Copying all dlls (and maybe all pdbs for debug builds?) from the `bin` folders seems line a fine approach to me.

I like the approach in the accepted answer for https://stackoverflow.com/questions/10671916/how-to-copy-dll-files-into-the-same-folder-as-the-executable-using-cmake since it puts it in the correct directory regardless of the project configuration.

CMake/CopyVcpkgDlls.cmake
```diff
@@ -0,0 +1,16 @@
+# Add an extra step to copy built DLLs on MSVC
+IF (MSVC AND USE_VCPKG_DEPS)
+# Have to do this in two custom steps as we have some DLLs in a top level
+# and some in a debug folder so the $<CONFIGURATION> flag won't help
+
+    add_custom_command(TARGET Copy_Release_DLLS CorsixTH
```
TheCycoONE: Haven't tried it yet, but these don't look right: https://cmake.org/cmake/help/v3.2/command/add_custom_command.html#build-events

Disclaimer, I haven't tested it, but I think it should read like:
```
add_custom_command(
  TARGET CorsixTH POST_BUILD
  COMMAND ${CMAKE_COMMAND} -E copy_directory
    "${VCPKG_INSTALLED_PATH}/<$CONFIGURATION:Debug:debug/>bin"
    $<TARGET_FILE_DIR:CorsixTH>)
```

TheCycoONE:
Release build is working, the debug build is not. It seems that some libraries SDL2 and freetype in particular are given a 'd' suffix for their debug dlls, but CMake is using the release libs which don't match that name. It's annoying but we might be better off just respecting CMAKE_BUILD_TYPE up front instead of the CONFIG at build time.

I also have concerns about updating vcpkgs packages (iirc the old versions need to be manually removed, but we don't want to do that every time, so we'll have to check for updates), and maybe how to handle builds that require old versions (e.g. release tags), though I'm willing to ignore them for now.

TheCycoONE:
With respect to my above comment, it seems debug builds are broken even with CMAKE_BUILD_TYPE=debug. In the vcpkg toolchain file, they put the debug path before the non-debug path on debug (and unspecified) builds: https://github.com/Microsoft/vcpkg/blob/master/scripts/buildsystems/vcpkg.cmake#L73

DavidFair:
> Release build is working, the debug build is not. It seems that some libraries SDL2 and freetype in particular are given a 'd' suffix for their debug dlls, but CMake is using the release libs which don't match that name. It's annoying but we might be better off just respecting CMAKE_BUILD_TYPE up front instead of the CONFIG at build time.

I've added the changes from the link you suggested. SDL2 was giving some issues, it either found the debug OR release. In the end I tweaked the find module to search for both the debug and release version and package them together in the same variable.
This now links correctly on my end, does it resolve the problems you were seeing?

> I also have concerns about updating vcpkgs packages (iirc the old versions need to be manually removed, but we don't want to do that every time, so we'll have to check for updates), and maybe how to handle builds that require old versions (e.g. release tags), though I'm willing to ignore them for now.

The script adds the commit id for the last build to a file within vcpkg. Using this file we can see if we have checked out forwards or backwards. When we checkout backwards we run `git clean -fx` and build from scratch (I don't foresee this happening many times hence the nuclear approach).
In rare cases like this: https://github.com/Microsoft/vcpkg/issues/501 where `vpckg update` couldn't determine the correct behaviour the user will have to clean or delete the vcpkg folder. 
However I want to put the mechanism for that in when we have to otherwise it forces a clean rebuild each time.

I've also moved the `VCPKG_COMMIT_SHA` into the cache so users can see it. From the GUI it only appears under the advanced tab. This should allow us to test new commits without having to change the CMake file until were ready.

scripts/build_vcpkg_deps.ps1
```diff
@@ -0,0 +1,173 @@
...
+            # Git did not recognise this commit id. We have gone backwards
+            # completely clean before continuing
+            Write-Output "Dependencies has been rolled back. Cleaning for safety."
+            run_command "git clean -fx"
+            # Bootstrap again
+            run_command ".\bootstrap-vcpkg.bat"
+        } else {
+            # We have moved forwards, we can use normal mechanism
+            Write-Output "Dependencies have been updated."
+            run_command ".\vcpkg update"
```
TheCycoONE: Unfortunately at this time `vcpkg update` doesn't actually perform an update, it just finds a list of ports that have updates and tells you to remove and install them again. I don't know how fancy you want to get here.

scripts/build_vcpkg_deps.ps1
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    # Update the SHA we last saw
+    Set-Content -Path $commit_id_filename -Value $VcpkgCommitSha
+
+    # Always make sure we are using the latest files
+    run_command ".\vcpkg update"
```
TheCycoONE: Redundant?

scripts/build_vcpkg_deps.ps1
DavidFair: Okay lets save ourselves lots of headaches and just always clean build dependencies if we change the commit that we point to.

TheCycoONE: I think you would also be ok with `./vcpkg remove --outdated` then sending the whole list to install. On the train though so I can't test.

TheCycoONE: Checking the vcpkg source it sadly looks like this isn't allowed. So sure, clean and rebuild.

ras0219-msft: I don't have total context, but hopefully our design is that you can simply do
```
./vcpkg remove --outdated --recurse
./vcpkg install <list of all dependencies>
```
Any deps that are still installed will be a no-op, any deps that got removed will be installed.

ras0219-msft:
> Does the vcpkg team have a policy on the state of a library before it is accepted?

We want libraries that are usable by others :) For mature libraries, this means latest stable release, for immature libraries it's whatever the submitter wants. For unmaintained libraries, the caveat is that we _will_ break it by updating its dependencies, and if nobody fixes it it'll get removed.

scripts/build_vcpkg_deps.ps1
DavidFair: @ras0219-msft - Would these commands also work if we rolled back to a specific commit or would we have to take the clean build approach?

ras0219-msft: They will still work, though you should re-bootstrap the tool as well.
```
.\bootstrap-vcpkg.bat
```
We treat versions as "cookies" and don't try to say less-than/greater-than. If it's different, it's wrong :).

DavidFair: Great thanks for the heads up, this makes it significantly easier. 

Could we add this to the documentation in the vcpkg repository? 
The only details I could find about updating was [here](https://github.com/Microsoft/vcpkg/blob/master/docs/about/faq.md#how-do-i-update-libraries). Beyond this I couldn't find details that you've provided such as re-running the bootstrap script or how versions are handled when I was looking.

TheCycoONE:
@ras0219-msft What's the version policy if a package targets a git commit instead of an upstream release?

TheCycoONE:
Getting close now. The release configuration runs properly. The debug configuration still depends on the release SDL2.dll both directly (as well as the debug SDL2d.dll), and indirectly due to an upstream issue: https://github.com/Microsoft/vcpkg/issues/1951

ras0219-msft:
The sdl2 issue should be fixed in https://github.com/Microsoft/vcpkg/tree/deee8c17.

TheCycoONE:
I've synced CorsixTH/vcpkg with the latest upstream so we should be able to test. I'll run it tonight if no one beats me too it.

@DavidFair 
We will also need to consider how to handle the share folder resources. I'll put some thoughts on that later.

TheCycoONE:
Sorry I didn't post here after I tested yesterday. There is still an issue upstream with the smpeg2 package, and as I mentioned, our own project appears to end up depending on both SDL2 and SDL2d directly in Debug. I don't know why, but if you have no idea either I will try to do some reading and play around. (Multi-config cmake is new to me)

TheCycoONE:
Using vcpkg commit 23702360ce912b1827afe5559de936cc9a24cd0c (which I have pushed to CorsixTH/vcpkg), all transitive dependencies on SDL2 are fixed. We still have a direct dependency which I need to look into further.

TheCycoONE:
Ok, I got it working.

The solution is to skip the FindSDL2 module all together (evidently Find* modules are somewhat of a deprecated concept in CMake, replaced by Config files).

What you need is the following two lines:
```
FIND_PACKAGE(SDL2 CONFIG REQUIRED)
TARGET_LINK_LIBRARIES(CorsixTH SDL2::SDL2)
```

You also need to skip the normal Find SDL2 block. Since I don't know how prevalent SDL2 config modules are yet, it's probably best to do that conditionally on vcpkg usage.

TheCycoONE:
@DavidFair are you able to continue on this?

DavidFair:
@TheCycoONE Sorry I don't like leaving PRs in a half completed state but something came up on my end which I had to deal with.
Thanks for continuing with these changes in #1274, I'll close this PR which is now stale.

TheCycoONE:
No problem, thank you for getting the ball rolling on this.

Issue #1218, opened 2017-05-03
'Folders' menu renders values with default font


tiriana:
Despite I have custom font set, values of options in 'folders' menu are rendered with default font:

#### Steps to Reproduce
1. Set custom font
2. Enter Main menu -> settings -> Folders 

#### Actual Behavior
Values of settings are rendered with default font

#### Expected Behavior
Values of settings are rendered with custom font

#### System Information
**CorsixTH Version:** *0.60*

**Operating System:** *Ubuntu 16.04*

**Theme Hospital Version:** *CD*

#### Gamelog.txt
// file is empty
```
```

### Additional Info
![image](https://cloud.githubusercontent.com/assets/5216143/25673756/b2b428c4-3038-11e7-8a38-4c807eaec45f.png)

wolfy1339:
Could you post the values in English

tiriana:
@wolfy1339 

![image](https://cloud.githubusercontent.com/assets/5216143/25693766/6b841932-30ab-11e7-9b80-5db36cf4ef8e.png)

TheCycoONE:
By the default font, you mean whatever is specified in `config.txt` or always arialuni?

tiriana:
I mean that the values (on the right) are rendered by the orginial font despite unicode_font is specified.

mugmuggy:
Option labels are set with no font specified to the lower level functions
:setLabel(_S.folders_window.screenshots_label):setTooltip(_S.tooltip.folders_window.screenshots_location)

But the values are set with the 'built_in' font on the values
local built_in = app.gfx:loadBuiltinFont()

self.screenshots_panel:setLabel(app.config.screenshots and app.config.screenshots or tooltip_screenshots, **built_in**):setAutoClip(true)

Removing built_in gives this
![font](https://user-images.githubusercontent.com/8009292/31276660-698994cc-aae0-11e7-94e9-589d1246e75d.png)

You can see the effect by searching ', built_in' and replace with nothing in folder_settings.lua and then hitting the 'X' on the screen (this resets the font label in say resetScreenshotDir back to the built_in

TheCycoONE:
@mugmuggy Good find! I don't think we want to be using the built_in here - or anywhere? Would you mind making a PR?

Lego3:
Note that the built in font is used here because paths are incomprehensible in the original Th font. It is missing slashes and colon.

mugmuggy:
[...]

PR #1262, opened 2017-10-17
[RDY] 'Folders' menu renders values with default font

mugmuggy:
Refers to a solution for #1218 
Using the language font on both the labels and fields when its an option.
Using builtin when it isn't

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/folder_settings.lua
+++ b/CorsixTH/Lua/dialogs/resizables/folder_settings.lua
@@ -61,7 +61,7 @@ function UIFolder:UIFolder(ui, mode)
     .lowered = true
 
   -- Location of original game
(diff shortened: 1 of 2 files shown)


TheCycoONE:
LGTM

Ping @Lego3

Lego3:
Not having updated and started CorsxiTH for a while, as long as you still
get the built-in font if you haven't specified any other font file it
should be fine.

TheCycoONE:
I don't want 0e2303c as part of this; I want to fix that issue at it's source.

TheCycoONE:
Cool, can you squish this up, or would you like me to do it?

mugmuggy:
Any preference to squash or drop the #1263 related commit and its revert. Otherwise you can cover it as part of the merges

TheCycoONE:
Merged

Issue #723, opened 2015-03-18
Lock mouse pointer in the window for better windowed mode play


leaveboss:
What is it you would like in the game?
I think there should be an option to lock the mouse inside the window like dota 2. 
I have a big screen so I have to play in the window mode. However the mouse keeps going outside the window and it is impossible to scroll on the window edge with mouse. I googled this issue, it seems that the mouse cursor was locked inside the window in the previous version, but updated in a newer version.
I like the feeling when I am playing dota 2, the mouse is locked inside the window to enable edge-scroll, which will secure fast control like over 250 APM when playing. Right now i have to use arrow keys to scroll in window mode, and my left hand is jumping around as playing a piano.

What version of the product are you using? On what operating system?
0.40 RC3

TheCycoONE:
I like this solution better than the one in #431 because it matches the experience people would have playing with dosbox, and other window mode games.  Needs a key combo to unlock the cursor obviously.

PR #1264, opened 2017-10-21
[RDY]Lock mouse to window via key combo

mugmuggy:
#723 suggests locking mouse to window. I've set this up to match the Dosbox default Ctrl-F10 to capture and release. Defaults to uncaptured.

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -443,6 +443,12 @@ void THRenderTarget::setBlueFilterActive(bool bActivate)
     m_bBlueFilterActive = bActivate;
 }
 
+// Actiate and Deactivate SDL function to capture mouse to window
+void THRenderTarget::setWindowGrab(bool bActivate)
+{
+    SDL_SetWindowGrab(m_pWindow, (SDL_bool)(bActivate ? 1: 0));
+}
+
 uint32_t THRenderTarget::mapColour(uint8_t iR, uint8_t iG, uint8_t iB)
 {
     return THPalette::packARGB(0xFF, iR, iG, iB);
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -446,7 +446,7 @@ void THRenderTarget::setBlueFilterActive(bool bActivate)
 // Actiate and Deactivate SDL function to capture mouse to window
 void THRenderTarget::setWindowGrab(bool bActivate)
 {
-    SDL_SetWindowGrab(m_pWindow, (SDL_bool)(bActivate ? 1: 0));
+    SDL_SetWindowGrab(m_pWindow, bActivate ? SDL_TRUE : SDL_FALSE);
 }
 
 uint32_t THRenderTarget::mapColour(uint8_t iR, uint8_t iG, uint8_t iB)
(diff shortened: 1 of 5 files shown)


CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -443,6 +443,12 @@ void THRenderTarget::setBlueFilterActive(bool bActivate)
     m_bBlueFilterActive = bActivate;
 }
 
+// Actiate and Deactivate SDL function to capture mouse to window
+void THRenderTarget::setWindowGrab(bool bActivate)
+{
+    SDL_SetWindowGrab(m_pWindow, (SDL_bool)(bActivate ? 1: 0));
```
TheCycoONE: Prefer c++ style casts.
In this case though use `SDL_TRUE` and `SDL_FALSE` in the expression instead of casting.

CorsixTH/Src/th_gfx_sdl.h
```diff
@@ -194,6 +194,9 @@ class THRenderTarget
 
     //! Set the window caption
     void setCaption(const char* sCaption);
+	
+	//! Toggle mouse capture on the window.
```
TheCycoONE: Indentation  issue

CorsixTH/Src/th_lua_gfx.cpp
```diff
@@ -781,6 +781,14 @@ static int l_surface_get_renderer_details(lua_State *L)
     return 1;
 }
 
+// Lua to THRenderTarget->setWindowGrab
+static int l_surface_set_capture_mouse(lua_State *L)
+{
+    THRenderTarget* pCanvas = luaT_testuserdata<THRenderTarget>(L);
+    pCanvas->setWindowGrab((lua_isnoneornil(L, 2) != 0) ? false : (lua_toboolean(L, 2) != 0));
+    return 1;
```
TheCycoONE: Return the number of values you are pushing back to lua. In this case 0.

CorsixTH/Src/th_lua_gfx.cpp
```diff
@@ -942,6 +950,7 @@ void THLuaRegisterGfx(const THLuaRegisterState_t *pState)
     luaT_setfunction(l_surface_scale, "scale");
     luaT_setfunction(l_surface_set_caption, "setCaption");
     luaT_setfunction(l_surface_get_renderer_details, "getRendererDetails");
+	luaT_setfunction(l_surface_set_capture_mouse, "setCaptureMouse");
```
TheCycoONE: Indentation  issue

CorsixTH/Src/th_lua_gfx.cpp
mugmuggy: I'm not up to scratch on lua/c integration so I took my lead from an existing routine (l_surface_set_blue_filter_active) which is another void method call with same argument type, not doing a push back to lua and returns 1. I see others that have a lua_settop(L,1) return 1 when there is no explicit push. Maybe that is another bug.

TheCycoONE: Shouldn't be a bug, it will just pass the top of the stack back as the return value, which is useless.

Moving the top of the stack to 1 and returning 1 might be to return self, for a fluent interface.

TheCycoONE:
Merged.

An obvious future enhancement would be to add a default capture mouse option to the config file.

Issue #1105, opened 2016-06-29
Handyman stuck in Toilet, blocking patient


maugsburger:
#### Describe the issue

There is a handyman stuck in the toilet room, blocking  the patient washing her hands from moving.

#### Steps to Reproduce
1. *Load save game attached to this ticket*

2. Try to remove patient or handyman from the room.

3. Game becomes unusable (eg pickup will block mouse forever)

#### Save Game
autosave6 is the latest before the problem, in autosave7 it already happened. The handyman right behind the woman on the toilet doesn't move, neither pickup directly nor from the staff details popup works.

I didn't manage to reproduce it by waiting after autosave6.

#### Expected Behavior

- The Handyman leaves the room
- I can fire the handyman, send the patient home, and either will move.

#### System Information
**CorsixTH Version:** 0.60

**Operating System:** Windows 7 64bit

**Theme Hospital Version:** GOG.com

#### Gamelog.txt
```
Created game with savegame version 111.
Error in timer handler: 
C:\Program Files\CorsixTH\Lua\class.lua:136: Suspected infinite loop
stack traceback:
	[C]: in function 'error'
	C:\Program Files\CorsixTH\Lua\app.lua:919: in hook '?'
	C:\Program Files\CorsixTH\Lua\class.lua:136: in field 'is'
	C:\Program Files\CorsixTH\Lua\entity_map.lua:54: in method 'addEntity'
	C:\Program Files\CorsixTH\Lua\entity.lua:133: in field 'setTile'
	C:\Program Files\CorsixTH\Lua\entities\patient.lua:1057: in method 'setTile'
	C:\Program Files\CorsixTH\Lua\entity.lua:174: in method 'setTilePositionSpeed'
	C:\Program Files\CorsixTH\Lua\humanoid_actions\walk.lua:176: in function <C:\Program Files\CorsixTH\Lua\humanoid_actions\walk.lua:137>
	(...tail calls...)
	C:\Program Files\CorsixTH\Lua\entities\humanoid.lua:580: in upvalue 'Humanoid_startAction'
	C:\Program Files\CorsixTH\Lua\entities\humanoid.lua:672: in method 'finishAction'
	...ogram Files\CorsixTH\Lua\humanoid_actions\use_object.lua:325: in local 'timer_function'
	C:\Program Files\CorsixTH\Lua\entity.lua:219: in method 'tick'
	C:\Program Files\CorsixTH\Lua\world.lua:1075: in method 'onTick'
[...]

PR #1270, opened 2017-10-28
[RDY] Pathfinding broken by side objects

mugmuggy:
A proposed resolution to #1105 and related issues where pathfinding breaks from placement of side objects.

afterLoad in map and world to call their required methods to fix pathfinding on old saves.

To prevent the issue occuring, added a check and calls to those same functions in Object:onDestroy after the super class method is called when any side objects are moved.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local runDebugger = dofile "run_debugger"
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 117
+local SAVEGAME_VERSION = 120
 
 class "App"
 
--- a/CorsixTH/Lua/dialogs/place_objects.lua
+++ b/CorsixTH/Lua/dialogs/place_objects.lua
@@ -666,7 +666,9 @@ function UIPlaceObjects:setBlueprintCell(x, y)
         if not tile.shareable and is_object_allowed then
           -- Check 4: only one object per tile allowed original TH
           -- can build on litter and unoccupied tiles and only placeable if not on another objects passable footprint unless that too is a shareable tile
-          is_object_allowed = world:isTileExclusivelyPassable(xpos, ypos, 10)
+          local objchk = map:getCellFlags(xpos, ypos, flags)["thob"]
+          is_object_allowed = objchk == 0 or objchk == 62 or objchk == 64 -- no object, litter/puke, ratholes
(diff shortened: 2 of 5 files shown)


mugmuggy:
This is ready, not sure why Appveyor idn't pick it up and report it back. But as this is a lua change only should be ok, to take the travis-ci as an indicator its passing.

MarkL1961:
@mugmuggy  I could be wrong, but I think you need to update app.lua to `local SAVEGAME_VERSION = 118` as the current save game is 117 (I got this from the last build on appveryor)
Otherwise your afterload won't work as intended.
@TheCycoONE might do this at the merge stage, as it can be difficult to keep track if merges are done out of order

TheCycoONE:
I'd rather the version was bumped here, if it's out of order I'll get a merge conflict and can fix it. If it's not there at all I could miss it.

TheCycoONE:
There are hard coded litter check at:
* https://github.com/CorsixTH/CorsixTH/blob/fa0ae5362fdf1ac115d84b817513225cb75686c8/CorsixTH/Lua/dialogs/edit_room.lua#L1150
* https://github.com/CorsixTH/CorsixTH/blob/fa0ae5362fdf1ac115d84b817513225cb75686c8/CorsixTH/Lua/world.lua#L2721

The first one is almost an identical check to what you're doing here (though it doesn't also check rat tiles, which may be a bug.) As well these are the only place isTileExclusivelyPassable is called on. I would prefer you remove the redundant checks, if you must keep checks for particular object numbers instead of flags, put them in one function (e.g. isTileExclusivelyPassable), and preferably use another object property besides it's thob.

mugmuggy:
Litter has no footprint, so getWalkableTiles will be called, so with this check in isTileExclusivelyPassable, I'm basically saying ignore checking litter - which really means check for doors in object placement and check doors for opposing entrances in room construction. Could guard the for loop with an if statement checking if the type is litter to make it a little more obvious that we don't want to do it.

Rathole has an empty footprint table so basically skips over the first loop in isTileExclusivlePassable and a door can't be placed on an existing wall so not required in edit_room.

The check in edit_room you've highlighted, is to prevent say the drink machine actual tile, blocking a door and object physically in front of it,  the isTileExclusivelyPassable checks only the passable tile of the drinks machine is also not blocking the door. And there is the same process in placing an object. In Place Object I check to ensure that the only object on the tile is currently litter/rathole and if so check passable tiles of nearby objects.

TheCycoONE:
Some day I hope for a cleaner fix for this.

PR #1274, opened 2017-11-05
[RDY] CMake deps handling - continuation of pr #1261

TheCycoONE:
Compared to #1261 this PR:

* Handles SDL2 on multiple configuration targets when using built in vcpkg, by using the provided sdl2 cmake config.
* Doesn't modify FindSDL2.cmake (those modifications didn't work)
* Copies lua files from the `vcpkg/installed/**/share/lua/*` directory
* Checks out the correct commit when cloning vcpkg the first time
* Checks out a newer vcpkg commit, containing luasockets and newer sdl packages
* Fixes whitespace
* Modifies appveyor to take advantage of these changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,9 @@ Release/
 MinSizeRel/
 RelWithDebInfo/
 
+# Ignore the folder where libraries are pre-compiled
+vcpkg/
+
 # Build folders
 /build*/
 
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ CorsixTH/config.txt
 /Makefile
 /SDL
 /SDL_mixer-1.2.8
+/PrecompiledDeps
 /doc
 
 # For CorsixTH directory and Windows
--- a/CMake/CopyVcpkgLua.cmake
+++ b/CMake/CopyVcpkgLua.cmake
@@ -1,0 +1,29 @@
+# Add an extra step to copy LUA files from vcpkg
+add_custom_command(TARGET CorsixTH POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy_directory
+  "${VCPKG_INSTALLED_PATH}/share/lua"
+  $<TARGET_FILE_DIR:CorsixTH>)
+
+add_custom_command(TARGET CorsixTH POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E remove
+  $<TARGET_FILE_DIR:CorsixTH>/COPYRIGHT)
+
+add_custom_command(TARGET CorsixTH POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy
+  "${VCPKG_INSTALLED_PATH}/$<$<CONFIG:Debug>:debug/>bin/lfs.dll"
+  $<TARGET_FILE_DIR:CorsixTH>/lfs.dll)
+
+add_custom_command(TARGET CorsixTH POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy
+  "${VCPKG_INSTALLED_PATH}/$<$<CONFIG:Debug>:debug/>bin/lpeg.dll"
+  $<TARGET_FILE_DIR:CorsixTH>/lpeg.dll)
+
+add_custom_command(TARGET CorsixTH POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy
+  "${VCPKG_INSTALLED_PATH}/$<$<CONFIG:Debug>:debug/>bin/mime/core.dll"
+  $<TARGET_FILE_DIR:CorsixTH>/mime/core.dll)
+
+add_custom_command(TARGET CorsixTH POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy
+  "${VCPKG_INSTALLED_PATH}/$<$<CONFIG:Debug>:debug/>bin/socket/core.dll"
+  $<TARGET_FILE_DIR:CorsixTH>/socket/core.dll)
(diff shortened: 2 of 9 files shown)


CMake/CopyVcpkgDlls.cmake
```diff
@@ -0,0 +1,19 @@
+# Add an extra step to copy built DLLs on MSVC
```
ras0219-msft: We should do this automatically in Vcpkg!

Could you let me know what specifically isn't working?

ras0219-msft: Nevermind, I see below that you're not using the toolchain file.

ras0219-msft:
I'd like to help you use the toolchain file if possible!

Note that the toolchain file only appears to take effect during the first call to `project()`, so if you want to download the prebuilt libraries, you could do that first:
```cmake
# download stuff here
set(CMAKE_TOOLCHAIN_FILE /downloaded/path/to/vcpkg.cmake CACHE STRING "")

cmake_minimum_required(VERSION 3.8)
project(foo CXX)

message(STATUS "VCPKG_TARGET_TRIPLET=${VCPKG_TARGET_TRIPLET}")
```

Also, the architecture of the Visual Studio generator can be controlled with `-A x64`.

TheCycoONE:
@ras0219-msft I played for awhile tonight trying to use the CMAKE_TOOLCHAIN_FILE like you suggest, but so far haven't been able to turn it into a good experience.

I'll try again tomorrow, and document what I'm seeing then if I still can't get it to work.

TheCycoONE:
@ras0219-msft The first problem I run into with that approach is that MSVC isn't set before PROJECT, so I don't know if we are in an appropriate environment to use VCPKG.

TheCycoONE:
@ras0219-msft Disregard the above - I've gotten considerably closer.

Now the issue I run into when I use the vcpkg toolchain is:
```
Building CorsixTH
CMake Error at vcpkg/scripts/buildsystems/vcpkg.cmake:159 (_find_package):
  Could not find a package configuration file provided by "SDL2" with any of
  the following names:

    SDL2Config.cmake
    sdl2-config.cmake

  Add the installation prefix of "SDL2" to CMAKE_PREFIX_PATH or set
  "SDL2_DIR" to a directory containing one of the above files.  If "SDL2"
  provides a separate development package or SDK, be sure it has been
  installed.
Call Stack (most recent call first):
  CorsixTH/CMakeLists.txt:97 (FIND_PACKAGE)
```

Compared to the code I just committed, this was my changes:
```
diff --git a/CMake/VcpkgDeps.cmake b/CMake/VcpkgDeps.cmake
index 08933a19..1d778194 100644
--- a/CMake/VcpkgDeps.cmake
+++ b/CMake/VcpkgDeps.cmake
@@ -56,23 +56,4 @@ if(err_val)
       "\nIf this error persists try deleting the 'vcpkg' folder.\n")
 endif()

-# We cannot use a toolchain file at this point despite it being recommended by MS.
-# The arch is determined by the generator in use on Windows. For example
-# Visual Studio xx Win64 <= implies 64 bit build. If we use a toolchain this
-# always defaults to a 32 bit build.
-
-# If the user specified their own generator it is too late at this point.
-# We would need to restart CMake at this point, delete the
-# cache and invoke CMake again with the toolchain set.
-
-# For both of these reasons we will use CMAKE_PREFIX_PATH
-
-set(VCPKG_INSTALLED_PATH ${VCPKG_PARENT_DIR}/vcpkg/installed/${VCPKG_TARGET_TRIPLET})
-
-if(CMAKE_BUILD_TYPE MATCHES "^Debug$" OR NOT DEFINED CMAKE_BUILD_TYPE)
-  list(APPEND CMAKE_PREFIX_PATH ${VCPKG_INSTALLED_PATH}/debug)
-  list(APPEND CMAKE_LIBRARY_PATH ${VCPKG_INSTALLED_PATH}/debug/lib/manual-link)
-endif()
-
-list(APPEND CMAKE_PREFIX_PATH ${VCPKG_INSTALLED_PATH})
-list(APPEND CMAKE_LIBRARY_PATH ${_VCPKG_INSTALLED_DIR}/lib/manual-link)
+set(CMAKE_TOOLCHAIN_FILE ${VCPKG_PARENT_DIR}/vcpkg/scripts/buildsystems/vcpkg.cmake)
```

ras0219-msft:
It looks like SDL2 doesn't provide a `-config.cmake`, so you'll probably need to write a `FindSDL2.cmake` script manually (which you probably needed anyway?).

It looks like there are a few floating around that might be interesting: https://github.com/search?utf8=%E2%9C%93&q=filename%3Afindsdl2.cmake&type=Code

TheCycoONE:
It turned out that the VCPKG_TRIPLET was ending up as just `-windows`.

in vcpkg.cmake, if VCPKG_TARGET_TRIPLET is specified it skips setting `_VCPKG_TARGET_TRIPLET_ARCH`, then later in the file it sets `VCPKG_TARGET_TRIPLET` again. I don't know how this is normally handled when you use `-DVCPKG_TARGET_TRIPLET`, but when I was setting it in VcpkgDeps.cmake the ARCH was never set and then removed.

By avoiding setting VCPKG_TARGET_TRIPLET, it lets the toolchain detect it and set it properly.

TheCycoONE:
AppVeyor is broken because I haven't put `lfs.dll` in the lua cpath for lua.exe

TheCycoONE:
At @Alberth289346's request I've made the indentation for the two new CMake files internally consistent.

If no one else has anything to add, I'll squash these up and move this PR back to RDY tomorrow.

TheCycoONE:
All squished up and ready to go.

PR #1278, opened 2017-11-10
[RDY] Fix write to empty vector

TheCycoONE:
We were writing to an empty vector when poping the last element.


Alberth289346:
good find!

PR #1281, opened 2017-11-13
[RFC] Make WindowsInstaller test for 2017 Redistributable

TheCycoONE:
The Visual C++ Redistributable for Visual Studio 2017 is a bit different than
2015. The installed key doesn't tell us that we are using the right one because
both 2015 and 2017 install to the 14.0 path. 2017 is backwards compatible and
registers itself with the version 14.10.#####


wolfy1339:
It seems you reset all localization to English. Would it not be possible to only change 2015 to 2017 using a simple find and replace? As to not replace the localized text

TheCycoONE:
I didn't realise any of them were actually translated. It seems only French was, so I will restore it. The rest have the wrong English name, so replacing the number isn't sufficient.

TheCycoONE:
@Lego3 there is no good landing page anymore. You get it from here: https://www.visualstudio.com/downloads/ which I consider a little overwhelming.

You're right about the architecture issue. I used the 'bookmark' link from that page, which wasn't updated whether I selected x64 or x86. I thought it might be using some sort of host detection to serve the appropriate file but that looks not to be the case. I can fix it to link to the appropriate file; or we can make our own landing page and link there?

wolfy1339:
I did find this, it is some sort of a landing page: https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads

TheCycoONE:
@wolfy1339 that page also appears to only link to the x64 version?

wolfy1339:
Would this be what were looking for: https://www.visualstudio.com/downloads/#other

TheCycoONE:
Yeah, I think that's friendly enough for now. Certainly better than the 0.60 approach (nothing, except a note in the release notes).

Thanks @wolfy1339

TheCycoONE:
Pushed a couple changes to address the comments by @Lego3 and @wolfy1339 

Hopefully we can get more localizations done before the release.

TheCycoONE:
@Lego3 do you think you'll have a chance to look at my fixes this week?

Lego3:
Sorry, a little fast there on the merge. Should of course have squashed it. Getting rusty at this...

Issue #1155, opened 2016-11-09
Error removing litter


MarkL1961:
Error message by handyman near to reception desk

Whether this is reproducible is unknown, when I just tried and loaded the save again the error did not happen the second time.

Links to two saves below, one at the point the error is thrown and the second is the save I loaded just before the error.

Not sure if this is related and I doubt it is I loaded the save from yesterday and started to remove the plants.  The error was thrown just after this.  I had calls dispatcher open whilst removing plants so there were no pending calls to water

Created game with savegame version 115.
Warning: Trying to remove nonexistant room build callback (function: 00000250EAEE8DE0) from humanoid (table: 00000250ED377D30).
Warning: Trying to remove nonexistant room remove callback (function: 00000250EA0E4900) from humanoid (table: 00000250ED377D30).
Warning: Trying to remove nonexistant room build callback (function: 00000250E8310F90) from humanoid (table: 00000250E5ECB020).
Warning: Trying to remove nonexistant room remove callback (function: 00000250EB44C6D0) from humanoid (table: 00000250E5ECB020).
Warning: Trying to remove nonexistant room build callback (function: 00000250EF45AFF0) from humanoid (table: 00000250E8C2A050).
Warning: Trying to remove nonexistant room remove callback (function: 00000250E893FA30) from humanoid (table: 00000250E8C2A050).
Warning: Trying to remove nonexistant room build callback (function: 00000250EAEF64C0) from humanoid (table: 00000250E8EC3D60).
Warning: Trying to remove nonexistant room remove callback (function: 00000250ED6C7E40) from humanoid (table: 00000250E8EC3D60).
Warning: Trying to remove nonexistant room build callback (function: 00000250EAEF7180) from humanoid (table: 00000250EAE35E70).
Warning: Trying to remove nonexistant room remove callback (function: 00000250E8504AC0) from humanoid (table: 00000250EAE35E70).
Warning: Trying to remove nonexistant room build callback (function: 00000250EF456850) from humanoid (table: 00000250EAAFE8F0).
Warning: Trying to remove nonexistant room remove callback (function: 00000250ECBEF470) from humanoid (table: 00000250EAAFE8F0).
Warning: Trying to remove nonexistant room build callback (function: 00000250EA7D0950) from humanoid (table: 00000250E8A4E0B0).
Warning: Trying to remove nonexistant room remove callback (function: 00000250EAC6C820) from humanoid (table: 00000250E8A4E0B0).
[...]

PR #1286, opened 2017-11-25
[RDY] Remove handyman task upon sweep action start

mugmuggy:
Litter not 'tickable' as in to interrupt Handyman tasks when Litter:remove on build over.

I didn't update Litter:remove() to enforce the ticks = false, just left that to new litter only.

Removing the task in the action_sweep_floor_start is similar to Plant:restoreToFullHealth called from the end of action_use_object_start.
Hopefully addresses issues in #1155.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local runDebugger = dofile "run_debugger"
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 120
+local SAVEGAME_VERSION = 121
 
 class "App"
 
--- a/CorsixTH/Lua/entities/staff.lua
+++ b/CorsixTH/Lua/entities/staff.lua
@@ -900,6 +900,15 @@ function Staff:afterLoad(old, new)
     end
   end
 
+  if old < 121 then
+    if self.humanoid_class == "Handyman" and self.user_of and self.user_of.object_type.class == "Litter" then
+      local litter = self.user_of
+      local hospital = self.world:getHospital(litter.tile_x, litter.tile_y)
+      local taskIndex = hospital:getIndexOfTask(litter.tile_x, litter.tile_y, "cleaning", litter)
+      hospital:removeHandymanTask(taskIndex, "cleaning")
+    end
+  end
+
   Humanoid.afterLoad(self, old, new)
 end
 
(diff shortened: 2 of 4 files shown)


TheCycoONE:
I looked into the history of 
https://github.com/CorsixTH/CorsixTH/blob/f539713e462821df8f793166912a737e57fd953f/CorsixTH/Lua/hospital.lua#L2166 and found this curiosity:

```diff
index 94be3f83..c0bf692e 100644
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1513,7 +1513,7 @@ function Hospital:removeHandymanTask(taskIndex, taskType)
     local task = subTable[taskIndex]
     table.remove(subTable, taskIndex)
     if task.assignedHandyman then
-      if taskType ~= "cleaning" and task.object.ticks ~= true then
+      if task.object.ticks ~= true then
         task.assignedHandyman:intreruptHandymanTask()
       end
     end
```
Which effectively is the same thing that you are doing by disabling ticks on litter. There was no indication in the commit as to why they removed that test.

I don't like that the purpose of `ticks` was overridden like this - it's suppose to only indicate when we can call `enity.onTick()` - but the use of ticks for handyman tasks predates your PR by years so I will let it go in favour of a future refactor.

Otherwise, I was concerned about the action getting disrupted, and litter being abandoned, but it seems that the `must_happen` takes care of that.

I'm still concerned about the state deviation between old saves and new litter tasks. I means a code cleanup in the future could easily lead to broken state with little clue as to why. Should be able to fix that with an afterLoad on `staff.lua` that checks `humanoid_class`, and `user_of` and removes the task for any litter found?

mugmuggy:
For the ticks thing, I should have set it correctly as
`self.ticks = object_type.ticks` - which is false - like all other Objects do
https://github.com/CorsixTH/CorsixTH/blob/f539713e462821df8f793166912a737e57fd953f/CorsixTH/Lua/entities/object.lua#L40-L53

A quick dive into the history and the context I believe is the 'soot' in a crash room was invoking an addHandymanTask call in the litter constructor. The afterLoad version < 54 check was added to remove the existing tasks for soot in crashed rooms, calling removeHandymanTask and in doing so, to interrupt handyman assigned to clean soot in rooms, the `~= "cleaning"` was removed, with the upside it should fix the tasks, but alas I don't think it was due to the Entity constructor setting it ticks to true.

Yep currently `must_happen` prevents disrupting the task.

For cleaning up tasks already assigned, the minimum required is, but not necessarily providing coverage to potentially all edge cases litter afterLoad.
`  if old < 121 then`
`    self.ticks = object.ticks`
`  end`
As Litter:remove() still needs to call removeHandymanTask for when placing objects over litter, there is cleanup being done still there and that will remove the handyman task also. I'll look into the staff change to remove the task earlier as there could be edge cases I'm just not hitting.

TheCycoONE:
lgtm

TheCycoONE:
RDY @mugmuggy?

mugmuggy:
Yup

Issue #1285, opened 2017-11-21
Dying patient can receive invalid walk instructions


mugmuggy:
#### Describe the issue
In certain circumstance a dying patient at the end of a corridor, can receive an invalid walk action.

In the save in #1284 at (50,63) is a dying patient.

In action_die_start a randomly selected direction (south or east) is chosen for the fall. In this intance south is blocked. It also happens that, a humanoid was at the door of the pharmacy adjacent to the patient at the time the tile was checked, so neither South or East is chosen. The code then goes on to check the west and north tiles. The west tile first, here is checked, and has no objects/humanoids but is outside the hospital. The patient is then given a walk instruction to that west tile, that no path can be found to and becomes stuck.

The code will also behave incorrectly with objects (mainly side objects being a concern here) as isTileEmpty checks for humanoids and objects.

#### Steps to Reproduce
Unknown

#### Save Game
See #1284

#### Expected Behavior
Patient should either probably fall still to the east or find a free tile

#### Gamelog.txt
None

TheCycoONE:
I'm curious what the original does here if there is no valid walk tiles. I will test if no one beats me to it.

TheCycoONE:
I answered my question in the RFC; in short the original just allows the patient to fall through an invalid tile. There is a minor graphical issue (the patient falls through whatever was blocking them), but nothing game breaking for such a rare occurrence.

lewri:
May be resolved by PR #1288

PR #1288, opened 2017-11-26
[RDY] Dying patient walk instructions can lead to recurring invalid walk instructions

mugmuggy:
For #1285.
Still attempts the fall east/south randomly chosen, then should attempt the opposite. If neither then it should check west than north tiles to travel to so can fall east then south.

As patients/staff are wandering around some of these checks will fail still after walk actions, or immediately not be possible. So I have got a bit of a brute force search for tiles, and this can sometimes make the patients wander around a bit too much, especially in a single width corridor with patients passing through.

That brute force loop uses getAdjacentFreeTiles which has the issue in entity_map. Without that fix it will go in an infinite loop.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/die.lua
+++ b/CorsixTH/Lua/humanoid_actions/die.lua
@@ -264,42 +264,8 @@ local function action_die_start(action, humanoid)
   --If this isn't done their bald head will become bloated instead of suddenly having hair:
   if humanoid.disease.id == "baldness" then humanoid:setLayer(0,2) end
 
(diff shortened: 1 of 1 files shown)


TheCycoONE:
Avoid extremes, we still want to eventually support #826 which may mean that there is no free tile to fall in:

Screenshot from trapped patient dying in the original game:
hospital_005

mugmuggy:
I'll simplify this up to just drop dead in either direction where the patient is I think.

TheCycoONE:
This might be too much of a step back, but someone can improve it later if they think it's an issue.

Issue #1165, opened 2016-11-20
Death - is it working correctly?


MarkL1961:
Load the attached save, cancel buying items and focus on the pharmacies.
Last invis emergency enters, takes the cure and you hear the boo, advisor pops up and mentions that you have killed a patient.  The patient though shows the cured icon and is cured.  At this point pause the game and take a look round, there are not many places to be cured or die so it won't take long.
If it wasn't him who did die and why can't you see it?
I have also seen in another level where a patient falls down to die, gets up and walks out!  The above happens every time with this save, couldn't get a save to replicate the second observation though
https://www.dropbox.com/s/7nqjf23yew91hkg/Autosave2.sav?dl=0

mugmuggy:
Just took a look at this one. Kill 'em twice
https://github.com/CorsixTH/CorsixTH/blob/63afae1554a2abeba225d5865f5ea77b18e73c78/CorsixTH/Lua/hospital.lua#L1286-L1291

Already dead patients from emergency_patients will retrigger the patient:die() call upon end of emergency as they aren't in a room - like patients waiting to be treated which is the intent obviously here. Maybe we just need to check `not patient.dead` as well.

TheCycoONE:
Thanks @MarkL

PR #1305, opened 2017-12-10
[RDY] Kill patients only once during emergencies

mugmuggy:
Adds a check for patient.dead so Patient:die isn't called again as is happening in #1165


TheCycoONE:
Other people with contribute access, please hold off on merging this until after the release.

PR #1310, opened 2017-12-20
[RFC] Add mouse capture option to option menus

### LINKED ISSUE #1309: Add toggle option for mouse capture
In #1264 and #1280 logic was added to capture the mouse. Whilst this can be toggled with a keyboard shortcut (CTRL+F10) or in the configuration it is not an accessible option for novice users. 

Instead we should add another option called something like `Capture Mouse` in the main options window. Personally this was the first place I looked before having to look at the above PRs to disable mouse capture. 

I'm happy to look into this as the changes should be relatively simple. However I'm posting this issue if anybody has any concerns / feedback before the work is completed?

DavidFair:
This PR adds a new toggle to the options menu and in game tool-bar to toggle whether the mouse should be captured in the current window or not.

I've re-factored how the y positions of buttons in the options menu are calculated, instead of having to recalculate the Y position of all buttons when we add a new one. Instead we keep track internally of the y position and call a function which gives us a new co-ord and increments that counter.

I've adds three new strings which must be localised too:
- menu_options.capture_mouse
- options_window.capture_mouse
- tooltip.options_window.capture_mouse

I've also updated .gitignore with some additional files which should be ignored.

Fixes #1309 

----

To test:
- Start CorsixTH
- Go to options menu and check button is sane (e.g. is pressed on when the setting is on)
- Try toggling the option and check the you can move the mouse out of the window or cannot
- Check the configuration file is updated each time you click it
- Close and open the options window, is the state preserved
- Go into any level
- Go to the top toolbar and check that the state you left it in is preserved
- Check that if you toggle the option the mouse is correctly captured
- Check the config is saved each time you toggle it on or off


.gitignore
```diff
@@ -69,9 +69,26 @@ CorsixTH/CorsixTH.sln
...
+# Debugging related files
 CorsixTH/Lua/debug_script.lua
 CorsixTH/Lua/debugger.lua
+mobdebug.lua
 
 # For AnimView output
 VSPR-0.xml
 *export.log
+
+# CMake related files
```
TheCycoONE: Since you added this header, the top five items in this file are all CMake related as well.

CorsixTH/Lua/dialogs/menu.lua
```diff
@@ -636,12 +636,21 @@ function UIMenuBar:makeGameMenu(app)
   end
   options:appendCheckItem(_S.menu_options.lock_windows, boolean_runtime_config"lock_windows")
 
+  -- Edge Scrolling
   options:appendCheckItem(_S.menu_options.edge_scrolling,
     not app.config.prevent_edge_scrolling,
     function(item) app.config.prevent_edge_scrolling = not item.checked end,
```
TheCycoONE: Not relevant to your PR, but I just noticed this toggle doesn't save, when the others do.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -67,6 +86,13 @@ function UIOptions:UIOptions(ui, mode)
   self.default_button_sound = "selectx.wav"
   self.app = app
 
+  -- Tracks the current position of the object
+  self._current_option_index = 1
+
+  -- Constants for most button's width and height
+  local BTN_WIDTH = 135
```
TheCycoONE: +1

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -226,6 +273,14 @@ function UIOptions:buttonFullscreen(checked)
...
 end
 
+function UIOptions:buttonMouseCapture(checked)
+  local app = self.ui.app
+  app.config.capture_mouse = not app.config.capture_mouse
+  app:saveConfig()
+  app:setCaptureMouse()
+  self.mouse_capture_button:setLabel(app.config.capture_mouse and _S.options_window.option_on or _S.options_window.option_off)
+end
+
```
TheCycoONE: Remove this line if there are other changes, should be one empty line.

.gitignore
```diff
@@ -19,6 +19,9 @@ CorsixTH/config.txt
 *.exe
 Thumbs.db
 
+# For built binary
+CorsixTH/CorsixTH
```
TheCycoONE: I generally discourage building in the root source directory. Building out of source isn't well supported yet, but building in a build*/ subdirectory works really well.

CorsixTH/Lua/app.lua
```diff
@@ -472,6 +472,13 @@ function App:loadMainMenu(message)
   end
 end
 
+--! Sets the mouse capture to the state set within
+--! app.config.capture_mouse
+function App:setCaptureMouse()
+  self.video:setCaptureMouse(self.config.capture_mouse)
+end
+
+
```
TheCycoONE: 1 line

CorsixTH/Lua/dialogs/menu.lua
DavidFair: I'll open another issue to look into it. It should be minor but deserves it's own testing to be sure we have got it working properly

DavidFair: I've opened #1311 - I'll look into it another day as it looks like a simple job, but I suspect other buttons could also be broken too.

TheCycoONE: Thanks, I was going to do this but got sleepy 👍

TheCycoONE:
Code review LGTM
I'll play test this later today.

TheCycoONE:
Plays well to me too.
Ready to merge @DavidFair or did you want to fiddle with the history first?

DavidFair:
Let's merge as is. If we need to roll back the refactor we revert a single commit rather than all features at once.

Issue #789, opened 2015-07-11
Room queue and mouse over room


bobcanthelpyou:
![bildschirmfoto von 2015-07-11 10 49 11](https://cloud.githubusercontent.com/assets/4523264/8633130/dbd56d04-27ba-11e5-96e0-cff80b4a07c3.png)

When the room queue is opened it is sometimes over the room (mostly when it is a large room and the room is more or less in the center of the screen), now when i try to move a patient and i mouse over the room at the same time, the icons over the patients change. But this behavior is bad, when patients with unhappy smileys or skulls should be moved.

This only should happen when i mouse over the room directly and not when an other window like the room queue is between the mouse pointer and the room.

Not sure if the icons should change within the room queue when mouse over the room, but this also could be a feature to show 'this room queue belongs to a special room'.

mugmuggy:
GameUI:onCursorWorldPositionChange isn't hit testing the UI components for the queuing over room logic, pretty simple fix this one.

TheCycoONE:
Fixed by #1314

PR #1314, opened 2017-12-28
[RFC] Stored camera positions

mugmuggy:
When referring to the table here in [programming ideas](https://github.com/CorsixTH/CorsixTH/wiki/Programming-Ideas#key-bindings) I noted the only unimplemented original feature was having stored camera positions that you can recall with a key combo. Alt 0-9 to save a position and Ctrl 0-9 to restore them. Currently I don't have this changing the zoom level, so when zoomed out, the movement is limited but if zoomed in a bit, you can navigate to different part of the maps quickly.

I included the fix to address room queue dialog mood icons showing queue icons when dialog is interacted with - #789 - changes limited to onCursorWorldPositionChange.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 127
+local SAVEGAME_VERSION = 128
 
 class "App"
 
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -86,6 +86,8 @@ function GameUI:GameUI(app, local_hospital, map_editor)
   self.current_momentum = {x = 0.0, y = 0.0, z = 0.0}
   self.multigesturemove = {x = 0.0, y = 0.0}
 
+  self.recallpositions = {}
+
   self.speed_up_key_pressed = false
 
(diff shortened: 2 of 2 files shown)


TheCycoONE:
The hit test fix for 789 should be in a separate commit (I don't mind it in the same PR)

I don't understand the `--register` and `--deregister` comments where you are doing the key binding.

Have you tried with and without including the zoom level to see how it felt? I'm curious which is the better experience.

mugmuggy:
Sorry comments are probably a bit confusing, as I refactored and changed my mind a couple of times on the functionality, I'll fix that up with the commit being fixed up. Maybe I should have created an issue for discussion. I originally had one function, which I was going to call with true, false or nil parameter to set/recall/unset - unset is overkill.

Without the zoom, you could have 2 or more camera positions that are close and when you zoom right out it won't move, as the camera is limited, however with zoom, you can store a view that is zoomed all the way out, but just in typing that, sounds like a useful hotkey, a bit like transparent walls.

But with zoom feels better to me, but I haven't played much with the feature, as the intent really is to have a particular view that you can go back to easily.

TheCycoONE:
@mugmuggy this is still marked as WIP, what is the state of this PR?

mugmuggy:
It functionally works with the zoom, and it plays ok as it is so it is 'ready'. I'll just check that nothing is broken with a rebase and can push another commit and fix the conflict up too if you like.

TheCycoONE:
Sounds great. First feature for the next release!

TheCycoONE:
Bumped the SAVEGAME_VERSION and merged.

PR #1308, opened 2017-12-20
[RDY] Fix campaign not preserving salary between levels

DavidFair:
Fixes the salary not increasing between levels. 
In `app.lua : 592` we only carry across the player state if the `world.campaign_info` variable is set. This isn't set correctly in `dialogs/resizables/newgame.lua : 199`  so the salary was not preserved.

Also I've removed a check that `TheApp.world` has been created, `loadLevel` should handle this and bail if something goes wrong on our behalf instead of us checking and ignoring.

Fixes #1275 

----
Ideally the UI dialogue should not be setting states on the app behalf. Instead it should be tweaked so `app.lua` has a method to start the original campaign (we already have a function for custom campaigns) which keeps this logic in a single location. However I wanted to get this change into the RC so left the re-factoring for a future date.

-----
To test:
- Load a new campaign
- Either play through the level properly or cheat through a couple of months
- Check the end of year salary and make a note
- Win the level and proceed to next level (cheats also work but forfeit the bonus)
- Using cheats skip to end of year: 
- The salary should be greater than the end of year from the previous level (ie. not around 10,000-11,000)

Issue #144, opened 2014-01-16
FHS compliance - 2 issues


mattmill30:
1. CorsixTH binary is installed into /usr/share/games/corsix-th, with a symlink in /usr/games.
   The binary should be installed into /usr/games directly.
2. Original data folders should be expected in /usr/share/games/corsix-th.
   a) If the original data folders (and files within) can't be located the 'HOSP location' dialog should then be displayed.
   b) Original folders (e.g. LEVELS) should be checked for before new folders (e.g. Levels), to avoid 'HOSP location' always being displayed (and never satisfied) when both folders are stored in /usr/share/games/corsix-th

TheCycoONE:
I'm working on this now.

My thoughts are that a distro package should be able to install the CorsixTH executable in
/usr/bin/corsixth and the rest of the CorsixTH platform independent resource files under a directory configurable in CMake e.g. (/usr/share/[games/]corsixth).  I am currently investigating the best way to handle this in CMake - I'm sure it's a common requirement for programs in posix environments.

HOSP data from the original game will not be included by distributions as that would be a copyright violation.  Since that location is prompted at first run I don't see anything to do there, the user is free to store it in /home/user/theme-hospital or /usr/share/local/theme-hospital or anywhere else they feel is appropriate.

Lego3:
Have you looked at this anything more, TheCycoONE?

TheCycoONE:
Not recently. I know the fix but I haven't prioritized it. I suppose anyone else could take it over if they wanted.

TheCycoONE:
Part 1 was addressed in 0.62, part 2 is partially addressed. The default location of original game files is not a clear issue. Closing this.

PR #1318, opened 2018-01-01
[WIP] Organise C++ Files and add CPP Unit Testing

DavidFair:
Background

Following #1308 I started looking into refactoring the Lua scripts. However CorisxTH would run out of memory and be killed if the scripts were incorrect.

Digging deeper I discovered that we leak around 200MB when running correctly and if we error out can rapidly run into 30GB+ of leaked memory (using Valgrind). This is because we use raw pointers through the program and only return without releasing in many cases when there is an error.

As a roadmap I want to migrate the project from C style pointers to C++ RAII containers (such as unique_ptr). However this requires new unit tests to ensure we don't regress the game as part of the modernisation.

-----
Sorry this is a big PR from nowhere. I wasn't planning on completing this work but things gradually came together over a day.

This PR:
- Adds the Google Test framework as an external project to run the unit tests
- Separates CPP header and source files into respective folders (`inc` and `src`)
- Bumps the minimum C++ standard to C++-14 (allowing us to use `make_unique`). The minimum compiler required now is GCC 4.9+ / MSVC 2013+. However these are both old enough to feel confident about proposing this change.
- Enables target folders on IDEs that support it (such as MSVC) so we can move Unit Test related targets into their own folder.
- Enables using `vcpkg` by default on MSVC. (This allows use of cmake-gui for fresh projects and users who are comfortable providing their own libraries instead should be able to use `-DUSE_VCPKG_DEPS=OFF`)
- Adds post build steps to copy the required folders/file: Bitmaps, Campaigns, Levels, Lua and Corsixth.lua into the build directory using CMake.
- Moves `ConvertLegacySprite` into `sprite.h/cpp` and renames / clarifies the code
- Adds a two new unit tests, `example` which provides a testing framework sanity check. And `SampleTest.h` which was written and checked before the refactoring above took place.
- Adds the Unit Testing step to the Travis Build. I wasn't sure how to do something similar with AppVeyor so left it as is.

As always I'm happy for any / all feedback, suggestions or improvement ideas.

----

Testing:
- Build all targets (`CorsixTH` and `UnitTests`)
- Run the unit test target:
1. `cd CorsixTH/CppTest`
2. `./UnitTests`
- Check CorsixTH still loads correctly. However there have been no changes to the main codebase.

-----

Future work ideas:

One of the most important pieces of work is to resolve the memory leaks throughout the code base by using modern containers. There are three approaches I could take to modernising the codebase in personal preference:

- Adding a new data structure which internally holds the data as a `vector<uint8_t>` which has methods to get a reference or copy of the data as a vector or C array (depending if the caller is working with Lua or not). 
- Converting `uint8_t *` to `std::vector<uint8_t>` within the codebase. Then asking for the vector raw pointer or copying the vector into a new pointer at the last moment when we need to use Lua (depending on whether the Lua function takes ownership). 
- Wrapping all raw pointer arrays in `std::unique_ptr`. This loses cache locality and the benefits of being on the stack. The trade of is we don't need to do the conversion when going to Lua. However if the Lua function does not take ownership we must think about data structures to make sure the `unique_ptr` is not released after we forward to the Lua function.

Alternatively I am happy to consider any other suggestions or ideas?

If I do this work it will be gradual as I can consistently spare only a few hours per week for the upcoming months. I plan on creating two helper methods to convert from raw pointers -> new container (or vice versa) so the codebase can gradually be transitioned across multiple PRs. Before any function is converted a collection of unit tests will be created so we can avoid regressing.


CMakeLists.txt
```diff
@@ -16,24 +16,25 @@ CMAKE_MINIMUM_REQUIRED(VERSION 3.2)
...
-CHECK_CXX_COMPILER_FLAG("-std=c++0x" COMPILER_SUPPORTS_CXX0X)
-IF(COMPILER_SUPPORTS_CXX11)
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-ELSEIF(COMPILER_SUPPORTS_CXX0X)
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
-ENDIF(COMPILER_SUPPORTS_CXX11)
+# Allow CMake to use folders in IDEs which support it
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+
+# C++14 Required
```
TheCycoONE: I'm good with bumping the target language to C++14, but is it actually related to this commit in any way?

Such an important change should either go in it's own commit, or with the introduction of C++14 features into the code.

TheCycoONE:
Can you add a quick justification as to why you selected Google Test Framework (over another unit testing framework for C++). I would like some discussion on this before we jump on one or another.

TheCycoONE:
I would like input from other contributors on bd7a723

Note that:
* for FHS compliance we already need better support for selecting a datadir. Some packages patch our source code to change the location - should be a cmake variable.
* The install target already copies those files: https://github.com/DavidFair/CorsixTH/blob/bd7a72332f17023801d93cf68ef78cabd8a91924/CorsixTH/CMakeLists.txt#L234 (to the CorsixTH executable directory, not the proper location for FHS compliance)

TheCycoONE:
The sprite function doesn't need it's own file. It is logically connected to the other gfx code.

See similar anti-guideline: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rnr-lots-of-files

TheCycoONE:
@DavidFair I'm very much in favour of the things you've listed under future work. If you look at my own commit history I have been slowly pushing the code in that direction myself. Additionally I would like to eliminate code macros.

I am in favour of adding a unit test framework, but I have strong reservations about adding one without also including a substantial number of tests to prove it; otherwise we're likely to end up like the luacheck situation, with a test framework in place but serving no use.

All changes unrelated to adding the unit test framework, and adding tests I'd prefer were not in this PR. Many of the changes included are controversial, and worthy of their own discussion.

CorsixTH/CppTest/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+
+
+
+)
+
+enable_testing()
+
+add_custom_target(AllTests)
+
+# Download and unpack googletest at configure time
```
TheCycoONE: Downloading gtest shouldn't be the only (or even the default) behaviour.

If people have configured to use vcpkg, then gtest can be installed from there. We should use FIND_GTEST  to locate it and import the relevant macros.

CorsixTH/CppTest/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+                 "${CMAKE_BINARY_DIR}/googletest-build"
+                 EXCLUDE_FROM_ALL)
+
+# Move gtest into its own folder
+set_target_properties(gtest gtest_main PROPERTIES FOLDER UnitTests)
+
+# Link together all the unit test files
+add_executable(exampleTest ${TESTING_FILES})
+target_link_libraries(exampleTest gtest_main)
+add_test(NAME example COMMAND exampleTest)
```
TheCycoONE: I'd prefer `GTEST_ADD_TESTS` for the improved output when running ctest.

DavidFair:
> I oppose this commits changes.
> [...]
> It would be more confusing in a future where directories also represented namespaces.

Agreed, I was working on some conventions which talk about moving headers into their own folder without realising this applies to a public library interface. I'll move them back.

> I intentionally set this OFF by default.
> If it is off and you meant to have it on then the build will fail nearly immediately. If it is on when you meant to have it off the build will succeed, potentially with the wrong libraries and after a fairly long time.

My only concern was you cannot use vcpkg dependencies with the GUI unless you have used the command line first.
Initially I enabled it by default as doing a complete cache clean + build with developer warnings enabled as it was more convenient in the GUI. However I agree silently linking against the wrong libraries would be a nightmare to debug, so I'm not sure where to go from here as it requires discussion.

> Can you add a quick justification as to why you selected Google Test Framework (over another unit testing framework for C++). I would like some discussion on this before we jump on one or another.

No reason. I saw there wasn't unit tests in CorsixTH and wanted to start refactoring; this just happened to be the easiest to integrate as a proof of concept which is widely used.
I'm happy to use any testing framework if people have personal preferences and it's the reason I only wrote a single test....

> I am in favour of adding a unit test framework, but I have strong reservations about adding one without also including a substantial number of tests to prove it; otherwise we're likely to end up like the lua situation, with a test framework in place but serving no use.

This is why I added only a single test for a function which was proving problematic to refactor. Whilst it's easy for me to to say I think we should add unit tests going ahead and writing one shows I am committed to actually using a testing framework. Without this I worry we would we end up in the same situation as the Lua testing where the framework was added without any tests. 

However I didn't want to commit too much time in case the changes were not accepted.

> The install target already copies those files: https://github.com/DavidFair/CorsixTH/blob/bd7a72332f17023801d93cf68ef78cabd8a91924/CorsixTH/CMakeLists.txt#L234 (to the CorsixTH executable directory, not the proper location for FHS compliance)

I added the extra steps as I had to remember to copy the various Lua files between the source directory and build directory each time I modified them whilst testing. This depends on developer workflows largely I suspect and is worthy of discussion.

> The sprite function doesn't need it's own file. It is logically connected to the other gfx code.
> See similar anti-guideline: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rnr-lots-of-files

I initially moved it out as it didn't have it's own header in the `th_gfx_sdl.cpp`. I agree having a file just for a single function isn't correct, and just left it there for the purpose of writing a unit test. I'll move it into `th_gfx.cpp` as it is not SDL specific though.

> I'm very much in favour of the things you've listed under future work. If you look at my own commit history I have been slowly pushing the code in that direction myself. Additionally I would like to eliminate code macros.

> All changes unrelated to adding the unit test framework, and adding tests I'd prefer were not in this PR. Many of the changes included are controversial, and worthy of their own discussion.

Many of the changes in this PR happened in a single day trying to just understand a single function and I suspected they would be controversial (actually more controversial than this) but wanted to spark some discussion around them.

I'm glad to see there is interest in the ideas proposed. I'll try to open a discussion before prototyping changes in the future, as I agree with several points of feedback given here.

------

I see this PR being split squashed into several separate PRs, something like this?: 

Potentially least controversial :

1. Unit testing framework and initial unit tests. (Do we stick with Google Test or go with something else?)
2. Move `convertLegacySprite` from `th_gfx_sdl.cpp` to `th_gfx.cpp` and include the new header which was missing
3. Bump C++ standard to C++-14 for `make_unique` and enable folders for IDEs that support it

More controversial and discussion required:

4. Default option for vcpkg, do other developers link against their own system libraries instead or is this the exception to the rule?
5. Copying data directories / files into the build folder and should it be included in the build target steps?

Revert:

- Moving header files into their own folder
- Vcpkg on by default (depending on aforementioned discussion)

Alberth289346:
Make unit tests stuff optional in cmake?
Users building a version for themselves don't necessarily want to download and run external projects

TheCycoONE:
From my research so far, Google Test seems fine (widely available, reasonably expressive, used by quite a few other projects, cross platform.)

Every major distro seems to include gtest packages, which for me reinforces the idea we treat it exactly the same way as other dependencies. A flag to enable/disable building the tests would be appropriate.

DavidFair:
Due to work commitments I haven't won't be able to look at this PR until the 12th. 

When I next look at this I'll split out the testing commits, add some additional unit tests too and make them all optional.

TheCycoONE:
I've come around to being in favor of copying the data directories to the cmake_binary_dir.

I don't think they need to be copied to the target directory, so we could use the more ergonomic FILE commands instead of the external cmake calls.

@DavidFair do you mind if one of us handles this as a separate PR. I would like to attack #144 at the same time - they are closely related.

DavidFair:
@TheCycoONE Go for it, I'm more than happy for people to take any chunk they see fit from this PR into a separate one.

DavidFair:
Upstream *nix does not include a prebuilt version of GTest: [Readme](https://www.apt-browse.org/browse/ubuntu/trusty/main/all/libgtest-dev/1.6.0-1ubuntu6/file/usr/share/doc/libgtest-dev/README.Debian). 

The short version is that linking to a pre-compiled version might violate the one definition rule and cause link time / runtime errors.

The workarounds are:
- Require users to build their own copy of the library whilst checking their compilation options always match ours
- Build our own copy of GTest locally 
- Bundle the precompiled version in some manner

`VCPKG` works - but only by chance as the compiler flags (and ABI) match the default used by the packaging tool. However this is fragile.

The [recommended](https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#why-is-it-not-recommended-to-install-a-pre-compiled-copy-of-google-test-for-example-into-usrlocal) practise is to let CMake pull this in as an external project and build it using the same CMake file so it's guaranteed the same compilation flags are used, so I'm going to stick with this unless we can figure out a better way of doing this?

DavidFair:
Closing for #1550

PR #1322, opened 2018-01-08
Issue 1321: Adds default build path to .gitignore

pschoffer:
Even though the recommended path to build the game is outside the source, when running make without specifying path the game still gets build to `CorsixTH/CorsixTH`. That's why I created this PR anyway:) Please see the issue for more info.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,3 +92,6 @@ CorsixTH/CorsixTH.sln
 .cproject/
 .vscode/
 .cproject
+
+# Default build path
+CorsixTH/CorsixTH


TheCycoONE:
Since it's a path, please include the trailing `/`

pschoffer:
Ohhh. I was wrong. So I thought the whole `CorsixTH` (insede the `CorsixTH` repo is build), but is only a `CorsixTH/CorsixTH` executable file. Meaning only the file should be ignored in my opinion. Meaning I can't put the trailing `/` as git will then look for folder and not match files. Do you think we could keep it as is then?

PR #1324, opened 2018-01-09
[RDY] Path manager

pschoffer:
This is a first step to take out some logic from the config_finder to some more focused place. 

Next steps I would like to do are for starters:
* replace other usage of package.config:sub(1, 1)
* Move more stuff to packageManager (and unittested it) e.q `pathconcat`

Feel free to tear down this PR:) It is my first lua code after all, and I might be missing the mindset shift for sure.


CorsixTH/Luatest/spec/system/path_manager_spec.lua
```diff
@@ -0,0 +1,32 @@
...
+require("system.path_manager")
+
+describe("Path Manager -", function()
+    it("should return right separator", function()
+        local path_manager = PathManager()
+        local expected_separator = package.config:sub(1, 1)
+        assert.equal(expected_separator, path_manager:getSeparator())
+    end)
+end)
+
```
TheCycoONE: trailing whitespace here

CorsixTH/Lua/system/path_manager.lua
```diff
@@ -0,0 +1,37 @@
...
+--]]
+
+class "PathManager"
+
+---@type Room
+local PathManager = _G["PathManager"]
+
+
+function PathManager:PathManager()
+    self._separator = package.config:sub(1, 1)
```
TheCycoONE: We use a 2 space indent for lua code.

CorsixTH/Lua/system/path_manager.lua
```diff
@@ -0,0 +1,37 @@
...
+
+--[[
+    PathManager is meant to unify the logic connected with file path operations.
+--]]
+
+class "PathManager"
+
+---@type Room
+local PathManager = _G["PathManager"]
+
```
TheCycoONE: drop extra newline

CorsixTH/Luatest/spec/system/path_manager_spec.lua
```diff
@@ -0,0 +1,32 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE. --]]
+
+require("class_test_base")
+
+require("system.path_manager")
+
+describe("Path Manager -", function()
+    it("should return right separator", function()
+        local path_manager = PathManager()
```
pschoffer: This fails build with `CorsixTH/Luatest/spec/system/path_manager_spec.lua:27:30: (W113) accessing undefined variable PathManager`
Do you know why? And how to fix it?

TheCycoONE:
Instead of storing the path separator in self, you could store it as a file local. Then `getPathSeparator` could be called as a static.

Path might be a better name than PathManager.

CorsixTH/Lua/config_finder.lua
```diff
@@ -18,8 +18,10 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
+dofile "system/path_manager"
```
TheCycoONE: Should be require probably; we don't want to load this file twice.

CorsixTH/Luatest/spec/system/path_manager_spec.lua
```diff
@@ -0,0 +1,32 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE. --]]
+
+require("class_test_base")
+
+require("system.path_manager")
```
TheCycoONE: `/` not `.`

CorsixTH/Lua/config_finder.lua
pschoffer: I got the same error with it not being local here as well

CorsixTH/Lua/config_finder.lua
```diff
@@ -18,8 +18,10 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
+dofile "system/path"
```
pschoffer: This still needs to be fixed. When I try to do require instead. I get an error that the module loader can't find it.

CorsixTH/Luatest/spec/system/path_spec.lua
```diff
@@ -20,13 +20,11 @@ SOFTWARE. --]]
...
-        assert.equal(expected_separator, path_manager:getSeparator())
-    end)
-end)
-  
\ No newline at end of file
+  it("should return right separator", function()
+    local expected_separator = package.config:sub(1, 1)
+    assert.equal(expected_separator, Path.getSeparator())
+  end)
+end)
```
TheCycoONE: Still a problem with your file endings on all files - strange editor.
There should be a single newline at the end of the file.

pschoffer: I guess it should be fixed now?

CorsixTH/Lua/config_finder.lua
```diff
@@ -18,8 +18,10 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
+require("CorsixTH/Lua/system/path")
```
pschoffer: I don't know if there is a better way of doing this? This seems to work though

pschoffer:
Hi @TheCycoONE is there something you are missing here (there is a required change from you, but I think it was covered already maybe?)?

TheCycoONE:
No, there just isn't much advantage to the PR as it is right now. The standard lua way of getting a path separator is ugly but conventional to the language. path.concat is used for the same purpose as pathsep, but there isn't an effort to standardize path building. For corsixth.require we use yet another approach with gsub.

So it's a very nice PR but it needs to go further with unification to be worth merging.

PR #1325, opened 2018-01-10
[RDY] CMake Cleanup

TheCycoONE:
Style consistency and dead code removal for the CMake scripts.


CMakeLists.txt
```diff
@@ -11,133 +11,135 @@
...
 # Get precompiled dependencies before running the various find modules
-if ( USE_PRECOMPILED_DEPS )
+if(USE_PRECOMPILED_DEPS)
   message("Note: Using precompiled dependencies.")
   include( PrecompiledDeps )
 endif()
 
 # Environment handling
-CHECK_INCLUDE_FILES(inttypes.h CORSIX_TH_HAS_INTTYPES_H)
+check_include_files(inttypes.h CORSIX_TH_HAS_INTTYPES_H)
```
DavidFair: This can be dropped. C99 support was added in VS2013 so we don't need to handle this ourselves.

CorsixTH/CMakeLists.txt
```diff
@@ -75,143 +75,145 @@ IF(APPLE)
...
-  INCLUDE_DIRECTORIES(${MSINTTYPES_INCLUDE_DIRS})
-  SET(CORSIX_TH_HAS_INTTYPES_H 1)
-ENDIF(MSVC AND NOT CORSIX_TH_HAS_INTTYPES_H)
-
-IF(MSVC AND CORSIX_TH_USE_VLD)
-  FIND_PACKAGE(VLD REQUIRED)
-  IF(VLD_FOUND)
-    TARGET_LINK_LIBRARIES(CorsixTH ${VLD_LIBRARY})
-    INCLUDE_DIRECTORIES(CorsixTH ${VLD_INCLUDE_DIR})
+if(MSVC AND NOT CORSIX_TH_HAS_INTTYPES_H)
```
DavidFair: Ditto on dropping this block as MSVC has this header natively in VS2013+.
In addition we should change the only usage of `inttypes.h` in `SpriteEncoder/scanner.cpp` to use `<cinttypes>` instead.

CorsixTH/CMakeLists.txt
```diff
@@ -221,11 +223,12 @@ IF(APPLE)
...
     ")
-ELSE()
+else()
   install(TARGETS CorsixTH RUNTIME DESTINATION CorsixTH)
   install(DIRECTORY Campaigns Lua Levels DESTINATION CorsixTH PATTERN "*.svn" EXCLUDE)
   install(DIRECTORY Bitmap DESTINATION CorsixTH
-        FILES_MATCHING REGEX ".*\\.(tab|pal|dat|png)$"
-        PATTERN "*.svn" EXCLUDE)
+    FILES_MATCHING REGEX ".*\\.(tab|pal|dat|png)$"
+    PATTERN "*.svn" EXCLUDE
```
DavidFair: Do we need to explicitly exclude SVN files or is this a holdover from before the project used git?

CMake/FindLua.cmake
```diff
@@ -57,30 +56,30 @@ FIND_LIBRARY (LUA_LIBRARY NAMES ${LUA_LIBRARY_NAME}
...
+if(NOT LUA_LIBRARY)
+  message(FATAL_ERROR "Could Not Find Lua Library")
+endif()
 
 
-IF (LUA_INCLUDE_DIR AND LUA_LIBRARY)
-    SET (Lua_FOUND TRUE)
-ENDIF (LUA_INCLUDE_DIR AND LUA_LIBRARY)
+if(LUA_INCLUDE_DIR AND LUA_LIBRARY)
+  set(Lua_FOUND TRUE)
```
DavidFair: If `LUA_LIBRARY` or `LUA_INCLUDE_DIR` are not found `FATAL_ERROR` means we bail out. So this check can be removed entirely

CMake/FindLua.cmake
```diff
@@ -57,30 +56,30 @@ FIND_LIBRARY (LUA_LIBRARY NAMES ${LUA_LIBRARY_NAME}
...
-       MESSAGE(FATAL_ERROR "Could not find ${LUA_INTERPRETER_TYPE}")
-   ENDIF (Lua_FIND_REQUIRED)
-ENDIF (Lua_FOUND)
+if(Lua_FOUND)
+  if(NOT Lua_FIND_QUIETLY)
+    message(STATUS "Found ${LUA_INTERPRETER_TYPE} library: ${LUA_LIBRARY}")
+  endif()
+else (Lua_FOUND)
+ if(Lua_FIND_REQUIRED)
+   MESSAGE(FATAL_ERROR "Could not find ${LUA_INTERPRETER_TYPE}")
```
DavidFair: And we don't need to check `Lua_found` or the else condition at this point, we know it had to be set here.

TheCycoONE:
Actually we can drop FindLua.cmake entirely because of the CMake version bump recently. It's included upstream.
You're right about the msinttypes too; and there's some other convoluted and dead code I haven't addressed yet which could belong in this PR.

CorsixTH/CMakeLists.txt
TheCycoONE: Yes to both, but the second is out of scope for this PR.

CorsixTH/CMakeLists.txt
TheCycoONE: Hold over, I can drop these.

TheCycoONE:
I need to update the Deps repository first because FindLua.cmake considers `A/include/lua` to be higher president than `A/include/`

TheCycoONE:
Even after updating Deps we are still getting the wrong lua include. I think our FindLua.cmake was better (or at least incompatible with the upstream one), so I will restore it and fix it's issues.

CMakeLists.txt
```diff
@@ -11,133 +11,132 @@
 #   - WITH_LIBAV   : Whether to use LibAV (as opposed to FFMEPG) when building movies
 #   - WITH_VLD     : Build with Visual Leak Detector (requires Visual Studio)
 
-CMAKE_MINIMUM_REQUIRED(VERSION 3.2)
+cmake_minimum_required(VERSION 3.2)
```
DavidFair: Do we need to bump the CMake version to 3.7 for `FindLua`?

Issue #43, opened 2013-10-17
Counting dates in CorsixTH


sadger:
Currently world.lua and possibly other files track the date in months (this stems from the original level files operating in months) meaning all over the code base there is explicit conversion of date in months to the actual month something like `actual_months = 1 + ((months - 1) mod 12)` i.e. 
- months = 12 = actual_months = 12 and
- months = 32 = actual_months = 8 (August of year 2) 

Ideas for solving this would be one of:
1. Track the actual_months separately either keep them updated or introduce functions to calculate them on the fly `get_actual_months(months)` 
2. Rework how counting is done either count years and month separately so
we always have the actual month, or just count in months and work out what 
year it is when required (currently knowing the year isn't a very common operation but a necessary one) 

The second way is my preferred way but this is really an open discussion.

Alberth289346:
The issue seems a little bigger than months.

There are also 'day' and 'hour' variables. Lots of code seems to be checking for a point in time by comparing day and month variables, or setting up future points in time by adding amounts to days or hours. In case of the former one then has to check for end-of-month cases, etc (earth quakes).

A direction for this issue could be to have a central notion world.current_time measured in hours since game-start, and relate all timing and timers to that notion.

The simplest form of that time notion would be a single integer, but I can also imagine a table or class with several fields. A table or class would allow keeping shadow variables like month_in_year, etc for reducing cpu costs in converting.

It should be possible to create future points in time, and check whether points in time have already passed (a fundamental question here is how to handle different time granularities, eg how many hours is 'two days from now', if the current hour is 5 (with 50 hours in a day)).

Finally, one should be able to get human-readable dates from it, for display.

PR #1327, opened 2018-01-11
[RDY] 43 world counter

pschoffer:
This is related to #43 

The change contains:
* Moving month to dedicated `Date` class
* Unifying the `days per month` logic
* replace usage of `world.month` with the new `world.game_date` of `Date` class
* Move `world.day` to `world.game_date`
* Move `world.year` to `world.game_date`
* Move and unify date adjustment logic to `Date` (e.g. next VIP vs next Emergency)
* Cover the Date operations with unittests


TheCycoONE:
Don't worry about travis, it's broken until #1325 is merged.

TheCycoONE:
I'd rather you didn't make a new directory for this unless/until more classes can go in it. Date is fine to be top level for now.

TheCycoONE:
Looking forward to seeing this flushed out.

pschoffer:
@TheCycoONE Thanks for checking. For the folder in my opinion that root folder is way to crammed at the moment and some more structure would be useful, but I understand your point. What about creating a `util` folder with this `Date` + `Path` class  in it?

TheCycoONE:
I would like to see utilities.lua and Path.lua in the same space because they fill the same function (abstracting system details)

Date doesn't belong with them, it is game logic.

World is too general, unless you move entities and hospital and and objects etc.

But I don't see the problem with a relatively flat stucture. What problems is it causing you?

pschoffer:
None really, just seems hard to navigate. But I agree it is not that bad. I will move it to the root folder:)

CorsixTH/Lua/world.lua
```diff
@@ -1020,13 +1020,13 @@ function World:onTick()
         self.day = 1
         self.game_date = self.game_date:plusMonths(1)
         if self.game_date:monthOfYear() > 12 then
```
TheCycoONE: This condition will never be true now.

pschoffer: Yep thanks. I am aware of this and the next commit should fix that.

CorsixTH/Lua/world.lua
```diff
@@ -1010,29 +1002,24 @@ function World:onTick()
...
       self.hour = self.hour - self.hours_per_day
-      self.day = self.day + 1
-      if self.day > month_length[self.month] then
-        self.day = month_length[self.month]
+      if self.game_date:isLastDayOfMonth() then
         for _, hospital in ipairs(self.hospitals) do
           hospital:onEndMonth()
         end
-        -- Let the hospitals do what they need to do at end of month first.
+         -- Let the hospitals do what they need to do at end of month first.
```
pschoffer: fix extra space

.luacheckrc
```diff
@@ -37,8 +37,8 @@ globals = { -- Globals
             "Patient", "Plant", "Queue", "ResearchDepartment", "Room",
             "SaveGame", "SaveGameFile", "Staff", "StaffProfile", "StaffRoom",
             "Strings", "SwingDoor", "TheApp", "TreeControl", "Vip", "Window",
-            "World",
-
+            "World", "Date",
+
```
pschoffer: Another extra space

TheCycoONE:
It would be nice if some of the metatable methods could be implemented for this new Date type; specifically `__eq`, `__lt`, and `__lte`. That would simplify several of the comparisons in the code.

pschoffer:
@TheCycoONE  I totally agree. I forgot to ask about that. 

That was my initial plan, but I couldn't figure out how to do it. I tried to define 
```
function Data.__eq(one, other)
  -- code
end
```
and
```
function Data:__eq(one, other)
  -- code
end
```

But neither of them work. Then I tried some messing up with getmetatable(Data), but no luck either. I tried to find some reference in code but couldn't. But Maybe I was not looking hard enough:) Any pointers?

TheCycoONE:
Right, the facility isn't really there for it at the moment, but it can be added:

```diff
diff --git a/CorsixTH/Lua/class.lua b/CorsixTH/Lua/class.lua
index e601df8a..a40160b6 100644
--- a/CorsixTH/Lua/class.lua
+++ b/CorsixTH/Lua/class.lua
@@ -88,6 +88,7 @@ local function define_class(name, super, adopts_self)
   methods_mt.__class_name = name
 
   _G[name] = methods
+  _G[name .. '_mt'] = mt
 end
 
 strict_declare_global "class"
diff --git a/CorsixTH/Lua/date.lua b/CorsixTH/Lua/date.lua
index 185c3541..73df77c3 100644
--- a/CorsixTH/Lua/date.lua
+++ b/CorsixTH/Lua/date.lua
@@ -27,6 +27,7 @@ class "Date"
 
 ---@type Date
 local Date = _G["Date"]
+local Date_mt = _G["Date_mt"]
 
 local month_length = {
   31, -- Jan
@@ -100,6 +101,10 @@ function Date:clone()
   return Date(self._day, self._month, self._year)
 end
 
+Date_mt.__eq = function(self, other)
+  return self._year == other._year and self._month == other._month and self._day == other._day
+end
+
 -- PRIVATE
 
 function Date:_adjustDayOverflow()
diff --git a/CorsixTH/Luatest/spec/date_spec.lua b/CorsixTH/Luatest/spec/date_spec.lua
index 915681a7..0b524387 100644
--- a/CorsixTH/Luatest/spec/date_spec.lua
+++ b/CorsixTH/Luatest/spec/date_spec.lua
@@ -120,6 +120,12 @@ describe("Date", function()
     local origin_date = Date(12,3,2)
     local clone_date = origin_date:clone()
     assert.True(origin_date:equals(clone_date))
-    assert.False(origin_date == clone_date) -- different objects
+  end)
+  it("implements ==", function()
+    local date1 = Date(1,2,3)
+    local other_date1 = Date(1,2,3)
+    local date2 = Date(3,2,1)
+    assert.True(date1 == other_date1)
+    assert.False(date1 == date2)
   end)
 end)
```

TheCycoONE:
Diff above is a proof of concept. Of course it should be cleaned up, and the class extension should be it's own commit. I'm not sure that name_mt is the best convention either.

pschoffer:
Awesome thanks! I will try to do that in the afternoon.

CorsixTH/Lua/hospital.lua
```diff
@@ -1030,7 +1033,7 @@ function Hospital:onEndDay()
   if breakdown == 1 and not self.heating_broke and self.boiler_can_break and
       self.world.object_counts.radiator > 0 then
     if tonumber(self.world.map.level_number) then
-      if self.world.map.level_number == 1 and (self.world.month > 5 or self.world.year > 1) then
+      if self.world.map.level_number == 1 and (self.world:date():monthOfYear() > 5 or self.world:date():year() > 1) then
```
TheCycoONE: Could be simplified to `self.world:date() > Date(1,5,1)` there are several places like this.

CorsixTH/Lua/date.lua
```diff
@@ -0,0 +1,136 @@
...
+  30, -- Jun
+  31, -- Jul
+  31, -- Aug
+  30, -- Sep
+  31, -- Oct
+  30, -- Nov
+  31, -- Dec
+}
+
+function Date:Date(day, month, year)
```
TheCycoONE: year, month, day is more conventional.

CorsixTH/Lua/world.lua
```diff
@@ -1106,20 +1093,20 @@ function World:onEndDay()
   end
 
   -- check if it's time for an earthquake, and the user is at least on level 5
-  if (self.year - 1) * 12 + self.month == self.next_earthquake.start_month and
-      self.day == self.next_earthquake.start_day then
+  if self.game_date:monthOfGame() == self.next_earthquake.start_month and
```
TheCycoONE: Nicer to use Date for the other side of earthquake and emergency as well.

CorsixTH/Lua/hospital.lua
pschoffer: Done, could you please also point me to other places ? I found some other comparisons as well but they are mostly about some part of year but every year so I couldn't figure out how to compare them.

CorsixTH/Lua/world.lua
pschoffer: I did it for VIP, but both emergencies and earthquake boundaries are configured within level config, and to translate that might be messy. I would prefer to not mess with it within this PR.

pschoffer:
When I was doing some testing I found out this hip new class has some issue with persistence. I will have a look tomorrow.
```
Unhandled argument to th_getfenv
Unhandled argument to th_getfenv
Unhandled argument to th_getfenv
Unhandled argument to th_getfenv
nil	<no name>
Error while autosaving game: ./CorsixTH/Lua/persistance.lua:242: Lua functions must be given a unique persistable name in order to be persisted (attempt to persist ./CorsixTH/Lua/date.lua:112)
```

TheCycoONE:
I personally haven't hit that issue with this PR. It doesn't happen every time then? Do you know the cause?

pschoffer:
For me it happens every time I try to save (or autosave). From the error I guess it tries to persist the metatable functions `__lg` and `__eq` and fails because they are not unique or something like that. I was refreshing my C++ to understand better how it works, but I am not there yet:)

TheCycoONE:
@pschoffer I was on an out of date pr. I can reproduce it. I had the same hunch as you, but it's wrong. This issue seems to be that the method it's using to return the names of those pieces of code is returning nil.

At the moment, one workaround is this:
```diff
diff --git a/CorsixTH/Lua/date.lua b/CorsixTH/Lua/date.lua
index 9c991578..35adea79 100644
--- a/CorsixTH/Lua/date.lua
+++ b/CorsixTH/Lua/date.lua
@@ -105,11 +105,11 @@ end
 
 local Date_mt = Date._metatable
 
-function Date_mt.__eq(one, other)
+Date_mt.__eq = --[[persistable:date_eq]] function(one, other)
   return one._year == other._year and one._month == other._month and one._day == other._day
 end
 
-function Date_mt.__lt(one, other)
+Date_mt.__lt = --[[persistable:date_lt]] function(one, other)
   if one._year == other._year then
     if one._month == other._month then
       return one._day < other._day
```

I think that's ugly though. Perhaps we can do better.

pschoffer:
Man I got soo tired by this Lua <-> C integration so I just `settled` for your solution without understanding it. Feel free to explain that bit to me:)

TheCycoONE:
I don't blame you, the persistence code is pretty opaque. Functions can't be read on their own, because they assume the state of the lua stack, and what that state should be isn't documented.

At a high level, it crawls through the permanent objects tables, recursively. Functions (prototypes) are stored in a cache, keyed by source file and line number, which are indexed; so if the same function is encountered again later in the search (e.g. multiple callbacks storing the same function reference), the cache is checked and the existing index is pulled out so that the function info is only stored once.

Because source name and line number are not going to be constant between game versions, another table exists that maps names to these values. This table is also in the lua stack (put there by init()) and when we add a function we've encountered for the first time, we lookup in this table what the name should be. There are three ways for names to get into that table.

1. All top level functions and class methods are automatically added by persist.lua
2. A function can be annotated with `f = persistable"name"( function() ... end)`
3. A function can be annotated with the `--[persistable:name]]` comment

There are some fine differences between those methods. The problem with jumping directly to method 2 or 3, is that if we do add automatic annotation (which should be doable, given the ._metatable pattern) we would need to make sure the names match, otherwise saves won't be compatible.

https://github.com/CorsixTH/CorsixTH/wiki/Persistence has some more details.

TheCycoONE:
This is actually the fix I recommend:
```diff
diff --git a/CorsixTH/Lua/persistance.lua b/CorsixTH/Lua/persistance.lua
index a779221c..d5be8a21 100644
--- a/CorsixTH/Lua/persistance.lua
+++ b/CorsixTH/Lua/persistance.lua
@@ -103,6 +103,11 @@ local function MakePermanentObjectsTable(inverted)
         permanent[v] = name .. "." .. k
       end
     end
+    for k, v in pairs(class._metatable) do
+      if type(v) == "function" then
+        permanent[v] = name .. "._metatable." .. k
+      end
+    end
   until true end
 
   -- C/class/library methods
```

(I'm not sure if I still like the `_metatable` name - but that would be bike shedding and I don't think we need to dwell on that.)

pschoffer:
Cool I did your last recommendation @TheCycoONE  and it seems to work fine! I also added some more fixes. So I am happy with this.

CorsixTH/Lua/world.lua
```diff
@@ -984,35 +977,30 @@ function World:onTick()
...
       if not status then
         print("Error while autosaving game: " .. err)
       end
     end
-    if self.year == 1 and self.month == 1 and self.day == 1 and self.hour == 0 then
-      if not self.ui.start_tutorial then
-        self.ui:addWindow(UIWatch(self.ui, "initial_opening"))
-        self.ui:showBriefing()
-      end
+    if self.game_date == Date() and not self.ui.start_tutorial then
```
TheCycoONE: Technically correct, but Date(1) might be clearer? I'll leave that up to you.

CorsixTH/Lua/world.lua
```diff
@@ -1183,21 +1170,21 @@ function World:onEndMonth()
...
   -- either 6 months later or at the end of month 12 and then every 6 months
-  if self.month % 3 == 0 and self.month < 12 then
+  if self.game_date:monthOfYear() % 3 == 0 and self.game_date:monthOfYear() < 12 then
     self:checkIfGameWon()
   end
 
   local local_hospital = self:getLocalPlayerHospital()
   local_hospital.population = 0.25
-  if self.month >= self.map.level_config.gbv.AllocDelay then
+  if self.game_date:monthOfYear() >= self.map.level_config.gbv.AllocDelay then
```
TheCycoONE: Another bug in the old code. This should be month of game.

CorsixTH/Lua/world.lua
```diff
@@ -2104,7 +2069,7 @@ function World:objectPlaced(entity, id)
       self.ui.adviser:say(_A.room_requirements.reception_need_receptionist)
     elseif self.hospitals[1]:hasStaffOfCategory("Receptionist") and
         self.object_counts["reception_desk"] == 1 and
-        not self.hospitals[1].receptionist_msg and self.month > 3 then
+        not self.hospitals[1].receptionist_msg and self.game_date:monthOfYear() > 3 then
```
TheCycoONE: Almost certainly should also be month of game.

CorsixTH/Lua/world.lua
```diff
@@ -2469,7 +2434,6 @@ function World:afterLoad(old, new)
     self:nextEmergency()
   end
   if old < 31 then
-    self.hours_per_day = 50
```
TheCycoONE: Best to leave old modifications in, in case code in between relied on it. In this case it probably doesn't. Note that old saves will continue to have `self.hours_per_day` until you nil it out in your < 124 afterLoad.

CorsixTH/Lua/date.lua
```diff
@@ -0,0 +1,182 @@
...
+function Date:year()
+  return self._year
+end
+
+function Date:hourOfDay()
+  return self._hour
+end
+
+function Date:tostring()
+  return self._day .. "/" .. self._month .. "/" .. self._year .. ":" .. self._hour
```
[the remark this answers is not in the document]
pschoffer: Nice point. I kept ISO format apart of 4 year digit as we start from 1 digit and rarely go above 1 or 2 digits.

CorsixTH/Lua/hospital.lua
```diff
@@ -811,7 +814,7 @@ function Hospital:checkFacilities()
 end
 
 --! Called each tick, also called 'hours'. Check hours_per_day in
---! world.lua to see how many times per day this is.
+--! date.lua to see how many times per day this is.
 function Hospital:tick()
```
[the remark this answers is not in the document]
pschoffer: Hmm sorry I didn't get that one.

CorsixTH/Lua/world.lua
pschoffer: I would say either `Date()` or `Date(1,1,1,0)`. Defining some of the parameters and relying on the defaults for other ones seems a bit more confusing to me. But yeah I prefer the default solution.

CorsixTH/Lua/world.lua
pschoffer: I fixed most of the issues. I kinda refactor the whole annoucments section in hospital, changing some of it's logic. It would be nice if you could have a look especially on that.

TheCycoONE:
LGTM

Merge pending play test.

pschoffer:
Pending play test?

TheCycoONE:
I will test before merging so I don't repeat last time. Normally do and get burned whenever I don't.

pschoffer:
Oh I see I thought you want me to merge some `play` test:)

TheCycoONE:
All good. Could you squish?

CorsixTH/Lua/world.lua
```diff
@@ -1083,7 +1083,7 @@ function World:onEndDay()
   self.current_tick_entity = nil
 
   --check if it's time for a VIP visit
-  if self.game_date == self.next_vip_date then
+  if self.game_date >= self.next_vip_date then
```
pschoffer: This is the change i did after your review.

pschoffer:
So I squashed it a bit. Is this enough? I still wanted to keep some stuff in separate commits. Also please check the last commit I remembered and fixed one potential issue with `next_vip_date` after your review.

TheCycoONE:
That's not a good change, the next_vip_date isn't reset again until the previous VIP leaves, or you refuse the fax.

I suppose you could either add sameDay, or nil out the date once you enter the condition.

pschoffer:
Oh right thanks. Nice catch. I fixed it with a `isSameDay` method. Also I refactored the `nextVIP` method in world. This Way it seems a little more intuitive, but I am fine with reverting it back to having the old nextVIP as well.

TheCycoONE:
Sorry, I didn't think to mention it earlier. The new date class and all it's methods should have ldocgen comments.

pschoffer:
Not fun the docs part. But it looks nice in the html thats true:)

PR #1330, opened 2018-01-15
[RFC] Add GoogleTest framework and some unit tests

DavidFair:
This PR is large, however the majority is unit tests as I didn't want to add a testing framework then leave it unused.

Building on the feedback in #1318 this PR:

- Adds Google Testing Framework (we clone and build rather than link, explanation in #1318 )
- Bumps us to using C++-14 standard.
- Makes the unit test target optional and opt-in
- Enables CPP Tests for Travis builds
- Adds unit tests for `THChunkRenderer` 

---

To test:
- Check that by default unit tests are enabled
- Try enabling unit tests (the var is called `ENABLE_UNIT_TESTS` ) and check it correctly runs CMake
- Check the building `CorsixTH` does not build unit tests
- Check doing a bare build does
- Move into $BuildDir/CorsixTH
- Run `ctest -VV` check all tests pass correctly
- Code review.

As always feel free to suggest any changes / improvements ideas.


wolfy1339:
I would move the code relocation/reorganization/fixes (commits 19de544 b343d22 02fd0f3 d532b28) to another PR as they aren't really related to adding unit tests

DavidFair:
Okay I've split this PR and updated the PR message to reflect the current changes

DavidFair:
Closing for #1550

PR #1331, opened 2018-01-15
[RFC] Fixes and reorg from unit test changes

DavidFair:
Moves the changes from #1330 as per PR feedback:

Changes are:
-----

- Moves `convertLegacySprite()` into `th_gfx.cpp` as it is not SDL specific (no changes to source code) following feedback in previous PR.
- Moves the deceleration of `IntersectTHClipRect(..)` into `th_map.h` as the definition is in the respective cpp file (and I didn't think it warranted a separate PR)
  - Fixes up header includes following that move
- `THChunkRenderer` fix ups:

  - Store a `nullptr` instead of 0 when we call `takeData()`
  - Add doxgen in the header for all methods explaining in detail what the method does
  - Account for the size of the data when calling `memset` or `memcpy` 
  - Split decode chunks into two seperate methods: `decodeChunksSimple` and `decodeChunksComplex`
  - Document the code of those new methods, which can be a bit terse to understand at points. Note: There was no code changes for these methods
  - Rename the private method `_isDone()` to `_isEndOfBuffer()`

### DIFF
--- a/CorsixTH/Src/iso_fs.cpp
+++ b/CorsixTH/Src/iso_fs.cpp
@@ -21,6 +21,8 @@ SOFTWARE.
 */
 
 #include "iso_fs.h"
+
+#include <lua.h>
 #include <cstring>
 #include <cstdarg>
 #include <cstdlib>
--- a/CorsixTH/Src/th_gfx.cpp
+++ b/CorsixTH/Src/th_gfx.cpp
@@ -142,6 +142,47 @@ public:
     }
 };
 
+//! Convert legacy 8bpp sprite data to recoloured 32bpp data, using special recolour table 0xFF.
+/*!
+@param pPixelData Legacy 8bpp pixels.
+@param iPixelDataLength Number of pixels in the \a pPixelData.
+@return Converted 32bpp pixel data, if succeeded else nullptr is returned. Caller should free the returned memory.
+*/
+uint8_t *convertLegacySprite(const uint8_t* pPixelData, size_t iPixelDataLength)
+{
+	// Recolour blocks are 63 pixels long.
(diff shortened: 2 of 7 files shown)


lewri:
@DavidFair what's the current state of play with this one?

Alberth289346:
@TheCycoONE looks like a better person to answer that question.

lewri:
Going through the conflicts it looks like this PR should be able to be corrected course onto the current code -- then again some of them look a bit complex and the nature of these files may have changed. Would be good to get some update on what the plans are with this at the current time.

Alberth289346:
Likely at least some of the changes are good, however in its current form it's difficult to understand what exactly is modified and what is just moved. Eg the first commit says `Move decleration for convertLegacySprite`, yet the actual change deletes a function. Likely it gets added elsewhere, but it makes checking the changes difficult.

On the patch:
- A clean split between moves and (a sequence of small) changes would be very helpful in understanding what is being done here.
- Some added code comments should not tell about decimal values but about bit patterns being recognized. I am not sure it's actually needed, as the format being decoded and encoded is documented in a text file somewhere. The code follows that document very closely (if you think in bit patterns rather than numeric values).

The bigger issue is I think what to do with this code:
Part of the code is for legacy sprites, which is actually being used currently. Fixing/improving/documenting that would be fine. Another part is for new (open source) graphics. Since the code already exists for several years and we have no new graphics whatsoever, I think we can conclude that the current attempt is not working. Also, further experiments in this area indicated we may have the wrong approach. This implies rethinking how to solve the open source graphics problem, I believe.
Not sure if it makes sense to fix the current unused open source sprite loading code until the above has been done.

PR #1332, opened 2018-01-16
[Pending] Refactor THChunkRenderer

DavidFair:
**Note this PR cannot be merged / reviewed before #1330 and #1331 are merged.**

-----

Re-factors THChunkRenderer to modernise the code the class uses:

- Drop the optional parameter to an existing array of data in the constructor
  - This was only used in one place and it duplicated the internal logic
- Switches the internals of THChunkRenderer to use a `std::vector<uint8_t>` instead of `uint8_t *`
- Replaces hex values to more readable decimal values for the `decodeChunk` methods
- Removes the inline keyword (which is ignored by the compiler) and renames some private methods
- Uses member initialisation instead of a constructor 
- Switches all methods to accept size_t parameters instead of int

-----

Testing:
- Run unit tests
- Run CorsixTH, the opening animation to the menu uses all code paths of THChunkRenderer when assets are loaded in and converted
- Code Review


Alberth289346:
Hex is better than decimal if you're working with bits, since you can immediately see the bit patterns that the number expresses (I read eg 0xC0 as 1100.0000, I never even consider its numeric value 12*16).

CorsixTH/Src/th_gfx.cpp
```diff
@@ -1033,88 +1092,119 @@ inline void THChunkRenderer::_incrementPosition(int npixels)
...
+        {
+            // Copy the number of pixels represented by inputVal.
+            size_t amt = inputVal;
+            if (dataLen < amt)
+                amt = dataLen;
+            chunkCopy(amt, inputData);
+            inputData += amt;
+            dataLen -= amt;
+        }
+        // If the input has bit 128 set without bit 64
```
Alberth289346: bit 128 is not 0x80, bit 128 is 2**128, which is 340282366920938463463374607431768211456 or 0x100000000000000000000000000000000 (see how hex makes bits much more clear here?).

You mean bit 7 and bit 6 here (bit b has value 2**b, starting with b=0, with value 1.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -1033,88 +1092,119 @@ inline void THChunkRenderer::_incrementPosition(int npixels)
...
+        // i.e. in the range 128-191 
+        else if ((inputVal & 0xC0) == 0x80)
+        {
+            // Fill the with 0xFF for the number of pixels - 128
```
Alberth289346: "the with" ? needs something in-between?

CorsixTH/Src/th_gfx.cpp
```diff
@@ -1033,88 +1092,119 @@ inline void THChunkRenderer::_incrementPosition(int npixels)
...
+            chunkFill(inputVal - 0x80, 0xFF);
+        }
+        else
+        {
+            // B is between 64 - 127 or 192 - 255
```
Alberth289346: Doesn't seem to be a B around here.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -1033,88 +1092,119 @@ inline void THChunkRenderer::_incrementPosition(int npixels)
...
+                dataLen -= 2;
             }
             else
             {
-                chunkFill(0x100 - b, 0xFF);
+                // Bitwise And inputVal with 128, divide this by two
+                // then subtract this from inputVal. Then subtract
+                // 60 from inputVal. This gives the amount of pixels
+                // to copy.
+                amt = inputVal - 60 - (inputVal & 0x80) / 2;
```
Alberth289346: Comment that repeats what the code says is less than useless. At best, it actually repeats it (giving no new information), less optimal if it says something else (which can only give confusion). Instead, explain what is happening here at a higher level, and clarify the 60, as it's pure magic now.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -142,6 +146,47 @@ class Input
     }
 };
 
+//! Convert legacy 8bpp sprite data to recoloured 32bpp data, using special recolour table 0xFF.
+/*!
+@param pPixelData Legacy 8bpp pixels.
+@param iPixelDataLength Number of pixels in the \a pPixelData.
+@return Converted 32bpp pixel data, if succeeded else nullptr is returned. Caller should free the returned memory.
+*/
```
Alberth289346: Corsix documents non-static function in the header files.
(I don't agree with this policy, but that's another matter.)

lewri:
#1331 got closed. Assuming this one can't go forward now either?

PR #1333, opened 2018-01-17
[Pending] Create CorsixTH Library Target

DavidFair:
**Cannot be merged before #1330**

----

Creates a new target called `CorsixTHLib`, this builds all files except for `main.cpp` as a static library. 
Whilst there is an option for shared libraries these will not work on MSVC without additional work. If linking the static library is becoming time consuming we can revisit this, but for the moment linking takes a couple of seconds anyway.

This has no impact on the final executable produced, however it saves us building the source code twice when using unit tests. 

----

To test:
- Check CorsixTH builds and opens correctly
- Build the `UnitTests` target, make sure it doesn't rebuild the source that was built for `CorsixTHLib`
- Make a change anywhere in any file in `Src`
- Make the `UnitTests` target, make sure it correctly builds `CorsixTHLib` to include the change
- Build `CorsixTH` check it doesn't rebuild `CorsixTHLib` again


TheCycoONE:
We have a library now and are using catch2 rather than Google.

Issue #1334, opened 2018-01-19
Freeze Error


sionmaisley:
After starting a pandemic mission everyone in the hospital is frozen, ive tried adjusting the speed and reloading but still the same issue, restarting the level sent me to a blank screen.

Running: The timer handler.
A stack trace is included below, and the handler has been disconnected.
C:\Program Files\CorsixTH\Lua\epidemic.lua:551: attempt to index a nil value (local 'action')
stack traceback:
        C:\Program Files\CorsixTH\Lua\epidemic.lua:551: in upvalue 'is_static'
        C:\Program Files\CorsixTH\Lua\epidemic.lua:561: in method 'markedPatientsCallForVaccination'
        C:\Program Files\CorsixTH\Lua\epidemic.lua:109: in method 'tick'
        C:\Program Files\CorsixTH\Lua\hospital.lua:1410: in method 'manageEpidemics'
        C:\Program Files\CorsixTH\Lua\hospital.lua:834: in method 'tick'
        C:\Program Files\CorsixTH\Lua\world.lua:1039: in method 'onTick'
        C:\Program Files\CorsixTH\Lua\app.lua:1008: in function <C:\Program Files\CorsixTH\Lua\app.lua:1005>
        (...tail calls...)
        C:\Program Files\CorsixTH\Lua\app.lua:909: in function <C:\Program Files\CorsixTH\Lua\app.lua:904>

Warning: No event handler for timer

Version 0.61

OS: Windows 10

Origin Version

pschoffer:
Hi @sionmaisley, do you have a save game or a way to reproduce this?

sionmaisley:
I don't know where to find my save files

Sent from Yahoo Mail on Android 
 
  On Sat, 20 Jan 2018 at 11:42, pschoffer<notifications@github.com> wrote:   
Hi @sionmaisley, do you have a save game or a way to reproduce this?

sionmaisley:
Found it after much digging


Thanks 
Sent from Yahoo Mail on Android 
 
  On Sat, 20 Jan 2018 at 12:48, Christian Thorpe<sionmaisley@yahoo.co.uk> wrote:   I don't know where to find my save files

Sent from Yahoo Mail on Android 
 
  On Sat, 20 Jan 2018 at 11:42, pschoffer<notifications@github.com> wrote:   
Hi @sionmaisley, do you have a save game or a way to reproduce this?

pschoffer:
[...]

PR #1336, opened 2018-01-20
[RDY]1334: Adds a no action check

pschoffer:
Fix for #1334


TheCycoONE:
Just add some noise so it shows up in the game log along with information to identify the patient (type, x, y etc.)

Be clear in code that an invariant has been violated.

Longer term, we need to cut down on action queue meddling and enforce that invariant

pschoffer:
OK that sounds good. I will unify the `getCurrentAction` logic somewhere with logging. To be continued:)

pschoffer:
@Alberth289346 thanks for the comment. I see your point (from both of you:)). and as I mentioned I will prepare the contra solution as suggested.

TheCycoONE:
@pschoffer how does this error look in the output?

pschoffer:
I get this when not catching in in test. So that looks good I think.
screenshot from 2018-01-24 11-33-59

TheCycoONE:
Looks great to me, thanks

TheCycoONE:
We should improve error to dump a call stack, but that can be another issue.

PR #1339, opened 2018-01-23
[RDY] FHS and out of source build support

TheCycoONE:
Adds support for out of source builds in any path, by adding a shell script and visual studio project debugger command line option to set the interpreter path to the source directory.

Gets the proper FHS path to install binaries and data files to on non-apple UNIX systems.

Removes the search for CorsixTH.lua; it should always be correctly provided by the --interpreter command line argument or through cmake for installing. On Windows it is assumed to be in the same path as the executable.

Moves resources to the Resources folder on Apple.

Fixes doxygen for out of source `make doc`

Introduces `-DUSE_SOURCE_DATADIRS=ON` for specifying a development build will not be deployed and can store the location of the source directory as the data location.

What I haven't tested:
- [ ] Does this work if the source path contains spaces, quotes, or other unusual characters (all operating systems)
- [ ] Does this work if the install path contains spaces, quotes, or other unusual characters (linux)
- [x] Does this work at all on Apple
- [x] Makefile or Ninja generator on Windows
- [x]  XCode generator on Apple (may be limited to not creating app bundles)


CorsixTH/CMakeLists.txt
```diff
@@ -185,6 +198,14 @@ if(MSVC AND CORSIX_TH_USE_VLD)
   endif()
 endif()
 
+# Launch script to facilitate out of source builds on Linux
+if(UNIX)
+  configure_file(${CMAKE_CURRENT_SOURCE_DIR}/run-corsixth-dev.sh.in ${CMAKE_CURRENT_BINARY_DIR}/run-corsixth-dev.sh @ONLY)
```
tobylane: The file is run-corsix-th-dev.sh.in with a dash in corsix-th.

CorsixTH/CMakeLists.txt
```diff
@@ -14,6 +14,17 @@ elseif(MSVC)
   add_definitions(/D "_BIND_TO_CURRENT_VCLIBS_VERSION=1")
 endif()
 
+if(MSVC)
+  set(CORSIX_TH_DATADIR CorsixTH)
+  set(CORSIX_TH_INTERPRETER_PATH CorsixTH.lua)
+elseif(APPLE)
+  set(CORSIX_TH_DATADIR ${FULL_PRODUCT_NAME}/Contents/Resources/)
+  set(CORSIX_TH_INTERPRETER_PATH ${CORSIX_TH_DATADIR}/CorsixTH.lua)
```
tobylane: This is set to "/Contents/MacOS/CorsixTH.lua" which is correct relative to the app folder, but the working directory of an app is $home.

CorsixTH/CMakeLists.txt
TheCycoONE: Ah, thanks. I changed the file name in a later commit and didn't fix this line.

CorsixTH/CMakeLists.txt
TheCycoONE: This should be correct at install time, not build time. The initial slash is going to be a problem - otherwise do paths get evaluated relative to the app on OSX?

TheCycoONE:
I think that commit fixes linux, short of potential path issues with characters.

Apple is probably still very broken. I think it doesn't work out of source at all yet, and that rsync stuff has to go (it should be handled by cmake)

TheCycoONE:
@tobylane ok, the above changes work for me with the following issues:
lfs.so and lpeg.so need to be copied manually into the Resources path.
bundle location is fixed by CMAKE_INSTALL_PREFIX. In practice that might be acceptable because on MacOS bundles should always be in Applications, and the installer can enforce that.
I haven't tested at all with xcode, if the IDE mangles paths it might break - everyone uses make && make install right?...

tobylane:
We can remove RequiredResources.txt, these changes make it unused.

TheCycoONE:
I tested XCode with `USE_SOURCE_DATADIRS` and at least with that option it works well.
I know that the shell script to launch with the source directory when `USE_SOURCE_DATADIRS` is not set will not find the right directory; I'm not sure if there's other problems.

TheCycoONE:
I'm good with the state of this now. If anyone has something specific they'd like to change, mention it now or file another PR after this is merged.

tobylane:
Could the three new files go in the CMake folder? 👍 on everything else.

TheCycoONE:
I'd rather not move them; we normally put .in files in the same directory (relative to the source directory) that they output to (relative to the binary directory)
The only other place they might make sense is in the root, outputting to the root.

PR #1342, opened 2018-01-25
[RFC] 1326 fix imports take 2 - Not ready for merge

pschoffer:
Current version breaks persistance

Issue #1326, opened 2018-01-10
Change package.path instead of overloading dofile


pschoffer:
#### Describe the issue
Currently we are in CorsixTH.lua overloading dofile to work in similar way as require but to look for files in a code directory (ignoring your working directory)

The same effect could be achieve by setting `package.path` to check the code folder first.
We should also replace all the `dofile` with `require` (unless the dofile makes sense there).
In the end the overloading `dofile` implementation should be removed.

PR #1343, opened 2018-01-25
[RDY] Dofile move

TheCycoONE:
A less ambitious version of #1326 and #1342


CorsixTH/CorsixTH.lua
```diff
@@ -44,7 +47,10 @@ else
     return ...
   end
 end
-function dofile(name)
+
+_G['corsixth'] = {}
+
+corsixth.require = function(name)
   if pathsep ~= "/" then
```
pschoffer: This I would suggest to change to use "." as universal separator. That way it will behave the same way as Lua require works. Now this "/" was kinda weird to me. This of course means going through all the usages of `crosixth,require(.*\/.*)` and replace with `.`

TheCycoONE:
Agreed, I will switch to dot notation to be more 'require' like.

The other differences from a real require are that it doesn't search the LUA_PATH and it doesn't work with compiled lua.

TheCycoONE:
dot notation changes have been made.

Alberth289346:
Some documentation lines for the new require function, but that's all that seems to be missing.
Nice work!

TheCycoONE:
Added documentation
Rebased on latest master (since date pr landed)
Fixed one require that my sed script had missed in map.lua

Issue #490, opened 2014-12-24
Boiler breakdown rate should depend on stuff


TheCycoONE:
_From [henke@henke37.cjb.net](https://code.google.com/u/henke@henke37.cjb.net/) on April 12, 2012 19:32:18_

The boiler currently breaks randomly. According to the manual it should depend on the number of radiators and the number of handymen. I think that the set heating level should be a variable too.

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1220_

da4id:
Im new in CorsixTH, but i know very well the orignal game. I've made some thougs about this issue which i want to discuss.

Boiler Breakdown should depend on 
- number of Handymens
- number of Radiators
- heating level
- some random

My idea ist to have a breakdown score. Each day this score will be increased by some value depend on heating level, ratio radiators/handyman and some random value. If score reaches critical level boiler breaks down. But there should be some independent random part too. 

The decision if boiler breaks down is now implemented with
`local breakdown = math.random(1, 240)
if breakdown == 1 and not self.heating_broke and self.boiler_can_break and self.world.object_counts.radiator > 0 then`
this gives average brekdown every 3 Months.

What do you think about having something like:
`local breakdown = math.random(1, 10000)
self.boiler_breakdown_score += math.random()*self.radiator_heat*nb_radiators/nb_handyman

if (self.boiler_breakdown_score > criticalLevel or breakdown == 1) and not self.heating_broke and self.boiler_can_break and self.world.object_counts.radiator > 0 then`

with a very low probability it is possible that you have multiple breakdowns within short time, because there is still some random part. 
criticalLevel should be something like:
if you have 5 Radiators per Handyman and a heating level of 50% the boiler should breakdown in average after 5 years. This gives critical level = 5 years * 365 * 5 radiatorsPerHandyman / 4 = 2281.2

I've tested this algorithm in Matlab with 1000 runs and the boiler will breakdown between 58 and 63 months. If you run the same values but with 100% heating rate breakdown will occur every 28 to 32 months.

Im not sure about a good ratio radiators/handyman and average time for breakdown.
Do you think 5 years with 5 Radiators per Handyman is a good choice? Or should there be more Radiators per Handyman? Are 5 years to much? Is one or two years better?

TheCycoONE:
In my uninformed opinion both of the time frames are too long. Boiler breakdowns were a semiregular occurrence and most levels can be easily beaten before 5 years.

[...]

PR #1345, opened 2018-02-04
[WIP] boiler breakdown depends on handymans

da4id:
As suggested in #490 its like in the original game that 8 radiators per handyman will stop boiler breakdown
Boiler will now breakdown every 200 days.


MarkL1961:
If you are adding `  self.world.map.level_config.gbv.DisasterLaunch  ` can you look at adding the vomit wave at the same time?
Patients can already vomit, but from what I remember from TH it was like everyone had and it was hard to clean up.
There is some code in patient I think that should trigger one, but I have never seen it happen.  Logic says one would get triggered based on the current state of cleanliness, perhaps @mugmuggy can tell from TH what was being used?

da4id:
Ive added it now. On the custom maps this value is not available therefore i added 200 days per default if not defined. I will implement this function later. 
I know, on the orignal game this was really hard to control and you had to hire all handymans. This made the game much much more difficult. 
I started a new Issue #1347

MarkL1961:
It is something that could be added to custom maps in the future and map makers can update their maps. Good idea to add a fall back just in case.
Have you decided on what the trigger is going to be or is that what the other issue is for?

da4id:
The trigger is gbv.disasterLaunch with 50% chance of boiler breakdown, 25% chance of vomit wave and 25% chance that nothing happens. This makes shure that the player dont get hit by both at once.

MarkL1961:
So is it just random then?
With the boiler you said you can mitigate the chances by having more handymen per radiator, is there not something that can mitigate the vomit wave i.e. less rubbish/vomit/dead rats per tile means less or no chance and greater than a certain ratio/percentage of tiles with it is more likely to happen?

TheCycoONE:
@MarkL1961 he implemented the same logic that @mugmuggy mentioned, where the length and whether it happened depends on the handyman / radiator ratio.

.gitignore
```diff
@@ -95,3 +95,4 @@ CorsixTH/CorsixTH.sln
 
 # Default build path
 CorsixTH/CorsixTH
+/.metadata/
```
TheCycoONE: What is `.metadata`?

CorsixTH/Lua/hospital.lua
```diff
@@ -1005,31 +1015,44 @@ function Hospital:onEndDay()
...
+    if math.random(1,3) ~= 1 then
+      --check if boiler can break
+      if not self.heating_broke and self.boiler_can_break and radiators > 0 and 8 * handymancount < radiators then
         self:boilerBreakdown()
       end
-    else
-      self:boilerBreakdown()
     end
+  else
+    self.days_until_disaster = self.days_until_disaster -1
```
TheCycoONE: Space around `-`

MarkL1961:
@TheCycoONE I am not talking about boiler breakdowns :)

.gitignore
da4id: Eclipse projekt data
you can ignore this

pschoffer:
@da4id have you thought about moving boiler and radiator logic into its own class? I think that the hospital class is getting pretty big now. If you would be up for it and move all that disaster, radiator price and all that logic to dedicated class (could be called `Boiler` or `BoilerController`). Even awesomer would be if you would add some unittests around the new class (as testing these big classes is much more challenging). Just a thought though so feel free to ignore:)

da4id:
I don't know how its implemented in the original game.

I think it could be something similar to epidemics. If a person vomit the persons next to this persons get infected by this and do the same after infection time or something. This is more a self running chain reaction which get triggered by disaster launch and keep going until infection stopped by some criteria like people have enough space between each other, handyman's cleans mess up or people get cured.
maybe there is some other criteria to stop

Thats what i observed from the original one

@pschoffer 
Good point, another name could be `HeatingController` including breakdown, costs, heat spread and so on. Maybe a second class `disasterLauncher` for breakdown trigger and vomit waves.

da4id:
I moved all the boiler code from hospital.lua to new heatingController class. Some variables like `self.radiator_heat` or `self.heating_broke` are accessed from outside hospital.lua to control the radiator heat or to see if the heating works. From which class is this controlled? Another problem is that these variables are restored after saved game is loaded. What is the nicest way to pass these variables after load from hospital.lua to heatingController in order to keep it compatible with older save games? New safegames does not need this because they can store it directly in heatingController. When is `Hospital:afterLoad(old, new)` called? What is stored in old and new?

I think its not a good solutions to keep these variables in hospital.lua because they need to get passed from and to heatingcontroller all the time

TheCycoONE:
afterLoad gets called immediately after loading a save game. Old is the save game version of the file and new is the current save game version. When making changes to the lua representation you bump the SAVEGAME_VERSION in app, and put in a condition to do the conversion in the relevant classes.

da4id:
Wow nice feature. Another question: In which class is the heat spreading implemented (Temperature of Field xy)? How does this class know what the actual radiators_heat is?

TheCycoONE:
@da4id the heat spreading is in C++, in th_map.cpp.
It's called in world.lua:1041 `self.map.th:updateTemperatures()`

It is passed `0.25 + self.hospitals[1].radiator_heat * 0.3`

TheCycoONE:
The magic numbers there are never explained, which is not so nice.

da4id:
Alright. So i dont need to move this code to heatingController.lua. When i update the line in world.lua it should work

da4id:
Checking CorsixTH/Lua/HeatingController.lua       1 warning
    CorsixTH/Lua/HeatingController.lua:85:42: (W212) unused argument new
Checking CorsixTH/Lua/hospital.lua                2 warnings
    CorsixTH/Lua/hospital.lua:97:28: (W113) accessing undefined variable HeatingController
    CorsixTH/Lua/hospital.lua:640:30: (W113) accessing undefined variable HeatingController
Total: 3 warnings / 0 errors in 288 files

How can i fix these warnings?

CorsixTH/Lua/HeatingController.lua
```diff
@@ -0,0 +1,182 @@
...
+  --if disasterLaunch not defined in Levelfile
+  if self.days_until_boiler_breakdown == nil then
+    self.days_until_boiler_breakdown = 200
+  end
+end
+
+function HeatingController:checkHeatingFacilities(day,averagePatiensWarmth, averageStaffWarmth)
+  if not self.warmth_msg and not self.heating_broke then
+    if day == 15 then
+      if averagePatiensWarmth < 0.22 then
```
TheCycoONE: Can you move these magic numbers to 'constants, e.g. `local COLD_TOLERANCE_TEMP = 0.22`

CorsixTH/Lua/dialogs/fullscreen/town_map.lua
```diff
@@ -186,15 +186,15 @@ function UITownMap:draw(canvas, x, y)
   self.info_font:draw(canvas, radiators,    x +  95, y + 265)
 
   -- Heating costs
-  local heating_costs = math.floor(((hospital.radiator_heat *10)* radiators)* 7.5)
+  local heating_costs = math.floor(((hospital.heatingController.radiator_heat *10)* radiators)* 7.5)
```
TheCycoONE: Space on both sides of * (formatting was wrong before)

CorsixTH/Lua/hospital.lua
```diff
@@ -579,17 +554,12 @@ function Hospital:afterLoad(old, new)
       }
     end
   end
-  if old < 39 then
```
TheCycoONE: Do not remove old save conditions

TheCycoONE: Nm I see you just moved it. Preferably don't.

CorsixTH/Lua/hospital.lua
```diff
@@ -751,28 +741,7 @@ function Hospital:checkFacilities()
...
-      if day == 20 then
-        local avgWarmth = self:getAverageStaffAttribute("warmth", 0.25) -- Default value does not result in a message.
-        if avgWarmth < 0.22 then
-          self.world.ui.adviser:say(_A.warnings.staff_very_cold)
-        elseif avgWarmth >= 0.36 then
-          self.world.ui.adviser:say(_A.warnings.staff_too_hot)
-        end
-      end
-    end
+    self.heatingController:checkHeatingFacilities(day,self:getAveragePatientAttribute("warmth", 0.3),self:getAverageStaffAttribute("warmth", 0.25))
```
TheCycoONE: Space after each , and split long lines.

CorsixTH/Lua/hospital.lua
```diff
@@ -791,7 +760,7 @@ function Hospital:checkFacilities()
       self.toilet_msg = false
       self.bench_msg = false
       self.cash_msg = false
-      self.warmth_msg = false
+      self.heatingController.warmth_msg = false
```
TheCycoONE: Hospital shouldn't directly set warmth_msg. Add a clearWarmthMsg setter.

CorsixTH/Lua/world.lua
```diff
@@ -1039,7 +1039,7 @@ function World:onTick()
       self.current_tick_entity = nil
       self.map:onTick()
       self.map.th:updateTemperatures(outside_temperatures[self.game_date:monthOfYear()],
-          0.25 + self.hospitals[1].radiator_heat * 0.3)
+          0.25 + self.hospitals[1].heatingController.radiator_heat * 0.3)
```
TheCycoONE: Add getter

CorsixTH/Lua/HeatingController.lua
```diff
@@ -0,0 +1,182 @@
...
+  end
+
+  -- Calculate heating cost daily.  Divide the monthly cost by the number of days in that month
+  local heating_costs = (((self.radiator_heat * 10) * radiators) * 7.50) / self.world:date():lastDayOfMonth()
+  self.acc_heating = self.acc_heating + heating_costs
+end
+
+function HeatingController:getHeatingCostsForActualMonth()
+  local monthlyCosts = math.round(self.acc_heating)
+  self.acc_heating = 0
```
TheCycoONE: Getter should not have side effects. Reset acc_heating in setter or onEndMonth method.

CorsixTH/Lua/HeatingController.lua
```diff
@@ -0,0 +1,182 @@
...
+  end
+end
+
+function HeatingController:afterLoad(old,new)
+  if old < 125 then
+    self:resetBoilerBreakdownCounter()
+  end
+end
+
+function HeatingController:coldWarning(isPlayerHospital)
```
TheCycoONE: The test for announcements and the argument are redundant. These methods should be private (prefix with _ or specify outside of the class) and should not take isPlayerHospital.

CorsixTH/Lua/HeatingController.lua
```diff
@@ -0,0 +1,182 @@
...
+
+function HeatingController:HeatingController(world)
+  self.world = world
+  self.radiator_heat = 0.5
+  self:resetBoilerBreakdownCounter()
+  self.acc_heating = 0
+end
+
+function HeatingController:resetBoilerBreakdownCounter()
+  self.days_until_boiler_breakdown = self.world.map.level_config.gbv.DisasterLaunch
```
TheCycoONE: You can use `or 200` to specify the default. Preferably move the default to a file local const.

CorsixTH/Lua/HeatingController.lua
```diff
@@ -0,0 +1,182 @@
...
+  end
+end
+
+function HeatingController:warningTooHot()
+  local hot_msg = {
+    (_A.information.initial_general_advice.decrease_heating),
+    (_A.warnings.patients_too_hot),
+    (_A.warnings.patients_getting_hot),
+  }
+  if hot_msg then
```
TheCycoONE: Test is always true, remove.

TheCycoONE:
Add HeatingController to .luacheckrc globals list.

CorsixTH/Lua/dialogs/fullscreen/town_map.lua
da4id: I created a function for calculating the expected heating costs. I need the same calculation in the onEndDay and town_map

CorsixTH/Lua/HeatingController.lua
da4id: They need isPlayerHospital otherwise adviser will show messages

CorsixTH/Lua/HeatingController.lua
```diff
@@ -0,0 +1,241 @@
...
+        if isPlayerHospital == true then
+          self.world.ui.adviser:say(_A.warnings.staff_too_hot)
+        end
+        self.warmth_msg = true
+      end
+    end
+  end
+end
+
+function HeatingController:warningTooCold(isPlayerHospital)
```
TheCycoONE: Remove the isPlayerHospital argument here; it's not useful. Just don't call this method if you don't want to display anything.

Actually I don't think this advisor stuff (or anything that depends on whether it's the player hospital) belongs in this class. I think the class should just expose whether it's too hot or cold.

TheCycoONE:
@da4id Are you still interested in pursuing this? Perhaps in a more limited fashion without the refactor?

da4id:
@TheCycoONE Ou sorry, I was very busy the last months and forgot this pullrequest. I will do the refactoring this week.

lewri:
@da4id just going through stuff. Do you think you are still interested in pursuing this?

Alberth289346:
"Next week" in 2018 clearly didn't work, it seems safe to assume the original author will not work on it.

lewri:
Superseded by #1709

PR #1346, opened 2018-02-04
[RDY]67 moving clears usage

### LINKED ISSUE #67: Moving/replacing the equipment inside a room resets the usage counter of machine
1. A room with a hardware that used many times. and the status of hardware is not good.
2. Edit the room, move to other location, save.

Result:
Everything reset.

pschoffer:
Fixes #67

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -614,10 +614,12 @@ function UIEditRoom:returnToDoorPhase()
           obj.remove()
           break
         end
+        local obj_state = obj:getState()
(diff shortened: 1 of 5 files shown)


CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -618,6 +618,7 @@ function UIEditRoom:returnToDoorPhase()
         if not obj.master then
           self:addObjects({{
             object = TheApp.objects[obj.object_type.id],
+            state = obj:getState(),
```
TheCycoONE: You might want to pull the state before the object is destroyed, in case the destruction breaks the state?

PR #1352, opened 2018-02-16
[RDY] Enum modernization

TheCycoONE:
I reviewed all enum usage in CorsixTH and converted to `enum class` and/or added type specifiers as appropriate for improved type safety, and ease of analysis.

### DIFF
--- a/CorsixTH/Src/iso_fs.cpp
+++ b/CorsixTH/Src/iso_fs.cpp
@@ -63,7 +63,7 @@ void IsoFilesystem::setPathSeparator(char cSeparator)
     m_cPathSeparator = cSeparator;
 }
 
-enum IsoVolumeDescriptorType
+enum IsoVolumeDescriptorType : uint8_t
 {
     VDT_PRIVARY_VOLUME = 0x01,
     // Other type numbers are either reserved for future use, or are not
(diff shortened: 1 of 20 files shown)


Alberth289346:
Conversion of alignment string to enum value seems to be duplicated, and enumeration class names are not consistent, but both are out of scope for this patch I think.

TheCycoONE:
Yes, I noticed both issues too. I still don't really know what naming convention we want for classes, and haven't settled on one yet.

```
iso_names_their_classes_like_this
Stroustrop_names_his_classes_like_this
JavaStyleIsPopular
THPrefixesAreUsedBySomeVendorsLikeApple
```

Alberth289346:
I always use CamelCase for classes. Prefixes are silly, use a namespace if you're worried the code will be used in another library with clashing names (ie not going to happen is my guess).

I use lower_case for variables. Functions/methods are always a bit of a problem, I often use CamelCase too, as they have an active form DoSomething, while classes usually have a more passive wording like DataStore. My alternative form for functions is like_variables. notInitialCamelCase for functions could work too, but I find it less pretty, not sure why.

I would likely write enum values AS_ALL_UPPERCASE, as they are constants to me, but that may look bad together with the forced prefix.

TheCycoONE:
Strostrup and the C++ Core Guidelines have strong objections to ALL_CAPS being used for anything but preprocessor macros. I agree with their reasoning.

Alberth289346:
Except for the standard #include protection, I don't use preprocessor macros in c++. In c, I only use it for plain expressions, that should be 'inline' functions, but c89 (yes, indeed, that old :( ) doesn't have that.

PR #1356, opened 2018-02-19
[RFC] Fix VIP behaviour about hospitals with no functioning rooms

### LINKED ISSUE #1115: Vip rating computation on average queue length is incorrect
In the VIP evaluation code, the queue length is summed over all non-exploded rooms in the world, and then divided by the number of existing rooms in the world.

Clearly this is not how you would compute queue length, it should not take exploded rooms into account at all (number of existing rooms includes them).

Secondly, the VIP is evaluating one hospital only, so only rooms of the visited hospital should be taken into account. (Of course in single player, this makes no difference as there are no other hospitals)

tobylane:
Fixes #1115, counting only usable rooms for average queue size. Fixes the comment in that function about hospitals with no functioning rooms. The -100 rating is arbitrary but as no-one would expect a positive review of a hospital in that state does it seems fair?

### DIFF
--- a/CorsixTH/Lua/entities/vip.lua
+++ b/CorsixTH/Lua/entities/vip.lua
@@ -295,17 +295,22 @@ function Vip:setVIPRating()
 
   -- check for the average queue length
   local sum_queue = 0
+  local room_count = 0
   for _, room in pairs(self.world.rooms) do
-    -- this can be nil if there has been a room explosion
     if room.door.queue then
       sum_queue = sum_queue + room.door.queue:size()
     end
+    if not room.crashed then
+      room_count = room_count + 1
+    end
   end
 
-  if sum_queue == 0 then
+  if room_count == 0 then
+    self.vip_rating = self.vip_rating - 100
+  elseif sum_queue == 0 then
     self.vip_rating = self.vip_rating + 6
   else
-    local queue_ratio = sum_queue / #self.world.rooms
+    local queue_ratio = sum_queue / room_count
     local queue_ratio_rangemap = {
       {upper = 2, value = 6},
       {upper = 5, value = 3},


CorsixTH/Lua/entities/vip.lua
```diff
@@ -294,26 +294,31 @@ function Vip:setVIPRating()
...
-    if room.door.queue then
-      sum_queue = sum_queue + room.door.queue:size()
+  if #self.world.rooms ~= 0 then
+    local sum_queue = 0
+    local room_count = 0
+    for _, room in pairs(self.world.rooms) do
+      if room.door.queue then
+        sum_queue = sum_queue + room.door.queue:size()
+        room_count = room_count + 1
+      end
```
Alberth289346: The game seems to use `room.door.queue` as "this is a valid room" condition rather than `#self.world.rooms ~= 0`. In other words, your change opens the option that there are rooms in the world, but none has a queue, and thus none is counted.
Wouldn't it be simpler to keep this counting part as it was, and instead use `room_count` in the vip-rating above your new line 307 (if not rooms -> -100, elif no queue -> +6, else -> complicated calculation) ?

tobylane: It would, reverted back to an earlier version of the loop.

Alberth289346:
I am confused, the overall `if #self.world.rooms ~= 0` is still there.

You do realize that it's an unneeded check right? If there are no rooms in the world, the loop will never iterate, so `room_count` will be `0`. Checking that after counting will not only cover having no rooms in the world, but also cover rooms without `door.queue`, or only having crashed rooms, as what the current patch seems to exclude now as well.

tobylane:
When there are no functioning rooms in the hospital the loop though self.world.rooms will cause a crash. In testing this I made it print the name of each room looked at and the numbers in the ratio calculation, it was the whole hospital minus two crashed rooms.

Alberth289346:
Yes, `a / b` where `b` can be 0 is always dangerous :)

I was suggesting
```
for room in world.rooms do
  if room "is a valid room" then room_count = room_count + 1 end
end
-- room_count will be 0 if self,world.rooms == 0 or if no valid room exists.

if room_count == 0 then  -- viprating is -100
else if queue size == 0 then -- viprating is 6
else -- do ratio calculation
end
```
In that way you get -100 for no functioning rooms (which is more general than no rooms at all worldwide), and avoid the crash at the same time.

tobylane:
Thanks. I don't remember how I caused the problem (`bad argument #1 to 'for iterator' (table expected, got nil)`) needing that check but all the test savegames run fine now.

Alberth289346:
Indenting looks horribly wrong, don't use tab, use 2 spaces for indenting instead

PR #1360, opened 2018-03-01
[RDY] Add configurable option for room information dialogs

### LINKED ISSUE #337: debug mode changes game behaviour (hides first built room information)
This behaviour was intentional, it was introduced in 41c8b342b9ba162d1b2c4f1552cb4588ffd8279d but my feeling is that it's an unexpected / seemingly unrelated functionality change from the act of debugging, and in the case of #256 actually made debugging harder because toggling information off prevented the error.

@MarkL1961 mentioned that this had been discussed in google issues.  I haven't been able to find the issue in a quick skim, but if someone can find it there it would be interesting to see the reasons why this change was introduced in the first place.

My recommendation at this point is to introduce a config option specifically for whether to show first time room information dialogs, and to disassociate it with the debug flag.

tobylane:
Fixes #337

### DIFF
--- a/CorsixTH/Lua/config_finder.lua
+++ b/CorsixTH/Lua/config_finder.lua
@@ -138,6 +138,7 @@ local config_defaults = {
   new_graphics_folder = nil,
   use_new_graphics = false,
   check_for_updates = true,
+  room_information_dialogs = true
 }
 
 fi = io.open(config_filename, "r")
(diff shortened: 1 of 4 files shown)


tobylane:
I messed up my testing. This is already fixed.

Alberth289346:
Hmm, and I liked your positive name "room_information_dialogs" instead of the negated form "room_information_dialogs_off" which just complicates matters without good reason ;)

TheCycoONE:
@tobylane why did you think this was already resolved. It still depends on `debug`. Can we reopen this?

tobylane:
Fixed and ready.

Issue #84, opened 2013-10-29
Adding english names to all languages in the menu


Grimdoc:
As I mentioned in issue 1719 in the Google Project page:

Should we maybe add the english name for the languages to the menu so that everyone can easily identify the languages?
For example I can't tell chinese from korean, but it could sometimes be handy to know which is which.

Example of what the menu might would look like:
Castellano (Spanish)
Dansk (Danish)
Deutsch (German)
 Etc..

What do you think?

TheCycoONE:
Wouldn't it be more appropriate for the bracketed or tool-tip language to be the player's current language instead of always English?

bobcanthelpyou:
at the moment it looks like this:
![bildschirmfoto von 2015-07-23 00 58 53](https://cloud.githubusercontent.com/assets/4523264/8839526/96457b42-30d6-11e5-9c25-7ee3b7ffd1cd.png)

And i think it is a good idea to keep it, you could always switch to the languages you could read.

The idea of @TheCycoONE is also okay, to add the translation of the actually chosen language to the tool-tip or the brackets of the languages. 

But what is the befit of adding a translation to it? Would you choose a language if you can't read the origin name? 'Ah of cause it Korean let me chose that!'

lewri:
PR 1362 needs linking to this issue (i.e. fixes xxxx)

PR #1362, opened 2018-03-06
[RDY] Add English names of languages beside the native name in options

tobylane:
The change in strings.lua affects https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/app.lua#L721 and https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/app.lua#L779 but I'm not sure what to do.

The longer text spills well outside the box, but it looked bad simply widening the background element. How do you want it to look?

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -730,7 +730,7 @@ function App:dumpStrings()
 
   self:checkMissingStringsInLanguage(dir, self.config.language)
   -- Uncomment these lines to get diffs for all languages in the game
-  -- for _, lang in ipairs(self.strings.languages_english) do
+  -- for _, lang in pairs(self.strings.languages_english) do
   --   self:checkMissingStringsInLanguage(dir, lang)
   -- end
   print("")
@@ -746,7 +746,6 @@ end
 --!param dir The directory where the file to write to should be.
 --!param language The language to check against.
 function App:checkMissingStringsInLanguage(dir, language)
-
   -- Accessors to reach through the userdata proxies on strings
   local LUT = debug.getregistry().StringProxyValues
   local function val(o)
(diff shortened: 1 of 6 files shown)


MarkL1961:
What about doing this as an hover text as those only show when you hover and can expand easily ?

tobylane:
```lua
langs[#langs + 1] = {text = lang, font = font, tooltip = _S.tooltip.options_window.language_dropdown_item:format(lang, eng_name)}
```

That would look like this. I don't get that tooltip, not even with a new config.txt or without my changes.

MarkL1961:
Does the tool tip work on hover the button or the name label?  It only needs to work from one of them and you could use the other

MarkL1961:
I have just looked at the game to see what you are doing and I can see the problem with my suggestion.
Only the selected language as a hover text.
You could widen the menu and then widen the text area of the dropdown to show the extended text

TheCycoONE:
Linking to the relevant issue and discussion: #84

MarkL1961:
From the original issue it is clear some can't see the benefit of this, but let's suppose your pc lacks the correct font and all you see is squares or you accidently select the wrong language it is useful to see the language selected and then perhaps the English name to correct your mistake..

Yes you can correct it in config, but how may times have we had someone not know where to find it?

Tooltips is deactivated, can that be reversed so the option stated above becomes a solution?
Can we widen the drop down to show more text?  (also suggested above)
What about undo so that you can go back to the way things were?
Does someone have another idea?

tobylane:
This enables tooltips with sensible defaults, which are also used for the difficulty setting of a new campaign.

TheCycoONE:
Code looks good to me @tobylane - is this RDY?

tobylane:
It is, sorry for the delay.

TheCycoONE:
No worries, holding for after the release. Make sure I remember right after :smile:

Issue #1329, opened 2018-01-15
Handyman Stuck forever (also happened with the Grim Reaper)


lloyd-lejack:
Hey LloydJara here! I was testing a custom map i made for the current version of CorsixTH (v0.61) and got a bug where a Handyman (and in a past save i currently don't have, the Death shortly before a patient died) got stuck forever on the hospital doing an action. Not sure if a similar ticket exists but i wanted to post this regardless.
I have no save prior to the bug, but i have the savegame when my Handyman got stuck forever while cleaning trash. The Death on the other hand, got stuck after letting the patient go to helll. Both him and his lava got stuck forever in the Hospital.

Here's the mentioned save: 
https://dl.dropboxusercontent.com/s/renpn3vye1lagfz/Doctor%20Frog.sav?dl=0

My Operating System is Windows 10. Not sure if this behavior tends to happen on older versions.

https://prnt.sc/i0jcrv Here's also a pic showing the handyman.

#### Gamelog.txt
If im not wrong, this is the gamelog of the bug:

Created game with savegame version 122.
Error in timer handler: 
C:\Program Files\CorsixTH\Lua\world.lua:2134: attempt to perform arithmetic on a nil value (local 'y')
stack traceback:
	C:\Program Files\CorsixTH\Lua\world.lua:2134: in method 'removeObjectFromTile'
	C:\Program Files\CorsixTH\Lua\objects\litter.lua:99: in method 'remove'
	...gram Files\CorsixTH\Lua\humanoid_actions\sweep_floor.lua:45: in local 'timer_function'
	C:\Program Files\CorsixTH\Lua\entity.lua:216: in field 'tick'
	C:\Program Files\CorsixTH\Lua\entities\staff.lua:149: in method 'tick'
	C:\Program Files\CorsixTH\Lua\world.lua:1051: in method 'onTick'
	C:\Program Files\CorsixTH\Lua\app.lua:1008: in function <C:\Program Files\CorsixTH\Lua\app.lua:1005>
	(...tail calls...)
	C:\Program Files\CorsixTH\Lua\app.lua:909: in function <C:\Program Files\CorsixTH\Lua\app.lua:904>

LloydJara:
Sadly i dont have the save right before the Handyman got suck. Not sure if i can replicate if perfectly but i could try.

stavrossk:
Hi, is CorsixTH playable in it's current state?

Thanks,
Stavros.

LloydJara:
[...]

PR #1363, opened 2018-03-07
[RDY] Prevent double remove on same litter object

mugmuggy:
Provides a quick fix for handyman issue in #1329. I actually recall having this change in there previously, but can't recall testing this exact scenario, alas at somepoint I had removed it and forgot to put it back in.

Appears to be a more extensive issue of placing objects where a humanoid is already standing which was prevented in original.


mugmuggy:
Added log spam

TheCycoONE:
LGTM
I'll merge if you are ready

PR #1370, opened 2018-03-27
[RDY] Naming & Layout Part 1

TheCycoONE:
This pull request updates CorsixTH to consistently use ISO standard naming conventions as part of the effort to conform to the C++ Core Guidelines.

In particular, type prefixes have been removed from all class variables as per: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#nl5-avoid-encoding-type-information-in-names

ISO Standard style class and function names are used throughout, which conforms to:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#nl8-use-a-consistent-naming-style
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#nl10-avoid-camelcase

Constants have also been changed to lower snake case conforming to:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#nl9-use-all_caps-for-macro-names-only

Future pull requests in this series are planned.

### DIFF
--- a/CorsixTH/Src/iso_fs.cpp
+++ b/CorsixTH/Src/iso_fs.cpp
@@ -27,58 +27,58 @@ SOFTWARE.
 #include <vector>
 #include <algorithm>
 
-IsoFilesystem::IsoFilesystem()
+iso_filesystem::iso_filesystem()
 {
-    m_fRawFile = nullptr;
-    m_sError = nullptr;
-    m_pFiles = nullptr;
-    m_iNumFiles = 0;
-    m_iFileTableSize = 0;
-    m_cPathSeparator = '\\';
+    raw_file = nullptr;
+    error = nullptr;
+    files = nullptr;
+    file_count = 0;
+    file_table_size = 0;
+    path_seperator = '\\';
 }
 
-IsoFilesystem::~IsoFilesystem()
+iso_filesystem::~iso_filesystem()
 {
-    _clear();
+    clear();
 }
 
-void IsoFilesystem::_clear()
+void iso_filesystem::clear()
 {
-    delete[] m_sError;
-    m_sError = nullptr;
+    delete[] error;
+    error = nullptr;
 
-    if(m_pFiles)
+    if(files)
     {
-        for(size_t i = 0; i < m_iNumFiles; ++i)
-            delete[] m_pFiles[i].sPath;
-        delete[] m_pFiles;
-        m_pFiles = nullptr;
-        m_iNumFiles = 0;
-        m_iFileTableSize = 0;
+        for(size_t i = 0; i < file_count; ++i)
+            delete[] files[i].path;
+        delete[] files;
+        files = nullptr;
+        file_count = 0;
+        file_table_size = 0;
     }
 }
 
-void IsoFilesystem::setPathSeparator(char cSeparator)
+void iso_filesystem::set_path_separator(char cSeparator)
 {
-    m_cPathSeparator = cSeparator;
+    path_seperator = cSeparator;
 }
 
-enum IsoVolumeDescriptorType : uint8_t
+enum iso_volume_descriptor_type : uint8_t
 {
-    VDT_PRIVARY_VOLUME = 0x01,
+    vdt_privary_volume = 0x01,
     // Other type numbers are either reserved for future use, or are not
     // interesting to us.
-    VDT_TERMINATOR = 0xFF,
+    vdt_terminator = 0xFF,
 };
 
-enum IsoDirEntFlag : uint8_t
+enum iso_dir_ent_flag : uint8_t
 {
-    DEF_HIDDEN = 0x01,
-    DEF_DIRECTORY = 0x02,
-    DEF_MULTI_EXTENT = 0x80,
+    def_hidden = 0x01,
+    def_directory = 0x02,
+    def_multi_extent = 0x80,
 };
 
-template <class T> static inline T ReadNativeInt(const uint8_t *p)
+template <class T> static inline T read_native_int(const uint8_t *p)
 {
     // ISO 9660 commonly encodes multi-byte integers as little endian followed
     // by big endian. Note that the first byte of iEndianness will be a zero on
@@ -87,31 +87,31 @@ template <class T> static inline T ReadNativeInt(const uint8_t *p)
     return reinterpret_cast<const T*>(p)[*reinterpret_cast<const uint8_t*>(&iEndianness)];
 }
 
-bool IsoFilesystem::initialise(FILE* fRawFile)
+bool iso_filesystem::initialise(FILE* fRawFile)
 {
-    m_fRawFile = fRawFile;
-    _clear();
+    raw_file = fRawFile;
+    clear();
 
     // Until we know better, assume that sectors are 2048 bytes.
-    m_iSectorSize = 2048;
+    sector_size = 2048;
 
     // The first 16 sectors are reserved for bootable media.
     // Volume descriptor records follow this, with one record per sector.
-    for(uint32_t iSector = 16; _seekToSector(iSector); ++iSector)
+    for(uint32_t iSector = 16; seek_to_sector(iSector); ++iSector)
     {
         uint8_t aBuffer[190];
-        if(!_readData(sizeof(aBuffer), aBuffer))
+        if(!read_data(sizeof(aBuffer), aBuffer))
             break;
         // CD001 is a standard identifier, \x01 is a version number
         if(std::memcmp(aBuffer + 1, "CD001\x01", 6) == 0)
         {
-            if(aBuffer[0] == VDT_PRIVARY_VOLUME)
+            if(aBuffer[0] == vdt_privary_volume)
             {
-                m_iSectorSize = ReadNativeInt<uint16_t>(aBuffer + 128);
-                _findHospDirectory(aBuffer + 156, 34, 0);
-                if(m_iNumFiles == 0)
+                sector_size = read_native_int<uint16_t>(aBuffer + 128);
+                find_hosp_directory(aBuffer + 156, 34, 0);
+                if(file_count == 0)
                 {
-                    _setError("Could not find Theme Hospital data directory.");
+                    set_error("Could not find Theme Hospital data directory.");
                     return false;
                 }
                 else
@@ -119,22 +119,22 @@ bool IsoFilesystem::initialise(FILE* fRawFile)
                     return true;
                 }
             }
-            else if(aBuffer[0] == VDT_TERMINATOR)
+            else if(aBuffer[0] == vdt_terminator)
                 break;
         }
     }
-    _setError("Could not find primary volume descriptor.");
+    set_error("Could not find primary volume descriptor.");
     return false;
 }
 
-int IsoFilesystem::_fileNameComp(const void* lhs, const void* rhs)
+int iso_filesystem::filename_compare(const void* lhs, const void* rhs)
 {
     return std::strcmp(
-        reinterpret_cast<const _file_t*>(lhs)->sPath,
-        reinterpret_cast<const _file_t*>(rhs)->sPath);
+        reinterpret_cast<const file_metadata*>(lhs)->path,
+        reinterpret_cast<const file_metadata*>(rhs)->path);
 }
 
-char IsoFilesystem::_normalise(char c)
+char iso_filesystem::normalise(char c)
 {
     if(c == '_') // underscore to hyphen
         return '-';
@@ -144,7 +144,7 @@ char IsoFilesystem::_normalise(char c)
         return c;
 }
 
-void IsoFilesystem::_trimIdentifierVersion(const uint8_t* sIdent, uint8_t& iLength)
+void iso_filesystem::trim_identifier_version(const uint8_t* sIdent, uint8_t& iLength)
 {
     for(uint8_t i = 0; i < iLength; ++i)
     {
@@ -156,7 +156,7 @@ void IsoFilesystem::_trimIdentifierVersion(const uint8_t* sIdent, uint8_t& iLeng
     }
 }
 
-int IsoFilesystem::_findHospDirectory(const uint8_t *pDirEnt, int iDirEntsSize, int iLevel)
+int iso_filesystem::find_hosp_directory(const uint8_t *pDirEnt, int iDirEntsSize, int iLevel)
 {
     // Sanity check
     // Apart from at the root level, directory record arrays must take up whole
@@ -176,12 +176,12 @@ int IsoFilesystem::_findHospDirectory(const uint8_t *pDirEnt, int iDirEntsSize, 
             continue;
         }
 
-        uint32_t iDataSector = ReadNativeInt<uint32_t>(pDirEnt + 2);
-        uint32_t iDataLength = ReadNativeInt<uint32_t>(pDirEnt + 10);
+        uint32_t iDataSector = read_native_int<uint32_t>(pDirEnt + 2);
+        uint32_t iDataLength = read_native_int<uint32_t>(pDirEnt + 10);
         uint8_t iFlags = pDirEnt[25];
         uint8_t iIdentLength = pDirEnt[32];
-        _trimIdentifierVersion(pDirEnt + 33, iIdentLength);
-        if(iFlags & DEF_DIRECTORY)
+        trim_identifier_version(pDirEnt + 33, iIdentLength);
+        if(iFlags & def_directory)
         {
             // The names "\x00" and "\x01" are used for the current directory
             // the parent directory respectively. We only want to visit these
@@ -194,13 +194,13 @@ int IsoFilesystem::_findHospDirectory(const uint8_t *pDirEnt, int iDirEntsSize, 
                     iBufferSize = iDataLength;
                     pBuffer = new uint8_t[iBufferSize];
                 }
-                if(_seekToSector(iDataSector) && _readData(iDataLength, pBuffer))
+                if(seek_to_sector(iDataSector) && read_data(iDataLength, pBuffer))
                 {
-                    int iFoundLevel = _findHospDirectory(pBuffer, iDataLength, iLevel + 1);
+                    int iFoundLevel = find_hosp_directory(pBuffer, iDataLength, iLevel + 1);
                     if(iFoundLevel != 0)
                     {
                         if(iFoundLevel == 2)
-                            _buildFileLookupTable(iDataSector, iDataLength, "");
+                            build_file_lookup_table(iDataSector, iDataLength, "");
                         delete[] pBuffer;
                         return iFoundLevel + 1;
                     }
@@ -217,7 +217,7 @@ int IsoFilesystem::_findHospDirectory(const uint8_t *pDirEnt, int iDirEntsSize, 
                 int i = 0;
                 for(; i < 10; ++i)
                 {
-                    if(_normalise(pDirEnt[33 + i]) != sName[i])
+                    if(normalise(pDirEnt[33 + i]) != sName[i])
                         break;
                 }
                 if(i == 10)
@@ -232,7 +232,7 @@ int IsoFilesystem::_findHospDirectory(const uint8_t *pDirEnt, int iDirEntsSize, 
     return 0;
 }
 
-void IsoFilesystem::_buildFileLookupTable(uint32_t iSector, int iDirEntsSize, const char* sPrefix)
+void iso_filesystem::build_file_lookup_table(uint32_t iSector, int iDirEntsSize, const char* sPrefix)
 {
     // Sanity check
     // Apart from at the root level, directory record arrays must take up whole
@@ -244,7 +244,7 @@ void IsoFilesystem::_buildFileLookupTable(uint32_t iSector, int iDirEntsSize, co
         return;
 
     uint8_t *pBuffer = new uint8_t[iDirEntsSize];
-    if(!_seekToSector(iSector) || !_readData(iDirEntsSize, pBuffer))
+    if(!seek_to_sector(iSector) || !read_data(iDirEntsSize, pBuffer))
     {
         delete[] pBuffer;
         return;
@@ -259,11 +259,11 @@ void IsoFilesystem::_buildFileLookupTable(uint32_t iSector, int iDirEntsSize, co
             continue;
         }
 
-        uint32_t iDataSector = ReadNativeInt<uint32_t>(pDirEnt + 2);
-        uint32_t iDataLength = ReadNativeInt<uint32_t>(pDirEnt + 10);
+        uint32_t iDataSector = read_native_int<uint32_t>(pDirEnt + 2);
+        uint32_t iDataLength = read_native_int<uint32_t>(pDirEnt + 10);
         uint8_t iFlags = pDirEnt[25];
         uint8_t iIdentLength = pDirEnt[32];
-        _trimIdentifierVersion(pDirEnt + 33, iIdentLength);
+        trim_identifier_version(pDirEnt + 33, iIdentLength);
 
         // Build new path
         char *sPath = new char[iLen + iIdentLength + 2];
@@ -271,29 +271,29 @@ void IsoFilesystem::_buildFileLookupTable(uint32_t iSector, int iDirEntsSize, co
 #ifdef _MSC_VER
 #pragma warning(disable: 4996)
 #endif
-        std::transform(pDirEnt + 33, pDirEnt + 33 + iIdentLength, sPath + iLen, _normalise);
+        std::transform(pDirEnt + 33, pDirEnt + 33 + iIdentLength, sPath + iLen, normalise);
 #ifdef _MSC_VER
 #pragma warning(default: 4996)
 #endif
         sPath[iLen + iIdentLength] = 0;
 
-        if(iFlags & DEF_DIRECTORY)
+        if(iFlags & def_directory)
         {
             // None of the directories which we're interested in have length 1.
             // This also avoids the dummy "current" and "parent" directories.
             if(iIdentLength > 1)
             {
-                sPath[iLen + iIdentLength] = m_cPathSeparator;
+                sPath[iLen + iIdentLength] = path_seperator;
                 sPath[iLen + iIdentLength + 1] = 0;
-                _buildFileLookupTable(iDataSector, iDataLength, sPath);
+                build_file_lookup_table(iDataSector, iDataLength, sPath);
             }
         }
         else
         {
-            _file_t *pFile = _allocFileRecord();
-            pFile->sPath = sPath;
-            pFile->iSector = iDataSector;
-            pFile->iSize = iDataLength;
+            file_metadata *file = allocate_file_record();
+            file->path = sPath;
+            file->sector = iDataSector;
+            file->size = iDataLength;
             sPath = nullptr;
         }
         delete[] sPath;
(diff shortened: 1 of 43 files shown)


TheCycoONE:
MSVC failure is from doxygen error that is likely a simple oversight. g++ error is more fundamental; it seems to dislike variables with the same names as classes even when the classes are explicitly namespaced (allowed in clang and msvc)

TheCycoONE:
For reasons I'm not clear on, g++ was only concerned with one of the class / variable name collisions (map in map_scanline_iterator). Still plenty of complaints from MSVC though.

TheCycoONE:
Opening this up to bike shedding now.
It's probably the best time to chime in if you don't like any of the new names, or to point out that I missed something with an old style name.

Note:
`L` is still used as the name of the Lua_state variable - this is a lua convention.
`l_` is still used as a prefix for functions to be called from lua. This is temporary until those functions can be moved into another namespace.
Likewise with the `luaT_` prefix.

I did not touch local variable or parameter names in this PR. It's big enough without them.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -32,49 +32,49 @@ SOFTWARE.
...
-        m_pData = pData;
-        m_iLength = iLength;
+        data = pData;
+        remaining_bytes = iLength;
     }
 
-    const uint8_t *m_pData; ///< Pointer to the remaining data.
-    size_t m_iLength;       ///< Remaining number of bytes.
+    const uint8_t *data; ///< Pointer to the remaining data.
+    size_t remaining_bytes;       ///< Remaining number of bytes.
```
Alberth289346: Documentation comment alignment is messed up, that needs to be addressed at some point.
While I like vertical alignment, if we want to change convention, now is likely a good time.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -308,104 +308,104 @@ inline static void _setmin(int& iLeft, int iRight)
...
         iLoadedElements++;
         iNumElements--;
     }
     return iFirst;
 }
 
-size_t THAnimationManager::makeListElements(size_t iFirstElement, size_t iNumElements,
+size_t animation_manager::make_list_elements(size_t iFirstElement, size_t iNumElements,
                                             size_t &iLoadedListElements,
                                             size_t iListStart, size_t iListCount)
```
Alberth289346: Indenting of other arguments is incorrect now

CorsixTH/Src/th_gfx.cpp
```diff
@@ -663,179 +663,179 @@ bool THAnimationManager::loadCustomAnimations(const uint8_t* pData, size_t iData
...
-    *pY = m_vFrames[iFrame].iSecondaryMarkerY;
+    *pX = frames[iFrame].secondary_marker_x;
+    *pY = frames[iFrame].secondary_marker_y;
     return true;
 }
 
-bool THAnimationManager::hitTest(size_t iFrame, const THLayers_t& oLayers,
+bool animation_manager::hit_test(size_t iFrame, const ::layers& oLayers,
                                  int iX, int iY, uint32_t iFlags,
                                  int iTestX, int iTestY) const
```
Alberth289346: Ha, this works, just find a new name that is equally long!  :D

CorsixTH/Src/th_gfx.h
```diff
@@ -484,129 +489,129 @@ class THAnimationManager
...
+    void set_layers_from(const animation_base *pSrc) {layers = pSrc->layers;}
 
    // bool isMultipleFrameAnimation() { return false;}
 protected:
     //! X position on tile (not tile x-index)
-    int m_iX;
+    int x;
     //! Y position on tile (not tile y-index)
-    int m_iY;
+    int y;
```
Alberth289346: Perhaps a bit more expressing what the comment says, `xpos_on_tile` or `xpos_within_tile` or so?
Apparently this is currently confusing.

CorsixTH/Src/th_gfx.h
```diff
@@ -484,129 +489,129 @@ class THAnimationManager
...
+    size_t animation_index; ///< Animation number.
+    size_t frame_index;     ///< Frame number.
     union {
         struct {
-            //! Amount to change m_iX per tick
+            //! Amount to change x per tick
             int x;
-            //! Amount to change m_iY per tick
+            //! Amount to change y per tick
             int y;
```
Alberth289346: This becomes ambiguous, the comment points to a different `y` than the variable `y` here, but that difference disappears with your renaming. Arguably, `y` here is wrong, I'd make it `dy` at least, or `dy_per_tick` or so

CorsixTH/Src/th_gfx.h
```diff
@@ -484,129 +489,129 @@ class THAnimationManager
...
+    void draw(render_target* pCanvas, int iDestX, int iDestY);
+    bool hit_test(int iDestX, int iDestY, int iTestX, int iTestY);
 
-    void setSheet(THSpriteSheet* pSheet) {m_pSpriteSheet = pSheet;}
-    void setSpeed(int iX, int iY) {m_iSpeedX = iX, m_iSpeedY = iY;}
-    void setLifetime(int iLifetime);
-    void appendSprite(size_t iSprite, int iX, int iY);
-    bool isDead() const {return m_iLifetime == 0;}
+    void set_sheet(sprite_sheet* pSheet) {sheet = pSheet;}
+    void set_speed(int iX, int iY) {speed_x = iX, speed_y = iY;}
```
Alberth289346: a comma?! :p

CorsixTH/Src/th_gfx.cpp
TheCycoONE: Formatting is my next PR in this series

CorsixTH/Src/th_gfx.cpp
```diff
@@ -297,7 +297,7 @@ bool THAnimationManager::loadFromTHFile(
     @param [inout] iLeft Left value to check and update.
     @param iRight Second value to check.
  */
-inline static void _setmin(int& iLeft, int iRight)
+static void set_left_to_min(int& iLeft, int iRight)
```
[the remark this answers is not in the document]
TheCycoONE: We need to start a list

CorsixTH/Src/th_gfx.cpp
TheCycoONE: Ok, it'll get clobbered in the format PR by I will fix it now too.

CorsixTH/Src/th_gfx.h
TheCycoONE: Esoteric C operator for no reason... another on for the list.

CorsixTH/Src/th_gfx.h
Alberth289346: In freerct, I used 256 sub-tile positions, multiplied the tile index by 256, and merged both numbers so I had a single position number that expressed the precise position in the game.

CorsixTH/Src/th.h
```diff
@@ -26,59 +26,63 @@ SOFTWARE.
...
 
 //! Generic linked list class (for inheriting from)
-struct THLinkList
+class link_list
 {
-    THLinkList() ;
-    ~THLinkList();
+public:
+    link_list() ;
+    ~link_list();
```
Alberth289346: Given its purpose is inheriting (as stated in the comment), destructor should be virtual

CorsixTH/Src/persist_lua.cpp
```diff
@@ -406,40 +401,40 @@ class LuaPersistBasicWriter : public LuaPersistWriter
...
                 if(iTable != 3)
                     goto table_resume;
                 break; }
 
             case LUA_TFUNCTION:
                 if(lua_iscfunction(L, 2))
                 {
-                    setErrorObject(2);
-                    setError("Cannot persist C functions");
+                    set_error_object(2);
```
Alberth289346: Is 2 a magic number that should be eliminated?

CorsixTH/Src/persist_lua.cpp
```diff
@@ -450,37 +445,37 @@ class LuaPersistBasicWriter : public LuaPersistWriter
...
                     lua_getfenv(L, 2);
-                    writeStackObject(-1);
+                    write_stack_object(-1);
                     lua_pop(L, 1);
                 }
                 break;
 
             case LUA_TUSERDATA:
-                if(!_checkThatUserdataCanBeDepersisted(2))
+                if(!check_that_userdata_can_be_depersisted(2))
```
Alberth289346: bit long, but probably fine

CorsixTH/Src/persist_lua.cpp
```diff
@@ -555,7 +548,7 @@ class LuaPersistBasicWriter : public LuaPersistWriter
             {
                 if(lua_objlen(L, iIndex) != 0)
                 {
-                    setError("Can only persist non-empty userdata if "
+                    set_error("Can only persist non-empty userdata if "
                         "they have a __depersist_size metafield");
```
Alberth289346: Should such multi-line strings be merged onto a single line at some point?

CorsixTH/Src/th_map.h
```diff
@@ -154,32 +154,32 @@ struct th_map_node_flags
...
+enum class temperature_theme {
     red,         //!< Default warmth colouring (red gradients)
-    multiColour, //!< Different colours (blue, green, red)
-    yellowRed    //!< Gradients of yellow, orange, and red
+    multi_colour, //!< Different colours (blue, green, red)
+    yellow_red    //!< Gradients of yellow, orange, and red
 };
 
-struct THMapNode : public link_list
+struct map_node : public link_list
```
Alberth289346: `link_list` is a class, make this one too?

CorsixTH/Src/th_map.h
```diff
@@ -381,60 +381,61 @@ class THMap
...
     //! \param pNode A pointer to the current node being tested.
-    //! \param prevTemp The array index into THMapNode::aiTemperature that currently stores the temperature of the node (prior to this calculation).
+    //! \param prevTemp The array index into map_node::temperature that currently stores the temperature of the node (prior to this calculation).
     //! \return The weight of the connection, 0 if there is no neighbour, 1 through walls, and 4 through air.
-    uint32_t thermalNeighbour(uint32_t &iNeighbourSum, bool canTravel, std::ptrdiff_t relative_idx, THMapNode* pNode, int prevTemp) const;
+    uint32_t thermal_neighbour(uint32_t &iNeighbourSum, bool canTravel, std::ptrdiff_t relative_idx, map_node* pNode, int prevTemp) const;
 
     //! Create the adjacency matrix if it doesn't already exist
-    void _makeAdjacencyMatrix();
+    void mae_adjacency_matrix();
```
Alberth289346: typo "mae"

CorsixTH/Src/th_movie.h
```diff
@@ -321,70 +321,71 @@ class THMovie
...
     //!
     //! \param pFrame An empty frame which gets populated by the data in the
     //! video packet queue.
     //! \returns 1 if the frame was received, 0 if it was not, and < 0 on error
-    int getVideoFrame(AVFrame *pFrame);
+    int get_video_frame(AVFrame *pFrame);
 #endif
 
-    SDL_Renderer *m_pRenderer; ///< The renderer to draw to
+    SDL_Renderer *renderer; ///< The renderer to draw to
```
Alberth289346: th_gfx_font has code like `render_target* pCanvas`, while here the `*` is near the variable.
I prefer the latter, as it matches the true semantics of the pointer declaration, but at least it's inconsistent.

CorsixTH/Src/iso_fs.cpp
```diff
@@ -370,78 +370,78 @@ IsoFilesystem::file_handle_t IsoFilesystem::findFile(const char* sPath) const
...
     else
-        return m_pFiles[iFile - 1].iSize;
+        return files[iFile - 1].size;
 }
 
-bool IsoFilesystem::getFileData(file_handle_t iFile, uint8_t *pBuffer)
+bool iso_filesystem::get_file_data(file_handle iFile, uint8_t *pBuffer)
 {
-    if(iFile <= 0 || static_cast<size_t>(iFile) > m_iNumFiles)
+    if(iFile <= 0 || static_cast<size_t>(iFile) > file_count)
```
Alberth289346: As a general policy, would it make sense to have this 0 vs 1 stuff in the Lua <-> C++ bridge only?

CorsixTH/Src/th_gfx_font.h
```diff
@@ -206,80 +206,80 @@ class THFreeTypeFont final : public THFont
...
-        int iNumRows;
+        int row_count;
 
         //! Alignment of the message in the box
-        eTHAlign eAlign;
+        align alignment;
 
-        //! True when the pData reflects the sMessage given the size constraints
-        bool bIsValid;
+        //! True when the data reflects the message given the size constraints
```
Alberth289346: Such a shame that we loose the link to the variables by the rename operation, doxygen would need a `#` prefix (or a full `classname::` prefix) to preserve the link.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -1512,49 +1512,49 @@ void line::depersist(lua_persist_reader *pReader)
...
     {
-        SDL_DestroyTexture(pCacheEntry->pTexture);
-        pCacheEntry->pTexture = nullptr;
+        SDL_DestroyTexture(pCacheEntry->texture);
+        pCacheEntry->texture = nullptr;
     }
 }
 
-void THFreeTypeFont::_makeTexture(render_target *pEventualCanvas, cached_text_t* pCacheEntry) const
+
```
Alberth289346: additional empty line

CorsixTH/Src/th_lua.h
```diff
@@ -205,19 +205,19 @@ template <> struct luaT_classinfo<raw_bitmap> {
     static inline const char* name() {return "RawBitmap";}
 };
 
-class THFont;
-template <> struct luaT_classinfo<THFont> {
+class font;
+template <> struct luaT_classinfo<font> {
```
Alberth289346: could merge the class declaration in the template, I think

CorsixTH/Src/th_gfx_font.h
```diff
@@ -28,38 +28,38 @@ SOFTWARE.
 #include FT_FREETYPE_H
 #endif
 
-enum class eTHAlign {
+enum class align {
```
Alberth289346: `text_align` ?

CorsixTH/Src/th_map_overlays.h
```diff
@@ -64,13 +64,13 @@ class map_typical_overlay : public map_overlay
     virtual ~map_typical_overlay();
 
     void set_sprites(sprite_sheet* pSheet, bool bTakeOwnership);
-    void set_font(THFont* font, bool take_ownership);
+    void set_font(::font* font, bool take_ownership);
```
Alberth289346: the need for `::` prefix signals to me that `font` may be too generic as class name.

CorsixTH/Src/run_length_encoder.cpp
```diff
@@ -25,79 +25,79 @@ SOFTWARE.
 #include <new>
 #include <algorithm>
 
-IntegerRunLengthEncoder::IntegerRunLengthEncoder()
+integer_run_length_encoder::integer_run_length_encoder()
```
Alberth289346: `integer` has no meaning does it?
Also `run_length` could be considered an implementation detail, but that somewhat depends on what other encoders we have.

CorsixTH/Src/run_length_encoder.cpp
```diff
@@ -25,79 +25,79 @@ SOFTWARE.
...
+bool integer_run_length_encoder::initialise(size_t iRecordSize)
 {
-    _clean();
-    m_iRecordSize = iRecordSize;
+    clean();
+    record_size = iRecordSize;
 
     // Buffer must hold at least 7 + 2 * 8 records, as the maximum object size
     // is 8 records, 2 of which are needed to detect a repeat, and 7 for the
     // offset at which the objects are found.
```
Alberth289346: so this sets requirements for `iRecordSize`. Imhi a better solution is to silently fix any given record size to the requirements.

CorsixTH/Src/run_length_encoder.cpp
```diff
@@ -129,199 +129,199 @@ void IntegerRunLengthEncoder::_flush(bool bAll)
...
+        != buffer[(iObjIdx2 + i) % buffer_capacity])
         {
             return false;
         }
     }
     return true;
 }
 
-bool IntegerRunLengthEncoder::_output(size_t iObjSize, size_t iObjCount)
+bool integer_run_length_encoder::move_object_to_output(size_t iObjSize, size_t iObjCount)
```
Alberth289346: encoder shouldn't store the output imho

CorsixTH/Src/th_lua.h
```diff
@@ -227,8 +227,8 @@ template <> struct luaT_classinfo<layers> {
     static inline const char* name() {return "Layers";}
 };
 
-class THPathfinder;
-template <> struct luaT_classinfo<THPathfinder> {
+class pathfinder;
```
Alberth289346: `path_finder` is perhaps better in line with the other code??

CorsixTH/Src/th_pathfind.h
```diff
@@ -121,129 +121,129 @@ class BasePathing
...
 
-    inline bool findIdleTile(const level_map* pMap, int iStartX, int iStartY, int iN)
+    inline bool find_idle_tile(const level_map *pMap, int iStartX, int iStartY, int iN)
     {
-        return m_oIdleTileFinder.findIdleTile(pMap, iStartX, iStartY, iN);
+        return idle_tile_finder.find_idle_tile(pMap, iStartX, iStartY, iN);
     }
 
-    inline bool findPathToHospital(const level_map* pMap, int iStartX, int iStartY)
+    inline bool find_path_to_hospital(const level_map *pMap, int iStartX, int iStartY)
```
Alberth289346: Always fun, these `* ` to ` *` conversions :p

CorsixTH/Src/th_sound.cpp
```diff
@@ -26,49 +26,49 @@ SOFTWARE.
...
+    if(iDataLength < sizeof(uint32_t) + sizeof(th_header))
         return false;
 
     uint32_t iHeaderPosition = reinterpret_cast<const uint32_t*>(pData + iDataLength)[-1];
-    if(static_cast<size_t>(iHeaderPosition) >= iDataLength - sizeof(th_header_t))
+    if(static_cast<size_t>(iHeaderPosition) >= iDataLength - sizeof(th_header))
         return false;
 
-    m_oHeader = *reinterpret_cast<const th_header_t*>(pData + iHeaderPosition);
+    header = *reinterpret_cast<const th_header*>(pData + iHeaderPosition);
```
Alberth289346: `th_header` looks wrong, I guess this is the weak point of not using camel case for types. I'd go for renaming the class to something more specific, eg `sound_header` ?

CorsixTH/Src/th_sound.cpp
```diff
@@ -26,49 +26,49 @@ SOFTWARE.
...
 
-    m_pFiles = reinterpret_cast<th_fileinfo_t*>(m_pData + m_oHeader.table_position);
-    m_iFileCount = m_oHeader.table_length / sizeof(th_fileinfo_t);
+    sound_files = reinterpret_cast<th_fileinfo*>(data + header.table_position);
+    sound_file_count = header.table_length / sizeof(th_fileinfo);
     return true;
 }
 
-size_t THSoundArchive::getSoundCount() const
+size_t th_sound_archive::get_sound_count() const
```
Alberth289346: `get_number_of_sounds` ?

CorsixTH/Src/th_sound.cpp
```diff
@@ -150,179 +150,179 @@ size_t THSoundArchive::getSoundDuration(size_t iIndex)
...
-SDL_RWops* THSoundArchive::loadSound(size_t iIndex)
+SDL_RWops* th_sound_archive::load_sound(size_t iIndex)
 {
-    if(iIndex >= m_iFileCount)
+    if(iIndex >= sound_file_count)
         return nullptr;
 
-    th_fileinfo_t *pFile = m_pFiles + iIndex;
-    return SDL_RWFromConstMem(m_pData + pFile->position, pFile->length);
+    th_fileinfo *pFile = sound_files + iIndex;
```
Alberth289346: th prefix looks wrong/inconsistent

CorsixTH/Src/th_sound.cpp
```diff
@@ -150,179 +150,179 @@ size_t THSoundArchive::getSoundDuration(size_t iIndex)
...
 {
-    return ms_pSingleton;
+    return singleton;
 }
 
-void THSoundEffects::setSoundArchive(THSoundArchive *pArchive)
+void sound_player::populate_from(th_sound_archive *pArchive)
 {
-    for(size_t i = 0; i < m_iSoundCount; ++i)
+    for(size_t i = 0; i < sound_count; ++i)
```
Alberth289346: looks hopelessly silly, the archive should know the number of sounds, or it should be provided as parameter

CorsixTH/Src/th_sound.cpp
```diff
@@ -150,179 +150,179 @@ size_t THSoundArchive::getSoundDuration(size_t iIndex)
...
-    ms_pSingleton = this;
-    m_iCameraX = 0;
-    m_iCameraY = 0;
-    m_fCameraRadius = 1.0;
-    m_fMasterVolume = 1.0;
-    m_fSoundEffectsVolume = 0.5;
-    m_iPostionlessVolume = MIX_MAX_VOLUME;
-    m_bSoundEffectsOn = true;
+    sounds = nullptr;
+    sound_count = 0;
```
Alberth289346: why does a sound player need to know how many sounds exist

CorsixTH/Src/th_sound.cpp
```diff
@@ -150,179 +150,179 @@ size_t THSoundArchive::getSoundDuration(size_t iIndex)
...
 
-int THSoundEffects::reserveChannel()
+int sound_player::reserve_channel()
 {
     // NB: Callers ensure that m_iChannelStatus != 0
     int iChannel = 0;
-    for(; (m_iChannelStatus & (1 << iChannel)) == 0; ++iChannel) {}
-    m_iChannelStatus &=~ (1 << iChannel);
+    for(; (available_channels_bitmap & (1 << iChannel)) == 0; ++iChannel) {}
+    available_channels_bitmap &=~ (1 << iChannel);
```
Alberth289346: always nice if there is no channel available

CorsixTH/Src/xmi2mid.cpp
```diff
@@ -31,75 +31,75 @@ SOFTWARE.
 /*!
     Utility class for reading or writing to memory as if it were a file.
 */
-class MemoryBuffer
+class memory_buffer
```
Alberth289346: looks like something that should be shared and re-used at every point where we have a buffer

CorsixTH/Src/th_lua_ui.cpp
```diff
@@ -25,9 +25,9 @@ SOFTWARE.
 #include "th_map.h"
 #include <algorithm>
 
-struct THWindowBase_t {};
+class abstract_window {};
```
Alberth289346: not sure about `abstract_window`, sounds more like something that deals with rectangular areas of some size with generic operations on it, rather than a base class that provides common functionality for a gui, but ymmv.

CorsixTH/Src/th_lua_strings.cpp
```diff
@@ -51,7 +51,7 @@ SOFTWARE.
     environment.
 */
 
-struct THStringProxy_t {};
+class string_proxy {};
```
Alberth289346: It doesn't look like a proxy, much more like a string computation storage, which I totally do not get.

CorsixTH/Src/th_lua.h
```diff
@@ -277,8 +277,8 @@ template <> struct luaT_classinfo<string_proxy> {
     static inline const char* name() {return "StringProxy";}
 };
 
-struct THLfsExt;
-template <> struct luaT_classinfo <THLfsExt> {
+class lfs_ext;
```
Alberth289346: looks very short compared to the other names

Alberth289346:
I went through the changes in varying states of awakiness, hopefully I covered most of the problems. Lots of comments are not even about the names, I think, at some point I just started writing things that looked wrong, so feel free to move them to the list, or even totally ignore them.
The same somewhat applies to renaming suggestions, in the end, as long as the name is unique, it's fine. There are likely more serious things wrong that need fixing than a name.

CorsixTH/Src/th_movie.h
TheCycoONE: I will be switching all of these to C++ style in the format PR.

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#nl18-use-c-style-declarator-layout

http://www.stroustrup.com/bs_faq2.html#whitespace

CorsixTH/Src/th_lua.h
TheCycoONE: Extensions to luafilesystem. Ok I will change this to be clearer

CorsixTH/Src/th_gfx_font.h
```diff
@@ -28,38 +28,38 @@ SOFTWARE.
...
     left = 0,
     center = 1,
     right = 2,
 };
 
 /** Structure for the bounds of a text string that is rendered to the screen. */
-struct THFontDrawArea
+struct text_draw_dimensions
```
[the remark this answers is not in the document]
TheCycoONE: I struggled with this name a bit (and for awhile I did have text_area_*. But a text_area is usually a container for text and I think it should be emphasized that this is the actual resulting size of the text as rendered.  Open to other suggestions.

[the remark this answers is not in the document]
Alberth289346: `text_layout`? At least, it looks like layout information. The fuzzy part is that there is no text in the structure, which makes `row_count` and `width` quite weird.

CorsixTH/Src/th_lua.h
Alberth289346: There doesn't seem to be anything "Lua specific" in lfs (eg special handling of lua files or so), just regular fs access for lua. Drop the "lua" part?

CorsixTH/Src/run_length_encoder.cpp
TheCycoONE: Perhaps should be uint32_run_length_encoder/decoder, because it encodes / decodes uint32s. Being a RLE affects the generated result and means the corresponding decoder has to be compatible (i.e. not an implementation detail). It's a fairly low level concept so I think the name makes sense (except 'integer').

CorsixTH/Src/th_lua.h
TheCycoONE: Pathfinder is one word in English.

CorsixTH/Src/th_sound.cpp
TheCycoONE: Maybe even `sound_dat_file_header`

CorsixTH/Src/th_gfx_font.h
```diff
@@ -28,38 +28,38 @@ SOFTWARE.
...
 {
     //! Number of rows the rendered text spans
```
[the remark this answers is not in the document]
TheCycoONE: Ok, I will make a point to address this in the format PR

CorsixTH/Src/th_gfx_font.h
TheCycoONE: Ok, I didn't know it generated the link before.

CorsixTH/Src/persist_lua.cpp
TheCycoONE: 2 is a Lua stack index. There are a lot of magic number lua stack indexes in the code - it makes it very hard to follow.

CorsixTH/Src/th_lua.h
TheCycoONE: luafilesystem is the name of a 3rd party module. This is an extension to make it play nicer with Windows.

CorsixTH/Src/th_gfx.h
TheCycoONE: I'll take a second look at this whole struct. Seems like it confuses what it is with how it's used.

CorsixTH/Src/persist_lua.cpp
TheCycoONE: Why?

CorsixTH/Src/iso_fs.cpp
TheCycoONE: Yes

CorsixTH/Src/th_gfx_font.h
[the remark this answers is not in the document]
TheCycoONE: I like `text_layout`, changed locally. I'll push my changes when I've addressed everything.

CorsixTH/Src/th_sound.cpp
TheCycoONE: renamed completely, file here is misleading at best.

CorsixTH/Src/th_sound.cpp
TheCycoONE: Because this sound player prepares an SDL sound for every piece of sound data so that they are ready to play, and because they don't use an stl container for sounds.

CorsixTH/Src/th_sound.h
```diff
@@ -28,35 +28,35 @@ SOFTWARE.
 #endif
 
 //! Utility class for accessing Theme Hospital's SOUND-0.DAT
-class THSoundArchive
+class th_sound_archive
```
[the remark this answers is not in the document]
TheCycoONE: fixed

CorsixTH/Src/th_lua_strings.cpp
TheCycoONE: No idea, I didn't follow this one enough to dispute the existing name.

TheCycoONE:
Still working on your suggestions. One other significant change I have made since then is renaming map_node to map_tile and everything related (because they are tiles, and node is needlessly general).

CorsixTH/Src/th_lua_ui.cpp
TheCycoONE: either way it's short for "abstract base class", I like abstract better these days, so I'm keeping it :p

CorsixTH/Src/th_lua.h
TheCycoONE: Keeping lfs, that's luafilesystem's internal name. e.g. `lfs.dir(...)`, `lfs.attributes(...)`

TheCycoONE:
Addressed all of the naming issues mentioned by @Alberth289346 as well as some of the formatting issues (I know I missed some).

Code issues are out of scope for this PR.

TheCycoONE:
Ok, let's get this merged since it basically blocks C++ work.

AnimView won't be updated to this naming scheme as it clashes with wxWidgets

PR #1383, opened 2018-04-12
[RDY] Remove jit_opt

TheCycoONE:
Removes jit_opt.h, a byte array of the bytecode of jit.opt from LuaJIT 1.x.

Since LuaJIT 2.x it is no longer necessary to call functions in this library to start the jit; it happens automatically.

Our LuaJIT usage in general is very low, and 1.x is quite old. I think we can drop it.


TheCycoONE:
I have tested this briefly on LuaJIT 2.1.0-beta3 and everything runs fine.

Issue #1385, opened 2018-04-13
Jukebox preventing save


mugmuggy:
#### Describe the issue
Same issue as #543, cannot save games, autosave or manual if jukebox is open.

#### Steps to Reproduce
1. Open Jukebox, wait for next month to tick over

2. Console fills with many 'Unhandled argument to th_getfenv'
followed by
```
nil     <no name>
Error while autosaving game: Lua\persistance.lua:247: Can only persist non-empty
 userdata if they have a __depersist_size metafield
```
If you have another dialog open an Information window is displayed with the error above also.

#### Expected Behavior
Saves should be created.

#### System Information
**CorsixTH Version:** *trunk*

#### Additional Info
I have a PR in the works if anyone else can confirm they indeed have the issue.

TheCycoONE:
Haven't looked yet but I'm curious what your PR does, and if this only affects certain lua versions. th_getfenv is completely different in Lua 5.1/LuaJIT than Lua 5.2+

mugmuggy:
Admittedly it might not be the most elegant. I switched to the setToggleState calls, as if you engage looping, then click a toggle it becomes `enabled`, then undo the loop it was toggling the state again and combined with the possibilty of going back to a save the ui became inconsistent, but the update in draw probably would have corrected that anyway.

I must admit these persisted funtions and the use of them confuses me a little but I guess one of the additional parameters was preventing it working.

TheCycoONE:
If you're curious, what you did was remove 'info' which is userdata (i.e. a c++ class/struct) from the upvalues (variables from outside a function, stored with and used in a function) from the persistable jukebox_toggle_track function. Now the only upvalue that function stores is the index i.

Evidently that userdata was not properly configured for storage. There are likely still problems with persisting the jukebox; e.g. if the music list changes between loads. 

I'm ok with merging this PR as is, because I think it's a bit cleaner, but preferably UIJukebox would be excluded from saves, and the callback wouldn't need to be persistable. Since this hasn't changed in many releases, it's probably safe to not even add a contingency for old saves.

MarkL1961:
As you cannot save with the jukebox open I tend to agree there won't or should not be any issues with saves. (I don't think you ever could)
I am not sure that anyone would even notice that the music was changed from a save to loading it later and as it would only happen if you had the jukebox open at the time it is even less likely to happen

mugmuggy:
[...]

PR #1386, opened 2018-04-13
[RFC] UIJukebox prevents save due to persistance error if open at time of save

mugmuggy:
An attempt to address #1385

### DIFF
--- a/CorsixTH/Lua/dialogs/jukebox.lua
+++ b/CorsixTH/Lua/dialogs/jukebox.lua
@@ -66,10 +66,10 @@ function UIJukebox:UIJukebox(app)
     end
     self.track_buttons[i] = self:addPanel(404, self.width - 61, y):makeToggleButton(19, 4, 24, 24, 405):setSound("selectx.wav")
     if not info.enabled then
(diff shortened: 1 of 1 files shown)


Alberth289346:
A layer woven into the entire program to fix the broken persistence?

TheCycoONE:
Not sure what you mean here @Alberth289346 

The issue is that info is a C++ object not a lua object, and doesn't have the prerequisite methods defined on it to be able to be persisted.

This PR doesn't address that directly, instead removing info from the persist list. In truth we don't want UIJukebox persisted at all.

To summarize: this PR is not the fix we want for this issue. The question is whether the new code is cleaner and should be merged, or is messier or just as messy and should be rejected independent of being a workaround for the bug at hand. I haven't decided yet.

TheCycoONE:
This PR is cleaner than without it in my opinion and should be merged.

TheCycoONE:
@mugmuggy are you ok with this being merged (RDY)? I still think it would be best not to persist the dialog, but I like this change too.

mugmuggy:
Yep

PR #1391, opened 2018-04-28
[RDY] Use the same rnc code for CorsixTH and AnimView

TheCycoONE:
CorsixTH/Lua specific code and AnimView specific code was extracted
into the relevant projects. rnc.h was modified into a public contract
for using rnc as a C++ library. Helper functions were added.

New common rnc code is in new $/common directory.


TheCycoONE:
Missed a file it seems.

Issue #1290, opened 2017-11-27
doctor's/nurses after resting abandon their consults and wont take new consults.


overlordchaos:
#### Describe the issue
every time a doctor (surgeon/doctor/psychiatrist/investigation/specialist/nurse) after going to rest, 90% of the times wont go back to their assigned station will just roam around. and very rarely when needed will go to a compatible station (for example if the nurse goes out the enfirmary to rest, then she does not go back, but might run into the pharmacy, those are rare cases like 10% or less) this also happened on the original game. it does not make it less of an issue that should be resolved.
#### Steps to Reproduce
1. run any hospital with several doctors/nurses in

2.let them to work then go to rest. once that is done they will start to roam around instead of go back to their assignated stations.

3. (just obserb)

#### Expected Behavior
they should go back to their stations, and if they are occupied by a new staff member, in the second they are needed in another place, auto assign that station to themselves.
#### System Information
**CorsixTH Version:** *e.g. 0.50 or 7d886f35ca*
6.1 (beta)
**Operating System:** *e.g. Windows 10*
windows 10 x64
**Theme Hospital Version:** *CD, GOG.com, Origin, or Demo*
gog.com
### Additional Info
it could be useful to check the psychiatrist code, considering the psychiatrist (even on the original game) in the second a regular doctor is needed, they usually run to take the place of the regular doctor. this code could be modded and applied to all the staff so this issue does not accur anymore.

TheCycoONE:
Classifying as a low priority enhancement, since micromanaging doctors after the leave for rest was part of the original game; but you're right it's kinda tedious and doesn't add much fun. The alternative was the stick to your room policy, in which case you'd have to micromanage how tired they were getting.

overlordchaos:
yes aldo if you force them to stick to your room, you will have to manually replace the doctor/nurse and when they are rested switch them back. it is quite tedious if you have 30+ staff members and you need to be tracking down each one. aldo i am sure its an original game bug to have with or without the room policy to doctors/nurses roaming around while the recepcionist asks for a replacement in a room and they ignore it while patients wait untill death or leaving angry. its a mix of game mechanic/glitch if you ask me ^_^

MarkL1961:
That is somewhat realistic, there are people where I work who just wander round all day whilst others are run off their feet :)

lewri:
Actually I think might be normal behaviour of the game.

[...]

PR #1405, opened 2018-05-21
Update calls_dispatcher.lua

LeDocteurr:
Made a change in calls_dispatcher.lua in the CallsDispatcher.verifyStaffForRoom function so that wandering staff will no longer steal already occupied rooms, preventing for example surgeons to end up in GP's office, as reported in issue #1290.

The change also requires a new function I added in #1406

### DIFF
--- a/CorsixTH/Lua/calls_dispatcher.lua
+++ b/CorsixTH/Lua/calls_dispatcher.lua
@@ -501,13 +501,15 @@ function CallsDispatcher.verifyStaffForRoom(room, attribute, staff)
   end
 
   local current_room = staff:getRoom()
+  local last_room = staff.last_room
   if not staff.hospital.policies["staff_allowed_to_move"] then
-    if current_room and current_room ~= room then
-      -- Staff is in another room, not usable.
+    if current_room and current_room ~= room
+    or last_room and last_room ~= room then
+      -- Staff is in another room or heading to their room, not usable.
       return false
     end
     if room:isFullyStaffed() then
-      -- The room is already fully staffed, don't let the staff steal the room.
+      -- The room is already fully staffed, not allowed to steal room.
       return false
     end
   end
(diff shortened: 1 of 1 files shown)


LeDocteurr:
Yes, if the staff are not allowed to leave their rooms, they shouldn't just jump and help if they are already allocated to a post. The old code rendered the "staff leave rooms" option useless by:

- allowing free staff to help already occupied stations when their owners were resting, taking them as their own
- allowing staff that owned a room to answer calls right after leaving the staffroom

So for now I fixed the "staff leave rooms" option, but further code can be added to enhance behaviour, for example a doctor could come and kick temporary staff out in order to reoccupy their posts (which will require a new variable to bind them to their posts since last_room only holds the last visited room), I don't think it's hard to do, but a more clear design is needed otherwise it could get messy.

Alberth289346:
You know the original TH aimed at a lot of micro-managing?
Automating shuffling of staff seems a move away from that.

LeDocteurr:
@Alberth289346 : Having the staff shuffled to allow micromanagement when "staff leave room" is enabled is one thing, but if it's disabled then it's a bug. What's the point of that option if the staff can leave their rooms at the first call? Even if the original game somehow allowed that, then it's still a bug and the purpose of CorsixTH is to fix those bugs and enhance the original. Looking back at issue #1290 there are lots of people bothered by that behaviour, me included :) Since now we can have custom maps, think about very large maps with 3+ OT, where all you do is bringing back surgeons to their OT... not fun.

Alberth289346:
Code-style for the multi-line 'if' statements is wrong, the 'and' or 'or' should not be at the beginninhg of the next line, and be indented twice. You may want to check the guldelines:
https://github.com/CorsixTH/CorsixTH/wiki/Coding-Conventions#lua-code

LeDocteurr:
I've included this changes in the "Private Clinic" patch, available on the CorsixTH website, if you are going to use my code, please include my pseudonym "LeDocteur" in the copyright list of each modified file.

PR #1406, opened 2018-05-21
Update room.lua

LeDocteurr:
Added a function Room:isFullyStaffed to test if a room is fully staffed even if the staff is not currently present in the room. The function is needed for pull #1405 to prevent wandering staff to steal rooms that do not belong to them when the staff are not allowed to leave their offices.


LeDocteurr:
I've included this changes in the "Private Clinic" patch, available on the CorsixTH website, if you are going to use my code, please include my pseudonym "LeDocteur" in the copyright list of each modified file.

PR #1418, opened 2018-06-04
[RDY] Fix one iteration loop warnings

TheCycoONE:
Use next instead of pairs to test if a table is empty.

### DIFF
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -258,10 +258,7 @@ function Room:testStaffCriteria(criteria, extra_humanoid)
     end
     return false
   else
-    for _, _ in pairs(missing) do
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/room.lua
```diff
@@ -258,10 +258,10 @@ function Room:testStaffCriteria(criteria, extra_humanoid)
     end
     return false
   else
-    for _, _ in pairs(missing) do
-      return false
+    if next(missing) == nil then
+      return true
     end
-    return true
+    return false
```
Alberth289346: `return next(missing) == nil` just like the previous change?

TheCycoONE: Sure

PR #1430, opened 2018-07-26
[RDY] Shape up scripts/*.py and SpriteEncoder/decode.py

antonv6:
I know this project is primarily Lua, but there's some Python code too, so I ran flake8 on it. flake8 is a Python linter/style checker, and can be a bit nitpicky at times, but I think these changes is an improvement. Feel free to only pick specific commits if you wish.

### DIFF
--- a/scripts/check_language_files_not_BOM.py
+++ b/scripts/check_language_files_not_BOM.py
@@ -27,14 +27,17 @@ top = os.getcwd()
 if len(sys.argv) == 2:
     top = os.path.join(top, sys.argv[1])
 
+count = 0
 offending_files = []
 for root, dirs, files in os.walk(top):
     for f in files:
         if f.endswith('.lua'):
+            count += 1
             path = os.path.join(root, f)
             if is_BOM_encoded_file(path):
                 offending_files.append(f)
 
+print('Checked {} files'.format(count))
 if offending_files:
     print('Found files with UTF-8 with BOM encoding:')
     for f in offending_files:
--- a/scripts/check_language_files_not_BOM.py
+++ b/scripts/check_language_files_not_BOM.py
@@ -14,27 +14,27 @@ import sys
 
 
 def is_BOM_encoded_file(path):
-  """ Returns whether |path| is a file that is encoded in UTF-8 with BOM. """
-  with open(path, 'rb') as f:
-    raw = f.read(4)
-    return raw.startswith(codecs.BOM_UTF8)
+    """ Returns whether |path| is a file that is encoded in UTF-8 with BOM. """
+    with open(path, 'rb') as f:
+        raw = f.read(4)
+        return raw.startswith(codecs.BOM_UTF8)
 
 
 if len(sys.argv) > 2:
-  sys.exit('Usage: {} [root]'.format(sys.argv[0]))
+    sys.exit('Usage: {} [root]'.format(sys.argv[0]))
 
 top = os.getcwd()
 if len(sys.argv) == 2:
-  top = os.path.join(top, sys.argv[1])
+    top = os.path.join(top, sys.argv[1])
 
 offending_files = []
 for root, dirs, files in os.walk(top):
-  for f in files:
-    if f.endswith('.lua'):
-         path = os.path.join(root, f)
-         if is_BOM_encoded_file(path):
-           offending_files.append(f)
+    for f in files:
+        if f.endswith('.lua'):
+            path = os.path.join(root, f)
+            if is_BOM_encoded_file(path):
+                offending_files.append(f)
 
 if offending_files:
-  sys.exit('Found files with UTF-8 with BOM encoding: ' + str(offending_files))
+    sys.exit('Found files with UTF-8 with BOM encoding: ' + str(offending_files))
 sys.exit(0)
--- a/scripts/check_language_files_not_BOM.py
+++ b/scripts/check_language_files_not_BOM.py
@@ -20,7 +20,7 @@ def is_BOM_encoded_file(path):
     return raw.startswith(codecs.BOM_UTF8)
 
 
-if (len(sys.argv) > 2):
+if len(sys.argv) > 2:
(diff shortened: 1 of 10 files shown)


TheCycoONE:
LGTM

scripts/check_lua_classes.py
```diff
@@ -22,7 +23,8 @@
           print("******* CHECK CLASS DECLARATIONS *******")
           problem_found = True
           print("Invalid/Improper Class Declarations Found:")
-        print("*" + print_root_regex.search(root).group(0) + "\\" + script + ":" + found_class)
+        path = print_root_regex.search(root).group(0)
+        print("*" + path + "\\" + script + ":" + found_class)
```
Alberth289346: Using str.format looks better:
``` python
print("*{}\\{}:{}".fprmat(path, script, found_class))
```

scripts/check_trailing_whitespaces.py
```diff
@@ -27,6 +26,7 @@ def has_trailing_whitespaces(path):
   handle.close()
   return False
 
+
 if (len(sys.argv) > 2):
```
Alberth289346: outermost parentheses are not needed

scripts/check_lua_classes.py
wolfy1339: Depending on what version of Python you want to target, there's f-string templates (3.6+):
```py
print(f"*{path}\\{script}:{found_class}")
```

Alberth289346: It would be a great addition, but 3.6 sounds a bit too new, my fedora which tends to deploy the newer versions doesn't even has 3.7 yet. There will be many users that use something older.

TheCycoONE: The real condition here is that it has to run in travis_ci; which is currently running Ubuntu 14.04. I don't mind bringing in a back port of a newer python version; but if we don't have to it's easier.

TheCycoONE: I believe that means 3.4.0 should be the target version.

Alberth289346:
Thanks for the update, looks good, hopefully travis doesn't get stuck again

antonv6:
The scripts work for me, they actually catch some files:
```
$ python ./scripts/check_language_files_not_BOM.py ./CorsixTH/Lua/languages
Found files with UTF-8 with BOM encoding:
norwegian.lua

$ python scripts/check_trailing_whitespaces.py 
Found files with trailing whitespaces:
.../CorsixTH/CorsixTH/Src/th_map.cpp
.../CorsixTH/CorsixTH/Src/th_map.h
.../CorsixTH/CorsixTH/Src/lua_rnc.cpp
.../CorsixTH/CorsixTH/Src/th_gfx_sdl.h
```

Which is surprising, considering the last build on travis is clean: https://travis-ci.org/CorsixTH/CorsixTH/builds/406569941

(check_lua_classes.py doesn't catch anything right now, but works as expected if I intentionally bork any class definition in CorsixTH/Lua/)

And I don't see my PR in https://travis-ci.org/CorsixTH/CorsixTH/pull_requests. It should be there, shouldn't it? Maybe webhooks didn't fire or something, but travis doesn't even know that it needs to build something, that's my impression.

Alberth289346:
It's there, at the top-right is "more options" dropdoen, with one item "requests". There are 2 "abuse detected" entries, but no other information that I could find

antonv6:
It's... possible that `$TRAVIS_BUILD_DIR` is an absolute path, in which case there are differences in path handling after 95a0430:

- `cwd + '/' + path` gives '/cwd//travis-build-dir'
- `os.path.join(cwd, path)` gives '/travis-build-dir'

Alberth289346:
Good point.

Naming something "root" but then appending it to cwd seems quite counter-intuitive anyway. That should probably be fixed. If you want to fix it further, moving away from python2 seems useful, as is adding a check that we actually do some scanning, to avoid running the program at some directory where no source exists without knowing.

On the other hand if
```
${TRAVIS_BUILD_DIR}/scripts/check_trailing_whitespaces.py $TRAVIS_BUILD_DIR
```
runs, then `$TRAVIS_BUILD_DIR` is obviously in the checked out repo, so at worst it would scan the entire checked out content.

Alberth289346:
@antonv6 I have had contact with travis-ci support, and they say

    The issue is that @antonv6 on GitHub, the author of the pull request
    (https://github.com/CorsixTH/CorsixTH/pull/1430) has not logged into Travis CI before.
    This causes our abuse detection system to false positive since the user can't be found
    in our database. The solution is to ask @antonv6 to login to https://travis-ci.org at least
    once so an account is created for them so we can run builds for them in the background
    when they push code to your (and other) repositories.

    No abuse has been done, just a false positive and requesting they log into our website with
    their GitHub user is the solution. Once they have done that, you can have them push a new
    commit to the pull request and we will start a build.

Not sure I like this requirement (it's new to us as well), but it's your choice.

I am not sure I like this very much, but it's your choice.

antonv6:
Thanks for contacting them, but it seems what they proposed isn't enough. I've logged in using by github account, and the request is still marked as abuse.

Anyway, the scripts now print the number of checked files, so you can see if it's 0 or not. I've decided not to do sys.exit(1) if count == 0, because after all, this is just for debugging your CI setup once (hopefully).

Alberth289346:
ok, shall send more email then :(

scripts/check_language_files_not_BOM.py
```diff
@@ -25,6 +28,8 @@ def is_BOM_encoded_file(path):
 
 top = os.getcwd()
 if len(sys.argv) == 2:
+    if not os.path.isdir(sys.argv[1]):
+          sys.exit('Error: ' + sys.argv[0] + ' is not a directory')
```
TheCycoONE: Error message prints wrong argument.

Alberth289346:
@antonv6  Quite a bumpy ride, but thanks for your time and contributions

PR #1441, opened 2018-08-15
[RDY] Announcement queue

Feanathiel:
I've noticed there was a TODO called "Announcement Queue System" which wasn't implemented yet. I was wondering if you guys could have a peek at my implementation, and provide feedback on where to improve. These are the first few lines of Lua that are typed by my hands, so let me know what you think.

### DIFF
--- a/CorsixTH/Lua/announcer.lua
+++ b/CorsixTH/Lua/announcer.lua
@@ -1,0 +1,210 @@
+--[[ Copyright (c) 2018 Rick "Feanathiel" Megens
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+of the Software, and to permit persons to whom the Software is furnished to do
+so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE. --]]
+
+--! Pr
+local AnnouncementPriority = {
+  Critical = 1,
+  High = 2,
+  Normal = 3,
+  Low = 4
+}
+
+strict_declare_global "AnnouncementPriority"
+_G["AnnouncementPriority"] = AnnouncementPriority
+
+local default_announcement_priority = AnnouncementPriority.Normal
+
+local default_announcement_decay_hours = {
+  [AnnouncementPriority.Critical] = -1, -- never decay
+  [AnnouncementPriority.High] = 365*24,
+  [AnnouncementPriority.Normal] = 31*24,
+  [AnnouncementPriority.Low] = 7*24
+}
+
+--! An announcement queue based on priority
+class "AnnouncementQueue"
+
+---@type AnnouncementQueue
+local AnnouncementQueue = _G["AnnouncementQueue"]
+
+--! Creates an announcement queue: a collection of announcements with queue semantics.
+function AnnouncementQueue:AnnouncementQueue()
+  -- might want to use a single table for this
+  self.priorities = {
+    [AnnouncementPriority.Critical] = {},
+    [AnnouncementPriority.High] = {},
+    [AnnouncementPriority.Normal] = {},
+    [AnnouncementPriority.Low] = {}
+  }
+
+  self.count = 0
+end
+
+--! Adds the announcement entry to the queue.
+--!param priority (int) the priority of the announcement
+--!param entry (AnnouncementEntry) the announcement entry
+function AnnouncementQueue:push(priority, entry)
+  local entries = self.priorities[priority]
+  table.insert(entries, entry)
+  self.count = self.count + 1
+end
+
+--! Dequeues the announcement with the highest priority, nil if the queue is empty.
+function AnnouncementQueue:pop()
+  for _, entries in ipairs(self.priorities) do
+    if entries[1] ~= nil then
+      local entry = table.remove(entries, 1)
+      self.count = self.count - 1
+      return entry
+    end
+  end
+
+  return nil
+end
+
+--! Returns true if the queue is empty, false otherwise.
+function AnnouncementQueue:isEmpty()
+  return self.count == 0
+end
+
+--! An announcement.
+class "AnnouncementEntry"
+
+---@type AnnouncementEntry
+local AnnouncementEntry = _G["AnnouncementEntry"]
+
+--! Creates an announcement
+function AnnouncementEntry:AnnouncementEntry()
+  self.name = nil -- filename to play
+  self.priority = default_announcement_priority
+  self.created_tick = nil -- when it has been created
+  self.decay_ticks = nil -- how long until the announcement isn't relevant anymore
+  self.played_callback = nil -- call me whenever the sound was played, ...
+  self.played_callback_delay = nil -- but not until delay has passed
+end
+
+--! Announces audible messages to the player.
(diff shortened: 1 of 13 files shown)


CorsixTH/Lua/announcer.lua
```diff
@@ -0,0 +1,157 @@
...
+  Normal = 2,
+  Low = 3
+}
+
+strict_declare_global "AnnouncementPriority"
+_G["AnnouncementPriority"] = AnnouncementPriority
+
+local default_announcement_priority = AnnouncementPriority.Normal
+
+local default_announcement_decay_time = {
```
Feanathiel: We can tweak values here. I don't think they make sense right now, but its a start.

CorsixTH/Lua/game_ui.lua
```diff
@@ -79,9 +80,6 @@ function GameUI:GameUI(app, local_hospital, map_editor)
   self.transparent_walls = false
   self.do_world_hit_test = true
 
-  self:setRandomAnnouncementTarget()
-  self.ticks_since_last_announcement = 0
```
Feanathiel: Do I need to clean these up, so they're not in the savegames anymore?

CorsixTH/Lua/game_ui.lua
```diff
@@ -665,16 +665,8 @@ function GameUI:onMouseWheel(x, y)
...
-  -- NB: Every tick is 30ms, so 2000 ticks is 1 minute
-  self.random_announcement_ticks_target = math.random(8000, 12000)
-end
-
-function GameUI:playAnnouncement(name, played_callback, played_callback_delay)
-  self.ticks_since_last_announcement = 0
-  if self.app.world:getLocalPlayerHospital():hasStaffedDesk() then
-    UI.playAnnouncement(self, name, played_callback, played_callback_delay)
-  end
+function GameUI:playAnnouncement(name, priority, played_callback, played_callback_delay)
```
Feanathiel: Adding priority shouldn't break anything. As far as I could find, there are no calls with more than one parameter.

CorsixTH/Lua/game_ui.lua
```diff
@@ -698,15 +690,6 @@ function GameUI:onTick()
     self.multigesturemove.x = 0.0
     self.multigesturemove.y = 0.0
   end
-  if not self.app.world:isCurrentSpeed("Pause") then
```
Feanathiel: Moved to Announcer:onTick

CorsixTH/Lua/game_ui.lua
```diff
@@ -1112,10 +1098,6 @@ function GameUI:afterLoad(old, new)
   if old < 23 then
     self.do_world_hit_test = not self:getWindow(UIPlaceObjects)
   end
-  if old < 28 then
```
Feanathiel: Do I actually need to keep this?

CorsixTH/Lua/game_ui.lua
```diff
@@ -1144,6 +1126,10 @@ function GameUI:afterLoad(old, new)
   if old < 122 then
     self.multigesturemove = {x = 0.0, y = 0.0}
   end
+  if old < 128 then
```
Feanathiel: Check if 128 is the right value...?

CorsixTH/Lua/world.lua
```diff
@@ -578,7 +579,7 @@ function World:tickEarthquake()
     -- start of warning quake
     if self.next_earthquake.warning_timer == earthquake_warning_period then
       self.ui:beginShakeScreen(0.2)
-      self.ui:playAnnouncement(announcements[math.random(1, #announcements)])
+      self.ui:playAnnouncement(announcements[math.random(1, #announcements)], AnnouncementPriority.Critical)
```
Feanathiel: Example call. We need to set priority to playAnnouncement calls. If its not specified, priority 'normal' is used.

CorsixTH/Lua/announcer.lua
```diff
@@ -0,0 +1,157 @@
...
+    local ticks_since_last_announcement = self.ticks_since_last_announcement
+    if ticks_since_last_announcement >= self.random_announcement_ticks_target then
+      self:playAnnouncement("rand*.wav", AnnouncementPriority.Low)
+      self:_setRandomAnnouncementTarget()
+    else
+      self.ticks_since_last_announcement = ticks_since_last_announcement + 1
+    end
+  end
+  
+  self.ticks = self.ticks + 1
```
TheCycoONE: I think it would be better to use the game_date from world than tracking your own hours/in game ticks here.

CorsixTH/Lua/game_ui.lua
TheCycoONE: We'll check this at the time it's merged - depends what else goes in between now and then.

CorsixTH/Lua/game_ui.lua
TheCycoONE: You probably want to drop it since you handle this in your < 128 condition.

CorsixTH/Lua/game_ui.lua
Feanathiel: Alright, thanks.

CorsixTH/Lua/announcer.lua
Feanathiel: I've added a commit handling this issue.

CorsixTH/Lua/game_ui.lua
TheCycoONE: Preferably yes, in the afterLoad.

CorsixTH/Lua/world.lua
TheCycoONE: Looks like an excellent use case.

TheCycoONE:
Looks pretty good to me. If you're able to add busted tests to the Luatest/specs folder that would be great. Our testing is still a bit immature though so I understand if it's too much trouble.

CorsixTH/Lua/announcer.lua
```diff
@@ -0,0 +1,160 @@
...
+    end
+  end
+end
+
+function Announcer:_setRandomAnnouncementTarget()
+  self.random_announcement_ticks_target = math.random(8000, 12000)
+end
+
+function Announcer:_play(entry)
+  self.playing = self.playing + 1
```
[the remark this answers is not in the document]
Feanathiel: This was added to be able to play multiple announcements when multiple desks were available. Now that I think about it, it doesn't make sense, so I've replaced it with a boolean.

Feanathiel:
Busted was giving me a hard time on Windows. I had to set default.lpath to "../Lua/?.lua;?.lua" in the .busted-file, although I didn't commit this (is it my configuration?). In the end I think I've managed to add some testcases, so feedback is appreciated.

CorsixTH/Lua/announcer.lua
```diff
@@ -50,7 +44,7 @@ function AnnouncementQueue:push(priority, entry)
 end
 
 function AnnouncementQueue:pop()
-  for _, entries in pairs(self.priorities) do
+  for _, entries in ipairs(self.priorities) do
```
TheCycoONE: Why did you end up changing this from pairs to ipairs?

TheCycoONE:
Doesn't help that their site is down right now. I don't have to change anything about lpath, but I've been running it exclusively on Linux. I'll look at Windows later.

I'm happy that you managed to get some tests going. I'll separate out the mocks later so tests can share better - as I mentioned, that whole aspect is a bit underdeveloped in CorsixTH at the moment.

Really what's left is going through the announcements and assigning sane priorities / decay times I think.

CorsixTH/Lua/announcer.lua
Feanathiel: I've read (http://lua-users.org/wiki/ForTutorial) that the order of pairs is not defined, which is not what we want if critical announcements need to be played before low priority annoucements. ipairs iterates over the table in a defined numeric order.

TheCycoONE: ipairs iterates in index order, but your table doesn't use numeric indexes so it's moot. If you need guaranteed order you will have to change it into an array instead.

```lua
  self.priorities = {
    AnnouncementPriority.Critical,
    AnnouncementPriority.High,
    AnnouncementPriority.Normal,
    AnnouncementPriority.Low
  }
```

TheCycoONE: Which doesn't work at all ... keep another table (array) of the priorities in order and loop through that instead.

TheCycoONE: ... disregard everything, those are numbers. Sorry.

CorsixTH/Lua/hospital.lua
```diff
@@ -881,15 +885,15 @@ function Hospital:coldWarning()
     "sorry002.wav", "sorry004.wav",
   }
   if announcements and self:isPlayerHospital() then
-    self.world.ui:playAnnouncement(announcements[math.random(1, #announcements)])
+    self.world.ui:playAnnouncement(announcements[math.random(1, #announcements)], AnnouncementPriority.Low)
```
TheCycoONE: I think Hot and Cold warnings should be priority Normal; save Low for the random banter?

Feanathiel: Yeah, I think that makes more sense. That way the heat issue can be fixed faster by the player.

TheCycoONE:
Merged

PR #1446, opened 2018-08-23
[RDY] Add man page for corsix-th

TheCycoONE:
New man page inspired by the one included in Debian. This one is written in mdoc; which has the advantage of being a largely semantic markup producing consistent man pages.

I plan to sort the resource files dumped into /CorsixTH in the future; at present this seems like it belongs; which allows me to put it in the CorsixTH CMake file, and likewise add a man page for e.g. AnimView that is only installed when AnimView is included in the build.

I do plan to add the install to cmake as part of this PR. I will likely also add a files section similar to what Debian has.

### DIFF
--- a/CorsixTH/corsix-th.6
+++ b/CorsixTH/corsix-th.6
@@ -1,0 +1,94 @@
+.\" Copyright (c) 2018 Stephen "TheCycoONE" Baker
+.\"
+.\" Permission is hereby granted, free of charge, to any person obtaining a copy of
+.\" this software and associated documentation files (the "Software"), to deal in
+.\" the Software without restriction, including without limitation the rights to
+.\" use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+.\" of the Software, and to permit persons to whom the Software is furnished to do
+.\" so, subject to the following conditions:
+.\" 
+.\" The above copyright notice and this permission notice shall be included in all
+.\" copies or substantial portions of the Software.
+.\" 
+.\" THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


TheCycoONE:
Need to exclude Apple

CorsixTH/corsix-th.6
```diff
@@ -0,0 +1,84 @@
...
+.\" Language reference: https://man.openbsd.org/mdoc.7
+.Dd August 22, 2018
+.Dt CORSIX-TH 6
+.Os
+.Sh NAME
+.Nm corsix-th
+.Nd "An" open source implementation "of" Theme Hospital
+.Sh SYNOPSIS
+.Nm
+.Op Fl -bitmap-dir Ns Ar =path
```
Alberth289346: Is it intentional that the `=` is concatenated with `path` here? other options do it differently, not to mention line 44

CorsixTH/corsix-th.6
```diff
@@ -0,0 +1,84 @@
...
+.Op Fl -config-file Ns = Ns Ar path
+.Op Fl -connect-lua-dbgp
+.Op Fl -dump=strings
+.Op Fl -interpreter Ns = Ns Ar path
+.Op Fl -load Ns = Ns Ar savegame
+.Op Fl -lua-dir Ns = Ns Ar path
+.Sh DESCRIPTION
+CorsixTH is a reimplementation of the 1997 Bullfrog business simulation game Theme Hospital.
+As well as faithfully recreating the original, CorsixTH adds support for modern operating systems, high resolution displays, and much more.
+A purchased copy of Theme Hospital, or the Theme Hospital demo is required to use CorsixTH.
```
Alberth289346: Isn't the demo useful for only a few levels? It now looks asif it's a full alternative for the purchased copy.
Would be fair to warn about that, I think.

Alberth289346:
Just minor stuff, looking great, nice job

Issue #1429, opened 2018-07-25
Dying patients and staff in crashed room don't Despawn


mugmuggy:
#### Describe the issue
I was playing around with the issue in #1394 and came across these issues;
* of patients remaining in the hospital.patients table when they would be expected to be removed. 
* of staff processing ticks and failing to find nearby objects

Room:crashRoom calls Patient:die() but if patient is in a room, it just queues up a Meander and Die action. As the patient might not be processing an action that now completes and perform those actions (including walking to Screen after operating table etc). They can never complete and the last part of the die action is Patient:despawn and World:destroyEntity and so remain being processed throughout.

The staff issue is essentially a dupe of #1195 and #1265 and has the same underlying cause related to the above, humanoid:despawn in this case creates a Spawn action with mode despawn to initiate the removal of the staff member from the world.entities which never gets executed. Eventually the staff member ticks, and they try to find nearby objects (litter) which causes a failure.

#### Save Game
I just used the savegame from #1401 sacked all handymen to allow room crashing and observed the issue.

#### System Information
**CorsixTH Version:** observed on 0.62-rc2
**Theme Hospital Version:** Origin

#### Gamelog.txt
N/A

PR #1451, opened 2018-09-01
[RDY] Crashed room despawn

mugmuggy:
Addresses #1429.

One additional change here was to remove the humanoidDeath call from Staff:die() as staff deaths didn't count to death count in TH or affect reputation (room being destroyed did).

There is one outstanding issue which might contribute (or not) to some bugs, is that Room:crashRoom calls World:destroyEntity, which removes multiple entities. From my understanding that can result in skipped tick calls as calling table.remove from within an iteration from ipairs is bad.

### DIFF
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -787,7 +787,7 @@ function Room:crashRoom()
   for humanoid, _ in pairs(self.humanoids) do
     remove_humanoid(humanoid)
   end
-  self.humanoids = nil
+  self.humanoids = {}
   -- There might also be someone using the door, even if that person is just about to exit
   -- he/she is killed too.
   local walker = self.door.user
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -779,6 +779,7 @@ function Room:crashRoom()
       table.remove(self.world:getLocalPlayerHospital().emergency_patients, humanoid.is_emergency)
     end
     humanoid:die()
+    humanoid:despawn()
     self.world:destroyEntity(humanoid)
   end
 
@@ -786,6 +787,7 @@ function Room:crashRoom()
   for humanoid, _ in pairs(self.humanoids) do
     remove_humanoid(humanoid)
   end
+  self.humanoids = nil
   -- There might also be someone using the door, even if that person is just about to exit
   -- he/she is killed too.
   local walker = self.door.user
--- a/CorsixTH/Lua/entities/staff.lua
+++ b/CorsixTH/Lua/entities/staff.lua
@@ -148,7 +148,7 @@ end
 function Staff:tick()
   Entity.tick(self)
   -- don't do anything if they're fired or picked up or have no hospital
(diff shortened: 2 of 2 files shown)


CorsixTH/Lua/room.lua
```diff
@@ -779,13 +779,15 @@ function Room:crashRoom()
...
     humanoid:die()
+    humanoid:despawn()
     self.world:destroyEntity(humanoid)
   end
 
   -- Remove all humanoids in the room
   for humanoid, _ in pairs(self.humanoids) do
     remove_humanoid(humanoid)
   end
+  self.humanoids = nil
```
TheCycoONE: nit pick, but I think I'd prefer you set this back to the empty set rather than nil, similar to what happens to `humanoids_enroute` in deactivate

TheCycoONE:
Action case is not safe, action calls machineUsed, called inside ipairs loop over entities in world:onTick

TheCycoONE:
I will create a new issue. crashRoom shouldn't happen in machineUsed, it should happen in a new loop in world near the end of tick()

mugmuggy:
Yep, I didn't update the issue, but I realise I made a mistake in that - the staff:despawn was being called - but it was the Staff:tick and the action being called that was stuffing it all up.

I haven't been able to find an obvious issue with destroyEntity, ie litter removal is also called from an action and happens all the time but we are probably only talking about 1 or 2 animation ticks when they are skipped because of that removal.

TheCycoONE:
If the destroyed entity is earlier in the entity list than the current entity, or is the current entity, then the next entity in the list will have it's `tick()` skipped. This cascades for every removal, so if 5 entities are removed (<= current), five will be skipped that loop through. Removed entities ahead of the current one cause no effect.

`ipairs` is basically equivalent to `for (i = 1; table[i] != nil; i++) { res = table[i]; ... }`, and `table.remove(t, x)` does something like `for (i = x; table[i] != nil; i++) { table[i] = table[i + 1] }` (excuse the C style notation)

An additional benefit to doing all the removals at once is that we can use a single pass (worse case order N, instead of order N^2)

mugmuggy:
I've done 3 solutions to this problem, as I noticed it before but got side tracked. I implemented a fix to change destroyEntity to set e.destroyed = true and processed in a loop after the enitity:tick() avoiding the 'tick' also within that loop.

And I believe with the additional check in entity:tick anyway, it would be just as good to process self.entities in reverse.
```
  for i = #self.entities, 1, -1 do
    if self.entities[i].destroyed then
      table.remove(self.entities, i)
    elseif self.entities[i].ticks then
      self.current_tick_entity = self.entities[i]
      self.entities[i]:tick()
    end
  end

TheCycoONE:
```lua
local destroyed_entities = 0
local entity_count = #self.entities
for i = 1, entity_count, 1 do
  if self.entities[i].destroyed then
    destroyed_entities += 1
    self.entities[i] = nil
  else
    self.entities[i - destroyed_entities] = self.entities[i]
  end
end

Issue #1425, opened 2018-07-19
Crash due to patient going to treatment being re-routed on error to GP, but going to treatment once it is its turn in GP


ElvarT:
#### Describe the issue
As soon as a GP's office was built, a baldness patient queued for it. (The bold patient in the middle of the screen).
But when a doctor arrives to the office, the patient instead of entering the GP office goes to the hair restoration, while its status remains "Queueing for GP's office".
As soon as it reaches the hair restoration clinic the crash happens.

#### Steps to Reproduce (A)
1. Load this savegame: https://nofile.io/f/ND0X6zEdKvp/befcrash2.sav.7z

2. Either just wait ~30 seconds for a doctor to arrive, or drop a doctor in the GP's office. See patient go to hair restoration.

3. Wait for him to get there and wait for crash.

#### Another way to reproduce from a few seconds earlier (B):
1. Load this other savegame: https://nofile.io/f/zZ0FuuD0v0s/causing_boldguy_bug.sav.7z

2. If you complete the GP's office build the bold guy will queue for it as described before.

2.b. If you cancel the build, the bold guy will continue in his way to the hair restoration without problems.

#### System Information
**CorsixTH Version:** 0.61

**Operating System:** Ubuntu 18.04

**Theme Hospital Version:** CD

Note: It is the same run as in here: https://github.com/CorsixTH/CorsixTH/issues/1329#issuecomment-406053152, a few minutes after working around the crash described there.

----------------

ElvarT:
If I send that patient to the reception desk (from the GP's office queue) he will go to the reception, then go to the hair restoration directly, and complete treatment without problems.

Note: In this workaround you can see a small issue (maybe worth its own GH issue?). The patient is placed in the reception desk queue as soon as he is routed manually from the UI. This blocks the reception desk until that patient is handled.

ElvarT:
In that same savegame you can observe an unrelated issue (should I create another GH issue for it?)
In the Operating Theater shown in the picture, there is a "confused" surgeon (circled in red), undressed and ready to get out, while at the same time waiting for the next patient (circled in blue). This patient cannot go in as the surgeon is not dressed. At the same time there is another surgeon (blue) waiting to replace the 1st surgeon.
![stucksurgeon_marked](https://user-images.githubusercontent.com/41392927/42930323-f597f552-8b3c-11e8-9bd3-26ec802be071.jpg)
That 1st surgeon will not get out, and the queue will just grow.

[...]

PR #1452, opened 2018-09-01
[RDY] Reroute if door/wall has been built on walk path

mugmuggy:
Addresses #1425 where a room is built on the already calculated walk path, and the walk path is interrupted by a door.


TheCycoONE:
This looks good to me. RDY?

Issue #1439, opened 2018-08-13
LUA Error: empty queue/time issue


yuk75:
So I have a game where staff are getting stuck in rooms and it's been causing patients to die...like the staff are frozen and they are in the room just waiting. I reloaded and it didn't help so I removed the doctors and placed them back down, didn't change things they acted like they couldn't get to their desk (this one was a GP). So I removed the doc from the room and sent the patient in question to be treated, then reloaded the game just in case. Later on my game stopped responding (everything except UI frozen) and I minimised to see an error in the console window which I will put here, it's probably not related? But I thought I should also mention the issue prior in the off chance it is related...

```


---------------------------------------------------------------

Welcome to CorsixTH v0.62!

---------------------------------------------------------------

This window will display useful information if an error occurs.

---------------------------------------------------------------

Checking for CorsixTH updates...
You are running the latest version of CorsixTH.
An error has occurred!
Almost anything can be the cause, but the detailed information below can help th
e developers find the source of the error.
Running: The timer handler.
A stack trace is included below, and the handler has been disconnected.
Lua\entities\humanoid.lua:868: Action queue was empty. This should never happen.

humanoid - class: Standard Male Patient
Warmth: 0.258   Happiness: 1.000   Fatigue: 0.000  Thirst: 0.231  Toilet_Need: 0
.239   Health: 0.000
Actions: []
stack traceback:
        [C]: in function 'error'
        Lua\entities\humanoid.lua:868: in method 'getCurrentAction'
        Lua\hospital.lua:690: in method 'countSittingStanding'
        Lua\hospital.lua:720: in method 'checkFacilities'
        Lua\hospital.lua:974: in method 'onEndDay'
        Lua\world.lua:998: in method 'onTick'
        Lua\app.lua:1014: in function <Lua\app.lua:1011>
        (...tail calls...)
        Lua\app.lua:915: in function <Lua\app.lua:910>

Warning: No event handler for timer
```

My gamelog.txt
```
Created game with savegame version 111.
Savegame version is 111 (v0.60), originally it was 111 (v0.60)
Savegame version is 111 (v0.60), originally it was 111 (v0.60)
Warning: Trying to remove nonexistant room build callback (function: 000000000FCE33B0) from humanoid (table: 0000000010280890).
Warning: Trying to remove nonexistant room remove callback (function: 0000000004D1F180) from humanoid (table: 0000000010280890).
[...]

PR #1453, opened 2018-09-01
[RDY] Interrupted walk should refer to original must_happen

mugmuggy:
Addresses #1439. Where a walk action is interrupted to a drinks machine resulting in an incorrect 'must_happen' walk after the drinks machine use.

### DIFF
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -372,7 +372,7 @@ function Patient:falling()
         name = current.name,
         x = current.x,
         y = current.y,
-        must_happen = current.must_happen,
+        must_happen = (current.name == "walk" and current.saved_must_happen) or (current.name == "idle" and current.must_happen),
         is_entering = current.is_entering,
       }, 2)
     else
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/entities/patient.lua
```diff
@@ -879,7 +879,7 @@ function Patient:tickDay()
             name = current.name,
             x = current.x,
             y = current.y,
-            must_happen = current.must_happen,
+            must_happen = current.name == "walk" and current.saved_must_happen or current.name == "idle" and current.must_happen,
```
TheCycoONE: Please make this clearer with brackets, it's too easy to forget precedence of `and` and `or`. As is it's a bit hard to read.

mugmuggy:
I've noticed the same logical bug in Patient:falling, vomit and pee methods, I'll push another commit to cover them also.

TheCycoONE:
I think this needs cleaned up later

mugmuggy:
I assume on the cleanup - there being 4 bits of code that basically do the same thing - that a refactor is in mind into a single function call. In preparing another PR I found some minor cosmetic issues in game around how that bit of code functions, so I plan to include it there, if that PR doesn't get up, then I can pull it out from there.

Issue #1468, opened 2018-09-16
CorsixTH\Lua\graphics.lua:260: assertion failed! - Error during startup


Nox13last:
#### Describe the issue
When loading the game, it refuses to reach the main menu. It displays the following error and then awaits the user pressing the red Exit button..

```An error has occurred in CorsixTH:
C:\Program Files\CorsixTH\Lua\graphics.lua:260: assertion failed!
stack traceback:
        [C]: in function 'assert'
        C:\Program Files\CorsixTH\Lua\graphics.lua:260: in method 'loadRaw'
        C:\Program Files\CorsixTH\Lua\ui.lua:402: in method 'setMenuBackground'
        C:\Program Files\CorsixTH\Lua\app.lua:284: in method 'init'
        C:\Program Files\CorsixTH\CorsixTH.lua:120: in main chunk
        [C]: in ?
```

Issue seems to exist in both 0.61 (with files copied from a computer with a working installation) and 0.62

#### Steps to Reproduce
0. Ensure the Prerequisite frameworks are installed.
1. Download CorsixTHInstaller.exe
2. Run it.
3. Specify where the \HOSPITAL\ directory is.
4. Run the CorsixTH.exe.lnk shortcut from the Start Menu.

#### Expected Behavior
The game loads to the main menu, asking either how the user wants to engage the game, or asking for the location of the original game's HOSPITAL directory.

#### System Information
**CorsixTH Version:** *0.61, 0.62*

**Operating System:** *Windows 7*

**Theme Hospital Version:** *CD*

#### Gamelog.txt
_(Gamelog.txt is not available, at least not where it is "supposed" to be. It seems it is not generated until after startup.)_

### Additional Info
**P.S.:** The Android fork is missing. It would be nice to have that back, too.

TheCycoONE:
The android port was an independent effort by @alanwoolley. I haven't heard from him in some time.

https://github.com/CorsixTH/CorsixTH/blob/8f31878e4fbd4805a5cc8bc4702d87991735722a/CorsixTH/Lua/graphics.lua#L260

The assertion suggests your original game data is corrupt. Where did you get it?

TheCycoONE:
Actually that is the main menu graphic that should be installed with the game, not original game data...

TheCycoONE:
Let me know what the result is of running the portable version, extracted to a different location.

Nox13last:
> The assertion suggests your original game data is corrupt. Where did you get it?

[...]

PR #1470, opened 2018-09-19
[RDY] Show errors when loading bitmaps

TheCycoONE:
Use exceptions and lua_error to show the cause of any error loading bitmaps
instead of simply false. This makes diagnosing errors much easier.

This patch was used to diagnose #1468 but does not fix the issue.


TheCycoONE:
Missed an include that MSVC didn't care about.

Alberth289346:
Nice! Sorry for not getting online, but terribly busy

PR #1471, opened 2018-09-19
[RDY] Annual report changes from AI Hospitals PR

mugmuggy:
The ordering of fields is broken when multiple hospitals have the same value and they aren't in index order. So my change insteads passes through the index of the hospital being processed, finds the index of that hospital in the sorted list and returns that directly instead of using the 'dup_' values to try and do this.

The current code will draw names over the top of another and can leave with gaps in the list when multiple hospitals have the same value.

Sort functions changed to always put the local player hospital first assuming its at index 1.

I also aligned the date of the first year with TH as the Date class has this as a 1 indexed value, and its already incremented at this point for the next year.

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/annual_report.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/annual_report.lua
@@ -133,42 +133,33 @@ function UIAnnualReport:UIAnnualReport(ui, world)
 
   -- Get and sort values used on the statistics screen.
   -- The six categories. The extra tables are used to be able to sort the values.
-    self.money = {}
     self.money_sort = {}
-    self.visitors = {}
     self.visitors_sort = {}
-    self.salary = {}
     self.salary_sort = {}
-    self.deaths = {}
     self.deaths_sort = {}
-    self.cures = {}
     self.cures_sort = {}
-    self.value = {}
     self.value_sort = {}
 
     -- TODO: Right now there are no real competitors, they all have initial values.
     for i, hospital in ipairs(world.hospitals) do
-      self.money[hospital.name] = hospital.balance - hospital.loan
-      self.money_sort[i] = hospital.balance - hospital.loan
-      self.visitors[hospital.name] = hospital.num_visitors
-      self.visitors_sort[i] = hospital.num_visitors
-      self.deaths[hospital.name] = hospital.num_deaths
-      self.deaths_sort[i] = hospital.num_deaths
-      self.cures[hospital.name] = hospital.num_cured
-      self.cures_sort[i] = hospital.num_cured
-      self.value[hospital.name] = hospital.value
-      self.value_sort[i] = hospital.value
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/fullscreen/annual_report.lua
```diff
@@ -133,39 +133,28 @@ function UIAnnualReport:UIAnnualReport(ui, world)
...
+      self.money_sort[i] = {hospital.balance - hospital.loan, i}
+      self.visitors_sort[i] = {hospital.num_visitors, i}
+      self.deaths_sort[i] = {hospital.num_deaths, i}
+      self.cures_sort[i] = {hospital.num_cured, i}
+      self.value_sort[i] = {hospital.value, i}
+      self.salary_sort[i] = {hospital.player_salary, i}
     end
 
-    local sort_order = function(a,b) return a>b end
+    local sort_order = function(a,b) return a[1] > b[1] or (a[1] == b[1] and a[2] == 1) end
```
TheCycoONE: Naming nit: asc_order / desc_order. Also comment that it keeps the player on top for equivalence, that was not obvious.
Naming the keys and giving the player_hospital a variable would reduce the number of magic numbers and make the code more readable IMO:

```lua
for i, hospital in ipairs(world.hospitals) do
  self.money_sort[i] = { value: hospital.balance - hospital.loan, hosp_index: i }
  ...
end

-- When sorting keep the player hospital on top when values are the same
local player_index = 1
local desc_order = function(a, b) return a['value'] > b['value'] or (a['value' == b['value'] and a[hosp_index] == player_index) end
local asc_order = function(a, b) return a['value'] < b['value'] or (a['value'] == b['value'] and a[hosp_index] == player_index) end
```

CorsixTH/Lua/dialogs/fullscreen/annual_report.lua
```diff
@@ -584,82 +574,64 @@ function UIAnnualReport:drawStatisticsScreen(canvas, x, y)
...
   local row_no_y = 106
-  local dup_money = 0
-  local dup_salary = 0
-  local dup_cures = 0
-  local dup_deaths = 0
-  local dup_visitors = 0
-  local dup_value = 0
-  for _, player in ipairs(world.hospitals) do
+
+  for idx, player in ipairs(world.hospitals) do
```
TheCycoONE: It'd be nice if this was switched from player to hospital to avoid confusion and be consistent with the loop above. For the same reason `idx` should match `i` above. I prefer `i` or `hosp_index` to `idx`.

TheCycoONE:
looks good to me

mugmuggy:
yeah I also removed magic numbers from the rest of it, tested it and it looked ok still.

TheCycoONE:
It occurs to me we could (and should) have used isPlayerHospital() rather than hard coding 1. It won't matter until multiplayer though.

PR #1472, opened 2018-09-20
[RDY] mainmenu background sizes

TheCycoONE:
Adds a couple smaller mainmenu background images (640x480 and 1280x720) and selects the one with the height no larger than your current screen (or the smallest image.)

Intended to address #1468


TheCycoONE:
Fix confirmed by @Nox13last

Issue #1315, opened 2017-12-28
Segfault if ffmpeg doesn't support smacker decoding.


Mailaender:
#### Steps to Reproduce
1. Install the game from https://software.opensuse.org/package/CorsixTH
2. Start the game

#### Save Game
None

#### Expected Behavior
Startup the game

#### System Information
**CorsixTH Version:** *both 0.60 and 0.61*

**Operating System:** @openSUSE Leap

**Theme Hospital Version:** *GOG.com*

#### Gamelog.txt
None written

### Additional Info
```
 CorsixTH 


---------------------------------------------------------------

Welcome to CorsixTH v0.61!

---------------------------------------------------------------

This window will display useful information if an error occurs.

---------------------------------------------------------------

/usr/bin/CorsixTH: Zeile 3: 18667 Speicherzugriffsfehler  (Speicherabzug geschrieben) ./CorsixTH
/home/matthias/.config/CorsixTH/Levels
```

TheCycoONE:
Not much to go on without debugging info. Are you able to compile yourself or does opensuse provide debug packages?

Alberth289346:
Discussions at SO suggest the computer tries to execute code at a place where no executable code is expected (such as a data area or a stack area). However, as German is not my native language, I don't quite understand what it says (yay for translated error messages in an international project).

We don't provide OpenSuse packages afaik, so your best bet is to report to whoever supplied you your package (probably opensuse). Once they established the fault is not in their code, they likely report an upstream error back to us.
I briefly looked for corsixth errors at opensuse, but they don't seem to exist, at least without having an account.

Mailaender:
Updated https://build.opensuse.org/request/show/560427 and enabled debug packages now. "Speicherzugriffsfehler" means segmentation fault. I don't really know how to give a more detailed error message. I tried deleting the config file. That made the game ask me again for the Theme Hospital folder, then it restarted and crashed again. Tried different folder locations. It doesn't seem to matter. See http://en.opensuse.org/CorsixTH for the instructions I followed. Maybe the error is in there.

TheCycoONE:
Ah, you are the package manager!

Ok, now that you have debug symbols, run the game through gdb. When it segfaults type `bt` and let us know the trace.

I also recommend switching to lua5.2 or 5.3 these days. Should still run with 5.1 but it has been awhile since I tried.

bobcanthelpyou:
[...]

PR #1473, opened 2018-09-20
[RDY] Handle when no supported video stream is found

TheCycoONE:
Fixes a segfault in environments where smacker support is disabled in ffmpg.

Bug #1315


TheCycoONE:
RFC because this is untested.

TheCycoONE:
Switched to the same error handling we use elsewhere in this routine. A future refactor could switch to exceptions.

TheCycoONE:
Marking RDY. I don't hold out much hope in attracting people from the issue.

PR #1474, opened 2018-09-23
[RDY] Litter placed before a door was placed was only found on thob checks

mugmuggy:
Using getObject with a table input to find either door type instead. Refers to #1425 still

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/walk.lua
+++ b/CorsixTH/Lua/humanoid_actions/walk.lua
@@ -206,12 +206,12 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
     -- if going to the corridor we don't care as we are at the door
     -- as this will be false we won't bother checking for rerouting
     recalc_route = flags_there.room
-    local door = TheApp.objects.door.thob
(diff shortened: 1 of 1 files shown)


TheCycoONE:
@mugmuggy I think https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/dialogs/edit_room.lua#L1153 is the same issue - if litter and something else was on the same tile, the thob could be 62 and allow a door where it shouldn't?

mugmuggy:
Thats if any object but not if its litter prevent the door being placed. Litter is removed from all tiles inside a room, so its only outside where this becomes an issue. The other thing is that as it is an 'or' it only breaks on particular room orientations as it may check inside the room first, find the door and skip the rest of the statements.

Nothing stopping you building a room with a door where the adjacent tile in corridor has litter. Unfortunately when I was testing the other future pr, I had vomit everywhere that then resulted an in infinite loop of recalculating routes, and upon checking the thob found the it was 62 instead of 3.

The old walk code would redirect only where there was a wall found, so it could have also been entirely possible at one stage to build a room on the path with a door and therefore just walked over litter as it was never rerouting unless it hit that wall.

Issue #1206, opened 2017-04-05
Insane stuck patients near psychiatry and invisible soda machine


l29ah:
https://dump.bitcheese.net/files/sipamub/buggie.sav
ea0e6dfe1ab6444ddcd876a9b4fa2d78172c463d, Gentoo Linux, nothing interesting in the log.

l29ah:
The soda machine is next to them but it turns visible only when someone uses it.

mugmuggy:
I happened upon an invisible soda machine in #867 and that was on 0.50 and this one is post 0.60

PR #1475, opened 2018-09-23
[RDY] Invisible drinks machine

mugmuggy:
Finally tracked down the invisible drinks machine (mentioned in #1206).

From memory checking watch, yawns and tapping foot were only done whilst waiting in queue or whilst waiting for a room to be built, I can't recall it very well.

As the patient had to be waiting for a room to be built in this instance, waiting for 30 or 50 days and also happen to be using the drinks machine at the same time, it happens rarely. I'm not sure if there is any value in the is_leaving, I assume that was put in some time back for some purpose but I've replicated it across to the yawn as well. Likewise even though the checkWatch/tapFoot calls were only ever done within the Patient:tickDay checks at the start, and therefore patient.waiting had to be ticking down, they would never be called for an idle state from my quick perusal.

Put this as RFC for the fact I'm not sure on the correct state.

### DIFF
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -494,20 +494,22 @@ function Patient:pee()
 end
 
 function Patient:checkWatch()
-  if self.check_watch_anim and not self:getCurrentAction().is_leaving then
+  local action = self:getCurrentAction()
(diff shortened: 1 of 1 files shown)


TheCycoONE:
I'm inclined to accept this as is.

mugmuggy:
Thats fine moved to RDY

Issue #1436, opened 2018-08-06
Epidemic evacuation can lead to invalid action states


mugmuggy:
#### Describe the issue
Patients vomiting outside the hospital.

A patient whilst leaving the hospital as part of an epidemic evacation, receives a VomitAction as part of Patient:tickDay.

This inserts itself as 2nd in action queue:
```lua
    self:queueAction(VomitAction(), 1)
```
Patient proceeds to spawn point (initial walk is not interrupted), then vomit action is started whilst outside the hospital and litter outside the hospital should not be possible.

#### Steps to Reproduce
1. Used save attached to 1394, this has an evacuation in progress.

#### Save Game
[Save from 1394](https://github.com/CorsixTH/CorsixTH/files/2034833/Ulti.XT.zip)

#### Expected Behavior
No vomiting outside hospital.
Epidemic:evacuate should probably just do a Patient:goHome("kicked") type call, but then all patients will affect statistics which I can't recall the original behaviour.  This should prevent the vomit action being added after being kicked.

#### System Information
**CorsixTH Version:** 0.62

#### Gamelog.txt
```
Savegame version changed from 122 (v0.61) to 127 (v0.62). The save was created using 122 (v0.61)
Error in timer handler: 
Lua\objects\litter.lua:90: attempt to index a nil value (local 'hospital')
stack traceback:
	Lua\objects\litter.lua:90: in method 'setLitterType'
	Lua\humanoid_actions\vomit.lua:33: in local 'timer_function'
	Lua\entity.lua:216: in method 'tick'
	Lua\world.lua:1036: in method 'onTick'
	Lua\app.lua:1014: in function <Lua\app.lua:1011>
	(...tail calls...)
	Lua\app.lua:915: in function <Lua\app.lua:910>
```

mugmuggy:
Found another case where a patient has died, fallen, lava hole and grim reaper has spawned, grim has walked to lava and patient has stood up but then evacuate hospital has been called and patient proceeds to walk out of the hospital. This leaves the grim reaper stuck in hospital.

TheCycoONE:
Was the vomit outside the hospital caused by invalid map flags? This ticket reads like that one

mugmuggy:
The decision to vomit appears to be when within the hospital itself, and the action is queued. However i don't think a leaving patient was supposed to vomit even within the hospital. Patient:canPeeOrPuke guards the vomit call.

This guards it:

```lua
[...]

PR #1476, opened 2018-09-24
[RDY] Implement a goHome handler for evacuated patients

mugmuggy:
Addressing #1436. I can't recall if an evacuation affected the statistics in the original for not treated, but have stopped the percentages being updated here.

The only negative is that goHome will get called twice, when a patient is 'dealtWithPatient' call is made in a callback after an action completes, the warning will be printed.

### DIFF
--- a/CorsixTH/Lua/entities/patient.lua
+++ b/CorsixTH/Lua/entities/patient.lua
@@ -556,7 +556,7 @@ function Patient:goHome(reason, disease_id)
 
   elseif reason == "evacuated" then
     self:clearDynamicInfo()
-    self:updateDynamicInfo('text', {_S.dynamic_info.patient.actions.epidemic_sent_home})
+    self:setDynamicInfo('text', {_S.dynamic_info.patient.actions.epidemic_sent_home})
     self:setMood("exit","activate")
 
   else
@@ -583,7 +583,7 @@ function Patient:goHome(reason, disease_id)
   if room then
     room:makeHumanoidLeave(self)
   end
-  self:despawn()
+  Humanoid.despawn(self)
 end
 
 -- Despawns the patient and removes them from the hospital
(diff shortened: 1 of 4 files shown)


MarkL1961:
Regarding "I can't recall if an evacuation affected the statistics in the original " see https://github.com/sadger/CorsixTH/issues/34  this was something we discussed when implementing epidemics and the feeling was you lose rep on sent home already so on a really busy level it would kill the game if you got stung twice if the evac also gave you a rep loss.

mugmuggy:
I realise that I don't need that check, around updatePercentages as I have excluded in the code updateNotCuredCounts which has the rep hit. So I do think I can get away with just removing that if, it doesn't do anything anyway.

TheCycoONE:
I like this change as is.

mugmuggy:
There are some issues here I need to address, one - ipairs as Patient:despawn calls Hospital:RemoveByValue via RemovePatient the other is I bad copy pasted from another branch where I was side by side testing this with updates and bungled the update/setDynamicInfo.

For the Patient:despawn issue I could call Humanoid.despawn(self) which enqueues the despawn action eventually and have that then call the Patient:despawn method as it currently does to finally remove the patient. But this then delays the patient from affecting warmth/thirst attributes as it currently also does with epidemic evacuations, but for everyone 'going_home' this would be the case.

mugmuggy:
I was just checking the original and how the stats progress dialog looks, when a patient is outside the bounds of the hospital - they don't count to the patient count (on town map) or affect warmth/thirst/happiness status.

However, if you send someone home and they are within the bounds of any walls, then they contribute to the statistics, if they reenter a building they recontribute to the statistics. They leave a building they drop from the statistics. So patients 'going_home' now, are dropped early as Patient:despawn calls RemovePatient. Which might give me scope to use the Humanoid.despawn(self) option and leave the removal from the hospital to the SpawnAction, however then adjusting stats to deal with patients between buildings outside the hospital.

Likewise they are also added too early to the stats, before they enter the bounds of the hospital. Which I thought I mentioned in another Issue but may not have got to it.

TheCycoONE:
Thanks @mugmuggy

The original behavior seems better, and fairly easy to implement.

mugmuggy:
This should be all ok now

mugmuggy:
In some testing I think there might be an issue with the latest changes, I'll mark it RDY again once I can hopefully confirm the issue, just I have a few changes to fix other issues that cause my other changes to break.

PR #1478, opened 2018-09-27
[RFC] Chinese and memory leak fixes

TheCycoONE:
Adds line breaking after Chinese punctuation characters, and fixes a couple memory leaks relating to freetype rendering.


TheCycoONE:
@Icenowy could you confirm that this displays the text correctly. It looks right to me, but since I can't read it I want to make sure I'm not corrupting characters or something.

CorsixTH/Src/th_gfx_font.cpp
```diff
@@ -702,7 +710,7 @@ text_layout freetype_font::draw_text_wrapped(render_target* pCanvas, const char*
             // Calculate the line height and baseline position.
             FT_Pos iLineHeight = 0;
             FT_Pos iBaselinePos = 0;
-            for(const char* s = itr->first; s != itr->second; )
+            for(const char* s = itr->first; s < itr->second; )
```
TheCycoONE: These changes aren't actually necessary - but I feel more comfortable with them. I got a buffer overflow when I messed up the number of skipped bytes above originally.

CorsixTH/Src/th_gfx_font.cpp
```diff
@@ -653,8 +653,16 @@ text_layout freetype_font::draw_text_wrapped(render_target* pCanvas, const char*
...
-            if(iCode == ' ')
+            if (iCode == ' ') {
                 sLineBreakPosition = sOldMessage;
+            } else if (iCode == 0x3002 ||
+                    iCode == 0xff0c ||
+                    iCode == 0xff0d ||
+                    iCode == 0xff1b ||
+                    iCode == 0xff1f) {
+
+                sLineBreakPosition = sOldMessage + 3; // break after this codepoint (3 bytes in utf-8)
```
Alberth289346: I assume the `' '` matched at 656 gets removed (by the looks of it at 633?), so would that mean the character after the line break match you set here also gets removed?

TheCycoONE:
Oh, I also added ideographic space (0x3000) to the list of characters, since it seemed like it fit in.

Issue #1480, opened 2018-10-06
Auto-autospy should be 'rediscoverable'


mugmuggy:
#### Describe the issue
Patients sent to the research room currently only allow discovery of the auto autospy once when its entirely possible to have it discovered mutliple times in the original game - just its probably unlikely as research may complete unless the research values are set low (ie 0) or something leads to a treatment room being completely removed and a patient being diagnosed with that disease

The chance of discovery is also less initially in CorsixTH, 10%, growing in steps of 10% until it is discovered once. The original either has a 1/3rd (average staff happiness >= 0.25), or 2/3rds chance (< 0.25) of discovery every use.

#### Steps to Reproduce
1. Observed on L2 sleepy hollow in Origin with research, gp and general diagnosis built and sending all patients to research.

#### System Information
**CorsixTH Version:** 0.62

mugmuggy:
Merged

PR #1483, opened 2018-10-11
[RDY] Autospy can be discovered multiple times - morale related chances

mugmuggy:
For Issue #1480


TheCycoONE:
Reads ok to me now 👍 
RDY?

Issue #469, opened 2014-12-23
Missing fax when built room has no staff available


TheCycoONE:
_From [mark.l1...@hotmail.co.uk](https://code.google.com/u/117099726060447259318/) on June 27, 2011 14:55:48_

What steps will reproduce the problem? 1. build op theatre 
2. be one or both surgeons short for it to be usable What is the expected output? What do you see instead? Dr diagnoses someone that needs surgery and sends them to the ward.
The casebook tells you that you can't treat this and tells you why, the same check should occur at diagnosis and the Dr should either send them home or you ought to get a fax so that you can decide.

I think that the same can happen with the pharmacy.  You build it, forget to employ a nurse and yet patients are still sent there.

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=906_

mugmuggy:
Yep

Issue #453, opened 2014-12-23
Hire a researcher should update existing faxes with autopsy option


TheCycoONE:
_From [roujin.mw](https://code.google.com/u/roujin.mw/) on November 05, 2010 08:24:57_

Currently it is decided at creation time of Messages if the autopsy option is available or not (depending on if a research room was built).

However with the possibility now to hold back a decision until later (with the new fax system in r999 ), it should be made possible to build a research room and then open the fax again and choose autopsy. This is currently not possible.

Care should be taken that even opened faxes are updated (someone with a high resolution could manage to build a research room while keeping the fax opened).

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=480_

Alberth289346:
In short, the options displayed by the fax should be computed at the moment it is opened rather than being generated beforehand. That way, new options to deal with the situation will also be mentioned in the fax.

MarkL1961:
What if you were to minimise the fax and they say build the research department?
Would the fax update to show the option of send to research?

Alberth289346:
That's how I read the enhancement indeed. Do you read something else?

For that reason, I believe that at the moment the fax gets opened for display, it should decide what to show. If you minimize the fax, change the hospital, and then open it again (re-computing what the fax show display), it will show the correct choices.

MarkL1961:
I wanted to be sure as it did not read that way to me in the previous comment.  If re-opening does a recalculate that is indeed what is needed

mugmuggy:
Digging this issue out from the grave.
This [commit](https://github.com/CorsixTH/CorsixTH/commit/426dd039b32456142d3ea76992b9cf7d6dfdaa5c#diff-af0b4fc49dace13a57ed7623269991de) original tried to address the old Issue 480.

It has a logic bug that partialy accounts for the above. In TH, if the room is buildable (ie discovered) but doesn't exist in the hospital then you should have the option to send to research still if research exists, currently its disabled in this case.

In the original it has a bug, in that an unoccupied research room will allow patients to 'go there', but when you start, and remove the researcher (sack them), it will disable the button again.

The other issue is also with the content of the faxes. Emergency does it correctly:
https://github.com/CorsixTH/CorsixTH/blob/f16a51011a611c1a66ba91bea8fa640739905ccf/CorsixTH/Lua/hospital.lua#L1232-L1240

[...]

Issue #1337, opened 2018-01-21
Newly built room causes confusion


wmwmw:
Patients are walking in a loop looking for suitable room.
This happens when you build a new room of type, that is overloaded, then most of the patients go to queue to the newly built room. After they arrive, they decide to walk to another room, this repeats for quite some time, until the situation stabilizes, causing some of them to die.

[Save](https://drive.google.com/open?id=16x3q1sdj4eZ0CNXlplPQVVOM66_Y37-M)
You can see such group in the bottom part, near operating theater and toilets.

**CorsixTH Version:** *0.61*

**Operating System:** *Windows 10*

**Theme Hospital Version:** *Origin*

LeDocteurr:
Suggested solution:

- compute the percentage of patients for each room of the same type based on the max queue setting
- use those percentages to split the waiting patients in between rooms

For example, for 3 rooms with 4, 4 and 2 max queue set the percentage for each room will be:
4 + 4 + 2 = 10
4 / 10 = 0.4
4 / 10 = 0.4
2 / 10 = 0.2
For 100 waiting patients, we get:
100 * 0.4 = 40 will go to room 1;
100 * 0.4 = 40 will go to room 2;
100 * 0.2 = 20 will go to room 3;

mugmuggy:
My analysis leads me to the 'build_callbacks' called via Humanoid:notifyNewRoom, cause a goto room action for all humanoids with the callbacks still registered. This steals all the patients with them as they will all get a createEnterAction occuring to the newly built room.

The individual rooms then when dealing with the last patient, call room:tryToFindNearbyPatients() which will then start stealing them from other rooms and this is slightly flawed as well as given in #441 results in disproportionate queues which then go through the cycle again and what appears to be again and again until things balance out.

mugmuggy:
So I think ignoring the callback, as that doesn't appear to be as much of an issue. Room:tryToFindNearbyPatients calling tryMovePatient evaluates the queue depth, distance via getUsageScore which attempts to serve the purpose of the ghost comment above.

The save attached to the issue points to an area of the map where the queue depth is set to a value of 2, however that really only takes into account the actually queued patients with a penalty applied when queue:isFull.

https://github.com/CorsixTH/CorsixTH/blob/1dd16721273d65ac81053748d8ef7ef7eccc196a/CorsixTH/Lua/room.lua#L605-L619

[...]

PR #1484, opened 2018-10-12
[RDY] Research fax options updating

mugmuggy:
When looking at the autospy discovery issue I raised, I also noted inconsistencies with seek treatment room fax as mentioned in #453. Some of these are documented, some are differences from the original (not necessarily desired though), and there are probably more better ways to perform the same functions.

Patients seeking a treatment room fax doesn't 'exist' for the 'need_to_employ' message when room exists but not the required staff - #469 
Patients seeking a treatment - seek the room when unstaffed - related to above (this change compromises on the diagnostic rooms) and patients will not attempt to queue until a staff member exists in the hospital
Research button disabled when the treatment room is discovered when it should remain enabled (research and researcher exist)
Interrupted seeks - don't remove message
Fax updates dynamically depending on the state

Just created for some more discussion

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/fax.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/fax.lua
@@ -166,6 +166,7 @@ function UIFax:choice(choice_number)
         owner:goHome("over_priced", owner.disease.id)
       end
     elseif choice == "research" then
+      owner:unregisterCallbacks()
       owner:setMood("idea", "activate")
       owner:setNextAction(SeekRoomAction("research"))
     end
--- a/CorsixTH/Lua/dialogs/message.lua
+++ b/CorsixTH/Lua/dialogs/message.lua
@@ -144,8 +144,10 @@ function UIMessage:removeMessage(choice_number)
       self.owner.message = nil
       self.owner.message_callback = nil
     end
-    self:onClose(false)
-    self.onClose = nil
+    if self.onClose then
+      self:onClose(false)
+      self.onClose = nil
+    end
     self:close()
   end
 end
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -300,6 +300,7 @@ function Humanoid:Humanoid(...)
 
   self.build_callbacks  = {--[[set]]}
   self.remove_callbacks = {--[[set]]}
+  self.staff_callbacks = {--[[set]]}
 end
 
 -- Save game compatibility
@@ -337,11 +338,15 @@ function Humanoid:afterLoad(old, new)
   if old < 83 and new >= 83 and self.humanoid_class == "Chewbacca Patient" then
     self.die_anims.extra_east = 1682
   end
+  if old < 133 then
+    self.staff_callbacks = {}
+  end
 
   for _, action in pairs(self.action_queue) do
     -- Sometimes actions not actual instances of HumanoidAction
     HumanoidAction.afterLoad(action, old, new)
   end
+
   Entity.afterLoad(self, old, new)
 end
 
@@ -819,6 +824,12 @@ function Humanoid:notifyNewRoom(room)
   end
 end
 
+function Humanoid:notifyOfStaff(staff)
+  for callback, _ in pairs(self.staff_callbacks) do
+    callback(staff)
+  end
+end
+
 -- Registers a new remove callback for this humanoid.
 --!param callback (function) The callback to call when a room has been removed.
 function Humanoid:registerRoomRemoveCallback(callback)
@@ -841,6 +852,29 @@ function Humanoid:unregisterRoomRemoveCallback(callback)
   end
 end
 
+
+-- Registers a new staff callback for this humanoid.
+--!param callback (function) The callback to call when a staff member has been hired or fired
+function Humanoid:registerStaffCallback(callback)
+  if self.staff_callbacks and not self.staff_callbacks[callback] then
+    self.staff_callbacks[callback] = true
+  else
+    self.world:gameLog("Warning: Trying to re-add staff callback (" .. tostring(callback) .. ") for humanoid (" .. tostring(self) .. ").")
+  end
+end
+
+-- Unregisters a remove callback for this humanoid.
+--!param callback (function) The callback to remove.
+function Humanoid:unregisterStaffCallback(callback)
+
+  if self.staff_callbacks and self.staff_callbacks[callback] then
+    self.staff_callbacks[callback] = nil
+  else
+    self.world:gameLog("Warning: Trying to remove nonexistant staff callback (" .. tostring(callback) .. ") from humanoid (" .. tostring(self) .. ").")
+  end
+end
+
+
 -- Function called when a humanoid is sent away from the hospital to prevent
 -- further actions taken as a result of a callback
 function Humanoid:unregisterCallbacks()
@@ -851,6 +885,10 @@ function Humanoid:unregisterCallbacks()
   -- Remove callbacks for removed rooms
   for cb, _ in pairs(self.remove_callbacks) do
     self:unregisterRoomRemoveCallback(cb)
+  end
+  -- Remove callbacks for removed rooms
+  for cb, _ in pairs(self.staff_callbacks) do
+    self:unregisterStaffCallback(cb)
   end
   -- Remove any message related to the humanoid.
   if self.message_callback then
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -300,7 +300,7 @@ function Humanoid:Humanoid(...)
 
   self.build_callbacks  = {--[[set]]}
   self.remove_callbacks = {--[[set]]}
-  self.staff_callbacks = {--[[set]]}
+  self.staff_change_callbacks = {--[[set]]}
 end
 
 -- Save game compatibility
@@ -338,8 +338,8 @@ function Humanoid:afterLoad(old, new)
   if old < 83 and new >= 83 and self.humanoid_class == "Chewbacca Patient" then
     self.die_anims.extra_east = 1682
   end
-  if old < 133 then
-    self.staff_callbacks = {}
+  if old < 134 and new >= 134 then
+    self.staff_change_callbacks = {}
   end
 
   for _, action in pairs(self.action_queue) do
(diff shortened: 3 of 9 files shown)


CorsixTH/Lua/dialogs/fullscreen/fax.lua
```diff
@@ -166,6 +166,7 @@ function UIFax:choice(choice_number)
         owner:goHome("over_priced", owner.disease.id)
       end
     elseif choice == "research" then
+      owner:unregisterCallbacks()
       owner:setMood("idea", "activate")
       owner:setNextAction(SeekRoomAction("research"))
     end
```
mugmuggy: This change I believe fixes the last issue in #133. Believe in this issue the callback may still be registered so when the edited room exists again the callback was executed requeuing the seek to the room. Yet to attempt to verify

CorsixTH/Lua/dialogs/message.lua
```diff
@@ -144,8 +144,10 @@ function UIMessage:removeMessage(choice_number)
       self.owner.message = nil
       self.owner.message_callback = nil
     end
-    self:onClose(false)
-    self.onClose = nil
+    if self.onClose then
+      self:onClose(false)
+      self.onClose = nil
+    end
     self:close()
```
mugmuggy: Calling the bottom_panel:removeMessage calls UIMessage:removeMessage twice and the onClose is nil.

CorsixTH/Lua/entities/patient.lua
```diff
@@ -1074,17 +1074,33 @@ function Patient:updateMessage(choice)
...
+              Doctor       = _S.staff_title.doctor,
+              Surgeon      = _S.staff_title.surgeon,
+              Psychiatrist = _S.staff_title.psychiatrist,
+          }
+          output_text = strings.need_to_employ:format(staffclass_to_string[next(req.staff)])
+        else
+          enabled = false
         end
+        self.message[3].text = output_text
+      else
```
mugmuggy: Recalculating message content, only issue here is I need to validate if getRoomNameAndRequiredStaffName will return a ward message, if checkDiseaseRequirements returns just it, if op theatre exists, I just can't recall seeing a fax with 'ward' and nurse on it. The staffclass_to_string was just taken from some other code

CorsixTH/Lua/hospital.lua
```diff
@@ -1764,6 +1767,10 @@ end
 --!param staff (Staff) Staff member to remove.
 function Hospital:removeStaff(staff)
   RemoveByValue(self.staff, staff)
+  -- update all messages for waiting patients
+  for _, patient in pairs(self.patients) do
+    patient:notifyOfStaff(nil)
+  end
 end
```
mugmuggy: The callbacks for hiring and firing.

CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -48,7 +48,7 @@ function SeekRoomAction:setDiagnosisRoom(room)
   return self
 end
 
-local function action_seek_room_find_room(action, humanoid)
+local action_seek_room_find_room = permanent"action_seek_room_find_room"( function(action, humanoid)
   local room_type = action.room_type
```
mugmuggy: Believe I had to do this for the callbacks to persist correctly

CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -128,22 +128,26 @@ local function action_seek_room_no_treatment_room_found(room_type, humanoid)
...
+    if #req.rooms > 0 then
+      if next(req.staff) then
+        output_text = strings.need_to_build_and_employ:format(room_name, staff_name)
+      else
+        output_text = strings.need_to_build:format(room_name)
+      end
     else
-      output_text = strings.need_to_build_and_employ:format(room_name, staff_name)
+      output_text = strings.need_to_employ:format(staff_name)
     end
```
mugmuggy: Like Patient:updateMessage except we have a room type to get the staff string for this time.

CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -246,115 +253,162 @@ local function action_seek_room_start(action, humanoid)
...
+    local room = action_seek_room_find_room(action, humanoid)
+    -- if we have the room but not the staff we shouldn't seek out the room either
+    if room then
+      if humanoid.message then
+        TheApp.ui.bottom_panel:removeMessage(humanoid)
+      end
+      action_seek_room_goto_room(room, humanoid, action.diagnosis_room)
+      return
+    end
+  end
```
mugmuggy: Restructuring the code a bit here as one of the 'differences' between corsix and th. Seek diagnosis or research rooms, and remove the message if it exists (this cleans up interrupted seeks)

CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -246,115 +253,162 @@ local function action_seek_room_start(action, humanoid)
...
```
mugmuggy: Similar logic for treatment as diagnosis, just checking if we alos have outstanding treatment requirements

CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -246,115 +253,162 @@ local function action_seek_room_start(action, humanoid)
...
-        end
-      end -- End of remove_callback function
-      action.remove_callback = remove_callback
-      humanoid:registerRoomRemoveCallback(remove_callback)
+    local remove_callback = --[[persistable:action_seek_room_remove_callback]] function(rm)
+      humanoid:updateMessage("research")
+    end -- End of remove_callback function
+    action.remove_callback = remove_callback
+    humanoid:registerRoomRemoveCallback(remove_callback)
+
```
mugmuggy: If the room exists and it crashes or edited the room or deleted the room - we also want to update the message so removed the research check.

CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -246,115 +253,162 @@ local function action_seek_room_start(action, humanoid)
...
-            for i = 1, #humanoid.available_diagnosis_rooms do
-              if humanoid.available_diagnosis_rooms[i].id == rm.room_info.id then
-                found = true
-              end
+          humanoid:unregisterStaffCallback(staff_callback)
+          action_seek_room_goto_room(room, humanoid, action.diagnosis_room)
+        end
+      end
+    end -- End of staff_callback function
+    action.staff_callback = staff_callback
```
mugmuggy: Callback for when a staff member is hired/fired to update the message will also enable the Research button on the fax, not just the change in message

TheCycoONE:
@mugmuggy what is the state of this PR? Ready to check?

mugmuggy:
I can recall testing each of the individual fax changes, by removing/adding rooms and hiring/firing staff. There are a few minor things, like a commented out assert that should probably just be removed, and the other concern I had at the time was around translated strings which I didn't switch languages on to check either.

And I included the unregister callback warnings, which makes for ugly game logs when executed which I think is the reason I haven't marked as RFC/RDY as yet. Getting a review on the current implementation and I'll rebase, fix app version, the commented assert and run it through some tests again.

mugmuggy:
I had a play around with the original to see what it did. If you have an op theatre and no ward, and a patient is diagnosed with say Spare Ribs. The fax says, _You cannot cure this disease at the moment._. Whereas, with this change it will Ward and Nurse (evn though you have a nurse), so its a little confused in the current state so I'll try and address that.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -300,6 +300,7 @@ function Humanoid:Humanoid(...)
 
   self.build_callbacks  = {--[[set]]}
   self.remove_callbacks = {--[[set]]}
+  self.staff_callbacks = {--[[set]]}
```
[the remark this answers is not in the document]
mugmuggy: The only difference between a hire/fire is the room checking is technically not needed after updating the message if you have sacked anyone as the message will be updated to reflect the state. Could rename to hospital_staff_change_callbacks.

CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -246,115 +260,162 @@ local function action_seek_room_start(action, humanoid)
...
+    humanoid:registerStaffCallback(staff_callback)
+
+    build_callback = --[[persistable:action_seek_room_build_callback]] function(rm)
```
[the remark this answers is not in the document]
mugmuggy: remove_callbacks does this for updating a message. Once a room exists and the patient is on way to the room (via the build_callback if required), destroying the room will start a new SeekRoomAction and the reallocation of the callbacks will occur.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -841,6 +852,29 @@ function Humanoid:unregisterRoomRemoveCallback(callback)
   end
 end
 
+
+-- Registers a new staff callback for this humanoid.
+--!param callback (function) The callback to call when a staff member has been hired or fired
+function Humanoid:registerStaffCallback(callback)
+  if self.staff_callbacks and not self.staff_callbacks[callback] then
+    self.staff_callbacks[callback] = true
```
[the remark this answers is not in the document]
mugmuggy: I just copied the existing callback related functions that do the same. To me in theory, it looks like we only should ever have one call back at a time registered for each type, but I haven't checked that specifically but the individual action callbacks are unregistered, so its when each callback is satisfied they should unregister themselves, but then you only 'seek one room at a time' - ward and operating theatre being special for treatment. Difference here is that for diagnosis rooms you might have a list of possible diagnosis rooms to try, and you'd get another callback for another SeekRoomAction but I'd have to check that too.

Maybe it was envisioned as having a list of actions that were toggled for activity for the life of the patient rather than recreating them.

Though I can see in issue #1337 for example, where a callback for a GPs office exists, when GPs offices already exist. In that particular issue, many patients get redirected to the newly built room, as the build callbacks exist, whereas room:tryToFindNearbyPatients should probably be called instead or the callback not exist at all.

TheCycoONE:
Is there much more to do on this mugmuggy? It seems better than now already, I'd like to put it in the release.

mugmuggy:
I updated based on some feedback and found another edge case that trainee doctors would cause the fax not to update. I now have them updating the messages when any special skill is added, however its only effective on surgeon skills because you need two.

PR #1486, opened 2018-10-17
[RFC] Added basic hotkey support and scroll speed and zoom speed options in Option menu.

leiget:
I plan on expanding the hotkeys to include all in-game panels and being able to change them in the options menu.


CorsixTH/Lua/app.lua
```diff
@@ -318,7 +328,7 @@ function App:init()
     if error_message then
       self.ui:addWindow(UIInformation(self.ui, error_message, true))
     end
-  end
+  end
```
TheCycoONE: Undo this whitespace change

CorsixTH/Lua/app.lua
```diff
@@ -836,6 +846,8 @@ function App:fixConfig()
         self.config[key] = 1
       end
     end
+
```
TheCycoONE: and this one

CorsixTH/Lua/config_finder.lua
```diff
@@ -18,6 +18,8 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
+------------------------------------------------------CONFIG------------------------------------------------------
```
TheCycoONE: Undo this change. The stylized section headers are only used within the config file.

CorsixTH/Lua/config_finder.lua
```diff
@@ -134,14 +139,17 @@ local config_defaults = {
   shift_scroll_speed = 4,
   new_graphics_folder = nil,
   use_new_graphics = false,
-  check_for_updates = true
+  check_for_updates = true,
 }
+
```
TheCycoONE: Ok to add a newline here, but just one

CorsixTH/Lua/config_finder.lua
```diff
@@ -476,6 +485,9 @@ audio_music = nil -- [[X:\ThemeHospital\Music]]
 'scroll_speed = ' .. tostring(config_values.scroll_speed) .. '\n' ..
 'shift_scroll_speed = ' .. tostring(config_values.shift_scroll_speed) .. '\n' .. [=[
 
+
```
TheCycoONE: Undo this change. Not consistent with other dividers. If you want to make another commit or pr making these section headers consistent it would be appreciated but not necessary.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -248,7 +266,7 @@ end
 function UIOptions:selectResolution(number)
   local res = self.available_resolutions[number]
 
-  local callback = --[[persistable:options_resolution_callback]] function(width, height)
+  local callback = function(width, height)
```
TheCycoONE: Did you verify this function was never persisted?

Either way this change belongs in its own commit.

CorsixTH/Lua/game_ui.lua
```diff
@@ -96,6 +98,15 @@ function GameUI:GameUI(app, local_hospital, map_editor)
   self.shake_screen_intensity = 0
 
   self.announcer = Announcer(app)
+  
+  
+  -- Set the scrolling keys.
+  self.scroll_keys = {
+    [tostring(app.hotkeys["scroll_up"])]   = {x =   0, y = -10},
```
TheCycoONE: I don't mind if you remove the extra spaces, or line up the table like the old scroll_keys.

CorsixTH/CorsixTH.lua
```diff
@@ -128,7 +128,8 @@ strict_declare_global "TheApp"
 TheApp = App()
 TheApp:setCommandLine(
   "--bitmap-dir=" ..base_dir.. "Bitmap",
-  "--config-file=" .. corsixth.require("config_finder"),
+  "--config-file=" .. select(1 , corsixth.require("config_finder") ),
+  "--hotkeys-file=" .. select(3 , corsixth.require("config_finder") ),
```
TheCycoONE: Should be removed.

TheCycoONE:
Thank you @leiget 

As mentioned above, some changes are requested - mostly related to formatting, a few more significant. I like what you are doing here and think it will improve the game. I hope you continue with this.

leiget:
Okay, thanks. I'll try to add code consistent with the current formatting.

TheCycoONE:
Yeah, the afterLoad stuff is wrong. Basically in all cases now you want to remove all keymaps and set them all on load as well as initialization.

For each UI object with keymaps make a function to set keys and call it both at the end of afterLoad and in the constructor. Clear any mappings.

CorsixTH/Lua/game_ui.lua
```diff
@@ -102,36 +103,36 @@ function GameUI:GameUI(app, local_hospital, map_editor)
...
+  self:addKeyHandler(self.app.hotkeys["ingame_poopLog"], self.app.world, self.app.world.dumpGameLog)
+  self:addKeyHandler(self.app.hotkeys["ingame_poopStrings"], self.app, self.app.dumpStrings)
+  self:addKeyHandler(self.app.hotkeys["ingame_toggleAnnouncements"], self, self.togglePlayAnnouncements)
+  self:addKeyHandler(self.app.hotkeys["ingame_toggleSounds"], self, self.togglePlaySounds)
+  self:addKeyHandler(self.app.hotkeys["ingame_toggleMusic"], self, self.togglePlayMusic)
   -- scroll to map position
   for i = 0, 9 do
     -- set camera view
-    self:addKeyHandler({"alt", tostring(i)}, self, self.setMapRecallPosition, i)
+    self:addKeyHandler({self.app.hotkeys["ingame_storePosition"], tostring(i)}, self, self.setMapRecallPosition, i)
```
mugmuggy: I haven't yet pulled this down to give it a try to visualise the change, but these image_storePosition/recallPosition are a list alt and ctrl modifiers for, 0 through 9. The change as it stands can only alter the default modifiers and leaving the numerical string as the 2nd element. I don't have an alternate suggestion.

TheCycoONE: Fixing it's as easy as making a separate handler option for each. Should be done since the rest of the options expect / allow a table, and these ones can only be strings.

CorsixTH/Lua/ui.lua
```diff
@@ -244,15 +247,14 @@ end
...
-  self:addKeyHandler({"shift", "f10"}, self, self.resetApp)
-  self:addKeyHandler({"ctrl", "f10"}, self, self.toggleCaptureMouse)
+  self:addKeyHandler(self.app.hotkeys["global_cancel"], self, self.closeWindow)
+  self:addKeyHandler(self.app.hotkeys["global_stop_movie"], self, self.stopMovie)
+  self:addKeyHandler(self.app.hotkeys["global_screenshot"], self, self.makeScreenshot)
+  self:addKeyHandler(self.app.hotkeys["global_fullscreen_toggle"], self, self.toggleFullscreen)
+  self:addKeyHandler(self.app.hotkeys["global_fullscreen_toggle_alt"], self, self.toggleFullscreen)
+  self:addKeyHandler(self.app.hotkeys["global_exitApp"], self, self.exitApplication)
+  self:addKeyHandler(self.app.hotkeys["global_resetApp"], self, self.resetApp)
+  self:addKeyHandler(self.app.hotkeys["global_CaptureMouse"], self, self.toggleCaptureMouse)
```
TheCycoONE: CaptureMouse is inconsistently cased compared to the rest of the options.

CorsixTH/Lua/utility.lua
```diff
@@ -312,3 +312,68 @@ function hasBit(value, bit)
   local p = 2 ^ bit
   return value % (p + p) >= p
 end
+
+-- Function to convert table to string. Pretty much just used in "config_finder.lua".
+function Table2String(table, recursive)
```
TheCycoONE: Can't you modify print_table to suit your uses instead of writing a whole new implementation?

leiget:
Okay, I'm still figuring this stuff out. What you are saying is that you want me to:

- Change all the "addKeyHandler" keys back to how they were before.
- Have them modified according to the hotkey.txt file when loading a save game and when initializing the app.
- Have them modified by making a function to do so within every object with "addKeyHandler" is created.

Is that right?

CorsixTH/Lua/ui.lua
leiget: Got that.

CorsixTH/Lua/utility.lua
leiget: I can try but Table2String returns a string while print_table prints directly to the terminal.
Table2String also outputs a table like this: `{"shift","q"}`
While print_table outputs like this:
`1 shift`
`2 q`
 It seems that having to change print_table would change it's original purpose a little too much.

Would making Table2String a local function inside config_finder.lua work for you? That's the only place it's used right now. I'll also change it to "table_toString" to fit the naming method, as I should of before.

CorsixTH/Lua/game_ui.lua
mugmuggy: Yes ingame_storePosition_0 etc is the solution, but if moving from the defaults to something else and a choice is made to rebind a subset. Say 0 to 4 is rebound to shift, QWERT, but 5-9 is not included and use the default.

Do we just leave the defaults for 6-0 or unbind them. I'm probably over thinking it and leaving defaults is fine and having them potentially disjoint is fine. As you could just want to remap a subset like the 6-0 to QWERT under the 1-5 row and then there is no need to explicitly set all of them.

mugmuggy:
Change back all the addKeyHandlers in afterLoad functions

Create a new 'handler' for afterLoad functions to remove all the old key bindings, then also register your new addKeyHandlers. This will be guarded by the oldversion < newversion check and and you change the SAVEGAME_VERSION see my previous PR #1314 just this isn't unregistering the old keyHandlers This covers your last 2 points. Old saves remove and register the new hotkeys binding, and new games, ignore the afterload change.

TheCycoONE:
Close @mugmuggy, but unfortunately in this case it can't be guarded because the keys could change. Just run the handler every time in afterLoad.

Ideally we wouldn't save them at all, but that can be future work.

leiget:
Okay, I get it.

So I'll need to bump SAVEGAME_VERSION to, say, 131, and then in an afterLoad class function I'll do basically something like:
```if old < 131 then
addKeyHandler(hotkeys["ingame_ingame_panel_bankManager"])
```
Although I think that examples wrong?

TheCycoONE:
Skip `if old` in this case, and also remove the old binding.

```lua
function setKeyHandlers(ui)
  ui:addKeyHandler(hotkeys[...], callback)
  ...
end

function UIPlaceObjects:afterLoad(old, new)
   ...
   setKeyHandlers(self)
end
```

And in ui.lua
```
UI:afterLoad(old, new)
  self.key_handlers = {}
  UI:setupGlobalKeyHandlers()
  Window.afterLoad(self, old, new)
end
```

leiget:
I'll do that. I have to go to work now :cry: but I'll get to work on it tomorrow. Thanks for the help.

TheCycoONE:
If you like later tonight I can write a patch file

mugmuggy:
Thanks @TheCycoONE I was coming back to fix up some of my comments. Question though as the key configs will be centralised to the file is it worthwhile passing the table still to add/removeKeyHandler rather than just the 'key' for the app.hotkeys lookup.
```lua
self:addKeyHandler(self.app.hotkeys["global_fullscreen_toggle_alt"], self, self.toggleFullscreen)
```
to
```lua
self:addKeyHandler("global_fullscreen_toggle_alt", self, self.toggleFullscreen)
```
and have the keyHandler routines deal with the lookup?

TheCycoONE:
Commit that patches your afterLoad handling; doesn't address @mugmuggy's comment about just using names and doing the mapping in ui.lua (which I think is a good idea)

https://github.com/TheCycoONE/CorsixTH/commit/c78e59c23aa173aebb1f0b98e395ef9c3af9ac1d

leiget:
Yep, you can. I'm new to git: is it called a diff patch?

leiget:
I merged it already.

TheCycoONE:
They were removed intentionally. How afterLoad works is that when you load a save game, it calls the afterLoad functions (you can follow the path from app, up through the various objects) to convert the save to the current format. Since the current format doesn't use any of the old keypress handlers, we do not want to add them to save games, no matter how old. I explicitly drop them all on all save games in ui.lua.

leiget:
My current issue is that I'm trying to add a setKeyHandlers() function to be used at the end of World:World(), but the ui hasn't been setup yet, so that means I can't use addKeyHandler() or removeKeyHandler().

TheCycoONE:
Why can't it go in setUI where I put all the others?

leiget:
I'm confused. Are hotkeys saved when a level is saved, and then they are loaded when a save game is loaded?

mugmuggy:
Yes state is persisted in the save. Hence unregister all handlers and reload the latest key mappings is required, otherwise the old mappings will work and not the new ones. Which is what your comment regarding https://github.com/CorsixTH/CorsixTH/pull/1486/commits/db9ad369d8affe396aef885517564ed88fc9e9fc might have been talking about here

TheCycoONE:
@leiget The commit I sent you removes all of the hotkeys when loading the save - prior to that the behavior was to keep them. Anything open during the save is not 'constructed' again, but loaded with all the variables as is at the time of save - which is why we have to run afterLoad to fix things.

The key line from my patch is: https://github.com/CorsixTH/CorsixTH/blob/22be2fbe4e27615e420d0e3f6022e444cd01c8af/CorsixTH/Lua/ui.lua#L846

TheCycoONE:
Please remove your backup folder from the commit.

CorsixTH/Lua/game_ui.lua
TheCycoONE: Yeah, I think that's fine.

TheCycoONE:
Coming along!

To summarize the outstanding issues:
[ ] I'd like to see the change mugmuggy mentioned in https://github.com/CorsixTH/CorsixTH/pull/1486#issuecomment-434100279 - the hotkey lookup moved to one spot in the addKeyHandler function, instead of everywhere.

[ ] Some white space and formatting changes to pass travis tests and conform to our code style; most of which I commented on from the start.

[ ] Change the storePosition / restorePosition to use a separate entry for each position rather than only storing the modifier to combine with 0-9.

[ ] Investigate code reuse between print_table and table_toString

[ ] tests?

mugmuggy:
App:afterLoad prevents it running again afterLoads running again

https://github.com/CorsixTH/CorsixTH/blob/e14d4ffa993a1d3fa71b6bca293d77602f388c74/CorsixTH/Lua/app.lua#L1419-L1423

leiget:
Any idea what I could do about this?

mugmuggy:
Probably as we're just talking UI elements, call, ui:afterLoad within that block. Most should just be registering keyHandlers which you need to do. This would then avoid the map/world afterLoads running. @TheCycoONE might have a different opinion as he is more familiar with persistence

leiget:
> that return should be removed.

I've tested it with a couple of saves with three different key combos and it works as far as I can tell.

leiget:
As far as  code reuse between print_table and table_toString, I'm not sure exactly what you'd like me to do. Do you not want too many global functions?
As said, print prints and toString converts table to string to be used now in menu items and the hotkey file.

Would something like this work?
https://gist.github.com/leiget/a6455105501a908ab8e94a60f3857a89

TheCycoONE:
@leiget No, that's not what I had in mind at all. My point was that both are functions to recurse through tables and convert them to human readable strings. Currently the implementations differ but I don't see much functional difference.

leiget:
Okay, how about this? (It's not complete, yet)
```lua
function table_toString(obj, forMenu, printout, max_level, level)
  assert(type(obj) == "table", " Tried to use " .. tostring(obj) .. " with table_toString().")
  pt_reflist[#pt_reflist + 1] = obj

  forMenu = forMenu or false
  printout = printout or false
  max_level = max_level or 0
  level = level or 0

  local temp_keyType
  local spacer = ""
  local result = ""
  
  if printout then
    spacer = spacer .. " "
  else
    result = "{"
  end

  for k, v in pairs (obj) do
    if printout then
      print(spacer .. tostring(k), v)
    end

    temp_keyType = type(k)

    if temp_keyType == "table" then
      result = result .. tostring(v)
      v = k
    elseif temp_keyType == "string" or "number" then
      result = result .. "\"" .. tostring(v) .. "\""
    elseif temp_keyType == "nil" then
      result = result .. "nil"
    end

    if type(v) == "table" and (not max_level or max_level > level) then
      local found_ref = false
      for _, ref in ipairs(pt_reflist) do
        if ref == v then
          found_ref = true
        end
      end

      if printout then
        if found_ref then
          print(spacer .. " " .. "<reference loop>")
        else
          print_table(v, max_level, level + 1)
        end
      end
    end
    
    result = result .. ","
  end
  
  if result ~= "" then
	  result = result:sub(1, result:len() - 1)
  end

  result = result .. "}"
  pt_reflist[#pt_reflist] = nil

  return result
end
```

TheCycoONE:
The forMenu code is rather specialized and should be a separate function, not part of utility.lua but simply local to menu.lua

e.g.
```lua
local function hotkeyToString(opt)
  if type(opt) == string then
    return string.upper(opt)
  end

  local result = ""
  for _, v in ipairs(opt) do
    result = result .. tostring(v) .. "+"
  end

  if result == "" then
    return ""
  end

  return result:sub(1, result:len() - 1)
end
```

I don't actually see any other use of table_toString in your code. print_table itself is only used in one place, where it is dumped to the console / gamelog in an error situation, so the exact output is open to change.

What I was actually looking for was using whatever table serialization function you needed to get the result you wanted, and either changing `print_table` to just call it, or remove `print_table` altogether; but as is the `print_table` serialization handles more cases than your's, in particular it has two safeguards against table cycles, whereas yours would lock up.

Additionally your type == '...' tostring() code is somewhat redundant (and tostring(nil) == 'nil' as well, so that special case doesn't need to be there).

However if you have no further need with forMenu moved out, maybe we can leave table_toString or whatever it ends up being, out.

leiget:
Also, would it be a problem to add "UIScrollSpeed", "UIShiftScrollSpeed", and "UIZoomSpeed" to ".luacheckrc"?

TheCycoONE:
@leiget No, I don't like passing in `for_X` to a general purpose function.

I think you'd be better suited by an `array_join(tbl, separator)` method anyway.

Table key order is undefined so iterating with pairs is unsuitable for generating a config file.

Yes your new classes need to be added to the globals list in luacheckrc

leiget:
It would probably be best if you wrote a small patch for this specific issue so we can have exactly what you want.

CorsixTH/Lua/config_finder.lua
```diff
@@ -493,4 +502,317 @@ audio_music = nil -- [[X:\ThemeHospital\Music]]
...
+-- Lines starting with two dashes (like this one) are ignored.
+-- Text settings should have their values between double square braces, e.g.
+-- Number settings should not have anything around their value,
+--  setting = [[value]]
+-- e.g. setting = 42
+
+----------------------------------------Global Keys----------------------------------------
+-- These are global keys to be used at anytime while the application is open.
+--]=] .. '\n' ..
+'global_confirm = [[' .. tostring(hotkeys_defaults.global_confirm) .. ']]' .. '\n' ..
```
TheCycoONE: All of these should support hotkeys with modifiers. I can put that in my patch if you want.

leiget: Oh, yeah. If you'd like, you can.

TheCycoONE: It ended up being bigger than I anticipated. A simple array join isn't suitable for the config and I'm ending up writing a general purpose serializer based loosely on the old print_table.

Maybe a few days due to my availability.

leiget: No problem. Take your time. I'll work on testing and a couple other things in the meantime.

TheCycoONE:
https://github.com/TheCycoONE/CorsixTH/commit/894588733054494358b4be89c6ef0bc9466c6a79

leiget:
I've merged but it says that loadstring is undeclared.

TheCycoONE:
Sorry, version compatibility issue: 
https://www.lua.org/manual/5.2/manual.html#8.2

I'll have to add a wrapper function that calls load on 5.2+ if there isn't already one.

TheCycoONE:
Try adding:
```
if not loadstring then
  function loadstring(str)
    load(str)
  end
end

**Comment by leiget** [comment]
Okay.

**Comment by leiget** [comment]
Soory, not sure where you mean that I should put that. You mean like this? (This is config_finder.lua around line 810)

```lua
for k, str_val in pairs(hotkeys_values) do
  if not loadstring then
    function loadstring(str)
      load(str)
    end
  end

  local status, lua_val = pcall(loadstring('return ' .. str_val))
  if status then
    hotkeys_values[k] = lua_val
  end
end
```

Regardless of how I do it, it says loadstring is undeclared, even if it's something simple like this:

```lua
if not loadstring then
end
```

TheCycoONE:
Ok I'll make the polyfill when I'm on a computer. Probably do a version check instead. And it belongs near the top of utility.lua

There should be other examples.

TheCycoONE:
Yes, like loadsting_envcall if you want to try.

leiget:
Sure thing. I see what you're talking about. Should be simple enough.

CorsixTH/Lua/strings.lua
```diff
@@ -294,7 +294,7 @@ end
 function Strings:_loadPrivate(language, env, ...)
   local chunk = self.language_to_chunk[language:lower()]
   if not chunk then -- If selected language could not be found, try to revert to English
-    print_table(self.language_to_chunk)
+    print(serialize(self.language_to_chunk, {detect_cycles=true, pretty=true})
```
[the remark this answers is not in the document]
TheCycoONE: Yes it does.

TheCycoONE:
In retrospect loadstring_envcall should have been what was used anyway. Just switch loadstring to loadstring_envcall.

CorsixTH/Lua/strings.lua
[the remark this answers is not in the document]
TheCycoONE: You can fix it now - we'll be squishing a lot of the commits later anyway.

TheCycoONE:
It is actually table.unpack that isn't supported. You were better leaving clone but changing the implementation. E.g.

```lua
function shallow_clone(tbl)
  local res = {}
  for k, v in pairs(tbl) do
    res[k] = v
  end
  return res
end

**Comment by leiget** [comment]
Sure thing.

**Comment by leiget** [comment]
Made quite a few edits since last push. Still needs a bit of work, but I uploaded it so you can see what's going on.

**Comment by leiget** [comment]
I'm trying to find out how to get "check_trailing_whitespaces.py" (run locally) to show exactly what lines I need to edit to conform to the TravisCI build. Know how? Also, I'm not sure what the problem with the TravisCI build is when it says something like this:

```
CorsixTH/Lua/app.lua:960:23: (W113) accessing undefined variable serialize
```

As I've already put "serialize" in ".luacheckrc".

**Comment by mugmuggy** [comment]
I normally do a check (well I moved to vs2017 and now have the hook to check), but as I use notepad++, I just search the same reg expressions [ \t][\r\n].

As for .luacheck, its not part of the PR (yet).

**Comment by TheCycoONE** [comment]
On a related note, the default scroll keys should stay the arrows

**Comment by TheCycoONE** [comment]
git itself can find those line problems. `git diff master --check`

**Comment by leiget** [comment]
> On a related note, the default scroll keys should stay the arrows

Okay.



> git itself can find those line problems. `git diff master --check`

Thanks.

File: CorsixTH/Lua/languages/english.lua [hunk 14]
```diff
@@ -450,6 +464,124 @@ tooltip.folders_window = {
...
+  ingame_recallPosition_5 = "5",
+  ingame_recallPosition_6 = "6",
+  ingame_recallPosition_7 = "7",
+  ingame_recallPosition_8 = "8",
+  ingame_recallPosition_9 = "9",
+  ingame_recallPosition_0 = "10",
+}
+
+tooltip.hotkey_window = {
+  button_accept = "Accept hotkey assingment above and save them to the disk",
```
Alberth289346: typo fix:
```suggestion
  button_accept = "Accept hotkey assignment above and save them to the disk",
```

CorsixTH/Lua/languages/english.lua
```diff
@@ -450,6 +464,124 @@ tooltip.folders_window = {
...
+  button_cancel = "Cancel the assingment and go back to the options menu",
```
Alberth289346:
```suggestion
  button_cancel = "Cancel the assignment and go back to the options menu",
```

CorsixTH/Lua/languages/brazilian_portuguese.lua
```diff
@@ -89,33 +89,33 @@ tooltip.research_policy.research_progress = "Progresso para a próxima descobert
 menu["player_count"] = "QUANTIDADE DE JOGADORES"
 
 menu_file = {
-  load =    " CARREGAR (SHIFT+L)   ",
-  save =    " SALVAR (SHIFT+S)   ",
-  restart = " REINICIAR (SHIFT+R)",
-  quit =    " SAIR (SHIFT+Q)   ",
+  load =    "  (%1%) CARREGAR  ",
```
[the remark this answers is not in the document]
leiget: Yes, to be the same order as the english table.

CorsixTH/Lua/ui.lua
```diff
@@ -354,20 +359,100 @@ end
...
+    end
+  end
+
+  if temp_keys ~= nil then
+    local has_enterOrPlus
+    local temp_keys_copy = {}
+
+    if type(temp_keys) == "table" then
+      temp_keys_copy = shallow_clone(temp_keys)
+    elseif type(temp_keys) == "string" then
```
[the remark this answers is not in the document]
leiget: Like this?

```lua
if type(temp_keys) == "table" then
      temp_keys_copy = shallow_clone(temp_keys)
    elseif type(temp_keys) == "string" then
      temp_keys_copy[#temp_keys_copy+1] = temp_keys
    end
```

CorsixTH/Lua/game_ui.lua
```diff
@@ -91,13 +91,46 @@ function GameUI:GameUI(app, local_hospital, map_editor)
...
+  -- This is the long version of the shift speed key.
+  -- i.e. if the "ingame_scroll_shift" key is "ctrl", then it will give us
+  --  "left ctrl" and "right ctrl" for reference against the rawchar in
+  --  "onKeyDown()" and "onKeyUp()"
+  self.shift_scroll_key_long = {}
+  self.shift_scroll_speed_pressed = false
+  local temp_table = {}
+  local shift_scroll_key_index = 1
+  if type(self.app.hotkeys["ingame_scroll_shift"]) == "string" then
+    temp_table = shallow_clone({self.app.hotkeys["ingame_scroll_shift"]})
```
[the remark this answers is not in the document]
leiget: Do you want me to make it the same as my response to your "game_ui.lua" comment below?

```lua
if type(self.app.hotkeys["ingame_scroll_shift"]) == "string" then
    temp_table[#temp_table+1] = self.app.hotkeys["ingame_scroll_shift"]
```

leiget:
Does anyone know how to make my fork have a successful travis-ci build? I'm completely confused as to how it works.

CorsixTH/Lua/ui.lua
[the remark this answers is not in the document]
TheCycoONE: `temp_keys_copy = { temp_keys }` no?

TheCycoONE:
There is a list of globals in .luacheckrc, add these functions and classes there and commit.

You said earlier you had, but I don't see the change in your PR.

CorsixTH/Lua/game_ui.lua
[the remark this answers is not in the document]
TheCycoONE: In that case you don't need clone, just if type is string, wrap with table. The original won't be affected.

CorsixTH/Lua/ui.lua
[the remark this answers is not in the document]
leiget: But won't that just be a reference, and when the first key is removed on line 409:
```lua
local key = table.remove(temp_keys, #temp_keys):lower()
```
Won't ```temp_keys_copy``` also lose its first element, as it's only a reference to ```temp_keys```?

CorsixTH/Lua/game_ui.lua
```diff
@@ -100,7 +100,7 @@ function GameUI:GameUI(app, local_hospital, map_editor)
   local temp_table = {}
   local shift_scroll_key_index = 1
   if type(self.app.hotkeys["ingame_scroll_shift"]) == "string" then
-    temp_table = shallow_clone({self.app.hotkeys["ingame_scroll_shift"]})
+    temp_table[#temp_table+1] = self.app.hotkeys["ingame_scroll_shift"]
```
TheCycoONE: the index is always 1, no reason to take the length. You could also assign the table here, like you do below for lifetime symmetry. e.g. temp_table = {self.app.hotkeys["ingame_scroll_shift"]}

CorsixTH/Lua/ui.lua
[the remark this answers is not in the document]
TheCycoONE: You need it for the table case; in the string case temp_keys_copy wouldn't reference temp_keys, it references a new table that contains that string.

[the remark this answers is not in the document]
leiget: Okay, I thought it would just be a reference and not a new table.

CorsixTH/Lua/game_ui.lua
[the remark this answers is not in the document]
leiget: Okay, I changed it.

CorsixTH/Lua/game_ui.lua
leiget: Okay, changed it to 
```lua
if type(self.app.hotkeys["ingame_scroll_shift"]) == "string" then
    temp_table = {self.app.hotkeys["ingame_scroll_shift"]}
```

leiget:
Whenever the numlock is off and I try to assign a keypad key, it doesn't seem to work. Is there a problem with keypad key input when the numlock is off? I'm trying to find out if there's something wrong with my code or not.

mugmuggy:
When numlock is on, modifier numlockactive is true. UI:onKeyDown I think is where you need to take a look.

TheCycoONE:
Is there anything else pending for this PR? I am ok with leaving better numpad handling for s follow up.

leiget:
Besides the keypad thing, the only things I can think of are:

- Disable the "global_exitApp" hotkey while assigning hotkeys? (Alt-F4 still closes, even without adding "global_exitApp" in ui.lua)
- Have modifier keys for other languages. Ex: "STRG" for German's "CTRL".

TheCycoONE:
The giant function with the table renamed several times should be split up. I can help with that if you are done with it.

leiget:
It'd probably be more expedient, since you seem to know what you want.

TheCycoONE:
Doesn't work with luajit 
`ui.lua:469 attempt to call field 'unpack' (a nil value)`

leiget:
I changed those lines in addKeyHandler() but forgot to change them in removeKeyHandler:

```lua
if self.app.hotkeys[keys] ~= nil then
  if type(self.app.hotkeys[keys]) == "table" then
    temp_keys = shallow_clone(self.app.hotkeys[keys])
  elseif type(self.app.hotkeys[keys]) == "string" then
    temp_keys = shallow_clone({self.app.hotkeys[keys]})
  end
else
```

Sorry.

TheCycoONE:
Sorry about the delay:

https://github.com/TheCycoONE/CorsixTH/tree/hotkey_refactor.

leiget:
Cool! So do you want me to merge this with my fork?

TheCycoONE:
@leiget the commit before the one you merged on the same branch had the pack fix for luajit.

TheCycoONE:
I also noticed in my testing that using modifiers doesn't work with scroll keys (e.g. ctrl+w) - we need to fix that or block them from being set.

leiget:
Do you mean if the player tries to set scroll up in the hotkey window as "CTRL+W", it won't work? You don't mean the "Speed Shift" modifer won't work, do you?

TheCycoONE:
Yes I mean the basic scroll without the speed shift.

leiget:
Okay, I'll write it to where it will get rid of any modifier, as it's not really nessecary and it might conflict with the speed shift modifier.

leiget:
Appveyor failed. Was it something I did? It says "C:/projects/corsixth/vcpkg/downloads/temp/bzip2-1.0.6.tar.gz" doesn't have the expected hash.

mugmuggy:
Appveyor cache appears to have been invalidated, and the vcpkg version we are pulling refers to a now invalid bzip2 domain for the package. Happened on my local vcpkg builds too a while back but I sat on raising an issue in case the domain came back - and time slips by.

Probably just need a separate issue to bump the vcpkg commit like #1415 ( I know a github mirror has been setup for bzip2).

CorsixTH/Lua/dialogs/menu.lua
```diff
@@ -546,10 +556,23 @@ end
...
-  menu:appendItem(_S.menu_file.load, function() self.ui:addWindow(UILoadGame(self.ui, "game")) end)
-    :appendItem(_S.menu_file.save, function() self.ui:addWindow(UISaveGame(self.ui)) end)
-    :appendItem(_S.menu_file.restart, function() app:restart() end)
-    :appendItem(_S.menu_file.quit, function() self.ui:quit() end)
+  local hotkeys = app.hotkeys
+
+  -- A function to set the menu hotkey strings, whether a string or table.
+  -- Usage: hotkey_value_label(hotkey name as string)
+  -- Ex: hotkey_value_label("ingame_loadMenu")
+  local function hotkey_value_label(hotkey_name)
```
TheCycoONE: here is the other one

CorsixTH/Lua/utility.lua
```diff
@@ -18,6 +18,12 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
+if _G._VERSION == "Lua 5.2" or _G._VERSION == "Lua 5.3" then
```
TheCycoONE: I think we don't use this function anymore, or shouldn't.

scripts/check_trailing_whitespaces.py
```diff
@@ -51,4 +51,4 @@ def has_trailing_whitespace(path):
         print(path)
     sys.exit(1)
 
-sys.exit(0)
+sys.exit(0)
```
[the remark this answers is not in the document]
TheCycoONE: Still need the newline here.

CorsixTH/Lua/config_finder.lua
```diff
@@ -493,4 +502,313 @@ audio_music = nil -- [[X:\ThemeHospital\Music]]
...
+]=]
+  fi = io.open(hotkeys_filename, "w")
+  if fi then
+    fi:write(string_01 .. string_02 .. string_03)
+  fi:close()
+  end
+end
+
+for k, str_val in pairs(hotkeys_values) do
+  local status, lua_val = pcall(loadstring('return ' .. str_val))
```
TheCycoONE: switch with loadstring_envcall

CorsixTH/Luatest/spec/utility_spec.lua
```diff
@@ -0,0 +1,109 @@
...
+  [[everywhere]],
+  42,
+  false
+}]=], res)
+  end)
+
+  it("serializes a simple key value table", function()
+    local test_tbl = {foo='bar', [true]='candy', bar=42}
+    local res = serialize(test_tbl)
+    local res_tbl = loadstring("return " .. res)()
```
TheCycoONE: switch with loadstring_envcall

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -96,37 +96,49 @@ function UIBottomPanel:UIBottomPanel(ui)
...
+  -- New bottom panel hotkeys for the first 4 buttons.
+  ui:addKeyHandler("ingame_panel_buildRoom", self, self.dialogBuildRoom)    -- Build room.
+  ui:addKeyHandler("ingame_panel_furnishCorridor", self, self.dialogFurnishCorridor)    -- Furnish corridor.
+  ui:addKeyHandler("ingame_panel_editRoom", self, self.editRoom)    -- Edit room.
+  ui:addKeyHandler("ingame_panel_hireStaff", self, self.dialogHireStaff)    -- Hire staff.
 
   -- "old" keyboard shortcuts for some of the fullscreen windows
-  ui:addKeyHandler("T", buttons[2], buttons[2].handleClick, "left") -- T for town map
-  ui:addKeyHandler("R", buttons[4], buttons[4].handleClick, "left") -- R for research
+  ui:addKeyHandler("ingame_panel_map_alt", buttons[2], buttons[2].handleClick, "left") -- T for town map
```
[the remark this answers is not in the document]
leiget: Changed.

CorsixTH/Lua/dialogs/menu.lua
```diff
@@ -529,9 +529,19 @@ end
 --!param app Application.
 function UIMenuBar:makeMapeditorMenu(app)
   local menu = UIMenu()
-  menu:appendItem(_S.menu_file.load, function() self.ui:addWindow(UILoadMap(self.ui, "map")) end)
-    :appendItem(_S.menu_file.save, function() self.ui:addWindow(UISaveMap(self.ui)) end)
-    :appendItem(_S.menu_file.quit, function() self.ui:quit() end)
+  local hotkeys = app.hotkeys
+
+  local function hotkey_value_label(hotkey_name)
```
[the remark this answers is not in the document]
leiget: Fixed.

CorsixTH/Lua/dialogs/menu.lua
leiget: Okay.

CorsixTH/Lua/utility.lua
leiget: It's still used here in "config_finder.h", at line 808:

https://github.com/leiget/CorsixTH/blob/9cdf151af7a5e4e0d995a0cdac1802486efc4d1b/CorsixTH/Lua/config_finder.lua#L807

If we should use a different function, what one would it be?

leiget: Nevermind, I just saw the second comment after this. Changed.

CorsixTH/Lua/config_finder.lua
leiget: Changed.

CorsixTH/Luatest/spec/utility_spec.lua
leiget: Changed.

leiget:
I added the fullscreen/close statements so that if the player wants to move from one dialog to the next they can do so without having to close the previous one, making it more streamlined. I should have discussed it with you before changing it.

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -565,34 +577,74 @@ end
...
 function UIBottomPanel:dialogBuildRoom()
   if self.world.user_actions_allowed then
-    local dlg = UIBuildRoom(self.ui)
-    self.ui:setEditRoom(false)
-    self.ui:addWindow(dlg)
-    self.ui:tutorialStep(3, 1, 2)
+    local w = self.ui:getWindow(UIBuildRoom)
+    local fullscreen = self.ui:getWindow(UIFullscreen)
+
+    if w then
```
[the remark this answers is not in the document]
leiget: The problem is that each "w" is a different dialog class. i.e. "UIBuildRoom" in "dialogBuildRoom()" and "UIHireStaff" in "dialogHireStaff()". Also, there are different tutorial steps in a few of these dialogs that are different.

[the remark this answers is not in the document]
leiget: Since "w" and "fullscreen" variables are used so often, maybe I should define them in file scope and just change them when I need them, to keep from having to create and destroy these variables over and over through the course of a level, as the player might open these dialogs hundreds of times. Would that be enough?

leiget:
If a python file doesn't have a newline at the end, does it have a little arrow pointing at a "no" symbol?

TheCycoONE:
> 
> 
> If a python file doesn't have a newline at the end, does it have a little arrow pointing at a "no" symbol?

If any *text* file is missing the training newline git (and github) will flag it. In github it shows up as the little arrow with the "no symbol." Many Unix tools expect a trailing newline on text files.

leiget:
Alright, then, the newline seems to be there. Anything else that needs to be done?

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -183,6 +201,10 @@ function UIOptions:UIOptions(ui, mode)
   self:addBevelPanel(165, customise_y_pos, BTN_WIDTH, 30, col_bg):setLabel(_S.options_window.folder)
     :makeButton(0, 0, BTN_WIDTH, 30, nil, self.buttonFolder):setTooltip(_S.tooltip.options_window.folder_button)
 
+  -- "Hotkeys" button
+  local hotkey_y_pos = self:_getOptionYPos() + 10
```
TheCycoONE: Does it look better centered?

leiget: How about this?
temp_01

CorsixTH/Lua/config_finder.lua
```diff
@@ -493,4 +502,313 @@ audio_music = nil -- [[X:\ThemeHospital\Music]]
...
+'ingame_recallPosition_3 = ' .. hotkeys_values.ingame_recallPosition_3 .. '\n' ..
+'ingame_recallPosition_4 = ' .. hotkeys_values.ingame_recallPosition_4 .. '\n' ..
+'ingame_recallPosition_5 = ' .. hotkeys_values.ingame_recallPosition_5 .. '\n' ..
+'ingame_recallPosition_6 = ' .. hotkeys_values.ingame_recallPosition_6 .. '\n' ..
+'ingame_recallPosition_7 = ' .. hotkeys_values.ingame_recallPosition_7 .. '\n' ..
+'ingame_recallPosition_8 = ' .. hotkeys_values.ingame_recallPosition_8 .. '\n' ..
+'ingame_recallPosition_9 = ' .. hotkeys_values.ingame_recallPosition_9 .. '\n' ..
+'ingame_recallPosition_0 = ' .. hotkeys_values.ingame_recallPosition_0 .. '\n' .. [=[
+
+----------------------------------------Toggle Various----------------------------------------
```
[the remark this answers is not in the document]
TheCycoONE: Yes

TheCycoONE:
I think this is ok. I'll play test before merging.

How do you feel about a rebase to clean up history. Normally we insist on it, but this is kinda big and collaborative. We made an exception previously for SDL2.

TheCycoONE:
And I'll have to look at the busted failures

leiget:
I don't know much about git. I've looked up rebasing and merging and I can't make heads or tails of it. Just do what you think is best.

TheCycoONE:
Ok. This is not the PR I would suggest cutting your teeth on so I'll handle it now. Thank you for your persistence!

TheCycoONE:
Nothing for you at the moment, unless you could rebase on the latest master to see if it fixes appveyor.

I need to test on Windows, then merge.

leiget:
I might as well try to rebase it.

Would I need to do this:
```
git rebase master
```

TheCycoONE:
Neither probably. If you have a remote called upstream for CorsixTH/CorsixTH then git fetch upstream; git rebase upstream/master

leiget:
I'm trying to rebase, but I can't seem to get it. I tried branching and other things, but the rebase doesn't seem to be doing anything.

TheCycoONE:
That last commit very much broke things... ok, just leave it to me.

TheCycoONE:
Probably not what you want to hear, but I found a bug:
1. Change some hotkeys
2. Click accept
3. Go back to hotkeys, change some more
4. Click cancel

Instead of going back to the previously configured hotkeys, it goes back to the original set from when the game was loaded.

Want to fix this after the merge?

leiget:
I go to bed in 3 hours, so I'll try to fix it tomorrow. If I'm not able to finish it by 1 pm EST, then I'll fix it after the merge.

leiget:
Fixed it. And I wanted my fork to be back to working order so I pushed it again. Sorry if that makes things messier.

The changes I made to fix the bug you mentioned are in "/CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua" It's just a `hotkeys_backedUp = false` in the accept buttons function, and I added a `shallow_clone` in the cancel buttons function so it would copy the backup table instead of just pointing to it, just in case not copying it causes problems.

TheCycoONE:
Ok, I had started the rebase/merge. I'll pull your changes into it.

TheCycoONE:
Verified your fix

TheCycoONE:
Merged

TheCycoONE:
Leiget, now that this is merged, please consider updating the wiki, specifically: https://github.com/CorsixTH/CorsixTH/wiki/Hotkeys and the relevant programming ideas section.

Also due to the way you used git, your private repo / master branch is going to be a mess. If you have upstream as a remote:
```bash
git fetch upstream
git reset --hard upstream/master
```

leiget:
Okay, I'll have to get on that tomorrow. As far as the git upstream thing you mention, I'm not sure I have upstream as remote.

TheCycoONE:
https://help.github.com/articles/configuring-a-remote-for-a-fork/

leiget:
I edited the "Hotkeys" and "Programming Ideas" wiki pages.

PR #1491, opened 2018-12-23
Dockerfile added for building corsixth

lastcosmonaut:
Tell me if it's useful to you. Thanks!

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,0 +1,8 @@
+# Dockerfile to build CorsixTH
+FROM ubuntu:latest
+
+WORKDIR /sourcecode
+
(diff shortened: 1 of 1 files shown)


TheCycoONE:
I'm not personally experienced enough with Docker to review this, but I don't see the harm in including it as long as it's maintained, and the documentation on how to use it is added either to this file (comments) or the wiki.

@emorrp1 are you able to look at this?

Should we specify a particular version of Ubuntu? `latest` makes me nervous.

emorrp1:
Yes, I'll look into this at the weekend. I suspect "latest" is probably fine.

romualdr:
well, FYI this still works in 2020 - I mostly used the `apt-get` commands to get the dependencies in my own Dockerfile to create a build environment to play with the source code, and it still works like a charm with ubuntu `latest`

TheCycoONE:
If it's useful to you, I'll merge it.

PR #1492, opened 2018-12-29
[RDY] Update vcpkg hash

TheCycoONE:
Update to latest vcpkg, should fix build errors.

Bump ffmpeg to 4.1
Bump sdl to 2.0.9


mugmuggy:
I got the same on an x86-windows triplet here. I made a small tweak in ffmpegs configure to avcodec_suggest to add ole32 but see you added the upstream defect.

TheCycoONE:
I doubt that hardware acceleration is implemented for smacker or would be very useful for the low resolution videos in CorsixTH.

PR #1501, opened 2019-03-12
[RDY] Speed scroll not handled by GameUI:setupGlobalKeyHandlers 

mugmuggy:
Speed Scrolling modifier was only included in the GameUI constructor, and not subsequenctly called from setupGlobalKeyHandlers, thus a save from a prior version will break (nil passed into pairs as shift_scroll_key_long not initialised) and if it is changed on a new one, it will remain using the old keys.

### DIFF
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -91,6 +91,25 @@ function GameUI:GameUI(app, local_hospital, map_editor)
 
   self.speed_up_key_pressed = false
 
+  -- The currently specified intensity value for earthquakes. To abstract
+  -- the effect from the implementation this value is a number between 0
+  -- and 1.
+  self.shake_screen_intensity = 0
+
+  self.announcer = Announcer(app)
+end
+
+function GameUI:setupGlobalKeyHandlers()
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/game_ui.lua
```diff
@@ -91,6 +91,25 @@ function GameUI:GameUI(app, local_hospital, map_editor)
...
+
+function GameUI:setupGlobalKeyHandlers()
+  UI.setupGlobalKeyHandlers(self)
+
+  -- Set the scrolling keys.
+  self.scroll_keys = {
+     [tostring(self.app.hotkeys["ingame_scroll_up"])] = {x = 0, y = -10},
+     [tostring(self.app.hotkeys["ingame_scroll_down"])] = {x = 0, y = 10},
+     [tostring(self.app.hotkeys["ingame_scroll_left"])] = {x = -10, y = 0},
+     [tostring(self.app.hotkeys["ingame_scroll_right"])]	= {x = 10, y = 0},
```
Alberth289346: Bit much whitespace before the `=`

TheCycoONE: That block was preexisting, just moved. I'm ok either way.

Issue #1490, opened 2018-11-30
Can place more than maximum staff in ward


mugmuggy:
#### Describe the issue
More than the maximum staff can be placed in the ward. This is more of a cheat and could potentially apply to other rooms. I've tested the ward remains functional.

I was able to get 3 surgeons gowned in the operating theatre with the same process (didn't test room is functional) as the first lot of steps to reproduce.

#### Steps to Reproduce
1. Build ward, just one desk and place *initial* nurse

2. Bring up the staff dialog and click sack, but don't confirm/cancel.

3. Bring up available staff dialog, and place as many nurses in the ward before the *initial* nurse leaves.

4. Before *initial* nurse leaves, confirm the firing.

5. Placing another nurse in the room, only one will leave.

Or:

1. Build ward, just one desk and place *initial* nurse, wait for nurse to use desk.

2. Quickly place as many other nurses in the room, before the *initial* nurse finishes at the desk

3. *Initial* nurse should leave, leaving the rest in the room. No need to fire them.

#### Save Game
N/A But I have restored the autosaves and there has been no issue there either.

#### Expected Behavior
Excess nurses should leave

#### System Information
**CorsixTH Version:** https://github.com/CorsixTH/CorsixTH/commit/c97629ed0bef3d027a4fc2f518e394a0c0151bab

#### Gamelog.txt
N/A

Issue #1489, opened 2018-11-30
Crash when placing Surgeons in operating theatre


mugmuggy:
#### Describe the issue
OperatingTheatre:getStaffMember attempts to compare number with nil as documented in https://github.com/CorsixTH/CorsixTH/issues/1462#issuecomment-441411345

There is 2 issues here
staff.fired is checked though in the if statement but doesn't check staff_member.fired, which is the cause of the issue here.

The other issue is 'minor' as the wrong staff member can be selected.
The else path sets staff to staff_member regardless of the 'fired' state, this can lead to staff members not being evaluated (as the if will bypass the checks).

A fix:

The if should be changed to staff_member.fired instead of staff. 

The else path should check if not staff_member.fired.

I haven't tested the above against a working game, just working around it, when reproducing the issue. Other rooms, research and ward also have a similar fired check in the instances of getStaffMember, where a fired staff member could be returned.

#### Steps to Reproduce
1. Place surgeon in op theatre

2. Sack surgeon

3. Place 3 other surgeons in op theatre before sacked one leaves

#### Save Game
N/A

#### Expected Behavior
One of the other surgeons leaves

#### System Information
**CorsixTH Version:** https://github.com/CorsixTH/CorsixTH/commit/c97629ed0bef3d027a4fc2f518e394a0c0151bab

#### Gamelog.txt
See https://github.com/CorsixTH/CorsixTH/issues/1462#issuecomment-441411345

Issue #757, opened 2015-06-20
Patient with serious radiation at research department


bobcanthelpyou:
The shower to heal the serious radiation is already researched. Possible i send a patient to the research department and when the patient arrived it was already researched (but i could be wrong). 

I only could send the patient home. If the shower is already researched, the patient should not access or leave the research department and visit the shower. Or do something different instead of waiting for the death.

http://filebin.ca/25vd3snDvgiG/stuck.sav

CorsixTH trunk (master branch)

MarkL1961:
do you have the save from before he went to research?
Was he already on the way there when you researched and built the room?

bobcanthelpyou:
I have a savegame before that issue occurs but i can't reproduce it.

If serious radiation is already researched then a doctor also take attention to the patient

Maybe something happens to the doctor with the attention, and then nobody else take care of the patient. Sometimes a doctor starts wandering around and don't leave the room he is in, like here #505 maybe that happened here too. But it doesn't happen here.

But you could force the issue, send the patient to the research department and pick up the doctor with the attention and put the doctor to the staff room. Now the patient is waiting until he is dead ...

Is there also a 'todo' list for doctors as there are for handymen? described here: https://github.com/CorsixTH/CorsixTH/issues/388#issuecomment-113790765 

If a doctor lost his attention somebody else should take care of the patient.

http://filebin.ca/25zcuEEapkSl/beforestuck.sav

MarkL1961:
This is as I thought, you get the fax, send him to the research and just after this the shower is researched before he gets there.  I think to get the error it would be bad timing/edge case.  It might even be the researcher with the task of putting him in autopsy leaves the room and the task is not reassigned as it should be

bobcanthelpyou:
It doesn't matter if the shower is researched before, he will also be put inside the autopsy machine.

The problem is the researcher that leaves the room (for a break, fired, ..) and the task is not reassigned to a different researcher.

PR #1505, opened 2019-03-22
[RDY] Account for leaving staff in getStaffMember for rooms with multiple staff

mugmuggy:
#757, #1489, #1490
All three of these issues are related to the one function not accounting for staff leaving the room correctly.


TheCycoONE:
Reads right to me. There may be room for consolidation here later.

PR #1506, opened 2019-03-23
[RDY] getStaffMember update

mugmuggy:
Didn't see push failed with the previous PR. hasLeavingAction required as staff member current action might not be is_leaving at the time of the getStaffMember call.


TheCycoONE:
Shouldn't there be a 3rd one?

mugmuggy:
Research already had it

PR #1508, opened 2019-03-25
[RDY] Queue dialog patient reordering fixes

### LINKED ISSUE #368: 0.40 beta2 - cannot reorder queue when an handyman is repairing a room
It happens also when the handyman is enqueued: when I dragged a patient in the queue the handyman pops in the first place.
Here is a savegame: https://www.dropbox.com/s/j1whztx62cxn2tc/Handyman.sav?dl=0
Look at the queue of the Operating Theater, you can't place the infectious patient at the first place until the handyman leaves the room.

mugmuggy:
#368 indicates some problems with queue reordering. I wasn't able to get an instance of the handyman at the front preventing a move of an epidemic patient, but I did manage to get a queue that happened to have 2 epidemic patients in it, and the order would not change.

I also got the other more easily observed issue of the handyman in room popping into and out of the queue when re-ordering. The changes here, insert re-ordered patients relative to the start of the 'reportedHumanoids' of the queue rather than the entire queue addresses both issues. And this appears to fix both issues.

### DIFF
--- a/CorsixTH/Lua/dialogs/queue_dialog.lua
+++ b/CorsixTH/Lua/dialogs/queue_dialog.lua
@@ -175,17 +175,16 @@ function UIQueue:onMouseUp(button, x, y)
       self.dragged = nil
       return
     end
-
     if x > 170 and x < 210 and y > 25 and y < 105 then -- Inside door bounding box
-      queue:move(index, 1) -- move to front
+      queue:movePatient(index, 'front') -- move to front
     elseif x > 542 and x < 585 and y > 50 and y < 105 then -- Inside exit sign bounding box
-      queue:move(index, num_patients) -- move to back
+      queue:movePatient(index, 'back') -- move to back
     elseif isInsideQueueBoundingBox(x, y) then
       local dx = 1
       if num_patients ~= 1 then
         dx = math.floor(width / (num_patients - 1))
       end
-      queue:move(index, math.floor((x - 220) / dx) + 1) -- move to dropped position
+      queue:movePatient(index, math.floor((x - 220) / dx)) -- move to dropped position
       self:onMouseMove(x, y, 0, 0)
     end
 
--- a/CorsixTH/Lua/queue.lua
+++ b/CorsixTH/Lua/queue.lua
@@ -294,6 +294,26 @@ function Queue:move(index, new_index)
   end
 end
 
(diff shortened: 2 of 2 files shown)


CorsixTH/Lua/dialogs/queue_dialog.lua
```diff
@@ -176,16 +176,17 @@ function UIQueue:onMouseUp(button, x, y)
       return
     end
 
+    local queue_reported_humans = queue:size() - num_patients
```
TheCycoONE: Funny name for this variable - not clear what it represents. (I assume from the context that it's people who need to use the door that are not part of the patient queue, like staff that need to enter the room.)

mugmuggy: Yeah a little funky, in light of this I've done a refactor and I'll push that out once I tested it, as its a little more readable also. This currently has this changed to first_patient_index

CorsixTH/Lua/dialogs/queue_dialog.lua
```diff
@@ -176,16 +176,17 @@ function UIQueue:onMouseUp(button, x, y)
       return
     end
 
+    local first_patient_index = queue:size() - num_patients + 1
```
TheCycoONE: Same calculation used in 169. Maybe move this up where num_patients is defined (or into queue)

TheCycoONE:
I'm concerned this breaks with emergencies. Emergency patients are at the front of the queue ahead of newly arriving VIPs etc.

TheCycoONE:
I'm not sure that was original behavior, nor that it makes sense. Maybe take a closer look at queue.lua

TheCycoONE:
Doesn't break with emergencies, as inspectors and VIPs are in the reported size, and staff/anyone leaving are at the front.

Still suggesting to handle this from queue.lua so it's not a leaked implementation detail.

mugmuggy:
Ok, queue:move is only used by the queue dialog already anyway, so I could probably move that detail there or implement the logic in a wrapper around queue:move.

TheCycoONE:
@mugmuggy if you've already started go ahead; otherwise just move the calculation up.

queue isn't quite sure how it wants to present indexes / sizes etc. 😝

TheCycoONE:
I like it. Will need to review the proper LuaDoc style later. In JSDoc it would have been (int|string) with one parameter.

PR #1511, opened 2019-04-05
[RDY] Modernizing C++

TheCycoONE:
A number of refactors aimed at modernizing and applying current best practices to the C++ code base.

* Bump to C++ 14 which is now the default version on a number of compilers.
* Use constexpr on compile time constants and a conservative set of functions that should be evaluated at compile time.
* Replace some magic numbers with named constants
* Greatly reduce the number of macros used
* Use anonymous namespaces instead of static for controlling linkage in C++ code.


TheCycoONE:
Looks like I'll need to fiddle with this under Visual Studio.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -82,6 +82,66 @@ static inline uint32_t makeSwapRedBlue(uint8_t iOpacity, uint8_t iR, uint8_t iG,
...
+    // Maps 0-63 to 0-255
+    0x00, 0x04, 0x08, 0x0C, 0x10, 0x14, 0x18, 0x1C,
+    0x20, 0x24, 0x28, 0x2D, 0x31, 0x35, 0x39, 0x3D,
+    0x41, 0x45, 0x49, 0x4D, 0x51, 0x55, 0x59, 0x5D,
+    0x61, 0x65, 0x69, 0x6D, 0x71, 0x75, 0x79, 0x7D,
+    0x82, 0x86, 0x8A, 0x8E, 0x92, 0x96, 0x9A, 0x9E,
+    0xA2, 0xA6, 0xAA, 0xAE, 0xB2, 0xB6, 0xBA, 0xBE,
+    0xC2, 0xC6, 0xCA, 0xCE, 0xD2, 0xD7, 0xDB, 0xDF,
+    0xE3, 0xE7, 0xEB, 0xEF, 0xF3, 0xF7, 0xFB, 0xFF,
+};
```
Alberth289346: LUTs are a bit obsolete with modern processors, you may want to use the `color` function that I wrote for this:
``` cpp
static unsigned char colors[] = {
    0x00, 0x04, 0x08, 0x0C, 0x10, 0x14, 0x18, 0x1C,
    0x20, 0x24, 0x28, 0x2D, 0x31, 0x35, 0x39, 0x3D,
    0x41, 0x45, 0x49, 0x4D, 0x51, 0x55, 0x59, 0x5D,
    0x61, 0x65, 0x69, 0x6D, 0x71, 0x75, 0x79, 0x7D,
    0x82, 0x86, 0x8A, 0x8E, 0x92, 0x96, 0x9A, 0x9E,
    0xA2, 0xA6, 0xAA, 0xAE, 0xB2, 0xB6, 0xBA, 0xBE,
    0xC2, 0xC6, 0xCA, 0xCE, 0xD2, 0xD7, 0xDB, 0xDF,
    0xE3, 0xE7, 0xEB, 0xEF, 0xF3, 0xF7, 0xFB, 0xFF,
};

/*
def f(i):
    return int(i*4.048 + 0.5)
*/

static unsigned char color(int index) {
    return static_cast<int>(index * 4.048 + 0.5);
}

#include <cstdio>
#include <cstdlib>

int main() {
    for (int i = 0; i <64; i++) {
        int t = colors[i];
        int c = color(i);
        if (t == c) continue;
        printf("%d: table=%d, func=%d\n", i, t, c);
    }
    return 0;
}
```
Complete with test, doesn't print any output at my system (both with the Python `f` and the cpp `color` functions).

CorsixTH/Src/th_sound.cpp
```diff
@@ -71,11 +71,19 @@ const char* sound_archive::get_sound_name(size_t iIndex) const
...
-    | static_cast<uint32_t>(static_cast<uint8_t>(c2) <<  8) \
-    | static_cast<uint32_t>(static_cast<uint8_t>(c3) << 16) \
-    | static_cast<uint32_t>(static_cast<uint8_t>(c4) << 24) )
+constexpr uint32_t fourcc(const char c1, const char c2, const char c3, const char c4)
+{
+    return (
+        static_cast<uint32_t>(static_cast<uint8_t>(c1) <<  0)
+        | static_cast<uint32_t>(static_cast<uint8_t>(c2) <<  8)
+        | static_cast<uint32_t>(static_cast<uint8_t>(c3) << 16)
+        | static_cast<uint32_t>(static_cast<uint8_t>(c4) << 24));
```
Alberth289346: Is this defined, shifting with a larger amount than the width of the data. Perhaps simply cast directly to `uint32_t` ?

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: Sorry, being silly, the better function is of course
``` cpp
return static_cast<int>(255 * index / 63.0 + 0.5);
```

CorsixTH/Src/th_sound.cpp
TheCycoONE: Good catch!

After a deep dive in the specification, I believe this is implementation defined in C++14 (not undefined) for systems with 32bit integers. In C++11 and earlier it would be undefined, and on C++20 I believe it would be defined as long as the target's int type is > 32bit. All of that assumes that it treats int like one of the fixed width integer types (2s complement is mandated for them, but not for e.g. int, short, etc. directly)

I'll change it.

CorsixTH/Src/th_gfx_sdl.cpp
TheCycoONE: I just moved this without looking too closely at how it was used. Will consider.

TheCycoONE:
The Visual Studio issue was simply a missing header (include what you use.)

I've also added a fix for the `fourcc` issue, and fixes for a number of other errors and warnings.

CorsixTH/Src/th_gfx_sdl.cpp
TheCycoONE: Agree, this isn't even a performance sensitive path.

TheCycoONE:
git did something weird, and dated my last commit for a week ago.

TheCycoONE:
Working on cleaning up the lua 'class' binding macros which cleans up a lot of the warnings in g++ if `-Wpedantic` is set.

TheCycoONE:
There is a regression to the town map view in be953b870ab4315cf9d53e2f946a2d5110545f8b

TheCycoONE:
Rebased on latest master so that travis can pass.

TheCycoONE:
That's all the changes I plan to make in this pull. I'd like it merged shortly after 0.63 lands. I have other work based on this branch.

TheCycoONE:
Ok, ready for this to land now 😄

Issue #1512, opened 2019-04-06
Assert in call dispatcher executeCall


mugmuggy:
#### Describe the issue

Appears to have been introduced as part of the staff classes. Handyman becomes stuck. Reproduce on new game and old saves.

#### Steps to Reproduce
1. Start any hospital, place bunch of plants so they will require watering around the same time.

2. Hire 2 handymen, prepare to hire a 3rd, but don't place him

3. Wait for the 2 handymen to start actioning the tasks, place 3rd handyman.

#### Save Game
N/A
#### System Information
**CorsixTH Version:** https://github.com/CorsixTH/CorsixTH/commit/f3d204e4deb5e52ce9d57ca6297c0cd008025231
**Operating System:** *e.g. Windows 10*

**Theme Hospital Version:** *CD, GOG.com, Origin, or Demo*

#### Gamelog.txt
```
Lua\calls_dispatcher.lua:448: call to be executed is still assigned
stack traceback:
	[C]: in function 'assert'
	Lua\calls_dispatcher.lua:448: in method 'executeCall'
	Lua\entities\humanoids\staff\handyman.lua:177: in method 'assignHandymanTask'
	Lua\entities\humanoids\staff\handyman.lua:124: in method 'searchForHandymanTask'
	Lua\calls_dispatcher.lua:346: in method 'answerCall'
	Lua\humanoid_actions\answer_call.lua:33: in field '?'
	Lua\entities\humanoid.lua:531: in upvalue 'Humanoid_startAction'
	Lua\entities\humanoid.lua:623: in method 'finishAction'
	Lua\humanoid_actions\walk.lua:193: in local 'timer_function'
	Lua\entity.lua:216: in field 'tick'
	Lua\entities\humanoids\staff.lua:105: in method 'tick'
	Lua\world.lua:1042: in method 'onTick'
	Lua\app.lua:1111: in function <Lua\app.lua:1108>
	(...tail calls...)
	Lua\app.lua:1012: in function <Lua\app.lua:1007>
```

### Additional Info
*Paste any screen shots or other additional information that might help
illustrate the problem.*

mugmuggy:
Looks like Handyman:onPlaceInCorridor, calling the meander in Staff:onPlaceInCorridor and allocating the task and answers a call, then its unassigned again immediately after but not dropping the call. Probably get away with unassigning the task before call Staff.onPlaceInCorridor from a quick test

TheCycoONE:
[...]

PR #1513, opened 2019-04-06
[RDY] Handyman:unassignTask before calling Staff.onPlaceInCorridor

mugmuggy:
Addresss #1512

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff/doctor.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff/doctor.lua
@@ -262,7 +262,7 @@ function Doctor:fulfillsCriterion(criterion)
   return false
 end
 
-function Staff:adviseWrongPersonForThisRoom()
+function Doctor:adviseWrongPersonForThisRoom()
   local room = self:getRoom()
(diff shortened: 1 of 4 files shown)


TheCycoONE:
See if fixing the other two changes I made resolves it. I don't see why it would, but I don't know how it wasn't an issue with the old code either - as far as I can tell call.assigned isn't unset when placing a handyman, but task.assignedHandyman is, which should make the task come up again and get dispatched again? (Both before and after my change)

mugmuggy:
Yeah its a little confusing, the specific instances of the 'interrupt' stack just occur by themselves, you don't have to place them like the issue indicates. Likewise the call doesn't appear to interrupt correctly either.

I had 2 handymen, one that entered the room and fixed the machine, the other started to enter via another room, exited that room, then started walking to the room they had the job for. Then when the handyman in the room finished up, the one outside the call was dropped on the one entering. 

There might be some more issues around the cancellation and prioritisation of these tasks, as I have a game where there could be 3 or 4 in a corridor and there is a call for watering plant in a room, and none pick up the call for some time, with the default 0.33 priority.

PR #1515, opened 2019-04-08
[WIP] Bump dependencies to latest vcpkg

TheCycoONE:
Newer patches (same upstream versions) for:
sdl2
ffmpeg
libmodplug


TheCycoONE:
The appveyor builds are timing out after an hour...

mugmuggy:
You seen https://github.com/appveyor/ci/issues/517 I wonder if that is enough time, just to bump dependencies to update cache

TheCycoONE:
@mugmuggy I'm investigating a bug. With newer versions of vcpkg the execution never gets past the powershell script. There is plenty of time left, at the point it freezes.

https://github.com/Microsoft/vcpkg/issues/5476

mugmuggy:
Taking a more solid look I could see that. I've had no troubles running it locally unlike what that issue states, but obviously there are differences. The only thing I noticed locally, and its in your comment on that issue was that gpg-agent was running still afterwards and I hadn't seen that in previous bumps.

That led me to https://help.appveyor.com/discussions/problems/16396-build-freezes-between-commands-in-appveyoryml so it could be a factor even though it doesn't appear to belong to a process tree.

mugmuggy:
Just FYI, I explored the gpg-agent and am getting success by doing stopping the gpg-agent forceably.

TheCycoONE:
@mugmuggy could you notify upstream.

PR #1516, opened 2019-04-08
[WIP] Use vcpkg preinstalled in appveyor

TheCycoONE:
Attempt to speed up the build by using the vcpkg included with appveyor instead of checking out our own copy.


TheCycoONE:
So regardless of how I build on appveyor, using the latest vcpkg, the script gets stuck after building the libraries. Since there are no important changes we could probably get by with not bumping the libraries now.

TheCycoONE:
I'd like to keep the flexible remote name from this PR, but nothing else is relevant.

PR #1527, opened 2019-04-16
[RDY] Fix undefined behaviour

TheCycoONE:
I compiled CorsixTH with g++ with ubsan (undefined behaviour sanitizer) enabled, and it flagged a couple errors.

These fixes are quick/minimal so that they can land before the 0.63 release.

### DIFF
--- a/CorsixTH/Src/th_sound.cpp
+++ b/CorsixTH/Src/th_sound.cpp
@@ -37,12 +37,37 @@ sound_archive::~sound_archive()
     delete[] data;
 }
 
+namespace {
+
+/**
+ * Convert 4 bytes representing uint32 in little endian representation into a
+ * uint32.
+ *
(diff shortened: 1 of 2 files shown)


CorsixTH/Src/th_sound.cpp
```diff
@@ -37,12 +37,30 @@ sound_archive::~sound_archive()
     delete[] data;
 }
 
+namespace {
+
+uint32_t bytes_to_uint32_le(const uint8_t* bytes)
```
Alberth289346: Could use a little documentation I guess, although it's hardly complicated at all.

TheCycoONE:
Added the suggested documentation. xmi2mid could use larger changes, but I wanted a minimal fix for the release.

TheCycoONE:
(Mainly it's yet another place that reimplements vectors)

TheCycoONE:
Pushed documentation update

PR #1528, opened 2019-04-18
[RDY] Bump travis to Xenial

TheCycoONE:
Trying to fix broken Travis builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ addons:
       - libssl-dev
       - luarocks
       - doxygen
-      - g++-5
       - sshpass
       - ant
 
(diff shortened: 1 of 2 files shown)


mugmuggy:
This on upstream https://github.com/brunoos/luasec/issues/136

TheCycoONE:
> This on upstream [brunoos/luasec#136](https://github.com/brunoos/luasec/issues/136)

Sounds relevant, bumping to xenial.

TheCycoONE:
Bingo, just need ant because it's not included by default anymore.

wolfy1339:
Any particular reason to still install GCC 5?

TheCycoONE:
@wolfy1339 no, it's default anyway, but it doesn't hurt either right?

TheCycoONE:
Removed explicit gcc version per @wolfy1339 's recommendation.

Issue #1529, opened 2019-04-25
Hotkeys broken on place object dialog


mugmuggy:
#### Describe the issue
When placing objects, typically you can use 'Space' by default to rotate the object. The hotkeys change appears to cause this functionality to break when the place object (when editing a room) dialog is visible, ie works fine with corridor objects.

There is also another side effect, in that it also appears to break previous version saves if the place object dialog was open; in Window:Draw as hotkeyboxes appears to be nil
```lua
  if not class.is(self, UI) then -- prevent UI (sub)class from handling the hotkeyboxes too
    for _, _ in ipairs(self.hotkeyboxes) do
      -- Empty right now. Don't know if I need it. -- LEIGET
    end
  end
```

Additionally the registerKeyHandler lists 3 handlers
```lua
function UIPlaceObjects:registerKeyHandlers()
  self:addKeyHandler("global_cancel", self.cancel)
  self:addKeyHandler("global_cancel_alt", self.cancel)
  self:addKeyHandler("ingame_rotateobject", self.tryNextOrientation)
end
```
However, the global_cancel_alt (q) doesn't appear to run, global_cancel does (esc) still work.

#### Steps to Reproduce
1. Build a room

#### Save Game
See #1439 for save game where placing an object is underway at the point of save.

#### Expected Behavior
Should be able to rotate objects within a room with the hot key.

#### System Information
**CorsixTH Version:** 0.63 beta 1

**Operating System:** *e.g. Windows 10*

**Theme Hospital Version:** *CD, GOG.com, Origin, or Demo*

#### Gamelog.txt
For the Window.Draw
```
Savegame version changed from 127 (v0.62) to 134 (Trunk). The save was created using 111 (v0.60)
Error in frame handler: 
attempt to index a nil value
stack traceback:
	[C]: in for iterator 'for iterator'
	Lua\window.lua:1476: in field 'draw'
	Lua\dialogs\place_objects.lua:556: in field 'draw'
	Lua\dialogs\edit_room.lua:984: in method 'draw'
	Lua\window.lua:1484: in field 'draw'
	Lua\game_ui.lua:258: in method 'draw'
	Lua\app.lua:1130: in function <Lua\app.lua:1123>
	(...tail calls...)
	Lua\app.lua:1012: in function <Lua\app.lua:1007>
```

### Additional Info
*Paste any screen shots or other additional information that might help
illustrate the problem.*

PR #1530, opened 2019-04-28
[RDY] Fix edit room hotkeys

TheCycoONE:
self:registerKeyHandlers was calling the function on the
subclass. Explicitly register on the current class.

Fixes the major issues in #1529 
I haven't yet tested the old save.

### DIFF
--- a/CorsixTH/Lua/dialogs/confirm_dialog.lua
+++ b/CorsixTH/Lua/dialogs/confirm_dialog.lua
@@ -99,5 +99,6 @@ function UIConfirmDialog:draw(canvas, x, y)
 end
 
 function UIConfirmDialog:afterLoad(old, new)
+  Window.afterLoad(self, old, new)
   self:registerKeyHandlers()
 end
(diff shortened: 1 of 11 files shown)


mugmuggy:
That works ok.

The old save has an afterLoad with a call to self:registerKeyHandlers also, but that doesn't get called at all. It also leads to the room requiring to be completed then edited again for the hotkeys to become active.

This occurs on new games, started on 0.63 beta1 as well, if the hotkey is changed and the save reloaded, and the UI is open it is disfunctional until the UI is reopened again.

The saves issue is more a nice to have, as there is a simple workaround, complete room then edit, or cancel room and start over.

TheCycoONE:
I had changed the afterLoad as well, but was tired at the time and neglected to include it in the commit. I'll fix UIPlaceObject's afterLoad not being called as well.

TheCycoONE:
I've fixed loading saves made with the current version with new hotkeys.

TheCycoONE:
Fixed loading old saves that had dialogs open. I think this now fixes everything from #1529

TheCycoONE:
@mugmuggy could you check and merge this if ok?

mugmuggy:
All good on trunk saves and happy to leave the previous saves with the editing of room.

Issue #832, opened 2015-07-25
New GP office built: multi_use_object.lua:313: attempt to index field '?' (a nil value)


J-Shep:
Reported by @bobcanthelpyou:

> > I moved the Pharmacy room and build a new GP's Office nearby and an error occurs:

```
Will not check for updates since this is the Trunk version.
Savegame version is 102 (Trunk), originally it was 102 (Trunk)
An error has occured!
Almost anything can be the cause, but the detailed information below can help the developers find the source of the error.
Running: The timer handler.
A stack trace is included below, and the handler has been disconnected.
./Lua/humanoid_actions/multi_use_object.lua:313: attempt to index field '?' (a nil value)
stack traceback:
    ./Lua/humanoid_actions/multi_use_object.lua:313: in function '?'
    ./Lua/entities/humanoid.lua:575: in function 'Humanoid_startAction'
    ./Lua/entities/humanoid.lua:667: in function 'finishAction'
    ./Lua/humanoid_actions/walk.lua:152: in function 'timer_function'
    ./Lua/entity.lua:218: in function 'tick'
    ./Lua/world.lua:1121: in function 'onTick'
    ./Lua/app.lua:876: in function <./Lua/app.lua:873>
    (tail call): ?
    ./Lua/app.lua:777: in function <./Lua/app.lua:772>

Error in timer handler: 
./Lua/humanoid_actions/multi_use_object.lua:313: attempt to index field '?' (a nil value)
stack traceback:
    ./Lua/humanoid_actions/multi_use_object.lua:313: in function '?'
    ./Lua/entities/humanoid.lua:575: in function 'Humanoid_startAction'
    ./Lua/entities/humanoid.lua:667: in function 'finishAction'
    ./Lua/humanoid_actions/walk.lua:152: in function 'timer_function'
    ./Lua/entity.lua:218: in function 'tick'
    ./Lua/world.lua:1121: in function 'onTick'
    ./Lua/app.lua:876: in function <./Lua/app.lua:873>
    (tail call): ?
    ./Lua/app.lua:777: in function <./Lua/app.lua:772>
Warning: No event handler for timer
```

> > savegame before error: http://www.xup.in/dl,10243463/congame2.sav/
> > savegame after error: http://www.xup.in/dl,94694418/conerror.sav/ (the room with the cast remover seems problematic)

TheCycoONE:
What version is this error from?

J-Shep:
[...]

PR #1531, opened 2019-04-30
[RDY] Check for lava hole use tile in room

mugmuggy:
Addresses #832


TheCycoONE:
Ah, makes sense here.

Issue #1536, opened 2019-05-04
UIConfirmDialog produces an inconsistent dialog box size, often leading to overlapping text on frame


terrorcide:
#### Steps to Reproduce
1. Click file, quit

2. Observe prompt
![corsixth](https://user-images.githubusercontent.com/10485456/57176349-ecbfad80-6e4e-11e9-9bfc-ebdc7c08b9ce.png)

#### Expected Behavior
Retail doesn't do this
![retail](https://user-images.githubusercontent.com/10485456/57176352-f9440600-6e4e-11e9-9845-abb32b1cd4ed.png)

System Information
**CorsixTH Version: commit 27f5bb8

**Operating System: XUbuntu 18.04.2 LTS

Theme Hospital Version: GOG

TheCycoONE:
The bug being that the bottom of the prompt overlaps the border, not that we have to match the retail size.

terrorcide:
Lua/dialogs/confirm_dialog.lua comments have a clue:

  -- Rough estimate of how many rows it will be when drawn.                      
  for y = 22, h * (w / **160**) * 1.4, 11 do -- Previous value: 136                  
    self:addPanel(358, 0, y)  -- Dialog background                               
    self.height = self.height + 11                                               
    last_y = last_y + 11                                                         
  end                 

Changing this back to 136 fixes this.
![image](https://user-images.githubusercontent.com/10485456/57181841-bdca2b80-6e90-11e9-9483-b21873989b1b.png)

This appears have been added in this commit from 2012:
https://github.com/CorsixTH/CorsixTH/commit/f24fb90e6936eaa9cfa8c09ccac5d610160d8570

terrorcide:
I've opened a pull request for this. I hope I've followed convention, but let me know if I've missed anything.

terrorcide:
Interestingly, there are other confirmation boxes in the game that are broken. This fix doesn't seem to affect any of those.

e.g. sack staff one

![image](https://user-images.githubusercontent.com/10485456/57182924-69797880-6e9d-11e9-909a-773fe8aab6e7.png)

terrorcide:
By tweaking this code further (moving to 120), you can fix the sack staff dialog:

![image](https://user-images.githubusercontent.com/10485456/57183600-2c19e880-6ea7-11e9-9f38-f6b3f83530f5.png)

  self:addPanel(357, 0, 0)  -- Dialog header                                     
  local last_y = 22                                                              
[...]

PR #1537, opened 2019-05-04
Fix size of quit confirmation dialog box

terrorcide:
#1536 Fix size of quit confirmation dialog box 

#### Old
corsixth

#### New
Screenshot_2019-05-04_17-19-08


terrorcide:
Wrong version checked in

PR #1538, opened 2019-05-04
Fix quit confirmation dialog box size

terrorcide:
Proposed fix for #1536 

#### Old
corsixth

#### New
Screenshot_2019-05-04_17-19-08


terrorcide:
Not happy that this doesn't fix the sack staff dialog box, I'm closing this until we can fix both

PR #1539, opened 2019-05-05
Fix confirmation dialog box sizes

### LINKED ISSUE #1536: UIConfirmDialog produces an inconsistent dialog box size, often leading to overlapping text on frame

terrorcide:
Calculates actual size of confirmation dialog boxes and uses this to draw them (rather than estimating)

Fixes #1536 
#### Examples

image

image

### DIFF
--- a/CorsixTH/Lua/dialogs/confirm_dialog.lua
+++ b/CorsixTH/Lua/dialogs/confirm_dialog.lua
@@ -58,7 +58,7 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
   local y = topFrameHeight;
   for i=1, totalLines do
     self:addPanel(358, 0, y)  -- Dialog background
-    y = y + textRowHeight
+    y = topFrameHeight + (i * textRowHeight)
   end
 
   self:addPanel(359, 0, y)  -- Dialog footer
--- a/CorsixTH/Lua/dialogs/confirm_dialog.lua
+++ b/CorsixTH/Lua/dialogs/confirm_dialog.lua
@@ -61,9 +61,11 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
   --local totalLines = 9
 
   local y = topFrameHeight;
-  for i=1, totalLines do
+  local i=0
+  while (i < totalLines) do
     self:addPanel(358, 0, y)  -- Dialog background
 	y = y + textRowHeight
+	i = i + 1
   end
 
   self:addPanel(359, 0, y)  -- Dialog footer
--- a/CorsixTH/Lua/dialogs/confirm_dialog.lua
+++ b/CorsixTH/Lua/dialogs/confirm_dialog.lua
@@ -27,6 +27,10 @@ local UIConfirmDialog = _G["UIConfirmDialog"]
 function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
   self:Window()
 
+  local topFrameIndex = 357;
+  local topFrameHeight = 22;
+  self:addPanel(topFrameIndex, 0, 0)  -- Dialog header
+
   local app = ui.app
   self.modal_class = "information"
   self.esc_closes = true
@@ -37,26 +41,35 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
   self:setDefaultPosition(0.5, 0.5)
   self.panel_sprites = app.gfx:loadSpriteTable("QData", "Req04V", true)
   self.white_font = app.gfx:loadFont("QData", "Font01V")
-  self.text = text
   self.callback_ok = callback_ok  -- Callback function to launch if user chooses ok
   self.callback_cancel = callback_cancel -- Callback function to launch if user chooses cancel
+  self.text = text
 
-  -- Check how "high" the dialog must be
-  local w, h = self.white_font:sizeOf(text)
+  local maxLineLengthPixels = 152 -- this should come from somewhere else
 
-  self:addPanel(357, 0, 0)  -- Dialog header
-  local last_y = 22
-  -- Rough estimate of how many rows it will be when drawn.
-  for y = 22, h * (w / 160) * 1.4, 11 do -- Previous value: 136
+  -- Text is varable width, calculate it
+  local lineWidthPixels, textHeight = self.white_font:sizeOf(self.text)
+
+  -- This was previously hard coded to 11, but textHeight is actually 12, one less
+  -- But the graphics appear to be drawn to 13 - one more
+  local textRowHeight = textHeight + 1
+
+  -- Calculate how many text rows we need
+  -- Retail has a fixed size box - but existing behaviour is to calculate size of box
+  -- Leave this as it is
+  local totalLines = math.ceil(lineWidthPixels / maxLineLengthPixels)
+  --local totalLines = 9
+
+  local y = topFrameHeight;
+  for i=1, totalLines do
     self:addPanel(358, 0, y)  -- Dialog background
-    self.height = self.height + 11
-    last_y = last_y + 11
+	y = y + textRowHeight
   end
 
-  self:addPanel(359, 0, last_y)  -- Dialog footer
-  self:addPanel(360, 0, last_y + 10):makeButton(8, 10, 82, 34, 361, self.cancel)
+  self:addPanel(359, 0, y)  -- Dialog footer
+  self:addPanel(360, 0, y + 10):makeButton(8, 10, 82, 34, 361, self.cancel)
     :setTooltip(_S.tooltip.window_general.cancel):setSound"No4.wav"
-  self:addPanel(362, 90, last_y + 10):makeButton(0, 10, 82, 34, 363, self.ok)
+  self:addPanel(362, 90, y + 10):makeButton(0, 10, 82, 34, 363, self.ok)
     :setTooltip(_S.tooltip.window_general.confirm):setSound"YesX.wav"
 
   self:registerKeyHandlers()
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/confirm_dialog.lua
```diff
@@ -37,26 +41,37 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
...
   self.callback_ok = callback_ok  -- Callback function to launch if user chooses ok
   self.callback_cancel = callback_cancel -- Callback function to launch if user chooses cancel
+  self.text = text
+
+  local maxLineLengthPixels = 152 -- this should come from somewhere else
+
+  -- Text is varable width, calculate it
+  local lineWidthPixels, textHeight = self.white_font:sizeOf(self.text)
+
+  -- This was previously hard coded to 11, but textHeight is actually 12, one less
```
TheCycoONE: I'd prefer not to comment on the way things use to be. Source control can be used for that.

CorsixTH/Lua/dialogs/confirm_dialog.lua
```diff
@@ -37,26 +41,37 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
...
-  local last_y = 22
-  -- Rough estimate of how many rows it will be when drawn.
-  for y = 22, h * (w / 160) * 1.4, 11 do -- Previous value: 136
+  local y = topFrameHeight;
+  local i=0
+  while (i < totalLines) do
     self:addPanel(358, 0, y)  -- Dialog background
-    self.height = self.height + 11
-    last_y = last_y + 11
+	y = y + textRowHeight
```
TheCycoONE: Too much indent.

CorsixTH/Lua/dialogs/confirm_dialog.lua
```diff
@@ -37,26 +41,37 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
...
+  local totalLines = math.ceil(lineWidthPixels / maxLineLengthPixels)
+  --local totalLines = 9
 
-  self:addPanel(357, 0, 0)  -- Dialog header
-  local last_y = 22
-  -- Rough estimate of how many rows it will be when drawn.
-  for y = 22, h * (w / 160) * 1.4, 11 do -- Previous value: 136
+  local y = topFrameHeight;
+  local i=0
+  while (i < totalLines) do
```
TheCycoONE: Lua tends to 1 based indexing, and doesn't use brackets around conditions. `for i = 1, totalLines do` would be conventional.

CorsixTH/Lua/dialogs/confirm_dialog.lua
```diff
@@ -37,26 +41,37 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
...
+  -- But the graphics appear to be drawn to 13 - one more
+  local textRowHeight = textHeight + 1
 
-  -- Check how "high" the dialog must be
-  local w, h = self.white_font:sizeOf(text)
+  -- Calculate how many text rows we need
+  -- Retail has a fixed size box - but existing behaviour is to calculate size of box
+  -- Leave this as it is
+  local totalLines = math.ceil(lineWidthPixels / maxLineLengthPixels)
+  --local totalLines = 9
```
mugmuggy: I'd suggest removing the fixed value comment here as it probably pertains to your testing or trying to match the fixed size but the dialog gets used with multiple actions.

CorsixTH/Lua/dialogs/confirm_dialog.lua
terrorcide: Thanks for the advice. This was actually akin to my first version, but it does not pass luacheck
``
  local y = topFrameHeight;                                                      
  for i=1, totalLines do                                                         
      self:addPanel(358, 0, y)  -- Dialog background                             
      y = y + textRowHeight                                                      
  end   
``
Checking confirm_dialog.lua                       1 warning
    confirm_dialog.lua:61:7: (W213) unused loop variable i
Total: 1 warning / 0 errors in 1 file

Any ideas - or can this be overridden?

mugmuggy: You can override the warning in .luacheck.rc or just calculate 'y' from i within the loop.

CorsixTH/Lua/dialogs/confirm_dialog.lua
terrorcide: No problem - removed

CorsixTH/Lua/dialogs/confirm_dialog.lua
terrorcide: Agreed - missed that one. I think it's now fixed

CorsixTH/Lua/dialogs/confirm_dialog.lua
terrorcide: No problem - done. I left it in to comment on the approach being different between retail and here, and you can swap by adjusting this here, but I'll remove if you'd prefer it.

CorsixTH/Lua/dialogs/confirm_dialog.lua
terrorcide: Thanks for the advice. I'll calculate y from i (although it looks a little strange).

Hope this is OK, let me know if not

TheCycoONE: Oh, sorry: the fix for the luacheck warning is simply to use `_` instead of `i`.

terrorcide: Brilliant - that looks a lot nicer. I have checked that in. Thanks

CorsixTH/Lua/dialogs/confirm_dialog.lua
```diff
@@ -58,7 +58,7 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
   local y = topFrameHeight;
   for i=1, totalLines do
     self:addPanel(358, 0, y)  -- Dialog background
-    y = y + textRowHeight
+    y = topFrameHeight + (i * textRowHeight)
```
TheCycoONE: Unused variable names can be replaced with `_` and then it will ignore them. That's better than making this more complicated.

terrorcide: Agreed - have just checked in.

TheCycoONE:
For me there is a regression in the behaviour with freetype fonts.

Master:
master_exit_freetype

This Branch:
1539_exit_freetype

TheCycoONE:
I see, the panel has a fixed height but you are separating them by your calculated textHeight. Instead you have to determine how many panels would be needed to add up to textHeight and add that many.

CorsixTH/Lua/dialogs/confirm_dialog.lua
```diff
@@ -37,26 +41,30 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
...
+
+  -- Text is varable width, calculate it
+  local lineWidthPixels, textHeight = self.white_font:sizeOf(self.text)
+  local textRowHeight = textHeight + 1
 
-  -- Check how "high" the dialog must be
-  local w, h = self.white_font:sizeOf(text)
+  -- Calculate how many text rows we need
+  -- Retail has a fixed size box - but current behaviour is to calculate size of box
+  local totalLines = math.ceil(lineWidthPixels / maxLineLengthPixels)
```
TheCycoONE: Old code did this but it's totally unnecessary and probably wrong. Just pass the max width as a second argument to font:sizeOf, and it will give you the total height of the resulting textbox.

CorsixTH/Lua/dialogs/confirm_dialog.lua
```diff
@@ -37,26 +41,30 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
   self:setDefaultPosition(0.5, 0.5)
   self.panel_sprites = app.gfx:loadSpriteTable("QData", "Req04V", true)
   self.white_font = app.gfx:loadFont("QData", "Font01V")
-  self.text = text
   self.callback_ok = callback_ok  -- Callback function to launch if user chooses ok
   self.callback_cancel = callback_cancel -- Callback function to launch if user chooses cancel
+  self.text = text
+
+  local maxLineLengthPixels = 152 -- this should come from somewhere else
```
TheCycoONE: The number is 153, at least where it is used in drawWrapped. Best to use a single file level variable for both.

CorsixTH/Lua/dialogs/confirm_dialog.lua
TheCycoONE: Which would probably fix the discrepancy between different dialogs, because actual line breaking would be taken into consideration...

CorsixTH/Lua/dialogs/confirm_dialog.lua
```diff
@@ -37,26 +41,30 @@ function UIConfirmDialog:UIConfirmDialog(ui, text, callback_ok, callback_cancel)
...
+  -- Calculate how many text rows we need
+  -- Retail has a fixed size box - but current behaviour is to calculate size of box
+  local totalLines = math.ceil(lineWidthPixels / maxLineLengthPixels)
 
-  self:addPanel(357, 0, 0)  -- Dialog header
-  local last_y = 22
-  -- Rough estimate of how many rows it will be when drawn.
-  for y = 22, h * (w / 160) * 1.4, 11 do -- Previous value: 136
+  local y = topFrameHeight;
+  for _=1, totalLines do
```
TheCycoONE: And since you have to fix other things anyway, might as well add the spaces here :smile:

TheCycoONE:
Closing due to lack of response in favour of #1572

PR #1541, opened 2019-05-08
Add desktop file and AppData manifest file

tim77:
Please add AppData manifest file for modern package managers and easy installation on searching via Gnome-Software/Discover KDE.

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html


TheCycoONE:
@emorrp1 anything from your Debian packaging experience that might help here?

CorsixTH/com.github.CorsixTH.CorsixTH.appdata.xml
```diff
@@ -1,16 +1,23 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Copyright 2019 Artem Polishchuk <ego.cordatus@gmail.com> -->
 <component type="desktop">
   <id>corsix-th.desktop</id>
+  <!-- Copyright 2019 Artem Polishchuk <ego.cordatus@gmail.com> -->
   <metadata_license>CC0</metadata_license>
-  <project_license>GPL-3.0+</project_license>
```
TheCycoONE: Sorry, I misspoke, our code is just MIT. Does the AppData license need to cover dependencies like ffmpeg as well though?

TheCycoONE: Though, the random.c code is BSD-3 clause... ok. 😄

emorrp1:
> @emorrp1 anything from your Debian packaging experience that might help here?

[Debian's .desktop](https://sources.debian.org/src/corsix-th/0.62-2/debian/corsix-th.desktop/) has a couple of extra details to help searching, but it is better if you supply one:

```
[Desktop Entry]
Name=CorsixTH
Exec=corsix-th
Comment=A Theme Hospital Clone
Comment[de]=Ein Klon von Theme Hospital
Icon=corsix-th
Type=Application
Terminal=false
StartupNotify=false
Categories=Game;Simulation;
Keywords=hospital;simulation;bullfrog;
```

I haven't looked into AppStream, but again, if it's provided (/usr/share/metainfo/) then apparently it'll be automatically included and detected in debian [source](https://wiki.debian.org/AppStream/Guidelines).

TheCycoONE:
Install locations are helpful, thanks. We can make cmake install put them in the right place.

CorsixTH/com.github.CorsixTH.CorsixTH.appdata.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>corsix-th.desktop</id>
```
TheCycoONE: Isn't id suppose to be the reverse DNS name, like the name of the file? https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic and https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps sound that way.

TheCycoONE:
@tim77 you're going to change the <id> element as well?

PR #1542, opened 2019-05-11
Add desktop and appdata in cmake install

TheCycoONE:
@emorrp1 / @tim77 I think this does the right thing.

My one concern is that now cmake install adds a desktop file, but not the required icons. Do you think this is ok for 0.63?

### DIFF
--- a/CorsixTH/CMakeLists.txt
+++ b/CorsixTH/CMakeLists.txt
@@ -251,6 +251,9 @@ if(NOT USE_SOURCE_DATADIRS)
 
   if(UNIX AND NOT APPLE)
     install(FILES corsix-th.6 DESTINATION ${CMAKE_INSTALL_MANDIR}/man6)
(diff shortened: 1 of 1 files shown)


tim77:
I will try to build and test this. We can add icons as well in cmake? Something like this:

```cmake
install(FILES Original_Logo.svg DESTINATION ${CMAKE_INSTALL_DATADIR}/icons/hicolor/scalable/apps/corsix-th.svg)
```

I put this icon file manually in last build and it works.
corsixth

TheCycoONE:
Yes, I will add something today before prepping rc1

TheCycoONE:
In the interest of time I simply followed @tim77's suggestion and added the existing svg as a scalable icon. I suggest that packagers continue to extract / include their own pngs at the common sizes for environments that don't support svg; and for the next version we'll get something more upstream.

PR #1544, opened 2019-05-21
Allow overriding data path via CMake

h3xx:
If unspecified, it defaults to the original.

Should not affect builds with USE_SOURCE_DATADIRS=ON, but if specified,
will override it.


h3xx:
Whoops, something isn't quite right.

TheCycoONE:
What's the use case for this?

h3xx:
My particular use case for this is thus:

- All icons on my system should go under /usr/share/icons
- All app .desktop files should go under /usr/share/applications
- All game data should go into subfolders under /usr/share/games/ (i.e. If it's in man section 6)

Using this patch, the way to have `make install` target these paths correctly is to invoke CMake with the following options:

    -DCORSIX_TH_DATADIR=/usr/share/games/corsix-th
    -DCMAKE_INSTALL_PREFIX=/usr

Without the patch, using the above options:

- Icons are installed under /usr/share/icons ✔️
- .desktop files are installed under /usr/share/applications ✔️
- Game data is installed under /usr/share/corsix-th ✘

There doesn't really exist another workaround to make the paths right.

TheCycoONE:
Alright. I have some concerns with this on Windows and perhaps MacOS targets. Because it was internal it wasn't always used consistently and I will have to test them. For Linux/BSDs It might make sense to make the other paths configurable independent of CMAKE_INSTALL_DATADIR as well.

No changes requested at the moment, this will land after 0.63.

TheCycoONE:
Ok, lets be a little more conservative with this option for now:

Wrap the initial set in `if(UNIX AND NOT APPLE)` and in the other place make it lower precedent, so something like:

```
if(USE_SOURCE_DATADIRS)
...
elseif(MSVC)
....
elseif(APPLE)
....
else()
  if(NOT CROSIX_TH_DATADIR)
    set(CORSIX_TH_DATADIR ${CMAKE_INSTALL_FULL_DATADIR/corsix-th)
  endif()
  set(CORSIX_TH_INTERPRETER_PATH ${CORSIX_TH_DATADIR/CorsixTH.lua})
endif()
```

TheCycoONE:
That way I don't have to worry about platforms where the path is meant to be relative.

TheCycoONE:
Actually I have a more canonical fix for your use case. CMAKE_INSTALL_DATADIR should be /usr/share/games and CMAKE_INSTALL_DATAROOTDIR should be /usr/share/ and I should be using DATAROOTDIR for things like icons.

TheCycoONE:
Closed in favour of #1545

PR #1545, opened 2019-05-25
[RDY] Use DATAROOTDIR for icons, metadata on *nix

TheCycoONE:
Use DATAROOTDIR instead of DATADIR for files that belong in the root
/usr/share instead of /usr/share/**/corsix-th.

This allows putting corsix-th under /usr/share/games.


TheCycoONE:
@h3xx could you confirm this works for you

h3xx:
Yes, I was able to get it targeting the paths I described in #1544 using the following configuration:

    CMAKE_INSTALL_DATAROOTDIR=/usr/share
    CMAKE_INSTALL_DATADIR=/usr/share/games

PR #1548, opened 2019-05-29
[RDY] The return of ISO support in CorsixTH

TheCycoONE:
Fixes all the bit rot that happened in the ISO code over the years, and possibly some bugs that have always been there.

Everything is functional except movies.

Commits need to be cleaned up and squashed as appropriate.


CorsixTH/Lua/dialogs/resizables/directory_browser.lua
```diff
@@ -70,6 +71,26 @@ function InstallDirTreeNode:createNewNode(path)
...
 
+local function is_iso(name)
+  if name == nil then
+    return false
+  end
+
+  local ext = 'iso'
+  if string.sub(name:lower(), -string.len(ext)) == ext then
+    return true
+  end
```
Alberth289346: Shouldn't it always return a boolean? ie
``` lua
return string.sub(name:lower(), -string.len(ext)) == ext
```

CorsixTH/Lua/dialogs/resizables/directory_browser.lua
```diff
@@ -70,6 +71,26 @@ function InstallDirTreeNode:createNewNode(path)
   return InstallDirTreeNode(path)
 end
 
+local function is_iso(name)
```
Alberth289346: Our coding style says
``` text
Function names should use CamelCase. Global function names should begin with an upper-case character, whereas locals and method names should begin with a lower-case character.
```
See also https://github.com/CorsixTH/CorsixTH/wiki/Coding-Conventions#lua-code

CorsixTH/Lua/dialogs/resizables/directory_browser.lua
```diff
@@ -70,6 +71,26 @@ function InstallDirTreeNode:createNewNode(path)
...
+    return false
+  end
+
+  local ext = 'iso'
+  if string.sub(name:lower(), -string.len(ext)) == ext then
+    return true
+  end
+end
+
+function InstallDirTreeNode:isValidFile(name)
```
Alberth289346: We do have lua documentation conventions nowadays

CorsixTH/Lua/filesystem.lua
```diff
@@ -99,8 +99,20 @@ function FileSystem:listFiles(virtual_path, ...)
   return self.files
 end
 
+local function get_full_path(virtual_path, ...)
+  if ... then
```
Alberth289346: lol :)

Might be useful to document the function with expected arguments, as this can be pretty much anything.

Also, more function name and function documentation stuff. Likely it exists at more spots.

CorsixTH/Src/th.h
```diff
@@ -92,4 +92,40 @@ class th_string_list
...
+    res |= bytes[1];
+    res <<= 8;
+    res |= bytes[0];
+
+    return res;
+}
+
+/**
+ * Convert 2 bytes representing uint16 in little endian representation into a
+ * uint16.
```
Alberth289346: Except it returns a uint32

CorsixTH/Src/iso_fs.cpp
```diff
@@ -70,33 +75,187 @@ constexpr int max_directory_depth = 16;
...
+        if(sIdent[i] == ';')
+        {
+            iLength = i;
+            return;
+        }
+    }
 }
 
-} // namespace
+char normalise(char c) {
```
Alberth289346: This curly braces layout is different from the previous one, code style says curly braces must be on their own line.

CorsixTH/Src/iso_fs.cpp
```diff
@@ -70,33 +75,187 @@ constexpr int max_directory_depth = 16;
...
+    for (size_t i = 0; i < length; i++) {
+        ret.push_back(normalise(static_cast<char>(*p)));
+        ++p;
+    }
+    return ret;
 }
 
+std::string normalise(const char* str) {
+    std::string ret;
+    const char* p = str;
```
Alberth289346: You can use `str` instead of introducing `p`, but this is ok too.

CorsixTH/Src/iso_fs.cpp
```diff
@@ -70,33 +75,187 @@ constexpr int max_directory_depth = 16;
...
+        uint8_t filename_length = b[filename_length_offset];
+        trim_identifier_version(b + filename_offset, filename_length);
+        filename = normalise(b + filename_offset, filename_length);
+    }
+
+    uint32_t data_sector;
+    uint32_t data_length;
+    uint8_t flags;
+    std::string filename;
+};
```
Alberth289346: I guess documentation should be here, due to lack of a header fie.

CorsixTH/Src/iso_fs.cpp
```diff
@@ -70,33 +75,187 @@ constexpr int max_directory_depth = 16;
...
+        }
+        return *this;
+    }
+
+    iso_directory_iterator operator++(int) {
+        iso_directory_iterator o(*this);
+        ++(*this);
+        return o;
+    }
+private:
```
Alberth289346: Empty line before `private` would be useful I think

CorsixTH/Src/iso_fs.cpp
```diff
@@ -70,33 +75,187 @@ constexpr int max_directory_depth = 16;
...
+    uint8_t flags;
+    std::string filename;
+};
+
+class iso_directory_iterator final {
+    using iterator_category = std::input_iterator_tag;
+    using value_type = const iso_file_entry;
+    using difference_type = ptrdiff_t;
+    using pointer = const iso_file_entry*;
+    using reference = const iso_file_entry&;
```
Alberth289346: new spiffy type syntax, apparently :)

CorsixTH/Src/iso_fs.cpp
```diff
@@ -133,63 +283,34 @@ bool iso_filesystem::initialise(FILE* fRawFile)
...
-        if(std::memcmp(aBuffer + 1, "CD001\x01", 6) == 0)
-        {
-            if(aBuffer[0] == vdt_privary_volume)
-            {
-                sector_size = read_native_int<uint16_t>(aBuffer + 128);
-                find_hosp_directory(aBuffer + 156, 34, 0);
-                if(file_count == 0)
-                {
+        if(std::memcmp(aBuffer + 1, "CD001\x01", 6) == 0) {
+            if(aBuffer[0] == vdt_primary_volume) {
```
Alberth289346: At other places you change it to `if (...`, ie with a space behind the keyword. You may want to look for `if(` and a the usual others (`for(` and `while(`). Wiki code style probably also needs adapting then.

CorsixTH/Src/iso_fs.cpp
```diff
@@ -133,63 +283,34 @@ bool iso_filesystem::initialise(FILE* fRawFile)
...
-    if(c == '_') // underscore to hyphen
-        return '-';
-    else if('a' <= c && c <= 'z') // ASCII lowercase to ASCII uppercase
-        return static_cast<char>(c - 'a' + 'A');
-    else
-        return c;
-}
-
-void iso_filesystem::trim_identifier_version(const uint8_t* sIdent, uint8_t& iLength)
+bool iso_filesystem::filename_compare(const file_metadata& lhs, const file_metadata& rhs)
```
Alberth289346: `compare` is a ambiguous, eg equality checking is also comparing. Perhaps `filename_sort_compare` or `filename_orderer` ?

CorsixTH/Src/iso_fs.cpp
```diff
@@ -203,62 +324,36 @@ int iso_filesystem::find_hosp_directory(const uint8_t *pDirEnt, int iDirEntsSize
...
-        uint32_t iDataLength = read_native_int<uint32_t>(pDirEnt + file_data_length_offset);
-        uint8_t iFlags = pDirEnt[file_flags_offset];
-        uint8_t iIdentLength = pDirEnt[filename_length_offset];
-        trim_identifier_version(pDirEnt + filename_offset, iIdentLength);
-        if(iFlags & def_directory)
-        {
+    iso_directory_iterator dir_iter(pDirEnt, pDirEnt + iDirEntsSize);
+    iso_directory_iterator end_iter(pDirEnt + iDirEntsSize, pDirEnt + iDirEntsSize);
+    for (; dir_iter != end_iter; ++dir_iter) {
+        iso_file_entry ent = *dir_iter;
```
Alberth289346: Reference to avoid copying?

Alberth289346:
Very good work, my comments are just stuff that gets missing when you switch to fixing and understanding code.

CorsixTH/Src/th.h
TheCycoONE: Oops. 👍

CorsixTH/Src/iso_fs.cpp
TheCycoONE: Yes, that's the plan. After this lands I'm going to run clang-format over everything and push a unified style.

CorsixTH/Lua/dialogs/resizables/directory_browser.lua
TheCycoONE: Interesting; lower_snake_case is more common for our lua local functions, though both are used:
```
rg 'local function [a-z_]+_[a-z]+\(' -q --stats

155 matches
155 matched lines
94 files contained matches
285 files searched
```
vs camelCase
```
rg 'local function [a-z][a-z-A-Z]+\(' -q --stats

95 matches
95 matched lines
40 files contained matches
285 files searched
```
For completeness PascalCase:
```
rg 'local function [A-Z][a-z-A-Z]+\(' -q --stats

12 matches
12 matched lines
6 files contained matches
285 files searched
```

CorsixTH/Lua/dialogs/resizables/directory_browser.lua
TheCycoONE: Will add.

CorsixTH/Src/iso_fs.cpp
TheCycoONE: Yeah, I don't like mutating arguments. Can easily miss if passed by reference or not.

CorsixTH/Src/iso_fs.cpp
TheCycoONE: C++ 11 version of a typedef yes. This is the standard C++ 17 (and higher) way of declaring an iterator, since std::iterator is deprecated. Works in C++14 so why not.

CorsixTH/Src/iso_fs.cpp
TheCycoONE: hmm, existing name, but the functionality was changed. Don't like confusing what it does with where it's used. otoh it's simple enough it could be a lambda. `file_path_less_than`?

TheCycoONE: It parallels std::less_than

CorsixTH/Src/iso_fs.cpp
TheCycoONE: Not a big fan that I missed these, but I will do a format patch later (soon) to fix the C++ format consistency project wide. Skip for now.

CorsixTH/Lua/dialogs/resizables/directory_browser.lua
TheCycoONE: :+1:

CorsixTH/Src/th.h
TheCycoONE: Done.

CorsixTH/Lua/dialogs/resizables/directory_browser.lua
TheCycoONE: I'll change them to the documented convention, and make a note to convert everything else at some point.

CorsixTH/Lua/filesystem.lua
TheCycoONE: Done

CorsixTH/Lua/dialogs/resizables/directory_browser.lua
TheCycoONE: Done.

CorsixTH/Src/iso_fs.cpp
```diff
@@ -70,33 +74,231 @@ constexpr int max_directory_depth = 16;
...
+        return *this;
+    }
+
+    /**
+     * Advance this iterator to the next file entry in the directory table,
+     * returning the result.
+     */
+    iso_directory_iterator& operator++() {
+        directory_ptr += *directory_ptr;
+        while (*directory_ptr == 0 && directory_ptr < end_ptr) {
```
Alberth289346: Swap the conditions to avoid accessing memory that you're not supposed to access

CorsixTH/Src/iso_fs.cpp
```diff
@@ -70,33 +74,231 @@ constexpr int max_directory_depth = 16;
...
+        }
+        return *this;
+    }
+
+    /**
+     * Advance this iterator to the next file entry in the directory table,
+     * returning a copy of the old iterator.
+     */
+    iso_directory_iterator operator++(int) {
+        iso_directory_iterator o(*this);
```
Alberth289346: `o` is a not so good variable name, maybe change to `old` ?

CorsixTH/Src/iso_fs.cpp
```diff
@@ -70,33 +74,231 @@ constexpr int max_directory_depth = 16;
...
+    /**
+     * Advance this iterator to the next file entry in the directory table,
+     * returning the result.
+     */
+    iso_directory_iterator& operator++() {
+        directory_ptr += *directory_ptr;
+        while (*directory_ptr == 0 && directory_ptr < end_ptr) {
+            ++directory_ptr;
+        }
+        if (directory_ptr == end_ptr) {
```
Alberth289346: For safety, compare with `>=` ?

TheCycoONE:
Added paranoia around the `iso_directory_iterator`. Force push now simplifies the exception handling in the advance pointer, so no state is changed until past the last fail point instead of trying to recover.

TheCycoONE:
Some more documentation and another mistake I noticed.

CorsixTH/Src/iso_fs.cpp
```diff
@@ -260,17 +275,32 @@ class iso_directory_iterator final {
...
+            ++new_dir_ptr;
+        }
+
+        // Catch a malformed directory entry where the size would extend past
+        // the end of the table.
+        if (new_dir_ptr < end_ptr && new_dir_ptr + *new_dir_ptr > end_ptr) {
+            throw std::runtime_error("The last directory entry was larger than the defined table region.");
+        }
+
+        if (new_dir_ptr < end_ptr) {
```
Alberth289346: Since the first check at line 294 is the same, you can move the comments, the second check, and the throw into here.

TheCycoONE: I felt it was clearer this way.

TheCycoONE:
Squashed up about half the commits. OK to merge as is now.

PR #1550, opened 2019-06-11
[RDY] Add Catch2 for unit testing

DavidFair:
Adds Catch2 and a simple sanity test to check the CI hooks correctly run

~~We use VCPKG to pull in Catch2 on MSVC, if your running *nix you will need to install it from your package manager to test.~~
User installation and package manager / VCPKG installations require different include paths in the cpp file. 
To avoid all this hassle I've bundled catch2, in the far future when it's widely available from package managers (when Ubuntu 16.04 is dropped) we can just switch to using cmake's find package instead.

This PR also moves everything that isn't `main` into a separate lib so we don't have to mess with dependencies for unit testing too


CorsixTH/CppTest/CMakeLists.txt
```diff
@@ -25,15 +25,21 @@ set (TESTING_FILES
     example.cpp
 )
 
-find_package(Catch2 REQUIRED)
+# Ubuntu 16.04 does not have catch and it requires a different
```
TheCycoONE: Better to leave find_package(Catch2 REQUIRED) and add a FindXXX module in CMake for grabbing the header on systems that don't have it installed with the Catch2config.cmake file. I don't want to maintain this in tree.

DavidFair: I looked into using a FindXXX module and `ExternalProject_add` to automatically download the deps. This won't work as a find module needs the header at configuration, but external project add runs at build time.

Digging around I discovered we can just use the CMake find package with a small workaround for users who don't have Catch2 in their package manager. When this is merged I'll add it to the wiki (a copy is in `.travis.yml`)

.travis.yml
```diff
@@ -1,4 +1,4 @@
-sudo: false
+sudo: true
```
TheCycoONE: Drop this line entirely instead of changing it. It was deprecated/removed from travis, and should have been dropped earlier.

.travis.yml
```diff
@@ -11,8 +11,18 @@ addons:
       - doxygen
       - sshpass
       - ant
+      - git
```
TheCycoONE: Both git and cmake should already be included in the image, no?

DavidFair: CMake is only installed for the current user, so when we run as root it falls over. Installing system-wide avoids us having to mess with paths

Adding git too was a result of mind block though

CorsixTH/CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ list(REMOVE_ITEM corsixth_source_files "${CMAKE_SOURCE_DIR}/CorsixTH/Src/main.cp
                                    "${CMAKE_SOURCE_DIR}/CorsixTH/SrcUnshared/main.cpp")
 list(REMOVE_ITEM corsixth_source_files "${CMAKE_SOURCE_DIR}/CorsixTH/Src/main.h")
 
-add_library(corsixth_lib STATIC ${corsixth_source_files})
+add_library(Corsixth_lib STATIC ${corsixth_source_files})
```
TheCycoONE: It's rather pedantic, but can we go with `CorsixTH_lib`?

CorsixTH/CMakeLists.txt
```diff
@@ -63,22 +82,28 @@ if(APPLE)
...
   set(MACOSX_BUNDLE_ICON_FILE Icon.icns)
 
-  add_executable(CorsixTH MACOSX_BUNDLE ${corsixth_source_files} ${corsixth_icon_file})
+  add_executable(CorsixTH MACOSX_BUNDLE ${corsixth_main_files} ${corsixth_icon_file})
 
   set_target_properties(CorsixTH PROPERTIES LINK_FLAGS_MINSIZEREL "-dead_strip")
   set_target_properties(CorsixTH PROPERTIES XCODE_ATTRIBUTE_LD_RUNPATH_SEARCH_PATHS "@executable_path/../Frameworks")
 
   target_link_libraries(CorsixTH SDL2main)
+  target_link_libraries(CorsixTH Corsixth_lib)
```
TheCycoONE: Linking CorsixTH_lib can be done outside of the `if` block.

scripts/build_vcpkg_deps.ps1
```diff
@@ -24,17 +24,17 @@
...
+    [Parameter(Mandatory = $true)][string]$VcpkgCommitSha
 )
 
 ################
 # Variables
 ################
 
 $anim_view_libs = "wxwidgets"
-$corsixth_libs = "ffmpeg", "freetype", "lua", "luafilesystem", "lpeg", "sdl2", "sdl2-mixer[libflac,mpg123,libmodplug,libvorbis]", "luasocket"
+$corsixth_libs = "ffmpeg", "freetype", "lua", "luafilesystem", "lpeg", "sdl2", "sdl2-mixer[libflac,mpg123,libmodplug,libvorbis]", "luasocket", "Catch2"
```
TheCycoONE: port name is `catch2` all lowercase. This might matter if we use vcpkg in other environments in the future.

scripts/build_vcpkg_deps.ps1
```diff
@@ -71,15 +71,16 @@ function run_script {
         run_command -command "git clone $vcpkg_git_url $dest_folder_name"
         Set-Location -Path $dest_folder_path
         run_command "git checkout $VcpkgCommitSha"
-    } else {
+    }
+    else {
```
TheCycoONE: Why did you change the formatting here (and at the catch)?

TheCycoONE:
Need to drop catch.hpp from the patch still.

TheCycoONE:
I think that's it. The above mentioned things, and adding a trailing newline to every file and we should be good for this PR.

scripts/build_vcpkg_deps.ps1
DavidFair: The auto formatting in VS Code is set to do this automatically, I'm going to be cheeky and leave it otherwise I will have to remember to undo this each time I open this file

PR #1551, opened 2019-06-11
[RDY] Make CMake src files explicit

DavidFair:
*Requires #1550 to be merged first*

Normally I avoid feature branches based on feature branches, but the previous one added a static library we will have merge conflicts with later.

Names all .cpp and .hpp files we use explicitly. This uses a separate file to keep the bloat out of the main CMake list, however I'm open for suggestions if anyone has a better way?

### DIFF
--- a/CorsixTH/CMakeLists.txt
+++ b/CorsixTH/CMakeLists.txt
@@ -28,21 +28,8 @@ else()
   set(CORSIX_TH_INTERPRETER_PATH ${CMAKE_INSTALL_FULL_DATADIR}/corsix-th/CorsixTH.lua)
 endif()
 
-# Modify the config.h based upon our selection of options
-configure_file(${CMAKE_SOURCE_DIR}/CorsixTH/Src/config.h.in ${CMAKE_BINARY_DIR}/CorsixTH/Src/config.h)
-include_directories(${CMAKE_BINARY_DIR}/CorsixTH/Src/)
 
-# Generate source files list
-# Note: Done after generating config.h
-add_subdirectory(${PROJECT_SOURCE_DIR}/Src)
-
-add_library(Corsixth_lib STATIC ${corsixth_lib_src_files} ${corsixth_lib_hdr_files})
-
-set(corsixth_main_files
-  ${CMAKE_SOURCE_DIR}/CorsixTH/Src/main.cpp
-  ${CMAKE_SOURCE_DIR}/CorsixTH/SrcUnshared/main.cpp
-  ${CMAKE_SOURCE_DIR}/CorsixTH/Src/main.h
-)
+add_library(CorsixTH_lib STATIC "")
 
 # Declaration of the executable
 if(APPLE)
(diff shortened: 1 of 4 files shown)


CorsixTH/Src/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/th_movie.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/th_pathfind.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/th_sound.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/xmi2mid.cpp
+
+    ${CMAKE_SOURCE_DIR}/common/rnc.cpp
+    PARENT_SCOPE
+)
+
+set(corsixth_lib_hdr_files
```
TheCycoONE: I don't think the headers are actually needed. Could you check?

TheCycoONE: To answer my own question. Yes headers should be included, they are automatically excluded from compilation, but used for IDE generators.

TheCycoONE:
Please rebase on the latest master now that #1550 was squashed and merged.

CorsixTH/CMakeLists.txt
```diff
@@ -99,6 +58,23 @@ if(UNIX AND NOT APPLE)
...
+# so we can't just do it in in the folder like above
+target_sources(CorsixTH_lib
+  PRIVATE
+  ${CMAKE_SOURCE_DIR}/common/rnc.cpp
+  ${CMAKE_SOURCE_DIR}/common/rnc.h
+)
+
+# Ensure config.h is picked up by cmake - moving this into subdir cmake files will
+# prevent it applying to the CorsixTH project
+include_directories(${CMAKE_BINARY_DIR}/CorsixTH/Src/)
```
TheCycoONE: I think what you are looking for is `target_include_directories` https://cmake.org/cmake/help/v3.0/command/target_include_directories.html

CorsixTH/CppTest/CMakeLists.txt
```diff
@@ -46,3 +41,10 @@ set_target_properties(UnitTests PROPERTIES FOLDER UnitTests)
 
 # Make sure we add it as a dependency to all tests (some day we can add LuaTest here)
 add_dependencies(AllTests UnitTests)
+
+# List of all test files
+target_sources(UnitTests
```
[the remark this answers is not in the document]
DavidFair: Yes, the only reason is it separates the target definition from the list of files physically, so as the number of tests grow its easy to think:
Adding new test files -> Scroll to the bottom of the file
Changing the CMake target(s) in some way -> Stay at the top

CorsixTH/CMakeLists.txt
```diff
@@ -73,7 +73,7 @@ target_sources(CorsixTH_lib
 
 # Ensure config.h is picked up by cmake - moving this into subdir cmake files will
```
TheCycoONE: This comment is no longer accurate.

CorsixTH/CMakeLists.txt
```diff
@@ -73,7 +73,7 @@ target_sources(CorsixTH_lib
...
 # prevent it applying to the CorsixTH project
-include_directories(${CMAKE_BINARY_DIR}/CorsixTH/Src/)
```
TheCycoONE: I think this should be `PRIVATE`?

DavidFair:
Making this PRIVATE only adds the headers to the current target. As `CorsixTH` and `CorsixTH_lib` both use the header I was relying on the implicit inheritance as the former depends on the latter.

We can either make it explicit by including it for both targets, or carry on using the CMake public keyword, @TheCycoONE which would you prefer?

TheCycoONE:
Alright, I'm ok with leaving it PUBLIC for now. We can always adjust later if we have a reason to.

Issue #1558, opened 2019-08-10
Error in timer handler


carlosrd:
#### Describe the issue
While playing, this dialog appeared several times:
![errorcorsix](https://user-images.githubusercontent.com/2896275/62825601-f6278700-bbae-11e9-9f32-b646606a2f80.png)

I click accept but the dialog appears repeteadly until the error seems recovered.

At least two times ocurred in this level:

- [STAGE6-ERROR.zip](https://github.com/CorsixTH/CorsixTH/files/3489211/STAGE6-ERROR.zip)
- [STAGE6-ERROR2.zip](https://github.com/CorsixTH/CorsixTH/files/3489212/STAGE6-ERROR2.zip)

I think the problem is related to the bald guy blocked in the corner in the entrance of the hospital. In the first saved game he is in a vomiting loop:

![errorcorsix3](https://user-images.githubusercontent.com/2896275/62825705-62ef5100-bbb0-11e9-9865-ee11a0722e50.png)
In the second I've tried to send it to home but he still in the hospital

![errorcorsix2](https://user-images.githubusercontent.com/2896275/62825698-39362a00-bbb0-11e9-9d2f-46fb1cddad82.png)

#### System Information
**CorsixTH Version:** *0.63*

**Operating System:** *Windows 7 Home Premium SP1*

**Theme Hospital Version:** *CD*

#### Gamelog.txt
```
Created game with savegame version 134.
Warning: Trying to remove nonexistant room build callback (function: 000000001BF04890) from humanoid (table: 000000001EE1E950).
Warning: Trying to remove nonexistant room remove callback (function: 000000001BEEA4C0) from humanoid (table: 000000001EE1E950).
Warning: Trying to remove nonexistant staff callback (function: 000000001BF04900) from humanoid (table: 000000001EE1E950).
Warning: Trying to remove nonexistant room build callback (function: 000000001BF058C0) from humanoid (table: 000000001F124050).
Warning: Trying to remove nonexistant room remove callback (function: 000000001C002B80) from humanoid (table: 000000001F124050).
Warning: Trying to remove nonexistant staff callback (function: 000000001BF05930) from humanoid (table: 000000001F124050).
Warning: Trying to remove nonexistant room build callback (function: 000000001F16FDC0) from humanoid (table: 00000000247D5F20).
Warning: Trying to remove nonexistant room remove callback (function: 000000001C243EC0) from humanoid (table: 00000000247D5F20).
[...]

PR #1560, opened 2019-08-31
Correct cell flags for original map

mugmuggy:
Addressing #1558 with map flags on original TH map


CorsixTH/Lua/map.lua
```diff
@@ -778,4 +784,8 @@ function Map:afterLoad(old, new)
     -- Issue #1105 update pathfinding (rebuild walls) potentially broken by side object placement
     self.th:updatePathfinding()
   end
+  if old < 136 then
```
TheCycoONE: This is not screening by hospital, likely to cause problems.

PR #1562, opened 2019-09-11
[RDY] Cpp fmt

TheCycoONE:
Introduces clang-format and new style for C++, based closely on Googles style.

Rules not enforced by clang-format e.g. including braces on conditionals are not yet included.

Some other minor issues (compiler warnings, non-portable casts) are resolved in this patch set.


AnimView/frmSprites.cpp
```diff
@@ -20,91 +20,126 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
...
+            0,
+            wxALIGN_CENTER_VERTICAL);
     pFiles->Add(pFilesGrid, 0, wxEXPAND | wxALL, 1);
-    wxButton *pTmp;
-    pFiles->Add(pTmp = new wxButton(this, ID_LOAD, L"Load Simple"), 0, wxALIGN_CENTER | wxALL, 1);
-    pFiles->Add(pTmp = new wxButton(this, ID_LOAD_COMPLEX, L"Load Complex"), 0, wxALIGN_CENTER | wxALL, 1);
-    pFiles->Add(pTmp = new wxButton(this, ID_NEXT, L"Next"), 0, wxALIGN_CENTER | wxALL, 1);
+    wxButton* pTmp;
+    pFiles->Add(
+            pTmp = new wxButton(this, ID_LOAD, L"Load Simple"),
```
Alberth289346: Code is being really silly here, at least I don't see the need for `pTmp`.

AnimView/frmSprites.cpp
```diff
@@ -113,34 +148,40 @@ void frmSprites::_onNext(wxCommandEvent& evt)
...
-     ||!m_oAnims.loadSpriteFile(m_txtData->GetValue().IsEmpty() ? m_txtTable->GetValue().BeforeLast('.')+L".DAT" : m_txtData->GetValue())
-     ||!m_oAnims.loadPaletteFile(m_txtPalette->GetValue()))
-    {
+void frmSprites::load(bool bComplex) {
+    if (!m_oAnims.loadTableFile(m_txtTable->GetValue()) ||
+        !m_oAnims.loadSpriteFile(
+                m_txtData->GetValue().IsEmpty()
+                        ? m_txtTable->GetValue().BeforeLast('.') + L".DAT"
+                        : m_txtData->GetValue()) ||
+        !m_oAnims.loadPaletteFile(m_txtPalette->GetValue())) {
```
Alberth289346: Indentation of this line versus the next one is wrong imho. There should always be a distinction between code in the condition and code in the 'then' part.

CorsixTH/Src/cp437_table.h
```diff
@@ -3,24 +3,22 @@
...
+        0x00A1, 0x00AB, 0x00BB, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F,
+        0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F,
+        0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F,
+        0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F,
+        0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F,
+        0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x003F, 0x03B1, 0x00DF, 0x0393,
+        0x03C0, 0x03A3, 0x03C3, 0x03BC, 0x03C4, 0x03A6, 0x0398, 0x03A9, 0x03B4,
+        0x221E, 0x03C6, 0x03B5, 0x2229, 0x2261, 0x00B1, 0x2265, 0x2264, 0x2320,
+        0x2321, 0x00F7, 0x2248, 0x00B0, 0x2219, 0x00B7, 0x221A, 0x207F, 0x00B2,
+        0x25A0, 0x00A0};
```
Alberth289346: Pretty sure 8 values at a line makes much more sense for 0x80 numbers.

CorsixTH/Src/persist_lua.h
```diff
@@ -146,27 +131,24 @@ class lua_persist_reader
...
             tValue = static_cast<T>(tWrittenValue >> 1);
         return true;
     }
 
     template <class T>
-    bool read_float(T& fValue)
-    {
-        if(!read_byte_stream(reinterpret_cast<uint8_t*>(&fValue), sizeof(T)))
+    bool read_float(T& fValue) {
+        if (!read_byte_stream(reinterpret_cast<uint8_t*>(&fValue), sizeof(T)))
```
Alberth289346: Code can be shorter by returning the value of the function call, but for formatting, does it make sense to have multi-line statements without curly brackets?

CorsixTH/Src/random.c
```diff
@@ -180,28 +173,21 @@ static int l_randomdump(lua_State *L)
...
         int i;
         size_t len;
-        const char *data = lua_tolstring(L, 1, &len);
-        if(len != N * 4 + 2)
-            luaL_argerror(L, 1, "Seed string wrong length");
-        for(i = 0; i < N; ++i)
-            mt[i] = ((uint32_t*)data)[i];
+        const char* data = lua_tolstring(L, 1, &len);
+        if (len != N * 4 + 2) luaL_argerror(L, 1, "Seed string wrong length");
+        for (i = 0; i < N; ++i) mt[i] = ((uint32_t*)data)[i];
```
Alberth289346: Many function calls are split in a lot of lines, yet this is allowed. While I like this compact style, not sure if it fits in the general idea of the used style.

CorsixTH/Src/run_length_encoder.cpp
```diff
@@ -65,138 +59,120 @@ bool integer_run_length_encoder::initialise(size_t iRecordSize)
...
-                {
+            for (size_t iNumRecords = 1; iNumRecords <= 8; ++iNumRecords) {
+                for (size_t iOffset = 0; iOffset < iNumRecords; ++iOffset) {
                     size_t iNumRepeats = 0;
                     size_t iObjSize = iNumRecords * record_size;
-                    while(iObjSize * (iOffset + iNumRepeats + 1) <= buffer_size
-                        && are_ranges_equal(0, iNumRepeats, iOffset, iObjSize))
-                    {
+                    while (iObjSize * (iOffset + iNumRepeats + 1) <=
+                                   buffer_size &&
```
Alberth289346: Really weird breaking point.

CorsixTH/Src/th_gfx_font.h
```diff
@@ -28,15 +28,18 @@ SOFTWARE.
 #include FT_FREETYPE_H
 #endif
 
+class render_target;
+
```
Alberth289346: empty line doesn't look useful to me

CorsixTH/Src/th_gfx_font.h
```diff
@@ -87,17 +91,21 @@ class font
...
+            int iX,
+            int iY) const = 0;
 
     //! Draw a single line of text, splitting it at word boundaries
     /*!
         This function still only draws a single line of text (i.e. any line
-        breaks like `\r` and `\n` in sMessage are ignored), but inserts line breaks
-        between words so that no single line is wider than iWidth pixels.
+        breaks like `\r` and `\n` in sMessage are ignored), but inserts line
+       breaks between words so that no single line is wider than iWidth pixels.
```
Alberth289346: Why is the indenting different?

Alberth289346:
I don't quite like the large amount of lines that get created with function definitions and function applications, it's systematic and all that, but having two lines for `int iX, int iY` is more clutter than useful imho.

For function applications, I can see the idea, especially if an argument is more than a simple variable or value. For function definitions, I see no use. There is a piece of documentation that explains each parameter in calling order already.

CorsixTH/Src/th_gfx_font.h
TheCycoONE: There's a few of these: seems that clang-format doesn't understand this style of comment.

CorsixTH/Src/th_gfx_font.h
TheCycoONE: It's consistent with a rule about leaving a blank line of white space between class definitions. Perhaps not useful in this case, but not worth an override either imho.

AnimView/frmSprites.cpp
TheCycoONE: Out of scope - I got in a trap of refactoring everything earlier, but it can always be done later.

AnimView/frmSprites.cpp
TheCycoONE: This example has a condition that's far too long to be sane, it should be split up and put in it's own function.

I agree, jquery/eslint a style that would require or allow a blank line at the top of the code in the condition in this case would be better. In the end I opted to just keep a fairly minimal clang-format based on google's style, and that doesn't allow the blank line here.

Rather than fight it, shorten the conditional so it's not confusing.

CorsixTH/Src/cp437_table.h
TheCycoONE: Yes, I have an override for this case elsewhere, should add it here as well.

CorsixTH/Src/persist_lua.h
TheCycoONE: No, the new style requires braces here; just wasn't automatically picked up and I missed it.

TheCycoONE:
Fine, that's bin packing, and it was an override I did make based on personal taste - but I'll disable it  and learn to live with it.

CorsixTH/Src/run_length_encoder.cpp
TheCycoONE: Doesn't look weird to me if buffer_size didn't fit on the line, it's the least connected to what's before it, and the expressions left and right of the && are even less connected.

That said, again, a well named function (or functor, or lamba) for the conditional would help readability.

TheCycoONE: Doesn't look weird to me if buffer_size didn't fit on the line, it's the least connected to what's before it, and the expressions left and right of the && are even less connected.

That said, again, a well named function (or functor, or lamba) for the conditional would help readability.

CorsixTH/Src/persist_lua.h
TheCycoONE: It's only a single line - Google style allows it.

TheCycoONE:
Went full Google, because I want to avoid nit picking more than I care what style is used; and clang-format makes that an easy option.

Alberth289346:
2-space indenting??

Ah well, I give up, do what you think is good and just merge.

Issue #1561, opened 2019-09-03
Race condition can rarely lead to invalid seek_toilets action


mugmuggy:
#### Describe the issue

Through a thought process mainly (state was manpiuated to get me to the conclusion) I believe I found a rare but looks possible race condition related to this https://github.com/CorsixTH/CorsixTH/issues/555#issuecomment-527120811. I've hit the resulting error a few times across two different saves, but sometimes I get it, many times I don't,ie I found the cause of #555 whilst trying to identify this process.

The state has to be similar. Except this will be for normal rooms. Also I believe its more likely to occur when running at full speed, as tickDay, is called before the ticks. Ie if now + ticks = tomorrow, tickDay. Then the loop afterwards is for each tick that needs to execute.

A patient is queued, at least 3rd patient queued ie more likely to have actions: use_object (bench), queue, walk

Patient:tickDay executes, and the use toilet logic fires (setNextAction), action queue is use_object (being interrupted), queue, seek_toilets. Patient going_to_toilet now is yes., queue action is set todo_interrupt, use_object will undergo an interrupt, ie patient animation to stand starts.

Within the next one or two game ticks, the queue for the queue action, will have queue:pop() called, which will initiate the callbacks for queue:onChangeQueuePosition. A bench has now become free that is closer to the room, and onChangeQueuePosition finds the queue task, and then inserts walk and use_object (for the closer bench) actions.

Patients action queue becomes, walk, use_object, queue (todo_interrupt), seek_toilets, 

Either during this walk, or whilst on the bench, the patient tickDay is called again and this time Patient:pee() executes, resetting the going_to_toilet flag to no.

As the patient is seated is still at least the 2nd position in the queue, they are eligible in Room:trytoFindNearbyPatients. As the going_to_toilet flag is now "no", tryMovePatient can succeed in routing the patient., This attempts to find the walk after the queue action (which is now no longer there) and replace it with the new enter action to the new room. And also in doing so, sets todo_interrupt on the seek_toilets action.

As no new action appears after seek_toilets, it results in an empty action queue dialog and as the patient is then given a SeekReceptionAction and can continue on, as they'll be redirected to the room (unless the room is the toilet where they will go into the loop of #555).

mugmuggy:
Being mindful I'm also playing with around with a wip testing stuff - I believe an opportunity also exists for an assert in humandoid_actions\queue.lua partially as a result of the change in #1453.

[...]

Issue #1081, opened 2016-05-31
Handymen trying to water plant blocks room


jorritb:
#### Describe the issue

I noticed that there are two handymen waiting outside my Jelly Vat who seem to want to go in and water the plant, but they never go in and block the room. When I pick one up and place him in the room he just stands in the room waiting (he stands exactly behind the Jelly Vat). When I pick up the handymen either inside or outside the room, more show up and also start waiting. If I fire all handymen, including the ones inside the room, they become stuck and the room remains unusable. The same happens if I try to delete the room.

![handymen-plant](https://cloud.githubusercontent.com/assets/11193663/15691761/ebfaaad4-278a-11e6-9b09-d04e521554df.png)
.
#### Steps to Reproduce

Check the save I attached (you probably also need the .map and .level since this is a custom map)

Save file: https://drive.google.com/open?id=0B1z_KXLTgDplMzhWSlpONGdoTDQ
Level: https://drive.google.com/open?id=0B1z_KXLTgDplUmxab3R1MEw2dFE
Map: https://drive.google.com/open?id=0B1z_KXLTgDplcm5vMnpKQ3ZkQ1k
#### Expected Behavior

This shouldn't happen.
#### System Information

CorsixTH 0.60 om Windows 10.
#### Gamelog.txt

It only says "savegame version 111"

TheCycoONE:
All these stuck room issues are very tricky. Thanks for the report.

mugmuggy:
Here the door is reserved by the patient that I assume just used the door. They show up as having left the hospital and checking the previous_action (on trunk), the last action appears to be leaving the map (walking to the edge of the map), but they still have the door reserved.

The hack I guess it to update the hack in Door:checkForDeadlock to account for despawn actions
`      if action.name == "queue" or action.name == "spawn" and action.mode == "despawn" then`

mugmuggy:
From some experimentation this bit of code appears to be contributing to the patients and potentially staff reserving the door.

https://github.com/CorsixTH/CorsixTH/blob/e14d4ffa993a1d3fa71b6bca293d77602f388c74/CorsixTH/Lua/humanoid_actions/walk.lua#L427-L429

I put an 

> if action.is_leaving then

check in place wrapping the assignment of the humanoid to the reserved_for and so far it seems fairly reliable in preventing the patient holding the door.

As the QueueAction and the Walk action both have reservations set.

[...]

PR #1565, opened 2019-10-01
[RDY] Fix patient reserving doors

mugmuggy:
Hopefully thisis a suitable refactor as mentioned in #1453
Addresses my comment in #1561 with saved_must_happen not existing, and therefore must_happen not being correctly set.

Patients holding reservations on doors (#1081), reserve_on_resume needs to be copied, as this was not copied to the requeued walk action, the interrupt handler, would not unreserve the door, when interrupting walk action. This is brought about for drinks machine (and would be for falling) as action.keep_reserved is set to true, which bypasses the door unreserve logic.

Then as it wasn't copied, if they were interrupted again, the door would remain reserved.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -367,28 +367,8 @@ function Patient:falling()
     if self.on_ground then
       self:setNextAction(GetUpAction())
     end
-    if current.name == "idle" or current.name == "walk" then
-      self:queueAction({
-        name = current.name,
-        x = current.x,
-        y = current.y,
-        must_happen = (current.name == "walk" and current.saved_must_happen) or (current.name == "idle" and current.must_happen),
-        is_entering = current.is_entering,
-      }, 2)
-    else
-      self:queueAction({
-        name = current.name,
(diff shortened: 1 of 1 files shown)


mugmuggy:
I've got some checks sitting in some code that I need to validate are actually required for this and the other PR, they might have just been related to trying to avoid warnings printed from leaving queue actions as well.

TheCycoONE:
Ok. Well I am good with landing this one when you are ready.

mugmuggy:
In #1081 I made the note of preventing the drinks machine use if reserve_on_resume was set. However I did quickly check TH and managed to witness, a patient popped from start of queue, make their way to the drinks machine before entering room, so this seems a possibilty.

This last change ensures that if the walk after leaving queue, into the room after the reservation is made is interrupted for drinks machine use as an example, then whilst the patient is using the drinks machine, they are evacuated/kicked etc, that the walk action into room will be interrupted and not just removed via setNextAction.

mugmuggy:
That last change is defective in particular scenarios and am exploring another option.

mugmuggy:
Corrected the issue with the code originally with #1568

Issue #1556, opened 2019-08-10
Urgency patients not cured if facility is built after accept the urgency


carlosrd:
#### Describe the issue
If you receive an urgency that requires a facility that is not yet built and then you accept the urgency and builds the facility during the patients arrive, makes the urgency patients stuck in the hospital without being cured

#### Steps to Reproduce
1. Receive an urgency of a treatment that needs a facility that is not already built.
2. Accept it
3. Build the necessary facility to accomplish the urgency after acepting the urgency
4. The urgency patients arrive and wait infinitely in the hospital until the times ends (then they die)

You can download this [saved game](https://github.com/CorsixTH/CorsixTH/files/3489174/STAGE7-UrgencyError.zip) and check that there is one urgency patient to be treated in the surgery room that was built while the patient was arriving, but the patient is stuck in the hospital without getting in the row of the surgery room 

#### Expected Behavior
Patients must be notified about new buildings to get cured or treated

#### System Information
**CorsixTH Version:** *0.63*

**Operating System:** *Windows 7 Home Premium SP1*

**Theme Hospital Version:** *CD*

mugmuggy:
The seek room action changes I made in 0.63 around the room detection scenarios I believe is specifically checking for the existence of both the ward and operating theatre here and not seeking. I found another issue with this last week where you could also destroy a room (diagnosis only) even though it wasn't seeking the treatment rooms specifically this prevented the seek action from executing. I'll tack this on to the other change.

Issue #1268, opened 2017-10-27
Patient being in treatment can't die nor be cured


vyguera:
#### Describe the issue
My issue can be 2 issues indeed:
When a patient triggers its death while entering or being at its treatment room (at least in psychiatry) he/she can't die nor be cured. The doctor doesn't treat the patient and walks around the room as if he were waiting for a new patient but the patient doesnt leave the room.

This makes the doctor get tired till he reaches 100% of fatigue. Even though he doesn't leave the room.

So, the 2 issues I observe are:
1- Treatment can't finish if patient triggers death while entering the room and/or being lied on the sofa.
2 - Doctors (and maybe nurses too) can't finish the treatment and walks around the room if they reaches its fatigue 100% limit (I think it also happens when the limit is at the usual 60%)

#### Steps to Reproduce
1. *Load save game attached to this ticket*

2. Doctor at psychiatry is already walking around the room and the patient lied on the sofa

3. Wait for a minute so the hour of death arrives to the patient.

4. The patient can't die nor be treated. Also the doctor can't leave the room as he reaches its fatigue limit.

#### Save Game
https://app.box.com/s/sfi5ef63ahiod7fxpy2j3d0ks3v50rny

#### Expected Behavior
Ttreatment must always finish.
Although if the patient's hour of death arrives I don't know if he/she should always die or might be cured
If Doctor gets tired while treating he also must always finish the treatment and then leave the room.

#### System Information
CorsixTH Version: 0.60

Operating System: Windows 7, 64bits

Theme Hospital Version: CD

#### Gamelog.txt
https://app.box.com/s/cuqfzkr46c6mm4hnq8632e5pttgz1oei

### Additional Info
![patient dead while in treatment](https://user-images.githubusercontent.com/1656373/32119112-7d5acad2-bb54-11e7-901e-47c368eef95e.png)

lewri:
I wonder if this issue could be caused by the savegame change. But nonetheless I tried to dig around. 
The log doesn't seem to provide anything useful so I'll glaze over that.

Are you able to provide a save before this step (i.e. before our dying patient entered the room)?

When I clicked onto our patient he seemed to have the 'guess at cure' and 'queue status' buttons despite being in the room. I was able to send him home, which resulted in an error.

> Warning: goHome called when the patient is already going home
> Savegame version is 111 (v0.60), originally it was 111 (v0.60)
> Savegame version is 111 (v0.60), originally it was 111 (v0.60)
[...]

Issue #1233, opened 2017-07-23
Error: CorsixTH/Lua/entities/humanoid.lua:665: Can only finish current action


hnsl:
#### Describe the issue

Running: The timer handler.
A stack trace is included below, and the handler has been disconnected.
CorsixTH/Lua/entities/humanoid.lua:665: Can only finish current action
stack traceback:
        [C]: in function 'assert'
        CorsixTH/Lua/entities/humanoid.lua:665: in method 'finishAction'
        CorsixTH/Lua/rooms/general_diag.lua:68: in field 'after_use'
        CorsixTH/Lua/humanoid_actions/multi_use_object.lua:316: in local 'timer_function'
        CorsixTH/Lua/entity.lua:216: in method 'tick'
        CorsixTH/Lua/world.lua:1051: in method 'onTick'
        CorsixTH/Lua/app.lua:1005: in function <CorsixTH/Lua/app.lua:1002>
        (...tail calls...)
        CorsixTH/Lua/app.lua:906: in function <CorsixTH/Lua/app.lua:901>

Warning: No event handler for timer

#### System Information
**CorsixTH Version:** 6a079fd02bc456be7410e4394d7661308e86b3bf

**Operating System:** Arch Linux 4.10.13-1-ARCH

**Theme Hospital Version:** CD

### Additional Info

Happened when I was tearing down a psychiatrists office that had a long queue.

TheCycoONE:
https://github.com/CorsixTH/CorsixTH/blob/6a079fd02bc456be7410e4394d7661308e86b3bf/CorsixTH/Lua/entities/humanoid.lua#L665 is the assert. No relevant information in that method.

https://github.com/CorsixTH/CorsixTH/blob/6a079fd02bc456be7410e4394d7661308e86b3bf/CorsixTH/Lua/rooms/general_diag.lua#L68 is more interesting.

It seems that between when the staff was assigned the idle action while the patient used the trolly, and when the patient finished the staff had already been assigned another action. I'm not sure whether tearing down the other room had any impact.

TheCycoONE:
It would be helpful for diagnosing if that assert gave a little more information about the state of the action queue; or if we had a state dump -

@hnsl is the issue reproducible or was it a one time thing?

hnsl:
Hi, I was in the middle of a game. I just kept playing, didn't try to reproduce anything. Sorry!

bacon1986:
**Describe the issue**
[...]

Issue #858, opened 2015-08-08
"Trying to remove nonexistant room build callback" freeze/crash


NikNaks:
##### What steps will reproduce the problem?
1. Run Autosave7.sav
2. Wait a few seconds
3. The game will freeze/crash
##### What version of CorsixTH are you using (e.g. "0.30", "4c66e39985")?

0.5 RC1
##### What operating system / compile settings are you using?

Windows 10
##### What level was this on (e.g. "Demo level", "Full game level 12")?

Level 10 (Eggsenham)
##### Please provide any additional information below.

Uploads of gamelog and autosaves here: https://db.tt/AgUDUY76
I've included Autosave6 as well, in case this helps shed light on what changed between the two saves to cause the crash.
##### gamelog.txt output

```
Created game with savegame version 105.
Warning: Trying to remove nonexistant room build callback (function: 0000000F2E6AFC80) from humanoid (table: 0000000F2ED67EE0).
Warning: Trying to remove nonexistant room remove callback (function: 0000000F2A485F80) from humanoid (table: 0000000F2ED67EE0).
[repeated mulitple times]
Warning: Trying to remove nonexistant room build callback (function: 0000000F2E159FE0) from humanoid (table: 0000000F23930A00).
Warning: Trying to remove nonexistant room remove callback (function: 0000000F2A343B40) from humanoid (table: 0000000F23930A00).
Error in timer handler: 
C:\Program Files\CorsixTH\Lua\entities\humanoid.lua:662: Can only finish current action
stack traceback:
    [C]: in function 'assert'
    C:\Program Files\CorsixTH\Lua\entities\humanoid.lua:662: in method 'finishAction'
    C:\Program Files\CorsixTH\Lua\rooms\general_diag.lua:75: in field 'after_use'
    ...Files\CorsixTH\Lua\humanoid_actions\multi_use_object.lua:264: in local 'timer_function'
    C:\Program Files\CorsixTH\Lua\entity.lua:218: in method 'tick'
    C:\Program Files\CorsixTH\Lua\world.lua:1121: in method 'onTick'
    C:\Program Files\CorsixTH\Lua\app.lua:878: in function <C:\Program Files\CorsixTH\Lua\app.lua:875>
    (...tail calls...)
    C:\Program Files\CorsixTH\Lua\app.lua:779: in function <C:\Program Files\CorsixTH\Lua\app.lua:774>
```

Alberth289346:
[...]

PR #1567, opened 2019-10-07
[RFC] Some assorted issues

mugmuggy:
Contagious patients can't be treated if they are seeking room.

Adviser says with override was basically truncating the entire list down to the existing first message rather than leaving only the override one.

Broken seek is from #1556 I had identified this. Ward in that instance being the diagnostic room.

The setNextAction when the queue is destroyed would leave patients going from drinks machine to bench, as action_queue_on_change_position gets called from the callback process of action_queue_get_soda. Queue:rerouteAllPatients, wasn't a candidate to flag them as not is_in_queue as this had other artefacts.

Patients appearing to leave is to correct #1233 and related #500, #362, #858 and #650

Staff answering calls to another room, will appear as within the room they are in, and allow patients to enter. If that call was then given to another staff member, they would meander in room (bad for GP) and the patient would normally enter causing a stuck room situation. (#1268)

### DIFF
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -544,7 +544,7 @@ end
   @param patient (Patient) the patient we wish to determine if they are static.]]
 local function is_static(patient)
   local action = patient:getCurrentAction()
-  return action.name == "queue" or action.name == "idle" or
+  return action.name == "queue" or action.name == "idle" or action.name == "seek_room" or
       (action.name == "use_object" and action.object.object_type.id == "bench")
 end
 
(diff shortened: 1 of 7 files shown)


CorsixTH/Lua/epidemic.lua
```diff
@@ -544,7 +544,7 @@ end
   @param patient (Patient) the patient we wish to determine if they are static.]]
 local function is_static(patient)
   local action = patient:getCurrentAction()
-  return action.name == "queue" or action.name == "idle" or
+  return action.name == "queue" or action.name == "idle" or action.name == "seek_room" or
```
TheCycoONE: This list still isn't exhaustive. There must be a better way to indicate whether a patient is currently moving - like annotating each of the actions?

PR #1574, opened 2019-12-29
[RDY] Clang tidy

TheCycoONE:
A clang-tidy rule set, and fixes for the issues it detected.

Travis CI is also updated to bionic, mainly to get a newer version of clang-tidy.
CorsixTH/deps are no longer used; instead this brings us back to using the distro packages for dependencies.

### DIFF
--- a/AnimView/frmSprites.h
+++ b/AnimView/frmSprites.h
@@ -48,9 +48,9 @@ class MyVScrolled : public wxVScrolledWindow {
     iMyCount = ROW_COUNT;
   }
 
-  wxCoord OnGetRowHeight(size_t row) const { return 1; }
+  wxCoord OnGetRowHeight(size_t row) const override { return 1; }
 
-  wxCoord EstimateTotalHeight() const { return iMyCount; }
+  wxCoord EstimateTotalHeight() const override { return iMyCount; }
 
   int iMyCount;
 };
@@ -58,7 +58,7 @@ class MyVScrolled : public wxVScrolledWindow {
 class frmSprites : public wxFrame {
  public:
   frmSprites();
-  ~frmSprites() = default;
+  ~frmSprites() override = default;
 
   enum {
     ID_LOAD = wxID_HIGHEST + 1,
--- a/AnimView/frmSprites.h
+++ b/AnimView/frmSprites.h
@@ -23,6 +23,7 @@ SOFTWARE.
 #pragma once
 
 #include "config.h"
+
 #include <wx/bitmap.h>
 #include <wx/button.h>
 #include <wx/checkbox.h>
@@ -32,7 +33,9 @@ SOFTWARE.
 #include <wx/textctrl.h>
 #include <wx/timer.h>
 #include <wx/vscroll.h>
+
 #include <vector>
+
 #include "th.h"
 
 static const int ROW_COUNT = 1000;
--- a/AnimView/frmSprites.h
+++ b/AnimView/frmSprites.h
@@ -58,7 +58,7 @@ class MyVScrolled : public wxVScrolledWindow {
 class frmSprites : public wxFrame {
  public:
   frmSprites();
-  ~frmSprites();
+  ~frmSprites() = default;
 
   enum {
     ID_LOAD = wxID_HIGHEST + 1,
--- a/AnimView/frmSprites.h
+++ b/AnimView/frmSprites.h
@@ -20,7 +20,8 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
-#pragma once
+#ifndef ANIMVIEW_FRMSPRITES_H_
+#define ANIMVIEW_FRMSPRITES_H_
 
 #include "config.h"
 
@@ -93,3 +94,5 @@ class frmSprites : public wxFrame {
   MyVScrolled* m_panFrame;
   DECLARE_EVENT_TABLE()
 };
+
+#endif
--- a/AnimView/th.cpp
+++ b/AnimView/th.cpp
@@ -47,15 +47,11 @@ static const unsigned char palette_upscale_map[0x40] = {
 
 class ChunkRenderer {
  public:
-  ChunkRenderer(int width, int height, unsigned char* buffer = nullptr) {
-    m_data = buffer ? buffer : new unsigned char[width * height];
+  ChunkRenderer(int width, int height)
+      : m_x(0), m_y(0), m_width(width), m_height(height), m_skip_eol(false) {
+    m_data = new unsigned char[width * height];
     m_ptr = m_data;
     m_end = m_data + width * height;
-    m_x = 0;
-    m_y = 0;
-    m_width = width;
-    m_height = height;
-    m_skip_eol = false;
   }
 
   ~ChunkRenderer() { delete[] m_data; }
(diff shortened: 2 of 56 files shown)


AnimView/frmSprites.h
```diff
@@ -23,6 +23,7 @@ SOFTWARE.
 #pragma once
 
 #include "config.h"
+
```
Alberth289346: I wonder about the `#pragma once` line, which is non-standard cpp afaik. Wikipedia claims it to be an alternative to the `#ifndef/#define/#endif` lines, which I just merged a fix for.
According to https://en.wikipedia.org/wiki/Pragma_once#Portability portability looks like a non-issue, but perhaps settle on one form?

Alberth289346:
Looks good, there are quite a few semantic fixes in there.

For the `path_node` vector, it might be a good idea to wrap a class around it so the 2D access to/from vector indexing gets localized to that class.

AnimView/frmSprites.h
TheCycoONE: Right, should replace that pragma. Thanks.

TheCycoONE:
I agree with the path_node comment, but will postpone it to a later fix.

TheCycoONE:
That's probably the full set of checks / code changes I want to implement for now. I'd like to have this run as part of CI - but what's there is ready for review.

TheCycoONE:
Ok, all running through CI now. There are a couple differences between my environment (clang-tidy-9 running with compile_commands.json generated on clang on aarch64) and travis CI that I don't love but I can live with.

AnimView/th.cpp
```diff
@@ -60,7 +60,7 @@ class ChunkRenderer {
 
   unsigned char* takeData() {
     unsigned char* buffer = m_data;
-    m_data = 0;
+    m_data = nullptr;
     return buffer;
```
Alberth289346: looks like a `std::move`, no idea if that is feasible in its environment. Different patch at least.

AnimView/th.cpp
```diff
@@ -374,7 +378,8 @@ void THAnimations::drawFrame(wxImage& imgCanvas, size_t iAnimation,
   int iFarX = 0;
   int iFarY = 0;
   while ((pElement = _getElement(iListIndex++))) {
-    if (pMask != NULL && !pMask->isSet(pElement->flags >> 4, pElement->layerid))
+    if (pMask != nullptr &&
+        !pMask->isSet(pElement->flags >> 4, pElement->layerid))
```
Alberth289346: field access could be moved into `pElement`

AnimView/th.cpp
TheCycoONE: This would be a case for std::move if we weren't working with raw pointers. As is the reference is switched without copying data already.

Now maybe we shouldn't be working with raw pointers...

AnimView/th.cpp
TheCycoONE: Right now th_element_t and a number of other structs in this class have exact sizing and layout requirements (#pragma pack(1)) making them difficult to extend, and to some extent implementation defined behaviour.

We do the same thing in CorsixTH proper but we immediately copy the properties on to another non-packed struct after loading. (CorsixTH/Src/th_gfx.cpp:253)

Room for improvements there.

mugmuggy:
@TheCycoONE, with MSVC this appears to be throwing exceptions as open_idx can be initialised to -1 (and remain that value), and size_t is unsigned. Resulting in access on the vector operator [] with undefined behaviour I believe.

https://github.com/CorsixTH/CorsixTH/blob/ac9ea2ebb7d1a059197c19aec98988d2e923e6af/CorsixTH/Src/th_pathfind.cpp#L532-L536

https://github.com/CorsixTH/CorsixTH/blob/ac9ea2ebb7d1a059197c19aec98988d2e923e6af/CorsixTH/Src/th_pathfind.cpp#L280

I'm not getting any warnings on the comparisons in latest trunk. I haven't looked to find if its driven from the lua side in any particular cases either as yet.

Alberth289346:
Since unsigned includes 0, the while condition effectively checks `i != 0`

TheCycoONE:
Thank mugmuggy, I'll check it out.

TheCycoONE:
Alright, there is no undefined behaviour. In C++14 negative unsigned values are well defined (2s complement representation of the same size byte converted).

That said -1 should only be on nodes that are not in the open list, and we shouldn't be promoting those nodes, so I'm not sure how you hit this.

The most suspect case looks like: https://github.com/CorsixTH/CorsixTH/blob/04ddfc6f8585c2f828c7884a2d0403b270745ec3/CorsixTH/Src/th_pathfind.cpp#L99

Could you open another issue for this, with the stack trace if you have it?

mugmuggy:
The 'undefined behaviour' is the 'vector[not a valid position]', not the conversion., ie vector.at() throws a specific exception, but its undefined when position > size() with  [] - from what I've read - I'm no c++ guru in the faintest.

For demonstration with that save I mentioned, I took this build https://ci.appveyor.com/project/TheCycoONE/corsixth/builds/30049191/artifacts overlayed binaries, filesystem and app.lua over the top of 0.63 and hit the exception, but didn't with just 0.63 only. Observing the same save I could move a handyman in the corridor, that caused it to skip, but a doctor answering a call triggers another.

The stack trace is indicating the line you've got as the culprit.
CorsixTH.exe!abstract_pathfinder::record_neighbour_if_passable(path_node * pNode, map_tile_flags neighbour_flags, bool passable, path_node * pNeighbour=0x0000000013d68b80) Line 99

This here appears to be the influence, if the node was already visited, and we set this back and it becomes the candidate node again, open_idx will be -1, due to the path taken by L99 above. 

https://github.com/CorsixTH/CorsixTH/blob/497ce86f23eec9765d453e8319d1024bf5f30024/CorsixTH/Src/th_pathfind.cpp#L279-L281

I put a simple check if `pNode->prev == pNode` around that assignment and it also prevents the crash. The push_to_open_heap should correctly set the open_idx the first time, its visited and the next time it should promote it - the 'int' code would have skipped the processing of the while loop in open_heap_promote entirely.

TheCycoONE:
You're right, that assignment in the while loop is bad. I'll need to determine what they were actually trying to do.

I also plan to add an assert that the index is in range and points to itself in the open_heap_list.

mugmuggy:
My guess is that we don't need a shortest path here, we just need to find an idle tile nearby that is accessible from a start point, and then find the most optimum path to that node with a WalkAction.

TheCycoONE:
Alternately a simple Dijkstra search until we hit an idle tile, or exceed out max cost.

TheCycoONE:
Actually - I bet that's what they were going for!

TheCycoONE:
@mugmuggy I've investigated this more or less completely now.
What this 'idle_tile_finder' is primarily built around (and the reason for the hacky open_idx = -1) is queuing. To get good results for queuing they want the nth found result to be preferably next to the n-1 found result, so instead of being a dijsktra algorithm, they turn A* on it's head and give one node a negative guess equal to it's distance so that the closest (as the bird flies) neighbour will be promoted immediately. This would fall apart if we promoted a node we've visited before so to combat that they set every visited node to get a -1 index to prevent it from bubbling up and getting visited again.

I can restore the old behaviour with any large magic value that we check to skip promotion; but I'd rather do something more obvious.

Meander ends up using this same work around even though it only cares about getting one random reachable nearby idle tile.

mugmuggy:
Makes sense. I should have realised this as I have previously looked at the MeanderAction as I noticed it likes to 'snake' on itself, when in an open area. Where TH would have staff wander around a bit more - they appear to be move random amount one direction, then a random amount in another, no zig zag movement and potentially further away - but CorsixTH tends to attract itself around the same spot, as obviously its trying to be as close as possible to its starting location as well as close to the previous tile.

PR #1576, opened 2020-01-20
Fix unsigned integer comparisons

tobylane:
Clang warns about this, are they genuine problems?

> [ 19%] Building CXX object CorsixTH/CMakeFiles/CorsixTH_lib.dir/Src/th_gfx_sdl.cpp.o
CorsixTH/CorsixTH/Src/th_gfx_sdl.cpp:1107:51: warning: 
      comparison of unsigned expression < 0 is always false [-Wtautological-compare]
  if (iWidth <= 0 || iHeight <= 0 || iPixelNumber < 0 ||
                                     ~~~~~~~~~~~~ ^ ~
1 warning generated.
[ 22%] Building CXX object CorsixTH/CMakeFiles/CorsixTH_lib.dir/Src/th_lua.cpp.o
[ 25%] Building CXX object CorsixTH/CMakeFiles/CorsixTH_lib.dir/Src/th_lua_anims.cpp.o
CorsixTH/CorsixTH/Src/th_lua_anims.cpp:320:13: warning: 
      comparison of unsigned expression < 0 is always false [-Wtautological-compare]
  if (iAnim < 0 || iAnim >= pManager->get_animation_count())
      ~~~~~ ^ ~
1 warning generated.
[ 27%] Building CXX object CorsixTH/CMakeFiles/CorsixTH_lib.dir/Src/th_lua_gfx.cpp.o
CorsixTH/CorsixTH/Src/th_lua_gfx.cpp:160:15: warning: 
      comparison of unsigned expression < 0 is always false [-Wtautological-compare]
  if (iSprite < 0 || iSprite >= pSheet->get_sprite_count())
      ~~~~~~~ ^ ~

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -1104,7 +1104,7 @@ namespace {
  */
 uint32_t get32BppPixel(const uint8_t* pImg, int iWidth, int iHeight,
                        const ::palette* pPalette, size_t iPixelNumber) {
-  if (iWidth <= 0 || iHeight <= 0 || iPixelNumber < 0 ||
+  if (iWidth <= 0 || iHeight <= 0 ||
       iPixelNumber >= static_cast<size_t>(iWidth) * iHeight) {
     return palette::pack_argb(0, 0, 0, 0);
   }
--- a/CorsixTH/Src/th_lua_anims.cpp
+++ b/CorsixTH/Src/th_lua_anims.cpp
@@ -316,8 +316,9 @@ int l_anim_get_crop(lua_State* L) {
 int l_anim_set_anim(lua_State* L) {
   animation* pAnimation = luaT_testuserdata<animation>(L);
   animation_manager* pManager = luaT_testuserdata<animation_manager>(L, 2);
-  size_t iAnim = luaL_checkinteger(L, 3);
-  if (iAnim < 0 || iAnim >= pManager->get_animation_count())
+  lua_Integer iAnim = luaL_checkinteger(L, 3);
+  if (iAnim < 0 ||
+      iAnim >= static_cast<lua_Integer>(pManager->get_animation_count()))
     luaL_argerror(L, 3, "Animation index out of bounds");
 
   if (lua_isnoneornil(L, 4)) {
@@ -326,7 +327,7 @@ int l_anim_set_anim(lua_State* L) {
     pAnimation->set_flags(static_cast<uint32_t>(luaL_checkinteger(L, 4)));
   }
 
-  pAnimation->set_animation(pManager, iAnim);
+  pAnimation->set_animation(pManager, static_cast<size_t>(iAnim));
   lua_settop(L, 2);
   luaT_setenvfield(L, 1, "animator");
   lua_pushnil(L);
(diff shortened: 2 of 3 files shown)


TheCycoONE:
size_t is tricky, it's platform dependent whether it's signed or unsigned.

TheCycoONE:
For the two lua_ ones, there's an implicit cast as well. Solve both issues by leaving it as an int for the first comparison check, explicitly static_cast to size_t later if that's what's being used in the api.

TheCycoONE:
lua_Integer rather

TheCycoONE:
nevermind, size_t is always unsigned. So I'm ok with the first change. The lua ones I'd still prefer to be lua_Integer until they're range checked.

TheCycoONE:
And for my own sanity:
"There is a potential problem with the size_t type and versions of GCC prior to release 2.4. ANSI C requires that size_t always be an unsigned type. For compatibility with existing systems' header files, GCC defines size_t in stddef.h to be whatever type the system's sys/types.h defines it to be. Most Unix systems that define size_t in sys/types.h, define it to be a signed type. Some code in the library depends on size_t being an unsigned type, and will not work correctly if it is signed"

I knew I'd heard of it being signed before ... looks like that is no longer an issue though.

tobylane:
Clang tidy answered my thought about how long lines can be.

TheCycoONE:
Ah, that's not where I meant: "set_animation and get_sprite_size_unchecked both expect size_t."

Thanks for working on this @tobylane

TheCycoONE:
LGTM, but we exchanged warnings in one compiler for another - much fun with types:

From Visual Studio:
C:\projects\corsixth\CorsixTH\Src\th_lua_anims.cpp(320): warning C4018: '>=': signed/unsigned mismatch [C:\projects\corsixth\CorsixTH\CorsixTH_lib.vcxproj]
  th_lua_gfx.cpp
C:\projects\corsixth\CorsixTH\Src\th_lua_gfx.cpp(160): warning C4018: '>=': signed/unsigned mismatch [C:\projects\corsixth\CorsixTH\CorsixTH_lib.vcxproj]

tobylane:
Clang warns about comparing unsigned size_t with get\_\*\_count, while VS warns about comparing signed lua_integer with the same get\_\*\_count? Are they doing something different with the lua type or the count?

TheCycoONE:
> Clang warns about comparing unsigned size_t with get_*_count, while VS warns about comparing signed lua_integer with the same get_*_count? Are they doing something different with the lua type or the count?

The clang warning was about the < 0 comparison which doesn't make sense for unsigned values. The Microsoft warning is about comparing unsigned values with signed values; which may promote the signed value to unsigned and break (because C++ inherited C's terrible integer promotion rules)

Now the value we can get from lua could be any lua_Integer (certainly less than 0); if we don't trust the lua layer (and if we trusted it why would we validate at all).

lua_Integer should be big enough to express any valid sprite / anim index; so bigger than get_*_count - so we want to cast get_*_count to lua_Integer as well.

I'd feel better if we were using narrow from the GSL for some of these casts; or asserting some of these size assumptions another way, but for now I guess just static_cast the counts.

tobylane:
Thanks. There's fewer warnings for using size_t in the comparison.

tobylane:
Thanks for the help and sorry for that spam.

TheCycoONE:
Thanks

PR #1577, opened 2020-01-21
Convert common source into libraries

TheCycoONE:


AnimView/th.cpp
```diff
@@ -34,7 +34,7 @@ SOFTWARE.
 #include <stdexcept>
 #include <vector>
 
-#include "../common/rnc.h"
+#include "../libs/rnc/rnc.h"
```
Alberth289346: Wouldn't it be nicer if source code doesn't have full paths to other code?

That is, do `#include "rnc/rnc.h"` here, and extend the include search path of the compiler with `-I../libs`. In that way, source code doesn't change if you move libraries or other code around in the project tree.

TheCycoONE: Yes, I might be able to do that.

TheCycoONE:
> Works for me. On Ubuntu 16.04 it needs CXXFLAGS='-std=c++11'.
> 👍

Oh, I'm missing the compile flags. I'll fix that.

TheCycoONE:
Fixed the target standard version for @tobylane 's issue.

Fiddled with `target_include_directories` a bit but in the end decided I prefer the #include point to the real path. It's a little nicer as is for text editors that aren't aware of the project but can follow paths.

TheCycoONE:
On the include path, I'm open to be persuaded otherwise.

Alberth289346:
Good point about the editors. In the end it doesn't matter much, it's not likely one of the parts will ever get moved without the other part.

PR #1581, opened 2020-02-05
Codespell changes

tobylane:
Made from this command `codespell --enable-colors --quiet-level 2 --skip="languages,world.lua,corsix-th.6" -L sav,unexpect,persistance,uint CorsixTH`, and some manual edits. Nearly all of it is the documentation comments. There is a false positive on currentY in CorsixTH/Lua/world.lua:2630 so the file is ignored.


Alberth289346:
There is a number of "iff" to "if" changes, which is not a simple type, "iff" means "if and only if"

TheCycoONE:
In every case I saw I consider the (and only if) redundant, so I'm ok with it.

Issue #1569, opened 2019-11-29
Linux portable build


legluondunet:
Hello,

I would like to compile a portable Corsix-TH build, is it possible to compile dependencies statically?
I didn't find documentation about that.
Thank you for your help.

TheCycoONE:
It is certainly possible, largely a matter of linking static libraries instead of dynamic - which means you'll likelt be compiling all the dependencies yourself. What OS?

legluondunet:
Linux.

legluondunet:
I can join libraries files and then use LD_LIBRARY_PATH ,this works but...I didn't find a way for LUA, do you know how to make this depency portable?

TheCycoONE:
liblua.a should be the same. The lua flles themselves are data and cannot currently be built in. The lua executable dependency is only for build time so it doesn't matter.

TheCycoONE:
lua_socket you won't need but you will need lfs. It can be anywhere in the lua cpath, so you could include it with the executable and still be 'portable' unless you are going for single file.

TheCycoONE:
-DCORSIX_TH_INTERPRETER_PATH will also need to be set to the relative path to CorsixTH.lua

legluondunet:
if I send you a package, do you have a Linux partition to test it? If so, which distribution?

TheCycoONE:
All of my existing linux boxes and vms have the dependencies installed already. I could spin up a fresh vm, but so could you.

Are you suggesting to add this to the releases? If so, let me know what glibc version and architecture you are targeting and I will verify it.

legluondunet:
I need help, I'm stuck here:


``
$ ./corsix-th.sh 
le répertoire courant est /home/legluondunet/Games/theme-hospital/Corsix-TH
An error has occurred in CorsixTH:
./Lua/strict.lua:66: module 'lfs' not found:
	no field package.preload['lfs']
	no file './Lua/lfs.lua'
	no file './Lua/lfs/init.lua'
	no file '/usr/local/share/lua/5.2/lfs.lua'
	no file '/usr/local/share/lua/5.2/lfs/init.lua'
	no file '/usr/local/lib/lua/5.2/lfs.lua'
	no file '/usr/local/lib/lua/5.2/lfs/init.lua'
	no file '/usr/share/lua/5.2/lfs.lua'
	no file '/usr/share/lua/5.2/lfs/init.lua'
	no file './lfs.lua'
	no file './lfs.so'
[...]

PR #1588, opened 2020-03-10
Search more locations for interpreter

yangfl:
Currently Corsix-TH only uses hardcoded `CORSIX_TH_INTERPRETER_PATH` if no `--interpreter` option passed. Search more locations (including the working directory and the program directory where corsix-th exists) for `CORSIX_TH_INTERPRETER_NAME` (i.e. CorsixTH.lua).

Partially address #1569 so the working directory does not have to be the program directory. Please leave a comment if I misunderstand the coding style for this project.


TheCycoONE:
I like that this should make the Apple app files run anywhere, and may assist with portable builds on linux.

I do not like that it includes a full unix path before the interpreter_path for unix. This prevents having a separate developer version with USE_SOURCE_DATADIRS. I think that should go away, and only subdirectories of the executable be considered outside of the specified paths.

I wonder what this does for platforms that package / use CorsixTH but are not supported by that library. In particular OpenBSD has long packaged CorsixTH.

I suggest making the change compile time optional and on by default for Apple, but off everywhere else.

The code is reasonable, though calling a function with multiple return points instead of a do {} while(false) with multiple break points is probably warranted with the size and nested loops involved.

yangfl:
The library https://github.com/gpakosz/whereami supports virtually all platforms you could imagine. But I also add check for empty string.

TheCycoONE:
https://github.com/gpakosz/whereami/issues/13

colinjmatt:
I can pull a test build together for this tomorrow evening (UK time) to test these changes work as intended.

CorsixTH/Src/main.cpp
```diff
@@ -52,6 +61,59 @@ inline void preload_lua_package(lua_State* L, const char* name,
...
+      size_t iLen;
+      const char* sCmd = lua_tolstring(L, i, &iLen);
+      if (iLen > 14 && std::memcmp(sCmd, "--interpreter=", 14) == 0)
+        return sCmd + 14;
+    }
+  }
+
+#ifdef CORSIX_SEARCH_LOCAL_DATADIRS
+  // 2. Find CorsixTH.lua in working dir and program dir
+  static const char* const asSearchDirs[] = {
```
TheCycoONE: Prefer std containers such as std::array

TheCycoONE:
> I can pull a test build together for this tomorrow evening (UK time) to test these changes work as intended.

Thanks @colinjmatt

TheCycoONE:
Ok, the new error message is expected based on the patch, but I would have hoped that the it was able to find the resources. @yangfl do you have anything to go on for fixing that?

yangfl:
I don't know anything about mac since I only use Linux... but you may fix it by editing `asSearchDirs` (`"Contents/Resources/"` is just my wild guess)

TheCycoONE:
> I don't know anything about mac since I only use Linux... but you may fix it by editing `asSearchDirs` (`"Contents/Resources/"` is just my wild guess)

It probably is that but relative to the ModulePath not the ExecutablePath, from the ExecutablePath you would have to go down a directory.

It's worth noting that the original linux use case of having a portable linux version was solved by using a relative path for the `CORSIX_TH_INTERPRETER_PATH`

yangfl:
Could you test this example https://github.com/gpakosz/whereami/blob/master/example/executable.c and see where is `executable path`?

tobylane:
> $ /Applications/executable ; exit;
executable path: /Applications/executable
  dirname: /Applications
  basename: executable
module path: /Applications/executable
  dirname: /Applications
  basename: executable

That works perfectly.

`  // 2. Find CorsixTH.lua in working dir and program dir
  static constexpr std::array<const char*, 4> asSearchDirs{
      "./",
      "CorsixTH/",
      "Contents/Resources/",
      "../Resources/",
  };`

With ../Resources this PR works for me.

lewri:
@yangfl Any news on this?

yangfl:
Fixed https://github.com/CorsixTH/CorsixTH/pull/1588#issuecomment-638014540

lewri:
@colinjmatt you able to test this one again?

TheCycoONE:
Pinging @tobylane / @colinjmatt

CorsixTH/Src/main.cpp
```diff
@@ -52,6 +65,60 @@ inline void preload_lua_package(lua_State* L, const char* name,
...
+      size_t iLen;
+      const char* sCmd = lua_tolstring(L, i, &iLen);
+      if (iLen > 14 && std::memcmp(sCmd, "--interpreter=", 14) == 0)
+        return sCmd + 14;
+    }
+  }
+
+#ifdef CORSIX_SEARCH_LOCAL_DATADIRS
+  // 2. Find CorsixTH.lua in working dir and program dir
+  static constexpr std::array<const char*, 3> asSearchDirs{
```
tobylane, line 82:
```suggestion
  static constexpr std::array<const char*, 4> asSearchDirs{
```
With that, approved.

TheCycoONE: @tobylane have you tested it?

tobylane: Yes, I can move it and run it.

CorsixTH/CMakeLists.txt
```diff
@@ -14,18 +14,23 @@ if(MSVC)
   add_definitions(-D_CRT_SECURE_NO_WARNINGS)
 endif()
 
+if(SEARCH_LOCAL_DATADIRS)
+  add_definitions(/DCORSIX_SEARCH_LOCAL_DATADIRS)
```
TheCycoONE, line 18: `-D` for consistency, or better yet:
`set(CORSIX_TH_SEARCH_LOCAL_DATADIRS ${SEARCH_LOCAL_DATADIRS)` here
`#cmakedefine CORSIX_TH_SEARCH_LOCAL_DATADIRS` in config.h.in

Note `CORSIX_TH_` not `CORSIX_` as the prefix, matching all other defines.

CorsixTH/Src/main.cpp
```diff
@@ -22,17 +22,30 @@ SOFTWARE.
...
+#endif
+
 #include "iso_fs.h"
 #include "lua.hpp"
 #include "lua_rnc.h"
 #include "lua_sdl.h"
 #include "persist_lua.h"
 #include "th_lua.h"
 
+#ifdef CORSIX_SEARCH_LOCAL_DATADIRS
```
TheCycoONE, line 45: `CORSIX_TH_SEARCH_LOCAL_DATADIRS` per name change recommended above

TheCycoONE (commented):
Final nits, promise!

TheCycoONE:
@yangfl ^ build checks are failing, typo in cmake.

CorsixTH/CMakeLists.txt
```diff
@@ -14,18 +14,21 @@ if(MSVC)
   add_definitions(-D_CRT_SECURE_NO_WARNINGS)
 endif()
 
+set(CORSIX_TH_SEARCH_LOCAL_DATADIRS ${SEARCH_LOCAL_DATADIRS)
```
TheCycoONE, line 17: missing closing }

yangfl: typo from your comment :( nevertheless I'll fix it later (busy now

lewri:
Does this PR have enough work to warrant putting your name in the copyright of ``main.cpp`` @yangfl ?

yangfl:
It depends on you. I'm ok with either option.

CorsixTH/Src/main.cpp
```diff
@@ -52,6 +65,60 @@ inline void preload_lua_package(lua_State* L, const char* name,
...
+  std::string strProgramDir = "";
+  {
+    int iProgramPathLength = wai_getExecutablePath(nullptr, 0, nullptr);
+    if (iProgramPathLength != 0) {
+      char* sProgramDir = new char[iProgramPathLength + 1];
+      int iProgramDirLength;
+      wai_getExecutablePath(sProgramDir, iProgramPathLength + 1,
+                            &iProgramDirLength);
+      sProgramDir[iProgramDirLength] = '/';
+      sProgramDir[iProgramDirLength + 1] = '\0';
```
Alberth289346, line 97: This writes out of bounds, `new char[iProgramPathLength + 1]` runs from `0` to `iProgramDirLength` inclusive.

lewri:
Looks like you got Appveyor errors. Unfortunately not ignorable this time :(

yangfl:
windows, hmm no idea at all :(

lewri:
Wonderful, looks fixed. Do you feel it's ready now?
As for the copyright comment, it is actually your preference almost exclusively. You're mentioned inside the general license.txt file though regardless.

yangfl:
Yes, of course!

CorsixTH/Src/main.cpp
```diff
@@ -22,17 +22,30 @@ SOFTWARE.
 
 #include "config.h"
 
+#include <array>
 #include <cstdio>
 #include <cstring>
 #include <string>
 
+#ifdef _WIN32
```
TheCycoONE, line 30: I don't love the idea of depending on (and assuming POSIX) everywhere outside of Windows.

We don't have C++ 17's `std::filesystem::exists` but we can detect what we really care about, whether we can read a file pretty easily with ifstream.

```
namespace {
bool file_exists(std::string f) {
  std::ifstream file(f);
  return file.is_open();
}
}
```

CorsixTH/Src/main.cpp
```diff
@@ -52,6 +65,77 @@ inline void preload_lua_package(lua_State* L, const char* name,
...
+      "../Resources/",
+  };
+  std::string strProgramDir = "";
+  {
+    int iProgramPathLength = wai_getExecutablePath(nullptr, 0, nullptr);
+    if (iProgramPathLength != 0) {
+      char* sProgramDir = new char[iProgramPathLength + 1];
+      int iProgramDirLength;
+      int iProgramPathLengthReal = wai_getExecutablePath(
+          sProgramDir, iProgramPathLength + 1, &iProgramDirLength);
```
TheCycoONE, line 95: whereami's documentation doesn't put a + 1 here. It might be safe because we don't use sProgramDir before injecting a null after iProgramDirLength, but there shouldn't be any reason to do it either.

CorsixTH/Src/main.cpp
```diff
@@ -83,28 +167,18 @@ int lua_main_no_eval(lua_State* L) {
...
-    lua_pushstring(L, CORSIX_TH_INTERPRETER_PATH);
+  auto scriptFilePath = search_script_file(L);
+  if (scriptFilePath.empty()) {
+    std::fprintf(stderr,
+                 "CorsixTH cannot find CorsixTH.lua. If you want use a custom "
+                 "location, specify it by --interpreter=FILE\n");
+    std::fflush(stderr);
+    exit(1);
   }
+  lua_pushstring(L, scriptFilePath.c_str());
```
TheCycoONE, line 181: The lua c API all translates to `assert(loadfile(scriptFIlePath))`  which is easier to read when it's in one spot. Move the lua_getglobal lines down with the lua_pushstring.

CorsixTH/Src/main.cpp
TheCycoONE: My comment was meant as a quickly written reference rather than an exact template.

Take a reference instead of a value (std::string&), also remove the .cstr()'s where you're calling this function.

CorsixTH/Src/main.cpp
```diff
@@ -52,6 +58,83 @@ inline void preload_lua_package(lua_State* L, const char* name,
       fn);
 }
 
+// relace me with C++17 std::filesystem::exists
+inline bool file_exists(std::string f) {
```
TheCycoONE, line 62: `std::string& f`

CorsixTH/Src/main.cpp
```diff
@@ -52,6 +58,83 @@ inline void preload_lua_package(lua_State* L, const char* name,
...
+      sProgramDir[iProgramDirLength] = '/';
+      sProgramDir[iProgramDirLength + 1] = '\0';
+      strProgramDir = sProgramDir;
+      delete[] sProgramDir;
+    }
+  }
+  for (auto sSearchDir : asSearchDirs) {
+    std::string strPathInWorkingDir =
+        std::string(sSearchDir) + CORSIX_TH_INTERPRETER_NAME;
+    if (file_exists(strPathInWorkingDir.c_str()))
```
TheCycoONE, line 121: remove c_str(), takes a string reference.

CorsixTH/Src/main.cpp
```diff
@@ -52,6 +58,83 @@ inline void preload_lua_package(lua_State* L, const char* name,
...
+      return strPathInWorkingDir;
+    if (!strProgramDir.empty()) {
+      std::string strPathInProgramDir = strProgramDir + strPathInWorkingDir;
+      if (file_exists(strPathInProgramDir.c_str()))
```
TheCycoONE, line 125: remove c_str()

CorsixTH/Src/main.cpp
```diff
@@ -52,6 +58,83 @@ inline void preload_lua_package(lua_State* L, const char* name,
...
+        return strPathInProgramDir;
+    }
+  }
+#endif
+
+  // 3. Check CORSIX_TH_INTERPRETER_PATH
+  if (file_exists(CORSIX_TH_INTERPRETER_PATH))
```
TheCycoONE, line 132: probably needs converted into a std::string?

TheCycoONE (changes requested):

CorsixTH/Src/main.cpp
yangfl: probably no?

TheCycoONE:
Nice. Ready for testing?

TheCycoONE (approved):

lewri:
@yangfl possible to remove the WIP label now I guess?

yangfl:
ready for merge now

PR #1589, opened 2020-03-11
Fix CJK rendering issue

yangfl:


CorsixTH/Src/th_gfx_font.cpp
```diff
@@ -615,7 +627,12 @@ text_layout freetype_font::draw_text_wrapped(render_target* pCanvas,
             iHandledRows++;
           }
           sMessage = sLineBreakPosition;
-          next_utf8_codepoint(sMessage);
+          // Skip one char only when it's .. (maybe a CJK character)
+          if (std::iswspace(iCode) ||  // Whitespace characters
+              iCode == '/' ||          // Line break marker
```
TheCycoONE: Do we have `/` line breaks still? There was an earlier commit where we removed them all. It potentially conflicts with other locales. Everywhere should be using \n.

yangfl: Strings from Chinese original datapack still have / line breaks. I don't pretty sure, though.

TheCycoONE: So - we use `/` frequently where it is not a line break. Where we use it as a line break we seem to do `//`

TheCycoONE: e.g. in spanish.lua
```
618:    [1] = "Diagnosis general//",
641:tooltip.research_policy.research_progress = "Progreso del siguiente descubrimiento de esta categoría: %1%/%2%"
```

TheCycoONE: Can you try without adding the / at all? I think that is already handled. We use it heavily in other languages.

yangfl: Hmm let's just delete it for now, but it will render Chinese texts like this

slash-in-chinese

PR #1612, opened 2020-05-21
Update changelog and LICENSE file

TheCycoONE:
We should be doing this as we go, but since no one does ....


TheCycoONE:
@mugmuggy please suggest any fixes/changes I may have missed. I know you resolved a lot of issues.

mugmuggy:
That list is ok with me

PR #1617, opened 2020-05-23
[RDY] Fix for #1615

### LINKED ISSUE #1615: Suggestion: Truncate decimal .0 from Annual Report
I've noticed in the annual report that the player's salary, all 'richest', and all total value figures are reported as xxxxx.0
In reality and in line with the game only using whole numbers for money, the .0 should be truncated. This is still present in the current version 0.63.

I would have tried to look at changing this myself in the relevant file (either using math.floor or converting to text) but to be honest I'm not wholly confident in tackling it.

image

lewri:
Adds math.floor to values used in the Annual Report that would be given .0 at the end of their value.

Closes #1615


lewri:
Screenshot just to show it is working
image

PR #1620, opened 2020-05-24
[RDY] Prevent vaccination actions outside vaccination mode

### LINKED ISSUE #1619: Vaccination graphic active outside of vaccination mode
#### Describe the issue
When you are in an epidemic, treating yourself, you do not need to enable vaccination mode to get the vaccination graphics. As a result, this may cause some confusion as clicking on a patient where vaccination mode is off, still triggers the vaccination graphic. It also changes the graphic on the patient to say they are marked for vaccination
It also does not open the patient dialog on first click on a contagious patient.

The only thing that doesn't happen is make a vaccine call to a nurse.

#### Steps to Reproduce

2. Accept to deal with the epidemic yourself

3. Find the contagious patient about to enter psychiatric (on way to be cured so may need to be quick), click on the patient without activating vaccination mode.

#### Save Game
https://www.dropbox.com/s/b42ujhtjqbzvwh2/epidemic.sav?dl=0

#### Expected Behavior
In the original, the vaccination action/graphics were reserved exclusively for vaccination mode.

#### System Information
**CorsixTH Version:** 0.63

**Operating System:** W10

#### Gamelog.txt
None.

### Additional Info
I checked through the previous issues mentioned vaccination from 2015 and 2017. They do mention a behaviour issue but I don't think this one was cleared up (or was a resulting error from those fixes).

I can only assume that when you are dealing with an epidemic, the normal graphics/action for patients are being overriden for contagious patients. I added a debug line to the function "Epidemic:markForVaccination(patient)" in epidemic.lua and it looks like it is definitely being called outside the vaccination mode.

lewri:
Fixes #1619

Actions inside function Epidemic:markForVaccination(patient) were getting triggered outside of vaccination mode. Causes a graphic update for contagious patients without making a call for the nurse to vaccinate.

Function now checks if vaccination mode is on before executing. Tested and working.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -66,17 +66,10 @@ function Patient:onClick(ui, button)
     else
       local hospital = self.hospital or self.world:getLocalPlayerHospital()
       local epidemic = hospital and hospital.epidemic
-      if not epidemic or
-          (not epidemic.coverup_in_progress or
-            (not self.infected or self.marked_for_vaccination) and
-            not epidemic.vaccination_mode_active) then
-        ui:addWindow(UIPatient(ui, self))
-      end
-      if epidemic and epidemic.coverup_in_progress and
-          self.infected and not self.marked_for_vaccination and
-          -- Prevent further vaccinations when the timer ends
-          not epidemic.timer.closed and epidemic.vaccination_mode_active then
+      if epidemic and epidemic.vaccination_mode_active then
         epidemic:markForVaccination(self)
+      else
+        ui:addWindow(UIPatient(ui, self))
       end
     end
   elseif self.user_of then
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -75,7 +75,7 @@ function Patient:onClick(ui, button)
       if epidemic and epidemic.coverup_in_progress and
           self.infected and not self.marked_for_vaccination and
           -- Prevent further vaccinations when the timer ends
-          not epidemic.timer.closed then
+          not epidemic.timer.closed and epidemic.vaccination_mode_active then
         epidemic:markForVaccination(self)
       end
     end
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -293,7 +293,7 @@ end
  the player and can only happen during a cover up. ]]
 function Epidemic:markForVaccination(patient)
   if patient.infected and not patient.vaccinated and
-      not patient.marked_for_vaccination and self.vaccination_mode_active then
+      not patient.marked_for_vaccination then
     patient.marked_for_vaccination = true
     patient:setMood("epidemy4","deactivate")
     patient:setMood("epidemy2","activate")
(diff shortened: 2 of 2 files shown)


TheCycoONE:
Looks right to me. @mugmuggy ok to you too?

lewri:
I've changed this to WIP as I noticed something else that bothered me about the vaccination function:

I noticed that when a nurse gets assigned the call to vaccinate a patient, their dynamic info stays as 'just wandering around'. I would like to also push an improvement that updates their dynamic info to "Vaccinating a patient" rather than the aforementioned one.

mugmuggy:
My only consideration would be putting the check in Patient:onClick, as its already used in the patient window decision right above the markForVaccination call.

lewri:
@mugmuggy moved the check as requested. Please could you make sure it's in the place intended.
Did test and worked the same, at least.

mugmuggy:
There is the possibility to simplify some more, the epidemic.timer upon closing in Epidemic:finishCoverUp(), will toggle off the vaccination_mode_active if it is set and the timer window will be closed, so we probably don't need the epidemic.timer.closed check either. 

Likewise the timer won't exist, to toggle that mode through the UIWatch until its created and this is the same point that coverup_in_progress is set to true.

ie can we get to this?
```lua
      if epidemic and epidemic.vaccination_mode_active and
          self.infected then
        epidemic:markForVaccination(self)
      else
        ui:addWindow(UIPatient(ui, self))
      end

and the 'not self.marked_for_vaccination' gets moved to markForVaccination (as it currently does anyway) and this prevents the patient window loading then in vaccination mode (click once to mark for vaccination, click again brings up patient window), and also preventing non-infected patients and staff dialogs from showing either. 
or would this be inconsistent with TH?

Game_UI already does a check on vaccination_mode_active and the patients infected state, thus it would appear not to need the self.infected check in patient:onClick either and we are basically checking which mode we are in at this point without worrying about the rest of the state.

**Comment by lewri** [comment]
Unfortunately that didn't work, and we get this error:

```
An error has occurred!
Lua\entities\humanoids\patient.lua:68: attempt to call a table value (field 'epidemic')
stack traceback:
        Lua\entities\humanoids\patient.lua:68: in method 'onClick'
        Lua\ui.lua:906: in function <Lua\ui.lua:889>
        (...tail calls...)

Error in buttonup handler:
Lua\entities\humanoids\patient.lua:68: attempt to call a table value (field 'epidemic')
stack traceback:
        Lua\entities\humanoids\patient.lua:68: in method 'onClick'
        Lua\ui.lua:906: in function <Lua\ui.lua:889>
        (...tail calls...)

```

So for now I'll leave it as is.

mugmuggy:
The whole function as it is, working for me:
```lua
function Patient:onClick(ui, button)
  if button == "left" then
    if self.message_callback then
      self:message_callback()
    else
      local hospital = self.hospital or self.world:getLocalPlayerHospital()
      local epidemic = hospital and hospital.epidemic
      if epidemic and epidemic.vaccination_mode_active then
        epidemic:markForVaccination(self)
      else
        ui:addWindow(UIPatient(ui, self))
      end
    end
  elseif self.user_of then
    -- The object we're using is made invisible, as the animation contains both
    -- the humanoid and the object. Hence send the click onto the object.
    self.user_of:onClick(ui, button)
  end
  Humanoid.onClick(self, ui, button)
end

lewri:
Sorry @mugmuggy it does work. Oversight on my part.
I was testing inside 0.63 rather than from the master.

Tested now and it works fine.

PR #1621, opened 2020-05-24
[RDY] Suggestion - Give nurse dynamic info when vaccinating

lewri:
Addition to #1620 that adds an improvement for when the nurse is assigned a patient to vaccinate.

Currently, the nurse has the dynamic info "Just wandering around" when going to vaccinate a patient.
This PR would change her dynamic info to "Vaccinating a patient" in that time frame.

Dyanmic info resets back to normal once vaccination is done.

### DIFF
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -260,6 +260,7 @@ dynamic_info.patient.actions.no_gp_available = "Waiting for you to build a GP's 
 dynamic_info.staff.actions.heading_for = "Heading for %s"
 dynamic_info.staff.actions.fired = "Fired"
 dynamic_info.patient.actions.epidemic_vaccinated = "I am no longer contagious"
+dynamic_info.patient.actions.vaccine = "Vaccinating a patient"
 
 progress_report.free_build = "FREE BUILD"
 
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/languages/english.lua
```diff
@@ -260,6 +260,7 @@ dynamic_info.patient.actions.no_gp_available = "Waiting for you to build a GP's
 dynamic_info.staff.actions.heading_for = "Heading for %s"
 dynamic_info.staff.actions.fired = "Fired"
 dynamic_info.patient.actions.epidemic_vaccinated = "I am no longer contagious"
+dynamic_info.patient.actions.vaccine = "Vaccinating a patient"
```
mugmuggy, line 263: staff not patient

lewri: Good spot, thanks!

TheCycoONE:
Does this get cleared properly after the patient is vaccinated (or the action is interrupted)?

lewri:
I'm not entirely sure. And I wouldn't know what to really check.
All I can tell you is after the patient is vaccinated it goes back to 'Just wandering around'.
I can also tell you if the nurse is mid-action and the patient starts moving, the "giveVaccinationCandidateStatus" is stopped and goes back to 'Just wandering around'.

If that's not what you're asking I'm not sure, sorry.

PR #1624, opened 2020-05-30
Handle current VCPKG and Visual Studio versions

TheCycoONE:
SDL_mixer now requires dynamic_load unless building statically, and no
longer brings over the dynamic libraries it depends on automatically.

Visual Studio 2019 defaults to x64 instead of x86.

Vcpkg is available on more versions of Visual Studio, as well as Linux
and MacOS these days, though our CMake is still not compatible with
using it on non-windows platforms yet.


TheCycoONE:
These are changes I'm using to generate the Windows builds for 0.64

PR #1632, opened 2020-06-02
[RDY] 0.64 - Add ISO References

lewri:
Updating CorsixTH's TH install dialogs to match that ISO files can now be used.

**One of the updates has caused a graphical glitch (see below), not sure how to fix**
image

Needs fixing before can be RDY

### DIFF
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -480,7 +480,7 @@ folders_window = {
   savegames_label = "Saves",
   screenshots_label = "Screenshots",
   -- next four are the captions for the browser window, which are called from the folder setting menu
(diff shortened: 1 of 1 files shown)


TheCycoONE:
There's a fix in this release concerning the size of the confirm quit dialog that was a similar issue. The size of most of the UI elements are fixed.

The quick fix of course would be to drop that last note, even though it would be helpful.

The proper fix would be to ask for the size of the text up front like I do here: https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/dialogs/confirm_dialog.lua#L52

and then adjust the top (and height) of the TreeControl to account for it here: https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/dialogs/resizables/directory_browser.lua#L214

lewri:
@TheCycoONE as fixing that seems a bit too confusing for me.... I've updated the text so it now fits.

Hopefully that's okay with you.

PR #1648, opened 2020-06-16
[RDY] Enhancement - Extinguishers reduce chance of machines exploding

### LINKED ISSUE #994: Fire extinguishers don't impact explosion chance
#### Describe the issue

Fire extinguishers currently impact happiness but don't impact equipment explosion chance.

From the game manual:
"✚ Fire Extinguishers are looked for by Very Important People - VIPs. VIPs like being safe. If
they don’t see a reassuring number of Fire Extinguishers when they visit, their reports to
the Ministry won’t be very favorable. Fire Extinguishers also increase safety levels around
Hospital equipment, and their presence makes Patients and Staff alike happier."
#### Expected Behavior

Fire extinguishers should lower the chance that a piece of equipment explodes either by increasing it's strength or implementing a % chance that when the threshold is hit the equipment doesn't explode.
#### System Information

**CorsixTH Version:** 0.50

**Operating System:** any

**Theme Hospital Version:** any

lewri:
Would resolve #994 

PR will implement a variable saved_by_extinguisher to each machine/object defaulted at 0.
When machine threshold falls <1, the variable is recalculated by math.random(0,1) to see if an extinguisher in the room will save the room (saved_by_extinguisher == 1 is a success). Variable is saved until the machine is next used (inc. earthquake tick), in which case it sets back to 0.

The addition of self.saved_by_extinguisher may not be needed in object.lua... it was added more as a safeguard.

As for machine_dialog.lua nothing needs to be changed. The health marker in the dialogue is not affected if the number of times a machine is used is higher than its strength.

### DIFF
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -181,7 +181,7 @@ function Machine:calculateSmoke(room)
     -- Display smoke, up to three animations per machine
     -- i.e. < 4 one plume, < 3 two plumes or < 2 three plumes of smoke
     setSmoke(self, true)
-    -- turn on additional ldayers of the animation for extra smoke plumes, depending on how damaged the machine is
+    -- turn on additional layers of the animation for extra smoke plumes, depending on how damaged the machine is
     if threshold < 3 then
       self.smokeInfo:setLayer(11, 2)
     end
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -106,7 +106,7 @@ function Machine:machineUsed(room)
   if threshold < 1 then
     -- If a fire extinguisher in the room, room has chance not to explode
     for object, _ in pairs(room.objects) do
-      if object.object_type.id == "extinguisher" then
+      if object.object_type.id == "extinguisher" and num_extinguishers < 4 then
         num_extinguishers = num_extinguishers + 1
       end
     end
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -20,8 +20,6 @@ SOFTWARE. --]]
 
 local TH = require("TH")
 
-corsixth.require("announcer")
-local AnnouncementPriority = _G["AnnouncementPriority"]
 --! An `Object` which needs occasional repair (to prevent explosion).
 class "Machine" (Object)
 
@@ -102,7 +100,7 @@ function Machine:machineUsed(room)
   -- Find a queued task for a handyman coming to repair this machine
   local taskIndex = self.hospital:getIndexOfTask(self.tile_x, self.tile_y, "repairing")
   local num_extinguishers = 0
-  local explosion_chance = 1
+  local explosion_chance
   local explode = false
   -- Room is set to explode
   if threshold < 1 then
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -20,6 +20,8 @@ SOFTWARE. --]]
 
 local TH = require("TH")
 
+corsixth.require("announcer")
+local AnnouncementPriority = _G["AnnouncementPriority"]
 --! An `Object` which needs occasional repair (to prevent explosion).
 class "Machine" (Object)
 
@@ -99,19 +101,25 @@ function Machine:machineUsed(room)
   local threshold = self:getRemainingUses()
   -- Find a queued task for a handyman coming to repair this machine
   local taskIndex = self.hospital:getIndexOfTask(self.tile_x, self.tile_y, "repairing")
-  -- extinguisher check must be reset after each use
-  local saved_by_extinguisher = 0
+  local num_extinguishers = 0
+  local explosion_chance = 1
   local explode = false
   -- Room is set to explode
   if threshold < 1 then
-    -- If a fire extinguisher in the room, make explosion chance 50%
+    -- If a fire extinguisher in the room, room has chance not to explode
     for object, _ in pairs(room.objects) do
       if object.object_type.id == "extinguisher" then
-        saved_by_extinguisher = math.random(0,1)
-        break
+        num_extinguishers = num_extinguishers + 1
       end
     end
-    explode = saved_by_extinguisher == 0
+    if num_extinguishers == 0 or threshold < -3 then
+      -- If no extinguisher in room, or machine used 5 times over its strength always explode
+      explode = true
+    else
+      -- Explosion chance increaes 20% with every use over strength, and reduced by 5% for every additional extinguisher in the room bar the first one
+      explosion_chance = 0.25 + (threshold * -0.2) - (num_extinguishers * 0.05)
+      explode = math.random() < explosion_chance
+    end
   end
   -- Fire extinguisher failed to save the room or none were present
   if explode then
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -112,7 +112,6 @@ function Machine:machineUsed(room)
     end
     -- Extinguisher failed to save the room
     if self.saved_by_extinguisher == 0 then
-      print("machine lost!")
       -- Clean up any task of handyman coming to repair the machine
       self.hospital:removeHandymanTask(taskIndex, "repairing")
       -- Blow up the room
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/entities/machine.lua
```diff
@@ -154,14 +175,15 @@ function Machine:calculateSmoke(room)
...
+  if threshold < 1 and self.saved_by_extinguisher == 0 then
     setSmoke(self, false)
+    return
   -- Else if urgent repair needed
   elseif threshold < 4 then
     -- Display smoke, up to three animations per machine
     -- i.e. < 4 one plume, < 3 two plumes or < 2 three plumes of smoke
     setSmoke(self, true)
-    -- turn on additional layers of the animation for extra smoke plumes, depending on how damaged the machine is
+    -- turn on additional ldayers of the animation for extra smoke plumes, depending on how damaged the machine is
```
TheCycoONE: oops

CorsixTH/Lua/entities/machine.lua
```diff
@@ -99,29 +99,50 @@ function Machine:machineUsed(room)
...
+      local window = self.world.ui:getWindow(UIMachine)
+      if window and window.machine == self then
+        window:close()
+      end
+      -- Clear the icon showing a handyman is coming to repair the machine
+      self:setRepairing(nil)
+      return true
+    else
+      -- Extinguisher saved room, machine sure a handyman has been called
+      print("machine saved!")
```
TheCycoONE: Remove the print

CorsixTH/Lua/entities/machine.lua
```diff
@@ -99,29 +99,50 @@ function Machine:machineUsed(room)
   local threshold = self:getRemainingUses()
   -- Find a queued task for a handyman coming to repair this machine
   local taskIndex = self.hospital:getIndexOfTask(self.tile_x, self.tile_y, "repairing")
-
-  -- Too late it is about to explode
+  -- extinguisher check must be reset after each use
+  self.saved_by_extinguisher = 0
```
TheCycoONE, line 103: Should be a local

CorsixTH/Lua/entities/machine.lua
```diff
@@ -99,29 +99,50 @@ function Machine:machineUsed(room)
...
+      if window and window.machine == self then
+        window:close()
+      end
+      -- Clear the icon showing a handyman is coming to repair the machine
+      self:setRepairing(nil)
+      return true
+    else
+      -- Extinguisher saved room, machine sure a handyman has been called
+      print("machine saved!")
+      if taskIndex == -1 then
```
TheCycoONE, line 136: Not a fan of repeating this; better to split into two conditions:
```lua
local explode = false
if threshold < 1
  -- check if saved by extinguisher, set explode = true otherwise
end
if explode then
  -- do explode type stuff
elseif threshold < 4
  -- do handyman task priority stuff
elseif theshold < 6
  -- other call stuff
end
```

CorsixTH/Lua/entities/machine.lua
```diff
@@ -154,14 +175,15 @@ function Machine:calculateSmoke(room)
   local threshold = self:getRemainingUses()
 
   -- If now exploding, clear any smoke
-  if threshold < 1 then
+  if threshold < 1 and self.saved_by_extinguisher == 0 then
```
TheCycoONE, line 176: pass explode into this function, or don't call calculateSmoke when explode is true, just setSmoke directly. It's kinda weird how we set smoke here and everything else in the other function.

CorsixTH/Lua/entities/machine.lua
lewri: Think you commented on the wrong commit here?

CorsixTH/Lua/entities/machine.lua
lewri: As above above, wrong commit?

CorsixTH/Lua/entities/machine.lua
lewri: This is to do with a function in room.lua . If we edit the room, the smoke has to be reset.

lewri: In terms of the Machine:machineUsed(room) function it is easier to set the smoke inside there, especially with now checking if an extinguisher saved the room. Nonetheless I've removed an offending bit of the if statement that would never be used in calculateSmoke(room)

TheCycoONE:
I feel that 50% is too generous given how inexpensive and small extinguishers are. Maybe 10-25% Thoughts?

lewri:
I don't have a problem per se reducing percentage but given the catastrophic nature of earthquakes I would say reducing it too much would nullify any effect of this enhancement.

There's something that I just posted in #1649 with mugmuggy that might be of relevance here with this enhancement (if we acted on it at all).

lewri:
What about 33%?

If it survives one use, it's second use puts the chance at ~11%. (Even though that's technically not real maths)

mugmuggy:
This is why I indicated what the original code does as the chance starts out say between 11% and 25% of crashing, to 80-90% and then guaranteed. The stronger a machine is - the more researched it is - the more likely its to survive. Could have something like 20% chance for each use above the strength value and potentially additional extinguishers, reduce that by 5%, maybe or diminishing return for each extinquisher. Obviously have to track additional information to record this.

lewri:
Do we want a calculation of that complexity these days? That's my own concern.
(It's not technically that complex, just would a simple flat value not suffice?)

TheCycoONE:
I like the escalating chance for all machines with a slight reduction for extinguishers. As I said, I don't think the effect of extinguishers should be nearly as pronounced as this PR makes them; and that would also help balance against earthquakes a bit.

lewri:
Working on updating the calculations now. But we might need to tweak some things as it appears even more lenient than before:

```lua
room scanner in danger and has an explosion chance of 0.45
room fracture_clinic in danger and has an explosion chance of 0.05
room x_ray in danger and has an explosion chance of 0.65
room slack_tongue in danger and has an explosion chance of 1
room cardiogram in danger and has an explosion chance of 0.45
room blood_machine in danger and has an explosion chance of 0.05
room inflation in danger and has an explosion chance of 0.05
room inflation in danger and has an explosion chance of 0.05
room decontamination in danger and has an explosion chance of 0.05
room hair_restoration in danger and has an explosion chance of 0.05
room jelly_vat in danger and has an explosion chance of 0.05
room blood_machine in danger and has an explosion chance of 0.25
room scanner in danger and has an explosion chance of 0.65
room fracture_clinic in danger and has an explosion chance of 0.25
room cardiogram in danger and has an explosion chance of 0.65
room blood_machine in danger and has an explosion chance of 0.05
room blood_machine in danger and has an explosion chance of 0.25
room inflation in danger and has an explosion chance of 0.25
room inflation in danger and has an explosion chance of 0.25
room decontamination in danger and has an explosion chance of 0.25
room dna_fixer in danger and has an explosion chance of 0.05
room hair_restoration in danger and has an explosion chance of 0.25
room jelly_vat in danger and has an explosion chance of 0.25
room decontamination in danger and has an explosion chance of 0.05
room dna_fixer in danger and has an explosion chance of 0.05
room dna_fixer in danger and has an explosion chance of 0.05
room dna_fixer in danger and has an explosion chance of 0.05
room blood_machine in danger and has an explosion chance of 0.45
room scanner in danger and has an explosion chance of 1
room fracture_clinic in danger and has an explosion chance of 0.45
room operating_theatre in danger and has an explosion chance of -0.05
room blood_machine in danger and has an explosion chance of 0.25
room blood_machine in danger and has an explosion chance of 0.05
room blood_machine in danger and has an explosion chance of 0.05
room scanner in danger and has an explosion chance of 0.05
room inflation in danger and has an explosion chance of 0.45
```

That was during an earthquake. The number of extinguishers affecting the explosion chance may need to be clamped too.

lewri:
I have implemented the old TH method as above.
I made one change, in that I think an extinguisher must be present to benefit from any chance of a room not exploding. No extinguisher in room = it will explode.

I think we might need to cap the grant from extinguishers in the room; as currently it reduces explosion chance by 5% for every extinguisher in the room. 

Maybe it should be capped at 4, so there is still 5% chance of exploding minimum.

mugmuggy:
That looks like a reasonable compromise between the too harsh now and the original, that was my suggestion and not exactly how TH works - there appears increased surivability chances with TH as strength is used in the calculation. An approximation to that would be to change the 0.25 constant to roughly 2 / strength instead.

lewri:
Sounds like a good consideration to me.

I would also like to ensure that explosion_chance can never get to 0 though. So I would suggest there's a minimum value of 0.05 at all times. And maybe also maxing out at 0.95 until we reach the guaranteed explosion conditions...

lewri:
This is ready for review again

lewri:
Marking as ready as I don't perceive we need any more changes here

PR #1651, opened 2020-06-18
Fix load savegame menu tooltip placements

### LINKED ISSUE #518: Load game window Tooltip-bug: wrong place
_From [maartenp...@hotmail.nl](https://code.google.com/u/103280305005036074093/) on February 02, 2013 06:05:33_

What steps will reproduce the problem? 1.Click on Load Game.
2.Hold you're cursor on ''Game'' or ''Modified.''
3.The tooltips appears on a whole different place as you would espect (more left and above.) What is the expected output? What do you see instead? See steps above. Please use labels and text to provide additional information.

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1484_

tobylane:
Fixes #518
Are the variable names ok?

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/file_browser.lua
+++ b/CorsixTH/Lua/dialogs/resizables/file_browser.lua
@@ -88,12 +88,16 @@ function FilteredTreeControl:FilteredTreeControl(root, x, y, width, height, col_
 
   self.num_rows = (self.tree_rect.h - self.y_offset) / self.row_height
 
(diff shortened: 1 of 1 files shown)


TheCycoONE:
Not the worst.

lewri:
I imagine the 1x and 2x and y refer to axis/position. I think making that more obvious like button_1x, button_2x, button_y is better?

TheCycoONE:
It's really fine as is for UI code.

PR #1652, opened 2020-06-18
Make SDL audio optional

### LINKED ISSUE #143: app.lua is hardcoded to require SDL with audio (undermines -DWITHOUT_AUDIO)
app.lua is hardcoded to require SDL with audio:
108:   if not SDL.init("audio", "video", "timer") then

In addition to undermining the -DWITHOUT_AUDIO feature, it also contradicts the audio.lua check:
56:  if not SDL.audio.loaded then
57:    if not _MAP_EDITOR then
58:      print "Notice: Audio system not loaded as CorsixTH compiled without it"

this causes a generic "Cannot initialise SDL" error, which is traced back to CorsixTH.lua:
119: assert(TheApp:init())

tobylane:
Fixes #143
This adds the first line to this output when SDL2 is compiled without audio.

> SDL_Init failed: SDL not built with audio support
Notice: Audio system not loaded as CorsixTH compiled without it

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -121,10 +121,13 @@ function App:init()
   self:initScreenshotsDir()
 
   -- Create the window
-  if not SDL.init("audio", "video", "timer") then
+  if not SDL.init("video", "timer") then
     return false, "Cannot initialise SDL"
   end
   local compile_opts = TH.GetCompileOptions()
+  if compile_opts.audio then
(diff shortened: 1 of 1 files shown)


TheCycoONE:
We have `compile_ops` here already, we could avoid the failing pcall when we know that corsix is compiled without audio. Mind adding that in? I assume as is you get a console message?

tobylane:
Made the changes.

PR #1660, opened 2020-06-23
[RDY] Drop HOSPITAL directory selection from Installer

TheCycoONE:
Drops InstallOptions in favour of nsDialogs.
Drops the selecton of the install directory.

The intention is to allow the first run selection dialog in the game to run, which provides the ability to select the ISO or a directory. This will hopefully be less confusing for Windows users.


lewri:
There's an odd conflict on this and not sure how it happened but assuming you're keeping this one until the RC.

TheCycoONE:
Going to revive this for 0.65

lewri:
Additional note for changelog - this'll only affect new installs , existing installs will see no change.

PR #1664, opened 2020-06-27
[RDY] Update Issue Templates and PR Templates

lewri:
A suggestion to replace the old issue template file with the new issue templates as provided by Github. Also includes a default PR text.

Issues can now be broken down into:
- Bug reports
- Feature requests

These are automatically labelled up as such, making them clearer in the issues page.

A default PR template was also added to encourage use of the new "fixes xxxx" feature for linking PRs to Issues.


Alberth289346:
If the `title` entry at the top should be edited by the user, it may need some improvement to express that, eg 'Write a one line description of the problem'.

lewri:
I'm tempted to add "Give a" to the front of those titles. But I'd rather not make that line any longer than it needs to be.

Worst case we just go in and update the title manually. There's been reports with stuff like that before and it probably isn't going to stop happening from now on either.

lewri:
Updates are done

PR #1665, opened 2020-07-01
[RDY] Plants have an affect

mugmuggy:
It doesn't appear much as a departure but you get to visually see the need for watering and watering more often.

Whereas without the change, if you have a well staffed hospital, you might not see any transitions with 10 remaining days check.

I've scaled the happiness change, but note this only searches with a distance of 2, so probably largely unaffected in many rooms/corridors as it also only occurs once per day.

Transitions with the temperature change occur within 39-67 days, without the change, it was 49-75 (TH uses constant 50).

setNextState change just gives 5 states (animation frames), instead of 6 as its zero based.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -681,13 +681,8 @@ function Patient:tickDay()
   -- It is nice to see plants, but dead plants make you unhappy
   self.world:findObjectNear(self, "plant", 2, function(x, y)
     local plant = self.world:getObject(x, y, "plant")
-  if not plant then
-    return
-  end
-    if plant:isPleasing() then
-      self:changeAttribute("happiness", 0.0002)
-    else
-      self:changeAttribute("happiness", -0.0002)
+    if plant then
+      self:changeAttribute("happiness", -0.0003 + (plant:isPleasingFactor() * 0.0001))
     end
   end)
   -- It always makes you happy to see you are in safe place
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -62,11 +62,7 @@ function Staff:tickDay()
   self.world:findObjectNear(self, "plant", 2, function(x, y)
     local plant = self.world:getObject(x, y, "plant")
     if plant then
-      if plant:isPleasing() then
-        self:changeAttribute("happiness", 0.002)
-      else
-        self:changeAttribute("happiness", -0.003)
-      end
+      self:changeAttribute("happiness", -0.003 + (plant:isPleasingFactor() * 0.001))
     end
   end)
   -- It always makes you happy to see you are in safe place
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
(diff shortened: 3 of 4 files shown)


CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -691,13 +691,8 @@ function Patient:tickDay()
...
     local plant = self.world:getObject(x, y, "plant")
-  if not plant then
-    return
-  end
-    if plant:isPleasing() then
-      self:changeAttribute("happiness", 0.0002)
-    else
-      self:changeAttribute("happiness", -0.0002)
+    if plant then
+      self:changeAttribute("happiness", -0.0003 + (plant:isPleasingFactor() * 0.0001))
```
Alberth289346, line 695: This goes from `-0.0003` to `-0.0003 + 0.0001 = -0.0002`, ie always decreasing happiness.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -62,11 +62,7 @@ function Staff:tickDay()
   self.world:findObjectNear(self, "plant", 2, function(x, y)
     local plant = self.world:getObject(x, y, "plant")
     if plant then
-      if plant:isPleasing() then
-        self:changeAttribute("happiness", 0.002)
-      else
-        self:changeAttribute("happiness", -0.003)
-      end
+      self:changeAttribute("happiness", -0.003 + (plant:isPleasingFactor() * 0.001))
```
Alberth289346, line 65: Same here, no plants is better!

CorsixTH/Lua/entities/humanoids/patient.lua
TheCycoONE: PleasingFactor is an integer 1 to 5

Alberth289346: Good point.

However, shouldn't the factor be interpolated then? I mean `self.phases - self.current_state` isn't hard-coded 1 to 5, `self.phases` is decided at runtime by parsing the animation. Change the animation, and the plant will have a different effect.

Alberth289346: Unless of course you decide that the length will always be 5, but in that case, it may need an assert or so to verify.

Alberth289346: Interpolation breaks the game if in multiplayer two players use different animations for the plant.

That seems to suggest it needs a fixed number of states, and the plant display should be interpolated?? (or just don't bother and decide 5).

TheCycoONE: @mugmuggy could you fix phases at 5 (or go all the way with supporting a variable number of animations) as per Albert's suggestion.

mugmuggy: Sure I can change the initialisation back to just 5 for the self.phases for now. You have the same problem in a multiplayer game with everything else if any animation sequence is different, as the 2 states would be completely different, but where state doesn't necessarily need to sync, yes there would be issues.

TheCycoONE: Yeah, we'd probably force base game animations to use the same number, and require multiplayer to have the same mods installed. Very out of scope.

lewri:
This all looks fine to me, good to see some variability in plant effects. Guess we're just waiting on phases to be fixed?

mugmuggy:
I've fixed the phases value to 5 and readjusted the calculation on the days back to what is was before but have the watering on a slightly different variance

PR #1667, opened 2020-07-16
Add a flag to allow builds that block off areas.

Alberth289346:
Addresses #826 (first meta issue) first step: "Implement debug flag to allow placing objects that would block off a portion of the map."

- Add a debug config setting "Allow blocking off areas"
- Use the flag to allow building rooms and placing objects even if they block off some area in the hospital or room.

Screenshot_20200716_144635

### DIFF
--- a/CorsixTH/Lua/dialogs/place_objects.lua
+++ b/CorsixTH/Lua/dialogs/place_objects.lua
@@ -697,7 +697,13 @@ function UIPlaceObjects:setBlueprintCell(x, y)
     end
     if self.object_anim and object.class ~= "SideObject" then
       if allgood then
-        allgood = not world:wouldNonSideObjectBreakPathfindingIfSpawnedAt(x, y, object, self.object_orientation, roomId)
+        if world:wouldNonSideObjectBreakPathfindingIfSpawnedAt(x, y, object, self.object_orientation, roomId) then
+          if self.world.app.config.allow_blocking_off_areas then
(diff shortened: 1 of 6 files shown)


TheCycoONE:
We may ultimately decide to keep the check inside the room. This will help us figure that out though, thanks!

Alberth289346:
I didn't do anything special for in-room blocking, it's all the same 'place_objects'. That code would need to be adjusted then.

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -697,7 +697,13 @@ function UIPlaceObjects:setBlueprintCell(x, y)
     end
     if self.object_anim and object.class ~= "SideObject" then
       if allgood then
-        allgood = not world:wouldNonSideObjectBreakPathfindingIfSpawnedAt(x, y, object, self.object_orientation, roomId)
+        if world:wouldNonSideObjectBreakPathfindingIfSpawnedAt(x, y, object, self.object_orientation, roomId) then
+          if self.world.app.config.allow_blocking_off_areas then
```
mugmuggy, line 701: @Alberth289346 I guess this should actually be self.ui.app.config in the 3 instances of this check as I can get it to crash fairly easily as app isn't indexable.

Alberth289346: :o I'll have a look at it. Thanks

PR #1670, opened 2020-07-19
[RDY] Make cheat.wav announceable

lewri:
- Moves the sound for when a cheat is used into the announcement queue as PriorityCritical. This means it will still play while the game is paused.

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/cheats.lua
+++ b/CorsixTH/Lua/dialogs/resizables/cheats.lua
@@ -134,7 +134,7 @@ function UICheats:buttonClicked(num)
     if self.cheats[num].name ~= "lose_level" then
       local announcements = self.ui.app.world.cheat_announcements
       if announcements then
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/resizables/cheats.lua
```diff
@@ -134,7 +134,7 @@ function UICheats:buttonClicked(num)
     if self.cheats[num].name ~= "lose_level" then
       local announcements = self.ui.app.world.cheat_announcements
       if announcements then
-        self.ui:playSound(announcements[math.random(1, #announcements)])
+       self.ui:playAnnouncement(announcements[math.random(1, #announcements)], AnnouncementPriority.Critical)
```
sadger: Looks like we've lost a space of indent here

lewri:
Just to point out it's actually the way the announcer works that makes it still play... Though I plan to look into it on #1671

PR #1674, opened 2020-07-22
[RDY] Switching faxes keeps game paused

### LINKED ISSUE #1625: Switching faxes unpauses game
#### Describe the issue
If you have two faxes, and try to switch between them the game will unpause. Even though a fax is open.

#### Steps to Reproduce
1. Get two faxes popup on the bottom of the game

2. Open one, then go to open the other.

#### Expected Behavior
Game should stay paused when viewing faxes 

#### System Information
**CorsixTH Version:** Latest trunk

**Operating System:** W10

#### Gamelog.txt
None

lewri:
*Fixes #1625*

- Game will now stay paused when you switch between faxes
- Game also recognises if a staff_rise window is open (and vice versa) and stays paused then too

The function call `` TheApp.world:setSpeed("Pause") `` had to be moved in message.lua to after the new window was made, which was the reason why the game would unpause.

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/fax.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/fax.lua
@@ -283,7 +283,6 @@ function UIFax:appendNumber(number)
 end
 
 function UIFax:close()
-  local world = self.ui.app.world
   self.icon.fax = nil
   self.icon:adjustToggle()
   UIFullscreen.close(self)
--- a/CorsixTH/Lua/dialogs/fullscreen/fax.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/fax.lua
@@ -282,7 +282,7 @@ function UIFax:close()
   self.icon.fax = nil
   self.icon:adjustToggle()
   UIFullscreen.close(self)
-  if world and world:isCurrentSpeed("Pause") then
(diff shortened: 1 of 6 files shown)


CorsixTH/Lua/dialogs/fullscreen/fax.lua
```diff
@@ -282,7 +282,7 @@ function UIFax:close()
   self.icon.fax = nil
   self.icon:adjustToggle()
   UIFullscreen.close(self)
-  if world and world:isCurrentSpeed("Pause") then
+  if world and world:isCurrentSpeed("Pause") and #self.ui:getWindows(UIStaffRise) == 0 then
```
sadger, line 285: I understand the point of this from our discussions in IRC but only from reading the code it's not clear why this is happening. Could we perhaps create a function to count all other pause dialogs that might be open to use in both cases?

TheCycoONE:
I'd much prefer pausing and unpausing for dialogs was in one spot. Dialogs could have a `should_pause` property, and the open dialogs could be iterated to see if we are in / should be in a paused on dialog state.

Willing to accept this patch for now, but if you could make that refactor happen I'd be happier.

lewri:
@TheCycoONE I did try to unify it into one spot but I wasn't successful. But this solution should be ok for now.

I tried to use world:onTick() which does work but every time you change a window, everyone moves 1 tick.

lewri:
Bear with me... I think I have something better

Issue #1193, opened 2017-01-28
Money bar broken with Russian or Dutch lanuage


ruPaladin:
#### Describe the issue
Money bar broken http://i.imgur.com/IS7DRYb.png

#### Steps to Reproduce
1. Start game with Russian or Dutch language.

#### System Information
**CorsixTH Version:** 0.6

**Operating System:** Windows 10

**Theme Hospital Version:** GOG.com

#### Gamelog.txt
*For information about where to find gamelog.txt see:
https://github.com/CorsixTH/CorsixTH/wiki/Frequently-Asked-Questions#where-do-i-find-the-configuration-or-the-gamelog-file*

Alberth289346:
![numbers](https://cloud.githubusercontent.com/assets/3254232/22434964/7f3d4bdc-e71e-11e6-9005-a0011e46a62d.png)

It's hard to see what is happening, so I enlarged the numbers a bit. I hope you added the red line, at least I don't know where that could come from otherwise.

You can see clearly that the font of the digits are too small, two numbers fit in one wheel.
Perhaps you are using a font that cannot provide digits of the right size?

What font do you use, and what happens if you try another font?

MarkL1961:
@Alberth289346  I wondered whether the red line is from reputation, but when I changed languages yesterday I got what you did and that was small numbers only.
Are numbers the same in all languages?  If they are could we just use the default font from TH here?

TheCycoONE:
Digits are not the same in all languages. The CJK languages use different digits as does Arabic (which we do not support) off the top of my head. That said we likely use ansi digits regardless. I didn't find any conversion for numbers. Alternatively we could use one text box per digit for this display.

ruPaladin:
Sorry for inconvenience. The problem is only with numbers. Red line was added in GIMP. I use Windows and didn't change default settings.

PenguiN560:
I also have this problem. Is there some recommendation which font is better to used for different languages for normal display?

TheCycoONE:
No unfortunately there is no known workaround for this issue. Regression bisects and/or patches are appreciated. Otherwise I'll try to get this fixed before the next major version.

PenguiN560:
@TheCycoONE oh, ok, thank you, we will hope on it. 

Also some messages on Russian language go away from message boxes and some others written over others text
[...]

PR #1679, opened 2020-07-30
Stopgap solution for #1193

lewri:
Forces the money bar to use the bitmap font for legibility reasons; until native language numbers are introduced (see #1193 discussion).

- Makes all languages use the bitmap font for moneybar for legibility's sake
- At some point we want to replace this with support for native language numbers (e.g. Korean) but for now this is better than the current improperly drawn moneybar numbers (added a TODO to reflect this)

### DIFF
--- a/CorsixTH/Lua/graphics.lua
+++ b/CorsixTH/Lua/graphics.lua
@@ -393,8 +393,8 @@ end
 function Graphics:loadFont(sprite_table, x_sep, y_sep, ...)
   -- Allow (multiple) arguments for loading a sprite table in place of the
   -- sprite_table argument.
-  -- TODO: Native number support for e.g. Korean languages. Current use of cach_x_sep is a stopgap solution for #1193 and should be eventually removed
-  local cache_x_sep = x_sep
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/graphics.lua
```diff
@@ -393,6 +393,8 @@ end
 function Graphics:loadFont(sprite_table, x_sep, y_sep, ...)
   -- Allow (multiple) arguments for loading a sprite table in place of the
   -- sprite_table argument.
+  -- TODO: Native number support for e.g. Korean languages. Current use of cach_x_sep is a stopgap solution for #1193 and should be eventually removed
+  local cache_x_sep = x_sep
```
TheCycoONE, line 397: This function is polymorphic, though I don't know it's ever used that way - as such the arguments have the wrong name for what you're using them for, but your variable probably shouldn't. Call it `font_name` or something.

lewri:
Local variable changed. I've rather leave the x_sep inside the function args as it is as it seems to match up with whatever y_sep does.

PR #1681, opened 2020-08-02
[RFC] Fix for CorsixTH#1673

### LINKED ISSUE #1673: [Bug] Reception queue does not re-route if unstaffed
#### Describe the issue
If you fire a receptionist from a staffed desk that has a queue, the patients do not requeue at another desk.

If you pick up the desk however, the patients do re-queue.

#### Steps to Reproduce
1. Set down two reception desks and two receptionists in an active hospital

2. Remove one of the receptionists but leave the desk as is.

#### Expected Behavior
Patients should requeue to a manned desk when their current one is unmanned.
Exception should probably be if we have no manned desks at all.

#### System Information
**CorsixTH Version:** 0.64

**Operating System:** W10

### Additional Info
This is probably an oversight as we don't even expect a receptionist to get fired.

lewri:
Covers the highly unlikely event we have an unmanned reception desk for whatever reason.

*Fixes #1673*

- If a reception desk with a queue becomes unmanned, all patients are now rerouted to another desk
- Vip also now gets handled correctly if this happens (otherwise he gets stuck in an infinite meander loop)
- If there are no manned desks available, patients meander until a manned desk is available.

### DIFF
--- a/CorsixTH/Lua/objects/reception_desk.lua
+++ b/CorsixTH/Lua/objects/reception_desk.lua
@@ -80,7 +80,7 @@ function ReceptionDesk:ReceptionDesk(...)
   self.queue:setMaxQueue(20) -- larger queues for reception desk
   self.hover_cursor = TheApp.gfx:loadMainCursor("queue")
   self.queue_advance_timer = 0
-  self.rerouted = false -- helps prevent multiple reroute actions in tick
+  self.queue_rerouted = false -- helps prevent multiple reroute actions in tick
 end
 
 function ReceptionDesk:onClick(ui, button)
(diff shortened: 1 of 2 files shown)


lewri:
Working in making travis happy

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -91,11 +91,15 @@ function ReceptionDesk:onClick(ui, button)
...
 
+local queue_rerouted = false
+
 function ReceptionDesk:tick()
   local queue_front = self.queue:front()
   local reset_timer = true
   if self.receptionist and queue_front then
-    if queue_front:getCurrentAction().name == "idle" then
+    -- If we interrupted a patient at the front of the queue they get use_object instead of idle
+    if queue_front:getCurrentAction().name == "idle" or queue_front:getCurrentAction().name == "use_object" then
```
lewri, line 101: I been told not to try and interfere with the action queue and handle the exception instead.

lewri: Though saying that if someone wants to work out why I have no complaints

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -135,6 +139,17 @@ function ReceptionDesk:tick()
         queue_front.has_passed_reception = true
       end
     end
+  -- A reception desk with patients has become unmanned, make sure we reroute patients
+  -- If there are no manned desks available, let patients meander until one is available
```
TheCycoONE, line 143: staffed would be better than manned throughout.

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -91,11 +91,15 @@ function ReceptionDesk:onClick(ui, button)
   end
 end
 
+local queue_rerouted = false
```
TheCycoONE, line 94: What is this extra bit of state doing out here on it's own. It's not part of the reception desk so it's global. Doesn't necessarily play nicely with save state, and I don't see why it's needed.

If it is in fact needed, but it on the desk.

lewri: If we don't have a flag... we execute "rerouteAllPatients()" about 5-10 times before it stops. The flag means it only happens once. I'll move it inside the desk itself.

lewri:
This fix has highlighted a graphical problem with the way rerouteAllPatients() works with the action queue.
See video: https://www.dropbox.com/s/m1hi2l1j775t59b/screen-recording.webm?dl=0

It should probably be fixed... but that is beyond my scope... and it's the action queue....

CorsixTH/Lua/queue.lua
```diff
@@ -321,11 +321,14 @@ function Queue:movePatient(index, new_index)
...
-function Queue:rerouteAllPatients(action)
+--! Called when reception desk is destroyed (or unstaffed), or when a room is destroyed from a crashed machine.
+--!param action (function) What should the patient do now
+--!param origin The room or object, if needed
+function Queue:rerouteAllPatients(action, origin)
   for _, humanoid in ipairs(self) do
     -- check by class type as staff/vips shouldn't get a SeekRoomAction
-    if class.is(humanoid, Patient) then
+    -- Except at reception desks, a Vip could otherwise get stuck
+    if class.is(humanoid, Patient) or (class.is(humanoid, Vip) and origin.class == "ReceptionDesk") then
```
mugmuggy, line 331: VIPs don't actually get queued when visiting rooms (might need to check what happens with a build a room on path scenario). They get a walk to the door, but don't actually queue so they only ever get the expected callback.
Thus technically we don't need to check they are at a reception desk. We also could just use the action.name to determine if we were after a reception desk or a room or doing something else.

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -207,7 +226,7 @@ function ReceptionDesk:onDestroy()
       end
     end)
   end
-  self.queue:rerouteAllPatients(SeekReceptionAction())
+  self.queue:rerouteAllPatients(SeekReceptionAction(), object)
```
mugmuggy, line 229: There is no object still?

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -94,45 +95,63 @@ end
...
+      self.queue_rerouted = false
+    end
+
+  -- A reception desk with patients has become unstaffed, make sure we reroute patients
+  -- If there are no staffed desks available, let patients meander until one is available
+  elseif not self.queue_rerouted and self.queue:size() > 0 then
+    local hospital = self.hospital
+    for _, staff in ipairs(hospital.staff) do
+      if staff.humanoid_class == "Receptionist" and staff.associated_desk then
+        self.queue:rerouteAllPatients(SeekReceptionAction(), object)
```
mugmuggy, line 152: There is no object here?

CorsixTH/Lua/objects/reception_desk.lua
lewri: Object is declared at line 21. It passes that into the function as the "origin". I tried using self but it just returns nil.

CorsixTH/Lua/queue.lua
lewri: If that's the case the code will work fine without the origin.class == "ReceptionDesk" ?

lewri:
Probably best to hold off this pr for a while, until 1682 is squashed...

mugmuggy:
I was taking a look at this again, and I think there is a nicer solution than the queue_rerouted variable. And that is to call the rerouteAllPatients from within ReceptionDesk:checkForNearbyStaff just before it returns false when there is no available receptionist. It gets called when desk is removed and also when the receptionist is picked up/fired.

As for VIP/Inspector processing, I'm in a few minds on. One of the changes that I made to rerouteAllPatients was preventing Staff getting the action passed in. We know currently that Vips/Inspectors don't join room queues. Vips go into the expected list but never atttempt to go through the door, so do not queue.

We could just apply the same logic as patient instead.

Another option would be to do the MeanderAction still, and have logic in there to reset the state, like so with support for restarting if they could actually queue at a room:

```lua
  if class.is(humanoid, Vip) then
    if not humanoid.has_passed_reception then
      humanoid:setNextAction(SeekReceptionAction())
    else
      humanoid:queueAction(IdleAction())
      humanoid.waiting = 1
      humanoid:finishAction()
    end
    return
  end
```

That solution I believe doesn't just work as SeekReceptionAction queues a MeanderAction from within for implementing the procrastination part, ie gets stuck in a cycle until a manned desk appears. So I would be more leaning on the do the same as patient logic without the additional object check - until someone decides to implement queuing of VIP and reservation of the door logic. The additional 'object' check could be derived from the action.name anyway.

lewri:
I'm happy to close this in favour of something else, whatever that may be.

PR #1683, opened 2020-08-07
[rfc] Switch to translation text files

Alberth289346:
Currently, translations are Lua code, which makes using a translation service complicated.
This PR is a proof of concept to simplify that by using text files for translations rather than program code.

The current patch contains all ingredients, but it doesn't fly yet, as calling the parser doesn't seem to work (Lua protests in one way or another.). I converted the Dutch translation as an example.

If found useful, there a mass-conversion of all the current language files has to be done. I wrote a Python program to this end, but it doesn't handle comments yet (by design, currently). Transferring comments is generally a mess, as they need to end up at the right spot. While converting the Dutch translation, I found the few comments it had already a burden so likely it needs a different approach from manually extracting and editing the comments back in.


TheCycoONE:
My understanding from @SimplyRyan was that the hardest thing was that we didn't have a full set of prompts for the primary language due to copyright.

Alberth289346:
Well, yes, that's another thing that's needed.

That seems to imply we need 2 English sets of strings? That is "original" and "free/base/cth-english" ?
I can translate Dutch back to English, at least we discussed such an option in the past iirc.

lewri:
Should've probably asked earlier but have you still got the errors lua popped up?

Alberth289346:
> Should've probably asked earlier but have you still got the errors lua popped up?

No idea, but imho that's not relevant currently.

This issue aims to convert translations to a form more friendly for translation management software. The current blocking point seems to be lack of a set original text strings that can be used for translation.

Until that is resolved none of the other problems are anywhere near relevant imho.

TheCycoONE:
And it seems like the whole reason for that block of strings being missing is moot per the discussion in chat - we should be writing new material for all the original strings.

mefistotelis:
It is rarely a good thing to author a new solution, if there are existing ones - which community knows and uses, and which is adopted by thousands of projects.

For translations - there are such existing solutions.
One is PO/POT files - I adopted it for my Dungeon Keeper reimplementation, look here for details:
https://github.com/dkfans/keeperfx/tree/master/lang

When building assets, I'm converting that into binary format used by the game with this tool:
https://github.com/dkfans/po2ngdat

The PO/POT files are widely used, there are continuously supported tools to work with these formats, and most people who work on translations know it.

Alberth289346:
> It is rarely a good thing to author a new solution, if there are existing ones - which community knows and uses, and which is adopted by thousands of projects.

Tools is currently not the problem, the set of strings to use is currently the problem. English strings of the original program have copyright problems and cannot be distributed. We need new strings.

Then we need to attach (hopefully the existing) translations to the new strings, and only then we have a small tool problem.

> The PO/POT files are widely used ...

It's equally unwise to ditch the current working solution that we have unless it's a drop-in alternative which I doubt.

PR #1684, opened 2020-08-07
Use 0 to return lack of room or staff rather than false

Alberth289346:
Both functions used `false` for `0`, else a number. Unified it to using numbers only.


TheCycoONE:
Looks like a low hanging optimization later would be to add the boolean functions which exit early.

PR #1686, opened 2020-08-09
[RDY] lua 5.4 support

TheCycoONE:
Adds support for running CorsixTH with Lua 5.4.

At this point I have not encountered problems playing, and I am able to save hospitals but loading hospitals fails with a stack overflow for all but trivial saves.


TheCycoONE:
Confirmed that this version as is works with the current git version of lua and will therefore likely work with lua 5.4.2

I don't plan to spend further effort trying to be compatible with lua 5.4.0 or 5.4.1, and will point out those versions are unsupported.

CorsixTH/CorsixTH.lua
```diff
@@ -82,7 +82,7 @@ end
 
 -- Check Lua version
 if _VERSION ~= "Lua 5.1" then
-  if _VERSION == "Lua 5.2" or _VERSION == "Lua 5.3" then
+  if _VERSION == "Lua 5.2" or _VERSION == "Lua 5.3" or _VERSION == "Lua 5.4" then
```
Alberth289346, line 85: We may want to abort execution of the user has a wrong version of Lua 5.4, once we established that 5.4.2 works.

CorsixTH/Lua/date.lua
```diff
@@ -273,6 +273,19 @@ function Date_mt.__lt(one, other)
...
+    if one._month == other._month then
+      if one._day == other._day then
+        return one._hour <= other._hour
+      end
+      return one._day <= other._day
+    end
+    return one._month <= other._month
+  end
+  return one._year <= other._year
+end
```
Alberth289346, line 287:
```suggestion
function Date_mt.__le(one, other)
  if one._year ~= other._year then return one._year <= other._year end
  if one._month ~= other._month then return one._month <= other._month end
  if one._day ~= other._day then return one._day <= other._day end
  return one._hour <= other._hour
end
```
Writing the test the other way around avoids nesting.

CorsixTH/Lua/utility.lua
```diff
@@ -106,7 +106,7 @@ function loadfile_envcall(filename)
   return loadstring_envcall(result, "@" .. filename)
 end
 
-if _G._VERSION == "Lua 5.2" or _G._VERSION == "Lua 5.3" then
+if _G._VERSION == "Lua 5.2" or _G._VERSION == "Lua 5.3" or _G._VERSION == "Lua 5.4" then
```
Alberth289346, line 109: Would it make sense to fold this check in a separate function?

CorsixTH/Src/iso_fs.cpp
```diff
@@ -359,15 +359,21 @@ void iso_filesystem::clear() {
   delete[] error;
   error = nullptr;
   files.clear();
+  if (raw_file) fclose(raw_file);
```
Alberth289346, line 362: Also set it to `nullptr` to avoid double closing (one time calling `clear` in some usage code, and one time while destroying the object)

CorsixTH/Src/persist_lua.cpp
```diff
@@ -270,217 +267,194 @@ class lua_persist_basic_writer : public lua_persist_writer {
       // things).
       lua_getfenv(L, 1);
       lua_pushvalue(L, iIndex);
-      lua_gettable(L, -2);  // Might (indirectly) call writeObjectRaw
-      uint64_t iValue = (uint64_t)lua_tonumber(L, -1);
-      lua_pop(L, 2);
+
+      lua_rawget(L, -2);
+      uint64_t iValue = static_cast<uint64_t>(lua_tonumber(L, -1));
```
Alberth289346, line 272: Old code, but shouldn't this be some `UINT64_MAX` or so?

CorsixTH/Src/th_lua.cpp
```diff
@@ -320,58 +320,47 @@ void luaT_pushtablebool(lua_State* L, const char* k, bool v) {
...
+    case LUA_TSTRING: /* strings */
+      std::printf("string: '%s'\n", lua_tostring(L, idx));
+      break;
+    case LUA_TBOOLEAN: /* booleans */
+      std::printf("boolean %s\n", lua_toboolean(L, idx) ? "true" : "false");
+      break;
+    case LUA_TNUMBER: /* numbers */
+      std::printf("number: %g\n", lua_tonumber(L, idx));
+      break;
+    default: /* other values */
```
Alberth289346, line 335: Comments are quite useless, but harmless I guess

CorsixTH/Src/th_lua.cpp
```diff
@@ -320,58 +320,47 @@ void luaT_pushtablebool(lua_State* L, const char* k, bool v) {
...
-        break;
-      case LUA_TNUMBER: /* numbers */
-        std::printf("number: %g\n", lua_tonumber(L, i));
-        break;
-      default: /* other values */
-        std::printf("%s\n", lua_typename(L, t));
-        break;
-    }
-    std::printf(" "); /* put a separator */
+    printf("(%d) ", i);
```
Alberth289346, line 348: Something debugging?

CorsixTH/Src/th_lua.cpp
```diff
@@ -320,58 +320,47 @@ void luaT_pushtablebool(lua_State* L, const char* k, bool v) {
...
-        break;
-    }
-    std::printf(" "); /* put a separator */
+  /* table is in the stack at index 't' */
+  lua_pushnil(L); /* first key */
+  while (lua_next(L, idx) != 0) {
+    /* uses 'key' (at index -2) and 'value' (at index -1) */
+    printf("key: ");
+    luaT_printvalue(L, -2);
+    printf("value: ");
```
Alberth289346, line 361: and more `printf`, at the left we have `std::printf`

CorsixTH/Src/th_lua.h
```diff
@@ -408,6 +427,10 @@ void luaT_execute(lua_State* L, const char* sLuaString, T1 arg1, T2 arg2,
 
 void luaT_pushtablebool(lua_State* L, const char* k, bool v);
 
+// Print debugging helper functions
+
+void luaT_printvalue(lua_State* L, int idx);
+
```
Alberth289346, line 433: Ah, probably this function thus, seems fine then.

I fail to see the point of having an empty line between the function declarations though. Double-spacing code doesn't make it simpler imho.

CorsixTH/Src/th_lua.cpp
TheCycoONE: The whole function is for debugging, along with the one above it. Otherwise looking at the lua state in gdb is hard.

TheCycoONE: Should switch to std::printf

CorsixTH/Src/th_lua.cpp
TheCycoONE: Good catch, seems every compiler I use wants to import c functions into the global scope.

CorsixTH/Src/th_movie.cpp
```diff
@@ -307,7 +307,7 @@ void av_packet_queue::push(AVPacket* pPacket) {
     (defined(CORSIX_TH_USE_FFMPEG) &&                      \
      LIBAVCODEC_VERSION_INT < AV_VERSION_INT(57, 12, 100))
   if (av_dup_packet(pPacket) < 0) {
-    throw - 1;
+    throw -1;
```
TheCycoONE, line 310: Different clang format versions seem to disagree on this - why am I throwing -1 anyway though? Replace with a more descriptive throw or add a no-format comment.

CorsixTH/Src/persist_lua.cpp
TheCycoONE: I don't follow, was this comment on the wrong line?

Alberth289346: `-1` looks like an attempt to write signed data into an unsigned value, in particular, the bit representation of `0xFFFF...`.

TheCycoONE: No, that -1 means the top of the lua stack. Return the item at the top of the lua stack as if it was a number.

TheCycoONE:
Requesting that people test this on lua 5.4 with larger save games. If you can't I would still appreciate if you could send me some larger saves. This implementation works for me with the saves I have.

TheCycoONE:
Setting the cstacklimit isn't working for me with my own test saves, so I'll retract that commit.

TheCycoONE:
Fixed support for lua 5.4.1 (and probably 5.4.0)

TheCycoONE:
Has anyone tried this out (other than me)? I can merge, but I'd like to know at least one other person has used it successfully (especially around loading and saving).

colinjmatt:
I'd like to test this before merge if I can? I'll have time over the weekend to do this.

colinjmatt:
All looks good. Played for a little over an hour and save/loaded 6 times. Exited the game fully, restarted and loaded 3 times all without any visible errors.

PR #1688, opened 2020-08-10
Use self.ui.app rather than self.world.app, apparently.

Alberth289346:
mugmuggy said in https://github.com/CorsixTH/CorsixTH/pull/1667#pullrequestreview-464171942
that I shouldn't use `self.world.app` but use `self.ui.app` instead.

Both cases work locally, so I don't know what the preferred solution is, but here goes.

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -864,7 +864,7 @@ function UIEditRoom:enterDoorPhase()
 
   -- check if all adjacent tiles of the rooms are still connected
   if not self:checkReachability() then
-    if self.world.app.config.allow_blocking_off_areas then
+    if self.ui.app.config.allow_blocking_off_areas then
       print("Blocking off areas is allowed with room " .. self.blueprint_rect.x .. ", " .. self.blueprint_rect.y .. ".")
     else
       -- undo passable flags and go back to walls phase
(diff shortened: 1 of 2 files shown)


mugmuggy:
Admittedly I was playing with the 5.4 PR, wanting to see if there was this improvement in performance, and self.world.app would show up as nil. The two instances of the check in place_objects in theory would be found via world, there its local to the function instead of self. In edit_room the other piece of code referring to config just uses TheApp.config. But I quickly changed world to ui and the issue resolved itself.

Once I say move the bin from a room to the corridor (I replicate well with op theatre bin to in front of the nearest ward door) it gets this first error, which may repeat, and then normally it errors again, with a empty action queue for a humanoid.

But I also have confirmed it on the PR 1667 appveyor build. I also can replicate the issue when building a new room and placing the objects, so there is probably something else broken (I don't even have to move the cursor object out of the room. And note this is with previous saves that I got it on, from 0.61, so there might be something that has changed but the persistence of the UIEditRoom/UIPlaceObjects may have changed between then.
```
Error in motion handler: 
Lua\dialogs\place_objects.lua:730: attempt to index a nil value (field 'app')
stack traceback:
	Lua\dialogs\place_objects.lua:730: in method 'setBlueprintCell'
	Lua\dialogs\place_objects.lua:818: in field 'onCursorWorldPositionChange'
	Lua\dialogs\edit_room.lua:1403: in method 'onCursorWorldPositionChange'
	Lua\window.lua:1502: in function <Lua\window.lua:1498>
	(...tail calls...)
	Lua\game_ui.lua:542: in function <Lua\game_ui.lua:530>
	(...tail calls...)
	Lua\app.lua:1014: in function <Lua\app.lua:1009>
```

TheCycoONE:
I'm interested in knowing why world.app is nil. This would presumably cause more problems.

TheCycoONE:
Interestingly we never use world.app anywhere, not even within world. I suspect the underlying issue is long standing then, though I don't yet have a good idea what it is. In world we consistently use TheApp.

Issue #148, opened 2014-01-16
Room (construction) selection menu doesn't automatically update


mattmill30:
When the room construction menu is already open, and a new room is made available, it is necessary to close and re-open the menu, rather than the menu automatically add the new room to the relevent section (visible or not).

To avoid mis-selection in the case of updating the visible room list, the new room could simply be appended to the bottom of the list, and placed into it's correct order when the list is next made invisible (either by switching to another list, or closing of the menu)

PR #1691, opened 2020-08-13
Reset Build Rooms dialog when new rooms are researched

### LINKED ISSUE #513: Build room-windows (open) isn't updated when a new room is researched
_From [maartenp...@hotmail.nl](https://code.google.com/u/103280305005036074093/) on January 18, 2013 14:03:24_

What steps will reproduce the problem? 1.Open the build room-window.
2.A new room is available (researched.)
3.You don't see this at the screen (you first have to close it and re-open it and then you can build the room.) What is the expected output? What do you see instead? I should see that the screen is directly updated and that I can build the researched room. Instead of first close and then re-open the window.

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1455_

tobylane:
*Fixes #513 #148*

- When a room is researched, if the build rooms dialog is open then reopen it

### DIFF
--- a/CorsixTH/Lua/dialogs/build_room.lua
+++ b/CorsixTH/Lua/dialogs/build_room.lua
@@ -96,22 +96,27 @@ function UIBuildRoom:UIBuildRoom(ui)
     _S.room_classes.clinics,
     _S.room_classes.facilities
   }
-  self.category_rooms = {
-  }
+
+  self:updateBuildableRooms()
+
+  self:makeTooltip(_S.tooltip.build_room_window.cost, 160, 228, 282, 242)
+end
+
+function UIBuildRoom:updateBuildableRooms()
+  local app = self.ui.app
+  self.category_rooms = {}
   for i, category in ipairs({"diagnosis", "treatment", "clinics", "facilities"}) do
     local rooms = {}
     self.category_rooms[i] = rooms
     for _, room in ipairs(app.world.available_rooms) do
       -- NB: Unimplemented rooms are hidden unless in debug mode
       if (app.config.debug or room.class) and room.categories[category] and
-          ui.hospital.discovered_rooms[room] then
+          self.ui.hospital.discovered_rooms[room] then
         rooms[#rooms + 1] = room
       end
     end
     table.sort(rooms, function(r1, r2) return r1.categories[category] < r2.categories[category] end)
   end
-
-  self:makeTooltip(_S.tooltip.build_room_window.cost, 160, 228, 282, 242)
 end
 
 local cat_label_y = {21, 53, 84, 116}
@@ -145,6 +150,7 @@ function UIBuildRoom:setCategory(index)
     self.ui:tutorialStep(3, 3, 2)
   end
   self.category_index = index
+  self:updateBuildableRooms()
   self.list_title = _S.build_room_window.pick_room_type
   self.list = self.category_rooms[index]
 


mugmuggy:
I took them as something along the lines of, if the dialog is open, and I'm on facilities I could switch back to diagnostics and the new room would show up, or if I'm on diagnostics, the room would show up at the end of the list, thus not affecting any current selection.

The opposite would be to list all rooms and disable the ones that haven't been discovered, and as they are discovered enable them. It does get a bit messy getting UIBuildRoom notified of updates and drawing out of order for the current list (as not to affect current selection). I didn't think what I had was ideal and shelved it for another day, https://github.com/mugmuggy/CorsixTH/tree/build_room_dialog

lewri:
I think the version @mugmuggy made seems less invasive then resetting the window. Then again I assume the original was completely devoid of this feature.

TheCycoONE:
My preference is not to ever reorder the list - which I guess puts me in the disabled rooms in list camp - perhaps leaving empty buttons? It would be good to mock up the UI before implementing.

mugmuggy:
Tested what TH does. Upon researching a new room, you have to click on the category again to have it update the list. It will put it back in the normal order, visually reshuffling the current list.

TheCycoONE:
sounds good to me, thanks @mugmuggy

tobylane:
Thanks mugmuggy, your branch helped.

Ready.

CorsixTH/Lua/research_department.lua
```diff
@@ -543,6 +543,7 @@ function ResearchDepartment:discoverObject(object, automatic)
           end
         end
       end
+      UIBuildRoom:updateList()
```
mugmuggy, line 546: Think you want to get the instance if its open still, like you had before.
I assume this is why you have
```lua
  local app = TheApp.ui.app
```
in update list instead of self.ui. The code I had avoided the dependency but with much more overhead, another reason I didn't like it. But the simpler behaviour avoids it.

CorsixTH/Lua/dialogs/build_room.lua
```diff
@@ -96,22 +96,27 @@ function UIBuildRoom:UIBuildRoom(ui)
...
   }
-  self.category_rooms = {
-  }
+
+  self:updateList()
+
+  self:makeTooltip(_S.tooltip.build_room_window.cost, 160, 228, 282, 242)
+end
+
+function UIBuildRoom:updateList()
```
mugmuggy, line 105: Being a little picky here, as the function name came from my branch - I did say I didn't like it. Technically setCategory updates a member, 'list' from the category_rooms from that this reprocesses. Maybe updateBuildableRooms or something.

CorsixTH/Lua/dialogs/build_room.lua
```diff
@@ -145,6 +150,7 @@ function UIBuildRoom:setCategory(index)
     self.ui:tutorialStep(3, 3, 2)
   end
   self.category_index = index
+  self:updateList()
```
mugmuggy, line 153: I don't think you need this call here, as research_department calls across and you'll have the updated list of rooms already or you keep this call here and not call from research_department. Doing it from research you only do it once per discovered/available room and this does the update every category change.

tobylane: It works when it's just this line and not when it's just the line in research_department.

PR #1696, opened 2020-08-17
wip: Refactor create emergency

Alberth289346:
*Fixes two TODOs in the hospital:createEmergency function. Also deletes a few duplicate staff class translation tables.

Place of the new function mey be less than optimal, but this one seemed not too bad and actually works.

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
@@ -184,17 +184,9 @@ function UICasebook:updateIcons()
       research = research and (_S.tooltip.casebook.cure_requirement.research_machine .. research .. "). ") or ""
       build    = build    and (_S.tooltip.casebook.cure_requirement.build_room .. build .. "). ") or ""
 
-      local staffclass_to_string = {
-        Nurse        = _S.staff_title.nurse,
-        Doctor       = _S.staff_title.doctor,
-        Surgeon      = _S.staff_title.surgeon,
-        Psychiatrist = _S.staff_title.psychiatrist,
-        Researcher   = _S.staff_title.researcher,
-      }
-
       -- Staff requirements
       for sclass, amount in pairs(req.staff) do
-        staff = (staff and (staff .. ", ") or " (") .. staffclass_to_string[sclass] .. ": " .. amount
+        staff = (staff and (staff .. ", ") or " (") .. translateStaffclass(sclass) .. ": " .. amount
       end
       staff = staff and (_S.tooltip.casebook.cure_requirement.hire_staff .. staff .. "). ") or ""
 
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -1012,13 +1012,7 @@ function Patient:updateMessage(choice)
             output_text = strings.need_to_build:format(room_name)
           end
         elseif #req.rooms == 0 and next(req.staff) then
-          local staffclass_to_string = {
-              Nurse        = _S.staff_title.nurse,
-              Doctor       = _S.staff_title.doctor,
-              Surgeon      = _S.staff_title.surgeon,
-              Psychiatrist = _S.staff_title.psychiatrist,
-          }
-          output_text = strings.need_to_employ:format(staffclass_to_string[next(req.staff)])
+          output_text = strings.need_to_employ:format(translateStaffclass(next(req.staff)))
         end
         self.message[3].text = output_text
       else
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1217,18 +1217,14 @@ function Hospital:createEmergency(emergency)
     end
 
     self.emergency = emergency
-    -- The last room in the list of treatment rooms is considered when checking for availability.
-    -- It works for all original diseases, but if we introduce new multiple room diseases it might break.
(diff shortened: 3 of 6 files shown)


CorsixTH/Lua/hospital.lua
```diff
@@ -1217,32 +1217,28 @@ function Hospital:createEmergency(emergency)
...
-    -- Check so that all rooms in the list are available
-    if self:countRoomOfType(emergency.disease.treatment_rooms[no_rooms]) > 0 then
-      room_name = nil
+    -- Check needed rooms and staff.
+    local room_name, staff_name
+    local required = self:checkDiseaseRequirements(emergency.disease.id)
+    if required then
+      if #required.rooms > 0 then room_name = required.rooms[1] end -- Take the first missing room
+
+      for sn, _ in pairs(required.staff) do staff_name = sn; break; end
```
mugmuggy, line 1227: When I added the additional fax, I looked at that TODO and checkDiseaseRequirements, and I think you have the same problem and I just worked around it to avoid a refactor of checkDiseaseRequirements.

Ward and Op Theatre, room and staff will mix in the fax.
ie build_and_employ, Ward and Surgeon or Op Theatre and Nurse depending on rooms built and available staff.

Alberth289346: I see a few ways out, but not sure what is preferred.

Current strings handle
- 'all covered',
- 'missing 1 person of 1 staff type',
- 'missing 1 room type',
- 'missing 1 person of 1 staff type and 1 room type'.

In the general sense, we can have an arbitrary number of missing room types, and need 1 or more employees of 1 or more staff types.

- One way out is to extend the strings to the point that you can have an arbitrary number missing of room types and/or staff types/persons. Likely this explodes in the number of strings, for English it's not too bad on first judgement, but in translations it may become messy.

- Second way out is to drop the staff things if it gets complicated, and generalize with an additional string to something like "fully staffed X, Y, and Z rooms" (where an arbitrary list of rooms is still a separate problem to deal with in translations). Note that at this point re-use of the checkDiseaseRequirements fails, as it doesn't tell about rooms that you need and already have. This needs to be coded separately or the function needs to be extended.

- Third way out is not to generate emergencies that don't fit in the currently available strings. This may however lead to never having an emergency if you build a sufficiently crippled hospital.

mugmuggy: In the seek room action its handled a bit differently, as the seek there has multiple rooms, whereas here for emergency in the current state of play, is always the last room to visit for treatment. In seek room there is no attempt for multiple rooms or staff in the message, you just '_can_not_cure_' them.

Could probably achieve the same by processing the checkDiseaseRequirements in reverse, and aborting after the first if a flag is set, or adding the room_id as input and checking against only that room. But that all just feels a little hacky. Or modify checkDiseaseRequirements to reer to the room for the staff as well and modify the callers to process that.

Also the text returned is different long_name vs name, ie Psychiatric Room vs Psychiatric so there is a few things to unpick. The method getRoomNameAndRequiredStaffName appears to have been only added for fax processing.

CorsixTH/Lua/world.lua
```diff
@@ -2279,19 +2279,7 @@ function World:getRoomNameAndRequiredStaffName(room_id)
...
-    staff_name = _S.staff_title.psychiatrist
-  elseif staff_name == "Researcher" then
-    staff_name = _S.staff_title.researcher
-  elseif staff_name == "Surgeon" then
-    staff_name = _S.staff_title.surgeon
-  elseif staff_name == "Doctor" then
-    staff_name = _S.staff_title.doctor
-  end
-  return room_name, required_staff, staff_name
+  return room_name, staff_name, translateStaffclass(staff_name)
```
lewri, line 2282: The problem is your capitalisation :)
``translateStaffClass`` not ``translateStaffclass``

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -1012,13 +1012,7 @@ function Patient:updateMessage(choice)
...
           end
         elseif #req.rooms == 0 and next(req.staff) then
-          local staffclass_to_string = {
-              Nurse        = _S.staff_title.nurse,
-              Doctor       = _S.staff_title.doctor,
-              Surgeon      = _S.staff_title.surgeon,
-              Psychiatrist = _S.staff_title.psychiatrist,
-          }
-          output_text = strings.need_to_employ:format(staffclass_to_string[next(req.staff)])
+          output_text = strings.need_to_employ:format(translateStaffclass(next(req.staff)))
```
lewri, line 1015: and here too

CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
```diff
@@ -184,17 +184,9 @@ function UICasebook:updateIcons()
...
-        Doctor       = _S.staff_title.doctor,
-        Surgeon      = _S.staff_title.surgeon,
-        Psychiatrist = _S.staff_title.psychiatrist,
-        Researcher   = _S.staff_title.researcher,
-      }
-
       -- Staff requirements
       for sclass, amount in pairs(req.staff) do
-        staff = (staff and (staff .. ", ") or " (") .. staffclass_to_string[sclass] .. ": " .. amount
+        staff = (staff and (staff .. ", ") or " (") .. translateStaffclass(sclass) .. ": " .. amount
```
lewri, line 189: One more here

lewri (commented):
It took me months to notice (and sure it stumped you too) but that should fix it.

Alberth289346:
Don't think I will do this anytime soon.

PR #1698, opened 2020-08-19
Add advisor messages about handymen and receptionists, staff happiness from windows and big rooms, all cured award

tobylane:
- Use the two remaining receptionist warnings, add a new string for the other similar event
- Award a trophy, award and bonus for successfully diagnosing and treating everyone
- Use the handyman hint for adjusting attributes, and have it come up once
- Count the windows and add staff happiness based on this. 0.001 is a starting number, it probably needs refining.
- Add staff happiness for larger rooms. This could be expanded, say use the ratio between the room requirements and the actual space as a basis for the factor, if that's realistic.
- --continue=1 will load Autosave1.sav, --continue=true (or any word) will load the youngest Autosav*.sav.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -319,6 +319,18 @@ function App:init()
           "Please make sure you have specified a font file in the config file."}))
     end
 
+    -- If the player wants to continue then load the youngest file in the Autosaves folder
+    -- If they give a month number then load that month's autosave
+    if self.command_line.continue then
+      local num = self.command_line.continue
+      local file = "Autosaves" .. pathsep .. "Autosave" .. num .. ".sav"
+      if num >= "1" and num <= "12" and lfs.attributes(self.savegame_dir .. file, "size") then
+        self.command_line.load = file
+      else
+        self.command_line.load = "Autosaves" .. pathsep ..
+          FileTreeNode(self.savegame_dir .. "Autosaves"):getMostRecentlyModifiedChildFile(".sav").label
+      end
+    end
     -- If a savegame was specified, load it
     if self.command_line.load then
       local status, err = pcall(self.load, self, self.savegame_dir .. self.command_line.load)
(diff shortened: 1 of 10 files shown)


CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -97,8 +97,21 @@ function Staff:tickDay()
...
+  if room then
+    local count = room:countWindows()
+    self:changeAttribute("happiness", 0.001 * count)
+    if room.room_info.id == "staff_room" then -- Pleased another bit
+      self:changeAttribute("happiness", 0.001 * count)
+    end
+  end
+
+  -- Extra space in the room you are in adds to your happiness
+  if room and room.minimum_size and room.width > room.minimum_size and
```
TheCycoONE, line 111: Only extra space in both dimensions no matter how much?

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -97,8 +97,21 @@ function Staff:tickDay()
    self:changeAttribute("happiness", 0.05)
   end
 
-  --TODO windows in your work space and a large space to work in add to happiness
-  -- working in a small space makes you unhappy
+  local room = self:getRoom()
+  -- It always makes you happy to see the outdoors (or windows to anywhere)
+  if room then
+    local count = room:countWindows()
+    self:changeAttribute("happiness", 0.001 * count)
```
TheCycoONE, line 104: We might want to fiddle with this. Seems like too much advantage to scale linear for something that is free / cosmetic to me?

CorsixTH/Lua/app.lua
```diff
@@ -319,6 +319,32 @@ function App:init()
           "Please make sure you have specified a font file in the config file."}))
     end
 
+    -- If the player wants to continue then load the youngest file in the Autosaves folder
+    -- If they give a month number then load that month's autosave
+    if self.command_line.continue then
```
TheCycoONE, line 324: Document the new option in the man page

CorsixTH/Lua/entities/humanoids/staff.lua
tobylane: It could be the amount of extra space in relation to minimum, say factor = (width * height) / (minimum_size * miminum size) / 1000. Could be log scale, to keep a 20x20 staff room from adding 0.025.

tobylane:
With log the happiness factor looks better, for a 15x size gp's office it gives 0.003, that room with 34 windows gives 0.004.

PR #1700, opened 2020-08-29
Update appdata.xml to metainfo.xml

tobylane:
- The identified missing parts from the previous PR

Tested the install but not the open from desktop.


CorsixTH/com.corsixth.corsixth.metainfo.xml
```diff
@@ -23,6 +23,9 @@
   <provides>
     <binary>corsix-th</binary>
   </provides>
+  <launchable>
+    <desktop-id>com.corsixth.corsixth</desktop-id>
```
TheCycoONE, line 27: My understanding is that desktop-id is suppose to be the value of the type attribute of the launchable tag: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-launchable

TheCycoONE: Also .desktop should be part of the value

TheCycoONE:
Was just about to post my own PR for this, but it looks like you beat me too it. Just a few concerns about the launchable tag.

TheCycoONE:
Looking at the spec closer I see that our component type should also be `desktop-application` rather than `desktop`

PR #1704, opened 2020-09-03
[RDY] Refactor cheats + Roujin Challenge

### LINKED ISSUE #1702: Reinstating Roujin's Challenge to the code
At some point in development the code for Roujin's challenge got lost. So right now entering the fax number for it does nothing except an adviser pop-up.

The challenge itself is inspired by the Anime film Roujin Z; where the Japanese population has become inundated with elderly patients requiring end-of-life care. To solve it they came up with a machine that cared for them without intervention (and things went wrong when the machine develops a mind of its own).

Aims are to:
- [x] Reintroduce the cheat
- [x] Make cheat trigger the "cheating" flag
- [x] Update the wiki page with its origins

Anyone interested reading up about the film (or watching it, it's pretty comfy) see here: https://myanimelist.net/anime/2000/Roujin_Z

lewri:
*Fixes #1702*

- Refactors the cheats into their own class to separate out from the UICheats
- Should make it easier to add cheats through the fax codes in the future
- Also reintroduces Roujin's Challenge. When active, spawn rate is locked at 40 from the next month until it is turned off. When it turns off, the spawns are removed for the current month so the player has a chance to recover from the chaos.


lewri:
@TheCycoONE need your help with the afterload in cheats.lua.
It needs to be done after hospital:afterload(). Currently it's doing it beforehand and as a result L160 in cheats.lua breaks.

I tested this by changing the order in app.lua's afterload:
https://github.com/CorsixTH/CorsixTH/blob/9bf8c8e819b74a5a5b5618a09097cffa7541fabc/CorsixTH/Lua/app.lua#L1554-L1556

TheCycoONE:
I'm ok if you just close the cheat window, then you don't have that problem right?

lewri:
Appveyor stalled. Not sure why:
```
Packages (1) pacman-5.2.1-6
Total Installed Size:  37.74 MiB
Net Upgrade Size:      -8.86 MiB
:: Proceed with installation? [Y/n] 
checking keyring...
checking package integrity...
loading package files...
checking for file conflicts...
checking available disk space...
:: Processing package changes...
upgrading pacman...
:: Synchronizing package databases...
downloading mingw32.db...
downloading mingw32.db.sig...
error: :: Import PGP key 4A6129F4E4B84AE46ED7F635628F528CF3053E04? [Y/n] mingw32: key "4A6129F4E4B84AE46ED7F635628F528CF3053E04" is unknown
error: ke
```

TheCycoONE (changes requested):
Cheat should own the cheat list as well - you are decoupling cheat logic from the UI. Maybe move to a different file.

lewri:
Changes implemented as requested.

CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
```diff
@@ -0,0 +1,147 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE. --]]
+
+corsixth.require("announcer")
+
+local AnnouncementPriority = _G["AnnouncementPriority"]
```
TheCycoONE: Not used in this module, neither is announcer above. You can/should remove these lines.

lewri: Cyco this is removed in 74cf952 think you clicked the wrong commit

PR #1706, opened 2020-09-12
Patients in a ward may stay in bed when the nurse leaves

### LINKED ISSUE #1277: Patients leave ward too early when Nurse is picked up
#### Describe the issue
Nurse leaving ward (at least on pickup) aborts the patients stay. Upon picking up the nurse from the ward, patients leave immediately when they should actually stay the visit.

#### Steps to Reproduce
1. Wait for patient to enter ward.

2. Pickup nurse.

3. Watch patient leave.

#### Save Game
[nursepickup.zip](https://github.com/CorsixTH/CorsixTH/files/1456904/nursepickup.zip)
Patient is already going to bed here, pick up nurse asap, patient leaves around the 27 Mar. Reload game, leave nurse along patient jumps out on the 31 Mar.

#### Expected Behavior
Origin TH leaves the patient in the room for the diagnosis to complete. Nurse only really required for patient to enter ward. Likewise a tired nurse should be able to leave the ward with patients.

#### System Information
**CorsixTH Version:** 0.60 + trunk

**Operating System:** Windows 7

**Theme Hospital Version:**  Origin

#### Gamelog.txt
None created

### Additional Info
None

tobylane:
https://github.com/Brainsample/CorsixTH/commit/f951589f50eea63519f88f2e7d090e331aa943ee, brought up to date.
*Fixes #1277*

- The healing factor of a nurse depends on her skill and the number of patients under her care
- When she leaves the room the patients stay, but don't heal until a nurse returns

Ping @Brainsample


CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -386,7 +386,7 @@ function Staff:checkIfNeedRest()
       self:changeAttribute("happiness", -0.001)
     end
     local room = self:getRoom()
-    if (self.staffroom_needed and ((room and not room:getPatient()) or not room)) or
+    if (self.staffroom_needed and (room and (not room:getPatient() or room.room_info.id == "ward")) or not room) or
         (room and self.going_to_staffroom) then
```
Alberth289346, line 390: An if-else with more than one 'and' and one 'or' is a sign is it too complicated.

No doubt it is correct, but it gives me a headache already without even trying to understand what it says.

Please refactor it into a bunch of simple understandable 'if' statements, of make a function that returns a boolean result for this computation.

CorsixTH/Lua/room.lua
```diff
@@ -573,16 +573,18 @@ function Room:onHumanoidLeave(humanoid)
...
+          call_for_new_staff = true
+          if self.room_info.id ~= "ward" then
+            self:makeHumanoidLeave(room_humanoid)
+            room_humanoid:queueAction(self:createEnterAction(room_humanoid))
+          end
         end
       end
       -- Call for staff if needed
-      if self.is_active and (patient_needs_to_reenter or self.door.queue:patientSize() > 0) then
+      if self.is_active and (call_for_new_staff or self.door.queue:patientSize() > 0) then
```
Alberth289346, line 587: You can replace `false` at line 576 by `(self.door.queue:patientSize() > 0)`  so you don't need it here.

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -122,11 +122,29 @@ function WardRoom:doStaffUseCycle(humanoid)
   humanoid:queueAction(MeanderAction():setLoopCallback(meanders_loop))
 end
 
+-- This function is called once per tick per ward, unless there's no patient in the ward
+function WardRoom:getHealingAmount()
+  self.patient_idx = 0 -- Will be reset to nil after handling the last patient each tick
```
Alberth289346, line 127: If for some reason it stays 0 (no patients or all are leaving).

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -135,14 +153,20 @@ function WardRoom:commandEnteringPatient(patient)
...
+      end
+      amount_to_heal = amount_to_heal - self.healing_amount
+      if amount_to_heal < 0 then -- The patient is healed or diagnosed
         action.prolonged_usage = false
       end
-      length = length - 1
+      if self.patient_idx == 1 then -- This was the last patient, set up for the next tick
+        self.patient_idx = nil -- Erase the temporary field
+      else -- Set up for the next patient
+        self.patient_idx = self.patient_idx - 1
```
Alberth289346, line 168: and is the index is 0, it decreases to -infinity here.

CorsixTH/Lua/rooms/ward.lua
Alberth289346: Please if you introduce member variables, document their existence in the constructor, instead of randomly adding new state throughout all functions. That is, in the constructor say
``` lua
self.patient_idx = nil -- With descriptive meaning of the variable.
```
I don't like the variable name btw, "index in patients" doesn't say anything about the purpose.

What happens with the index if a patient leaves?

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -122,11 +122,29 @@ function WardRoom:doStaffUseCycle(humanoid)
...
+  for humanoid in pairs(self.humanoids) do
+    if not humanoid:isLeaving() then
+      if class.is(humanoid, Patient) then
+        self.patient_idx = self.patient_idx + 1 -- Count the number of patients
+      elseif humanoid.humanoid_class == "Nurse" then
+        local nurse_quality = humanoid:getServiceQuality()
+        nurse_force = nurse_force + 1.0 + nurse_quality
+        if nurse_quality > best_staff then
+          best_staff = nurse_quality
+          self.staff_member = humanoid -- Used in \patient.lua\completeDiagnosticStep()
```
Alberth289346, line 139: So now the `completeDiagnosticStep` and `getHealingAmount` are intertwined in some complicated way. Please keep separate things separate. Let each function deal with its own computations.
Code is complicated enough already.

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -135,14 +153,20 @@ function WardRoom:commandEnteringPatient(patient)
     print("Warning: A patient was called into the ward even though there are no free beds.")
   else
```
Alberth289346, line 154: Instead of a long `else`, deal with the `if` case completely, and return from the function.
Code is simpler if it doesn't continue executing stuff 25 lines down which you can only find by finding the correspond `end`.

Also, the `else` can be dropped and the entire `else` case can be dedented 2 spaces.

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -135,14 +153,20 @@ function WardRoom:commandEnteringPatient(patient)
...
         action.prolonged_usage = false
       end
-      length = length - 1
+      if self.patient_idx == 1 then -- This was the last patient, set up for the next tick
+        self.patient_idx = nil -- Erase the temporary field
+      else -- Set up for the next patient
+        self.patient_idx = self.patient_idx - 1
+      end
     end
     local after_use = --[[persistable:ward_after_use]] function()
```
Alberth289346, line 171: Nothing wrong inserting an empty line when you start doing a next step. Such lines guide the reader on such context switches in the same way that you add an empty line between different pieces of text in a message.

CorsixTH/Lua/room.lua
```diff
@@ -573,16 +573,18 @@ function Room:onHumanoidLeave(humanoid)
...
+      local call_for_new_staff = false
       for room_humanoid in pairs(self.humanoids) do
-        if class.is(room_humanoid, Patient) and self:shouldHavePatientReenter(room_humanoid) then
-          self:makeHumanoidLeave(room_humanoid)
-          room_humanoid:queueAction(self:createEnterAction(room_humanoid))
-          patient_needs_to_reenter = true
+        if class.is(room_humanoid, Patient) and not room_humanoid:hasLeavingAction() then
+          call_for_new_staff = true
+          if self.room_info.id ~= "ward" then
+            self:makeHumanoidLeave(room_humanoid)
```
Alberth289346, line 581: Probably out of scope, but why does the room have a `makeHumanoidLeave' function, while the humanoid has a `hasLeavingAction` ?
Wouldn't it be more logical if the room asked the humanoid to leave? Ie actions are owned by the humanoid rather than owned by the room.

lewri:
I think when I played the original, the nurse was only needed for admittance. And then the patient could leave on their own.
But I feel like I might've been the person to push for having a nurse 'sign off' patients.

Alberth289346:
I like that idea, to be honest. Patient should not walk out by him/her self imho without staff present

Alberth289346:
Not healing without a nurse present is debatable imho, time is doing the healing, a good nurse just speeds up the process imho

CorsixTH/Lua/rooms/ward.lua
tobylane: It's finding the best nurse at the time, different to patient.lua which finds the one (and could find the best) staff member at the time of a diagnostic step. If one nurse leaves and a worse one comes in it should change the speed of progress? More relevant if you want progress to continue when there's no nurse in the room.

Alberth289346: My point is that you're counting patients, compute `nurse_force`, and compute `self.staff_member` all here now in one loop. This makes that all these things are now connected to each other, yet each value is independent. The number of patients doesn't change on `nurse_force`, both nurse values are not related to each other either.

By breaking this single computation in 3 separate computations (one for each result value), you're making the code more modular, and also show these values are really independent values.

The primary aim of code is to make it understandable to fellow human coders.

In the above loop anyone but you has to reverse-engineer how these variables are related currently (you're computing them together, the general assumption is that doing that has a reason.)
By splitting into several loops, you send the message "these things are independent". No need to decypher that information from the code, less chance of drawing wrong conclusions.

CorsixTH/Lua/room.lua
mugmuggy: A Humanoid:leaveRoom call would make sense, then have the patient determine if they need to use the screen to change back, instead of the makeHumanoidLeave and the makeHumanoidDressIfNecessaryThenLeave but rooms with screens normally have this driven by a callback or they are queued up upon getting sent to machine etc.

Alberth289346: Pass that into the humanoid function could work perhaps

TheCycoONE:
There was a long discussion about what to do in the issue already. I'd prefer to keep any discussion of the expected behavior there, and keep this for the PR itself (except if we want to say a PR is premature because there is still no consensus on the issue - my understanding was that we did reach some consensus in the issue though).

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -385,14 +385,20 @@ function Staff:checkIfNeedRest()
...
     end
+
     local room = self:getRoom()
-    if (self.staffroom_needed and ((room and not room:getPatient()) or not room)) or
-        (room and self.going_to_staffroom) then
-      if self:getCurrentAction().name ~= "walk" and self:getCurrentAction().name ~= "queue" then
-        self.staffroom_needed = nil
-        self:goToStaffRoom()
-      end
+    local may_leave = false
```
TheCycoONE, line 390: Doesn't match the old condition and I think this is wrong now. `staffroom_needed` use to be restricted by the other conditions (no patient if in a room) `going_to_staffroom` was not.

tobylane:
Fixed that. The time they take in bed, with varying room members, feels better now.

tobylane:
I feel it could be fine tuned endlessly but it's ready now.

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -123,36 +124,54 @@ function WardRoom:doStaffUseCycle(humanoid)
...
-    end
-    local after_use = --[[persistable:ward_after_use]] function()
-      self:dealtWithPatient(patient)
+    return
+  end
+
+  bed.reserved_for = patient
+  local healing_required = math.random(150, 600)
+  local --[[persistable:ward_loop_callback]] function loop_callback(action)
+    healing_required = healing_required - self.healing_amount
```
TheCycoONE, line 162: I think this will break on loading games that have `ward_loop_callback` in the queue because healing_required won't be defined - changing upvalues is tricky.

TheCycoONE (changes requested):
upvalues and persistence is tricky to the point where you almost don't want to deal with them at all. As the wiki says on the topic:
```
If new upvalues are added to a non-permanent persistable function (for example, the new code uses a local variable from a parent function which the old code didn't), then when an older savegame is loaded, accesses to these new upvalues will act like accesses to globals with the same names. Hence the introduction of new upvalues should be avoided if savegame compatibility is desired. Alternatively, if new upvalues need to be introduced, and savegame compatibility needs to be maintained, then a copy of the original persistable function can be made somewhere, and the new code given a new persistable name.
```
As that document suggests, you may be better off making an entirely new function with a new persisted name and leaving the old definition in the code for saves to use until those patients are done.

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -123,36 +124,63 @@ function WardRoom:doStaffUseCycle(humanoid)
...
--- TODO the nurse should not leave the ward if there are beds in use, therefore prevent her from being picked up
--- and have a system that stops patients entering the ward if she is in need of taking a break or being called elsewhere.
 function WardRoom:commandEnteringPatient(patient)
-  local staff = next(self.staff_member_set) or self.staff_member
   local bed, pat_x, pat_y = self.world:findFreeObjectNearToUse(patient, "bed")
   self:setStaffMembersAttribute("dealing_with_patient", nil)
   if not bed then
     patient:setNextAction(self:createLeaveAction())
     patient:queueAction(self:createEnterAction(patient))
     print("Warning: A patient was called into the ward even though there are no free beds.")
```
lewri, line 155: Outside of scope but if we can could we change this to a gamelog rather than a print? Though depends if it'll get in the way of #1833

CorsixTH/Lua/rooms/ward.lua
lewri: Can confirm this breaks, but looks like Toby has dealt with it now

```
Lua\rooms\ward.lua:143: use of undeclared variable 'healing_required'
stack traceback:
        [C]: in function 'error'
        Lua\strict.lua:43: in metamethod '__index'
        Lua\rooms\ward.lua:143: in method 'loop_callback'
        Lua\humanoid_actions\use_object.lua:324: in local 'timer_function'
        Lua\world.lua:1072: in method 'onTick'
        Lua\app.lua:1128: in function <Lua\app.lua:1125>
        (...tail calls...)
        Lua\app.lua:1028: in function <Lua\app.lua:1023>
```

CorsixTH/Lua/rooms/ward.lua
tobylane: Done, I'll fix it in that branch.

CorsixTH/Lua/rooms/ward.lua
tobylane: Fixed, if I understand it right. It works for me.

tobylane:
Dated the old callback and removed the upvalues.

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -123,36 +124,64 @@ function WardRoom:doStaffUseCycle(humanoid)
...
+      action.prolonged_usage = false
     end
-    local after_use = --[[persistable:ward_after_use]] function()
-      self:dealtWithPatient(patient)
+    length = length - 1
+  end
+  -- New callback
+  patient.healing_required = math.random(150, 600)
+  local --[[persistable:ward_new_loop_callback]] function new_loop_callback(action)
+    patient.healing_required = patient.healing_required - self.healing_amount
```
TheCycoONE, line 171: Patient is another upvalue. What i would suggest is setting this on the action instead (it really is a property of the action.)

I have some thoughts on formalizing a usage counter for prolonged usage in action so it's less duck type spaghetti but injecting a propery is fine for now.

tobylane:
How do you set the initial value? This way isn't right.

Alberth289346:
My first guess is to stick the `required_work` value into the `action`

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -166,12 +165,14 @@ function WardRoom:commandEnteringPatient(patient)
...
   -- New callback
-  patient.healing_required = math.random(150, 600)
+  local required_work = math.random(150, 600)
   local --[[persistable:ward_new_loop_callback]] function new_loop_callback(action)
-    patient.healing_required = patient.healing_required - self.healing_amount
-    if patient.healing_required <= 0 then -- The patient is diagnosed or healed
+    if not action.remaining_work and required_work then
+      action.remaining_work = required_work
+    end
+    action.remaining_work = action.remaining_work - self.healing_amount
```
Alberth289346, line 173: When you write an `if`, please also carefully consider _all_ possible cases.

1. `not action.remaining_work` holds and `required_work` holds, 
2. `not action.remaining_work` fails and `required_work` holds,
3. `not action.remaining_work` fails and `required_work` fails, and
4. `not action.remaining_work` holds and `required_work` fails.

In case 1, the `then` code is performed, giving `action.remaining_work` a value, and the code works.
In case 2, `action.remaining_work` already has a value, and the code works.
In case 3, `action.remaining_work` already has a value, and the code works.
In case 4, `action.remaining_work` will not get a value, and the **program crashes** at this line.

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -123,37 +124,62 @@ function WardRoom:doStaffUseCycle(humanoid)
   humanoid:queueAction(MeanderAction():setLoopCallback(meanders_loop))
 end
 
+function WardRoom:calculateHealingFactor()
```
Alberth289346, line 127: When you "calculate" something, I expect the function to return a (calculated) result.
This function doesn't, so a better name is `updateHealingFactor()`, or probably better, `updateHealingAmount()`, as the result is eventually stored in `healing_amount` (Don't use several names for the same thing, it causes confusion.)

CorsixTH/Lua/rooms/ward.lua
```diff
@@ -123,37 +124,62 @@ function WardRoom:doStaffUseCycle(humanoid)
...
+    if not humanoid:isLeaving() then
+      if class.is(humanoid, Patient) then
+        patient_count = patient_count + 1
+      elseif humanoid.humanoid_class == "Nurse" and not humanoid.fired then
+        nurse_factor = nurse_factor + 0.5 + humanoid:getServiceQuality()
+      end
+    end
+  end
+  -- Difficulty of healing - 10% faster on easier, 10% slower on hard mode
+  local difficulty_factor = 0.8 + self.world.map:getDifficulty() * 0.1
```
Alberth289346, line 140: You can move this inside the `nurse_factor > 0` case, as there is the only place where you need the value.
In that way, you don't waste computing time for values that you eventually don't use.

CorsixTH/Lua/rooms/ward.lua
TheCycoONE: This is incidentally not the way I would recommend setting the value; you're checking something every loop that only needs to be initialized once. What I would suggest instead is creating the action, setting your property on the action, and then queuing it on the patient (line 180)

The same action there is passed into your callback.

tobylane: Done, if I understood.

TheCycoONE: Yeah, that's what I meant. Only nit is that I would fully instantiate the action outside of the queueAction method now that you're starting to do it above.

```lua
local bed_action = UseObjectAction(bed)
  :setProlongedUsage(true)
  :setLoopCallback(new_loop_callback)
  :setAfterUse(after_use)
bed_action.remaining_work = math.random(150, 600)
patient.queueAction(bed_action)
```

TheCycoONE: One other nit is the word new. If we ever replace it what does the replacement get called, new new? for the persist string I suggest the mundane ward_loop_callback_2, and for the actual function, loop_callback is fine, the old one can just be given the anonymous name `_`

lewri:
@tobylane @TheCycoONE this one looks ready to be gone over again?

tobylane:
Ready.

PR #1708, opened 2020-09-13
[RDY] Align gates_to_hell use tiles with footprint orientation

mugmuggy:
Die action calculates the use positions of the gates_to_hell using the wrong orientation for the offsets, they should align for the only passable tiles.

Spawn of a gates_to_hell can change the passable flags upon placement/removal since the footprint tile might be in a room. Can walk through a toilet up against a wall or prevent placing an object if the room is being constructed.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/die.lua
+++ b/CorsixTH/Lua/humanoid_actions/die.lua
@@ -102,8 +102,8 @@ local action_die_tick_reaper; action_die_tick_reaper = permanent"action_die_tick
     local mirror_grim = 0
 
     local spawn_scenarios = {
-      {"south", humanoid.tile_x, humanoid.tile_y + 4, 0, 1, "north", 0, -1, {{after_spawn_idle_direction = "east", hole_x_offset = -5, hole_y_offset = 2}, {hole_x_offset = 0, hole_y_offset = 3}} },
-      {"east", humanoid.tile_x + 4, humanoid.tile_y, 1, 0, "west",  -1,  0, {{hole_x_offset = 3, hole_y_offset = 0}} }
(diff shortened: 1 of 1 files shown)


mugmuggy:
Also added checks around the accessibility of the patient and grim use tile as it maybe possible the findIdleTile finds an isolated section of the map. Also adjusted the mirroring so the grim swipes whilst facing the patient. Also the direction for when the grim spawns on the use tile, ie scenarios where he can't walk

TheCycoONE:
Change looks good to me.

CorsixTH/Lua/humanoid_actions/die.lua
```diff
@@ -114,15 +114,20 @@ local action_die_tick_reaper; action_die_tick_reaper = permanent"action_die_tick
...
         humanoid.hole_use_tile_x = hole_x + spawn_scenario[7]
         humanoid.hole_use_tile_y = hole_y + spawn_scenario[8]
-        -- tile can't be in a room
-        if humanoid.world:getRoom(humanoid.hole_use_tile_x, humanoid.hole_use_tile_y) then
+        -- tile can't be in a room and must be accessible by the patient
+        if not humanoid.world:getPathDistance(humanoid.tile_x, humanoid.tile_y, humanoid.hole_use_tile_x, humanoid.hole_use_tile_y)
+            or humanoid.world:getRoom(humanoid.hole_use_tile_x, humanoid.hole_use_tile_y) then
+          return false
+        end
+        -- ensure grim won't be in a a room
```
lewri, line 129: Small typo here "a a"

TheCycoONE: @mugmuggy missed this one ^?

mugmuggy: My bad, I also observed something I need to check related to lava hole. Completely unrelated to this change, so I might see if I can replicate that and include that with this.

mugmuggy: I wasn't able to replicate the error, maybe it was a different save than I was thinking, playing around with a few different ones when it occured, but the error was around persistence in player_hospital line 272. So I guess progress with this anyway.

PR #1709, opened 2020-09-14
[rdy] Rework boiler disasters.

### LINKED ISSUE #490: Boiler breakdown rate should depend on stuff

Alberth289346:
Fixes #490 
Closes #1345

- Implements the boiler breakdown as described in issue #490
- Code in #1345 did too much, so I implemented it again.
- Adds starting point for a vomit wave disaster, which we don't seem to have yet.

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -98,7 +98,16 @@ function Hospital:Hospital(world, avail_rooms, name)
   -- level; e.g. Easy: 0.4 / Difficult: 0.2)
   self.over_priced_threshold = 0.3
 
-  self.radiator_heat = 0.5
+  -- (int) Number of days until the next boiler or vomit wave disaster.
+  -- TODO: Implement the vomit wave.
+  self.disasterless_days = self:daysTillNextDisaster()
+
+  self.curr_setting = nil -- (float) Saved radiator heat when boiler has broken down.
+  self.boiler_can_break = nil -- (bool) Prevents boiler breakdown before the hospital is open.
+  self.heating_broke = nil -- (bool) Whether the boiler is broken down currently.
+  self.boiler_countdown = nil -- (int) Number of days before the heating works properly again.
+  self.radiator_heat = 0.5 -- (float) [0..1] fraction of heating by a radiator.
+
   self.num_visitors = 0
   self.num_deaths = 0
   self.num_deaths_this_year = 0
@@ -150,7 +159,7 @@ function Hospital:Hospital(world, avail_rooms, name)
   self.ownedPlots = {1} -- Plots owned by the hospital
   self.ratholes = {} -- List of table {x, y, wall, parcel, optional object} for ratholes in the hospital corridors.
   self.is_in_world = true -- Whether the hospital is in this world (AI hospitals are not)
-  self.opened = false
+  self.opened = false -- Whether the hospital is opened (timer was cleared or ended).
   self.transactions = {}
   self.staff = {}
   self.patients = {}
@@ -680,6 +689,10 @@ function Hospital:afterLoad(old, new)
     self.hosp_cheats = Cheats(self.world.ui)
   end
 
+  if old < 142 then
+    self.disasterless_days = self:daysTillNextDisaster()
+  end
+
   -- Update other objects in the hospital (added in version 106).
   if self.epidemic then self.epidemic.afterLoad(old, new) end
   for _, future_epidemic in ipairs(self.future_epidemics_pool) do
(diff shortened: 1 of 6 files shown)


CorsixTH/Lua/hospital.lua
```diff
@@ -98,7 +98,16 @@ function Hospital:Hospital(world, avail_rooms, name)
   -- level; e.g. Easy: 0.4 / Difficult: 0.2)
   self.over_priced_threshold = 0.3
 
-  self.radiator_heat = 0.5
+  -- (int) Number of days until the next boiler or vomit wave disaster.
+  -- TODO: Implement the vomit wave.
+  self.disasterless_days = self:daysTillNextDisaster()
+
+  self.curr_setting = nil -- (float) Saved radiator heat when boiler has broken down.
```
TheCycoONE, line 105: curr_setting seems a bit vague? For earthquake I put all the disaster related parameters in one table to separate them from the other world parameters - might be nice to do here too. https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/world.lua#L100

TheCycoONE (commented):
Looks good otherwise, thanks for linking to relevant discussions.

Alberth289346:
Moved the variables into a table, and dropped the `hospital.boiler_can_break` variable, as it is always equal to `hospital.opened`.

CorsixTH/Lua/hospital.lua
```diff
@@ -915,18 +942,37 @@ function Hospital:hotWarning()
...
--- It will remain broken for a certain period of time.
-function Hospital:boilerBreakdown()
-  self.curr_setting = self.radiator_heat
-  self.radiator_heat = math.random(0, 1)
-  self.boiler_countdown = math.random(7, 25)
 
-  self.heating_broke = true
+--! Decide how many days the hospital functions within specification.
+--!return (int) Number of disaster-free days in the hospital.
+function Hospital:daysTillNextDisaster()
```
lewri, line 948: I'm not a fan of "Till", I would argue "Until" is better but that's a personal preference

CorsixTH/Lua/hospital.lua
```diff
@@ -936,12 +982,32 @@ function Hospital:boilerBreakdown()
...
-  self.heating_broke = false
-  if self:isPlayerHospital() then
-    self.world.ui.adviser:say(_A.boiler_issue.resolved)
+--! Boiler broke down and work is done to get it fixed.
+function Hospital:_fixBoiler()
+  local heat_vars = self.heating
+
+  if not heat_vars.heating_broke then return end -- Not broken, done!
+
+  -- Repair the boiler or radiators, more handyman speeds up repair, see also github #490
```
lewri, line 991: handymen

CorsixTH/Lua/hospital.lua
```diff
@@ -936,12 +982,32 @@ function Hospital:boilerBreakdown()
...
+  local num_radiators = self:countRadiators()
+  local num_handyman = self:countStaffOfCategory("Handyman")
+  if num_radiators < 5 * num_handyman then
+    heat_vars.boiler_repair_count = heat_vars.boiler_repair_count - 3
+  elseif num_radiators < 8 * num_handyman then
+    heat_vars.boiler_repair_count = heat_vars.boiler_repair_count - 2
+  elseif num_handyman > 0 then
+    heat_vars.boiler_repair_count = heat_vars.boiler_repair_count - 1
+  end
+  -- No repair without handyman!
```
lewri, line 1001: I have a concern that we get stuck in a state of permanent breakage here. Did this happen in the original?
I feel we should add an adviser message if this case happens (though very unlikely).

lewri (commented):
General nitpick comments, chances are you can ignore them if you want.

CorsixTH/Lua/hospital.lua
Alberth289346: resolved, boiler can now repair itself

CorsixTH/Lua/hospital.lua
Alberth289346: I always have a bit of fun with this word :)

CorsixTH/Lua/hospital.lua
Alberth289346: Fixed

Alberth289346:
All tweaked, thanks

PR #1711, opened 2020-09-16
[rdy] Create player hospital

Alberth289346:
*Fixes the future

- Introduces `PlayerHospital`, a derived class from `Hospital`
- Moves the hospitals into their own folder.
- Deletes the `World:initCompetitors` function as it is used one time only during a level.

`PlayerHospital` is still empty, aiming to get the basics right. In particular, starting a new game works, but how do you modify a loaded savegame to use the new class? (especially in the future when it's going to have its own data (I would guestimate).


TheCycoONE:
I handled breaking a class into subclasses and save games before.

See: https://github.com/CorsixTH/CorsixTH/blob/8722145d23d400d198bb709e1032e941ca516ec6/CorsixTH/Lua/entities/humanoids/staff.lua#L650

Alberth289346:
Nice. That doesn't run the constructor does it, so variables don't get initialized?
(which is fine currently as I don't have any yet)

TheCycoONE:
No it doesn't - the new objects would inherit all their current values, though you could take care of them in the afterLoads of the respective objects.

PR #1714, opened 2020-09-17
[rdy] Fix blue taint issue

### LINKED ISSUE #1707: [Bug] Blue screen while checking out the rewards
#### Describe the issue
Blue screen while checking out the rewards.

#### Steps to Reproduce
1. Play the game, save just before the end of the year.

2. At the end of the year, the game stops and overlays the screen with a blue taint, and displays the rewards you received.

3. You can inspect the rewards (clicking on them works), although reading is a bit difficult due to the the blue colour.

4. If you close the reward window, the blue taint disappears and the game continues.

#### Expected Behavior
While I don't like the blue taint at all, I can see the use. However, the rewards screen is fully functional, and the player even need to click on the close button, so at least that area should not be blue.

#### System Information
**CorsixTH Version:** *2dcb109e* currently, but that won't get into the repoistory.

**Operating System:** *Linux Ciinamon Mint 20*

**Theme Hospital Version:** *CD*

### Additional Info
Screenshot from 2020-09-13 13-02-47

lewri:
*Fixes #1707*

- Makes it so launching a "mustPause" window does not cause blue taint.

### DIFF
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1031,6 +1031,7 @@ function UI:addWindow(window)
   end
   if window:mustPause() then
     self.app.world:setSpeed("Pause")
+    self.app.video:setBlueFilterActive(false) -- mustPause windows shouldn't cause tainting
   end
   if window.modal_class == "main" or window.modal_class == "fullscreen" then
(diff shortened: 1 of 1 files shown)


TheCycoONE:
This is another area that could use refinement.

PR #1724, opened 2020-09-27
Fix reception desk placement on pause

maxxk:
When placing reception desk while being on pause (and in some other conditions, not sure which), non-existent in class Hospital field game_date causes an exception. The exception in its turn disconnects mouseup handler which makes the game unplayable.


lewri (approved):
Not sure why we're caching a variable ``game_date`` if we never use it again. Maybe it was a remnant of old code from when we did.

Edit: I'm okay with this but looks like game_date was moved to world. i.e. ``self.world.game_date``

maxxk:
Ok, I can change it. However, other instances of `getMonthOfGame` and `getMonthOfYear` in this file use `self.world:date()`.

lewri:
Found the reasons why this appeared and happy to continue with this as is :)

PR #1725, opened 2020-09-28
Always query the game date from world:date()

Alberth289346:
*Fixes external use of `world.game_date`

- Replace the last few uses of `world.game_date` by `world:date()`.
- Document the meaning of `world.game_date`.


lewri:
Looks fine. Just need to fix the busted test for announcer

TheCycoONE:
LGTM outside of the test failures that lewri mentioned.

Future improvement might be to make date immutable so that we don't need to clone it as much (out of scope here)

lewri (approved):

PR #1726, opened 2020-10-01
[RDY] Adds option to remove destroyed rooms for a fee

### LINKED ISSUE #969: Game error and glitch: doctor sprite stuck between doors
There was a sudden error and then I noticed a doctor inside a room being behind a door and in front of another (which should not be there), and the sprite is frozen.
[Screenshot here](https://i.imgur.com/uCY0YB4.jpg)

[gamelog.txt](https://github.com/CorsixTH/CorsixTH/files/109006/gamelog.txt)
[stuckdoctor.zip](https://github.com/CorsixTH/CorsixTH/files/109009/stuckdoctor.zip)

### LINKED ISSUE #991: Cleanup/remove exploded rooms
What is it you would like in the game?
The ability to delete broken rooms. Ideally this should be a non-default game setting.

What version of the product are you using? On what operating system?
Windows 10, latest production release of CorsixTH installed ontop of the GOG version.

mugmuggy:
*Fixes #991*
*Fixes #969*

- Adds option to customize dialog to turn on the option to remove destroyed rooms
- Costs twice as much minus some 'scrap' recovery to remove the room.
- Destroyed rooms now reduce hospital value appropriately

### DIFF
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -107,7 +107,7 @@ function GameUI:setupGlobalKeyHandlers()
      [tostring(self.app.hotkeys["ingame_scroll_up"])] = {x = 0, y = -10},
      [tostring(self.app.hotkeys["ingame_scroll_down"])] = {x = 0, y = 10},
      [tostring(self.app.hotkeys["ingame_scroll_left"])] = {x = -10, y = 0},
-     [tostring(self.app.hotkeys["ingame_scroll_right"])]	= {x = 10, y = 0},
+     [tostring(self.app.hotkeys["ingame_scroll_right"])] = {x = 10, y = 0},
   }
 
   -- This is the long version of the shift speed key.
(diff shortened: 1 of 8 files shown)


CorsixTH/Lua/game_ui.lua
```diff
@@ -643,10 +643,82 @@ function GameUI:onMouseUp(code, x, y)
...
-          self:setCursor(self.waiting_cursor)
-          self.edit_room = room
-          room:tryToEdit()
+        if room then
+          if not room.crashed then
+            self:setCursor(self.waiting_cursor)
+            self.edit_room = room
+            room:tryToEdit()
+          else
+            if self.app.config.remove_destroyed_rooms then
```
Alberth289346, line 652: Fold this new code into a function, probably even more than one. Having to find a matching 'end' 80 lines down is a mess!

Alberth289346:
I would consider this to be a cheat, at least wrt original behavior.

Maybe we should reconsider how we want to treat behavior extensions in general? Currently we just add it all in the configuration file, but maybe not all options are the same "kind"?

CorsixTH/Lua/game_ui.lua
mugmuggy: Indeed, it was just a slightly massaged guts of UIEditRoom:returnToDoorPhase, so can probably massage another method there to cover both scenarios.

mugmuggy:
I considered it the same as build on pause, something that varied from the original giving the player some sort of an advantage, a plain configuration item.

By using twice the room cost, for the cheaper rooms, its nearly a break even opportunity for the most part to just expand to another small plot, more expensive rooms its more certainly a given to just expand.

The other advantage it gives by freeing up that space is the value can go up again, with this change subtracting away the crashed room value - which it should have done anyway. That gain is generally neutralised if the same room is built compared to not subtracting the value like occurs now.

Due to research reducing machine cost, the more researched a room is between building it and removing it further down the track, the buying another plot scenario diminshes, with the caveat that you don't want to expand anyway. Though if you have two or more crashed rooms, expansion is probably more cost effective in the short term.

mugmuggy:
I'll squash this before merge if there are no other changes required

TheCycoONE:
LGTM, I think some players will like this a lot.

lukejames1111:
In the dialogue box I wish it told you what room you are about to destroy. After a massive earthquake I forget what rooms need rebuilding. If it said "Would you like to remove X-Ray for..." it would make it easier for me to know what to rebuild.

lewri:
> In the dialogue box I wish it told you what room you are about to destroy. After a massive earthquake I forget what rooms need rebuilding. If it said "Would you like to remove X-Ray for..." it would make it easier for me to know what to rebuild.

No need to necro a merged PR. :) Make a feature request in the issues section and it can be addressed on its own merits then

PR #1727, opened 2020-10-03
[RDY] Handymen announcements more often

### LINKED ISSUE #1649: "Handyman to repair x" announcements rarely trigger
#### Describe the issue
While testing a lot of earthquakes I noticed that the receptionist wasn't announcing machines that were in need of urgent repair (i.e. Handyman to repair x). It also appeared in some other saves that the task priority wasn't being bumped in the calls dispatcher (some handymen would still action low priority tasks over heightened ones).

This also doesn't seem specific to earthquakes, if regular machine use takes it over the threshold it still fails to announce.

There's only been one save I managed to get the announcements on. And I think that was from 0.60 or something. So maybe something has broke along the way since then?

#### Steps to Reproduce

2. Make some earthquakes happen, may need to reduce handymen numbers.

3. Watch machines with smoke not triggering any announcements.

#### Save Game
https://www.dropbox.com/s/5akhnc9azhbaxyf/vipfor138.sav?dl=0

#### Expected Behavior
Receptionist should be announcing machines in disrepair.

#### System Information
**CorsixTH Version:** 0.64

**Operating System:** W10

lewri:
*Fixes #1649*

- Handyman repair announcements now trigger if base strength was >6. Originally they only triggered if base strength was <4
- Also ensures earthquakes don't make the announcer get spammed.

I expect some tidying up needed on this. At the moment this was to get it out of my old test environment.

### DIFF
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -75,18 +75,11 @@ end
 function Machine:announceRepair(room)
   local sound = room.room_info.handyman_call_sound
   local earthquake = self.world.next_earthquake
-  -- If an earthquake is happening limit the amount of machine warnings
-  if earthquake.active and earthquake.warning_timer == 0 then
-    if not earthquake.machwarn_trigger then
-      self.world.ui:playAnnouncement("machwarn.wav", AnnouncementPriority.Critical)
-      earthquake.machwarn_trigger = true -- resets with each damage phase
-    end
-  else
-    -- TODO: Don't announce handyman call sound if there are no handymen
-    -- Must be same priority
-    self.world.ui:playAnnouncement("machwarn.wav", AnnouncementPriority.Critical)
-    if sound then self.world.ui:playAnnouncement(sound, AnnouncementPriority.Critical) end
-  end
+  self.world.ui:playAnnouncement("machwarn.wav", AnnouncementPriority.Critical)
+  -- If an earthquake is happening don't play the call sound to prevent spamming
+  if earthquake.active and earthquake.warning_timer == 0 then return end
+  -- TODO: Don't announce handyman call sound if there are no handymen
+  if sound then self.world.ui:playAnnouncement(sound, AnnouncementPriority.Critical) end
 end
 
 --! Set whether the smoke animation should be showing
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -87,7 +87,7 @@ function Machine:announceRepair(room)
     if sound then self.world.ui:playAnnouncement(sound, AnnouncementPriority.Critical) end
   end
 end
-  
+
 --! Set whether the smoke animation should be showing
 local function setSmoke(self, isSmoking)
   -- If turning smoke on for this machine
@@ -179,8 +179,8 @@ function Machine:machineUsed(room)
       self.hospital:addHandymanTask(self, "repairing", 2, self.tile_x, self.tile_y, call)
       self:announceRepair(room)
     else -- Otherwise the task is already queued. Increase the priority to above that of machines with at least 4 uses left
-      local subTable = self.hospital:findHandymanTaskSubtable("repairing")
-      if subTable[taskIndex].priority == 1 then
+       -- Upgrades task from low (1) priority to high (2) priority
+      if self.hospital:getHandymanTaskPriority(taskIndex, "repairing") == 1 then
         self.hospital:modifyHandymanTaskPriority(taskIndex, 2, "repairing")
         self:announceRepair(room)
       end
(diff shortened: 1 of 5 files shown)


CorsixTH/Lua/entities/machine.lua
```diff
@@ -156,7 +160,23 @@ function Machine:machineUsed(room)
...
+        if earthquake.active and earthquake.warning_timer == 0 then
+          if not earthquake.machwarn_trigger then
+            self.world.ui:playAnnouncement("machwarn.wav", AnnouncementPriority.Critical)
+            earthquake.machwarn_trigger = true -- resets with each damage phase
+          end
+        else
+          -- Must be same priority
+          self.world.ui:playAnnouncement("machwarn.wav", AnnouncementPriority.Critical)
+          if sound then self.world.ui:playAnnouncement(sound, AnnouncementPriority.Critical) end
+        end
```
Alberth289346, line 177: This code looks duplicate to the calls dispatcher addition.

Don't clone code. If it contains a bug, or it should be modified in the future, you have practically a guarantee that not all clones will be modified in the same way, thus creating bugs, as the same thing works subtly different depending on the code path taken.

Instead, make a function for it, and call the function from every spot. Less code, no duplication, no missing of cloned code, less bugs

lewri:
Latest commit seems to have broken something in persistence or the dispatcher. Not sure what

CorsixTH/Lua/entities/machine.lua
```diff
@@ -155,8 +177,13 @@ function Machine:machineUsed(room)
     if taskIndex == -1 then
       local call = self.world.dispatcher:callForRepair(self, true, false, true)
       self.hospital:addHandymanTask(self, "repairing", 2, self.tile_x, self.tile_y, call)
+      self:announceRepair(room)
     else -- Otherwise the task is already queued. Increase the priority to above that of machines with at least 4 uses left
-      self.hospital:modifyHandymanTaskPriority(taskIndex, 2, "repairing")
+      local subTable = self.hospital:findHandymanTaskSubtable("repairing")
+      if subTable[taskIndex].priority == 1 then
+        self.hospital:modifyHandymanTaskPriority(taskIndex, 2, "repairing")
```
Alberth289346, line 184: The machine shouldn't hack the internals of the handyman task administration of the hospital.
Instead, wrap this in a new handymanTask function. In other words, be nice, *ask* the hospital to change the priority for you.

The goal is to keep the amount of code that changes some data structure as local and small as possible, with a well-defined (function) interface. In that way, you can change such a part, or move such code by modifying one or two files, instead of 100 files.

Alberth289346: Somewhat the same is your earthquake queries above.
That is however more of an edge case. I am not sure what part of the data is public and what is internal.

CorsixTH/Lua/world.lua
```diff
@@ -647,6 +647,7 @@ function World:tickEarthquake()
 
       self.next_earthquake.remaining_damage = self.next_earthquake.remaining_damage - 1
       self.next_earthquake.damage_timer = self.next_earthquake.damage_timer + earthquake_damage_time
+      self.next_earthquake.machwarn_trigger = false
```
Alberth289346, line 650: You added a new variable, its meaning should be documented (see near line 101 in `world.lua`).

CorsixTH/Lua/world.lua
```diff
@@ -99,6 +99,7 @@ function World:World(app)
   -- remaining_damage (integer) The amount of damage this earthquake has yet to inflict.
   -- damage_timer (integer) The number of hours until the earthquake next inflicts damage if active.
   -- warning_timer (integer) The number of hours left until the real damaging earthquake begins.
+  -- machwarn_trigger (boolean) True if a machine has already announced it is breaking during a damage phase (resets each phase)
```
lewri, line 102:
```suggestion
  -- machwarn_trigger (boolean) True if any machine has already announced it is breaking during a damage phase (resets each phase)
```

CorsixTH/Lua/entities/machine.lua
mugmuggy: I have a feeling having just these 2 priorities for repair tasks is a little lacking when the plants have 5. I'm not expecting to address this now, but maybe something to consider later

TheCycoONE:
Not really a fan of this implementation; I don't think earthquakes need to be special cased, and I especially don't like leaking the warning flag into the world earthquake structure - that's not local to the player.

Seems to me the problem of not playing the same announcement too often could be generalized and attached to the announcement queue system.

lewri:
The announcer should be able to handle duplication of machwarn by itself actually. I'll need to look at how the duplication checker works there (even though I wrote it 😛  )
If it's good enough then we don't need to limit machwarns inside machine.

Issue #1173, opened 2016-12-07
Dynamic information showing incorrect information


MarkL1961:
Dynamic information is not always correct.

In this case it should not be showing diagnosis progress it should be something like fed up and going home  
![fedup-going home](https://cloud.githubusercontent.com/assets/4965487/20983667/39fc44b2-bcb5-11e6-8b69-a8ef86d396fb.png)

MarkL1961:
This time it says he is queuing for GP office and did so when he was sat on bench, but he is in x-ray!

![queuing for gp office](https://cloud.githubusercontent.com/assets/4965487/20984152/025ae9e4-bcb7-11e6-8908-2f4164a8bc00.png)

MarkL1961:
Now he is going home for some reason, but it still says diagnosis progress.
![going home](https://cloud.githubusercontent.com/assets/4965487/20984303/8b05c8e0-bcb7-11e6-9682-89e3f46573d2.png)

TheCycoONE:
So is it a matter of simply not updating often enough?

MarkL1961:
How often does it update?
I thought it was more or less instant, but I started to notice it showing the wrong information so started taking screenshots

TheCycoONE:
It is instant, but not necessarily every action updates it.

MarkL1961:
Interesting, they should do.  Especially things like going home as you need to know why

MarkL1961:
![cured and fed up](https://cloud.githubusercontent.com/assets/4965487/21468559/cf99956c-ca0f-11e6-8b31-93c2284ccc2f.png)
Spotted someone who is cured and also fed up and leaving, I don't think this should be possible.  You can leave at any point up to being treated and then you either leave cured or dead

mugmuggy:
This last issue, I just seen happen when I was mucking about so took a deeper dive - World:onEndDay calls Patient:tickDay.

It has this bit of code
https://github.com/CorsixTH/CorsixTH/blob/a549a809217a135d51214eba8f33d263178d90e6/CorsixTH/Lua/entities/patient.lua#L637-L646

I'm guessing as onEndDay gets called in World:onTick, whilst the treatment in this case Pharmacy is underway when health meets that criteria and that random value condition is also true. This sets the mood to exit.

The treatment is finished in the Room:dealtWithPatient is being called updating the cured state etc. When it calls Patient:goHome again, its already set to the onEndDay/tickDay mood and not updating the value as the patient is already going_home.

[...]

PR #1732, opened 2020-10-09
[RDY] Emergency balance changes

mugmuggy:
*Fixes # nothing in particular*

- Emergency timers are around 50 days, I just made it 52 for the even segmentation for the timer
- Helicopter arrives later, more consistent with the original and the patients spawn slightly further apart, half a day
- The ability for emergency patients to leave due to cost. - Corsix allows you to accept an emergency and increase the cost and patients just pay it. ~~Likewise this change locks the last accepted price in, thus all emergency patients pay the same. These are features of the original.~~ 
image
- sad_money priority bumped up so shows on refusing to pay emergency patients, but shouldn't on epidemy# ones when they leave (same priority)
- different to the original - if you kick or they refuse to pay - the timer is cut short at the earliest opportunity
- dynamic info displays 'Emergency: %s' instead of 'Diagnosed: %s' and was the reason I just jumbled Helicopter:spawnPatient so it was set initially to Emergency instead of having to update it due to the issue below it wasn't technically required
- Addressed the comment https://github.com/CorsixTH/CorsixTH/issues/1173#issuecomment-630137026
- Patient:goHome updates to use existing local variable and reflect the original with dynamic info in the evacuated/over priced of using the middle text line

The patient:agreesToPay I just manipulated for testing, I didn't run it through a natural run to see how discerning it is.

### DIFF
--- a/CorsixTH/Lua/dialogs/watch.lua
+++ b/CorsixTH/Lua/dialogs/watch.lua
@@ -25,6 +25,10 @@ class "UIWatch" (Window)
 ---@type UIWatch
 local UIWatch = _G["UIWatch"]
 
+local TICK_DAYS = 100
+local TICK_DAYS_EMERGENCY = 52
+local TIMER_SEGMENTS = 13
+
 --!param count_type (string) One of: "open_countdown" or "emergency" or "epidemic"
 function UIWatch:UIWatch(ui, count_type)
   self:Window()
@@ -33,8 +37,13 @@ function UIWatch:UIWatch(ui, count_type)
 
   self.esc_closes = false
   self.modal_class = "open_countdown"
-  self.tick_rate = math.floor((100 * Date.hoursPerDay()) / 13)
-  self.tick_timer = self.tick_rate  -- Initialize tick timer
+  if count_type == "emergency" then
+    self.tick_rate = math.floor((TICK_DAYS_EMERGENCY * Date.hoursPerDay()) / TIMER_SEGMENTS)
+    self.tick_timer = self.tick_rate
+  else
+    self.tick_rate = math.floor((TICK_DAYS * Date.hoursPerDay()) / TIMER_SEGMENTS)
+    self.tick_timer = self.tick_rate  -- Initialize tick timer
+  end
   self.open_timer = 12
   self.ui = ui
   self.hospital = ui.hospital
--- a/CorsixTH/Lua/objects/helicopter.lua
+++ b/CorsixTH/Lua/objects/helicopter.lua
@@ -47,7 +47,7 @@ function Helicopter:Helicopter(world, object_type, hospital, direction, etc)
   self:Object(world, object_type, x, y, direction, etc)
   self.th:makeInvisible()
   self:setPosition(0, -600)
-  self.phase = -60
+  self.phase = -120
   self.hospital = hospital
   -- TODO: Shadow: 3918
   hospital.emergency_patients = {}
@@ -61,7 +61,7 @@ function Helicopter:tick()
   elseif phase == 60 then
     self:setSpeed(0, 0)
     self.spawned_patients = 0
-  elseif phase == 80 then
+  elseif phase == 85 then
     if self.spawned_patients < self.hospital.emergency.victims then
       self:spawnPatient()
       phase = 60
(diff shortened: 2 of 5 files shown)


CorsixTH/Lua/dialogs/watch.lua
```diff
@@ -65,6 +69,9 @@ function UIWatch:UIWatch(ui, count_type)
     self.end_button = self:addPanel(end_sprite, 4, 0)
       :makeButton(4, 0, 27, 28, end_sprite + 1, self.onCountdownEnd)
       :setTooltip(tooltips[count_type])
+  elseif count_type == "emergency" then
```
TheCycoONE, line 72: Just else, or probably better move this condition up and have the currently ~= "emergency" condition as the else clause.

TheCycoONE: Actually I think it'd be better to do this emergency clause for tick_rate up where tick_rate is defined, instead of in this block which is deciding the end_button.

CorsixTH/Lua/objects/helicopter.lua
```diff
@@ -82,17 +82,25 @@ function Helicopter:spawnPatient()
...
+  hospital.emergency_patients[self.spawned_patients] = patient
   local x, y = hospital:getHeliportSpawnPosition()
   patient:setNextAction(SpawnAction("spawn", {x = x, y = y}):setOffset({y = 1}))
   patient:setHospital(hospital)
   -- TODO: If new combined diseases are added this will not work correctly anymore.
   patient.cure_rooms_visited = #patient.disease.treatment_rooms - 1
   local no_of_rooms = #patient.disease.treatment_rooms
-  patient:queueAction(SeekRoomAction(patient.disease.treatment_rooms[no_of_rooms]))
+  -- a spawning emergency patient might rather leave than pay
+  if not patient:agreesToPay(patient.disease.id) then
```
TheCycoONE, line 96: agreesToPay here is based on the current price, which - if not the first patient (as set in your condition below) might not match.

Better to set the emergency treatment_price before this check and perhaps pass it into agreesToPay (which would then have to filter through priceDistortion)

TheCycoONE:
Another approach might be to check the agreesToPay for each patient up front instead of on the tick they come out. That would be a more contained change than passing price throughout.

CorsixTH/Lua/dialogs/watch.lua
mugmuggy: I'll rejig

CorsixTH/Lua/objects/helicopter.lua
mugmuggy: TH does it that way, sets the price then calls the equivalent of agreesToPay based on the current price, not the emergency price. In priceDistortion, there are 3 factors, happiness is equal to 1, if the there is no effectiveness for the treatment its '100%' and takes it away so then just becomes a weighting on the reputation - proxy for price anyway. Effectiveness is proxied into reputation as well as more effective less chance of death, higher rep.

TH agreesToPay equivalent is a little different as its only taking into account the 'fair' cost (price * rep / 500) then checks to see if the 'asking' price is less than random('fair' cost, 1.5 * 'fair' cost) and if it is agree. The other mechanic in TH which we don't have, is that when insurance pays 'fair' cost is inflated by a percentage which is unique per insurance company. And insurance companies pay out at different times as well. I haven't checked to see if the inflated prices may actually be agreed on or just added to the insurance payouts, I assume they should.

The other part of agreesToPay in TH (not on emergnecy), is when the patients agree to pay, they are also agreeing on a price that may change in favour or not, rather than the price agreed. TH saves the agreed to price on the patient and thats what is paid. Thus after that decision is made you are free to change the price in Corsix with the only ramification that you alter the perception for new patients, but you would have to be keen to do this on a persistent basis, maybe if your room crashed and you had patients waiting again you can do this trick to recover some build cost nearly immediately.

mugmuggy:
I double checked and the price can vary, naturally (reputation) as well as through changing the price and they may or may not accept it and it is done at the time of spawn. I removed the code making a consistent fee, which still leaves it exploitable for manipulation at this stage.

Though in the future I think some scrutiny of agreesToPay is probably worthwhile.

TheCycoONE:
Could you rebase to fix the conflict.

PR #1735, opened 2020-10-11
[rdy] Move more advice to player hospital

Alberth289346:
*Fixes Move more advice and feedback to the player hospital.

- `cashLow` was previously merged into another function, but not deleted in the original code.
- `countPatients` ran every tick, and was only used just below for generating background noise, and in the townmap.
  Now, it's run around 1% of the time for background noise, and has better performance for busy hospitals. Independently, it may also run from the townmap.
- Background noise code is now in player hospital
- Advice code about receptionsists or desks is now in player hospital
- Advice code about plants wrt handman is now in player hospital
- Gates to hell animation code is now in player hospital

Best reviewed one commit at a time.

### DIFF
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -253,6 +253,15 @@ function PlayerHospital:msgReceptionDesk()
   end
 end
 
+--! Give advice to the user about maintenance of plants.
+function PlayerHospital:msgPlant()
+  local num_handyman = self:countStaffOfCategory("Handyman")
+
+  if num_handyman == 0 then
+    self:giveAdvice({_A.staff_advice.need_handyman_plants})
+  end
+end
+
 --! Advises the player.
 --!param msgs (array of string) Messages to select from.
 --!param rnd_frac (optional float in range (0, 1]) Fraction of times that the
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -192,29 +198,126 @@ function PlayerHospital:monthlyAdviceChecks()
...
+end
+
+-- Update the hospital each tick.
+function PlayerHospital:tick()
+  -- Add some random background sounds, ringing phones, coughing, belching etc.
+  if math.random(1, 100) == 3 then
+    -- Wait until there are some patients in the hospital and a room, otherwise you
+    -- will wonder who is coughing or who is the receptionist telephoning!
+    -- Opted for gp as you can't run the hospital without one.
+    if self:countRoomOfType("gp") > 0 and self:countPatients(3) > 2 then
```
TheCycoONE, line 302: This sounds stuff doesn't belong here - I'm not quite sure where it belongs, but (as far as I can remember) you can hear the environmental sounds of other hospitals in multiplayer.

Alberth289346: Moved the code back while keeping the perfoamnce improvements, and added a TODO multiplayer note.

PR #1739, opened 2020-10-21
[WIP] Add queue warnings, skip level board movie on restarts, fix edge scrolling bugs, shortern win faxes

### LINKED ISSUE #474: The handyman can water a plant through the outer wall
_From [Grim...@gmail.com](https://code.google.com/u/118083579813208703807/) on July 26, 2011 18:50:14_

The handyman can water a plant through the outer wall
See attached screenshot and savegame.

What version of CorsixTH are you using (e.g. "Beta 2", "SVN r123 ")? r1353 What operating system / compile settings are you using? Win7 64

Level 11

**Attachment:** [wateringplantthroughwall.png watering.sav](http://code.google.com/p/corsix-th/issues/detail?id=955)

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=955_

### LINKED ISSUE #525: Save a savegame when you're move around the map; map will still move if you open it
_From [maartenp...@hotmail.nl](https://code.google.com/u/103280305005036074093/) on February 10, 2013 13:48:33_

What steps will reproduce the problem? 1.Move at the map (scroll) and save the game (with ''enter'') at the same time.
2.Open that savegame (like the one attached.)
3.The scrolling goes on when opening it. What is the expected output? What do you see instead? See above. I'm using the latest snapshot.

**Attachment:** [test scroll.sav](http://code.google.com/p/corsix-th/issues/detail?id=1504)

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1504_

### LINKED ISSUE #526: Bad for new players: Unable to place new receptionists on top of the desk
_From [Tustin2...@gmail.com](https://code.google.com/u/101022002590040899166/) on February 14, 2013 21:43:54_

What steps will reproduce the problem? 1. Place down a Receptionist Desk
2. Hire a Receptionist, and place her down on the desk. What is the expected output? What do you see instead? Actual: When placing her on the desk, on the square she normally operates from, she is greyed out as if she can't be placed there. Upon clicking, she vanishes from your mouse pointer and the desk queue dialog appears. She returns to the hiring list, but at the end of the list instead of where she was originally.
Expected (ie How the original works): When placing her on the desk, on the square she normally works from, she CAN be placed and is not greyed out. If attempting to place her somewhere on the desk where she can't be placed (such as on the corner) she is greyed out and clicking will blink her once, as if attempting to place her, but returns her to the player's cursor immediately. The desk queue never appears during this.

Release Version 0.11 What operating system / compile settings are you using? Win 7 64x, Installed

First career level Attach a saved game file (preferably an autosave from before the error) if you have one and it is relevant. Please provide any additional information below.

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1515_

### LINKED ISSUE #1311: Edge scrolling not saved in config when changed by user
In #1310 we noticed edge scrolling is not saving into the config (`menu.lua : 640`) . 

I suspect that other options might also have similar issues. Separately each option in `menu.lua` needs to be checked that is sets the right variable in `app` and that the save method is invoked if applicable.

### LINKED ISSUE #1534: Restarting a level plays the board piece moving video (not same as retail)
#### Describe the issue
Restarting a level replays the board piece moving video (not same as retail)

#### Steps to Reproduce
* Restart a level

#### Expected Behavior
Retail doesn't do this from what I can see
System Information
**CorsixTH Version: commit 27f5bb8

**Operating System: XUbuntu 18.04.2 LTS

Theme Hospital Version: GOG

### LINKED ISSUE #1663: Demo game win message has text overlap
#### Describe the issue
Seems the win message for the demo is too big for our fax. As a result, it encroaches on the choices section.

#### Expected Behavior
Do not overlap the text

#### System Information
**CorsixTH Version:** 0.64

**Operating System:** W10

**Theme Hospital Version:** Demo

### Additional Info
image

tobylane:
*Fixes #*

- Sometimes idle staff can be called to a room they can help in
- Use two more announcer strings about long queues
- Fix #1534 
- Fix #1311 
- Fix #525 
- Fix #474
- Fix #526 Caveat: receptionists can only be placed on the centre tile of the desk
- Fix #1663 There are many possible versions of this. We will want to guide translators to the original strings and have them do it within the same limit

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -608,7 +608,7 @@ end
 -- Loads the specified level. If a string is passed it looks for the file with the same name
 -- in the "Levels" folder of CorsixTH, if it is a number it tries to load that level from
 -- the original game.
-function App:loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor)
+function App:loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor, skip_movie)
   if self.world then
     self:worldExited()
   end
@@ -662,7 +662,7 @@ function App:loadLevel(level, difficulty, level_name, level_file, level_intro, m
   self.ui = GameUI(self, self.world:getLocalPlayerHospital(), map_editor)
   self.world:setUI(self.ui) -- Function call allows world to set up its keyHandlers
 
-  if tonumber(level) then
+  if tonumber(level) and not skip_movie then
     self.moviePlayer:playAdvanceMovie(level)
   end
 
@@ -1534,7 +1534,7 @@ function App:restart()
       self.ui:addWindow(UIInformation(self.ui, {_S.information.cannot_restart}))
       return
     end
-    local status, err = pcall(self.loadLevel, self, level, difficulty, name, file, intro)
+    local status, err = pcall(self.loadLevel, self, level, difficulty, name, file, intro, nil, true)
     if not status then
       err = "Error while loading level: " .. err
       print(err)
--- a/CorsixTH/Lua/dialogs/menu.lua
+++ b/CorsixTH/Lua/dialogs/menu.lua
@@ -657,7 +657,10 @@ function UIMenuBar:makeGameMenu(app)
   -- Edge Scrolling
   options:appendCheckItem(_S.menu_options.edge_scrolling,
     not app.config.prevent_edge_scrolling,
-    function(item) app.config.prevent_edge_scrolling = not item.checked end,
+    function(item)
+      app.config.prevent_edge_scrolling = not item.checked
+      app:saveConfig()
+    end,
     nil,
     function() return not app.config.prevent_edge_scrolling end)
 
--- a/CorsixTH/Lua/dialogs/place_staff.lua
+++ b/CorsixTH/Lua/dialogs/place_staff.lua
@@ -77,15 +77,26 @@ end
 
 local flag_cache = {}
 local flag_altpal = 16
+-- Is this coordinate a valid tile for this staff member?
+local function isValidPlace(self)
+  self.world.map.th:getCellFlags(self.tile_x, self.tile_y, flag_cache)
+  local room = self.world:getRoom(self.tile_x, self.tile_y)
+  -- Is the tile inside the player's hospital building?
+  local isplayer = self.ui.hospital:getPlayerIndex() == flag_cache.owner and flag_cache.hospital
+  -- On a tile humanoids can walk on?
+  local walkable = flag_cache.passable and (not room and true or not room.crashed)
+  -- and in an area the regular staff (doctors, nurses and handymen) can go?
+  local staffable = (self.allow_in_rooms or flag_cache.roomId == 0)
+  local valid = walkable and staffable
+  -- Or is it a receptionist placed on the centre tile of an unstaffed desk?
+  local reception = self.staff and self.staff.profile.humanoid_class == "Receptionist" and
+      self.world:getObject(self.tile_x, self.tile_y, "reception_desk")
+  return isplayer and (reception or valid)
+end
+
 function UIPlaceStaff:draw(canvas)
   if self.world.user_actions_allowed then
-    self.world.map.th:getCellFlags(self.tile_x, self.tile_y, flag_cache)
-    local room = self.world:getRoom(self.tile_x, self.tile_y)
-    local player_id = self.ui.hospital:getPlayerIndex()
-    local valid = flag_cache.hospital and flag_cache.passable and
-        (self.allow_in_rooms or flag_cache.roomId == 0) and
-        (not room and true or not room.crashed) and
-        flag_cache.owner == player_id
+    local valid = isValidPlace(self)
     self.anim:setFlag(valid and 0 or flag_altpal)
     local zoom = self.ui.zoom_factor
     if canvas:scale(zoom) then
@@ -107,13 +118,7 @@ function UIPlaceStaff:onMouseUp(button, x, y)
       return true
     elseif button == "left" then
       self:onMouseMove(x, y)
-      self.world.map.th:getCellFlags(self.tile_x, self.tile_y, flag_cache)
-      local room = self.world:getRoom(self.tile_x, self.tile_y)
-      local player_id = self.ui.hospital:getPlayerIndex()
-      if flag_cache.hospital and flag_cache.passable and
-          (self.allow_in_rooms or flag_cache.roomId == 0) and
-          (not room or not room.crashed) and
-          flag_cache.owner == player_id then
+      if isValidPlace(self) then
         if self.staff then
           self.staff:setTile(self.tile_x, self.tile_y)
         else
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -208,6 +208,21 @@ function PlayerHospital:monthlyAdviceChecks()
   end
 
   self:checkReceptionAdvice(current_month, current_year)
+
+  -- Once a month the advisor will warn about long queues.
+  -- Warnings are less likely for the diagnosis rooms to keep the quantity of warnings down
+  for _, room in pairs(self.world.rooms) do
+    if #room.door.queue > 5 then
+      local info = room.room_info
+      if math.random(1, 3) == 1 and info.required_staff["Doctor"] and info.categories["diagnosis"] then
+        self:giveAdvice({_A.warnings.queue_too_long_send_doctor:format(info.name)})
+        break
+      elseif info.categories["treatment"] or info.categories["clinics"] then
+        self:giveAdvice({_A.warnings.queues_too_long})
+        break
+      end
+    end
+  end
 end
 
 --! Make players aware of the need for a receptionist and desk.
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -299,6 +299,17 @@ function PlayerHospital:onEndDay()
     self:dailyAdviceChecks()
   end
 
+  -- Occasionally unstaffed rooms with a queue will call for staff
+  for _, room in pairs(self.world.rooms) do
+    if math.random(1, 20) == 1 and #room.door.queue > 0 and not room.staff_member and
+        not room:testStaffCriteria(room:getRequiredStaffCriteria()) then
+          self.world.dispatcher:callForStaff(room)
+          break
+    end
+  end
+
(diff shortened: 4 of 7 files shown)


CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -299,6 +299,17 @@ function PlayerHospital:onEndDay()
...
+  -- Occasionally unstaffed rooms with a queue will call for staff
+  for _, room in pairs(self.world.rooms) do
+    if math.random(1, 20) == 1 and #room.door.queue > 0 and not room.staff_member and
+        not room:testStaffCriteria(room:getRequiredStaffCriteria()) then
+          self.world.dispatcher:callForStaff(room)
+          break
+    end
+  end
+
+  -- TODO: Do other regular things? Such as making plants need water.
```
Alberth289346, line 326: They already do, so delete this

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -299,6 +299,17 @@ function PlayerHospital:onEndDay()
...
   end
 
+  -- Occasionally unstaffed rooms with a queue will call for staff
+  for _, room in pairs(self.world.rooms) do
+    if math.random(1, 20) == 1 and #room.door.queue > 0 and not room.staff_member and
+        not room:testStaffCriteria(room:getRequiredStaffCriteria()) then
+          self.world.dispatcher:callForStaff(room)
+          break
+    end
+  end
```
Alberth289346, line 324: The whole point of the dispatcher is that you don't need this. That is, if all is well, the dispatcher should already know the room needs staff.

That said, I am not convinced that the dispatcher is the right solution. I'd prefer something much simpler, where things that need other things just look for the latter directly.
It avoids the whole administration of needs, and cleanup of callbacks which is a source of bugs.

If you want to go this direction, it would be a separate patch.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -299,6 +299,17 @@ function PlayerHospital:onEndDay()
     self:dailyAdviceChecks()
   end
 
+  -- Occasionally unstaffed rooms with a queue will call for staff
+  for _, room in pairs(self.world.rooms) do
+    if math.random(1, 20) == 1 and #room.door.queue > 0 and not room.staff_member and
+        not room:testStaffCriteria(room:getRequiredStaffCriteria()) then
+          self.world.dispatcher:callForStaff(room)
+          break
```
Alberth289346, line 322: I understand what you're trying to do here, but this looks like there are 2 nested constructs, while in reality there is only 1 construct.
To solve it, do
``` suggestion
    if math.random(1, 20) == 1 and #room.door.queue > 0 and not room.staff_member and
        not room:testStaffCriteria(room:getRequiredStaffCriteria()) then
      self.world.dispatcher:callForStaff(room)
      break
```
The 'then' part is indented as normal, while the 'not' line has 'weird' indenting, sending a signal 'watch out!'

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -208,6 +208,21 @@ function PlayerHospital:monthlyAdviceChecks()
...
+  -- Warnings are less likely for the diagnosis rooms to keep the quantity of warnings down
+  for _, room in pairs(self.world.rooms) do
+    if #room.door.queue > 5 then
+      local info = room.room_info
+      if math.random(1, 3) == 1 and info.required_staff["Doctor"] and info.categories["diagnosis"] then
+        self:giveAdvice({_A.warnings.queue_too_long_send_doctor:format(info.name)})
+        break
+      elseif info.categories["treatment"] or info.categories["clinics"] then
+        self:giveAdvice({_A.warnings.queues_too_long})
+        break
```
Alberth289346, line 222: I find this `if` extremely complicated to understand. The random call influences the room check you're doing, but random and room-type are not related.

I'd suggest to make it a 2-level decision. First decide whether the room is 'bad', then decide if the advisor should report it.

Alternatively, first collect bad rooms, then decide which one to report.

CorsixTH/Lua/app.lua
```diff
@@ -1534,7 +1534,7 @@ function App:restart()
       self.ui:addWindow(UIInformation(self.ui, {_S.information.cannot_restart}))
       return
     end
-    local status, err = pcall(self.loadLevel, self, level, difficulty, name, file, intro)
+    local status, err = pcall(self.loadLevel, self, level, difficulty, name, file, intro, nil, true)
```
Alberth289346, line 1537: Adding more parameters gets tricky at some point, there is a thing named "too many parameters", and to me, this one is one of them.
So the question is whether you can avoid adding parameters, and if possible, even reduce the number of parameters.

I can see at least 2 options.
1. Playing the movie seems to not affect other parts of that function. Thus, it should be possible to pull that play code out of the function. Then add it again as a separate call after a call to the above function. At this spot you can simply leave it out completely, although a comment with a reason why it is not played here is likely useful to avoid someone else trying to be consistent and add it again.
2. Break the `loadLevel` function into pieces, and call each piece separately.

Both alternatives somewhat prefer this being a separate patch, as the amount of changes grows.

CorsixTH/Lua/dialogs/menu.lua
```diff
@@ -657,7 +657,10 @@ function UIMenuBar:makeGameMenu(app)
   -- Edge Scrolling
   options:appendCheckItem(_S.menu_options.edge_scrolling,
     not app.config.prevent_edge_scrolling,
-    function(item) app.config.prevent_edge_scrolling = not item.checked end,
+    function(item)
+      app.config.prevent_edge_scrolling = not item.checked
+      app:saveConfig()
+    end,
```
Alberth289346, line 663: While this change seems to follow an existing pattern so it's fine here, the whole approach here looks wrong.

The better spot for saving would be on closing the window, or even just before ending the program, imho.

CorsixTH/Lua/world.lua
```diff
@@ -2642,6 +2642,9 @@ function World:afterLoad(old, new)
     self:resetSideObjects()
   end
 
+  -- Cancel any saved screen movement from edge scrolling
+  self.ui.tick_scroll_amount_mouse = nil
+
```
Alberth289346, line 2647: Shouldn't this belong in `ui.afterLoad` ?

CorsixTH/Lua/dialogs/place_staff.lua
```diff
@@ -77,15 +77,26 @@ end
...
+  -- On a tile humanoids can walk on?
+  local walkable = flag_cache.passable and (not room and true or not room.crashed)
+  -- and in an area the regular staff (doctors, nurses and handymen) can go?
+  local staffable = (self.allow_in_rooms or flag_cache.roomId == 0)
+  local valid = walkable and staffable
+  -- Or is it a receptionist placed on the centre tile of an unstaffed desk?
+  local reception = self.staff and self.staff.profile.humanoid_class == "Receptionist" and
+      self.world:getObject(self.tile_x, self.tile_y, "reception_desk")
+  return isplayer and (reception or valid)
+end
```
Alberth289346, line 95: Breaking up that long condition is great!

Try adding a few empty lines in there, so you get 'blocks' of code that are closely connected. It makes reading it much simpler. [EDIT: Below I suggest to rewrite the whole thing, consider this afterwards.]
Also:
- Line 80 is not complete, it should have luadoc explanations
- `isPlayer` seems the wrong name for an inside-tile check.
- Move 85 closer to83, so 'room things' are near each other.
- If you rename `walkable` to `walkable_tile`, line 86 becomes redundant. This may be possible to other lines too.
- `allow_in_rooms` makes no sense as name at all, but that's not part of this patch, probably.
- `valid` is too wide, it doesn't make clear what validness you decided here.

The final `return` statement possibly invalidates several computations, ie, you're wasting CPU time on doing computations that you don't need.

Finally, line 92 is a whole different branch of computations, not related at all to any previous computed value, it's a big twist in the plot :) 

In all, the direction is good I think, it just needs a second sweep. I'd suggest something along the lines of
``` lua
local function isValidPlace(self)
  -- Insert code to compute isPlayer here
  if not isPlayer return false end

  -- Insert code for reception here
  if reception then return true end

  -- Insert code for valid here
  return valid
end
```

CorsixTH/Lua/languages/english.lua
```diff
@@ -104,8 +104,13 @@ multiplayer.everyone_failed = "Everyone failed to satisfy that last objective. S
 -- Override for a disease patient choice typo
 disease_discovered_patient_choice.need_to_employ = "Employ a %s to be able to handle this situation."
 
---Win message override typo
-letter[12][2] = "Your successful career as the best hospital administrator since Moses is nearing an end. However, such has been your impact on the cosy world of medicine, the Ministry would like to offer you a salary of $%d simply to appear on our behalf, opening fetes, launching ships, and doing chat shows. The whole world is clamouring for you, and it would be great PR for us all!//"
+-- Override for shorter messages and a typo
```
Alberth289346, line 107: We now lost the information which line had a typo.

Alberth289346:
Great work!

Two general advices:

1. You may want to split patches like these into more smaller parts. Merging patches like this has the effect that it is difficult to understand afterwards (say in 2 years by a new developer) how each individual bug was fixed, in particular now that changes are requested, which are likely not going to be near the previous commit.
As a general guideline, avoid having unrelated changes in a single patch.
It does create more merge requests and branches, but reviewing and fixing one patch is simpler, as it deals with only one thing.

2. A second advice is that the result that you get is not always optimal. This is normal. A reviewer has a fresh pair of eyes, and sees new things.
You can find those too by reviewing what you made afterwards. It works best if you don't touch it for a few days, and possibly even for a week at first. By that time you forgot all the details. Then look at it asif someone else has written the code, and think if you can improve it. Usually you can.
My usual practice is to at least review anything I wrote or rewrote at least once, unless the change is extremely simple. In 70-80% of the cases I can find further improvements. In complicated code, I even rewrite patches from scratch at times.
You can get more training in these things by trying to do reviews yourself, and reading reviews of others. To a large part it's about getting experience, and the best way to get that is DIY sufficiently often.

CorsixTH/Lua/app.lua
```diff
@@ -608,7 +608,7 @@ end
 -- Loads the specified level. If a string is passed it looks for the file with the same name
 -- in the "Levels" folder of CorsixTH, if it is a number it tries to load that level from
 -- the original game.
-function App:loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor)
+function App:loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor, skip_movie)
```
Alberth289346, line 611: Fix the luadoc

PR #1722, opened 2020-09-25
[WIP] Load initial reputation and competitor names from levels, adjust win fax intervals, varied staff spawns, more

tobylane:
*Fixes #470,#1633*

**Describe what the proposed changes do**
- Jukebox dialog is closed and disabled when it can't be used, the loop button handles there being no music to loop
- Carry forward to the next level the policies, and the disable staff requests option, which could be reconsidered a cheat
- Add a staff initial, a random letter from the name parts. This could well not make sense or be rude, by random combinations like S. Hatfield.
- Load initial reputation and competitor names from levels, change the current levels to match this
- If the intended save game is unwritable use the OS's temp dir and warn the player about where it is
- Various todos from comments and the wiki

PR #1741, opened 2020-10-25
[WIP] Use initial reputation, overdraft interest rate and competitor names from custom levels

tobylane:
A small portion of #1722 

- Level config can set starting reputation
- Competitor names are already in the custom levels, this uses them
- Overdraft interest rate is set by a per mille called OverdraftDiff, ie 200 means it's 0.02 on top of the standard rate
- Picking insurance company names can take an unknown number of names in the language file

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -36,15 +36,19 @@ function Hospital:Hospital(world, avail_rooms, name)
   local reputation_min = 0
   local reputation_max = 1000
   local reputation = math.round((reputation_min + reputation_max) / 2)
+  -- The difference between the normal and overdraft interest rates, multiplied by 10000
+  local overdraft_differential = 200
   if level_config then
     if level_config.towns and level_config.towns[level] then
       balance = level_config.towns[level].StartCash
       interest_rate = level_config.towns[level].InterestRate / 10000
       reputation = level_config.towns[level].StartRep or reputation
+      overdraft_differential = level_config.towns[level].OverdraftDiff or overdraft_differential
     elseif level_config.town then
       balance = level_config.town.StartCash
       interest_rate = level_config.town.InterestRate / 10000
       reputation = level_config.town.StartRep or reputation
+      overdraft_differential = level_config.town.OverdraftDiff or overdraft_differential
     end
   end
   self.name = name or "PLAYER"
@@ -84,6 +88,7 @@ function Hospital:Hospital(world, avail_rooms, name)
   -- Initial values
   self.interest_rate = interest_rate
   self.inflation_rate = 0.045
+  self.overdraft_interest_rate = interest_rate + overdraft_differential / 10000
   self.salary_incr = level_config.gbv.ScoreMaxInc or 300
   self.sal_min = level_config.gbv.ScoreMaxInc / 6 or 50
   self.reputation = reputation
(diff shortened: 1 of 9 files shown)


CorsixTH/Lua/hospital.lua
```diff
@@ -80,11 +88,12 @@ function Hospital:Hospital(world, avail_rooms, name)
   -- Initial values
   self.interest_rate = interest_rate
   self.inflation_rate = 0.045
+  self.overdraft_interest_rate = interest_rate + overdraft_differential / 10000
```
TheCycoONE, line 91: Must be set in afterLoad to keep compatibility with save games.

CorsixTH/Lua/hospital.lua
```diff
@@ -184,18 +193,24 @@ function Hospital:Hospital(world, avail_rooms, name)
...
-    -- NOTE: Will not work if more companies are added
-    if math.random(1, 11) < 4 or 11 - no < #self.insurance + 3 then
-      self.insurance[#self.insurance + 1] = local_name
-    end
-    if #self.insurance > 2 then
-      break
-    end
+    companies[no] = local_name
   end
+  repeat
```
TheCycoONE, line 205: This block could use a comment explaining what's going on and why. It's not clear to me why we prefer (but don't require) the first quarter of companies from the list.

CorsixTH/Lua/world.lua
```diff
@@ -153,6 +153,9 @@ function World:World(app)
   for key, value in pairs(level_config.computer) do
     if value.Playing == 1 then
       self.hospitals[#self.hospitals + 1] = AIHospital(tonumber(key) + 1, self, avail_rooms)
+      if value.Name then
```
TheCycoONE, line 156: Pass the name as the third parameter, just like the player hospital. Change the ... in aihospital to explicitly define the parameters the same as the parent class, only use the competitor table when name is nil, and remove this block.

TheCycoONE (changes requested):
A handful of changes requested; particularly the addition of an afterLoad.

TheCycoONE:
@mugmuggy do you know if toby's change to make the reputation half way between the win / lose conditions matches the original, vs. the 500 that we have been starting with. I'm not confident that change is right, but I haven't checked myself.

Alberth289346:
```
hospital.lua:
85:  self.reputation = 500
86:  self.reputation_min = 0
87:  self.reputation_max = 1000
```
ie it's still `500`.

CorsixTH/Lua/hospital.lua
tobylane: I may have read that it was like that in TH, but can't find it now.

TheCycoONE:
@Alberth289346 that block was replaced. It does indeed still default to 500, but written as
```
  local reputation_min = 0
  local reputation_max = 1000
  local reputation = math.round((reputation_min + reputation_max) / 2)
```
later reputation is replaced with the level config value, which is nice, then self.reputation is set to that local reputation value.

So it does indeed still default to 500, though it does a bit of extra work for it. I recommend changing that back to `= 500` since other things need fixing in that file anyway.

mugmuggy:
@TheCycoONE Just coming back to this as you are correct reputation is always initialised to 500 * (good factor/bad divisor) and good and bad values are 20 constant and are clamped to 0-1000

PR #1743, opened 2020-10-29
Handyman can only water plants from inside the hospital

tobylane:
- Requires the handymen to respect walls. Best seen on L12.


Alberth289346:
LGTM

PR #1744, opened 2020-10-29
Cleanup temperature update code.

### LINKED ISSUE #1742: Max number of radiators at a tile in TH?
In 99245e26403b108b600f1b803729b4edcf8b28bf (diff is 2896 lines, so reading is not recommended), having multiple objects at a tile was introduced.
There, the heating code started counting radiators at a tile in `iRadiatorNumber`, and computes for the case that `iRadiatorNumber > 0`:
``` diff
                 iMergeTemp = iRadiatorTemperature;
-                iMergeRatio = 2;
+                iMergeRatio = 2 - (iRadiatorNumber - 1) * 0.5;
```
I tried to place more than one radiator at a tile, and the game didn't allow that. In other words, this change does exactly nothing.

So the question is, are we supposed to allow more than one radiator at a tile, or is this code more complicated than needed?

Alberth289346:
*Fixes #1742

- We can have only one radiator at a tile, so the code can be simplified.
- While at it, also renamed a few variables and made computations a bit more clear.

### DIFF
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -1319,8 +1319,8 @@ namespace {
     \param ratio Weight of the old node temperature, \c N-1 parts of
         the node temperature and \c 1 part of the external temperature.
 */
-void merge_temperatures(map_tile& node, size_t temp_idx,
-                        uint32_t other_temp, double ratio) {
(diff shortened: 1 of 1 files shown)


TheCycoONE:
All done with this? (there's always more to clean)

Alberth289346:
Ah yes, sorry about that.
Yes I am done here for now.

PR #1745, opened 2020-10-30
Add initial to staff names, from users language

tobylane:
- Picks a random letter for the first name initial from the name parts.


CorsixTH/Lua/staff_profile.lua
```diff
@@ -143,8 +143,20 @@ function StaffProfile:randomise(month)
...
+    local initials = {}
+    for uchar in string.gmatch(letters,
+      "([%z\1-\127\194-\244][\128-\191]*)") do
+      initials[#initials + 1] = uchar
+    end
+    self.name = initials[math.random(1, #initials)] .. ". "
+  else
+    local num = math.random(1, letters:len())
+    self.name = letters:sub(num, num) .. ". "
+  end
```
Alberth289346, line 158: Seems a bit expensive to compute all possible letters each time, since it's always the same unless you change language. So I propose that you make a cache of available initials.

Initialize the cache when you arrive here for the first time (ie when the cache is empty). I'd always simply assume we have utf-8 to simplify the code.
If you want to be resistant against change of language, you can store the relevant `_S` strings too in the cache, and compare them with the current value (and rebuild the cache if the cache is empty or the `_S` strings don't match).

tobylane:
The surname changes straight away when loading a different language, but the old initial stays until the new month. I can't improve on that so far.

Alberth289346:
I didn't like storing the cache in `app`. If we add more and more data to it, it eventually becomes one pile of random data without a clue where things come from.
So I made another proposal: https://github.com/Alberth289346/CorsixTH/commit/7551a2142d40244977c331f437e25c9f60c69026

How does that look?

tobylane:
That looks better.

A bug with any cache is that even if you hire the doctor generated in another language they'll still have the previous language's initial. Going from Chinese to English the debug output name is 艾. WYSMITH but ingame it's ?. WYSMITH forever.

tobylane:
Is this a good approach?

tobylane:
Improved and rebased.

CorsixTH/Lua/dialogs/hire_staff.lua
```diff
@@ -169,7 +169,7 @@ function UIHireStaff:draw(canvas, x, y)
     if not profile then
       return
     end
-    font:draw(canvas, profile.name, x + 79, y + 21)
+    font:draw(canvas, profile.initial .. ". " .. profile.name, x + 79, y + 21)
```
Alberth289346, line 172: As you can see in the diff, you have this sequence everywhere.

This implies it may be a good idea to move that concatenation into the `profile` class.
I think that is defensible, the profile manages all kinds of personal properties, so the full name with initials fits in it naturally.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -26,6 +26,29 @@ local PlayerHospital = _G["PlayerHospital"]
 local num_sitting_ratios = 15 -- Number of stored recent sitting ratio measurements.
 local ratio_interval = 2 -- Measurement interval in days.
 
+strict_declare_global "staff_initials_cache"
+staff_initials_cache = {}
+
+local function our_concat(t)
```
Alberth289346, line 32: For some reason I constantly run into this function. (Maybe you are moving it all the time :p )

It's not a big problem, but I think eventually it should be a global function with a better name and some documentation.
The question is 'where', but I have no answer to that.

@TheCycoONE  Maybe you have some idea?

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -386,5 +425,16 @@ function PlayerHospital:afterLoad(old, new)
...
+  updateInitialsCache()
+  -- Fix the initial of all employed and employable staff
+  for _, staff in ipairs(self.staff) do
+    localiseInitial(staff.profile)
+  end
+  for _, subtable in pairs(self.world.available_staff) do
+    for _, staff in pairs(subtable) do
+      localiseInitial(staff)
+    end
+  end
```
Alberth289346, line 437: In multi-player, we will have several players, each possibly with a different language.
As such, I agree `PlayerHospital` is perhaps the right spot for the hospital staff.

The trouble is with `self.world.available_staff`. That will be shared between players. If three players each have a different language, what should be the name of the staff?
Difficult question.

As it is now, a staff member has as many names as there are languages, I think.

Alberth289346: Keeping it in `world.lua` would also work, it's a central point for all hospitals, and it does global things.

CorsixTH/Lua/dialogs/hire_staff.lua
tobylane: Putting the two parts together any earlier breaks the last name translation.

CorsixTH/Lua/hospitals/player_hospital.lua
tobylane: Utility.lua?

CorsixTH/Lua/hospitals/player_hospital.lua
tobylane: For now: moved to world.lua

For multiplayer: Use the name seed for both parts of the name? As part of multiplayer the seed is generated once then each user makes the local name.

CorsixTH/Lua/dialogs/hire_staff.lua
Alberth289346: So you have to do this when the name is requested.
Fair enough.

what about making a new function
``` lua
function StaffProfile:getFulleName()
  return self.initial .. ". " .. self.name
end
```
and use that?

That eliminates the clutter everywhere too.

CorsixTH/Lua/hospitals/player_hospital.lua
Alberth289346: That would be indeed the obvious spot.

I looked through it yesterday, but it doesn't have any string other function, so if you don't know that the function exists, you will never find it. Obviously, that's the case now as well, but I was hoping to avoid that.

There is also `strings.lua`, but that is also not a good spot it seemed on my first scan.

Ah well, we can always keep it around like it is :)

CorsixTH/Lua/hospitals/player_hospital.lua
Alberth289346: That could work.

As persistence saves all lua data, it also saves your customization to the strings. If 2 players each have a different language and they both save the game, their files will be different. Don't know if it's a problem, but it's a little weird.

Perhaps it's just another aspect of multi-player I didn't know about until now.

Alberth289346:
For some time now, this patch is bothering me (@toby: and you probably even more :) ).
It all seems rather simple, but each try (before this one) somewhat twisted wrong in some way.
I wondered why for some time already.

Yesterday I finally understood it (or at least I think I do).
Before you started, the profile class was just a set of data, initialized once, and that was it. (Maybe it already handled translations, not sure about it).
The patch introduced localized initial names, which creates the need to do some calculations that you don't want to repeat for each profile, you want to share the results. But the data class has no useful spot for that, so we tried a local variable, you moved things to the player hospital and now to world.

What we maybe should have done is to create a new spot dedicated to shared profile data, a new class.
One instance of that new class is created, and each created staff profile gets a link to that instance, so each staff profile can query the shared profile data about language and initials (and maybe even more), and get its data from there.

To give you a more concrete picture of it, the cache would be in the shared instance, and instead of pushing language changes to all existing profiles like you do now, you just change the current language in the shared instance, and that's it.
Each staff profile checks with the shared instance if its data is still valid with respect to the currently selected language each time it gives information away, and if needed updates itself first.

Note I am not saying you should now change it in this way. I am fine with the current solution, and assuming thecycoone is happy as well, we can leave this change to somewhere in the future.

tobylane:
I think I understand your plan, and will try it later.

This fails on luatest, which I don't understand well enough.

> Error → spec/entities/humanoids/staff/doctor_spec.lua @ 34
Doctor: Can represent doctor as a string
../Lua/entities/humanoid.lua:925: attempt to call a nil value (method 'getFullName')`

lewri:
Looking through your luatest error, I think the error has happened because the test doesn't know about getFullName()
So I think you need to either forcibly give the outcome of the function... or maybe include() the lua file
The latter would probably be considered an actual test... I imagine the former is just making the test output to your whims
Try adding at the top (after line 23) require("staff_profile")

tobylane:
Fixed the test, ready.

CorsixTH/Lua/staff_profile.lua
```diff
@@ -142,12 +145,17 @@ function StaffProfile:randomise(month)
...
-  -- TODO: Randomise a letter from the current language's alphabet.
-  self.name = "" -- string.char(string.byte"A" + math.random(0, 25)) .. ". "
+  local initials = staff_initials_cache.initials
+  local num = self.name_seed % #initials
+  self.initial = initials[num]
+  self.name = ""
   for _, part_table in ipairs(name_parts) do
     self.name = self.name .. part_table.__random
   end
+  print(self.initial, self.name)
```
lewri, line 157: Is this intentional or a leftover print? Hard to tell on mobile

tobylane: Thanks for the catch.

CorsixTH/Lua/world.lua
```diff
@@ -2304,6 +2304,43 @@ function World:resetAnimations()
...
+  end
+  staff_initials_cache.initials = initials
+end
+
+-- Change the staff name first letter to one from the current language
+-- from the seed (generated here or in staff_profile.lua)
+function World:localiseInitial(profile)
+  if not profile.name_seed then
+    profile.name_seed = math.random(1, 1000)
+    profile.name_lang = "false"
```
TheCycoONE, line 2335: String `false` is a little odd. I guess the idea is that it doesn't match any language name, but wouldn't the empty string (or even nil) work just as well for that?

CorsixTH/Lua/world.lua
```diff
@@ -2304,6 +2304,43 @@ function World:resetAnimations()
...
+    initials[#initials + 1] = uchar
+  end
+  staff_initials_cache.initials = initials
+end
+
+-- Change the staff name first letter to one from the current language
+-- from the seed (generated here or in staff_profile.lua)
+function World:localiseInitial(profile)
+  if not profile.name_seed then
+    profile.name_seed = math.random(1, 1000)
```
TheCycoONE, line 2334: Why is this 1000 here and 1009 in StaffProfile? The comment in StaffProfile makes the max sound like it's important (though I have doubts)

tobylane:
Fixed, better documentation. Ready.

CorsixTH/Lua/hospitals/player_hospital.lua
TheCycoONE: I don't understand why we're calling updateInitialCache and managing the afterLoad here now that the implementation is all in world. I think both the initial call and the afterLoad handling should be moved there too.

tobylane: The afterload from PlayerHospital happens before World's, and World is too late.

lewri: That sounds like a different issue. Logically the world should be loading in before the hospitals.

TheCycoONE: @tobylane that's not true, the afterLoad sequence is:
map (which doesn't call anything else)
ui (which calls afterLoad on all the child windows)
world (which calls afterLoad on the hospitals and other objects)

So you can control whether your update happens before or after hospitals and staff in world's afterLoad by where in the function you put your update.

tobylane: Made the move to world.

TheCycoONE:
Looks good to me now. Going to play test before merging. If anyone else has play tested let me know what you've tried.

CorsixTH/Lua/hospitals/player_hospital.lua
TheCycoONE: Surprised you put it at the end of World's afterLoad, after saying that was too late. It seems to play test well for me though. What was the dependency concern?

CorsixTH/Lua/hospitals/player_hospital.lua
TheCycoONE: This isn't strictly resolved, but I don't mind solving it later.

TheCycoONE:
Plays well for me, old saves, new games, switching language on a save, switching alphabet (i.e. chinese) and loading a save.

CorsixTH/Lua/world.lua
```diff
@@ -2365,6 +2366,44 @@ function World:resetAnimations()
...
+  -- The standard table.concat function doesn't like our userdata strings :(
+  local result = ""
+  for _, s in ipairs(t) do
+    result = result .. s
+  end
+  return result
+end
+
+function World:updateInitialsCache()
+  -- Generate new set of staff name initials in case the language was changed
```
lewri, line 2382: This "in case language changed" check isn't done here. Plus it doesn't reflect the function's use in World:World.

lewri (commented):
Just nitpicking here

CorsixTH/Lua/world.lua
```diff
@@ -2365,6 +2366,44 @@ function World:resetAnimations()
...
+
+-- Change the staff name first letter to one from the current language
+-- from the seed (generated here or in staff_profile.lua)
+function World:localiseInitial(profile)
+  if not profile.name_seed then
+    -- 1009 is a prime number which avoids a modulo of 0 when we need
+    --  a positive number to randomly pick the initial letter
+    profile.name_seed = math.random(1, 1009)
+  end
+  if profile.name_lang == TheApp.config.language then return end
```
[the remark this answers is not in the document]
TheCycoONE: Out of scope for this PR. Not a bad idea though.

CorsixTH/Lua/world.lua
```diff
@@ -2723,6 +2762,17 @@ if old < 153 then
     end
   end
 
+  -- Fix the initial of staff names
```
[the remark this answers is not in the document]
TheCycoONE: That's fine, it's cheap enough.

CorsixTH/Lua/world.lua
tobylane: Corrected the comments.

TheCycoONE:
@tobylane waiting for an answer about the afterLoad function. That's all that's holding this.

CorsixTH/Lua/hospitals/player_hospital.lua
tobylane: I must have tested wrong. The issue was writing a new initials cache before it was used, it's now in the right order.

PR #1746, opened 2020-10-30
[RDY] Win fax comes less often after it was declined

tobylane:
Previous feedback in #1722 

- Change the timing of win faxes per the todo
- This then uses the second set of fax texts, which ask the player to reconsider staying at this hospital

### DIFF
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -39,6 +39,8 @@ function PlayerHospital:PlayerHospital(world, avail_rooms, name)
     sitting_ratios = {}, -- Measurements of recent sitting/standing ratios.
     sitting_index = 1 -- Next entry in 'sitting_ratios' to update.
   }
+
+  self.win_declined = false -- Has not yet declined the level win fax
 end
 
 --! Give advice to the player at the end of a day.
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/world.lua
```diff
@@ -1174,15 +1174,15 @@ end
...
-  if self.game_date:monthOfYear() % 3 == 0 and self.game_date:monthOfYear() < 12 then
-    self:checkIfGameWon()
-  end
+  -- Check if a player has won the level at months 3, 6 and 9. The annual report
+  -- window will perform this check at month 12 when it has been closed.
+  -- If the offer is declined then the next check is at month 6 and the annual report.
+  local check_months = {
+    [3] = not self.hospitals[1].win_declined,
+    [6] = true,
+    [9] = not self.hospitals[1].win_declined
```
Alberth289346, line 1183: Please don't add more `self.hospitals[1]` references, they must all be modified when we go to multi-player.
Instead query the local player hospital by `world:getPlayerHospital()` and use the returned value.

lewri (approved):
The 3, 6, 9 isn't sitting right with me but I also can't see it as a blocking factor.

Alberth289346:
The number of months isn't likely to change much :)

TheCycoONE:
Not blocking this PR, but if anyone feels like cleanup, all this check if won logic belongs in hospital not world.

Alberth289346:
It's not on my current shortlist.

tobylane:
Moved that section to player hospital. Ready.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -39,6 +39,8 @@ function PlayerHospital:PlayerHospital(world, avail_rooms, name)
     sitting_ratios = {}, -- Measurements of recent sitting/standing ratios.
     sitting_index = 1 -- Next entry in 'sitting_ratios' to update.
   }
+
+  self.win_declined = false -- Has not yet declined the level win fax
```
TheCycoONE, line 43: Should have an afterLoad to make this.

PR #1747, opened 2020-11-08
[RDY] Place initial staff in a level randomly around the starting tile

tobylane:
- Noticeably on level 5 they can be seen before closing the starting message. Previous feedback in #1722

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1517,8 +1517,30 @@ function Hospital:initStaff()
       if added_staff then
         local staff = self.world:newEntity(profile.humanoid_class, 2)
         staff:setProfile(profile)
-        -- TODO: Make a somewhat "nicer" placing algorithm.
-        staff:setTile(self.world.map.th:getCameraTile(1))
+
+        -- Identify a safe starting place and
+        -- try to place the staff member randomly near it or on it
+        local map = self.world.map.th
+        local map_x_length, map_y_length = map:size()
+        local map_offset = 10
+        local x, y = map:getCameraTile(self:getPlayerIndex())
+        -- Make sure the random numbers (x_attempt, y_attempt) will be safely within the map
+        local x_safe = math.max(map_offset + 1, math.min(x, map_x_length - map_offset))
+        local y_safe = math.max(map_offset + 1, math.min(y, map_y_length - map_offset))
+        local x_attempt, y_attempt
+        local attempts = 0
+        repeat
+          x_attempt = x_safe + math.random(-map_offset, map_offset)
+          y_attempt = y_safe + math.random(-map_offset, map_offset)
+          attempts = attempts + 1
+        until attempts > 100 or self:isInHospital(x_attempt, y_attempt)
+        if attempts <= 100 then
+          staff:setTile(x_attempt, y_attempt)
+        elseif self:isInHospital(x, y) then
+          staff:setTile(x, y)
+        else
+          staff:setTile(map_x_length / 2, map_y_length / 2)
+        end
         staff:onPlaceInCorridor()
         self.staff[#self.staff + 1] = staff
         staff:setHospital(self)


CorsixTH/Lua/hospital.lua
```diff
@@ -1517,8 +1517,16 @@ function Hospital:initStaff()
       if added_staff then
         local staff = self.world:newEntity(profile.humanoid_class, 2)
         staff:setProfile(profile)
-        -- TODO: Make a somewhat "nicer" placing algorithm.
-        staff:setTile(self.world.map.th:getCameraTile(1))
+
+        local x, y = self.world.map.th:getCameraTile(1)
```
Alberth289346, line 1521: Use `Hospital:getPlayerIndex()` to get the number.

CorsixTH/Lua/hospital.lua
```diff
@@ -1517,8 +1517,16 @@ function Hospital:initStaff()
...
+        local x2, y2
+        local attempts = 0
+        repeat
+          x2 = x + math.random(-10, 10)
+          y2 = y + math.random(-10, 10)
+          attempts = attempts + 1
+        until self:isInHospital(x2, y2) or attempts > 100
+        staff:setTile(x2, y2)
```
Alberth289346, line 1529: So you tried a 100 times and failed, and then put the staff member down to the last randomly selected tile.
Depending on how bad the camera location is, it may even be off-map. [Question: does the `isInHospital` function handle off-map coordinates in case the camera is close to the edge of the map, or does it simply crash?]

One way to handle it would be to make sure the original `x`, `y` pair is in the hospital and use that if 100 tries all fail. A second way is to query the world for a tile that is in the hospital (don't think we have that function yet), in case all attempts fail.

tobylane:
> Does the isInHospital function handle off-map coordinates in case the camera is close to the edge of the map, or does it simply crash?

It crashes with `bad argument #1 to 'getCellFlags' (Map coordinates out of bounds)`.

Used your other two notes and made it a little more efficient.

Alberth289346:
> It crashes with `bad argument #1 to 'getCellFlags' (Map coordinates out of bounds)`.

I am not sure why you don't address this.
The general idea is that a program should never ever crash, no matter what you try. It may behave strangely if you try weird things, but not crash.

The first thing to establish is what the old code did, and whether it could crash. The idea behind it is that you must not introduce new ways to crash the program. It's preferable but not required to fix already existing ways to make it crash.
In this way, the number of ways to crash the program should not increase, and hopefully decrease until we ultimately we eliminated all ways, and have a stable program.

As it's all about camera positions, let's establish what you can assume there. Note that none of it is documented (probably, I didn't check), so it's all just educated guesses. [If you know more promises we can assume here, those can be included too.]

- I would say it is fair to assume that the camera position is on-map. (If it isn't, the level-file loader is wrong and should be fixed, in my opinion.)
- It would also be highly desirable that the camera position is inside the (correct) hospital, I think. I don't know if we get that guarantee from the level-file loader.

Unless the level-file loader gives more promises (I didn't check that code, so I don't know), I think this is the most you can assume for the camera position.

In the old version, the staff member was simply dropped at the camera position at all times.

- As we just assumed above camera position is on-map, the old version thus never crashed on off-map positions.
- The old code didn't do anything for ensuring it's inside a hospital, so it completely relied on a camera position being inside.

With that established, what does your patch do?

Starting with the 2nd point, you do try to get the staff member inside a hospital, so that's an improvement. If all your attempts fail, you use the camera position to drop the staff member, the same as in the old code. Conclusion, your patch improved on the 2nd point.

The first point however doesn't work any more. We didn't assume that the camera position would not be near an edge. In other words, tile `(2, 32)` is a perfectly valid on-map camera position. [It may be unlikely to happen but it is not an invalid position, so it could happen, it's not impossible.]
With your offset added though, it can easily become an off-map position. Random may give an offset of `(-5, 7)`, leading to `x2, y2` being `(-3, 39)` which is off-map. Calling `self:isInHospital(x2, y2)` then will crash the program.
Conclusion, you created a new way to make the program crash. That should not  happen, and must be fixed in the same patch.

The simple local solution is to verify and/or correct the computed position so it is never off-map, and you can call the `isInHospital` function safely.

A more elaborate solution comes from realising that if you have this problem, all other uses of this function have this problem too. In other words, it may be better to fix the `isInHospital` function itself, so all uses throughout the entire code benefit.
Such a change is however a separate patch. It may also be more involved than just hacking the `isInHospital` function.

mugmuggy:
Out of interest, TH does it by going from the first tile, if that tile belongs to the players hospital, it then does a remainder == 0 on a random number with a configured divisor. The random number has its own custom seed and that means L5 places the doctors always on 60,61 initially. A similar algorithm, is also used for placing the litter, initial rat holes and the handymen in the mini game.

tobylane:
I only caused that crash by adding 100 where this change adds 10. The camera tile can only get so near the edge of the world. I could normalise x and y to >10 and <118 which would keep the random numbers always valid.

Alberth289346:
> I only caused that crash by adding 100 where this change adds 10. The camera tile can only get so near the edge of the world. I could normalise x and y to >10 and <118 which would keep the random numbers always valid.

That works indeed. Normally you would query the map size from a variable or a function and use that, but we don't seem to have such a value or function (at least nothing `128` seems to point in that direction).

In the optimal case, you would create such a variable somewhere, and rework all code that uses the map size as hard-coded value to use that variable.

If that seems too much at this time (and I can imagine), then create a local variable like
```
local map_size = 128  -- TODO Use actual map size.
```
in the file, and then use `map_size` to compute the upperbound.

In that way, you're improving the chance that at the time of adding a mapsize, your spot is also found and taken into account.

Alberth289346:
Note that you do need the original `x` and `y` in case all attempts fail.

TheCycoONE:
We definitely do have an authoritative source for the map size:
`lcb.add_function(l_map_getsize, "size")` i.e. th.map.size

CorsixTH/Lua/hospital.lua
```diff
@@ -1517,8 +1517,27 @@ function Hospital:initStaff()
...
+        local map_size = (map:size())
+        local x, y = map:getCameraTile(Hospital:getPlayerIndex())
+        -- Make sure the random numbers (x_attempt, y_attempt) will be safely within the map
+        local x_safe = math.max(10, math.min(x, map_size - 10))
+        local y_safe = math.max(10, math.min(y, map_size - 10))
+        local x_attempt, y_attempt
+        local attempts = 0
+        repeat
+          x_attempt = x_safe + math.random(-10, 10)
+          y_attempt = y_safe + math.random(-10, 10)
```
Alberth289346, line 1533: Unfortunately this is not always within the map.
Your `x` and `y` are `10` or bigger from line 1527 and 1528. The reference manuals says that `math.random(m ,n)` gives a random number `[m, n]` (ie boundaries included).

The limits checks:
- Map coordinates for `x` run from `1` upto-and-including `xsize`.
- `x=10`, random = `-10`, result is `0`, which is off-map.
- `x=xsize-10`, random = `10`, result is `xsize`, which  is on-map.

These things are always tricky to get right. I made it a habit to run a check like above on any limit I set or read.

As it needs changing  anyway, perhaps add a `max_offset` variable as well, instead of the hard-coded `10` everywhere? In the current code hardcoded `10` was just still feasible, as it's `10` all over the place. But with the change, you're going to have an `11` as well somewhere, and then it becomes complicated how the `10` and `11` are related.

tobylane:
Fixed that, and made one last attempt to spawn staff, in the centre of the map.

CorsixTH/Lua/hospital.lua
```diff
@@ -1517,8 +1517,30 @@ function Hospital:initStaff()
...
         staff:setProfile(profile)
-        -- TODO: Make a somewhat "nicer" placing algorithm.
-        staff:setTile(self.world.map.th:getCameraTile(1))
+
+        -- Identify a safe starting place and
+        -- try to place the staff member randomly near it or on it
+        local map = self.world.map.th
+        local map_x_length, map_y_length = map:size()
+        local map_offset = 10
+        local x, y = map:getCameraTile(Hospital:getPlayerIndex())
```
mugmuggy, line 1526: Want self:getPlayerIndex() here

tobylane: Yes, changed.

lewri (approved):
Change works fine for Level 5 by my analysis.
```
Attempts was    1        . Staff placed at      64      76
Attempts was    1        . Staff placed at      61      81
Attempts was    1        . Staff placed at      53      78
Created game with savegame version 143.
[GAME RESTARTED]
Attempts was    1        . Staff placed at      64      72
Attempts was    1        . Staff placed at      69      80
Attempts was    1        . Staff placed at      59      70
Created game with savegame version 143.
[GAME RESTARTED]
Attempts was    1        . Staff placed at      67      73
Attempts was    2        . Staff placed at      57      77
Attempts was    1        . Staff placed at      61      83
Created game with savegame version 143.
```

Also tested with the preloaded "Avatar" single scenario.
```
Warning: Removing disease "pregnant" due to missing treatment room "operating_theatre".
Attempts was    4        . Staff placed at      67      63
Created game with savegame version 143.
[GAME RESTARTED]
Warning: Removing disease "pregnant" due to missing treatment room "operating_theatre".
Attempts was    4        . Staff placed at      67      64
Created game with savegame version 143.
[GAME RESTARTED]
Warning: Removing disease "pregnant" due to missing treatment room "operating_theatre".
Attempts was    4        . Staff placed at      59      67
Created game with savegame version 143.
[GAME RESTARTED]
Warning: Removing disease "pregnant" due to missing treatment room "operating_theatre".
Attempts was    1        . Staff placed at      63      68
Created game with savegame version 143.
```

PR #1749, opened 2020-11-09
[RDY] Allow dmg files as disk images for TH data

tobylane:
- Generalises iso to disk images, currently .iso and .dmg.

### DIFF
--- a/CorsixTH/Lua/filesystem.lua
+++ b/CorsixTH/Lua/filesystem.lua
@@ -74,14 +74,24 @@ function FileSystem:_enumerate()
   end
 end
 
+--! Test if file name has an .iso or .dmg extension
+function FileSystem:isIso(name)
+  if name == nil then
+    return false
+  end
+
+  local ext = name:lower():match("%.(.+)$")
+  return ext == 'iso' or ext == 'iso9660$' or ext == 'dmg'
+end
+
 --! Set the root physical path for this FileSystem.
 -- If the path is an ISO then set the provider. If the path is a directory
--- then set the pysical_path and populate the files and sub_dirs.
+-- then set the physical_path and populate the files and sub_dirs.
 --
 --!param physical_path (string) a path on the filesystem to either a directory
 -- or theme hospital ISO file.
 function FileSystem:setRoot(physical_path)
-  if physical_path:match"%.[iI][sS][oO]$" or physical_path:match"%.[iI][sS][oO]9660$" then
+  if self:isIso(physical_path) then
     self.provider = ISO_FS()
     self.provider:setPathSeparator(pathsep)
     return self.provider:setRoot(physical_path)
--- a/CorsixTH/Src/th_lua_iso.cpp
+++ b/CorsixTH/Src/th_lua_iso.cpp
@@ -44,7 +44,7 @@ int l_isofs_set_root(lua_State* L) {
   const char* path = luaL_checkstring(L, 2);
   if (pSelf->initialise(path)) {
     lua_pushvalue(L, 2);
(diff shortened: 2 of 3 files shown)


lewri:
Are there any knock on behaviours for accessing our files via DMG e.g. is it restricted to the same issues as being unable to play videos?

tobylane:
Yes, this gives the dmg to the same iso_fs, so it will have that limitation.

CorsixTH/Lua/filesystem.lua
```diff
@@ -75,13 +75,14 @@ function FileSystem:_enumerate()
...
+-- If the path is a disk image then set the provider. If the path is a directory
 -- then set the pysical_path and populate the files and sub_dirs.
 --
 --!param physical_path (string) a path on the filesystem to either a directory
--- or theme hospital ISO file.
+-- or theme hospital ISO or DMG file.
 function FileSystem:setRoot(physical_path)
-  if physical_path:match"%.[iI][sS][oO]$" or physical_path:match"%.[iI][sS][oO]9660$" then
+  local path = physical_path:lower()
+  if path:match("%.iso$") or path:match("%.iso9660$") or path:match("%.dmg$") then
```
driverpt, line 85: Is this Case-Insensitive ?

driverpt (changes requested):
`.dmg` is Mac-specific. You should add some `if os.type =='darwin'` check to prevent issues.

tobylane:
In this case it's just another name for the same format, a name that the native tool uses. I could transfer my disk image to Windows and not need to rename it.

TheCycoONE:
dmg's can be a variety of formats, and the only format we support is ISO9660 so I think the isIso / ISO comments can remain as they are; or maybe explicitly `isISO9660`.

tobylane:
Reverted those parts and fixed a bug.

CorsixTH/Lua/filesystem.lua
TheCycoONE: @driverpt There's a tolower just above now. This path is only being used for the match check.

CorsixTH/Src/th_lua_iso.cpp
```diff
@@ -44,7 +44,7 @@ int l_isofs_set_root(lua_State* L) {
   const char* path = luaL_checkstring(L, 2);
   if (pSelf->initialise(path)) {
     lua_pushvalue(L, 2);
-    luaT_setenvfield(L, 1, "file");
```
TheCycoONE, line 47: good catch

TheCycoONE:
LGTM, ready when you are.

CorsixTH/Lua/filesystem.lua
```diff
@@ -76,12 +76,13 @@ end
...
 -- If the path is an ISO then set the provider. If the path is a directory
--- then set the pysical_path and populate the files and sub_dirs.
+-- then set the physical_path and populate the files and sub_dirs.
 --
 --!param physical_path (string) a path on the filesystem to either a directory
 -- or theme hospital ISO file.
 function FileSystem:setRoot(physical_path)
-  if physical_path:match"%.[iI][sS][oO]$" or physical_path:match"%.[iI][sS][oO]9660$" then
+  local path = physical_path:lower()
+  if path:match("%.iso$") or path:match("%.iso9660$") or path:match("%.dmg$") then
```
TheCycoONE, line 85: Any easy way to share this check with the filesystem check for the same thing? (I like the way it was done there a little better, if you had to choose implementation)

tobylane:
Both now use FileSystem:isIso. Ready.

PR #1751, opened 2020-11-18
Fix file selector on OSX

### LINKED ISSUE #1750: [Bug] AnimView File select OSX
#### Describe the issue
It's not possible to select `FONT00V.DAT` in AnimView's Sprite Viewer on OSX.

#### Steps to Reproduce
1. Open AnimView.app
2. Select Sprite Viewer
3. Try to open the FONT00V.DAT

#### Expected Behavior
`FONT00V.DAT` should be clickable and selectable in the OS File Dialog

#### System Information
**CorsixTH Version:** N/A

**Operating System:** *OSX 10.15*

#### Gamelog.txt
N/A

### Additional Info
N/A

driverpt:
*Fixes #1750*

Remove Regular Expressions from File Filter. Simplify Wildcard Filters

### DIFF
--- a/AnimView/frmSprites.cpp
+++ b/AnimView/frmSprites.cpp
@@ -52,7 +52,7 @@ frmSprites::frmSprites()
                   wxALIGN_CENTER_VERTICAL | wxALIGN_RIGHT);
   pFilesGrid->Add(
       m_txtTable = new wxTextCtrl(
-          this, wxID_ANY, L"X:\\ThemeHospital\\hospital\\QData\\Font00V.tab"),
+          this, wxID_ANY, L"X:\\ThemeHospital\\hospital\\QData\\FONT00V.TAB"),
       1, wxALIGN_CENTER_VERTICAL | wxEXPAND);
   pFilesGrid->Add(new wxButton(this, ID_BROWSE_TABLE, L"Browse..."), 0,
                   wxALIGN_CENTER_VERTICAL);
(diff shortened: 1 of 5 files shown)


Alberth289346:
Afaik there is no single well-defined way to transfer filenames from a case-insensitive file system to a case-sensitive file system.
As such your change is likely going to break other systems.

For example, the program completely breaks on use of the back-slash directory separator at my system.

If you want to fix it, I think it should be a generalisation of the file system that can safely handle all cases and directory separators.
A possible other solution is to extract all the hard-coded paths out of the executable. Not sure how feasible that is.

Personally, I find either too much work for this program, and keep a local patch that I apply before compiling it, if I use it.

The long term goal of this code is probably conversion of this thing, `wxWidgets` is a pain to build and maintain.

TheCycoONE:
Albert is right, I can't accept this change because it breaks linux. Does it work if you leave the regex but also include DAT and dat?

TheCycoONE:
Alternatively, good old precompiler flags.

lewri:
@tobylane any thoughts on if this one can be rescued/needs reviving?

tobylane:
It's needed, I can only use Animview with it. I'll do the precompiler flags.

TheCycoONE:
But MacOS isn't case sensitive usually?

TheCycoONE:
Okay

tobylane:
The first two changes in Animview aren't needed anymore, fixed in master.  
The other changes for Animview can be smaller, #1980

I don't need the lua changes, it seems like that is handled correctly already?

TheCycoONE:
Thanks @tobylane

PR #1754, opened 2020-11-21
[RDY] Prevent emergencies of undiscovered diseases

### LINKED ISSUE #506: Possible to treat emergency illness that you have not discovered!
_From [mark.l1...@hotmail.co.uk](https://code.google.com/u/117099726060447259318/) on November 13, 2012 15:39:42_

I don't know whether this should be possible or not, I suspect not.
I received a fax for several patients with invis' and it said I could treat them.  I have not yet discovered invis, but most were treated.  After the emergency was over I had an invis' go to the GP and invis' was then discovered.
You can see from the pic that invis' is not in the casebook yet.

Win7 using directx testing 0.11RC with level 6 and the emergency was in Nov of the first year

**Attachment:** [3a.png](http://code.google.com/p/corsix-th/issues/detail?id=1384)

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1384_

### LINKED ISSUE #1156: Error with button up handler
Accepted an emergency for slack tongue which I can treat, but it has not been discovered yet!

Load the save and accept the emergency.
As patients arrive open patient dialogue and click on casebook icon
Normally this would then open the casebook with the diagnosed illness selected.

Because of the way this has played out and due to a lack of funds I have discovered and built a treatment room for something yet to be discovered.  If I had the funds to build an x-ray I may then discover slack tongue.
Perhaps it should not be possible to have an emergency for something you have not discovered.
Patients with slack tongue have been arriving for sometime, but as I can't diagnose it yet I can only send them home (which I think is correct)

https://www.dropbox.com/sh/5pc51vpf1vcgsfs/AADTcwdeFKr_Eb1hsyDkf8Tca?dl=0

lewri:
Prevent emergencies being available to player if the relevant disease has not been discovered.

*Fixes #1156 Fixes #506*

- Checks if the game's generated emergency disease is discovered at the player hospital
- Where the disease isn't discovered, the emergency is discard
- If the player tried to make an emergency by cheating, they get a nice message instead if it failed.

### DIFF
--- a/CorsixTH/Lua/cheats.lua
+++ b/CorsixTH/Lua/cheats.lua
@@ -83,12 +83,13 @@ function Cheats:cheatResearch()
 end
 
 function Cheats:cheatEmergency()
-  local err = self.hospital:createEmergency() -- no err is nil
+  local err = self.hospital:createEmergency()
   local ui = self.hospital.world.ui
   if err == "undiscovered_disease" then
-    ui:addWindow(UIInformation(ui,{_S.misc.cant_treat_emergency}))
-  elseif err == "no_helipad" then
+    ui:addWindow(UIInformation(ui, {_S.misc.cant_treat_emergency}))
+  elseif err == "no_heliport" then
     ui:addWindow(UIInformation(ui, {_S.misc.no_heliport}))
+  -- else 'err == nil', meaning success. The case doesn't need special handling
   end
 end
 
--- a/CorsixTH/Lua/cheats.lua
+++ b/CorsixTH/Lua/cheats.lua
@@ -83,7 +83,10 @@ function Cheats:cheatResearch()
 end
 
 function Cheats:cheatEmergency()
-  if not self.hospital:createEmergency() then
+  local outcome, err = self.hospital:createEmergency()
+  if err == "undiscovered_disease" then
+    self.hospital.world.ui:addWindow(UIInformation(self.hospital.world.ui,{_S.misc.cant_treat_emergency}))
+  elseif not outcome then
     self.hospital.world.ui:addWindow(UIInformation(self.hospital.world.ui, {_S.misc.no_heliport}))
   end
 end
--- a/CorsixTH/Lua/cheats.lua
+++ b/CorsixTH/Lua/cheats.lua
@@ -83,11 +83,12 @@ function Cheats:cheatResearch()
 end
 
 function Cheats:cheatEmergency()
-  local outcome, err = self.hospital:createEmergency()
+  local err = self.hospital:createEmergency() -- no err is nil
+  local ui = self.hospital.world.ui
   if err == "undiscovered_disease" then
-    self.hospital.world.ui:addWindow(UIInformation(self.hospital.world.ui,{_S.misc.cant_treat_emergency}))
-  elseif not outcome then
-    self.hospital.world.ui:addWindow(UIInformation(self.hospital.world.ui, {_S.misc.no_heliport}))
+    ui:addWindow(UIInformation(ui,{_S.misc.cant_treat_emergency}))
+  elseif err == "no_helipad" then
+    ui:addWindow(UIInformation(ui, {_S.misc.no_heliport}))
   end
 end
 
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1042,9 +1042,9 @@ end
 
 -- Creates complete emergency with patients, what disease they have, what's needed
 -- to cure them and the fax.
---!return outcome (bool), err (string) If emergency was scheduled, and if not what the error was
+--!return err (string) If emergency failed what was the issue
 function Hospital:createEmergency(emergency)
-  local outcome = false
+  local err = nil
   local random_disease = self.world.available_diseases[math.random(1, #self.world.available_diseases)]
   local disease = TheApp.diseases[random_disease.id]
   local number = math.random(2, disease.emergency_number)
@@ -1063,7 +1063,8 @@ function Hospital:createEmergency(emergency)
 
     -- If disease chosen isn't discovered, cancel emergency
     if not self.disease_casebook[emergency.disease.id].discovered then
-      return outcome, "undiscovered_disease"
+      err = "undiscovered_disease"
+      return err
     end
 
     self.emergency = emergency
@@ -1112,9 +1113,10 @@ function Hospital:createEmergency(emergency)
       },
     }
     self.world.ui.bottom_panel:queueMessage("emergency", message, nil, Date.hoursPerDay() * 16, 2) -- automatically refuse after 16 days
-    outcome = true
+    return err -- sucessfully created
   end
-  return outcome
+  err = "no_helipad"
+  return err
 end
 
 -- Called when the timer runs out during an emergency or when all emergency patients are cured or dead.
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1042,9 +1042,8 @@ end
 
 -- Creates complete emergency with patients, what disease they have, what's needed
 -- to cure them and the fax.
---!return err (string) If emergency failed what was the issue
+--!return (optional string) Textual reason for failure, else nil
 function Hospital:createEmergency(emergency)
-  local err = nil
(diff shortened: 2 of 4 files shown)


CorsixTH/Lua/cheats.lua
```diff
@@ -83,7 +83,10 @@ function Cheats:cheatResearch()
 end
 
 function Cheats:cheatEmergency()
-  if not self.hospital:createEmergency() then
+  local outcome, err = self.hospital:createEmergency()
```
Alberth289346, line 86: It's nicer if `err` always contains the reason (and is `nil` if there is no error, probably), and then you have no need for `outcome`

CorsixTH/Lua/cheats.lua
```diff
@@ -83,7 +83,10 @@ function Cheats:cheatResearch()
...
+  if err == "undiscovered_disease" then
+    self.hospital.world.ui:addWindow(UIInformation(self.hospital.world.ui,{_S.misc.cant_treat_emergency}))
```
Alberth289346, line 88: If you make a local `ui` reference, that saves many `self.hospital.world.ui` de-references, and it makes the lines of code shorter as well.

CorsixTH/Lua/languages/english.lua
```diff
@@ -306,6 +306,7 @@ install = {
 
 misc.not_yet_implemented = "(not yet implemented)"
 misc.no_heliport = "Either no diseases have been discovered yet, or there is no heliport on this map. It might be that you need to build a reception desk and hire a receptionist"
+misc.cant_treat_emergency = "Your hospital cannot treat this emergency because its disease has not been discovered. Feel free to try again, cheater."
```
Alberth289346, line 309: I find `cheater` offensive in that sentence.
I am sure you are trying to be funny, but the reader can't see your big smile, and may mis-interpret the message.

CorsixTH/Lua/world.lua
```diff
@@ -1095,7 +1096,7 @@ function World:onEndDay()
   --check if it's time for a VIP visit
   if self.game_date:isSameDay(self.next_vip_date) then
     if #self.rooms > 0 and self.ui.hospital:hasStaffedDesk() then
-      self.hospitals[1]:createVip()
+      local_hospital:createVip()
```
Alberth289346, line 1099: And now you hope `self.ui.hospital` and `local_hospital` are actually the same thing.

lewri: Might need some elaboration on this. Are you suggesting to use ``local_hospital:hasStaffedDesk()`` on 1098 instead?

CorsixTH/Lua/world.lua
```diff
@@ -1082,6 +1082,7 @@ end
 
 -- Called immediately prior to the ingame day changing.
 function World:onEndDay()
```
Alberth289346, line 1084: While historically, `world` may have been the central boss, the code that you touch here hasn't much to do with `world`.
Instead, it seems to belong in the hospital class instead.

CorsixTH/Lua/languages/english.lua
lewri: I was on the edge about it. Chances are it's my british humour taking over generic humour. I'll just leave it without that word :)

CorsixTH/Lua/cheats.lua
lewri: This one is getting fixed in next revision

CorsixTH/Lua/world.lua
Alberth289346, line 1099: In my view, an object should not make decisions for another object without that explicit as design decision.
Here, world thus shouldn't decide for hospital.

All objects are grown-ups, they can decide things for themselves. If an object needs a decision, ask the other object about it. Don't make a decision without consulting the object that is affected.

Basically, apply real-life courtesy.


In code, that means you write a function in the hospital class that decides it is ready for a VIP visit, and call that.
[If you move the entire scheduling into hospital, the guideline also applies, but since the code is completely in the affected object then, it can simply compute the result directly (for short code at least).]

lewri: I think this is going outside the scope of the PR

CorsixTH/Lua/world.lua
lewri: Again the purpose of this PR is not lifting out potentially misplaced actions in ``onEndDay()``

CorsixTH/Lua/world.lua
Alberth289346: That's valid way of looking at it

CorsixTH/Lua/hospital.lua
```diff
@@ -1042,9 +1042,9 @@ end
 
 -- Creates complete emergency with patients, what disease they have, what's needed
 -- to cure them and the fax.
---!return outcome (bool), err (string) If emergency was scheduled, and if not what the error was
+--!return err (string) If emergency failed what was the issue
```
Alberth289346, line 1045: Alternative formulation:  `--!return (optional string) Textual reason for failure, else nil`

The doc string is intended for persons only reading the function header, but not the code. Those persons don't know "err", they only want to know what the return value means.
Added "optional" to express that you sometimes don't return a string.

CorsixTH/Lua/hospital.lua
```diff
@@ -1112,9 +1113,10 @@ function Hospital:createEmergency(emergency)
       },
     }
     self.world.ui.bottom_panel:queueMessage("emergency", message, nil, Date.hoursPerDay() * 16, 2) -- automatically refuse after 16 days
-    outcome = true
+    return err -- sucessfully created
   end
-  return outcome
+  err = "no_helipad"
+  return err
```
Alberth289346, line 1119: There is actually no reason to have an `err` variable in this function. You can simply return the value directly, For the line that I marked, it would be `return "no_helipad"` (but you likely already guessed that).

CorsixTH/Lua/cheats.lua
```diff
@@ -83,11 +83,12 @@ function Cheats:cheatResearch()
 end
 
 function Cheats:cheatEmergency()
-  local outcome, err = self.hospital:createEmergency()
+  local err = self.hospital:createEmergency() -- no err is nil
```
Alberth289346, line 86: It is not customary to denote meaning of return values near a call. The reason is that the comment will become misleading when the function is modified, and you typically don't want to dig around in the code finding all spots that document the return value.

Instead, there is one place for this information, the `--!return` line.

What you can do however, is mention it just below line 91:
``` lua
   -- else 'err == nil', meaning success. This case doesn't need special handling.
```
Technically, you're not documenting the return values of the function with such a comment, you're documenting explicitly that you considered the `err == nil` case in this function, and that you decided to do nothing.

Alberth289346:
Nice work

Alberth289346:
LGTM

CorsixTH/Lua/hospital.lua
```diff
@@ -1106,9 +1111,9 @@ function Hospital:createEmergency(emergency)
       },
     }
     self.world.ui.bottom_panel:queueMessage("emergency", message, nil, Date.hoursPerDay() * 16, 2) -- automatically refuse after 16 days
-    created_one = true
+    return -- sucessfully created
   end
-  return created_one
+  return "no heliport"
```
TheCycoONE, line 1116: This error is currently a catch all that includes staff desk, heliport, and according to the prompt 'not yet discovered' though you have a new condition for that.
You may break out a new error for the staff desk too?
Either way it's best to exit the function up when you know the condition doesn't hold rather than indenting the whole body. (Not a new issue)

TheCycoONE (commented):
I think when selecting the random disease you want to use hospital.discovered_diseases instead, and then the only time you would not have a discovered disease would be if there are no discovered diseases at all.

That table gives you the casebook index and only contains discovered diseases.

CorsixTH/Lua/hospital.lua
lewri: The new original prompt actually only happens if we have no diseases discovered at all. I'll need to revisit the code again to see if this actually happens outside level 1 (and maybe 2)

lewri:
As for the random disease selection, I ran with the idea here from Albert that we shouldn't compensate the player in any way if they get an emergency for a disease they haven't discovered yet.
https://github.com/CorsixTH/CorsixTH/issues/1156#issuecomment-699593110

PR #1755, opened 2020-11-22
[rdy] cpp cleanup of animation data

Alberth289346:
- Remove `link_list` inheritance of map_tile for storing entities
- Move `drawing_layer` value to `drawable` struct.

### DIFF
--- a/CorsixTH/Src/th_gfx.cpp
+++ b/CorsixTH/Src/th_gfx.cpp
@@ -1295,7 +1295,7 @@ bool THAnimation_is_multiple_frame_animation(drawable* pSelf) {
 
 }  // namespace
 
-animation_base::animation_base() {
+animation_base::animation_base() : drawable() {
   x_relative_to_tile = 0;
   y_relative_to_tile = 0;
   for (int i = 0; i < 13; ++i) {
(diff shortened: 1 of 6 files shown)


CorsixTH/Src/th_map.cpp
```diff
@@ -1600,12 +1601,12 @@ void level_map::depersist(lua_persist_reader* pReader) {
     if (!pReader->read_stack_object()) {
       return;
     }
-    pNode->next = (link_list*)lua_touserdata(L, -1);
-    if (pNode->next) {
-      if (pNode->next->prev != nullptr) {
+    pNode->entities.next = (link_list*)lua_touserdata(L, -1);
```
TheCycoONE, line 1605: Good, can you replace all these C style casts at the same time. Pretty sure all the pointer ones can be static_cast<>

CorsixTH/Src/th_gfx.cpp
```diff
@@ -1508,12 +1509,12 @@ void animation_base::attach_to_tile(map_tile* pMapNode, int layer) {
...
     pList = &pMapNode->oEarlyEntities;
   } else {
-    pList = pMapNode;
+    pList = &pMapNode->entities;
   }
 
   this->set_drawing_layer(layer);
 
-  while (pList->next && pList->next->get_drawing_layer() < layer) {
+  while (pList->next && ((drawable*)pList->next)->get_drawing_layer() < layer) {
```
TheCycoONE, line 1517: Just this one left

TheCycoONE:
Much better!

PR #1757, opened 2020-11-28
[RDY] Stop faxes being openable on pause

### LINKED ISSUE #1715: Should message windows be openable on pause?
Currently when paused and "editing while paused" is disabled we can open fax messages, staff rise windows. This has an avenue of unpausing the game. Should we be stopping this behaviour during pause? This came about while making #1714 

I've got no issue with these messages being openable if "editing while paused" is enabled.

Opening a Fax etc also removes the blue taint (even without the above PR). So it definitely is a 'break-out' of some sort.

lewri:
*Fixes #1715*

<s>I haven't tested this with staff rise windows yet...</s> works fine. so if anyone is able to tell me something's wrong let me know. Hopefully I've added functions in the right place for once.

- Prevents fax and staff rise windows being opened on pause where editing is not allowed.
- Added a "wrong2" sound when player tries to click on a message icon when not allowed.
- If the game gets paused because of a fax, then you can still switch between faxes as before.

### DIFF
--- a/CorsixTH/Lua/dialogs/message.lua
+++ b/CorsixTH/Lua/dialogs/message.lua
@@ -105,6 +105,11 @@ function UIMessage:adjustToggle()
 end
 
 function UIMessage:openMessage()
+  if TheApp.world:checkPauseBehaviour() and not self.ui:checkForMustPauses() then
+    self.ui:playSound("wrong2.wav")
+    self:adjustToggle()
+    return
+  end
   if TheApp.world:isCurrentSpeed("Speed Up") then
     TheApp.world:previousSpeed()
   end
--- a/CorsixTH/Lua/dialogs/message.lua
+++ b/CorsixTH/Lua/dialogs/message.lua
@@ -106,7 +106,7 @@ end
 
 --! Displays the fax/strike message to the player when opened from the bottom_panel.
 function UIMessage:openMessage()
-  if TheApp.world:preventActionsWhenPaused() and not self.ui:checkForMustPauseWindows() then
+  if TheApp.world:isUserActionProhibited() and not self.ui:checkForMustPauseWindows() then
     self.ui:playSound("wrong2.wav")
     self:adjustToggle()
     return
--- a/CorsixTH/Lua/dialogs/message.lua
+++ b/CorsixTH/Lua/dialogs/message.lua
@@ -104,8 +104,9 @@ function UIMessage:adjustToggle()
   end
 end
 
+--! Displays the fax/strike message to the player when opened from the bottom_panel.
 function UIMessage:openMessage()
-  if TheApp.world:checkPauseBehaviour() and not self.ui:checkForMustPauses() then
+  if TheApp.world:checkPauseBehaviour() and not self.ui:checkForMustPauseWindows() then
     self.ui:playSound("wrong2.wav")
     self:adjustToggle()
     return
--- a/CorsixTH/Lua/dialogs/message.lua
+++ b/CorsixTH/Lua/dialogs/message.lua
@@ -106,7 +106,7 @@ end
 
 --! Displays the fax/strike message to the player when opened from the bottom_panel.
 function UIMessage:openMessage()
-  if TheApp.world:checkPauseBehaviour() and not self.ui:checkForMustPauseWindows() then
+  if TheApp.world:preventActionsWhenPaused() and not self.ui:checkForMustPauseWindows() then
     self.ui:playSound("wrong2.wav")
     self:adjustToggle()
     return
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1041,7 +1041,7 @@ end
 
 function UI:removeWindow(closing_window)
   if Window.removeWindow(self, closing_window) then
-    local pauseGame = false
+    local pauseGame
     local class = closing_window.modal_class
     if class and self.modal_windows[class] == closing_window then
       self.modal_windows[class] = nil
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1047,12 +1047,7 @@ function UI:removeWindow(closing_window)
       self.modal_windows[class] = nil
     end
     if self.app.world and self.app.world:isCurrentSpeed("Pause") then
-      for i in pairs(self.windows) do
-        if self.windows[i].mustPause() then
-          pauseGame = true
-          break
-        end
-      end
+      pauseGame = self:checkForMustPauseWindows()
       if not pauseGame and closing_window.mustPause() then
         self.app.world:setSpeed(self.app.world.prev_speed)
       end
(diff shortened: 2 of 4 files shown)


lewri:
Hopefully at a later point I can replace some code with ``World:checkPauseBehaviour()`` in a future PR

CorsixTH/Lua/dialogs/message.lua
```diff
@@ -105,6 +105,11 @@ function UIMessage:adjustToggle()
 end
 
 function UIMessage:openMessage()
```
Alberth289346, line 107: As a rule of the thumb, please document the function when you touch it.

You have spend time to understand what it does, don't make others spend that same effort again, it's a waste of time.

CorsixTH/Lua/dialogs/message.lua
```diff
@@ -105,6 +105,11 @@ function UIMessage:adjustToggle()
...
+  if TheApp.world:checkPauseBehaviour() and not self.ui:checkForMustPauses() then
+    self.ui:playSound("wrong2.wav")
+    self:adjustToggle()
+    return
+  end
```
Alberth289346, line 112: The topic says you want to block opening faxes, so this spot is too late judging from the function name.
At this point the window already exists and the decision to open it has already been made (it seems to be called from the window being opened).

Blocking in my view means not even attempt to create or open it.

CorsixTH/Lua/ui.lua
```diff
@@ -1066,6 +1066,14 @@ function UI:removeWindow(closing_window)
   end
 end
 
+--! Function to check if we have any must pause windows open
+--!return (bool) Returns true if a must pause window is found
+function UI:checkForMustPauses()
+  for i in pairs(self.windows) do
+    if self.windows[i].mustPause() then return true end
+  end
+end
```
Alberth289346, line 1075: `nil` is not a valid boolean value to return.

The function name doesn't cover the computation. "check for" reads to me as "check if everything is ok" (and return true/false). Try to be more precise in what you compute.

Boolean functions often have "isX" or "hasX" as name, eg "haveMustPauseWindows()" is very concrete and also explains what a "true" as returned value means.

A sign here is when your longer explanation mentions relevant things not expressed in the function name.

CorsixTH/Lua/world.lua
```diff
@@ -954,6 +954,14 @@ function World:pauseOrUnpause()
   end
 end
 
+--! Function to check if player can perform actions when paused
+--!return (bool) Returns true if player hasn't allowed editing while paused
+function World:checkPauseBehaviour()
+  if self:isCurrentSpeed("Pause") and not self.user_actions_allowed then
+    return true
+  end
+end
```
Alberth289346, line 963: Ever noticed how easy it is to understand thre result value of `self:isCurrentSpeed("Pause")`?

The `and` expression already results in a boolean true/false, so you can reduce this function to `return self:isCurrentSpeed("Pause") and not self.user_actions_allowed`.
That also solves no returning `nil` as boolean value.

Function name seems not quite right to me, much alike the previous comment.

lewri:
On some further testing for https://github.com/CorsixTH/CorsixTH/pull/1757#discussion_r532183801 I tested moving the code into window.lua ``function Button:handleClick(mouse_button)`` but this broke other buttons on the UI.
To enable such a move would require all buttons to be re-evaluated and changed to solely be handled by the above. But that could be much more difficult work.

CorsixTH/Lua/dialogs/message.lua
```diff
@@ -104,8 +104,9 @@ function UIMessage:adjustToggle()
   end
 end
 
+--! Displays the fax/strike message to the player when opened from the bottom_panel.
 function UIMessage:openMessage()
-  if TheApp.world:checkPauseBehaviour() and not self.ui:checkForMustPauses() then
+  if TheApp.world:checkPauseBehaviour() and not self.ui:checkForMustPauseWindows() then
```
Alberth289346: The `checkPauseBehaviour` function has a different name too.

lewri: Looking at the wrong commit :)

CorsixTH/Lua/ui.lua
```diff
@@ -1066,6 +1066,15 @@ function UI:removeWindow(closing_window)
   end
 end
 
+--! Function to check if we have any must pause windows open
+--!return (bool) Returns true if a must pause window is found
+function UI:checkForMustPauseWindows()
+  for i in pairs(self.windows) do
+    if self.windows[i].mustPause() then return true end
+  end
```
mugmuggy, line 1074: Probably better to do

```lua
for _, window in pairs(self.windows) do
  if window:mustPause() then return true end
```
note the colon not the full stop.

Same issues manifest here from the previous change
https://github.com/CorsixTH/CorsixTH/blob/b7e6806f6953dee7bce47e5ce39eee8e4436faba/CorsixTH/Lua/ui.lua#L1050-L1056

But can refactor for code reuse.

mugmuggy (changes requested):

lewri:
Not sure why but my changes split itself into two commits... Anyway done.

CorsixTH/Lua/ui.lua
mugmuggy: Note still
```lua
if not pauseGame and closing_window.mustPause() then
```

full stop vs colon

CorsixTH/Lua/world.lua
```diff
@@ -954,6 +954,12 @@ function World:pauseOrUnpause()
   end
 end
 
+--! Function to check if player can perform actions when paused
```
TheCycoONE, line 957: Description and function name are slightly off. Maybe simply `isUserActionProhibited()` would be more appropriate. Pause and the config value are not necessarily important details to the caller.

CorsixTH/Lua/world.lua
```diff
@@ -954,6 +954,12 @@ function World:pauseOrUnpause()
...
+--!return (bool) Returns true if player hasn't allowed editing while paused
+function World:preventActionsWhenPaused()
```
lewri, line 959:
```suggestion
function World:isUserActionProhibited()
```

CorsixTH/Lua/dialogs/message.lua
```diff
@@ -104,7 +104,13 @@ function UIMessage:adjustToggle()
   end
 end
 
+--! Displays the fax/strike message to the player when opened from the bottom_panel.
 function UIMessage:openMessage()
+  if TheApp.world:preventActionsWhenPaused() and not self.ui:checkForMustPauseWindows() then
```
lewri, line 109:
```suggestion
  if TheApp.world:isUserActionProhibited() and not self.ui:checkForMustPauseWindows() then
```

CorsixTH/Lua/world.lua
lewri: This has been updated

PR #1763, opened 2020-12-13
[RDY] Maintain queue push-priority order

### LINKED ISSUE #1645: Chance of stuck doctors
#### Describe the issue
With the changes around the expected counts in 0.64rc, if you edit and cancel a room, that is treating patients, patients and staff in the room go through a leave and requeue process.

The logic may result in emergency patients arriving at the door to requeue before the staff member has popped from the queue. Thus inserting themselves before the staff member. tryAdvanceQueue will then prevent them from leaving as queue:front() is not able to enter the room.

This has quirked me in the past when looking at the code, it used to just work, but it would appear that Queue:push assumes the humanoids in the queue are already in the correct position with the humanoid.is_emergency check just looking for other emergency patients to queue behind.

I added a call to Queue:rerouteAllPatients in the Room:deactivate method that would appear to result in this and thus those already queued are resent ot the room upon the cancel. The old behaviour of reroute expected doesn't occur in TH, indeed they keep arriving whilst the edit room dialog is shown.

With 0.63 if you edit and cancel the edit for a room and have another of the same type the leaving patient will seek out that room as they cannot now find the one they left.

I could probably add another mode to 'rerouteAllPatients' or another method for just the epxected case to restore the functionality to the same as 0.63.

#### Steps to Reproduce
1. Load save game from #1616

2. Before unpausing, edit psych and cancel the edit

#### Save Game
*It is often useful for us if you have a save game from shortly before the
issue occurs that can be used to recreate the problem. Sometimes a save that
shows the issue happening may be useful as well. As GitHub currently only
supports uploading of images, you will have to upload your your savegames to
an another source such as [Google Drive](https://drive.google.com),
[Dropbox](https://dropbox.com), or
[SkyDrive/OneDrive](https://onedrive.live.com). Add the link to the file to
your issue and make the file(s) public, so we can access your gamelog or
savegame. The most important thing is to not remove these files after you
uploaded them!*

#### System Information

#### Gamelog.txt
```
Paste gamelog.txt output here
```

### Additional Info
illustrate the problem.*

mugmuggy:
*Fixes #1645*

- Checking increment_reported_size prevents changes to the queue order when its already been determined from previous priority - staff leaving the room will not move in front of a patient also leaving the room
- Before checks extended to include the priority from the previous checks - if a staff member is entering a room and a staff member and patient are queued to exit, the entering staff member will not move ahead of the leaving patient.

The Vip and Inspector don't actually queue at the moment, they just arrive at the door tile. They do at reception though, but emergency patients probably shouldn't but could go to reception I guess on some rare cases, and staff don't use it either, but the logic should be ok in those cases.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/queue.lua
+++ b/CorsixTH/Lua/humanoid_actions/queue.lua
@@ -208,7 +208,7 @@ local action_queue_on_change_position = permanent"action_queue_on_change_positio
   local queue = action.queue
   if not must_stand then
     for i = 1, queue.bench_threshold do
-      if queue[i] == humanoid then
+      if queue:reportedHumanoid(i) == humanoid then
         must_stand = true
         break
       end
--- a/CorsixTH/Lua/queue.lua
+++ b/CorsixTH/Lua/queue.lua
@@ -149,49 +149,40 @@ function Queue:setPriorityForSameRoom(entity)
   self.same_room_priority = entity
 end
 
+local function _isLeaving(queue, humanoid)
+  return queue.same_room_priority and queue.same_room_priority:getRoom() == humanoid:getRoom()
+end
+
+
+local function _humanoidQueuePriority(queue, humanoid)
+  if _isLeaving(queue, humanoid) then
+    return 1
+  elseif class.is(humanoid, Staff) then
+    return 2
+  elseif humanoid.is_emergency or class.is(humanoid, Vip) or class.is(humanoid, Inspector) then
+    return 3
+  else
+    return 4
+  end
+end
+
+local reported_priority_threshold = 2
+
 function Queue:push(humanoid, callbacks_on)
   local index = #self + 1
-  local increment_reported_size = true
-  if self.same_room_priority then
-    -- If humanoid in the priority room, then position them in the queue before
-    -- humanoids not in the room (because if they are in the room and in the
-    -- queue, then they are trying to leave the room).
-    local room = self.same_room_priority:getRoom()
-    if humanoid:getRoom() == room then
-      while index > 1 do
-        local before = self[index - 1]
-        if before:getRoom() == room then
-          break
-        end
-        index = index - 1
-      end
-      increment_reported_size = false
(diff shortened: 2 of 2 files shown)


TheCycoONE:
Queue.push is already a complex function and these changes (while correcting bugs) increase the complexity further.

I've been trying to think of what could be done to simplify it.

To start, lets make a friendly named private (or local) function:
```lua
function _isInsideRoom(humanoid)
  return self.same_room_priority and self.same_room_priority:getRoom() == humanoid:getRoom()
end
```
Then we can collapse one level of indentation (the condition starting on 155 and the condition starting on 160.

But really what we want is a table of priorities
```lua
function _humanoidQueuePriority(humanoid)
  if _isInsideRoom(humanoid) then
    return 1
  elseif class.is(humanoid, Staff) then
    return 2
  elseif humanoid.is_emergency or class.is(humanoid, Vip) or class.is(humanoid, Inspector) then
    return 3
  else then
    return 4
  end
end
```
Then you can collapse that function significantly because you only need one loop block to move the index to the appropriate position by priority.
You'll need a second condition to determine if the reported size should change.
`local increment_reported_size = not (self:_isInsideRoom(humanoid) or class.is(humanoid, Staff))`

I expect those changes will make it quite a bit shorter, and clearer.

mugmuggy:
Indeed that is a much cleaner solution and I did wonder why we just didn't have a queue priority. The increment_reported_size is just a proxy then for priority being > 2 and you need that anyway. As doors (except entrance), all call setPriorityForSameRoom, there are no FIFO doors implemented, so this same room priority check only prevents a queue push from another room, which obviously shouldn't happen and you could just humanoid:getRoom(). 

What it does do is associate it with a room via the door, rather than reception desk which allows the QueueAction to try and advance it at the bottom of the start function there. I'll get this included and update the queue action for the bench_threshold bug I mentioned in #1825

TheCycoONE:
I had kept the reported independent of the priorities on purpose because I think they're only coincidentally the same rather than fundamentally; but it's a minor thing, I'll take it as is.

PR #1764, opened 2020-12-15
[RDY] Use map size rather than hard coded 128 throughout

TheCycoONE:
Also fixes a clang-tidy warning about a variable that is set and never
used.


TheCycoONE:
I was just fixing issues found with newer clang tools, but the map size thing bothered me.

TheCycoONE:
New clang-format and old clang-format don't agree on the th_movie comment, so I'll revert it for now.

Alberth289346:
LGTM

PR #1768, opened 2020-12-24
[rdy] Objects belong to a hospital

Alberth289346:
- Previously, objects just lived in 'world', and sometimes used `world:getLocalPlayerHospital()` or `ui.hospital` for hospital access.
- This patch gives all objects a hospital reference of their own.
- Currently, the used hospital is always `world:getLocalPlayerHospital()`

*Other changes*
- Door deadlock resolve code didn't output anything which may be useful, so added that.
- Object creation code used continue magic, which was eliminated.

### DIFF
--- a/CorsixTH/Lua/objects/door.lua
+++ b/CorsixTH/Lua/objects/door.lua
@@ -166,6 +166,8 @@ function Door:checkForDeadlock()
     for _, action in ipairs(self.reserved_for.action_queue) do
       if action.name == "queue" then
         if action.queue ~= self.queue or self.queue[1] ~= self.reserved_for then
+          print("Warning: Trying to resolve door deadlock at ("
+              .. self.tile_x .. ", " .. self.tile_y .. ")")
           self.reserved_for = nil
           self:getRoom():tryAdvanceQueue()
         end
--- a/CorsixTH/Lua/objects/door.lua
+++ b/CorsixTH/Lua/objects/door.lua
@@ -166,8 +166,8 @@ function Door:checkForDeadlock()
     for _, action in ipairs(self.reserved_for.action_queue) do
       if action.name == "queue" then
         if action.queue ~= self.queue or self.queue[1] ~= self.reserved_for then
-          print("Warning: Trying to resolve door deadlock at ("
-              .. self.tile_x .. ", " .. self.tile_y .. ")")
+          self.world:gameLog("Warning: Trying to resolve door deadlock at ("
+              .. tostring(self.tile_x) .. ", " .. tostring(self.tile_y) .. ")")
           self.reserved_for = nil
           self:getRoom():tryAdvanceQueue()
         end
--- a/CorsixTH/Lua/objects/doors/entrance_right_door.lua
+++ b/CorsixTH/Lua/objects/doors/entrance_right_door.lua
@@ -36,22 +36,22 @@ class "EntranceDoor" (Object)
 ---@type EntranceDoor
 local EntranceDoor = _G["EntranceDoor"]
 
-function EntranceDoor:EntranceDoor(world, object_type, x, y, direction, etc)
+function EntranceDoor:EntranceDoor(hospital, object_type, x, y, direction, etc)
   self.is_master = object_type == object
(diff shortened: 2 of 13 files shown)


CorsixTH/Lua/objects/door.lua
```diff
@@ -166,6 +166,8 @@ function Door:checkForDeadlock()
     for _, action in ipairs(self.reserved_for.action_queue) do
       if action.name == "queue" then
         if action.queue ~= self.queue or self.queue[1] ~= self.reserved_for then
+          print("Warning: Trying to resolve door deadlock at ("
```
lewri, line 169: While I myself am a fan of ``print()`` I don't like it in the end code. A better convention perhaps (albeit achieves the same thing) is switching to ``self.world:gameLog()`` instead.

CorsixTH/Lua/objects/doors/entrance_right_door.lua
```diff
@@ -36,22 +36,22 @@ class "EntranceDoor" (Object)
 ---@type EntranceDoor
 local EntranceDoor = _G["EntranceDoor"]
 
-function EntranceDoor:EntranceDoor(world, object_type, x, y, direction, etc)
+function EntranceDoor:EntranceDoor(hospital, object_type, x, y, direction, etc)
   self.is_master = object_type == object
```
lewri, line 40: Food for thought: Slave & Master should be deprecated because of their historical background.
I know this isn't your fault actually but may be worth considering if you felt like it.

(Also present at circa line 45)

CorsixTH/Lua/objects/litter.lua
```diff
@@ -138,19 +137,23 @@ function Litter:anyLitter()
 end
 
 function Litter:afterLoad(old, new)
+  if old < 151 then
```
lewri, line 140: New ``afterLoad()``s should be at the bottom. (ascending order)

lewri (changes requested):
Generally looks fine. But as we're updating so many objects I feel like there may not be enough ``afterLoad()`` to handle existing saves.

I could be wrong though, and if so ignore me.

CorsixTH/Lua/objects/doors/entrance_right_door.lua
Alberth289346: I agree that master/slave in historic context was an extremely horrible idea, but I don't believe trying to erase these words from our language is the right answer.

The reason is that in historic context, the bad part is not master/slave as name for a type of relation that may exist. The bad part here is to enforce something (anything, in my view) onto other beings without their consent or opt-out. The latter is what we should fight, in my opinion. In historic context, the former is just one form of the latter. There are many other forms, eg weapon, war, sex.

Should we simply delete half our dictionary, because lots of things can be applied in a bad way? Does that solve anything? Will people in 5 generations remember and know about the bad things we erase now, so they can avoid them?

Alberth289346:
> But as we're updating so many objects I feel like there may not be enough afterLoad() to handle existing saves.

Litter derives from Entity, and has a new hospital parameter.
All other objects derived from Object, which also has a new hospital parameter.
As such, I think all objects have been covered.

CorsixTH/Lua/objects/door.lua
Alberth289346: done

CorsixTH/Lua/objects/litter.lua
Alberth289346: Done.

Code didn't have a hospital already, and below my addition it uses `self.hospital` without either introducing it or removing it in any existing afterLoad.
I wouldn't be surprised if the code just dies if you try loading old stuff like version 52.

lewri (approved):

TheCycoONE:
Not a fan of the createMapObjects refactor stylistically. If you're going to that much effort to rewrite the loop, why not put the body in a separate function e.g. `_createMapObject` and return early on the fail conditions. Lot less nesting to keep straight that way and no continue hack.

Alberth289346:
I thought I needed more invasive changes to `newObject` but it turned out not to be the case (at least for now).

PR #1777, opened 2020-12-28
[rdy] Leveledit fixes

Alberth289346:
Level editor crashed on the command line when following the README instructions.

- Use generics where needed.
- Allow startup without icon files.
- Update the README file.

### DIFF
--- a/LevelEdit/README.txt
+++ b/LevelEdit/README.txt
@@ -14,7 +14,8 @@ to compile. An alternative is to issue the following commands while in this
 folder:
 
 Compile:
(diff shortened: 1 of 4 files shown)


TheCycoONE:
Interesting - what version are we supporting for level edit? Is it worth using jlink etc. and making a real distributable out of it?

Alberth289346:
I used Java 8 and that runs.

It could use more love though, or we should add a new mode to the lua program perhaps so we can remove the java version

Alberth289346:
Missed adding [rdy], sorry

PR #1779, opened 2020-12-28
[RDY] Forbid loading new saves on old versions when outside debug

lewri:
- Completing a TODO to prevent newer saves opening in older games.
- This change will only have a real effect post savegame 152.
- Also saves the savegame_release version to world so an old game knows where the newer save came from (for compatibility it uses Trunk as fallback/nil). 
- Allows saves to be opened in debug but shows a similar warning.

When debug is on: (NB: updated 29/12)
image

When debug is off: (NB: updated 29/12)
image

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1519,19 +1519,13 @@ end
 
 --! Function to check the savegame version matches or is older than the installation of CorsixTH
 --!param save_version (num)
---!param save_release (string) Game's human readable game version
 --!return true if compatible, otherwise false
-function App:checkCompatibility(save_version, save_release)
+function App:checkCompatibility(save_version)
   local app_version = self.savegame_version
-  local version_name = self:getVersion(app_version)
-  local save_release_ver = save_release or "Trunk" -- For compatibility
-  local app_version_humanize = tostring(version_name .. " [" .. app_version .. "]")
-  local save_version_humanize = tostring(save_release_ver .. " [" .. save_version .. "]")
-
   if app_version >= save_version or self.config.debug then
     return true
   else
-    local err = _S.errors.compatibility_error:format(save_version_humanize, app_version_humanize)
+    local err = _S.errors.compatibility_error
     UILoadGame:loadError(err)
     return false
   end
@@ -1596,18 +1590,24 @@ function App:afterLoad()
     self.world.original_savegame_version = old
   end
   local first = self.world.original_savegame_version
+
+  -- Generate the human-readable version number (old [loaded save], new [game], first [original])
+  local first_version = first .. " (" .. self:getVersion(first) .. ")"
+  local old_version = old .. " (" .. self:getVersion(old) .. ")"
+  local new_version = new .. " (" .. self:getVersion() .. ")"
+
   if new == old then
-    self.world:gameLog("Savegame version is " .. new .. " (" .. self:getVersion() ..
-        "), originally it was " .. first .. " (" .. self:getVersion(first) .. ")")
+    local msg_same = "Savegame version is %s, originally it was %s."
+    self.world:gameLog(msg_same:format(new_version, first_version))
     self.world:playLoadedEntitySounds()
   elseif new > old then
-    self.world:gameLog("Savegame version changed from " .. old .. " (" .. self:getVersion(old) ..
-                       ") to " .. new .. " (" .. self:getVersion() ..
-                       "). The save was created using " .. first ..
-                       " (" .. self:getVersion(first) .. ")")
+    local msg_older = "Savegame changed from %s to %s. The save was created using %s."
+    self.world:gameLog(msg_older:format(old_version, new_version, first_version))
   else -- Save is newer than the game and can only proceed in debug mode
-    self.world:gameLog("Warning: loaded savegame version " .. old .. " (" .. self:getVersion(old) ..
-                       ")" .. " in older version " .. new .. " (" .. self:getVersion() .. ").")
+    local get_old_release_version = self.world.release_version or "Trunk" -- For compatibility
+    old_version = old .. " (" .. get_old_release_version .. ")"
+    local msg_newer = "Warning: loaded savegame version %s in older version %s."
+    self.world:gameLog(msg_newer:format(old_version, new_version))
     self.ui:addWindow(UIInformation(self.ui, {_S.warnings.newersave}))
   end
   self.world.release_version = self:getVersion()
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1517,6 +1517,26 @@ function App:quickLoad()
   end
 end
 
+--! Function to check the savegame version matches or is older than the installation of CorsixTH
+--!param save_version (num)
+--!param save_release (string) Game's human readable game version
+--!return true if compatible, otherwise false
+function App:checkCompatibility(save_version, save_release)
+  local app_version = self.savegame_version
+  local version_name = self:getVersion(app_version)
+  local save_release_ver = save_release or "Trunk" -- For compatibility
+  local app_version_humanize = tostring(version_name .. " [" .. app_version .. "]")
+  local save_version_humanize = tostring(save_release_ver .. " [" .. save_version .. "]")
+
+  if app_version >= save_version or self.config.debug then
+    return true
+  else
+    local err = _S.errors.compatibility_error:format(save_version_humanize, app_version_humanize)
+    UILoadGame:loadError(err)
+    return false
+  end
+end
+
 --! Restarts the current level (offers confirmation window first)
 function App:restart()
   assert(self.map, "Trying to restart while no map is loaded.")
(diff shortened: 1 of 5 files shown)


Alberth289346:
Quite sure you should not mention the work-around there for non-debug, as users will simply do what you tell them to without understanding.
Even if you tell them it's on their own risk they will continue, and then report a problem when it crashes etc.

Maybe only allow this stuff in a trunk version with debug on.

CorsixTH/Lua/app.lua
```diff
@@ -1517,6 +1517,26 @@ function App:quickLoad()
...
+--! Function to check the savegame version matches or is older than the installation of CorsixTH
+--!param save_version (num)
+--!param save_release (string) Game's human readable game version
+--!return true if compatible, otherwise false
+function App:checkCompatibility(save_version, save_release)
+  local app_version = self.savegame_version
+  local version_name = self:getVersion(app_version)
+  local save_release_ver = save_release or "Trunk" -- For compatibility
+  local app_version_humanize = tostring(version_name .. " [" .. app_version .. "]")
+  local save_version_humanize = tostring(save_release_ver .. " [" .. save_version .. "]")
```
Alberth289346, line 1529: Do one string at a time, line 1527 now switches to another version text, and 1528 switches back again to the previous conversion. Intuitively, you read it as 1 sequence of assignments, (`b = ..a .. ; c = .. b ..; d = ..c..;` etc) but it's not.

Use an empty line between them to show the reader how to read it.

Can't you fold the string concatenation stuff at 1528 and 1529 into the `_S.errors.compatibility_error` ? (would only work if it's a new string I guess)

TheCycoONE (commented):
I think this PR is overkill for an issue I have never actually seen in real life. Have you heard of anyone running into this @lewri ?

I'd prefer we didn't add additional complexity or state - on the off chance it does happen `This save is not compatible with your copy of CorsixTH` is probably sufficient. I don't see the value in a debug mode workaround either.

CorsixTH/Lua/app.lua
```diff
@@ -1517,6 +1517,26 @@ function App:quickLoad()
...
+
+  if app_version >= save_version or self.config.debug then
+    return true
+  else
+    local err = _S.errors.compatibility_error:format(save_version_humanize, app_version_humanize)
```
Alberth289346, line 1534: Why do you compute strings to report as error before you decided there is an error?

CorsixTH/Lua/app.lua
```diff
@@ -1585,11 +1605,12 @@ function App:afterLoad()
                        ") to " .. new .. " (" .. self:getVersion() ..
                        "). The save was created using " .. first ..
                        " (" .. self:getVersion(first) .. ")")
```
Alberth289346, line 1607: Would be nice if you could refactor this to a string.format call, this looks beyond ugly imho

CorsixTH/Lua/app.lua
```diff
@@ -1585,11 +1605,12 @@ function App:afterLoad()
...
-  else
-    -- TODO: This should maybe be forbidden completely.
+  else -- Save is newer than the game and can only proceed in debug mode
     self.world:gameLog("Warning: loaded savegame version " .. old .. " (" .. self:getVersion(old) ..
                        ")" .. " in older version " .. new .. " (" .. self:getVersion() .. ").")
+    self.ui:addWindow(UIInformation(self.ui, {_S.warnings.newersave}))
   end
+  self.world.release_version = self:getVersion()
   self.world.savegame_version = new
```
Alberth289346, line 1614: Variables are created in world. Is that documented in World?

I am somewhat on a crusade to at least state in a constructor of a class which variables may exist, and what they mean. As the code is filled with this random creation of new variables outside the class, my journey may never end though :)

CorsixTH/Lua/app.lua
```diff
@@ -1585,11 +1605,12 @@ function App:afterLoad()
                        ") to " .. new .. " (" .. self:getVersion() ..
                        "). The save was created using " .. first ..
                        " (" .. self:getVersion(first) .. ")")
-  else
-    -- TODO: This should maybe be forbidden completely.
+  else -- Save is newer than the game and can only proceed in debug mode
     self.world:gameLog("Warning: loaded savegame version " .. old .. " (" .. self:getVersion(old) ..
                        ")" .. " in older version " .. new .. " (" .. self:getVersion() .. ").")
```
Alberth289346, line 1610: Slightly less ugly, but using `format` is likely an improvement.

CorsixTH/Lua/dialogs/resizables/file_browsers/load_game.lua
```diff
@@ -41,12 +41,16 @@ function UILoadGame:choiceMade(name)
...
+    self:loadError(err)
   end
 end
 
+function UILoadGame:loadError(err)
+  print(err)
+  TheApp:loadMainMenu()
+  TheApp.ui:addWindow(UIInformation(TheApp.ui, {err}))
+end
+
```
Alberth289346, line 53: Not sure why you added a function, as it seems used only one time.

It's missing documentation.

CorsixTH/Lua/dialogs/resizables/file_browsers/load_game.lua
```diff
@@ -41,12 +41,16 @@ function UILoadGame:choiceMade(name)
   local status, err = pcall(app.load, app, name)
   if not status then
     err = _S.errors.load_prefix .. err
-    print(err)
-    app:loadMainMenu()
-    app.ui:addWindow(UIInformation(self.ui, {err}))
+    self:loadError(err)
```
Alberth289346, line 44: You moved from multiple uses of the updated `err` to one use of it in a function call. At that point a perhaps good alternative is not to assign to the variable, but instead move the entire expression into the call.

CorsixTH/Lua/dialogs/resizables/file_browsers/load_game.lua
lewri: This one is used from App:checkCompatibility and the pcall in the function above loadError. I'll get the documentation added.

CorsixTH/Lua/app.lua
lewri: The "error" happens here at 1531 if false.

CorsixTH/Lua/persistance.lua
```diff
@@ -260,10 +260,13 @@ function SaveGameFile(filename)
   f:close()
 end
 
+--! Loads the game's save into the game
```
Alberth289346, line 263: Nice recursion putting games in games :) 

As you're loading here, perhaps avoid mention 'save'?  "loaded file' is clear enough I think.
Also, it misses a description for `data` (which is somewhat too generic to my taste, literally everything is 'data').

CorsixTH/Lua/persistance.lua
```diff
@@ -260,10 +260,13 @@ function SaveGameFile(filename)
...
 function LoadGame(data)
   --local status, res = xpcall(function()
   local objtable = MakePermanentObjectsTable(true)
   local state = assert(persist.load(data, objtable))
+  -- Here we want to check the game we're loading is same or older than the current version
+  if not TheApp:checkCompatibility(state.world.savegame_version, state.world.release_version) then return end
```
Alberth289346, line 269: You want to check or are you checking?

If the former, it likely needs a `TODO`, if the latter, the code line seems to quite nicely express what you're aiming for.
(Same or older than current version  is what you actually check, ie it's the implementation of `checkCompatibility`, so that belong more inside the function than above it.)

CorsixTH/Lua/world.lua
```diff
@@ -130,6 +130,7 @@ function World:World(app)
   self.floating_dollars = {}
   self.game_log = {} -- saves list of useful debugging information
   self.savegame_version = app.savegame_version
```
Alberth289346, line 132: Please write for both variables that you use their meaning here in a line of text. In that way you avoid that you or someone else has to reverse engineer the meaning that you already found or created.

CorsixTH/Lua/app.lua
Alberth289346: yes, so computing the text can be done after line 1533

Alberth289346:
Generally nice work, just missing a few dots on the `ı` here and there.

CorsixTH/Lua/app.lua
```diff
@@ -1517,6 +1517,20 @@ function App:quickLoad()
   end
 end
 
+--! Function to check the savegame version matches or is older than the installation of CorsixTH
+--!param save_version (num)
+--!return true if compatible, otherwise false
+function App:checkCompatibility(save_version)
```
Alberth289346, line 1523: Abstraction level of 1520 doesn't match with function name.

It should describe what "compatible" means (ie loaded game can be run in the program), rather than how you compute it. If you find a newer or better way to decide compatibility, the function description shouldn't need to change.

CorsixTH/Lua/app.lua
```diff
@@ -1517,6 +1517,20 @@ function App:quickLoad()
...
+  local app_version = self.savegame_version
+  if app_version >= save_version or self.config.debug then
+    return true
+  else
+    local err = _S.errors.compatibility_error
+    UILoadGame:loadError(err)
```
Alberth289346, line 1529: Could merge 1528 and 1529

CorsixTH/Lua/app.lua
```diff
@@ -1576,20 +1590,27 @@ function App:afterLoad()
     self.world.original_savegame_version = old
   end
   local first = self.world.original_savegame_version
+
+  -- Generate the human-readable version number (old [loaded save], new [game], first [original])
```
Alberth289346, line 1594: 'game' is a bit generic, maybe 'program' ?

lewri: Or "App" these days :)

CorsixTH/Lua/persistance.lua
```diff
@@ -260,10 +260,14 @@ function SaveGameFile(filename)
   f:close()
 end
 
+--! Puts loaded file into the game
+--!param data The file
 function LoadGame(data)
   --local status, res = xpcall(function()
   local objtable = MakePermanentObjectsTable(true)
   local state = assert(persist.load(data, objtable))
+  -- Check the game we're loading is same or older than the current version
```
Alberth289346, line 269: No, you're checking compatibility, which today is comparing versions, and tomorrow can be different.

PR #1790, opened 2020-12-31
[RDY] Add default macos font file path

tobylane:
- Detect the OS - Windows/macOS/Unix. In the future this could look for Android, Wii, whatever needs custom UI?
- Use this to select a default unicode font on macos.

### DIFF
--- a/CorsixTH/CMakeLists.txt
+++ b/CorsixTH/CMakeLists.txt
@@ -46,6 +46,15 @@ else()
   add_executable(CorsixTH "")
 endif()
 
+# Report operating system in compile_opts.os
+if(WIN32)
+  set(CORSIX_TH_OS "windows")
+elseif(APPLE)
+  set(CORSIX_TH_OS "macos")
+else()
+  set(CORSIX_TH_OS "unix")
+endif()
+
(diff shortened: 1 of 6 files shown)


TheCycoONE:
I'd prefer this was exposed from c++, e.g. we have a pretty good idea of the OS in cmake, we could pass that in through config.h.in and expose it through that GetCompileOptions function (bumping the api version).

I can merge this as is and open another issue for that if you like @tobylane ?

tobylane:
Made the change you described. Another option is to pass [CMAKE_SYSTEM_NAME](https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-To-Write-Platform-Checks#cmake-system) through and clean it up in lua?

TheCycoONE:
Yes, I like this. CMAKE_SYSTEM_NAME might be useful in the future if we need to discriminate more than windows, apple, and other; and is potentially more accurate. On the other hand it's harder to use, for the moment we don't need the granularity, and breaking the api isn't a big deal. I'd take what you have.

tobylane:
Ready.

PR #1794, opened 2021-01-02
Serious Radation - Stage 1

### LINKED ISSUE #1659: Serious Radiation Graphics
#### Describe the issue
In the original Theme Hospital, patients with serious radiation had a green filter on them. For some reason, patients with Serious Radiation in CorsixTH aren't distinguishable from other patients.
#### Steps to Reproduce

2. Wait until a patient with Serious Radiation comes in.

3.

#### Save Game
https://drive.google.com/file/d/1KG4ZDYoZavyKyDXmbJShr_XMi09_KH__/view?usp=sharing

#### Expected Behavior
I expect the same green filter on Serious Radiation patients as is found in the original game.
#### System Information
**CorsixTH Version:** 0.64rc-2

#### Gamelog.txt
```
Paste gamelog.txt output here
```

### Additional Info
image How it's supposed to look.
Serious Radiation How it looks in my game.

driverpt:
*Fixes #1659*

This is the first iteration of applying Transformations to existing textures.


.luacheckrc
```diff
@@ -42,7 +42,7 @@ globals = { -- Globals
             "SaveGame", "SaveGameFile", "Staff", "StaffProfile", "StaffRoom",
             "Strings", "SwingDoor", "TheApp", "TreeControl", "Vip", "Window",
             "World", "Date", "Doctor", "Handyman", "Nurse", "Receptionist",
-            "Cheats",
+            "Cheats", "AnimationOverlayFlags",
```
lewri, line 45: Why I was lazy and put "Cheats" at the end of this list is a bad move by myself.
Ideally we want to keep them in alphabetic order (but blame myself for this :) )

CorsixTH/Src/singleton.h
```diff
@@ -0,0 +1,25 @@
...
+};
+
+template <class T>
+T& THSingleton<T>::get()
+{
+    static T instance;
+    return instance;
+}
+
+#endif // CORSIX_TH_TH_SINGLETON_H_
```
lewri, line 24: Looks like we want an empty line here

CorsixTH/Src/th_gfx_common.h
```diff
@@ -0,0 +1,39 @@
+/*
+Copyright (c) 2009-2013 Peter "Corsix" Cawley and Edvin "Lego3" Linge
```
lewri, line 2: You can delete Corsix and Lego3 out of this copyright, as it's "your" file :)

CorsixTH/Src/th_gfx_common.h
```diff
@@ -0,0 +1,39 @@
...
+enum animation_overlay_flags : uint32_t {
+  //! No Overlay
+  thaof_none = 0,
+  //! Serious Radiation Overlay
+  thaof_glowing = 1 << 0,
+  //! Jellyitis animation
+  thaof_jelly = 1 << 1,
+};
+
+#endif
```
lewri, line 39: Again looks like we need a new line here at the end

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -37,6 +37,8 @@ SOFTWARE.
 
 #include "th_map.h"
 
+#define PI 3.14159265
```
lewri, line 40: This may be undesired and instead use actual pi (but who am I to know).

CorsixTH/Src/th_lua_anims.cpp
```diff
@@ -701,6 +709,8 @@ void lua_register_anims(const lua_register_state* pState) {
     lcb.add_function(l_anim_tick<animation>, "tick");
     lcb.add_function(l_anim_draw<animation>, "draw", lua_metatable::surface);
     lcb.add_function(l_anim_set_drawable_layer, "setDrawingLayer");
+    lcb.add_function(l_anim_set_glowing, "setOverlay");
+    // TODO: Add Jellyatis condition overlay
```
lewri, line 713: Jellyitis

lewri (commented):
Won't pretend I know about CPP so will leave that to someone else. But here's the help I could give.

CorsixTH/Src/th_gfx_common.h
lewri: And year should be 2020

CorsixTH/Lua/diseases/serious_radiation.lua
```diff
@@ -43,6 +43,7 @@ disease.initPatient = function(patient)
   patient:setLayer(1, math.random(0, 3) * 2)
   patient:setLayer(3, 0)
   patient:setLayer(4, 0)
+  patient.th:setOverlay(AnimationOverlayFlags.Glowing)
```
lewri, line 46: Can't add my comment at line 1. I know most of your work is in other files but as a summation of that I would recommend changing the top to:
``--[[ Copyright (c) 2010 Manuel "Roujin" Wolf, 2020 <name> "driverpt" <surname>``
(Fill name and surname of course only if you want to)

.luacheckrc
TheCycoONE: I also blame lewri :p

CorsixTH/Lua/diseases/serious_radiation.lua
TheCycoONE: Certainly in LICENSE.txt - if you don't, I will before the next release. Here I don't think it's justified, but files with more substantial changes, certainly.

CorsixTH/Src/sdl_core.cpp
```diff
@@ -29,6 +29,23 @@ SOFTWARE.
 #include "lua_sdl.h"
 #include "th_lua.h"
 
+
+GlobalTickCounter::GlobalTickCounter() {
+  counter = 0;
```
TheCycoONE: Assign defaults in the initializer list not the body of the constructor.

`GlobalTickCounter::GlobalTickCounter() : counter(0) {}`

CorsixTH/Src/singleton.h
TheCycoONE: A trailing newline not an empty line, but yes.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -1298,7 +1299,7 @@ bool THAnimation_is_multiple_frame_animation(drawable* pSelf) {
 animation_base::animation_base() : drawable() {
   x_relative_to_tile = 0;
   y_relative_to_tile = 0;
-  for (int i = 0; i < 13; ++i) {
+  for (int i = 0; i < MAX_NUMBER_OF_LAYERS; ++i) {
```
TheCycoONE, line 1305: We only use screaming snake case for macros per: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-all-caps

CorsixTH/Src/th_gfx.h
```diff
@@ -22,10 +22,14 @@ SOFTWARE.
 
 #ifndef CORSIX_TH_TH_GFX_H_
 #define CORSIX_TH_TH_GFX_H_
+
+#define MAX_NUMBER_OF_LAYERS 13
```
TheCycoONE, line 26: This should be a constexpr not a macro.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -1479,6 +1481,15 @@ void animation::depersist(lua_persist_reader* pReader) {
   pReader->set_error("Cannot depersist animation instance");
 }
 
+void animation::set_overlay(animation_overlay_flags flags) {
+  // Sanity Check
```
TheCycoONE, line 1492: Fail early, or don't do this check and just accept the flag. I don't like silently changing the input.

CorsixTH/Src/th_gfx_common.h
TheCycoONE: th_gfx.h is the common graphics file. I'm not sure why this is needed.

CorsixTH/Src/th_gfx_common.h
```diff
@@ -0,0 +1,39 @@
...
+*/
+#ifndef CORSIX_TH_TH_GFX_COMMON_H_
+#define CORSIX_TH_TH_GFX_COMMON_H_
+
+#include "config.h"
+
+// This header is meant to be shared with GFX and GFX_SDL
+
+//! Animation Overlay Flags for drawing an overlay on top of all the frame
+enum animation_overlay_flags : uint32_t {
```
TheCycoONE, line 30: You can't have both, I'm not sure a bit set is the right thing here. You could use an `enum class` here and enforce that exactly one of these options is set, which removes the need for your earlier check. You will need to explicitly cast at the lua binding though.

.luacheckrc
driverpt: I can put alphabetically. I prefer to add chronologically by the time it was added to the code.

CorsixTH/Src/th_gfx_common.h
driverpt: Should be fixed now

CorsixTH/Src/th_gfx_sdl.cpp
TheCycoONE: Nope, not until c++ 20 and were at 14. Should be constexpr pi = ... though.

CorsixTH/Lua/diseases/serious_radiation.lua
driverpt: I think VCS will do that for you

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -95,6 +97,57 @@ uint8_t convert_6bit_to_8bit_colour_component(uint8_t colour_component) {
       (colour_component & mask_6bit) * static_cast<double>(0xFF) / mask_6bit));
 }
 
+double hueToRGB(double p, double q, double hue) {
```
TheCycoONE, line 100: we're weird and use snake case, like iso c++.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -95,6 +97,57 @@ uint8_t convert_6bit_to_8bit_colour_component(uint8_t colour_component) {
...
+  if (hue < 0) {
```
TheCycoONE, line 101: Should these be `while` instead of if - or is hue only ever slightly out of range from the 1/3rd adjustment?

CorsixTH/Src/th_gfx_sdl.cpp
TheCycoONE, line 100: This method is full of magic numbers. They probably mean something to someone that knows more about color space than I do, but I would appreciate some guidance in comments or named consts. Explaining the whole thing more will be useful.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -1044,10 +1103,33 @@ void sprite_sheet::draw_sprite(render_target* pCanvas, size_t iSprite, int iX,
     }
   }
 
+  if ((overlayFlags & thaof_glowing) != 0) {
```
TheCycoONE, line 1110: This will change if you end up using the enum class.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -1044,10 +1103,33 @@ void sprite_sheet::draw_sprite(render_target* pCanvas, size_t iSprite, int iX,
...
+    // We want this to vary between 155 -> 205 -> 255.
+    // We're using increments of 10 degrees within the Sin function and converting them to rad
+    // float currentVariation = sin(tickNumber*10*PI/180) * 20;
+    uint32_t tick = gTickCount.currentTick();
```
TheCycoONE, line 1114: SDL_GetTick - you don't need your custom tick class / code at all.

TheCycoONE (changes requested):
All the global tick counter / singleton stuff can go away, you just need the time the program has been running which is available with SDL_GetTick. I would also suggest making your animations distinct rather than able to apply together as the game doesn't require it and it complicates the code.

CorsixTH/Src/th_gfx_sdl.cpp
driverpt: Already ahead of you :)

CorsixTH/Lua/diseases/serious_radiation.lua
driverpt: @TheCycoONE , i'm already there

```
Copyright (C) 2012-2013 Luis "Driver" Duarte
```

CorsixTH/Src/th_gfx.cpp
driverpt: What do you suggest to use ?

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -1044,10 +1057,34 @@ void sprite_sheet::draw_sprite(render_target* pCanvas, size_t iSprite, int iX,
...
+      throw std::runtime_error(SDL_GetError());
+    }
+  }
+
   SDL_Rect rcSrc = {0, 0, sprite.width, sprite.height};
   SDL_Rect rcDest = {iX, iY, sprite.width, sprite.height};
 
   pCanvas->draw(pTexture, &rcSrc, &rcDest, iFlags);
+
+  if ((overlayFlags & thaof_glowing) != 0) {
```
TheCycoONE, line 1081: Can this be an else of the above, or is it important for some reason that this is after the draw and the other condition is before the draw?

CorsixTH/Src/lua_sdl.h
```diff
@@ -40,6 +40,8 @@ SOFTWARE.
 // SDL_USEREVENT_SOUND_OVER - informs script of a played sound finishing.
 #define SDL_USEREVENT_SOUND_OVER (SDL_USEREVENT + 4)
 
+#define TICKS_MS_TO_USERTICKS_CONVERSION 30
```
TheCycoONE, line 43: `constexpr usertick_period_ms = 30;` is a better name.

CorsixTH/Src/th_gfx_sdl.cpp
driverpt: I don't think so, since it's two different things for what we're branching the code.

Ideally we would have some sort of Lifecycle to Draw the Animations and just add Hooks like `onPreRender`

CorsixTH/Src/lua_sdl.h
driverpt: Should we drop the Macro ?

CorsixTH/Lua/diseases/serious_radiation.lua
TheCycoONE: Ah, welcome back, just need to add a `, 2020` or change `2013` to `2020` - up to you.

CorsixTH/Src/th_gfx.cpp
lewri: Going by the lower comment of constexpr I would say normal snake case...

CorsixTH/Src/lua_sdl.h
TheCycoONE: Yes. http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es31-dont-use-macros-for-constants-or-functions

.luacheckrc
lewri: I think my "Cheats" should be moved onto Line 35. Yours would go nicely before "App" on Line 34.

TheCycoONE:
Not sure if you saw the M_PI failure for MSVC. Unfortunately M_PI is non-standard. C++ 20 adds std::numbers::pi but we're not on 17 yet much less 20. I recommend just dumping a pi constexpr in config.h or th.h

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -196,12 +202,15 @@ void full_colour_renderer::decode_image(const uint8_t* pImg,
...
+        uint32_t iColour;
         pImg++;  // Skip reading the opacity for now.
         if (iTable == 0xFF) {
           // Legacy sprite data. Use the palette to recolour the
           // layer. Note that the iOpacity is ignored here.
           while (iLength > 0) {
-            push_pixel(pColours[*pImg++]);
+            iColour = pColours[*pImg++];
+
+            push_pixel(iColour);
```
Alberth289346, line 213: This looks like a unneeded change, code hasn't changed functionally, and you don't use the new variable elsewhere.

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: @TheCycoONE  An else? condition is the same as at 1060

CorsixTH/Src/th_lua_anims.cpp
```diff
@@ -701,6 +710,8 @@ void lua_register_anims(const lua_register_state* pState) {
     lcb.add_function(l_anim_tick<animation>, "tick");
     lcb.add_function(l_anim_draw<animation>, "draw", lua_metatable::surface);
     lcb.add_function(l_anim_set_drawable_layer, "setDrawingLayer");
+    lcb.add_function(l_anim_set_glowing, "setOverlay");
```
Alberth289346, line 713: lua function name doesn't quite match the cpp function name

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: if teh new variable is required, at least make it local to its use.

TheCycoONE:
@driverpt are you still interested in this PR, or would you prefer one of us run with it? I think it's too late for 0.65 but given how close you got I'd love to have this feature in the next version.

driverpt:
Hello. No

TheCycoONE:
Superseded by #1938

PR #1798, opened 2021-01-09
[RDY] Disable changing audio or movie settings when compiled without audio

tobylane:
- Prevents the button up error when changing global audio setting without audio in the build.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1706,6 +1706,10 @@ function App:finishVideoUpdate()
   self.video:startFrame()
 end
 
+function App:isAudioEnabled()
+  return TH.GetCompileOptions().audio
+end
+
 -- Do not remove, for savegame compatibility < r1891
 local app_confirm_quit_stub = --[[persistable:app_confirm_quit]] function()
 end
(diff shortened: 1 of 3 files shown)


TheCycoONE:
I suppose, but why let them press the button at all?

lewri:
Are you proposing to make the ``sorry.wav`` sound instead?
Unless that build setting also also turns off any sounds at all...

tobylane:
> I suppose, but why let them press the button at all?

It could be set to off and unchangeable? I don't recall any other buttons which can be disabled.
My thought for this was to let them do it and don't break.

> Are you proposing to make the `sorry.wav` sound instead?
> Unless that build setting also also turns off any sounds at all...

Yes, no sounds at all.

Alberth289346:
Not crashing has a lot of advantages :) 

I am fine with the button just glued in the "off" position (ie clicking has no effect). Disabling would be nice, but not required imho.
Other option is to pop up a warning that you cannot turn on sound.

tobylane:
This disables the three buttons that control audio dependant parts.

TheCycoONE:
LGTM

tobylane:
Ready.

PR #1800, opened 2021-01-16
[RDY]Save to OS temp directory if the intended file can't be written to

### LINKED ISSUE #470: Configuration and save games reset on Mac OS X
_From [forward....@gmail.com](https://code.google.com/u/102141427359698628586/) on July 08, 2011 15:45:59_

What steps will reproduce the problem? Unfortunately nothing seemed to go wrong while playing, the game exited as normal. When starting up an error during utility.loadfile line 123 occurred (result is nil).

Looking at the code this occurs when the config.txt file exists but it has no data in it.

After finding the config file it was zero bytes and had no data.

Deleting the config file caused it to be recreated from scratch when corsix started up, however the previous active save game file and the autosave file were both zero bytes (we were at level 11!). What is the expected output? What do you see instead? I expect the save game files to remain in the correct state.
My suggestion would be that corsix backs up the file it is about to overwrite and then performs the overwrite. This will at least allow a manual restoration of game state.

Beta 7. What operating system / compile settings are you using? Mac OS X

Full game level 11. Attach a saved game file if you have one and it is relevant. Please provide any additional information below.

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=928_

tobylane:
*Fixes #470*

- On failing to write to the intended save game file it saves to a temp file and tells the user

### DIFF
--- a/CorsixTH/Lua/persistance.lua
+++ b/CorsixTH/Lua/persistance.lua
@@ -255,7 +255,7 @@ end
 --!param filename (string) Path of the file to write.
 function SaveGameFile(filename)
   local data = SaveGame()
-  local f = assert(io.open(filename, "wb"))
+  local f = TheApp:writeToFileOrTmp(filename)
   f:write(data)
   f:close()
 end
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -869,7 +869,8 @@ end
 
 function App:saveConfig()
   -- Load lines from config file
-  local fi = io.open(self.command_line["config-file"] or "config.txt", "r")
+  local config_file = self.command_line["config-file"] or "config.txt"
+  local fi = io.open(config_file, "r")
   local lines = {}
   local handled_ids = {}
   if fi then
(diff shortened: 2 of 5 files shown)


CorsixTH/Lua/persistance.lua
```diff
@@ -255,7 +255,14 @@ end
 --!param filename (string) Path of the file to write.
 function SaveGameFile(filename)
   local data = SaveGame()
-  local f = assert(io.open(filename, "wb"))
+  local f, err = io.open(filename, "wb")
+  if err then
+    local file = os.tmpname()
+    f = io.open(file, "wb")
+    TheApp.ui:addWindow(UIInformation(TheApp.ui,
+        {"Saved " .. filename .. " to " .. file, "Error:", err}))
```
TheCycoONE, line 263: This is (potentially) user facing. Should use the language file and make a user friendly ish message.

tobylane: Done, added _S.errors.save_to_tmp.

tobylane:
Made it generic and applied it to the other main io.writes. A few minor ones were left out, the first two here https://grep.app/search?q=%22wb%3F%22&regexp=true&filter[repo][0]=CorsixTH/CorsixTH&filter[lang][0]=Lua&filter[path][0]=CorsixTH/Lua/

Not sure about the App:writeToFileOrTmp name.

Apologies for the multiple pushes.

tobylane:
Rebased to pass tests.

tobylane:
I can't re run the AppVeyor test but it's a Lua only change. Ready.

PR #1801, opened 2021-01-16
[RDY] Receptionists can be placed on unstaffed reception desks

### LINKED ISSUE #526: Bad for new players: Unable to place new receptionists on top of the desk

tobylane:
*Fixes #526*
Previous feedback #1739 

- Rewrite placing staff
- Permit receptionists being put on any tile of a desk

### DIFF
--- a/CorsixTH/Lua/dialogs/place_staff.lua
+++ b/CorsixTH/Lua/dialogs/place_staff.lua
@@ -77,15 +77,34 @@ end
 
 local flag_cache = {}
 local flag_altpal = 16
+--! Private function. Is the staff member (self) permitted
+-- to be placed where the user is hovering or clicking?
+function UIPlaceStaff:_isValidStaffPlacement()
+  local world, x, y = self.world, self.tile_x, self.tile_y
+  world.map.th:getCellFlags(x, y, flag_cache)
+  -- Is the tile inside the player's hospital building?
+  if not (self.ui.hospital:getPlayerIndex() == flag_cache.owner and
+      flag_cache.hospital) then
+    return false
+  end
+  local room = world:getRoom(x, y)
+  -- On a tile humanoids can walk on?
+  local walkable = flag_cache.passable and (not room and true or not room.crashed)
+  -- and in an area the regular staff (doctors, nurses and handymen) can go?
+  local staffable = (self.allow_in_rooms or flag_cache.roomId == 0)
+  -- Or is it a receptionist placed on an unstaffed reception desk?
+  local reception = false
+  if self.profile.humanoid_class == "Receptionist" then
+    local desk = world:getObject(x, y, "reception_desk") or
+        world:findObjectNear(self, "reception_desk", 0)
+    reception = desk and not desk.receptionist
+  end
+  return (walkable and staffable) or reception
+end
+
 function UIPlaceStaff:draw(canvas)
   if self.world.user_actions_allowed then
-    self.world.map.th:getCellFlags(self.tile_x, self.tile_y, flag_cache)
-    local room = self.world:getRoom(self.tile_x, self.tile_y)
-    local player_id = self.ui.hospital:getPlayerIndex()
-    local valid = flag_cache.hospital and flag_cache.passable and
-        (self.allow_in_rooms or flag_cache.roomId == 0) and
-        (not room and true or not room.crashed) and
-        flag_cache.owner == player_id
+    local valid = self:_isValidStaffPlacement()
     self.anim:setFlag(valid and 0 or flag_altpal)
     local zoom = self.ui.zoom_factor
     if canvas:scale(zoom) then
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/place_staff.lua
```diff
@@ -75,17 +75,39 @@ function UIPlaceStaff:onCursorWorldPositionChange(x, y)
   self.tile_y = math_floor(y)
 end
 
+-- Check the given tile and the four around it for (the centre of) a reception desk.
+local function nearReceptionDesk(world, x, y)
+  return world:getObject(x, y, "reception_desk") or
```
lewri, line 80: Is there a better way of representing this? If not then that's fine too.

tobylane: It could be made generally useful by reading the footprint of objects, but if this is the only use for it then it would be harder to read.

Alberth289346: Just add a function that searches a specified area for an object? (It sounds like we may already something like that, but don't know.)

Alberth289346: There are also searches for the opposite part of the entrance door, maybe there are more

tobylane: Changed to using two world functions together.

CorsixTH/Lua/dialogs/place_staff.lua
```diff
@@ -77,15 +77,31 @@ end
 
 local flag_cache = {}
 local flag_altpal = 16
+-- Is this coordinate a valid tile for this staff member?
+local function isValidPlace(self)
```
Alberth289346, line 81: If you only read `function isValidPlace(self)`, this function can mean just about anything. `self` is quite literally any object that exists, and "valid place" applies to almost all objects too in one way or another.
In other words, the names are too generic, it can be interpreted too wide. The function name and the parameter don't give enough clues what use is intended.

Your comment says it's for a staff member, the `self` is thus easily replaced by a better name, eg `staff_member`,
`isValidPlace(staff_member)` is already much better.

Valid places for staff members are complicated, it could check the staff member for being in the right room, or inside the hospital, or not in the toilet, or a bunch of other notions of 'valid' for staff.
I don't quite know what the function is supposed to do, but since this is `place_staff.lua`, I am guessing something like `canBeDroppedHere(staff_member)` ?

(see how much difference a few well-picked names make?)

One thing that still bothers me is that 'dropping' or 'here' implies needing a position, and the function has no position parameter. I tried wording it in the function name, but I couldn't come up with a good solution for it (maybe you can).
Given my failure, I would describe that in the lua doc comments now.

CorsixTH/Lua/dialogs/place_staff.lua
```diff
@@ -77,15 +77,31 @@ end
...
+  self.world.map.th:getCellFlags(self.tile_x, self.tile_y, flag_cache)
+  -- Is the tile inside the player's hospital building?
+  if not (self.ui.hospital:getPlayerIndex() == flag_cache.owner and flag_cache.hospital) then
+    return false
+   end
```
Alberth289346, line 86: wrong indentation

CorsixTH/Lua/dialogs/place_staff.lua
tobylane: `-- Is this staff member permitted to be placed where the user is hovering or clicking?`

It describes the bigger picture rather than just what that function is there to judge? The user's actions set the coordinates.

Alberth289346: Yeah, I messed up big time in understanding what the function was doing. Sorry about that.

As for your question, if you write an explanation in comment, there is only one spot where you can read it. If you document it in the function name, it's everywhere for free. This is why a good function name has a lot of value, and a good parameter name helps near the definition, and in the code (and in the generated documentation if you ever read that).

You changed it to
``` lua
-- Is this staff member permitted to be placed where the user is hovering or clicking?
local function isValidStaffPlacement(staff_member)
```
Given my confusion I am surprised you changed the parameter name to my (what I now believe, wrong) suggestion. Please don't assume I am always right. In fact, if you aren't convinced by my arguments, don't change things. I may have played the game of code changing more than you, but that doesn't make me a gospel of truth :) 
It's your code change, you must aim for what you believe is the best possible solution.

As the parameter was the window and not the staff member, I now believe `self` was the right name after all.

The general idea of the comment is fine. The comment part `this staff member` feels a bit confusing to me. It suggests to me there is more than one in the class (there is this staff and there is thus also other staff). Phrasing it as "the staff member" doesn't trigger the 'this/other' link for me.

I have another question, why is this function not a method of the UI class, ie
``` lua
function UIPlaceStaff:_isValidStaffPlacement()
```
This makes it fully clear that the function is part of the ui class. The leading `_` means it's a method intended for internal used (a private method). It's not in the coding conventions it seems, but we do have code that uses it.

tobylane: With the function and parameter names both options work for me.

For this/that I mildly agree. It's the one given to the function.

tobylane:
Ready.

tobylane:
Minor fix and rebase. Ready.

PR #1806, opened 2021-01-23
[RDY] Fix updater

tobylane:
- Add the two github urls to the trusted domains. This fixes the updater
- Add a different open command for macOS, fixing it on that platform
- Add the requirement for the given download url to be https. This will break the ability to update, can you convert the url?
- When the user doesn't have luasec warn them

I'm not sure on the usefulness of the third and fourth commits. http://github.com is a redirect to https, so this only protects interference local to the user? If the package doesn't have luasec it's just an unhelpful nag? We may do better just asking packagers to add luasec if they've added luasockets to the dependencies.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1643,7 +1643,7 @@ function App:checkForUpdates()
   local current_version = self:getVersion()
 
   -- Only URLs that match this list of trusted domains will be accepted.
-  local trusted_domains = { 'corsixth.com', 'code.google.com' }
+  local trusted_domains = { 'corsixth.com', 'github.com', 'corsixth.github.io' }
 
   -- Only check for updates against released versions
   if current_version == "Trunk" then
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1639,7 +1639,7 @@ function App:checkForUpdates()
 
   -- Default language to use for the changelog if no localised version is available
   local default_language = "en"
-  local update_url = 'www.corsixth.com/check-for-updates'
+  local update_url = 'https://corsixth.github.io/CorsixTH/check-for-updates'
   local current_version = self:getVersion()
 
   -- Only URLs that match this list of trusted domains will be accepted.
@@ -1651,25 +1651,14 @@ function App:checkForUpdates()
     return
   end
 
-  local success, _ = pcall(require, "socket")
-
-  if not success then
-    -- LuaSocket is not available, just return
-    print("Cannot check for updates since LuaSocket is not available.")
+  local luasocket, _ = pcall(require, "socket")
+  local luasec, _ = pcall(require, "ssl.https")
+  if not (luasocket and luasec) then
+    print("Cannot check for updates since LuaSocket and/or LuaSec are not available.")
     return
-  else
-    self.lua_socket_available = true
   end
   local http = require("socket.http")
   local url = require("socket.url")
-
-  -- Safely attempt to use luasec
-  local hassec, _ = pcall(require, "ssl.https")
-  if hassec then
-    update_url = "https://" .. update_url
-  else
-    update_url = "http://" .. update_url
-  end
 
   print("Checking for CorsixTH updates...")
   local update_body, status, _ = http.request(update_url)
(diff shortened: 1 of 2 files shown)


lewri:
Regarding your fourth commit, if github is always https can't you just always force https?

tobylane:
Possibly later. I'm not sure what portion of users _could_ get their updates from within the game (currently none) rather than their package manager. Or whether this improves anything to the security of executing the remote code. That change would look like this,

```lua
  -- Require luasec
  local luasec, _ = pcall(require, "ssl.https")
  if not luasec then
    print("Will not check for updates since Luasec is not available.")
    return
  end
```

tobylane:
> Regarding your fourth commit, if github is always https can't you just always force https?

I misunderstood you. I don't necessarily want to force https because some users may lose out. The url in question at that point is corsixth.com/check-for-updates which works on http and https. https://tobylane.github.io/check-for-updates isn't as accessible because its https only.

lewri:
See what you mean about your securing download url to https...

```
return {
  major = 0,
  minor = 6,
  revision = 4,
  changelog_en = 'Lots of fixes and experimental ISO support awaits in the latest version.',
  download_url = 'http://github.com/CorsixTH/CorsixTH/releases',
}
```
If @thecycoone is unable to change could you try changing the http to https within lua and seeing if it returns status 200?

tobylane:
Removed the [third](https://github.com/CorsixTH/CorsixTH/pull/1806/commits/37a11bcf61f5f38c94ae32c85dc77c6d02a68ceb) and [fourth](https://github.com/CorsixTH/CorsixTH/pull/1806/commits/c0277eacc9584f13c30ca92d06627247883d5e79) commits and rebased. Ready.

tobylane:
This uses the file added in #1844 (it doesn't matter which way round these are accepted) for updating. The old url is left in for people who don't have luasec for https. I want to add a note to 0.65 packagers strongly recommending that if they include luasocket they add luasec, so we could deprecate checking on http-only.

TheCycoONE:
The non-sec path is probably not needed.

Nearly every distribution disables (or doesn't intentionally enable) update checks, because the package manager covers that functionality. It's really only useful for our own builds because those are the only ones available on the release pages in the links, and since they're our builds we can make sure to include luasec.

CorsixTH/Lua/app.lua
```diff
@@ -1638,12 +1638,11 @@ function App:checkForUpdates()
   self.check_for_updates = false
 
   -- Default language to use for the changelog if no localised version is available
-  local default_language = "en"
-  local update_url = 'www.corsixth.com/check-for-updates'
+  local default_language, update_url = "en"
```
TheCycoONE, line 1641: Why is update_url set to "en" here? You overwrite it later.

.github/workflows/Linux.yml
```diff
@@ -142,7 +142,7 @@ jobs:
           git fetch origin gh-pages
           git checkout --force gh-pages
           rsync --recursive build/doc/ .
-          git add animview/ corsixth_engine/ corsixth_lua/ index.html leveledit/
+          git add animview/ check-for-updates corsixth_engine/ corsixth_lua/ index.html leveledit/
```
TheCycoONE, line 145: Is check-for-updates here?

CorsixTH/Lua/dialogs/resizables/update.lua
```diff
@@ -109,8 +101,10 @@ end
 
 function UIUpdate:buttonDownload()
 
-  if self.os_is_windows then
+  if self.app.os == "windows" then
     os.execute("start " .. self.download_url)
+  elseif self.app.os == "macos" then
```
TheCycoONE, line 106: Nice

.github/workflows/Linux.yml
lewri: It is now: http://corsixth.github.io/CorsixTH/check-for-updates

tobylane:
Rewrote the last commit, the linux.yml change wasn't necessary.

TheCycoONE:
LGTM. I'll have to get luasec, preferably through vcpkg next time I worry about packaging.

tobylane:
Ready.

PR #1807, opened 2021-01-23
[RDY] Add advisor warnings for room queues

tobylane:
Previous feedback #1739

- Uses two strings to warn the user of long queues. I chose to define long as 6+, and to check monthly. Do we know what TH97 does?

### DIFF
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -180,6 +180,10 @@ function PlayerHospital:dailyAdviceChecks()
     end
   end
 
+  if day == 10 then
+    self:warnForLongQueues()
+  end
+
   -- Reset advise flags at the end of the month.
   if day == 28 then
     self.adviser_data.temperature_advice = false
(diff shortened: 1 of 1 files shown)


Alberth289346:
In general looks ok.

There are two things that worry me somewhat.

- If you have 1 room with a long queue, you get the same warning every month. Not sure that's an actual problem though.
- For large hospitals, having 6+ queue length may be normal.

If you want more precise advice, you could pick rooms of at least 6 length and 1.5 * average queue length or so. That gives the top 25% rooms at most (less if busier rooms don't reach the 1.5 * average, this happens when you have many rooms that are slightly busier). This should give you the real peak lengths only. Likely the 1.5 factor needs some tuning.

lewri:
You could also base queue length around how the VIP scores, which actually I think more or less lines up.

@mugmuggy may have some more ideas about TH97 specific triggers

https://github.com/CorsixTH/CorsixTH/blob/361b7faf7c4605fb860d6540c94498e4f95a1a27/CorsixTH/Lua/entities/humanoids/vip.lua#L404-L415

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -332,6 +334,27 @@ function PlayerHospital:msgKilled()
...
+  if #busy_rooms == 0 then return end
+  local chosen_room = busy_rooms[math.random(1, #busy_rooms)]
+  -- Required staff that is not nurse is doctor, researcher, surgeon or psych
+  if chosen_room.required_staff and not chosen_room.required_staff["Nurse"]
+      and math.random(1, 3) > 1 then
+    self.world.ui.adviser:say(
+        _A.warnings.queue_too_long_send_doctor:format(chosen_room.name))
+  else
+    self.world.ui.adviser:say(_A.warnings.queues_too_long)
+  end
```
lewri, line 355: Not sure we should guarantee an adviser pop-up every month from long queues if the player is constantly having long queues.

tobylane:
Using both your feedback, this only considers a queue long if it's 8 long, one short of a VIP considering it bad. This only warns for an above average queue. In my L12 hospital with ~40 queueing this is likely to give advice. Perhaps the advised-upon room could be stored to make sure there's not an exact duplicate month to month?

TheCycoONE:
Probably belongs in the dailyAdviceChecks on a specific date. It feels closer to the bench / drink warnings than the month end financial advice.

tobylane:
Done. Kept as a function, it's easier to read with returns.

Alberth289346:
That's good, it makes the `dailyAdviceChecks` dispatch function nicely short.

lewri:
> Perhaps the advised-upon room could be stored to make sure there's not an exact duplicate month to month?

You could probably add a flag to each room as default of e.g. has_warned that then has a function to clear them all at the end of year (you'd need an afterload for that).

Alternatively have an array or something inside player_hospital (?) that stores the warned room; which also clears at the end of year.

TheCycoONE:
Not a fan of lewri's suggestion and don't personally see a problem with warning about the same room each time if it's not fixed.

tobylane:
Ready.

mugmuggy:
Just as toby did ask, I had tried to find how these are used. I couldn't see the message specifying room name being used.
These were the ones I could see being used.
[11,8] "You need more Doctors. Try putting your best Doctors in the rooms with the longest queues."
[11,12] "Your queues are too long."

They are determined separately, the latter I believe takes just the average total waiting 'cycles', I don't know if they align to days, ticks/hours or some other value, the former the same logic but with some extra logic with some values yet to be determind. The average here is also a little off, due to the TH bug of starting the patient count at 1 to remove all these div by zero issues I assume.

tobylane:
Thanks!. 11 ,8 is _A.staff_advice.need_doctors. Is it better to strictly copy TH97, or use all strings that fit? Using all three strings looks like:

```lua
    local warn_msgs = {
      _A.warnings.queue_too_long_send_doctor:format(chosen_room.name)
      _A.staff_advice.need_doctors
    }
    self:giveAdvice(warn_msgs) 
```

tobylane:
I chose to use all three strings. Rebased and ready.

lewri (approved):

PR #1809, opened 2021-01-30
[rdy] Save settings from ingame menu when menu closes

### LINKED ISSUE #1311: Edge scrolling not saved in config when changed by user

tobylane:
*Previous feedback #1739, fixes #1311*

- Saves all menu bar setting changes when the dropdown menus go away (a second after the cursor goes off them)

### DIFF
--- a/CorsixTH/Lua/dialogs/menu.lua
+++ b/CorsixTH/Lua/dialogs/menu.lua
@@ -71,6 +71,7 @@ function UIMenuBar:onTick()
       self.menu_disappear_counter = nil
     else
       if self.menu_disappear_counter == 0 then
+        self.ui.app:saveConfig()
         self.menu_disappear_counter = nil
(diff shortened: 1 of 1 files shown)


lewri:
Assuming the gh-actions error is upstream, can you confirm Toby?

TheCycoONE:
Not related to his changes anyway.

tobylane:
Ready.

The actions error is the same type as the one before, it's downloading the wrong file for that release.

lewri (approved):

tobylane:
Rebased to pass tests.

PR #1812, opened 2021-02-20
Add cmake 3.5 to workflow

lewri:
Cmake version isn't defined in the LuaJIT and Lua 5.1 tests and now seems to be failing.
Attempting a fix by defining the same cmake version in all tests (3.5)

### DIFF
--- a/.github/workflows/Linux.yml
+++ b/.github/workflows/Linux.yml
@@ -16,14 +16,14 @@ jobs:
         include:
           - static_analysis: 0
             docs: 0
-            cmake:
+            cmake: 3.5
             lua: LuaJIT
             luac: luac5.1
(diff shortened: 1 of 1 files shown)


lewri:
Failure is now:
``  cmake: error while loading shared libraries: libidn.so.11: cannot open shared object file: No such file or directory
Error: Process completed with exit code 127. ``
(known about)

TheCycoONE:
I don't want cmake 3.5 defined for all of them, but perhaps latest wasn't the right thing either. It might make the most sense to require 3.5 in the lua 5.1 test (old on old), and cmake 3.19 on lua 5.4 / latest.

TheCycoONE:
Obsoleted by Toby's PR

PR #1813, opened 2021-02-20
Animview overflow

### LINKED ISSUE #1811: Shifting signed 32-bit value by 31 bits is undefined behaviour.
[See condition at line 94](https://github.com/CorsixTH/CorsixTH/blob/361b7faf7c4605fb860d6540c94498e4f95a1a27/AnimView/th.h#L94)
https://github.com/CorsixTH/CorsixTH/blob/361b7faf7c4605fb860d6540c94498e4f95a1a27/AnimView/th.h#L95   

 [See condition at line 101](https://github.com/CorsixTH/CorsixTH/blob/361b7faf7c4605fb860d6540c94498e4f95a1a27/AnimView/th.h#L101)
https://github.com/CorsixTH/CorsixTH/blob/361b7faf7c4605fb860d6540c94498e4f95a1a27/AnimView/th.h#L102 

[See condition at line 106](https://github.com/CorsixTH/CorsixTH/blob/361b7faf7c4605fb860d6540c94498e4f95a1a27/AnimView/th.h#L106)
https://github.com/CorsixTH/CorsixTH/blob/361b7faf7c4605fb860d6540c94498e4f95a1a27/AnimView/th.h#L107

TheCycoONE:
*Fixes #1811*

Resolves undefined behavior by shifting unsigned longs.

clang-tidy suggested removing the unused assignment to pTmp, and fixing the name _merge (underscore names are reserved). I moved to using anonymous namespaces for file local code, per the c++ guidelines / style in CorsixTH

### DIFF
--- a/AnimView/frmMain.cpp
+++ b/AnimView/frmMain.cpp
@@ -442,7 +442,7 @@ void frmMain::load() {
   m_oAnims.markDuplicates();
 
   m_txtAnimCount->SetValue(
-      wxString::Format(L"%u", (int)m_oAnims.getAnimationCount()));
+      wxString::Format(L"%zu", m_oAnims.getAnimationCount()));
 
   m_imgBackground.Create(400, 400);
   {
@@ -542,7 +542,7 @@ void frmMain::_onGhostIndexChange(wxSpinEvent& evt) {
 
 void frmMain::_onAnimChange(size_t iIndex) {
   m_iCurrentAnim = iIndex;
-  m_txtAnimIndex->ChangeValue(wxString::Format(L"%u", (int)iIndex));
+  m_txtAnimIndex->ChangeValue(wxString::Format(L"%zu", iIndex));
   m_iCurrentFrame = 0;
 
   THLayerMask oMask;
@@ -560,7 +560,7 @@ void frmMain::_onAnimChange(size_t iIndex) {
   m_panFrame->Refresh(false);
   m_txtFrameIndex->SetValue(wxString::Format(L"0"));
   m_txtFrameCount->SetValue(
-      wxString::Format(L"%u", (int)m_oAnims.getFrameCount(iIndex)));
+      wxString::Format(L"%zu", m_oAnims.getFrameCount(iIndex)));
 }
 
 void frmMain::_onPlayPause(wxCommandEvent& evt) {
@@ -579,7 +579,7 @@ void frmMain::_onPrevFrame(wxCommandEvent& evt) {
   else
     m_iCurrentFrame =
         (m_iCurrentFrame - 1) % m_oAnims.getFrameCount(m_iCurrentAnim);
-  m_txtFrameIndex->SetValue(wxString::Format(L"%u", m_iCurrentFrame));
+  m_txtFrameIndex->SetValue(wxString::Format(L"%zu", m_iCurrentFrame));
   m_panFrame->Refresh(false);
 }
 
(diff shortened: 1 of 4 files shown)


Alberth289346:
I have a rule for myself not to assign to variables inside expressions, like
``` cpp
pFiles->Add(pTmp = new wxButton(this, ID_NEXT, L"Next"), 0,
              wxALIGN_CENTER | wxALL, 1);
```
It may have been necessary a few decades ago to save some fraction of time, but even then I preferred readability above being fast. I trust compilers to pull these tricks for me if feasible.

Don't know how you feel about it, but it could be something to fix.

----
The other thing was the question whether `1ul` is the right solution, since mask is a `std::uint32_t` and these two types are not always the same. To me a `static_cast` to the type of `mask` would have been more logical/safe, but you even removed that.

C++ standard seems to say `ul` suffix is at least 32 bit wide, so your solution is correct too.


So all seems fine to me.

TheCycoONE:
Split out all the other assignment's within Add spots I saw because I agree it's a bit ugly.
I removed setting the background colour to the button colour entirely. It had no effect on my laptop (gtk3) and I can't imagine it being good anywhere?
I also fixed a bug with the format strings and size_t types that caused errors for me when picking frames with > and <.

AnimView/frmMain.cpp
```diff
@@ -314,61 +313,61 @@ frmMain::frmMain()
...
+  m_chkFrameFlags[6] = new wxCheckBox(this, wxID_ANY, L"2^6");
+  m_chkFrameFlags[7] = new wxCheckBox(this, wxID_ANY, L"2^7");
+  pFlags1->Add(m_chkFrameFlags[0], 0, wxEXPAND | wxALL, 2);
+  pFlags1->Add(m_chkFrameFlags[1], 0, wxEXPAND | wxALL, 2);
+  pFlags1->Add(m_chkFrameFlags[2], 0, wxEXPAND | wxALL, 2);
+  pFlags1->Add(m_chkFrameFlags[3], 0, wxEXPAND | wxALL, 2);
+  pFlags1->Add(m_chkFrameFlags[4], 0, wxEXPAND | wxALL, 2);
+  pFlags1->Add(m_chkFrameFlags[5], 0, wxEXPAND | wxALL, 2);
+  pFlags1->Add(m_chkFrameFlags[6], 0, wxEXPAND | wxALL, 2);
+  pFlags1->Add(m_chkFrameFlags[7], 0, wxEXPAND | wxALL, 2);
```
Alberth289346, line 339: These array assignments start to look like it could use a loop, although I can also live without it.

Alberth289346:
Changes look fine to me.
Hacked the cpp file somewhat as well, as I wanted to look at some animations, but the build was broken.
https://github.com/Alberth289346/CorsixTH/commit/c92abf97f2f0acc07d5d4f6dc5836e47944047f4

Lines 129 fixes a signed/unsigned comparison warning at line 130.
Then I started converting `unsigned char` to `std:uint8_t` and modifying casts.

The h file also fixes a comparison warning.

Feel free to apply or copy or redo what looks useful to you.

AnimView/frmMain.cpp
TheCycoONE: Yeah, I'll live with it unrolled, they aren't that long. Two sets of 8 flags / checkboxes.

Alberth289346:
And also updated, can't make a PR it seems.

https://github.com/CorsixTH/CorsixTH/compare/ac5c317c...Alberth289346:pr1813-warn_fixes-type_cleanup

PR #1814, opened 2021-02-20
Actions - fix cmake

tobylane:
- Download cmake 3.5 better
- Less duplication in the different jobs
- Adds a test for unwanted tabs, ignoring the language files

Output https://github.com/tobylane/CorsixTH/actions/runs/578805335

### DIFF
--- a/.github/workflows/Linux.yml
+++ b/.github/workflows/Linux.yml
@@ -16,31 +16,27 @@ jobs:
         include:
           - static_analysis: 0
             docs: 0
-            cmake:
+            cmake: 1
             lua: LuaJIT
             luac: luac5.1
-            packages: libluajit-5.1-dev luajit
-            cmakejit: '-DWITH_LUAJIT=ON -DLUA_LIBRARY=/usr/lib/x86_64-linux-gnu/libluajit-5.1.so'
+            animview: and AnimView
+            packages: libluajit-5.1-dev luajit libwxgtk3.0-gtk3-dev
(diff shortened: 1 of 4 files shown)


.github/workflows/Linux.yml
```diff
@@ -50,11 +46,14 @@ jobs:
...
         run: |
-          sudo apt-get install doxygen yamllint clang-tidy-9 libsdl2-dev \
-           libsdl2-mixer-dev libwxgtk3.0-gtk3-dev libavcodec-dev \
-           libavformat-dev libavutil-dev libswresample-dev libswscale-dev
+          sudo apt-get install libsdl2-dev libsdl2-mixer-dev \
+           libavcodec-dev libavformat-dev libavutil-dev libswresample-dev libswscale-dev
+      - name: Install static analysis requirements
+        if: matrix.static_analysis
+        run: |
+          sudo apt-get install doxygen yamllint clang-tidy-9
```
TheCycoONE, line 56: doxygen isn't for static analysis, it's for docs target.

.github/workflows/Linux.yml
```diff
@@ -16,31 +16,27 @@ jobs:
         include:
           - static_analysis: 0
             docs: 0
-            cmake:
+            cmake: 1
```
TheCycoONE, line 19: Can this be 3.5.0 instead of 1. Perhaps we inject this value into the download path? We always use cmake.

TheCycoONE (approved):
Good changes, a couple minor nits.

Issue #1816, opened 2021-02-21
[Bug] cmake fails to find or compile RncLib for Animview


Alberth289346:
#### Describe the issue

Cmake doesn't compile RncLib for Animview

Version
``` bash
$ cmake --version
cmake version 3.16.3
```

Build output
``` bash
$ mkdir build
$ cd build
$ cmake ..
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found wxWidgets: -L/usr/lib/x86_64-linux-gnu;-pthread;;;-lwx_gtk3u_xrc-3.0;-lwx_gtk3u_html-3.0;-lwx_gtk3u_qa-3.0;-lwx_gtk3u_adv-3.0;-lwx_gtk3u_core-3.0;-lwx_baseu_xml-3.0;-lwx_baseu_net-3.0;-lwx_baseu-3.0 (found version "3.0.4") 
  wxWidgets found
CMake Warning (dev) in CMakeLists.txt:
  No cmake_minimum_required command is present.  A line of code such as

    cmake_minimum_required(VERSION 3.16)

  should be added at the top of the file.  The version specified may be lower
  if you wish to support older CMake versions for this project.  For more
  information run "cmake --help-policy CMP0000".
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Configuring done
-- Generating done
-- Build files have been written to: ..../AnimView/build

$ make
Scanning dependencies of target AnimView
[ 40%] Building CXX object CMakeFiles/AnimView.dir/frmSprites.o
[ 40%] Building CXX object CMakeFiles/AnimView.dir/th.o
[ 80%] Building CXX object CMakeFiles/AnimView.dir/frmMain.o
[ 80%] Building CXX object CMakeFiles/AnimView.dir/app.o
[100%] Linking CXX executable AnimView
/usr/bin/ld: cannot find -lRncLib
[...]

PR #1817, opened 2021-02-23
[RFC] warn fixes type cleanup

TheCycoONE:
@Alberth289346 's PR. Probably needed my change merged before it could be created.

### DIFF
--- a/AnimView/th.cpp
+++ b/AnimView/th.cpp
@@ -38,7 +38,7 @@ SOFTWARE.
 
 namespace {
 
-constexpr unsigned char palette_upscale_map[0x40] = {
+constexpr uint8_t palette_upscale_map[0x40] = {
     0x00, 0x04, 0x08, 0x0C, 0x10, 0x14, 0x18, 0x1C, 0x20, 0x24, 0x28,
     0x2D, 0x31, 0x35, 0x39, 0x3D, 0x41, 0x45, 0x49, 0x4D, 0x51, 0x55,
     0x59, 0x5D, 0x61, 0x65, 0x69, 0x6D, 0x71, 0x75, 0x79, 0x7D, 0x82,
@@ -51,7 +51,7 @@ class ChunkRenderer {
  public:
   ChunkRenderer(int width, int height)
       : m_x(0), m_y(0), m_width(width), m_height(height), m_skip_eol(false) {
-    m_data = new unsigned char[width * height];
+    m_data = new uint8_t[width * height];
     m_ptr = m_data;
     m_end = m_data + width * height;
   }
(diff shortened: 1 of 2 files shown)


AnimView/th.cpp
```diff
@@ -179,10 +179,14 @@ void decode_chunks_complex(ChunkRenderer& renderer, const uint8_t* data,
   renderer.chunkFinish(transparent);
 }
 
+inline uint8_t average_colour(unsigned int channel_a, unsigned int channel_b) {
+    return (channel_a + channel_b) / 2;
```
TheCycoONE: This is a narrowing operation, let's cast.

TheCycoONE (commented):
Check that you didn't introduce any other implicit conversions. We 'try' to avoid them.

AnimView/th.cpp
Alberth289346: Arguably the function could be
``` cpp
inline uint8_t average_colour(uint8_t channel_a, uint8_t channel_b) {
    unsigned int ch_a = channel_a;
    unsigned int ch_c = channel_b;
    return (ch_a + chl_b) / 2;
}
```
to clarify that channels are 8 bit, but I fail to understand why casting the return expression does anything. Clearly the return type says we return 8 bits only?

Alberth289346:
> Check that you didn't introduce any other implicit conversions. We 'try' to avoid them.

- 7619a86 widens unsigned values to larger or equal unsigned values, and removes conversions and/or comparisons to signed values.
- 27a5656 uses `uint8_t` where `unsigned char` was used, no change at all in types.
- c64c7cc 
  -  `amt = (int)data[0];` widens to the same type as `amt`, so it doesn't do anything.
  - `Format(L"a%04ue.png", (int)iSprite)` narrowed `iSprite` type, which is removed in the patch
  - `unsigned long outlen` is changed to `size_t` which is the same thing
  - other changes keep the cast, but convert from old-style to new-style casting.

- 0cd789a4921e74e305c03002df64be19f33dc3d0 starts with 8 bit channel values and ends with 8 bit channel values, but the addition needs one additional bit. I didn't like the `dst.x = static_cast<uint8_t>((static_cast<unsigned int>(dst.x) + static_cast<unsigned int>(src.x)) / 2);` lines, so I split the computation with a new function.

AnimView/th.cpp
TheCycoONE: because implicit conversion is a warning, and narrowing conversions in particular should be pointed out explicitly because they can overflow (though this one can't). I do prefer that signature.

Alberth289346:
I'll just wait until cmake is fixed (see #1816 ), so I get a useful set of compile flags that tell me about implicit conversions.

lewri:
Albert has 1816 been resolved to continue, it looks like it in part?

Alberth289346:
The problem was me trying to build the library separately which is not how it's supposed to be used.

Trying to build the library as part of CTH build avoided the problem of #1816 afaik, but then (and today as well with current master), debug build still doesn't give warnings. (documentation build does give warning, and lua docs crashes it seems, but that's another topic.)

I had another try with this after finding out how to build the library and adding my own warning options, but it turned out that changes cascade all over the place, and run out of intended scope very quickly, so I am not much tempted to continue on this path,

Alberth289346:
Let's put it to bed.

PR #1818, opened 2021-02-26
The scope of the variable's can be reduced.

ZanyXDev:
The scope of the variable 'found' can be reduced.
The scope of the variable 'cont' can be reduced.

*Fixes #*

**
Reducing the scope of variables, in my opinion, will make the code more readable.
**
-
-
-

### DIFF
--- a/AnimView/frmMain.cpp
+++ b/AnimView/frmMain.cpp
@@ -759,8 +759,8 @@ void frmMain::_drawCoordinates(wxPaintDC& DC, int i, int j) {
 
 wxString frmMain::_getCaseSensitivePath(const wxString& sInsensitivePathPart,
                                         const wxString& sPath) {
-  bool found;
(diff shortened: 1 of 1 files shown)


AnimView/frmMain.cpp
```diff
@@ -759,8 +759,8 @@ void frmMain::_drawCoordinates(wxPaintDC& DC, int i, int j) {
 
 wxString frmMain::_getCaseSensitivePath(const wxString& sInsensitivePathPart,
                                         const wxString& sPath) {
-  bool found;
-  bool cont;
+  
+
```
Alberth289346, line 763: Just delete these lines, no point in keeping them

Alberth289346:
Also, the build fails as there is a line with trailing whitespace (see "Details" of the failing entry).
Please fix that as well.

PR #1820, opened 2021-03-01
The scope of the variable's can be reduced.

ZanyXDev:
The scope of the variable 'found' can be reduced.
The scope of the variable 'cont' can be reduced.

Fixes #

**
Reducing the scope of variables, in my opinion, will make the code more readable.
**

*Fixes #*

-
-
-

### DIFF
--- a/AnimView/frmMain.cpp
+++ b/AnimView/frmMain.cpp
@@ -759,9 +759,6 @@ void frmMain::_drawCoordinates(wxPaintDC& DC, int i, int j) {
 
 wxString frmMain::_getCaseSensitivePath(const wxString& sInsensitivePathPart,
                                         const wxString& sPath) {
-  bool found;
-  bool cont;
-
   if (!wxFileName::IsCaseSensitive()) {
     return sPath + sInsensitivePathPart;
   }
@@ -779,9 +776,9 @@ wxString frmMain::_getCaseSensitivePath(const wxString& sInsensitivePathPart,
     wxString pathPart = pathTokenizer.GetNextToken();
 
     wxString realName;
-    cont = dir.GetFirst(&realName, wxEmptyString,
+    bool cont = dir.GetFirst(&realName, wxEmptyString,
                         wxDIR_DIRS | wxDIR_FILES | wxDIR_HIDDEN | wxDIR_DOTDOT);
-    found = false;
+    bool found = false;
     while (cont) {
       if (realName.Upper() == pathPart.Upper()) {
         if (retStr.Last() != wxFileName::GetPathSeparator()) {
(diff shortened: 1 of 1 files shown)


ZanyXDev:
I'm upset, but the error message is too vague.

Alberth289346:
First of all, something for the next time: You can force-push a new version of your patch into your branch. Github will pick that up automatically and update the information in the pull request. That is, you don't have to close an existing PR and make a new one about the same topic.

As for the error message, yes it's vague.

What happens is that `clang-format` get run over the source code which formats all source code to some standard (don't know its setting, but it should be in the repository somewhere if you want to know).
If the output of the formatter is different from the input, some code formatting rule was broken by your change, and you should fix it.

The diff tells you how.
``` diff
diff --git a/AnimView/frmMain.cpp b/AnimView/frmMain.cpp
index 19da031..e08bf6a 100644
--- a/AnimView/frmMain.cpp
+++ b/AnimView/frmMain.cpp
@@ -776,8 +776,9 @@ wxString frmMain::_getCaseSensitivePath(const wxString& sInsensitivePathPart,
     wxString pathPart = pathTokenizer.GetNextToken();
 
     wxString realName;
-    bool cont = dir.GetFirst(&realName, wxEmptyString,
-                        wxDIR_DIRS | wxDIR_FILES | wxDIR_HIDDEN | wxDIR_DOTDOT);
+    bool cont =
+        dir.GetFirst(&realName, wxEmptyString,
+                     wxDIR_DIRS | wxDIR_FILES | wxDIR_HIDDEN | wxDIR_DOTDOT);
     bool found = false;
     while (cont) {
       if (realName.Upper() == pathPart.Upper()) {
```
The file is near the top, in this case `AnimView/frmMain.cpp`. The `-776,8` gives you the range of lines in your version. The lines with a "-" in front should be removed, and the lines with a "+" in front should be inserted to get the new version.

So what this is actually saying, is that your version
``` cpp
    bool cont = dir.GetFirst(&realName, wxEmptyString,
                        wxDIR_DIRS | wxDIR_FILES | wxDIR_HIDDEN | wxDIR_DOTDOT);
```
must be replaced by
``` cpp
    bool cont =
        dir.GetFirst(&realName, wxEmptyString,
                     wxDIR_DIRS | wxDIR_FILES | wxDIR_HIDDEN | wxDIR_DOTDOT);
```
where indentation must match exactly.

Alberth289346:
LGTM (looks good to me)

TheCycoONE:
For what it's worth, if you have clang-format installed locally you can run `clang-format -i myfile.cpp` to have it automatically format the code.

Many IDEs/editors have some sort of clang-format integration as well, which may be easier than hand tuning the file to match.

PR #1822, opened 2021-03-04
[RFC]Unify room discovery in hospital

lewri:
Been a while since I've done a PR and I already feel rusty. After discussing with @Alberth289346 about things in the Lua that could be made better this was one of the things that seemed pretty nice to do.

- hospital.discovered_rooms and hospital.undiscovered_rooms are unified under one hospital.room_discovery
- Introduced a new function to check for room discovery to save on repeating some code due to the new array.

I don't feel comfortable with how I'm using [i] for things, and sure there's a nicer looking way :)

### DIFF
--- a/CorsixTH/Lua/dialogs/build_room.lua
+++ b/CorsixTH/Lua/dialogs/build_room.lua
@@ -102,16 +102,18 @@ function UIBuildRoom:UIBuildRoom(ui)
   self:makeTooltip(_S.tooltip.build_room_window.cost, 160, 228, 282, 242)
 end
 
+--! Checks what rooms are now available to build
 function UIBuildRoom:updateBuildableRooms()
   local app = self.ui.app
   self.category_rooms = {}
   for i, category in ipairs({"diagnosis", "treatment", "clinics", "facilities"}) do
     local rooms = {}
     self.category_rooms[i] = rooms
-    for _, room in ipairs(app.world.available_rooms) do
+    for i, room_idx in ipairs(self.ui.hospital.room_discovery) do
+      local room = room_idx.room
       -- NB: Unimplemented rooms are hidden unless in debug mode
       if (app.config.debug or room.class) and room.categories[category] and
-          self.ui.hospital.discovered_rooms[room] then
+          self.ui.hospital.room_discovery[i].is_discovered then
         rooms[#rooms + 1] = room
       end
     end
--- a/CorsixTH/Lua/dialogs/build_room.lua
+++ b/CorsixTH/Lua/dialogs/build_room.lua
@@ -109,11 +109,11 @@ function UIBuildRoom:updateBuildableRooms()
   for i, category in ipairs({"diagnosis", "treatment", "clinics", "facilities"}) do
     local rooms = {}
     self.category_rooms[i] = rooms
-    for i, room_idx in ipairs(self.ui.hospital.room_discovery) do
+    for j, room_idx in ipairs(self.ui.hospital.room_discovery) do
       local room = room_idx.room
       -- NB: Unimplemented rooms are hidden unless in debug mode
       if (app.config.debug or room.class) and room.categories[category] and
-          self.ui.hospital.room_discovery[i].is_discovered then
+          self.ui.hospital.room_discovery[j].is_discovered then
         rooms[#rooms + 1] = room
       end
     end
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -167,15 +167,8 @@ function Hospital:Hospital(world, avail_rooms, name)
   self.policies = {}
   self.discovered_diseases = {} -- a list
 
-  self.discovered_rooms = {} -- a set; keys are the entries of TheApp.rooms, values are true or nil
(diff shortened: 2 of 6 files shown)


CorsixTH/Lua/hospital.lua
```diff
@@ -256,6 +249,18 @@ function Hospital:Hospital(world, avail_rooms, name)
   end
 end
 
+--! Checks if a room has been discovered
+--!param room_id (string) The name of the room
+--!return (boolean) true if found
+function Hospital:checkRoomDiscovery(room_id)
```
Alberth289346, line 255: I don't expect an answer with 'check', I expect that whatever you check ends up in a desired state.
When you check the oil in your car, you ensure that the oil level is (or becomes) good.

The common form for this kind of functions is to make it question-like, eg `isRoomDiscovered(room_id)`. It shows it's a query rather than an action, and it shows how to interpret the answer `true`, making it nicely readable in code like
``` lua
if hosp:isRoomDiscovered(room_id) then ...
```

CorsixTH/Lua/hospital.lua
```diff
@@ -167,15 +167,8 @@ function Hospital:Hospital(world, avail_rooms, name)
...
-  self.undiscovered_rooms = {} -- NB: These two together must form the list world.available_rooms
-  for _, avail_room in ipairs(avail_rooms) do
-    if avail_room.is_discovered then
-      self.discovered_rooms[avail_room.room] = true
-    else
-      self.undiscovered_rooms[avail_room.room] = true
-    end
-  end
+  -- Populate the level's rooms available to this hospital, including their discovery status
+  self.room_discovery = avail_rooms
```
Alberth289346, line 171: Document the structure of the variable, what fields are expected to exist?
For an example, see world.lua line 102 `self.next_earthquake`

CorsixTH/Lua/hospital.lua
```diff
@@ -256,6 +249,18 @@ function Hospital:Hospital(world, avail_rooms, name)
   end
 end
 
+--! Checks if a room has been discovered
+--!param room_id (string) The name of the room
+--!return (boolean) true if found
+function Hospital:checkRoomDiscovery(room_id)
+  for _, room_idx in ipairs(self.room_discovery) do
```
Alberth289346, line 256: The `_idx` suffix implies 'index', but the second returned value from `ipairs` is an element from the `self.room_discovery` array, which is not an index, if I may believe `world:getAvailableRooms()`.

CorsixTH/Lua/research_department.lua
```diff
@@ -505,39 +506,41 @@ function ResearchDepartment:discoverObject(object, automatic)
...
-    end
-    if discovery then
-      self.hospital.discovered_rooms[room] = true
-      self.hospital.undiscovered_rooms[room] = nil
-      if self.hospital == self.world.ui.hospital then
-        if automatic then
-          self.world.ui.adviser:say(_A.research.new_available:format(object.name))
-        else
-          self.world.ui.adviser:say(_A.research.new_machine_researched:format(object.name))
+  for i, room_id in ipairs(self.hospital.room_discovery) do
```
Alberth289346, line 509: It's also not a `room_id`. This one is more confusing, as the code already has `room_id`, with a very specific meaning, and you are not using the same meaning here I think.

CorsixTH/Lua/research_department.lua
```diff
@@ -505,39 +506,41 @@ function ResearchDepartment:discoverObject(object, automatic)
...
-        for research_object, progress in pairs(self.research_progress) do
-          if research_object.default_strength then
-            -- Don't improve those that already have the max strength
-            if progress.start_strength < max_strength then
-              if progress.discovered and progress.start_strength < min_strength then
-                self.research_policy["improvements"].current = research_object
-                min_strength = progress.start_strength
+      if discovery then
+        self.hospital.room_discovery[i].is_discovered = true
+        if self.hospital == self.world.ui.hospital then
```
Alberth289346, line 522: There is a standard check for this (`isLocalPlayerHospital` iirc), please use that

CorsixTH/Lua/dialogs/build_room.lua
```diff
@@ -102,16 +102,18 @@ function UIBuildRoom:UIBuildRoom(ui)
...
-    for _, room in ipairs(app.world.available_rooms) do
+    for j, room_idx in ipairs(self.ui.hospital.room_discovery) do
+      local room = room_idx.room
       -- NB: Unimplemented rooms are hidden unless in debug mode
       if (app.config.debug or room.class) and room.categories[category] and
-          self.ui.hospital.discovered_rooms[room] then
+          self.ui.hospital.room_discovery[j].is_discovered then
         rooms[#rooms + 1] = room
       end
     end
```
Alberth289346, line 119: I think the hospital should do this rather than the UI.

One thing to note here is that you don't introduce a shortcut variable to the long `self.ui.hospital.room_discovery` sequence. If you find repetitive use of the same thing, it's often useful to introduce a shorter notation for it.

lewri:
I think after speaking with Albert yesterday I'll kill off this PR and restart it now knowing that sets do more than I thought they did :D

lewri:
Will be superseded

PR #1823, opened 2021-03-06
[RDY] Unify room discovery

lewri:
Re-submission of #1822 
- hospital.discovered_rooms and hospital.undiscovered_rooms are unified under one hospital.room_discovery
- Introduced a new function to check for room discovery to save on repeating some code due to the new array.

Also cleaned up an if statement as requested in the Research Dept.


CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
```diff
@@ -173,7 +172,7 @@ function UICasebook:updateIcons()
       if #req.rooms > 0 then
         for _, room_id in ipairs(req.rooms) do
```
Alberth289346, line 173: If `#req.rooms == 0`, the `for` loop will not perform its body, and effectively just jump to the end as well.
As such, this `if` is of very little value, since the same check is also done inside `for`.

CorsixTH/Lua/hospital.lua
```diff
@@ -608,6 +615,28 @@ function Hospital:afterLoad(old, new)
...
+    -- Has the player discovered rooms since?
+    for room_new_id, room_new in pairs(self.room_discovery) do
+      for room_old, _ in pairs(self.discovered_rooms) do
+        if room_new_id == room_old.id then
+          room_new.is_discovered = true
+          break
+        end
+      end
+    end
+  end
```
Alberth289346, line 638: remove the `self.[un]discovered_rooms` variables.

CorsixTH/Lua/hospital.lua
```diff
@@ -2274,14 +2303,12 @@ end
...
+--!param disease (string): the disease to be checked.
 function Hospital:roomNotYetResearched(disease)
   local req = self:checkDiseaseRequirements(disease)
   if type(req) == "table" and #req.rooms > 0 then
     for _, room_id in ipairs(req.rooms) do
-      if not self.discovered_rooms[self.world.available_rooms[room_id]] then
-        return true
-      end
+      if not self:isRoomDiscovered(room_id) then return true end
     end
```
Alberth289346, line 2312: As a suggestion for a next patch, negated logic (true if something is not the case) is mostly horrible for reasoning, code is much easier to understand and reason about if a positive result yields true.

CorsixTH/Lua/research_department.lua
```diff
@@ -343,8 +343,8 @@ function ResearchDepartment:addResearchPoints(points, autopsy_room)
   ---------------------- An autopsy has been done ---------------------------
   if autopsy_room then
     -- Do something only if the room is among those not yet discovered.
-    for room, _ in pairs(self.hospital.undiscovered_rooms) do
-      if room.id == autopsy_room then
+    for room_id, room in pairs(self.hospital.room_discovery) do
```
Alberth289346, line 346: Your `room` is not a room, it's a table containing a room and a discovery status. Don't use well-known names from the code if it is something different, it puts people on the wrong foot.
Perhaps `room_disc_status` ?

A trick I often employ is to use a plural word for a container, eg `hospital.room_discoveries`, which immediately gives you a trivial solution for the name of an element from that container (that is, `room_discovery` or shorter `room_disc`).

Alberth289346: Perhaps `room_availability` ?

CorsixTH/Lua/research_department.lua
```diff
@@ -343,8 +343,8 @@ function ResearchDepartment:addResearchPoints(points, autopsy_room)
...
+      if room_id == autopsy_room and not room.is_discovered then
```
Alberth289346, line 347: Why do you iterate over the map if you eventually only allow the key `autopsy_room` ? (`autopsy_room` seems a quite wrong name, but that's a different matter)

CorsixTH/Lua/research_department.lua
```diff
@@ -505,39 +505,42 @@ function ResearchDepartment:discoverObject(object, automatic)
...
-      if self.hospital == self.world.ui.hospital then
-        if automatic then
-          self.world.ui.adviser:say(_A.research.new_available:format(object.name))
-        else
-          self.world.ui.adviser:say(_A.research.new_machine_researched:format(object.name))
+  local room_discovery = self.hospital.room_discovery
+  for room_id, _ in pairs(room_discovery) do
+    if not room_discovery[room_id].is_discovered then
+      local room = room_discovery[room_id].room
+      local discovery = true
```
Alberth289346, line 512: I know you just moved existing code, but `discovery` is highly generic, can you find a better name?

CorsixTH/Lua/hospital.lua
```diff
@@ -256,6 +256,13 @@ function Hospital:Hospital(world, avail_rooms, name)
   end
 end
 
+--! Checks if a room has been discovered
+--!param room_id (string) The name of the room
+--!return (boolean) true if found
+function Hospital:isRoomDiscovered(room_id)
+  return room_id and self.room_discovery[room_id].is_discovered
```
TheCycoONE, line 263: Why do we protect against nil / false room_id here (but not otherwise invalid room_ids)?

CorsixTH/Lua/hospital.lua
TheCycoONE: To be fair to lewri, this method already existed with that logic.

CorsixTH/Lua/research_department.lua
TheCycoONE: Yeah, this point is moot because the whole loop goes away and the condition below is just `if not self.hospital.room_discovery[autopsy_room].is_discovered then`

The logic in the loop depended on `room` which was different than the value lewri was using - so this is broken as is. Lewri, you need a way to get the room object from the room_id still.

CorsixTH/Lua/research_department.lua
```diff
@@ -343,8 +343,8 @@ function ResearchDepartment:addResearchPoints(points, autopsy_room)
   ---------------------- An autopsy has been done ---------------------------
```
TheCycoONE, line 343: An aside, this function badly needs to be split into two functions, one for autospy research and one for general research.

CorsixTH/Lua/hospital.lua
lewri: I'm not sure if you're saying we should remove the room_id check or whether we should check room_id is a string?

TheCycoONE: I wasn't saying to do anything, I was asking why their was a nil check / what sort of bad input we expect.

CorsixTH/Lua/dialogs/build_room.lua
```diff
@@ -102,16 +102,19 @@ function UIBuildRoom:UIBuildRoom(ui)
...
 
+--! Checks what rooms are now available to build
 function UIBuildRoom:updateBuildableRooms()
   local app = self.ui.app
   self.category_rooms = {}
   for i, category in ipairs({"diagnosis", "treatment", "clinics", "facilities"}) do
     local rooms = {}
     self.category_rooms[i] = rooms
-    for _, room in ipairs(app.world.available_rooms) do
+    local room_discoveries = self.ui.hospital.room_discoveries
```
Alberth289346, line 112: The right-hand side doesn't depend on `i`, `category`, `rooms`, or `self.category_rooms[i]`, this statement here does 4 times the same thing. Perhaps move it outside the loop?

CorsixTH/Lua/dialogs/build_room.lua
```diff
@@ -102,16 +102,19 @@ function UIBuildRoom:UIBuildRoom(ui)
...
+    for room_id, _ in pairs(room_discoveries) do
+      local room = room_discoveries[room_id].room
       -- NB: Unimplemented rooms are hidden unless in debug mode
       if (app.config.debug or room.class) and room.categories[category] and
-          self.ui.hospital.discovered_rooms[room] then
+          room_discoveries[room_id].is_discovered then
```
Alberth289346, line 117: The 2nd result of `pairs` at line 113 is the value of `room_discoveries[room_id]`. I am not sure why you discard that value at line 113, and then query it twice at lines 114 and 117

CorsixTH/Lua/dialogs/build_room.lua
```diff
@@ -102,16 +102,19 @@ function UIBuildRoom:UIBuildRoom(ui)
...
 function UIBuildRoom:updateBuildableRooms()
   local app = self.ui.app
   self.category_rooms = {}
   for i, category in ipairs({"diagnosis", "treatment", "clinics", "facilities"}) do
     local rooms = {}
     self.category_rooms[i] = rooms
-    for _, room in ipairs(app.world.available_rooms) do
+    local room_discoveries = self.ui.hospital.room_discoveries
+    for room_id, _ in pairs(room_discoveries) do
+      local room = room_discoveries[room_id].room
```
Alberth289346, line 114: You obtain the `room` here, even though you need it only at line 118 if lines 116 and 117 yield true.
Depending on how often that doesn't happen, you're getting `room`s for nothing.

Perhaps move the line after you decided it's going to be used? (and perhaps even delete it there by simplifying the code)

CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
```diff
@@ -170,16 +169,14 @@ function UICasebook:updateIcons()
...
-          end
-          -- Researched, but not built. TODO: maybe make this an else clause to not oversize the tooltip that much
-          build = (build and (build .. ", ") or " (") .. TheApp.rooms[room_id].name
+      for _, room_id in ipairs(req.rooms) do
+        -- Not researched yet?
+        if not hosp:isRoomDiscovered(room_id) then
+          known = false
+          research = (research and (research .. ", ") or " (") .. TheApp.rooms[room_id].name
         end
+        -- Researched, but not built. TODO: maybe make this an else clause to not oversize the tooltip that much
```
Alberth289346, line 178: Not your fault, and likely another patch for another day, but I don't see the logic of this conclusion.

CorsixTH/Lua/dialogs/build_room.lua
Alberth289346: If you fix the thing below, you may find there is only one use left for `room_discoveries`, at which point it may be beneficial to completely remove this variable, and copy  the right-hand side at that spot instead.

CorsixTH/Lua/research_department.lua
```diff
@@ -505,39 +497,42 @@ function ResearchDepartment:discoverObject(object, automatic)
...
-      if (not current_improvement_research or current_improvement_research.dummy) then
-        for research_object, progress in pairs(self.research_progress) do
-          if research_object.default_strength then
-            -- Don't improve those that already have the max strength
-            if progress.start_strength < max_strength then
-              if progress.discovered and progress.start_strength < min_strength then
-                self.research_policy["improvements"].current = research_object
-                min_strength = progress.start_strength
+      if unveil_room then
+        room_discoveries[room_id].is_discovered = true
```
Alberth289346, line 513: similar to `buildroom.lua`, line 117, why don't you use the `_` value from `pairs here? (and twice further up too)

CorsixTH/Lua/research_department.lua
```diff
@@ -326,79 +326,71 @@ function ResearchDepartment:getResearchRequired(thing)
...
-Otherwise they will be divided according to the research policy
-into the different research areas.
-!param points (integer) The total amount of points (before applying
-any level specific divisors to add to research.
-!param autopsy_room (string) If a specific room should get points following
-an autopsy, then this is the id of that room.
-]]
-function ResearchDepartment:addResearchPoints(points, autopsy_room)
-
+--! This function will add progress to discovering a room/object to treat a diagnosed patient after an autopsy.
```
Alberth289346, line 329: So you kill a patient first, then treat it?
Sounds a bit fishy :p

Perhaps something like "Give research points to the treatment of illness of the autopsied patient."?

CorsixTH/Lua/research_department.lua
```diff
@@ -326,79 +326,71 @@ function ResearchDepartment:getResearchRequired(thing)
...
+--! Research points are added as according to the level config.
+--!param target_room_id (string) The name of the room
```
Alberth289346, line 331: Something like "Id of the room that treats the illness, target for getting research points." ?
(yes, an id is also a name in this program), but that's generally not the case in other programs (IDs are usually numbers rather than strings).

Line 330 seems better fitting around line 342.  This text describes the call interface (ie what information to provide and what does it do and return as result). Points are not in that interface, so not highly essential to mention here.

CorsixTH/Lua/research_department.lua
```diff
@@ -326,79 +326,71 @@ function ResearchDepartment:getResearchRequired(thing)
...
-
-            -- Maybe we now have enough to discover the object?
-            if research.points > required then
-              self:discoverObject(TheApp.objects[object])
-            end
-            break
-          end
+  local room_discovery = self.hospital.room_discoveries[target_room_id]
+  -- Do something only if the room is among those not yet discovered.
+  if not room_discovery.is_discovered then
```
Alberth289346, line 336: If you invert that check and return, all code below can dedent 1 level.

Alberth289346:
LGTM

CorsixTH/Lua/app.lua
```diff
@@ -28,7 +28,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 152
```
TheCycoONE, line 31: Bump this to 154 and I will merge.

lewri: @TheCycoONE  bumped

TheCycoONE: You have to do the afterLoad too...

PR #1824, opened 2021-03-07
AnimView tidying

TheCycoONE:
Mainly switches from separate layout arguments and ord bitflags to wxSizerFlags, which are clearer, particularly with respect to borders.

Also make some narrowing / sign conversions explicit.

### DIFF
--- a/AnimView/frmMain.cpp
+++ b/AnimView/frmMain.cpp
@@ -27,6 +27,7 @@ SOFTWARE.
 #include <wx/bitmap.h>
 #include <wx/dcclient.h>
 #include <wx/dcmemory.h>
+#include <wx/defs.h>
 #include <wx/dir.h>
 #include <wx/dirdlg.h>
 #include <wx/filename.h>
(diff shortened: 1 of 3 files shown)


AnimView/frmMain.cpp
```diff
@@ -520,9 +531,10 @@ void frmMain::_onLastAnim(wxCommandEvent& evt) {
...
 
 void frmMain::_onAnimChar(wxCommandEvent& evt) {
-  long iAnim;
-  if (evt.GetString().ToLong(&iAnim)) {
-    if (iAnim >= 0 && iAnim < (long)m_oAnims.getAnimationCount()) {
+  long conv;
+  if (evt.GetString().ToLong(&conv)) {
+    size_t iAnim = static_cast<size_t>(conv);
+    if (iAnim >= 0 && iAnim < m_oAnims.getAnimationCount()) {
       _onAnimChange((size_t)iAnim);
```
Alberth289346, line 538: iAnim is already a `size_t`

That also makes the  `>= 0` check one line higher useless.

AnimView/frmMain.cpp
```diff
@@ -66,25 +67,40 @@ BEGIN_EVENT_TABLE(frmMain, wxFrame)
...
   wxSizer* pMainSizer = new wxBoxSizer(wxHORIZONTAL);
 
   wxSizer* pSidebarSizer = new wxBoxSizer(wxVERTICAL);
 
-#define def wxDefaultPosition, wxDefaultSize
+  wxSizerFlags fillSizerFlags(0);
+  fillSizerFlags.Expand().Border(wxALL, 0);
+
+  wxSizerFlags fillBorderSizerFlags(0);
+  fillSizerFlags.Expand().Border(wxALL, 1);
```
Alberth289346, line 86: Missing "Border" in the variable name?

AnimView/frmMain.cpp
```diff
@@ -95,232 +111,226 @@ frmMain::frmMain()
...
-  pSearchButtons->Add(new wxButton(this, ID_SEARCH_FRAME, L"Frame"), 0, wxALL,
-                      1);
-  pSearchButtons->Add(new wxButton(this, ID_SEARCH_SOUND, L"Sound"), 0, wxALL,
-                      1);
+  pSearchButtons->Add(new wxButton(this, ID_SEARCH_LAYER_ID, L"Layer/ID"),
+                      buttonSizerFlags);
+  pSearchButtons->Add(new wxButton(this, ID_SEARCH_FRAME, L"Frame"),
+                      buttonSizerFlags);
+  pSearchButtons->Add(new wxButton(this, ID_SEARCH_SOUND, L"Sound"),
+                      buttonSizerFlags);
```
Alberth289346, line 322: Sure these buttonSizer things are right, it's missing a `wxALIGN_CENTER_VERTICAL`

Alberth289346: ?

AnimView/frmMain.cpp
TheCycoONE: Yep. Sloppy. I have to range check then cast

AnimView/frmMain.cpp
TheCycoONE: Looks good to me. I thought it was better to be consistent with our buttons.

AnimView/frmMain.cpp
TheCycoONE: Alright. I bet that changed appearances. Didn't notice from a sight inspection.

AnimView/frmMain.cpp
Alberth289346: Likely a good idea, I just verified whether the flags were the same before and after the change, and these failed that test :)

AnimView/frmMain.cpp
```diff
@@ -520,11 +531,13 @@ void frmMain::_onLastAnim(wxCommandEvent& evt) {
...
-  if (evt.GetString().ToLong(&iAnim)) {
-    if (iAnim >= 0 && iAnim < (long)m_oAnims.getAnimationCount()) {
-      _onAnimChange((size_t)iAnim);
+  long conv;
+  if (evt.GetString().ToLong(&conv)) {
+    if (conv < 0 || conv >= m_oAnims.getAnimationCount()) {
+      return;
     }
+    size_t iAnim = static_cast<size_t>(conv);
+    _onAnimChange((size_t)iAnim);
```
Alberth289346, line 540: Still doing a cast to its own type.

Alberth289346:
Looks fine to me

PR #1830, opened 2021-03-12
[RFC/RDY] Dedication to the IRC

lewri:
Just some fun additions to staff description.

### DIFF
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -75,8 +75,9 @@ staff_descriptions.bad[14] = "Sly, cunning, and subversive. "
 staff_descriptions.misc[11] = "Distils whisky. "
 
 -- Some fun, in honor of the IRC :)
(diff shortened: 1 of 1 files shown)


lewri:
Working on making this code friendly :(

lewri:
Not easily fixable right now to be properly keyed. But in all ways we init the language makes a mismatched key number not matter. Have discussed with @Alberth289346 and it will need to be made better at some point but there are several hurdles to overcome before that.

TheCycoONE:
Well this did encourage me to take a deeper look at how __random is implemented in `th_lua_strings.cpp`. It's an 'interesting' implementation. I may add a unit test and re-implement it with fewer C<->Lua transitions.

lewri:
Just to add a note so I don't keep forgetting -- I believe Corsix's original workaround for dedications was putting it inside the original strings lua file. But I don't think it'd work in this case.

lewri:
Closing until there's a better implementation for this

PR #1831, opened 2021-03-13
[RDY] Select insurance companies from all available company names

tobylane:
Previous feedback #1722 

- Picks three insurance companies, keeping the preference for early ones, from the full range in the language file.

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -183,18 +183,22 @@ function Hospital:Hospital(world, avail_rooms, name)
   self.policies["stop_procedure"] = 1 -- Note that this is between 1 and 2 ( = 100% - 200%)
   self.policies["goto_staffroom"] = 0.6
   self.policies["grant_wage_increase"] = TheApp.config.grant_wage_increase
-  -- Randomly select three insurance companies to use, only different by name right now.
-  -- The first ones are more likely to come
+
+  -- Semi-randomly select three insurance companies to use, only different by name right now.
+  -- The companies in the first quarter of the list are more likely to be selected
   self.insurance = {}
+  -- Make a local writeable copy table of the translated company names.
+  local companies = {}
   for no, local_name in ipairs(_S.insurance_companies) do
-    -- NOTE: Will not work if more companies are added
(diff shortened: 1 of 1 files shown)


tobylane:
Ready.

CorsixTH/Lua/hospital.lua
```diff
@@ -183,18 +183,24 @@ function Hospital:Hospital(world, avail_rooms, name)
   self.policies["stop_procedure"] = 1 -- Note that this is between 1 and 2 ( = 100% - 200%)
   self.policies["goto_staffroom"] = 0.6
   self.policies["grant_wage_increase"] = TheApp.config.grant_wage_increase
+
   -- Randomly select three insurance companies to use, only different by name right now.
-  -- The first ones are more likely to come
+  -- The companies in the first quarter of the list are more likely to be selected
```
lewri, line 188: I feel like these two comments go against each other:
1. Insurance companies are random
2. Insurance companies are shifted towards the start of the list.
Unless this was original behaviour we should instead try and make the selection more random. Though it could be out of scope of this PR.

CorsixTH/Lua/hospital.lua
```diff
@@ -183,18 +183,24 @@ function Hospital:Hospital(world, avail_rooms, name)
...
+    companies[no] = local_name
   end
+  repeat
+    local num = math.random(1, 2) == 1 and math.random(1, math.ceil(#companies / 4)) or
+        math.random(1, #companies)
+    if companies[num] then
+      self.insurance[#self.insurance + 1] = companies[num]
+      table.remove(companies, num)
+    end
+  until #self.insurance == 3 or #companies < 1
```
lewri, line 202: Is there any consequence of using a repeat loop where #companies is actually <1? I wonder if a while loop may be better to protect against that. Actually if #companies is <1, line 197 will throw an error as arg 1 > arg 2.

lewri: (That line above might only actually apply if we started with companies<1 ... if it naturally diminishes that's fine)

tobylane:
I agree on both, the intent is clearer. Ready.

PR #1832, opened 2021-03-13
[RDY] Disable/close jukebox when audio is disabled

tobylane:
- User can't open the jukebox if audio is disabled
- Will close jukebox if it was left open in a save where audio was enabled

### DIFF
--- a/CorsixTH/Lua/dialogs/jukebox.lua
+++ b/CorsixTH/Lua/dialogs/jukebox.lua
@@ -176,3 +176,10 @@ function UIJukebox:draw(canvas, x, y)
     end
   end
 end
+
+function UIJukebox:afterLoad(old, new)
+  if not (self.app.config.audio and self.app:isAudioEnabled()) then
+    self:close()
+  end
+  Window.afterLoad(self, old, new)
+end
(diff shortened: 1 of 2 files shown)


lewri:
My only comment would be should the afterLoad function be standardised? Even if we don't use the ``old`` and ``new`` elements.

TheCycoONE:
We don't really want the jukebox in the save at all (see mugmuggy's PR.) Will take this though.

tobylane:
Ready.

CorsixTH/Lua/dialogs/jukebox.lua
```diff
@@ -176,3 +176,9 @@ function UIJukebox:draw(canvas, x, y)
     end
   end
 end
+
+function UIJukebox:afterLoad()
+  if not (self.app.config.audio and self.app:isAudioEnabled()) then
+    self:close()
+  end
+end
```
Alberth289346, line 184: Apparently you need to call the base-class too, see eg `dialogs/adviser.lua`. You also need to pass the `old` and `new` values.

Not sure if it needs an `old` test like the other `afterLoad` function, I guess it won't hurt much not to.

tobylane:
Added base class call. I think this shouldn't check for old, if you made a save in savegame version 160 with audio on and jukebox open then later loaded it with audio off, this should run.

lewri (approved):

PR #1833, opened 2021-03-13
Put system information at the start of gamelog.txt

### LINKED ISSUE #1633: Suggestion: Move system info from title bar to gamelog.txt
I believe from discussion, the SDL, Lua version, Arch, and Direct 3D version are reported after CorsixTH in the game window's title for debugging purposes.

In reality when we want this info of this level for issues, we would ask for it in the form of the gamelog.txt
To improve the aesthetic of the window I would suggest we move this info into the gamelog. We could also probably stretch it out a bit to be a bit more informative about the data that matters.

The end result being the window title just becomes "CorisxTH" only; which is often the format other games present their game window.

tobylane:
*Previous feedback #1722*

- Move the game log from world to app
- Log all available info to the game log, fix #1633, example

> Compiled with api_version: 2679, arch_64: true, os: macos, audio: false, 
Renderer: SDL, metal
Lua 5.4 run with game version: Trunk, savegame version: 152, api version: 2679

### DIFF
--- a/CorsixTH/CMakeLists.txt
+++ b/CorsixTH/CMakeLists.txt
@@ -55,6 +55,13 @@ else()
   set(CORSIX_TH_OS "unix")
 endif()
 
+# Report system architecture in compile_opts.arch
+if(MSVC)
+  set(CORSIX_TH_ARCH ${CMAKE_CXX_COMPILER_ARCHITECTURE_ID})
+else()
+  set(CORSIX_TH_ARCH ${CMAKE_SYSTEM_PROCESSOR})
+endif()
+
 # Ensure config.h is picked up by cmake - moving this into subdir cmake files will
 # prevent it applying to the CorsixTH project
 set(CorsixTH_generated_src_dir ${CMAKE_BINARY_DIR}/CorsixTH/Src/)
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -166,18 +166,7 @@ function App:init()
   SDL.wm.setIconWin32()
 
   self:setCaptureMouse()
-
-  local caption_descs = {self.video:getRendererDetails()}
-  if compile_opts.jit then
-    caption_descs[#caption_descs + 1] = compile_opts.jit
-  else
-    caption_descs[#caption_descs + 1] = _VERSION
-  end
-  if compile_opts.arch_64 then
-    caption_descs[#caption_descs + 1] = "64 bit"
-  end
-  self.caption = "CorsixTH (" .. table.concat(caption_descs, ", ") .. ")"
-  self.video:setCaption(self.caption)
+  self.caption = "CorsixTH"
 
   -- Prereq 2: Load and initialise the graphics subsystem
   corsixth.require("persistance")
@@ -1748,6 +1737,21 @@ end
 
 function App:isAudioEnabled()
   return TH.GetCompileOptions().audio
+end
+
+function App:getSystemInfo()
+  local compile_opts = TH.GetCompileOptions()
+  local comp_details = {}
+  for key, value in pairs(compile_opts) do
+    table.insert(comp_details, key .. ": " .. tostring(value))
+  end
+  table.sort(comp_details)
+  local compiled = string.format("Compiled with %s\nSDL renderer: %s\n",
+      table.concat(comp_details, ", "), self.video:getRendererDetails())
+  local running = string.format("%s run with api version: %s, game version: %s, savegame version: %s\n",
+      compile_opts.jit or _VERSION, tostring(corsixth.require("api_version")),
+      self:getVersion(), tostring(SAVEGAME_VERSION))
+  return(compiled .. running)
 end
 
 -- Do not remove, for savegame compatibility < r1891
--- a/CorsixTH/Src/th_lua.cpp
+++ b/CorsixTH/Src/th_lua.cpp
@@ -185,15 +185,9 @@ int l_get_compile_options(lua_State* L) {
   lua_settop(L, 0);
   lua_newtable(L);
 
-#ifdef CORSIX_TH_64BIT
-  lua_pushboolean(L, 1);
-#else
-  lua_pushboolean(L, 0);
-#endif
-  lua_setfield(L, -2, "arch_64");
-
-  lua_pushliteral(L, "SDL");
-  lua_setfield(L, -2, "renderer");
+  // Report architecture
+  lua_pushliteral(L, CORSIX_TH_ARCH);
+  lua_setfield(L, -2, "arch");
 
 #ifdef CORSIX_TH_USE_SDL_MIXER
   lua_pushboolean(L, 1);
--- a/.github/workflows/Linux.yml
+++ b/.github/workflows/Linux.yml
@@ -67,11 +67,11 @@ jobs:
           cd ../../
           rm -rf Catch2
 
-      - name: Install CMake 3.5
+      - name: Install CMake 3.10
         if: matrix.cmake
         run: |
-          # Install CMake version 3.5, the oldest CorsixTH-supported version, which does not support Lua 5.4
-          curl -L https://github.com/Kitware/CMake/releases/download/v3.5.0/cmake-3.5.0-Linux-i386.sh -o cmake.sh
+          # Install CMake version 3.10, the oldest CorsixTH-supported version, which does not support Lua 5.4
+          curl -L https://github.com/Kitware/CMake/releases/download/v3.10.3/cmake-3.10.3-Linux-x86_64.sh -o cmake.sh
           sudo bash cmake.sh --prefix=/usr/local/ --exclude-subdir --skip-license
           cmake --version
       - name: Create CorsixTH ${{ matrix.animview }} makefiles with ${{ matrix.lua }}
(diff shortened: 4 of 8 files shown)


CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -494,10 +494,10 @@ local function Humanoid_startAction(self)
     end
 
     -- Tell the player what just happened.
-    self.world:gameLog("")
-    self.world:gameLog("Empty action queue!")
-    self.world:gameLog("Last action: " .. self.previous_action.name)
-    self.world:gameLog(debug.traceback())
+    self.world.app:gameLog("")
```
TheCycoONE, line 497: FWIW TheApp is a global, you don't need to go through world.

tobylane:
Ready, unless you want that TheApp change.

lewri:
Would personally recommend using ``TheApp`` but not up to me

TheCycoONE:
Yeah, let's use TheApp for this. Nice and uniform. Then I can pretend it's a service for dependency injection and not a circular dependency.

tobylane:
Done.

TheCycoONE:
Play tested this a little. It appears to have lost the old gamelog data when I continue a game.
Here is a dump of the gamelog from my save from master:
```
Created game with savegame version 139.
Savegame version is 139 (Trunk), originally it was 139 (Trunk)
Savegame changed from 139 (Trunk) to 155 (Trunk). The save was created using 139 (Trunk).
Savegame version is 155 (Trunk), originally it was 139 (Trunk).
```

And here is the same game dumped from this PR.
```
Compiled with os: unix, arch_64: true, audio: true, api_version: 2679, 
Renderer: SDL, opengl
Lua 5.4 run with game version: Trunk, savegame version: 155, api version: 2679
Savegame version is 155 (Trunk), originally it was 139 (Trunk).
```

tobylane:
I have a fix for that, but it's making me wonder if the system info should be put into every game log, and keep game log in World.

TheCycoONE:
I was thinking the new system info should maybe just be added as a header when it gets dumped.  Open to other thoughts.

TheCycoONE:
https://github.com/CorsixTH/CorsixTH/blob/edeba9fc767c4d011bb24ad48867694a1b09d6ab/CorsixTH/Lua/persistance.lua#L235 around here is where you would save/load a new top level object.

tobylane:
Ready.

TheCycoONE:
On my system this was the output:
```
Compiled with os: unix, renderer: SDL, opengl, arch_64: true, api_version: 2679, audio: true
Lua 5.4 run with game version: Trunk, savegame version: 155, api version: 2679
Created game with savegame version 155.
```

lewri:
> arch_64: true

Why not just ``64-bit``? Though I would guess we also have ARM now... That should probably have its own identifier.

tobylane:
That sounds good, I'll do that in another PR.

TheCycoONE:
Can we avoid using the same separater that's used for renderer, or quote them.

TheCycoONE:
also what's the purpose if having api_version listed twice.

tobylane:
Replaced arch_64 with arch, outputs are x86, x64, ARM, ARM64 and Other. Any others worth including?

API version is from the C++ and Lua sides.

Removed the SDL part as it's unchanging, so this reports the SDL renderer.

> Compiled with arch: x64, audio: true, api_version: 2681, os: macos
SDL renderer: metal
Lua 5.4 run with game version: Trunk, savegame version: 158, api version: 2681

TheCycoONE:
Ok, spiffy - thanks

TheCycoONE:
Crashes on launch with MSVC:
```
An error has occurred in CorsixTH:
attempt to index a function value
stack traceback:
        [C]: in function 'TH.GetCompileOptions'
        E:/GitHub/CorsixTH/CorsixTH/Lua\app.lua:127: in method 'init'
        E:/GitHub/CorsixTH/CorsixTH/CorsixTH.lua:143: in main chunk
        [C]: in ?
```

CorsixTH/Src/th_lua.cpp
```diff
@@ -185,15 +185,23 @@ int l_get_compile_options(lua_State* L) {
   lua_settop(L, 0);
   lua_newtable(L);
 
+#ifdef __x86_64__
```
TheCycoONE, line 188: These re non-standard headers, they are not defined in MSVC

CorsixTH/Src/th_lua.cpp
```diff
@@ -185,15 +185,23 @@ int l_get_compile_options(lua_State* L) {
...
 
-  lua_pushliteral(L, "SDL");
-  lua_setfield(L, -2, "renderer");
+#elif __arm__
+#ifdef CORSIX_TH_64BIT
+  lua_pushliteral(L, "ARM64");
+#else
+  lua_pushliteral(L, "ARM");
+#endif
+
```
TheCycoONE, line 201: You are missing an else here, so Other isn't pushed when neither __x86_64__ nor __arm__ are defined.

TheCycoONE (changes requested):
You'll need another approach to getting the compile host, I would suggest bringing it in from CMake instead of trying to use non-standard compiler headers.

https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html as a start.

tobylane:
Done.

> Compiled with api_version: 2681, audio: true, os: macos, arch: x86_64
SDL renderer: metal
Lua 5.4 run with game version: Trunk, savegame version: 158, api version: 2681

TheCycoONE:
Looks good on my linux chromebook:
```
Compiled with os: unix, api_version: 2681, arch: aarch64, audio: true
SDL renderer: opengl
Lua 5.4 run with game version: Trunk, savegame version: 158, api version: 2681
```
Just want to test 32 and 64 bit windows builds before I merge.

TheCycoONE:
Heh, I see we're seeing the non-deterministic order of pairs - our header orders don't match.

CorsixTH/Lua/app.lua
```diff
@@ -1750,6 +1739,20 @@ function App:isAudioEnabled()
   return TH.GetCompileOptions().audio
 end
 
+function App:getSystemInfo()
+  local compile_opts = TH.GetCompileOptions()
+  local comp_details = ""
+  for key, value in pairs(compile_opts) do
```
TheCycoONE, line 1745: If we want a consistent order we'd need to build a list of keys and sort it (or hardcode it, but that's not as nice is it?)

Alberth289346: Python2 also had non-deterministic dictionaries, and settled in Python3 for deterministic insertion order. Not sorted, but faster.

Simplest long term solution would be to change the lua table itself, but that may be out of scope, as you don't have a standard lua implementation any more in that case.

Sorting will work for tables with sortable keys, but likely not everything is fully ordered.

tobylane:
The first line is sorted, the last line adjusted to match.

>Compiled with api_version: 2681, arch: x86_64, audio: true, os: macos
SDL renderer: metal
Lua 5.4 run with api version: 2681, game version: Trunk, savegame version: 158

CorsixTH/Lua/app.lua
TheCycoONE: I'm talking about a particular case where the UI (or gamelog header in this case) depended on the order of the pairs. That's a broader discussion - factorio for one modified lua's table to have a deterministic order, but that's likely out of scope for us. We can hack it with lists of keys where we need to (perhaps encapsulated)

tobylane:
Ready.

TheCycoONE:
Yes I know. I need to verify whether this works on Windows the way we package. There is a cmake issue about it, but I'm not sure if it applies the way I generate.

lewri:
Here's mi windows portable from Appveyor's output if it helps:

```
Compiled with api_version: 2681, arch: AMD64, audio: true, os: windows
SDL renderer: direct3d
Lua 5.4 run with api version: 2681, game version: Trunk, savegame version: 158
```

TheCycoONE:
This is the output from a Win32 (x86) build for Windows as we package it:

```
Compiled with api_version: 2681, arch: AMD64, audio: true, os: windows
SDL renderer: direct3d
Lua 5.4 run with api version: 2681, game version: Trunk, savegame version: 158
Created game with savegame version 158.
```

So yes, it seems to be broken there 😞

TheCycoONE:
This is the upstream bug: https://gitlab.kitware.com/cmake/cmake/-/issues/15170

tobylane:
We could keep arch_64 until that bug is fixed?

TheCycoONE:
I tested this block and it seems to work (CMake documentation warns that it's internal - but...)
```
# Report system architecture in compile_opts.arch
if(MSVC)
  set(CORSIX_TH_ARCH ${CMAKE_CXX_COMPILER_ARCHITECTURE_ID})
else()
  set(CORSIX_TH_ARCH ${CMAKE_SYSTEM_PROCESSOR})
endif()
```

CorsixTH/CMakeLists.txt
```diff
@@ -55,6 +55,13 @@ else()
   set(CORSIX_TH_OS "unix")
 endif()
 
+# Report system architecture in compile_opts.arch
+if(MSVC)
+  set(CORSIX_TH_ARCH ${CMAKE_CXX_COMPILER_ARCHITECTURE_ID})
```
TheCycoONE, line 60: On discord I checked / pointed out that CMAKE_CXX_COMPILER_ARCHITECTURE_ID is CMake 3.10+ only. I don't mind bumping the cmake_minimum_version, or version guarding this condition.

PR #1836, opened 2021-03-14
[RDY] Fix Lua Console

lewri:
Clears the ``strict_declare_global`` of ``_`` whenever the Lua console closes so it does not kill saves.

For some reason the save seems to try to preserve whatever value _ was set to when using the console -- and when you reopen the game it no longer exists.

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/file_browsers/save_game.lua
+++ b/CorsixTH/Lua/dialogs/resizables/file_browsers/save_game.lua
@@ -82,7 +82,7 @@ end
 --! Try to save the game with given filename; if already exists, create confirmation window first.
 function UISaveGame:trySave(filename)
   -- Saving without closing the Lua Console will break the save
-  if self.ui:getWindow(UILuaConsole) then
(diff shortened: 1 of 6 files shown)


CorsixTH/Lua/dialogs/resizables/file_browsers/save_game.lua
```diff
@@ -81,6 +81,10 @@ end
 
 --! Try to save the game with given filename; if already exists, create confirmation window first.
 function UISaveGame:trySave(filename)
+  -- Saving without closing the Lua Console will break the save
+  if self.ui:getWindow(UILuaConsole) then
```
TheCycoONE, line 85: How does this react to autosave?

lewri:
Updated for autosavess

lewri:
NB: Needs to wait until #1833 is merged before this one goes. I've prepped for the change of gamelog's location.

lewri:
Killing in favour of a less restrictive solution

Issue #422, opened 2014-12-10
Emergencies stopped - without reason (?)


Glimmer-Man:
I tried to set emergencies at semi-random, with this values: days: 30, variance 20. I made it in level editor. I was expected a lot of emegencies and they came. But after time, emergecies stops and no one comes again.

I told myself "maybe semi-random is limited" and put it away...  and set back random mode.

After some time, I was edited world.lua, function "World:nextEmergency()", and set:

local mean = control[0].Mean or 30
local variance = control[0].Variance or 10

and looking for any reason, why emergencies could stop at random mode. I see no reason, but my skills are limited. So maybe there is a reason...

So I played, and after some time emergencies stops again. I found it weird.

I think about that and I have an idea - with this setting ( 30 - 10) is possible that next emergency could be set on same month as previous one ended. And I think this is the source of error. But I didn't find it in code... (not skilled with lua).

So I tried to set new values: days - 45, variance - 10 (always in next mont). In game I've returned back before last emergency ends (using autosaves). And it seems OK. Emergencies don't stop and appears according to this settings. 

Maybe I missed something in manual - sorry about that, but nevertheless I deside to report this finding.

bobcanthelpyou:
@Glimmer-Man could you share your maps with different emergency values for testing?

mugmuggy:
The Mean-'Variance' values where the issue stems from, is probably from negative values and thus emergencies will never start again once the start date is set back in time.

The documentation does refer to the 'variance' on a normal distribution - https://github.com/CorsixTH/CorsixTH/wiki/Custom-Levels#random-emergencies - but then alludes to the range relating to that figure as more than likely being with the 'variance' days of the mean.

However, the 'variance' is actually the standard deviation in the Box-Muller transform (math.n_random), thus in the original example, 99.7% would be in the range of 30 +- (3*20) resulting in those negative numbers. 68% would be in the range when interpretting the documentation.

Suggest we close this issue then for an update to that documentation instead.

lewri:
Recommended action as above

PR #1843, opened 2021-03-21
[RDY] Rework Emergencies

### LINKED ISSUE #1839: [Bug] Overlapping emergencies stop emergencies
#### Describe the issue
If next_emergency is set to a date that has already passed the game will be stuck in a state where it can no longer take emergencies. I don't have a savegame for this right now but am developing the fix already.

#### Expected Behavior
Handle emergency generation when the emergency date has passed (see below).

#### System Information
**CorsixTH Version:** *0.64*

**For discussion (if any)**
Fix 1: Skip the missed emergency and go to the next one. (So far coding for this version).
Fix 2: Postpone the missed emergency so it happens.

lewri:
*Fixes #1839*

- nextEmergency code will now more decisively schedule the next emergency
- New functions separate the random emergency scheduling from controlled emergencies
- Should an emergency end up scheduled in the past, the code will attempt the next emergency
- A complementary afterLoad has been added to try and fix any previous saves where emergencies got "stuck" because they were scheduled in the past. There isn't a guarantee this will work and has a small risk of skipping an emergency.

Ideally I need to make a separate PR to handle ``cheatMonth()`` and ``cheatYear()`` as these can break emergencies (and other things).

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1295,6 +1295,7 @@ function World:nextEmergency()
       self.next_emergency = nil
       return
     end
+    self.next_emergency = emergency
     self.next_emergency_no = self.next_emergency_no + 1
   until self:computeNextEmergencyDates(emergency)
 end
@@ -1322,7 +1323,6 @@ end
 function World:computeNextEmergencyDates(emergency)
   -- Generate the next month and day the emergency should occur at.
   -- Make sure it doesn't happen in the past.
-  self.next_emergency = emergency
   local start = math.max(emergency.StartMonth, self.game_date:monthOfGame())
   if (emergency.EndMonth < start) then
     return false
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -2685,7 +2685,6 @@ if old < 153 then
       self.next_emergency_date = Date(1, self.next_emergency_month, self.next_emergency_day)
       -- Complementary afterLoad to see if emergencies got stuck in the level.
       -- There's no guarantee we can unstick the level, however.
-      local hosp = self:getLocalPlayerHospital()
       local next_emer_date = Date(1, self.next_emergency_month, self.next_emergency_day)
       --[[ UIWatch's emergency timer is 52 days but this is local.
       The emergency fax also is held for 16 days.
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -201,7 +201,12 @@ function World:World(app)
   self:calculateSpawnTiles()
 
   -- Next Events dates
+  -- emergencies
+  -- The emergency control level data starts with an array of 0
+  self.next_emergency_no = -1 -- Properly adjusted in nextEmergency()
   self:nextEmergency()
+
+  -- vip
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/world.lua
```diff
@@ -2678,20 +2678,26 @@ function World:afterLoad(old, new)
...
-        self:nextEmergency()
+if old < 153 then
+    -- Set the new variable next_emergency_date
+    -- In previous code month == 0 meant emergencies were over
+    if self.next_emergency_month ~= 0 then
+      self.next_emergency_date = Date(1, self.next_emergency_month, self.next_emergency_day)
+      -- Complementary afterLoad to see if emergencies got stuck in the level.
+      -- There's no guarantee we can unstick the level, however.
+      local hosp = self:getLocalPlayerHospital()
+      local next_emer_date = Date(1, self.next_emergency_month, self.next_emergency_day)
```
Alberth289346, line 2688: `hosp` isn't used in the patch, `next_emer_date` looks like the same computation as already done in `self.next_emergency_date`.

lewri: You're looking at the wrong commit :)

lewri:
Ready. But should wait for 1845 to be ready too.

CorsixTH/Lua/world.lua
```diff
@@ -1267,56 +1272,68 @@ function World:getReputationImpact(hospital)
...
+
+--! If a level file specifies random emergencies we make the next one as defined by the mean/variance given
+--!param control (table) Contains emergency information from level file
+function World:scheduleRandomEmergency(control)
+  -- Support standard values for mean and variance
+  local mean = control[0].Mean or 180
+  local variance = control[0].Variance or 30
+  -- How many days until next emergency?
+  local days = math.round(math.n_random(mean, variance))
+  local emergency_date = self.game_date:plusDays(days)
```
mugmuggy, line 1309: There is potential that #422 issue still exists within this logic where a level config overrides mean and variance with bad values and the next emergency is scheduled in the past.

lewri: So days needs to be +ve? Can probably do that easy

lewri:
Addressed #422 as pointed out. Should be good now.

TheCycoONE:
Could this parallel earthquakes a bit closer. I think it would reduce cognitive complexity. For earthquakes we have a current index (which is really a next index) starting from 0 and incremented at the end of the nextEarthquake function. We also put most of the details into a single object which I'd prefer if it doesn't increase the complexity.

lewri:
I plan to change to earthquakes actually to be more like this (it has no checks if it gets scheduled in the past). But making next_emergency increment at the end could be better, so I'll move that.

CorsixTH/Lua/world.lua
```diff
@@ -1143,7 +1148,7 @@ function World:onEndDay()
         -- The level uses random emergencies, so just create one.
         local_hospital:createEmergency()
       else
-        control = control[self.next_emergency_no]
+        control = self.next_emergency
```
TheCycoONE, line 1151: Predates you, but control shouldn't be used for both the list of all emergencies and for a particular one. You could directly use self.next_emergency below and get rid of this line.

PR #1844, opened 2021-03-23
Add check-for-updates to github.io

tobylane:
- Add the same check-for-updates file from corsixth.com to corsixth.github.io/CorsixTH.


TheCycoONE:
Thanks for doing this @tobylane

PR #1848, opened 2021-04-05
Changes supporting falling actions

### LINKED ISSUE #476: Falling over animations
_From [mark.l1...@hotmail.co.uk](https://code.google.com/u/117099726060447259318/) on September 22, 2011 15:12:11_

What are the major things which your patch changes / adds? Adds yawning for some patients that are waiting or if they have sleeping illness.
Added queue to the list of possible vomit/pee patients to test it out
Also add falling over and shaking of fist for some patients. What version of the product are you using? On what operating system? The falling animations are for when there is an earth quake, as some patients should fall down. Please provide any additional information below. As earth quakes are not in the game yet (fingers are crossed that they never will be) I have made it that from time to time if you click on a patient he or she could fall over.  This is not in TH, but I thought it may make people smile.

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1020_

mugmuggy:
Screenshot 2026-06-15 at 14 11 37

*Fixes #476*

- Corrects animations for Female patient (this probably doesn't work if they are a Stripped version at the time)
- Fixes the must happen to be in the constructor, otherwise any setNextAction calls will remove them
- Patients randomly choose a fall direction then that direction doesn't change, the random would have them fall one way, lay another and then stand up again in another
- Queues the actions more correctly for falling

I didn't activate the falling code in world, so these don't actually initiate anything but can be used for testing.

Doesn't as pointed out in the discussion #1846 attempt to address the issue of interrupted actions being incorrectly requeued for example, walking to room that is destroyed at the same time the fall action starts, which would requeue the interrupted walk to that room.

I had put a has_interrupted flag on the action and detected it as not to requeue, but not all interruptions are driven through setNextAction (interrupt_head in queue for example) and I think the other issue was actions like the walk which starts and does something, sets the interrupt handler, which means the action starts, rather than a todo_interrupt being detected to basically skip the action in the action start (seek_toilets_action_start for example and this is same issue with teleporting the drinks machine or jumping to/from bench when queue actions are interrupted as use_object also must start)

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -29,7 +29,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 189 -- Refactor earthquakes
+local SAVEGAME_VERSION = 190 -- Improve falling actions for debug purposes
 
 class "App"
 
--- a/CorsixTH/Lua/earthquake.lua
+++ b/CorsixTH/Lua/earthquake.lua
@@ -145,7 +145,18 @@ function Earthquake:tick()
   self.remaining_damage = self.remaining_damage - 1
   self.damage_timer = self.damage_timer + damage_time
 
-  -- Patient falling: work in progress, see PR 1848
+  -- The below code triggers random patient falls during an earthquake.
+  -- It is currently disabled except for debugging purposes in the config file.
+  -- Current behaviour can cause empty action queues or other undesired behaviours.
+  -- Once working, the debugging flag can be removed.
+  if not TheApp.config.debug_falling then return end
+  local hospital = self.world:getLocalPlayerHospital()
+  -- loop through the patients and allow the possibility for them to fall over
+  for _, patient in ipairs(hospital.patients) do
+    if not patient.in_room and patient.falling_anim then
+      patient:falling(false)
+    end
+  end
 end
 
 --! Check if it's time for an earthquake
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -361,20 +361,14 @@ function Patient:falling()
   local current = self:getCurrentAction()
   current.keep_reserved = true
   if self.falling_anim and self:canPeeOrPuke(current) and self.has_fallen == 1 then
-    self:setNextAction(FallingAction(), 0)
     self.has_fallen = 2
-    if self.has_fallen == 2 then
-      self:setNextAction(OnGroundAction())
-      self.on_ground = true
-    end
-    if self.on_ground then
-      self:setNextAction(GetUpAction())
-    end
-    -- TODO: falls are broken, the will change once implemented
-    self:interruptAndRequeueAction(current, 2)
-    self.on_ground = false
-    if math.random(1, 5) == 3 then
-      self:shakeFist()
+    self:queueAction(FallingAction(), 1)
+    self:queueAction(OnGroundAction(), 2)
+    self:queueAction(GetUpAction(), 3)
+    if math.random(1, 5) == 3 and self:shakeFist() then
+      self:interruptAndRequeueAction(current, 5)
+    else
+      self:interruptAndRequeueAction(current, 4)
     end
     self:fallingAnnounce()
     self:changeAttribute("happiness", -0.05) -- falling makes you very unhappy
@@ -398,7 +392,8 @@ end
 --! Perform 'shake fist' action.
 function Patient:shakeFist()
   if self.shake_fist_anim then
-    self:queueAction(ShakeFistAction(), 1)
+    self:queueAction(ShakeFistAction(), 4)
+    return true
   end
 end
 
@@ -1124,6 +1119,11 @@ function Patient:afterLoad(old, new)
           action.must_happen = false
         end
       end
+    end
+  end
+  if old < 190 then
+    if self.humanoid_class == "Standard Female Patient" then
+      self.on_ground_anim = 1764
     end
   end
   self:updateDynamicInfo()
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -72,6 +72,14 @@ function Patient:onClick(ui, button)
         ui:addWindow(UIPatient(ui, self))
       end
     end
+  elseif TheApp.config.debug_falling and button == "right" then
+    -- Attempt to push patient over
+    -- Currently debug-only, enable in config file for testing.
+    -- Once confirmed working, the debugging flag can be removed.
+    if not self.world:isPaused() and not (self.cured or self.dead or self.going_home)
+         and math.random(1, 2) == 2 then
+      self:falling(true)
+    end
   elseif self.user_of then
     -- The object we're using is made invisible, as the animation contains both
     -- the humanoid and the object. Hence send the click onto the object.
@@ -341,10 +349,10 @@ function Patient:die()
   self:setDynamicInfoText(_S.dynamic_info.patient.actions.dying)
 end
 
--- Actions we can interrupt for in the canPeeOrPuke function
+-- Actions we can interrupt for in the canPeePukeOrFall function
 local good_actions = {walk=true, idle=true, seek_room=true, queue=true}
 
-function Patient:canPeeOrPuke(current)
+function Patient:canPeePukeOrFall(current)
   if not good_actions[current.name] then return false end
   if self.going_home then return false end
 
@@ -356,24 +364,25 @@ function Patient:canPeeOrPuke(current)
   return parcel ~= 0 and th:getPlotOwner(parcel) == self.hospital:getPlayerIndex()
 end
 
---! Animations for when there is an earth quake
-function Patient:falling()
+--! Falling animations for when there is an earth quake or the player is being very mean
+--!param player_init (bool) if true, player triggered the fall
+function Patient:falling(player_init)
   local current = self:getCurrentAction()
   current.keep_reserved = true
-  if self.falling_anim and self:canPeeOrPuke(current) and self.has_fallen == 1 then
+  if self.falling_anim and self:canPeePukeOrFall(current) and self.has_fallen == 1 then
     self.has_fallen = 2
     self:queueAction(FallingAction(), 1)
     self:queueAction(OnGroundAction(), 2)
     self:queueAction(GetUpAction(), 3)
-    if math.random(1, 5) == 3 and self:shakeFist() then
+    -- show the patient is annoyed, if possbile
+    if math.random(1, 5) == 3 and self.shake_fist_anim then
+      self:queueAction(ShakeFistAction(), 4)
       self:interruptAndRequeueAction(current, 5)
     else
       self:interruptAndRequeueAction(current, 4)
     end
-    self:fallingAnnounce()
+    if player_init then self:fallingAnnounce() end
     self:changeAttribute("happiness", -0.05) -- falling makes you very unhappy
-  else
-    return
   end
 end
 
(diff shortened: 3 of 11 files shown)


lewri:
Iirc patients can't fall inside rooms. So think you should be fine unless females are stripping around the hospital

lewri (commented):
Maybe out of scope a little but falling's line 28 should be a colon and not a full stop (I think)

i.e. this line needs correcting
```lua
self.setMustHappen(true)
```

mugmuggy:
Forgot to copy that one across. The Stripped Female Patients won't get the updated on_ground animation in the afterload, so once they become normal ones, they will be restored with the old one, also if they transition from Standard to Stripped then back to Standard. Obviously not a problem on new saves.

lewri:
@mugmuggy May need to be it's own issue but it looks like some recent update triggering falls is causing the adviser to repeatedly saying the "These people are ill" "How would you like it" etc. when the falling is on.

lewri:
I think we lost the GitHub logs on this one, if you wanted to try and regenerate them.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -381,7 +375,8 @@ end
 --! Perform 'shake fist' action.
 function Patient:shakeFist()
   if self.shake_fist_anim then
-    self:queueAction(ShakeFistAction(), 1)
+    self:queueAction(ShakeFistAction(), 4)
```
Alberth289346, line 378: There is exactly one use of this function, so not much need for having it as a function.

Alberth289346:
Looks good to me, except for the problem spotted by Lewri. My suggestion is just further improvement of the code while you are at it.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -342,20 +342,14 @@ function Patient:falling()
...
-      self:shakeFist()
+    self:queueAction(FallingAction(), 1)
+    self:queueAction(OnGroundAction(), 2)
+    self:queueAction(GetUpAction(), 3)
+    if math.random(1, 5) == 3 and self:shakeFist() then
+      self:interruptAndRequeueAction(current, 5)
+    else
+      self:interruptAndRequeueAction(current, 4)
     end
     self:fallingAnnounce()
```
lewri, line 354: Ah, the problem I identified a while back sits here. This ends up spamming the announcer with every person falling.
It's out of scope so just needs noting to be fixed (or fixed here, if you wanted to).

lewri: In fact, the falling Adviser messages are irrelevant to earthquakes.... They're when you click a patient I believe to knock them over.

mugmuggy:
> > Patients randomly choose a fall direction then that direction doesn't change, the random would have them fall one way, lay another and then stand up again in another
> 
> Unfortunately I don't think this one is true. I'm still see this even with your changes.
Specifically this is the on_ground/get_up where it could change direction whilst on the ground. They still fall in a random direction, but don't flop around on the ground.

lewri:
> > > Patients randomly choose a fall direction then that direction doesn't change, the random would have them fall one way, lay another and then stand up again in another
> > 
> > 
> > Unfortunately I don't think this one is true. I'm still see this even with your changes.
> > Specifically this is the on_ground/get_up where it could change direction whilst on the ground. They still fall in a random direction, but don't flop around on the ground.

Hmm. I've attached a webm here showing what's happening still.

https://user-images.githubusercontent.com/20030128/172935256-198499dc-a34e-402b-9cb6-f3bb35b00531.mp4

CorsixTH/Lua/entities/humanoids/patient.lua
lewri: Could you provide a suggestion for this code spot? Lets see if we can get this one in a tidier place and we can merge what we have since falling is still disabled and for debugging only.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -356,27 +364,22 @@ function Patient:canPeeOrPuke(current)
...
+    self:queueAction(FallingAction(), 1)
+    self:queueAction(OnGroundAction(), 2)
+    self:queueAction(GetUpAction(), 3)
+    if math.random(1, 5) == 3 and self:shakeFist() then
+      self:interruptAndRequeueAction(current, 5)
+    else
+      self:interruptAndRequeueAction(current, 4)
     end
-    self:fallingAnnounce()
+    if player_init then self:fallingAnnounce(true) end
```
lewri, line 382: the `true` isn't needed. Not sure what I was doing.

CorsixTH/Lua/entities/humanoids/patient.lua
Alberth289346:
``` bash
$ ag shakeFist 

CorsixTH/Lua/entities/humanoids/patient.lua
377:      self:shakeFist()
399:function Patient:shakeFist()
```

Usage is thus at line 377, maybe change it to something like
``` lua
    if math.random(1, 5) == 3 and self.shake_fist_anim then
      self:queueAction(ShakeFistAction(), 4)
      self:interruptAndRequeueAction(current, 5)
    else
      self:interruptAndRequeueAction(current, 4)
    end
```

Alberth289346: This also helps in understanding the magic 4 :)

lewri: > EDIT: Hard-coding the queue position in the middle of the action queue makes it quite unusable. If you want to keep this, pass the index as a parameter.

Does your suggestion resolve this too? Or should this till be factored.

Alberth289346: Nope, didn't fix that. I would have said so if it did.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -356,27 +364,22 @@ function Patient:canPeeOrPuke(current)
...
+    if math.random(1, 5) == 3 and self:shakeFist() then
+      self:interruptAndRequeueAction(current, 5)
+    else
+      self:interruptAndRequeueAction(current, 4)
     end
-    self:fallingAnnounce()
+    if player_init then self:fallingAnnounce() end
     self:changeAttribute("happiness", -0.05) -- falling makes you very unhappy
   else
     return
```
Alberth289346, line 385: Invert the condition so the return can be done immediately after deciding it's not happening.

Alberth289346:
Not sure why it needs a new section in the config file. Can't we have another setting in the "debug" settings section? May somewhat depend on the public though.

lewri:
> Not sure why it needs a new section in the config file. Can't we have another setting in the "debug" settings section? May somewhat depend on the public though.

As an example, for soundfont override on fluidsynth builds we don't give any indicators the setting `config.soundfont` exists. We could do that here too. Though I'm not keen on some magic undocumented flag either.

CorsixTH/Lua/entities/humanoids/patient.lua
lewri: `if not self.falling_anim or not self:canPeeOrPuke(current) and not self.has_fallen == 1 then return end`?

Alberth289346: Since the function ends there as well, just delete the "else return" instead?

Alberth289346: Instead of your long condition, you can break it into 3 parts, and bail out as soon as one fails.

Alberth289346: An unsure why it needs a "pee or puke" function. Probably it was re-used without adapting the name.

lewri: I think there's 3 options listed here now. Which one is the preferred one here? 😅 

> An unsure why it needs a "pee or puke" function. Probably it was re-used without adapting the name.

Very old code :) https://github.com/CorsixTH/CorsixTH/commit/df84d21f926a5868f86904b7d968bdf76d20cba6

Alberth289346: > Which one is the preferred one here?

Hmm, you're asking such difficult questions :)
I'd probably go for deleting the `else return`, as it's not doing anything useful.

> Very old code :)

Point is that it's being used for something else than pee or puke. The name of the function doesn't cover all it's purposes anymore. You can extend it with the new purpose, but maybe a more general name is better.? I pondered about a name and came up with `atFreeTile` or `atOpenTile`. Not sure these are good names though.

lewri: The function checks more than tiles, it has 3 steps:
1. Is the current action interruptible?
2. Does the patient have going home status?
3. Finally check if the current tile is buildable, owned, and in a hospital

It's likely the extend name option is going to be the easiest here; can't think of a good general name either.

Alberth289346: I am fine with extending the name.

Good point, so perhaps this is not just 1 function, it's 2 or 3 functions mixed into one?  
The first 2 steps seem related to the question "is it possible and does it make sense to insert a new action in the queue?", while the 3rd step seems about an empty tile (which probably should be abstracted into an `isAtEmptyTile` function, but that is out of scope here).

I am not sure about the "inHospital" check for the current usage of this function. For pee and puke it makes sense, since a handyman needs to cleanup the mess, but falling doesn't need to be cleaned. Not a serious problem though, falling patients at the middle of some road are never seen anyway.  
If you want to fix things in that function, doing that in a separate patch is likely the better idea, imho.

Alberth289346 (approved):
Not sure how good it is exactly, but let's find out :)

CorsixTH/Lua/app.lua
```diff
@@ -29,7 +29,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 188 -- Add groups to win and lose criteria.
+local SAVEGAME_VERSION = 183 -- WIP support falling actions
```
lewri, line 32: Will refresh, I just needed a new artifact for this PR

lewri:
Everything should be cleared up now and ready. I also made this a debug menu item for easier toggling,

TheCycoONE:
LGTM too

CorsixTH/Lua/world.lua
```diff
@@ -594,26 +594,19 @@ function World:tickEarthquake()
...
-        -- end up with nil action queues, and sometimes the resumed
-        -- actions throw exceptions. Also, patients in the hospital
-        -- who have not yet found reception throw exceptions after
-        -- they visit reception. Some debugging needed here to get
-        -- this working.
-
-        patient:falling()
+        patient:falling(false)
       end
     end
```
lewri, line 609: Relocate to earthquake class before merge

CorsixTH/Lua/earthquake.lua
```diff
@@ -145,7 +145,18 @@ function Earthquake:tick()
...
+  -- The below code triggers random patient falls during an earthquake.
+  -- It is currently disabled except for debugging purposes in the config file.
+  -- Current behaviour can cause empty action queues or other undesired behaviours.
+  -- Once working, the debugging flag can be removed.
+  if not TheApp.config.debug_falling then return end
+  local hospital = self.world:getLocalPlayerHospital()
+  -- loop through the patients and allow the possibility for them to fall over
+  for _, patient in ipairs(hospital.patients) do
+    if not patient.in_room and patient.falling_anim then
+      patient:falling(false)
```
lewri, line 157: Should there be a probability roll here?

Alberth289346: Hmm, "tick" function eh?  Should be a very low probability if you want to notice any of a delay, I guess.

Quick falling is perhaps more useful for testing?

lewri: Looks like it's every quake tick (which different from a regular tick), which in essence is a damage cycle completed.

Potentially its fine for now. If a patient can't fall at one part of the quake they may move and be able to later on. I think if it becomes too uniform then we can adjust later

Alberth289346: I agree :)   First make it work at all before starting to fine-tune it :)

PR #1849, opened 2021-04-07
[RDY] Use service quality in picking staff member from multi occupancy rooms

tobylane:
- Merge the specific (and slightly different) getStaffMembers from the rooms that extend it into the original in room.
- Use service quality rather than skill or tiredness alone.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -670,19 +670,20 @@ function Staff:getDrawingLayer()
   return 4
 end
 
---! Estimate staff service quality based on skills, fatigue and happiness.
+--! Estimate staff service quality based on skills, restfulness (inverse of fatigue) and happiness.
 --!return (float) between [0-1] indicating quality of the service.
 function Staff:getServiceQuality()
   -- weights
   local skill_weight = 0.7
-  local fatigue_weight = 0.2
+  local restfulness_weight = 0.2
   local happiness_weight = 0.1
 
   local weighted_skill = skill_weight * self.profile.skill
-  local weighted_fatigue = fatigue_weight * self.attributes["fatigue"]
+  -- Less fatigue is better
+  local weighted_restfulness = restfulness_weight * (1 - self.attributes["fatigue"])
   local weighted_happiness = happiness_weight * self.attributes["happiness"]
 
-  return weighted_skill + weighted_fatigue + weighted_happiness
+  return weighted_skill + weighted_restfulness + weighted_happiness
 end
 
 --[[ Return string representation
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -429,10 +429,20 @@ function Room:createDealtWithPatientCallback(humanoid)
 end
 
 --! Get the current staff member.
--- Can be overridden in rooms with multiple staff members to return the desired one.
+-- In multi-occupancy rooms this returns the staff member with the minimum service quality
 --!return (staff) The current staff member.
 function Room:getStaffMember()
-  return self.staff_member
+  if not self.staff_member_set then return self.staff_member end
+
+  local staff
+  for staff_member, _ in pairs(self.staff_member_set) do
+    if not staff_member.fired and not staff_member:hasLeavingAction() then
+      if not staff or staff:getServiceQuality() > staff_member:getServiceQuality() then
+        staff = staff_member
+      end
+    end
+  end
+  return staff
 end
 
 --! Set the current staff member.
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -929,13 +929,14 @@ function Humanoid:tostring()
 
   local result = string.format("%s - class: %s", full_name, class)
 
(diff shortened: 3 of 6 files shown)


CorsixTH/Lua/room.lua
```diff
@@ -429,10 +429,24 @@ function Room:createDealtWithPatientCallback(humanoid)
...
+  if not self.staff_member_set then return self.staff_member end
+
+  local staff
+  for staff_member, _ in pairs(self.staff_member_set) do
+    if staff and not staff_member.fired and not staff_member:hasLeavingAction() then
+      if staff:getServiceQuality() > staff_member:getServiceQuality() then
+        staff = staff_member
+      end
+    else
+      if not staff_member.fired and not staff_member:hasLeavingAction() then
```
Alberth289346, line 444: In a room with more staff, if `staff` is not nil but `staff_member:hasLeavingAction()` is true, you first check all conditions at line 439. That fails, and then you check the last 2 again at line 444.

The simpler and cheaper form is to first check the conditions like at line 444, and if these hold, decide whether to also take the service quality into account based on the value of `staff`.

CorsixTH/Lua/room.lua
```diff
@@ -429,10 +429,24 @@ function Room:createDealtWithPatientCallback(humanoid)
 end
 
 --! Get the current staff member.
--- Can be overridden in rooms with multiple staff members to return the desired one.
+-- In multi-occupancy rooms this returns the staff member with the minimum service quality
 --!return (staff) The current staff member.
 function Room:getStaffMember()
-  return self.staff_member
+  if not self.staff_member_set then return self.staff_member end
```
Alberth289346, line 435: The simpler system is to always have a set of staff members, even for rooms with 1 staff member. It allows to always use the same system to access staff members for every room, instead of having to distinguish for different cases all throughout the code.

tobylane:
Improved the loop. The change for all rooms to have a staff member set will be separate.

CorsixTH/Lua/room.lua
```diff
@@ -429,10 +429,20 @@ function Room:createDealtWithPatientCallback(humanoid)
...
+
+  local staff
+  for staff_member, _ in pairs(self.staff_member_set) do
+    if not staff_member.fired and not staff_member:hasLeavingAction() then
+      if not staff or (staff and staff:getServiceQuality() > staff_member:getServiceQuality()) then
```
Alberth289346, line 440: If `not staff` holds, the second part of the `or` isn't evaluated.

If `not staff` fails, the second part of the `or` is evaluated, but there `staff` then always holds, since `not staff` failed.
In other words, no need to test for `staff` in the second part, as you already checked that by a failing `not staff` in the first part of the `or`.

tobylane: Good, done.

Alberth289346:
Looks fine to me.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -670,19 +670,19 @@ function Staff:getDrawingLayer()
...
 end
 
---! Estimate staff service quality based on skills, fatigue and happiness.
+--! Estimate staff service quality based on skills, restfulness (inverse of fatigue) and happiness.
 --!return (float) between [0-1] indicating quality of the service.
 function Staff:getServiceQuality()
   -- weights
   local skill_weight = 0.7
-  local fatigue_weight = 0.2
+  local rest_weight = -0.2 -- Less fatigue is better
```
lewri, line 678: With this new variable, we should probably protect against the service quality going <0. It's highly unlikely but could happen if you have a verrrryyy tired unskilled staff member who's miserable.

lewri: Also that now this is a negative number we the variable name may need a tweak

tobylane: The function doc says [0-1] so a minimum of 0 will be enforced.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -670,19 +670,21 @@ function Staff:getDrawingLayer()
...
   -- weights
   local skill_weight = 0.7
-  local fatigue_weight = 0.2
+  local restfulness_weight = 0.2
   local happiness_weight = 0.1
 
   local weighted_skill = skill_weight * self.profile.skill
-  local weighted_fatigue = fatigue_weight * self.attributes["fatigue"]
+  -- Less fatigue is better
+  local weighted_restfulness = restfulness_weight * -self.attributes["fatigue"]
```
lewri, line 683: You've introduced a caveat here where service quality can no longer go above 0.8. Instead consider this. 
```suggestion
  local weighted_restfulness = 0.2 + (restfulness_weight * -self.attributes["fatigue"])
```
Or to be more logical make it so the + is the -

CorsixTH/Lua/room.lua
```diff
@@ -429,10 +429,20 @@ function Room:createDealtWithPatientCallback(humanoid)
...
+
+  local staff
+  for staff_member, _ in pairs(self.staff_member_set) do
+    if not staff_member.fired and not staff_member:hasLeavingAction() then
+      if not staff or staff:getServiceQuality() > staff_member:getServiceQuality() then
+        staff = staff_member
+      end
+    end
+  end
+  return staff
```
Alberth289346, line 445: No need to actually change imho, but if you used `selected_staff` instead of `staff` code would have been very slightly more easy to read.

tobylane:
Ready.

PR #1851, opened 2021-04-10
[RDY] Deal with staff attribute checks

### LINKED ISSUE #1838: [Bug] False missing Psychiatrist fax
#### Describe the issue
At some point between 0.64 and Trunk 152 the game is thinking I don't have a Psychiatrist even though a Consultant Surgeon Psychiatrist is given at the start of the level.

#### Steps to Reproduce
1. *Load save game attached to this ticket* (may be useless)

2. Have a Psych built and hire no new doctors etc.

3. Wake for a psych patient to need psych.

#### Save Game

[emergencytests.zip](https://github.com/CorsixTH/CorsixTH/files/6140466/emergencytests.zip)

#### Expected Behavior
See the Psychiatrist

#### System Information
**CorsixTH Version:** *Trunk 151*

#### Gamelog.txt

```
Paste gamelog.txt output here
```

### Additional Info
illustrate the problem.*
image

You can temporarily fix it by editing Psychiatrist and going back 1 stage to replace door, then finish editing. But it comes back.

mugmuggy:
*Fixes #1838*

- Use staff.profile to avoid inline class conversions, for Surgeons
- Reuse fulfillsCriterion function to support counting staff Category

fulfillsCriterion seemed to only really be used for flow of logic when a room was in scope in the calls dispatcher where a room is in scope or a room calling it within its methods so Receptionist seems safe to implement.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff/doctor.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff/doctor.lua
@@ -249,17 +249,18 @@ local profile_attributes = {
   Psychiatrist = "is_psychiatrist",
   Surgeon = "is_surgeon",
   Researcher = "is_researcher",
+  Junior = "is_junior",
+  Consultant = "is_consultant",
 }
 
 -- Helper function to decide if Staff fulfills a criterion
--- (one of "Doctor", "Nurse", "Psychiatrist", "Surgeon", "Researcher" and "Handyman")
+-- (one of "Doctor", "Nurse", "Psychiatrist", "Surgeon", "Researcher" and "Handyman", "Receptionist", "Junior", "Consultant")
 function Doctor:fulfillsCriterion(criterion)
   if criterion == "Doctor" then
     return true
-  elseif criterion == "Psychiatrist" or criterion == "Surgeon" or criterion == "Researcher" then
-    if self.profile and self.profile[profile_attributes[criterion]] == 1.0 then
-      return true
-    end
+  end
+  if self.profile and self.profile[profile_attributes[criterion]] == 1.0 then
+    return true
   end
   return false
 end
--- a/CorsixTH/Lua/entities/humanoids/staff/receptionist.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff/receptionist.lua
@@ -83,9 +83,9 @@ end
 
 
 -- Helper function to decide if Staff fulfills a criterion
--- (one of "Doctor", "Nurse", "Psychiatrist", "Surgeon", "Researcher" and "Handyman")
-function Receptionist:fulfillsCriterion(criterion) -- luacheck: no unused args
-  return false
+-- (one of "Doctor", "Nurse", "Psychiatrist", "Surgeon", "Researcher" and "Handyman", "Receptionist", "Junior", "Consultant")
+function Receptionist:fulfillsCriterion(criterion)
+  return criterion == "Receptionist"
 end
 
 function Receptionist:getDrawingLayer()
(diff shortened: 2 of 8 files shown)


CorsixTH/Lua/entities/humanoids/staff/receptionist.lua
```diff
@@ -84,7 +84,10 @@ end
 
 -- Helper function to decide if Staff fulfills a criterion
 -- (one of "Doctor", "Nurse", "Psychiatrist", "Surgeon", "Researcher" and "Handyman")
-function Receptionist:fulfillsCriterion(criterion) -- luacheck: no unused args
+function Receptionist:fulfillsCriterion(criterion)
+  if criterion == "Receptionist" then
+    return true
+  end
   return false
```
Alberth289346, line 91: The entire body could be `return criterion == "Receptionist"`

mugmuggy: Kept it in line with Handyman and Nurse instances, where it does it the long way too.

Doco wise I thought I would check fulfillsCriterion was ok as its used for checking staff for the room. As that same comment is on all the derived classes where this is overridden also and then it has Junior/Consultant in the Doctor case, which aren't criterion for a room anymore, and maybe they need to be separated in another method.

Alberth289346: We have had quite a few starting programmers, and often they don't know that "==" delivers a boolean value. Simple beginner programs don't manipulate boolean values so they never see such things. For this reason starting programmers think that comparisons exclusively belong in an `if` or `while`, and booleans exist as `true` and `false` only. And then you get `if x==y then return true else return false end` as result.
As we have had quite a few starting programmers, there are lots of such instances in the code.

As for documentation, I am fine if you change allowed input or change output or change aim of the code, but please reflect that in changes of the documentation.

Alberth289346:
looks fine to me

CorsixTH/Lua/entities/humanoids/staff/doctor.lua
```diff
@@ -249,17 +249,21 @@ local profile_attributes = {
...
--- (one of "Doctor", "Nurse", "Psychiatrist", "Surgeon", "Researcher" and "Handyman")
+-- (one of "Doctor", "Nurse", "Psychiatrist", "Surgeon", "Researcher" and "Handyman", "Receptionist", "Junior", "Consultant")
 function Doctor:fulfillsCriterion(criterion)
   if criterion == "Doctor" then
     return true
   elseif criterion == "Psychiatrist" or criterion == "Surgeon" or criterion == "Researcher" then
     if self.profile and self.profile[profile_attributes[criterion]] == 1.0 then
       return true
     end
+  elseif criterion == "Junior" or criterion == "Consultant" then
```
TheCycoONE, line 265: This is the same logic as the condition above, but written differently which makes the function harder to read.

This staff is not a Nurse or Handyman so they wouldn't have those profile attributes.

Couldn't the whole thing be reduced to:
```
if criterion == "Doctor" then
  return true
end
if self.profile and self.profile[profile_attributes[criterion]] then
  return true
end
return false
```
To me that's  easier to read.

TheCycoONE (commented):
Looks ok, just the one comment.

CorsixTH/Lua/entities/humanoids/staff/doctor.lua
mugmuggy: Indeed, I took the countStaffOfCategory as boolean checks and just mapped them over, but really for is_consultant and is_junior it was 1 or nil (an existence check) and they do fall into that so can be simplified, just with the value check still, rather than just an existence check otherwise partially trained doctors will meet the skill.

TheCycoONE: Good catch. Please do.

PR #1852, opened 2021-04-10
[RDY] Add to reception queue only when near reception #372

### LINKED ISSUE #372: New arrivals claim position in Reception Queue too soon
Everyone joins the queue for reception as soon as they enter the hospital.  If you have your desk near to the only entrance this is not so noticeable.  However as you can see in the picture the doors at the bottom are a long way from the desk, but he is already second in the queue.  This can cause a queue to build up at reception when there are multiple entrances and lots of visitors.  VIPs stop others being dealt with as they are pushed to the front, even though they may be a long way from the desk.

This is not a regression, it has been this way for a while now.

reception_queue

mugmuggy:
*Fixes #372*

- Walk is truncated to a distance of 6 tiles away from the reception desk use tile.
- Updates walk actions destination tile to that of the truncated walk (incase of vomit)

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/seek_reception.lua
+++ b/CorsixTH/Lua/humanoid_actions/seek_reception.lua
@@ -30,10 +30,10 @@ function SeekReceptionAction:SeekReceptionAction()
   self:HumanoidAction("seek_reception")
 end
 
-local function can_join_queue_at(humanoid, x, y)
+local function can_join_queue_at(humanoid, x, y, dist)
   local flag_cache = humanoid.world.map.th:getCellFlags(x, y)
   return flag_cache.hospital and not flag_cache.room and
-      humanoid.hospital and
+      dist <= can_join_queue_distance and
       flag_cache.owner == humanoid.hospital:getPlayerIndex()
 end
 
@@ -81,7 +81,7 @@ local function action_seek_reception_start(action, humanoid)
 
     -- We don't want patients which have just spawned to be joining the queue
     -- immediately, so walk them closer to the desk before joining the queue
-    if can_join_queue_at(humanoid, humanoid.tile_x, humanoid.tile_y) and dist <= can_join_queue_distance then
+    if can_join_queue_at(humanoid, humanoid.tile_x, humanoid.tile_y, dist) then
       local face_x, face_y = best_desk:getSecondaryUsageTile()
       humanoid:setNextAction(QueueAction(x, y, best_desk.queue):setMustHappen(action.must_happen)
           :setFaceDirection(face_x, face_y))
@@ -90,8 +90,9 @@ local function action_seek_reception_start(action, humanoid)
       humanoid:queueAction(walk, 0)
 
       -- Trim the walk to finish once it is possible to join the queue
-      for i = #walk.path_x, 2, -1 do
-        if can_join_queue_at(humanoid, walk.path_x[i], walk.path_y[i]) and dist - i < can_join_queue_distance then
+      local pathindex = #walk.path_x
+      for i = pathindex - 1, 2, -1 do
+        if can_join_queue_at(humanoid, walk.path_x[i], walk.path_y[i], pathindex - i) then
           walk.path_x[i + 1] = nil
           walk.path_y[i + 1] = nil
           walk.x = walk.path_x[i]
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/humanoid_actions/seek_reception.lua
```diff
@@ -71,13 +74,14 @@ local function action_seek_reception_start(action, humanoid)
...
     local y = best_desk.tile_y + orientation.use_position[2]
+    local dist = humanoid.world:getPathDistance(humanoid.tile_x, humanoid.tile_y, x, y)
     humanoid:updateDynamicInfo(_S.dynamic_info.patient.actions.on_my_way_to
       :format(best_desk.object_type.name))
     humanoid.waiting = nil
 
     -- We don't want patients which have just spawned to be joining the queue
     -- immediately, so walk them closer to the desk before joining the queue
-    if can_join_queue_at(humanoid, humanoid.tile_x, humanoid.tile_y) then
+    if can_join_queue_at(humanoid, humanoid.tile_x, humanoid.tile_y) and dist <= can_join_queue_distance then
```
Alberth289346, line 84: Your addition at all calls invalidates the name of the function. The function doesn't make the decision it claims to make any more since the additional condition could now block that decision. To clean up, I would suggest to either push the distance test inside the function, rename and possibly move the function so it doesn't make claims that aren't true, or even remove it entirely.

The function currently checks the humanoid for "inside the correct hospital". A function you'd expect as part of the humanoid code (it may even exist, didn't check) rather than part of reception.
Obviously "in the hospital" is better than joining at the spawn point so in that sense the function does what it claims compared to not having such a function, but you'd expect such a function to take the desk position into account, like you are adding. In other words, judging by the name of the the function it could also fit as part of the desk instance.

On the other hand, your addition makes a much tighter check, so arguably, "inside the right hospital" isn't very relevant any more. (With pathdistance, you also eliminated the option to walk outside near the desk, but still having to walk around the entire building.)

mugmuggy: Indeed Hospital:isInHospital plus a check for being in a room is all that can_join_queue_at is.

I agree with the can_join_queue_at with the extra parameter as thats an obvious alteration or break it down into the constituent calls, easier though to have just one function I guess. More I didn't integrate it in case there was something obvious I missed for now.

Outside you aren't in the hospital (well shouldn't be), but still possible within the distance, so can't join anyway before this distance check, the walk was truncated to within the bounds of the hospital already. But you can be in the hospital but not the same building when performing a SeekReception action so distance is important.

A patient can enter to join one and re-evaluate. But now they are on the otherside of the map. and choosing another desk and with the penalty in getUsageScores for a full queue, if you have more than one and one is full and one not, they might walk across the map entirely. You also don't want that QueueAction to start as they will queue from the other side of the map and lock the reception desk again, and given VIPs get priority. So either way its not ideal to rely on the hospital itself.

It maybe better to do the same best desk logic in the from outside the hospital and once they are in the hospital get nearest, but would probably want the TH feature of the queues being shared between reception desks placed close together before contemplating that.

Alberth289346: I always wondered why we don't treat reception just like a normal room, except for the door. It would allow all the queue management logic for rooms to be applied.

As for queue sharing, ideally they should be shared, but on the other hand, we don't share queues between several same  normal rooms close to each other either, eg 3 GPs next to each other.

CorsixTH/Lua/humanoid_actions/seek_reception.lua
```diff
@@ -23,6 +23,9 @@ class "SeekReceptionAction" (HumanoidAction)
 ---@type SeekReceptionAction
 local SeekReceptionAction = _G["SeekReceptionAction"]
 
+-- distance from queue that patient can join a reception desk queue
+local can_join_queue_distance = 6
```
lewri, line 27: Looking at room.lua and app.lua there are hints this variable should be ALL_CAPS...
Though the coding conventions do not give strict guidance to support this. @TheCycoONE ?

TheCycoONE: We're not consistent, see e.g. hours_per_day in date.lua or case_pattern in strings.lua. I don't feel strongly about it and I'm not going to push the matter either way. (In C++ we reserve screaming snake case exclusively for preprocessor macros, but in lua there's no such considerations.)

lewri:
Overall not strongly opposed to the PR, if the function names are adjusted as per Alberth's suggestion should be fine.
Noted you talk about sharing queues of nearby reception desks -- I like this (but never used it actively) but likely a separate enhancement.

mugmuggy:
Just made the local function make that decision now. Just replaced the 'and humanoid.hospital' as there is an assert that checks humanoid has a hospital anyway for the distance check. And through a small change to the for loop for the walk action truncation, skip one unnecessary iteration, and consolidate the comparison to <= in both instances.

PR #1853, opened 2021-04-14
[RDY] Use initial reputation and overdraft interest rate if set in level config

tobylane:
*Previous feedback #1722, #1741*

- Reputation and the overdraft interest differential can be set in levels
- Change interest and overdraft interest to a number over 10,000. Consistent with how it's set in levels.

### DIFF
--- a/CorsixTH/Lua/base_config.lua
+++ b/CorsixTH/Lua/base_config.lua
@@ -28,6 +28,7 @@ local configuration = {
   town = {
     InterestRate = 0.01,
     StartCash = 40000,
+    StartRep = 500,
   },
 
   -- New value, but should only be defined if starting staff is included.
@@ -133,19 +134,19 @@ local configuration = {
   },
 
   towns = {
-    {StartCash = 40000, InterestRate = 100}, -- Level 1
-    {StartCash = 40000, InterestRate = 200}, --  Level 2
-    {StartCash = 50000, InterestRate = 300}, --  Level 3
-    {StartCash = 50000, InterestRate = 400}, --  Level 4
-    {StartCash = 50000, InterestRate = 500}, --  Level 5
-    {StartCash = 50000, InterestRate = 600}, --  Level 6
-    {StartCash = 50000, InterestRate = 700}, --  Level 7
-    {StartCash = 60000, InterestRate = 700}, --  Level 8
-    {StartCash = 60000, InterestRate = 800}, --  Level 9
-    {StartCash = 60000, InterestRate = 800}, --  Level 10
-    {StartCash = 70000, InterestRate = 900}, --  Level 11
-    {StartCash = 70000, InterestRate = 900}, --  Level 12
-    {StartCash = 70000, InterestRate = 900}, --  Level 12
+    {StartCash = 40000, InterestRate = 100, StartRep = 500}, -- Level 1
+    {StartCash = 40000, InterestRate = 200, StartRep = 500}, --  Level 2
+    {StartCash = 50000, InterestRate = 300, StartRep = 500}, --  Level 3
+    {StartCash = 50000, InterestRate = 400, StartRep = 500}, --  Level 4
+    {StartCash = 50000, InterestRate = 500, StartRep = 500}, --  Level 5
+    {StartCash = 50000, InterestRate = 600, StartRep = 500}, --  Level 6
+    {StartCash = 50000, InterestRate = 700, StartRep = 500}, --  Level 7
+    {StartCash = 60000, InterestRate = 700, StartRep = 500}, --  Level 8
+    {StartCash = 60000, InterestRate = 800, StartRep = 500}, --  Level 9
+    {StartCash = 60000, InterestRate = 800, StartRep = 500}, --  Level 10
+    {StartCash = 70000, InterestRate = 900, StartRep = 500}, --  Level 11
+    {StartCash = 70000, InterestRate = 900, StartRep = 500}, --  Level 12
+    {StartCash = 70000, InterestRate = 900, StartRep = 500}, --  Level 12
   },
   popn = {
     [0] = {Month = 0, Change = 4}, -- Standard: 4 patients the first month.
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/base_config.lua
```diff
@@ -25,9 +25,12 @@ local configuration = {
   -----------------------------------------------------------
   --      New configuration values added in CorsixTH       --
   -----------------------------------------------------------
+  -- Interest rate and overdraft differential values will be divided by 10,000
```
lewri, line 28: Not your fault, more the original game's... but silly imo :)

CorsixTH/Lua/hospital.lua
```diff
@@ -32,18 +32,25 @@ function Hospital:Hospital(world, avail_rooms, name)
...
+  assert(level_config, "Cannot load level configuration.")
+  if level_config.towns and level_config.towns[level] then
+    balance = level_config.towns[level].StartCash
+    interest_rate = level_config.towns[level].InterestRate
+    reputation = level_config.towns[level].StartRep
+    overdraft_differential = level_config.towns[level].OverdraftDiff
+  elseif level_config.town then
+    balance = level_config.town.StartCash
+    interest_rate = level_config.town.InterestRate
+    reputation = level_config.town.StartRep
```
lewri, line 49: The TODO at 141 (old) is confusing here in regards to the change. Is the TODO saying start rep must always be 500? Or are we now honouring other start-reps?

CorsixTH/Lua/hospital.lua
```diff
@@ -612,6 +620,10 @@ function Hospital:afterLoad(old, new)
     end
   end
 
+  if old < 154 then
+    self.overdraft_interest_rate = self.interest_rate + 0.02
```
lewri, line 624: Could the overdraft differential be attempted to be got from the config instead? Though I'd also be happy with this if it's to play safe.

lewri (commented):
Only some clarifications from me, not against the setup per se.

CorsixTH/Lua/hospital.lua
tobylane: They don't exist yet. It could re read the level file, but I'd rather the less surprising thing of they continue with what they started.

CorsixTH/Lua/hospital.lua
tobylane: The latter, I was setting it at 600 for testing.

CorsixTH/Lua/hospital.lua
```diff
@@ -32,18 +32,25 @@ function Hospital:Hospital(world, avail_rooms, name)
...
-    elseif level_config.town then
-      balance = level_config.town.StartCash
-      interest_rate = level_config.town.InterestRate / 10000
-    end
+  local reputation_min = 0
+  local reputation_max = 1000
+  local reputation = 500
+  local interest_rate, overdraft_differential
+
+  assert(level_config, "Cannot load level configuration.")
```
lewri, line 40: Should we assert this? In map.lua if we fail to load a level config the base_config takes its place instead.

CorsixTH/Lua/base_config.lua
TheCycoONE: I suspect the original did this to avoid dealing with floating point. When it comes to multiplayer we may wish we had followed their example.

tobylane:
This takes the defaults out of hospital, where they were repeating the base config.

CorsixTH/Lua/base_config.lua
```diff
@@ -133,19 +137,19 @@ local configuration = {
...
-    {StartCash = 50000, InterestRate = 500}, --  Level 5
-    {StartCash = 50000, InterestRate = 600}, --  Level 6
-    {StartCash = 50000, InterestRate = 700}, --  Level 7
-    {StartCash = 60000, InterestRate = 700}, --  Level 8
-    {StartCash = 60000, InterestRate = 800}, --  Level 9
-    {StartCash = 60000, InterestRate = 800}, --  Level 10
-    {StartCash = 70000, InterestRate = 900}, --  Level 11
-    {StartCash = 70000, InterestRate = 900}, --  Level 12
-    {StartCash = 70000, InterestRate = 900}, --  Level 12
+    {StartCash = 40000, InterestRate = 100, StartRep = 500, OverdraftDiff = 200}, -- Level 1
```
Alberth289346, line 140: This line has 1 space in `-- Level` while the other have 2. As you're changing the line anyway, please correct that as well.

Alberth289346: That is, make it consistent, either by editing this line, or by editing all other lines :)

CorsixTH/Lua/hospital.lua
```diff
@@ -78,11 +80,12 @@ function Hospital:Hospital(world, avail_rooms, name)
   self.concurrent_epidemic_limit = level_config.gbv.EpidemicConcurrentLimit or 1
 
   -- Initial values
-  self.interest_rate = interest_rate
+  self.interest_rate = interest_rate / 10000
```
Alberth289346, line 83: This is weird, you have an interest rate, divide by 1000, and it's again an interest rate.
Obviously one of them is a factor 1000 higher than the other, so giving them both the same name is confusing.

Either `interest_rate` should have a different name imho, or you should apply the division at the line above where you copy the configuration value into the `interest_rate`, so you don't have to do that here.

CorsixTH/Lua/hospital.lua
```diff
@@ -78,11 +80,12 @@ function Hospital:Hospital(world, avail_rooms, name)
...
   self.inflation_rate = 0.045
+  self.overdraft_interest_rate = self.interest_rate + overdraft_differential / 10000
   self.salary_incr = level_config.gbv.ScoreMaxInc or 300
   self.sal_min = level_config.gbv.ScoreMaxInc / 6 or 50
-  self.reputation = 500
+  self.reputation = reputation
   self.reputation_min = 0
   self.reputation_max = 1000
```
Alberth289346, line 90: Shouldn't `self.reputation` be clamped against the stated bounds?
Likely stuff goes wrong when the reputation is outside expected bounds.

tobylane:
Fixed and ready.

lewri:
Looks fine here

PR #1854, opened 2021-04-19
[RDY] Service Quality affects treatment

### LINKED ISSUE #1829: Nurse can kill too readily
In the original the rate of killing a patient in Pharmacy was lower than that in CorsixTH -- such that it is even easier to kill your first patient at the hospital. It may be better to try and match the failure rate of drugs of the original to patient death.

I haven't looked into how treatment success is calculated yet but I'm thinking the calculation is just raw % effectiveness * diagnosis confidence.

lewri:
*Closes #1829*

- Service Quality now considered in treatment. As realistically the fatigue/skill etc or the doctor/nurse can affect effectiveness of a cure.
- Service Quality will only have a small impact cure_chance to not become unbalanced. Though now has a higher impact where cure_chance is lower, and less when higher.

Just a side note, inside staff_profile we actually set the midpoint as skill 0.55; should we copy that here too with service quality's midpoint?

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -301,8 +301,13 @@ function Patient:isTreatmentEffective()
   local cure_chance = self.hospital.disease_casebook[self.disease.id].cure_effectiveness
   cure_chance = cure_chance * self.diagnosis_progress
 
-  local die = self.die_anims and math.random(1, 100) > cure_chance
-  return not die
+  -- Service quality has a factor on cure chance +/- 10%
+  local room = self:getRoom()
+  local base, divisor = 0.9, 5
+  local service_factor = base + (room:getStaffServiceQuality() / divisor)
+  cure_chance = cure_chance * service_factor
+
+  return (cure_chance >= math.random(1,100))
 end
 
 --! Change patient internal state to "cured".
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -303,9 +303,11 @@ function Patient:isTreatmentEffective()
 
   -- Service quality has a factor on cure chance
   local room = self:getRoom()
-  local min, base, divisor = 20, -0.1, 5
-  local service_base = math.max(100 - cure_chance, min)
-  local service_factor = base + (room:getStaffServiceQuality() / divisor)
+  local min_impact = 20
+  local service_base = math.max(100 - cure_chance, min_impact)
+
+  local scale = 0.2 -- Quality scaled to +-10%
+  local service_factor = (room:getStaffServiceQuality() - 0.5) * scale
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -301,8 +301,13 @@ function Patient:isTreatmentEffective()
   local cure_chance = self.hospital.disease_casebook[self.disease.id].cure_effectiveness
   cure_chance = cure_chance * self.diagnosis_progress
 
-  local die = self.die_anims and math.random(1, 100) > cure_chance
-  return not die
+  -- Service quality has a factor on cure chance +/- 10%
+  local room = self:getRoom()
+  local base, divisor = 0.9, 5
+  local service_factor = base + (room:getStaffServiceQuality() / divisor)
+  cure_chance = cure_chance * service_factor
```
Alberth289346, line 308: So when I have a bad treatment the service quality has little impact, while if I have a good treatment the service quality has a lot of impact.
So effectively it's shortening the last part of the trajectory to reach an ultimate treatment with no deaths occurring.

Have you considered affecting the chance  by adding some of the remaining chance instead? That is, use `1 - cure_chance` as multiplication factor for the service_factor (and adding that to the chance)?
Then service quality has a larger effect to a bad treatment, and not much impact on a good treatment.

lewri: You'd need 100 - cure_chance. But also cure chance can be >100

Alberth289346: `math.max(100 - cure_chance, 0)` thus

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -301,11 +301,12 @@ function Patient:isTreatmentEffective()
...
 
-  -- Service quality has a factor on cure chance +/- 10%
+  -- Service quality has a factor on cure chance
   local room = self:getRoom()
-  local base, divisor = 0.9, 5
+  local min, base, divisor = 20, -0.1, 5
+  local service_base = math.max(100 - cure_chance, min)
   local service_factor = base + (room:getStaffServiceQuality() / divisor)
-  cure_chance = cure_chance * service_factor
+  cure_chance = cure_chance + (service_base * service_factor)
```
Alberth289346:
``` lua
  -- Amount of cure chance where service quality makes a difference.
  local min_impact = 20
  local service_base = math.max(100 - cure_chance, min_impact)

  local scale = 0.2 -- Quality scaled to [-scale / 2, +scale / 2]
  local service_factor = (room:getStaffServiceQuality() - 0.5) * scale
  cure_chance = cure_chance + (service_base * service_factor)
```

Alberth289346:
Looks fine to me

lewri:
Ready for cyco :)
When #1849 is merged this one will also balance correctly for service quality.

PR #1863, opened 2021-04-30
[RDY] Carry hospital policies to the next hospital

tobylane:
- Use the previous hospitals' policies in the next one when changing level

I always set the policies to the same, regardless of the level. It makes sense to have them reused..

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -626,22 +626,7 @@ function App:loadLevel(level, difficulty, level_name, level_file, level_intro, m
     return
   end
   -- If going from another level, save progress.
-  local carry_to_next_level
-  if self.world and self.world.campaign_info then
-    carry_to_next_level = {
-      world = {
-        room_built = self.world.room_built,
-        campaign_info = self.world.campaign_info,
-        },
-      hospital = {
-        player_salary = self.ui.hospital.player_salary,
-        message_popup = self.ui.hospital.message_popup,
-        handyman_popup = self.ui.hospital.handyman_popup,
-        hospital_littered = self.ui.hospital.hospital_littered,
-        has_seen_pay_rise = self.ui.hospital.has_seen_pay_rise,
-      },
-    }
-  end
+  local campaign_data = self.world and self.world:getCampaignData()
 
   -- Make sure there is no blue filter active.
   self.video:setBlueFilterActive(false)
@@ -668,8 +653,8 @@ function App:loadLevel(level, difficulty, level_name, level_file, level_intro, m
   self.world:setUI(self.ui) -- Function call allows world to set up its keyHandlers
 
   -- Now restore progress from previous levels.
-  if carry_to_next_level then
-    self.world:initFromPreviousLevel(carry_to_next_level)
+  if campaign_data then
+    self.world:setCampaignData(campaign_data)
   end
 end
 
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -2392,3 +2392,25 @@ end
 function Hospital:changeValue(changeValue)
   self.value = self.value + changeValue
 end
+
+--! Collect the hospital level settings relevant for the next hospital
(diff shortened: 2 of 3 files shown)


Alberth289346:
I wondered if some of the values shouldn't be in the configuration settings, but not sure.

Not your fault, but `self.ui.hospital` is wrong imho. I believe that `ui.hospital` should the hospital that can be modified by the UI (if we manage a clean separation, two different players could co-operatively in a single hospital in theory).
Here the UI isn't even involved in running the code. Use `self.world` for finding the hospital instead.

Also not your fault, but this code is the "insert tea into stomach" solution instead of giving a person a cup of tea. External entity "app" digs magic values from the world and the hospital, and insert them again in the next instance.
It may look simple, but the huge cost is that world and hospital cannot move or change such values without breaking this code.
(In short, `App` shouldn't need to know internal data layout of world and hospital, let each object handle that by itself. In that way, you can modify object data without needing to change code of them in other objects.)

The solution to this is to ask both world and hospital for campaign data that they think should be rescued with a function, and after setting up the new world and hospital provide the rescued data again to the new world and the new hospital so they can integrate the values to where-ever they store it.

Finally, for bonus points, include all hospitals. CTH has 4 hospitals, I see no reason why only one of them should have the option of copying over campaign data, even if the other 3 currently don't have useful data to transfer.

tobylane:
World is unloaded, then loaded. It's taking a tea order from one customer and giving it to the next. The order has to be held by App.

https://github.com/CorsixTH/CorsixTH/blob/81db3338d89df253e85fc6b7ace9d937092dafaa/CorsixTH/Lua/dialogs/fullscreen/fax.lua#L208
Much of this section, plus the carry_to_next_level, could be a new Hospital function for moving between levels in a deliberate progression? Hospital:advanceLevel which uses app.carry_to_next_level?

Alberth289346:
Sorry, I wasn't clear enough, apparently.

I understand that world and everything in it disappears, and is created again. So indeed, world or anything inside it cannot hold the data that needs to be transferred. App is the one and only solution for that, and that is fine.

What bother me is that App is doing code like `guess_cure = self.ui.hospital.policies["guess_cure"],` to extract the tea from the previous owner, and then `for key, value in pairs(tab) do  hosp.policies[key] = value  end` into the stomach of the new owner.

When the `policies` get moved or renamed in the hospital code, and you forget about App, things break and that's not desired. To avoid that, you have to search all files for names that look like they could belong to hospital variables, verify that (and don't make a mistake!), and possibly change it too. Lot of work, lots of room for making an error.

The better fix is not to export variable names outside the owning object. You do that by not accessing variables of another object directly. Instead use a function call.

What App should do here is ask politely through a function, something like `campaign_data = self.world.getCampaignData()`.
``` lua
function World:getCampaingData()
  return {
     world = {
       room_built = self.room_built,
       campaign_info = self.campaign_info,
       debug_disable_salary_raise = self.debug_disable_salary_raise,
      },
    hospital = self.getPlayerHospital():getCampaignData()  -- And Hospital:getCampaignData() does something similar for the hospital data
  }
end
```
Each object that owns part of the campaign data copies its values into a new table that is returned to the caller. World also has a coordinating role with hospitals, since it manages the hospitals too, but it doesn't know what is in the `hospital` field. Likewise, App gets a data value from the request, but doesn't know what's in it, nor does it look inside.

Restoring the data after creating a new world works the same but in the other direction (something like `self.world:restoreCampaignData(campaign_data)` in App).

Now world or the hospital can freely move or change campaign data. It only needs to be reflected in the rescue and restore functions like the `World:getCampaingData()` (if world is modified). There are no campaign variable names of the world or the hospital in App. If the hospital adds new fields to the data, neither App nor World have to be changed, as both never look inside the data of the hospitals.

So instead of reading and writing variables directly from/to other objects, let each object handle the read or write by itself, by means of a function call.

Alberth289346:
LGTM, hope you like it too

tobylane:
I do, and thanks for the detailed help. Ready.

TheCycoONE:
I like it too.

Issue #1416, opened 2018-06-02
Add support for Flatpak packages


LeandroStanger:

TheCycoONE:
I am open to the idea, but not willing to take it on myself. Do you have any experience?

LeandroStanger:
@nedrichards @TingPing could help?

f35f0ef9d0e827dae86552d3899f78fc:
I've never made a Flatpak before, or contributed to this project, but I think I could learn how. I will look into this.

LeandroStanger:
Publish the app on Flathub
https://github.com/flathub/flathub/wiki/App-Submission

TheCycoONE:
So is the first step to fork flathub under CorsixTH and create the branch? I can do that. I would like someone else to write the manifest file. The App ID should be com.corsixth.CorsixTH

TheCycoONE:
https://github.com/CorsixTH/flathub/tree/com.corsixth.CorsixTH

rmnvgr:
Hello,

I've written a manifest for CorsixTH: https://github.com/rmnvgr/flathub/tree/com.corsixth.CorsixTH

If it looks good to you, I'll create a PR to make it available on Flathub.

TheCycoONE:
It's not clear to me if it's possible but it would be nice to provide wayland support if available. I see the fallback-x11 socket in the documentation, but I can't tell how you would toggle SDL_VIDEODRIVER=wayland based off that.

Certainly not a requirement if it's not possible.

TheCycoONE:
One other note, since my prior comments on this ticket our appstream metadata id and desktop id was changed to com.corsixth.corsixth, in accordance with the guidelines: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic so the flatpak should have the same application id.

rmnvgr:
> It's possible to trim down ffmpeg quite a bit, if that's a concern. We don't need any of the binaries, and for now we only use the smacker and smackaud decoders, and smacker demuxer.

I was already stripping the binaries, I didn't know exactly which components where needed, but thanks to your input I've edited the manifest to only build the necessary components.

> It's not clear to me if it's possible but it would be nice to provide wayland support if available. I see the fallback-x11 socket in the documentation, but I can't tell how you would toggle SDL_VIDEODRIVER=wayland based off that.

[...]

PR #1871, opened 2021-05-10
Metainfo: Provide the old app ID

rmnvgr:
Providing the old app ID prevents app stores from showing duplicate entries if some sources still use it.


lewri:
#1416

PR #1874, opened 2021-05-10
Boiler breakdown advice is local player only

tobylane:
- Add PlayerHospital:makeAnnouncement, a copy of giveAdvice adapted for announcements
- Fix plots being unbuyable when walls are transparent
- Add checks for local hospital before telling user
- Move no receptionist advice away from check for receptionists

### DIFF
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -750,6 +750,27 @@ function GameUI:onMouseWheel(x, y)
   return UI.onMouseWheel(self, x, y)
 end
 
+--! Announcements to the player.
+--!param msgs (array of string) Messages to select from.
+--!param priority (optional valid AnnouncementPriority selection) Priority of announcement
+--!param chance_to_play (optional float in range (0, 1]) Fraction of times that the
+--    call actually says something.
+--!return (boolean) Whether a message was given to the user.
+function GameUI:playRandomAnnouncement(msgs, priority, chance_to_play, played_callback, played_callback_delay)
+  local max_rnd = #msgs
+  if chance_to_play and chance_to_play > 0 and chance_to_play < 1 then
+    -- Scale by the fraction.
+    max_rnd = math.floor(max_rnd / chance_to_play)
+  end
+
+  local index = (max_rnd == 1) and 1 or math.random(1, max_rnd)
+  if index <= #msgs then
+    self:playAnnouncement(msgs[index], priority, played_callback, played_callback_delay)
+    return true
+  end
+  return false
+end
+
 function GameUI:playAnnouncement(name, priority, played_callback, played_callback_delay)
   self.announcer:playAnnouncement(name, priority, played_callback, played_callback_delay)
 end
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -20,8 +20,6 @@ SOFTWARE. --]]
 
 corsixth.require("announcer")
 
-local AnnouncementPriority = _G["AnnouncementPriority"]
-
 class "Hospital"
 
 ---@type Hospital
@@ -777,22 +775,6 @@ function Hospital:isInHospital(x, y)
   local flags = self.world.map.th:getCellFlags(x, y)
   return flags.hospital and flags.owner == self:getPlayerIndex()
 end
-function Hospital:coldWarning()
-  local announcements = {
-    "sorry002.wav", "sorry004.wav",
-  }
-  if announcements and self:isPlayerHospital() then
-    self.world.ui:playAnnouncement(announcements[math.random(1, #announcements)], AnnouncementPriority.Normal)
-  end
-end
-function Hospital:hotWarning()
-  local announcements = {
-    "sorry003.wav", "sorry004.wav",
-  }
-  if announcements and self:isPlayerHospital() then
-    self.world.ui:playAnnouncement(announcements[math.random(1, #announcements)], AnnouncementPriority.Normal)
-  end
-end
 
 --! Decide how many days the hospital functions within specification.
 --!return (int) Number of disaster-free days in the hospital.
@@ -821,16 +803,13 @@ function Hospital:boilerBreakdown(broken_heat)
   heat_vars.boiler_repair_count = math.random(10, 30)
   heat_vars.heating_broke = true
 
-  -- Only show the message when relevant to the local player's hospital.
-  if self:isPlayerHospital() then
-    if heat_vars.radiator_heat == 0 then
-      self.world.ui.adviser:say(_A.boiler_issue.minimum_heat)
-      self:coldWarning()
-    else
-      self.world.ui.adviser:say(_A.boiler_issue.maximum_heat)
-      self:hotWarning()
-    end
-  end
+  -- Warn the player of the boiler's breakdown
+  self:adviseBoilerBreakdown(broken_heat)
+end
+
+--! Select a relevant message to be displayed to the user
+function Hospital:adviseBoilerBreakdown()
+  -- Nothing to do, override in a sub-class.
 end
 
 --! Boiler broke down and work is done to get it fixed.
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -378,6 +378,19 @@ function PlayerHospital:adviseDiscoverDisease(disease)
   end
 end
 
+--! Select a relevant message to be displayed to the user
+--!param broken_heat (0 or 1) Boiler output due to being broken.
+function PlayerHospital:adviseBoilerBreakdown(broken_heat)
+  local ui = self.world.ui
+  if broken_heat == 0 then
+    ui.adviser:say(_A.boiler_issue.minimum_heat)
+    ui:playRandomAnnouncement({ "sorry002.wav", "sorry004.wav" })
+  else
+    ui.adviser:say(_A.boiler_issue.maximum_heat)
+    ui:playRandomAnnouncement({ "sorry003.wav", "sorry004.wav" })
+  end
+end
+
 --! Called at the end of each day.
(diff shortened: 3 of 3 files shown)


CorsixTH/Lua/hospital.lua
```diff
@@ -774,22 +774,6 @@ function Hospital:isInHospital(x, y)
...
-  end
-end
-function Hospital:hotWarning()
-  local announcements = {
-    "sorry003.wav", "sorry004.wav",
-  }
-  if announcements and self:isPlayerHospital() then
-    self.world.ui:playAnnouncement(announcements[math.random(1, #announcements)], AnnouncementPriority.Normal)
-  end
-end
```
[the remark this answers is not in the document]
tobylane: They're also local player only, https://github.com/CorsixTH/CorsixTH/blob/ae4cf31be63008bc9ed4c9dd265f5f443e695baf/CorsixTH/Lua/hospital.lua#L822

Do you want stubs to be overridden?

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -367,6 +366,40 @@ function PlayerHospital:warnForLongQueues()
...
 
+--! Announcements to the player.
+--!param msgs (array of string) Messages to select from.
+--!param rnd_frac (optional float in range (0, 1]) Fraction of times that the
+--    call actually says something.
+--!param priority (optional - Low, Normal, High, Critical) Priority of announcement
+--!return (boolean) Whether a message was given to the user.
+function PlayerHospital:makeAnnouncement(msgs, rnd_frac, priority)
+  if priority then
+    priority = AnnouncementPriority.priority
```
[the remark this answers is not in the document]
tobylane: Now defaults to Normal if there's a typo.

CorsixTH/Lua/hospital.lua
```diff
@@ -819,15 +803,7 @@ function Hospital:boilerBreakdown(broken_heat)
...
-  if self:isPlayerHospital() then
-    if heat_vars.radiator_heat == 0 then
-      self.world.ui.adviser:say(_A.boiler_issue.minimum_heat)
-      self:coldWarning()
-    else
-      self.world.ui.adviser:say(_A.boiler_issue.maximum_heat)
-      self:hotWarning()
-    end
-  end
+  if self:isPlayerHospital() then self:warnBoilerBreakdown(broken_heat) end
```
Alberth289346, line 806: Don't call functions that don't exist in a class, ever.

It's fine to create empty a place-holder function in this class, and then re-implement it in the player hospital. That also allows you to remove the condition. Conceptually always warn, just some versions of "warn" don't do anything useful.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -367,6 +366,40 @@ function PlayerHospital:warnForLongQueues()
...
+
+  local index = (max_rnd == 1) and 1 or math.random(1, max_rnd)
+  if index <= #msgs then
+    self.world.ui:playAnnouncement(msgs[index], AnnouncementPriority[priority])
+    return true
+  end
+  return false
+end
+
+function PlayerHospital:warnBoilerBreakdown(broken_heat)
```
Alberth289346, line 393: Needs function documentation.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -382,6 +415,9 @@ function PlayerHospital:onEndMonth()
   -- Advise the player on cash flow.
   if self:hasStaffedDesk() then
     self:monthlyAdviceChecks()
+  else
+    local today = self.world:date()
+    self:checkReceptionAdvice(today:monthOfYear(), today:year())
```
Alberth289346, line 420: The function doesn't do anything if you have a staffed desk, so it already takes care of the pre-condition by itself. No need to do that here as well.

CorsixTH/Lua/hospital.lua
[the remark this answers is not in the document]
Alberth289346: The point is that by making calls to non-existing functions you're relying on always using a derived class. In other languages you can express that, in Lua you cannot.

I know there is some protection now, but code isn't stable, so somewhen an error will be introduced in the conditions and a call will be made, leading to a crash. Avoid trying to live on the edge, we make much more errors in code than you think.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -367,6 +366,40 @@ function PlayerHospital:warnForLongQueues()
   end
 end
 
+--! Announcements to the player.
+--!param msgs (array of string) Messages to select from.
+--!param rnd_frac (optional float in range (0, 1]) Fraction of times that the
+--    call actually says something.
+--!param priority (optional string - Low, Normal, High, Critical) Priority of announcement
+--!return (boolean) Whether a message was given to the user.
+function PlayerHospital:makeAnnouncement(msgs, rnd_frac, priority)
```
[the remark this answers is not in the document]
tobylane: I think of it as pre UI where it selects the specific message to be shown. Does that sound right?

tobylane:
Reduced to the boiler breakdown, including adding GameUI:makeAnnouncement. Removed isPlayerHospital checks in favour of PlayerHospital functions and Hospital stubs.

tobylane:
Using the prefix advise for consistent style. Ready.

CorsixTH/Lua/game_ui.lua
```diff
@@ -750,6 +752,30 @@ function GameUI:onMouseWheel(x, y)
...
 
+--! Announcements to the player.
+--!param msgs (array of string) Messages to select from.
+--!param rnd_frac (optional float in range (0, 1]) Fraction of times that the
+--    call actually says something.
+--!param priority (optional string - Low, Normal, High, Critical) Priority of announcement
+--!return (boolean) Whether a message was given to the user.
+function GameUI:makeAnnouncement(msgs, rnd_frac, priority)
+  if not priority or not AnnouncementPriority[priority] then priority = "Normal" end
+  assert(priority, "Invalid announcement priority")
```
Alberth289346, line 763: One line above you say `if  not priority .... then priority = "Normal" end`
How can this assert ever trigger?

tobylane: I'm not sure how it could trigger, but I think that about other asserts too.

Alberth289346: > I'm not sure how it could trigger, but I think that about other asserts too.

All asserts are for protecting against things that are not supposed to be happening. The point is indeed what you say, we mostly don't know if they can happen. In such a case the `assert` is kept, mostly as a safety measure. (Basically, it's too much work or too complicated to proof the assert never triggers, keeping it is cheaper.)

However, if you can easily proof that you have code equivalent to `assert(true)`, you should remove it, as it protects against an impossibility, and only clutters the code.

There are 3 cases here:
- `priority` evaluates to `false` or `nil`: In this case, the `if` jumps into the `then`, and sets the variable to a string which does not evaluate to `false` or `nil`. This case thus doesn't trigger the assert.
- `priority` evaluates to `true`, and `AnnouncementPriority[priority]` evaluates to `false`/`nil`. In this case too, the `if` fixes the value like the previous case, assert again doesn't trigger.
-  `priority` evaluates to `true`, and `AnnouncementPriority[priority]` evaluates to `true`. The `if` doesn't do anything. Since `priority` stays the same, the assert doesn't trigger.

So for all possible cases that I see, the assert doesn't trigger. My conclusion therefore is that the assert isn't useful.

Note that this can be decided only because you can here locally reason for assert not triggering. This is quite cheap to do. If code becomes more complicated, you can reach the point where it is not clear if a problem can happen. In such a case, you would add the assert again at that time.

CorsixTH/Lua/game_ui.lua
```diff
@@ -750,6 +752,29 @@ function GameUI:onMouseWheel(x, y)
   return UI.onMouseWheel(self, x, y)
 end
 
+--! Announcements to the player.
+--!param msgs (array of string) Messages to select from.
+--!param rnd_frac (optional float in range (0, 1]) Fraction of times that the
+--    call actually says something.
+--!param priority (optional string - Low, Normal, High, Critical) Priority of announcement
+--!return (boolean) Whether a message was given to the user.
+function GameUI:makeAnnouncement(msgs, rnd_frac, priority)
```
TheCycoONE, line 761: This name seems wrong; you're trying to distinguish from playAnnouncement below, but the difference isn't that the announcement is made, it's that there is an element of chance involved.

rnd_frac should be after priority, matching the name, priority order of the function it delgates to. Either msgs or names is named wrong because msgs is an array of names.

Perhaps `playRandomAnnouncement(names, priority, played_callback, played_callback_delay, chance_to_play)`

CorsixTH/Lua/game_ui.lua
```diff
@@ -750,6 +752,29 @@ function GameUI:onMouseWheel(x, y)
...
+  if not priority or not AnnouncementPriority[priority] then priority = "Normal" end
```
TheCycoONE, line 762: Not clear why `playAnnouncement` takes a priority and this takes a key into AnnouncementPriority. It's not used that way anyway, so let's assume it's already an AnnouncementPriority that gets passed in.

TheCycoONE (changes requested):

CorsixTH/Lua/game_ui.lua
tobylane: I was planning to use it that way once its possible, eg [epidemic start/end](https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/epidemic.lua#L228). I'm using the key to avoid needing to include the AnnouncementPriority at the top of each file where its used (or where you wanted to set any later parameters).

tobylane:
Agreed and done on both.

TheCycoONE:
Please resolve merge conflicts. Then I think this is ready.

tobylane:
Ready.

PR #1885, opened 2021-05-20
[RDY] Fix the search for Theme Hospital installs on Windows

TheCycoONE:
The macos directories were inserting nulls in the array which ipairs
would stop on without iterating through all options. Fixed by replacing
with empty strings. \\ directories in the search caused considerable
pause in windows and were invalid, so skip checking them. Add up to date
"GOG Games" folder which has been the default for awhile now.

This was a somewhat recent regression. Please test non-Windows OSs to ensure I didn't break anything (remove your config.txt and run, with/without a hospital directory that can be located.)

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1253,19 +1253,26 @@ function App:checkInstallFolder()
     -- then linux Filesystem Hierarchy Standard, then Windows Program Files
     -- mac_app_dir is the macOS app base directory named CorsixTH.app
     local mac_app_dir = debug.getinfo(1).short_src:match("(.*)/Contents/.")
-    local user_dir = os.getenv("HOME") or os.getenv("HOMEPATH") or ""
-    local possible_locations = { user_dir, user_dir .. pathsep ..  "Documents",
+    local user_dir = os.getenv("HOME") or os.getenv("HOMEPATH")
+    local possible_locations = {
+      user_dir,
+      user_dir and (user_dir .. pathsep ..  "Documents"),
       select(1, corsixth.require("config_finder")):match("(.*[/\\])"):sub(1, -2),
-      mac_app_dir, mac_app_dir and mac_app_dir:match("(.*)/.*%.app"),
+      mac_app_dir,
+      mac_app_dir and mac_app_dir:match("(.*)/.*%.app"),
       "/Applications/Theme Hospital.app/Contents/Resources/game/Theme Hospital.app/" ..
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/app.lua
```diff
@@ -1254,20 +1254,27 @@ function App:checkInstallFolder()
...
+      "/usr/share/games/corsix-th",
+      "/usr/local/share/games/corsix-th",
+      os.getenv("ProgramFiles") or "",
+      os.getenv("ProgramFiles(x86)") or "",
+      [[C:]], [[D:]], [[E:]], [[F:]], [[G:]], [[H:]] }
     local possible_folders = { "ThemeHospital", "Theme Hospital", "HOSP", "TH97",
-      [[GOG.com\Theme Hospital]], [[Origin Games\Theme Hospital\data\Game]] }
+      [[GOG.com\Theme Hospital]], [[GOG Games\Theme Hospital]],
+      [[Origin Games\Theme Hospital\data\Game]] }
     for _, dir in ipairs(possible_locations) do
```
lewri, line 1273: Couldn't these just be changed to normal pairs if ipairs was finishing prematurely?

tobylane:
This works, and Lewri's suggestion (with the extra or "" taken out) works too.

I'm surprised that I must have introduced this in 1858, because I had the reverse situation there in testing.

TheCycoONE:
Removing the or "" and changing to ipairs means I'd have to do something different about userdir (changing it to nil and verifying it exists before concatenating with Documents) - which I suppose isn't the end of the world /Documents is unlikely where we want to look anyway.
Also loses deterministic search order - which probably isn't too helpful anyway.
Sure.

What I'd like is OS dependent paths, but I was trying to be conservative in my changes nearing release for something I wanted in the release.

TheCycoONE:
Is there any problems with the drive letters on OSs MacOS?

tobylane:
> Is there any problems with the drive letters on MacOS?

None

> Removing the or "" and changing to ipairs means I'd have to do something different about userdir (changing it to nil and verifying it exists before concatenating with Documents)

If I change HOME to HME then the home directory isn't checked, and /Documents (a root folder) is checked. I see, or "" does fix it in all places.

tobylane (approved):

TheCycoONE:
Updated to use pairs and nil values as suggested.

TheCycoONE:
Build is failing for unrelated reasons, see #1886

PR #1886, opened 2021-05-21
[RDY] Bump clang version to keep up with GitHub

TheCycoONE:
GitHub removed clang-9 as part of their last 3 version policy so we are forced to bump.

The documentation seemed mixed on support for 12. In my tests it wasn't available for clang-tidy so I've opted to go with 11 (which is their new default as well.)


TheCycoONE:
See https://sourceware.org/bugzilla/show_bug.cgi?id=15366 for an explanation of the removed line clang-tidy had complained about.

mugmuggy (approved):

PR #1889, opened 2021-05-21
[RDY] Make computing price impact warnings local player specific

tobylane:
- Only advise the local hospital of the price level impacts.

### DIFF
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -392,6 +392,23 @@ function PlayerHospital:adviseCannotAffordPlot()
   self.world.ui.adviser:say(_A.warnings.cannot_afford_2, true, true)
 end
 
+--! Tell the player, through the advisor, about the impact of casebook prices
+--!param judgment (string - under, over, fair) The judgment of the price,
+-- from Hospital:computePriceLevelImpact
+--!param name (string) The name of the casebook entry of the diagnosis or disease
+function PlayerHospital:advisePriceLevelImpact(judgment, name)
+  local message
+  if judgment == "under" then
+    message = _A.warnings.low_prices:format(name)
+  elseif judgment == "over" then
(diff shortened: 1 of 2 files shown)


lewri:
Is this something that should actually be moved into player_hospital? (maybe out of scope though).

TheCycoONE:
Sort of. The price impact belongs where it is. There should be a stub method for advising for the hospital, and one that calls   self.world.ui.adviser:say in player_hospital. Then all these isPlayerHospital checks go away.

And yes that should replace this PR.

tobylane:
Rewrote it that way.

tobylane:
Ready.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -398,6 +398,24 @@ function PlayerHospital:onEndMonth()
...
+--!param judgment (string - under, over, fair) The judgment of the price,
+-- from Hospital:computePriceLevelImpact
+--!param name (string) The name of the casebook entry of the diagnosis or disease
+function PlayerHospital:advisePriceLevelImpact(judgment, name)
+  local message
+  if judgment == "under" then
+    message = _A.warnings.low_prices:format(name)
+  elseif judgment == "over" then
+    message = _A.warnings.high_prices:format(name)
+  elseif judgment == "fair" then
```
Alberth289346, line 411:
```suggestion
  else
    assert(judgment == "fair")
```
This ensures the code dies on typos in the `judgment` value.

TheCycoONE: Stylistically I prefer
```lua
if foo == "a" then
   ---
elseif foo == "b" then
  ---
elseif foo == "c" then
  ---
else
  error("foo must be a, b, or c")
end
```
but I suppose it doesn't matter.

lewri (approved):
Could you resolve the conflict then it can be merged.

PR #1891, opened 2021-05-22
Emscripten PoC

alexandre-lavoie:
*Fixes 1883*

# Summary

This is a PoC for building CorsixTH for the web with Emscripten.

## Building

To build using this PoC, you need to get the [nix packager manager](https://nixos.org/download.html#nix-quick-install) (OS should not be required). Once installed, you should be able to build with:

```
nix-build web.nix
``` 

This should generate a `result` folder containing all you need for the web build. You can then host the build using:

```
cd result/pkg
python3 -m http.server
```

You should then be able to find the game at: http://localhost:8000/ .

## Issues

The game is able to compile, but renders only a black screen (if you are lucky enough, you can get a single frame). From debugging, it does seem to actually be running, but visuals fail. I am pretty sure it's due to pthread and being required to use `emscripten_set_main_loop` to loop to keep in sync with the web JS frontend. From what I understand, it would be favorable to have a `do_frame()` method that would be called by JS every "JS frame" (I think the canvas is being flooded currently). Given I am not familiar with the C <-> Lua interactions, I would need people working on core to help!


TheCycoONE:
I will warn that in the past EA has been very sensitive to web ports. We must not host any game assets.

alexandre-lavoie:
> I will warn that in the past EA has been very sensitive to web ports. We must not host any game assets.

I'm considering a drag-and-drop of the HOSP folder (and somehow saving the folder in local browser cache). We'd only publicly host the engine.

alexandre-lavoie:
I also just managed to get in-browser debugging following: https://developer.chrome.com/blog/wasm-debugging-2020/

I'm getting a Segfault with my new changes loop - I'm not sure if it's on the core or one of the Lua libraries failing.

debug

lewri:
Note for this to be tidied

``CorsixTH/CMakeLists.txt:131: Line ends in whitespace [whitespace/eol]``

flackr:
If supported in emscripten, we might be able to use makecontext / setcontext to preserve the current context from the main game loop as we return control to the browser until the next frame at which point we can swapcontext back to that game context.

lewri:
@alexandre-lavoie Did you plan to continue with this one?

alexandre-lavoie:
@lewri I do not plan to continue on this.

PR #1892, opened 2021-05-22
[RDY] Avoid using recently depreacted FFMpeg functionality

### LINKED ISSUE #1890: FFmpeg AVPacketList deprecated
Some recent deprecations in ffmpeg 4.4 affect us:

```
2021-03-17 - f7db77bd87 - lavc 58.133.100 - codec.h
          Deprecated av_init_packet(). Once removed, si
          no longer be a part of the public ABI.
          Deprecated AVPacketList.
```
We will need to determine the replacements for these structures and methods, before they release a version where our movies don't work.

TheCycoONE:
*Fixes #1890*

- The only call to the now deprecated av_init_packet() was specific to an old path, so ifndef that out unless we are on very old FFMpeg versions
- The use of AVPacketList is still fairly critical and my attempts to replace it with modern collections tended to spiral. As a short term fix for the 0.65 release, implement our own struct with the same layout as AVPacketList as a drop in replacement.

### DIFF
--- a/CorsixTH/Src/th_movie.cpp
+++ b/CorsixTH/Src/th_movie.cpp
@@ -311,7 +311,7 @@ void av_packet_queue::push(AVPacket* pPacket) {
   }
 #endif
 
-  AVPacketList* pNode = (AVPacketList*)av_malloc(sizeof(AVPacketList));
+  th_packet_list* pNode = (th_packet_list*)av_malloc(sizeof(th_packet_list));
   pNode->pkt = *pPacket;
   pNode->next = nullptr;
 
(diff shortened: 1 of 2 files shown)


CorsixTH/Src/th_movie.cpp
```diff
@@ -311,7 +311,7 @@ void av_packet_queue::push(AVPacket* pPacket) {
   }
 #endif
 
-  AVPacketList* pNode = (AVPacketList*)av_malloc(sizeof(AVPacketList));
+  th_packet_list* pNode = (th_packet_list*)av_malloc(sizeof(th_packet_list));
```
Alberth289346, line 314: shouldn't this do `static_cast<th_packet_list *>(av_malloc(sizeof(th_packet_list));` ?

TheCycoONE: Yeah, this whole class has a lot more cisms than it should.

Alberth289346: given your intentions, I am fine if you just merge this as-is

PR #1893, opened 2021-05-23
Add 0.65 release notes

lewri:
Add release notes to changelog, initial drafting. No date input yet.

Feel free to cherrypick/add/adapt as seem fit.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,41 +2,39 @@
 Version 0.65 - released xxxx 2021
 -------------------------------------------------------------------------------
 # New Features/Enhancements
-* Support for Lua 5.4.2+! (NB: 5.4.0 and 5.4.1 are not recommended)
-* CMake: Support for version 3.5+
 * CorsixTH is now available as a Flatpak! (thanks to @rmnvgr !)
-* DMG files are now supported as disk images for Theme Hospital data
+* On MacOS DMG files are now supported as disk images for Theme Hospital data
 * Hospital policies will now carry over to your next level in the campaign
-* You can now remove destroyed rooms for a fee (custom option)
-* Room information dialogs (first built) can be turned off in the configuration file
+* You can now remove destroyed rooms for a fee (config option)
+* The information dialogs that appear when a room is first built can be
+  disabled in the configuration file
 * Fire extinguishers in rooms now have a chance to save a smoking machine
 * More adviser messages added
 * New trophy added
 * Windows and room size now impact staff happiness
-* Roujin's challenge is back
+* Roujin's challenge cheat is back
 
 # Changes
-* Emergencies of unknown diseases will no longer occur
+* Emergencies with unknown diseases will no longer occur
 * The VIP system has been overhauled to provide more balanced ratings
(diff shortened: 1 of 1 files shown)


TheCycoONE:
GitHub changed the encoding to utf-8, but I'm ok with that.

changelog.txt
```diff
@@ -1,3 +1,55 @@
...
+Version 0.65 - released xxxx 2021
+-------------------------------------------------------------------------------
+# New Features/Enhancements
+* CorsixTH is now available as a Flatpak! (thanks to @rmnvgr !)
+* On MacOS DMG files are now supported as disk images for Theme Hospital data
+* Hospital policies will now carry over to your next level in the campaign
+* You can now remove destroyed rooms for a fee (config option)
+* The information dialogs that appear when a room is first built can be
+  disabled in the configuration file
+* Fire extinguishers in rooms now have a chance to save a smoking machine
```
Alberth289346, line 11: + `from exploding` ?

changelog.txt
```diff
@@ -1,3 +1,55 @@
...
+* More adviser messages added
```
Alberth289346, line 12: You can't have less messages added, can you? :) 

More general could be 'Extended advice topics of the advisor', or 'Advisor gives more advice'.
Nicer but more work is to list what the new advice is about.

changelog.txt
lewri: It is now :)

changelog.txt
```diff
@@ -1,3 +1,55 @@
...
+* New trophy added
```
Alberth289346, line 13: 'New' can be interpreted as 'one more of the original game' and as 'a completely new one', you may want to distinguish between those.

Likely previous entry, listing the one new trophy would be nicer.

(and in a sense more useful for history if we want to find back when "foo trophy" was introduced)

changelog.txt
```diff
@@ -1,3 +1,55 @@
...
+* Greek language added! (thanks to dimitrios-iliadis!)
+* Updates to Brazilian-Portuguese language
+* Updates to English language
+* Fixes to Chinese unicode encoding
+
+# Bug Fixes
+* Issues scheduling emergencies have been resolved
+* Games will no longer crash on load if the jukebox was open
+* Games will no longer crash because of an empty music folder
+* Fixed a chance of Doctors getting stuck in a queue after editing a room
```
Alberth289346, line 40: `Fixed a possibility of Doctors getting stuck in a queue after editing a room` ?

changelog.txt
```diff
@@ -1,3 +1,55 @@
...
+* Fixed an issue where the game could continue scrolling on load
```
Alberth289346, line 41: `A game now stops scrolling after loading` ?

Alberth289346:
hmm, too late :(

lewri:
Not fully lost, there will be tweaks before release I imagine =]

TheCycoONE:
Not too late @Alberth289346. Too late is final release. I even made more changes between merging this and the release page that need incorporated.

PR #1894, opened 2021-05-24
[WIP] Use Zenity as a file chooser when in a Flatpak sandbox

### LINKED ISSUE #1887: Use File chooser portal

Use the File chooser portal to improve the privacy of the Flatpak.

**The theory**

As you know, CorsixTH is now [distributed as a Flatpak](https://flathub.org/apps/details/com.corsixth.corsixth). Flatpaks run in a sandboxed environment, and this creates an issue for programs that need to access the host filesystem.

To allow CorsixTH users to choose their game data nonetheless, the Flatpak requires read access to the whole home directory. This is the easiest solution because it doesn't require any change from in the program, but [some users have privacy concerns over this](https://github.com/flathub/com.corsixth.corsixth/issues/1).

It is possible, however, to plug this hole and use [portals](https://docs.flatpak.org/en/latest/sandbox-permissions.html#portals) to get access to the host. The [File chooser portal](https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-org.freedesktop.portal.FileChooser) spawns a system file chooser dialog on the host, and when the user selects files, they are mounted within the sandbox so the program can access them.

**In practice**

The program needs to make DBus calls to use the portals. However, as this is not the most practical, GUI toolkits have widgets that use these portals transparently.

Because CorsixTH doesn't use a GUI toolkit, there is another solution: spawn another program that will in turn call the portal and give back the result to CorsixTH. Such a program would be [`zenity`](https://help.gnome.org/users/zenity/), and [this is what OpenRCT2 uses](https://github.com/OpenRCT2/OpenRCT2/blob/a5a63f839aca539ad75abf09a765e590e5ba973b/src/openrct2-ui/UiContext.Linux.cpp#L153) for its file dialogs.

To not make `zenity` a dependency of CorsixTH, it is possible to only use it if it is already installed and/or the game detects it is in a Flatpak, and fall back to the current in-game file chooser.

There are two downsides:
- Because the dialog is part of the OS, it is not integrated into the game
- File validation can only happen after the user chooses them, because the program can't see them beforehand

**Conclusion**

Although not mandatory for the Flatpak, using the File chooser portal would be a nice convenience and comfort users about the privacy of their data. If that is something you are interested in seeing in CorsixTH, I can look into working on it, even if I'm not familiar with the code base and haven't written Lua in a while!

rmnvgr:
*Fixes #1887*

- Add a new method to the app to open a file chooser
- Call this method for all file choosers that need to access the host file system
- If the game is in a Flatpak sandbox, spawn a `zenity` dialog to access the host file system, else use the in-game file chooser

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -334,8 +334,7 @@ function App:init()
       debug.getregistry()._RESTART = true
       TheApp.running = false
     end
-
-    self.ui:addWindow(UIDirectoryBrowser(self.ui, nil, _S.install.th_directory, "InstallDirTreeNode", callback))
+    self:showFileChooser(self.ui, nil, _S.install.th_directory, "InstallDirTreeNode", callback)
     return true
   end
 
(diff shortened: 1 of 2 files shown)


rmnvgr:
@TheCycoONE I'm kinda stuck with it, I'm facing two issues:

- After selecting the folder, the game is stuck. If I close the window, the game restarts itself and loads fine.
- The mouse is captured in the game window, preventing it from going to the zenity dialog. I found the `app.video:setCaptureMouse()` method but the cursor is unable to appear in the dialog until it leaves the game's window.

Do you have any ideas to help me resolve them?

lewri:
Ping @TheCycoONE

TheCycoONE:
In swayvm I don't have a problem with the capture. I suspect I could solve it if I could reproduce - I think the function you need is ui:setMouseReleased(true); which should return the hardware cursor and stop capture.

The problem with the callback/exit is because the init() function which calls all this stuff is before the run method which sets the flag. Since you basically block the game you don't need to restart like callback.

The following worked for me:
```diff
diff --git a/CorsixTH/Lua/app.lua b/CorsixTH/Lua/app.lua
index 47b671f8..52a9e4db 100644
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -85,6 +85,7 @@ end
 
 function App:init()
   -- App initialisation 1st goal: Get the loading screen up
+  self.running = true
 
   print("")
   print("")
@@ -1027,7 +1028,9 @@ function App:run()
     local yield = coroutine.yield
     local dispatch = app.dispatch
     local repaint = true
-    while app.running do
+    local first = true
+    while app.running or first do
+      first = false
       repaint = dispatch(app, yield(repaint))
     end
   end)
@@ -1037,7 +1040,6 @@ function App:run()
     SDL.limitFPS(false)
   end
 
-  self.running = true
   do
     local num_iterations = 0
     self.resetInfiniteLoopChecker = function()
@@ -1104,6 +1106,7 @@ function App:run()
       -- might not otherwise be drawn for a while.
       pcall(self.drawFrame, self)
     end
+    self.running = true
     return self:run()
   end
 end
```

lewri:
@rmnvgr did we still this want PR kept open?

rmnvgr:
I unfortunately don't have time to work on this anymore, let's close it.

PR #1905, opened 2021-05-27
[RDY] Improve blue tint handling at load

lewri:
- In IRC it was reported Autosave1 was loading in with blue tint. This PR changes the checks on when persistance should apply the taint on load.


TheCycoONE:
Looks ok to me for now. In the future I'd like the blue filter logic consolidated so everywhere that might set or unset the filter calls a method that checks all the conditions and applies it appropriately.

PR #1908, opened 2021-05-29
Movie cleanup 2021

TheCycoONE:
Drops support for FFMPEG 3.0 and lower.
Drops support for LibAV (any version)
Less raw pointer use, and in particular raw owned pointers
No more implicit casts
Reworked audio conversion (no more audio buffer)
Replaced linked list implementation in packet queue with std::queue
Fix for possible undefined behavior around aborting


CorsixTH/Src/th_movie.cpp
```diff
@@ -296,59 +247,28 @@ int movie_picture_buffer::write(AVFrame* pFrame, double dPts) {
...
-    pPacket = (AVPacket*)av_malloc(sizeof(AVPacket));
-    *pPacket = pNode->pkt;
-    av_free(pNode);
+  av_packet_unique_ptr pPacket(nullptr);
+  if (!data.empty()) {
+    pPacket.swap(data.front());
+    data.pop();
   }
 
   return pPacket;
```
Alberth289346, line 274: Old problem, but this looks weird. It locks the mutex at the start of the function but doesn't release it here.

TheCycoONE: Yes it does. C++ mutexes release at end of scope.

Alberth289346:
LGTM

TheCycoONE:
Still in progress, hence draft, but good so far 👍

CorsixTH/Src/th_movie.h
```diff
@@ -411,22 +423,22 @@ class movie_player {
...
   int audio_buffer_size;      ///< The current size of audio data in
-                              ///< #m_pbAudioBuffer
+                              ///< #audio_buffer
   int audio_buffer_index;     ///< The current position for writing in
-                              ///< #m_pbAudioBuffer
-  int audio_buffer_max_size;  ///< The capacity of #m_pbAudioBuffer (allocated
+                              ///< #audio_buffer
+  int audio_buffer_max_size;  ///< The capacity of #audio_buffer (allocated
                               ///< size)
   uint8_t* audio_buffer;      ///< An audio buffer for playback
```
Alberth289346, line 431: an class with data storage and these administrative numbers could be useful?

TheCycoONE: std::array/std::vector can be used for that; will get around to replacing these in this patch set.

TheCycoONE:
I think that's just about enough things for one patch set.
Best to review each commit independently rather than the whole set - I mostly tried to keep each step compartmentalized.

CorsixTH/Src/th_movie.h
```diff
@@ -169,28 +183,40 @@ class movie_picture_buffer {
...
+  //! should only be accessed on the main thread
+  SDL_Texture* texture;
+
+  //! A mutex for restricting access to the picture buffer to a single thread
+  std::mutex mutex;
+
+  //! A condition for indicating access to the picture buffer
+  std::condition_variable cond;
+
+  //! The picture queue, aloo ping FIFO queue of movie_pictures
```
TheCycoONE, line 218: aloo ping!

CorsixTH/Src/th_movie.cpp
```diff
@@ -729,17 +693,11 @@ int movie_player::decode_audio_frame(bool fFirst) {
...
-#elif defined(CORSIX_TH_USE_LIBAV)
-  avresample_convert(audio_resample_context, &audio_buffer, 0, iOutSamples,
-                     (uint8_t**)&audio_frame->data[0], 0,
-                     audio_frame->nb_samples);
-#endif
   return iSampleSize;
 }
-#else   // CORSIX_TH_USE_FFMPEG || CORSIX_TH_USE_LIBAV
+#else   // CORSIX_TH_USE_FFMPEG
 movie_player::movie_player() {}
```
Alberth289346, line 681: We seem to indent code for `#` lines too, but not here apprarently?

Alberth289346: `th_movie.h` also doesn't always, maybe we have another rule?

Alberth289346:
lgtm

PR #1910, opened 2021-05-30
Add layer constant, refactor line class

Alberth289346:
- Introduces the `max_number_of_layers` constant (mostly cherry-picked from driverpt)
- Further reduces layer count magic numbers
- Removes Replaces the `link_list` base class, and replaces it with a `std::vector`
- Renames `line` to `line_sequence`, `line_operation` to `line_element`, `line_operation_type` to `line_command`
- Ensures persistence will always write the float data as `double`.


TheCycoONE:
LGTM

PR #1917, opened 2021-06-01
[RDY] Always prefix HOMEPATH with HOMEDRIVE

### LINKED ISSUE #1916: Theme Hospital finder attempts to access network locations
Win10, 0.65-beta2

On the portable, at first launch (config missing) if you press no to the prompt only the command window will display. About 20 seconds later the application loads -- locks the mouse but doesn't show the window until you Alt-tab.

Edit: TheApp:init() seems to be responsible for that.

Edit2: Something in this loop causes freezing up after the first pass (don't have line numbers at the mo), but it's where we look for game data.

```lua
    for _, dir in pairs(possible_locations) do
      if status then break end
      for _, folder in pairs(possible_folders) do
        local path = dir .. pathsep .. folder

        if lfs.attributes(path, "mode") == "directory" and self:isThemeHospitalPath(path) then
          print("Game data found at: " .. path)
          print("This will be written to the config file")
          self.config.theme_hospital_install = path
          status, _ = self.fs:setRoot(path)
          break
        end
      end
```

Looks like this directory entry (windows tries to connect to network locations with \\\\): ``dir     \\Documents``
i.e. this entry in ``possible_locations``: ``      user_dir and (user_dir .. pathsep ..  "Documents"):``

_Originally posted by @lewri in https://github.com/CorsixTH/CorsixTH/issues/1911#issuecomment-851643600_

TheCycoONE:
Search USERPROFILE as the closest equivalent to HOME. Search HOMEDRIVE..HOMEPATH if different from USERPROFILE.

Fixes #1916

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1253,10 +1253,16 @@ function App:checkInstallFolder()
     -- then linux Filesystem Hierarchy Standard, then Windows Program Files
     -- mac_app_dir is the macOS app base directory named CorsixTH.app
     local mac_app_dir = debug.getinfo(1).short_src:match("(.*)/Contents/.")
-    local user_dir = os.getenv("HOME") or os.getenv("HOMEPATH")
+    local user_dir = os.getenv("HOME") or os.getenv("USERPROFILE")
+    local win_home_dir = nil;
+    if os.getenv("HOMEDRIVE") and os.getenv("HOMEPATH") then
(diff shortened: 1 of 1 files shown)


lewri (approved):
This gets rid of the horrible lag now.

It still hangs on win_home_dir a little (but MUCH MUCH less -- like 1 second) but that's because it maps to ``N:\`` (network location). But I don't think we have a way to forecfully exclude network locations in their entirety.

TheCycoONE:
Possible yes, worth it; unlikely

Issue #1913, opened 2021-05-31
File browsers should have an OK button


kparal:
**What feature or enhancement would you like to add?**
When I wanted to select a different font in _Settings -> Folders -> Fonts_, I was confused how to actually select it. After clicking on the font file, there was no `OK` button (like in the _TH Data_ browser). So I used `Back`, but that turned out to mean `Cancel`. Hitting `Esc` also means `Cancel`. On my third attempt I finally figured out to do a double-click.

It shouldn't be this hard. Please add an `OK` button?

The same UX problem applies to the `Save Game` / `Load Game` dialog.

![1](https://user-images.githubusercontent.com/755451/120193858-312ffa00-c21d-11eb-9504-8213d6e6ddfd.png)

![3](https://user-images.githubusercontent.com/755451/120194089-6d635a80-c21d-11eb-8113-27f80dac42a7.png)

**What version of CorsixTH are you using?**
0.65-beta2, Flathub

lewri:
This will likely miss 0.65... but probably easy to implement in for 0.66

lewri:
#568 talks specifically about the save menu. So just tagging it here.

PR #1918, opened 2021-06-02
PoC: Add "OK" on filebrowser

lewri:
A mock-up of a solution for #1913 . It is not elegant, but hopefully a starting block.
Also missing doc stuff.

- Adds OK button to all file browsers (load, save, font)
- Sends the original window through the file browser system to reference back
- On saves, it appends the filename selected into the save textbox when single-clicked.

image
image

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/file_browser.lua
+++ b/CorsixTH/Lua/dialogs/resizables/file_browser.lua
@@ -207,8 +207,8 @@ function UIFileBrowser:UIFileBrowser(ui, mode, title, vertical_size, root, show_
   local indent = math.floor((h_size - (2*button_size))/3)
   self:addBevelPanel(indent, 340, button_size, 30, self.col_bg):setLabel(_S.menu_list_window.back)
     :makeButton(0, 0, button_size, 40, nil, self.buttonBack):setTooltip(_S.tooltip.menu_list_window.back)
-  
-  self:addBevelPanel(h_size - button_size - indent, 340, button_size, 30, 
+
+  self:addBevelPanel(h_size - button_size - indent, 340, button_size, 30,
   self.col_bg):setLabel(_S.menu_list_window.ok)
     :makeButton(0, 0, button_size, 40, nil, (--[[persistable:filebrowser_ok_callback]] function()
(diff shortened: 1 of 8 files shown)


CorsixTH/Lua/dialogs/resizables/file_browser.lua
```diff
@@ -189,31 +189,47 @@ function UIFileBrowser:UIFileBrowser(ui, mode, title, vertical_size, root, show_
...
-    :makeButton(0, 0, 160, 40, nil, self.buttonBack):setTooltip(_S.tooltip.menu_list_window.back)
+  local button_size = 135
+  local indent = math.floor((h_size - (2*button_size))/3)
+  self:addBevelPanel(indent, 340, button_size, 30, self.col_bg):setLabel(_S.menu_list_window.back)
+    :makeButton(0, 0, button_size, 40, nil, self.buttonBack):setTooltip(_S.tooltip.menu_list_window.back)
+
+  self:addBevelPanel(h_size - button_size - indent, 340, button_size, 30,
+  self.col_bg):setLabel(_S.menu_list_window.ok)
+    :makeButton(0, 0, button_size, 40, nil, (--[[persistable:filebrowser_ok_callback]] function()
+      if (self.origin.new_savegame_textbox and self.origin.new_savegame_textbox.text ~= "") or (self.origin.new_map_textbox and self.origin.new_map_textbox.text ~= "") then
```
TheCycoONE, line 214: Way too much outside knowledge in this line. Pass in a callback instead of origin; and pass the node to the callback.

TheCycoONE (changes requested):
Nice POC, but for a real implementation we do not want to pass origin in.

CorsixTH/Lua/dialogs/resizables/file_browser.lua
Alberth289346: `origin` is actually a derived class of `file_browser`, so a simple method overridden in the derived class is enough.

lewri:
Probably going to drop the "On saves, it appends the filename selected into the save textbox when single-clicked." so the original request can make `0.66`

PR #1919, opened 2021-06-05
[RDY] Fix audio on language change

### LINKED ISSUE #1906: [Bug] Change to unicode language (e.g. Czech) kills audio
#### Describe the issue
If you choose a language that needs its own font specified (e.g. Czech), then all audio stops until you either restart the game, or toggle "Global audio" off and on again.

#### Steps to Reproduce
1. Go into game and set a font e.g. arial

2. Choose one of the languages that needs a font specified (e.g. Czech).

3. Click back or any button and no sound occurs -- including if you load a game.

#### Expected Behavior
Don't lose audio on language change -- notably the languages not requiring a specific font don't cause this error.

#### System Information
**CorsixTH Version:** 0.65-beta1 (present in 0.64 too)

#### Gamelog.txt

```
Paste gamelog.txt output here
```

### Additional Info
illustrate the problem.*

mugmuggy:
*Fixes #1906*

- There are two return paths that assume these values should not change, if the sound file is not changed also.

### DIFF
--- a/CorsixTH/Lua/audio.lua
+++ b/CorsixTH/Lua/audio.lua
@@ -179,8 +179,6 @@ function Audio:init()
 end
 
 function Audio:initSpeech(speech_file)
-  self.sound_archive = nil
-  self.sound_fx = nil
   if self.not_loaded then
     return
   end
@@ -205,6 +203,10 @@ function Audio:initSpeech(speech_file)
     speech_file = "Sound-0.dat"
     archive_data = load_sound_file(speech_file)
   end
+
+  self.sound_archive = nil
+  self.sound_fx = nil
+  self.speech_file_name = nil
 
   if not archive_data then
     if self.app.good_install_folder then


CorsixTH/Lua/audio.lua
```diff
@@ -206,6 +204,9 @@ function Audio:initSpeech(speech_file)
     archive_data = load_sound_file(speech_file)
   end
 
+  self.sound_archive = nil
+  self.sound_fx = nil
+
```
TheCycoONE, line 209: I think we should be setting speech_file_name to nil here too? Otherwise it could error out below then not load again above.

TheCycoONE: The not_loaded variable is fishy - why aren't we setting it on all the not_loaded conditions but only there?

lewri:
This is probably being best snuck into 0.65, if you're able to address the query @mugmuggy

TheCycoONE:
No, at this point it missed. I'm not sneaking something in now. It's an old issue; workaround is easy.

CorsixTH/Lua/audio.lua
mugmuggy: Added speech_file_name, couldn't hurt. Not sure I follow the not_loaded comment, unless you mean it just looks odd, but I assume as its driven from that global audio, and it only sets it to true, if there is no music either loaded, which is done in Audio:init. No music, no sounds, disable it.

PR #1928, opened 2021-06-13
Fix warn queue check considering inactive rooms

### LINKED ISSUE #1927: [Bug] Queue warnings error on crashed rooms
#### Describe the issue
#1807 introduced queue length warning, but in the instance of a crashed room, the queue is nil from Door:closeDoor.

#### Steps to Reproduce
1. Load save game from #1462 for example and wait til the 10th day of month

#### Save Game
See #1462

#### Expected Behavior
Should skip over crashed rooms

#### System Information

#### Gamelog.txt
```
Savegame version is 155 (Trunk), originally it was 155 (Trunk).
Error in timer handler: 
Lua\hospitals\player_hospital.lua:344: attempt to get length of a nil value (field 'queue')
stack traceback:
	Lua\hospitals\player_hospital.lua:344: in method 'warnForLongQueues'
	Lua\hospitals\player_hospital.lua:184: in method 'dailyAdviceChecks'
	Lua\hospitals\player_hospital.lua:374: in method 'onEndDay'
	Lua\world.lua:1043: in method 'onTick'
	Lua\app.lua:1131: in function <Lua\app.lua:1128>
	(...tail calls...)
	Lua\app.lua:1031: in function <Lua\app.lua:1026>

```

### Additional Info
@tobylane had mentioned in irc the 'fixed' line was failing
> The line "if room.door.queue and #room.door.queue then" causes an error saying attempt to get length of nil value. That sounds impossible, why did the first part return true?
> 
Maybe testing against the code before the fix or forgot to push an update or something else.

tobylane:
*Fixes #1927*

- Only considers active rooms when calculating the queuing situation.

I think this is because a crashed room's door queue is not always instantly wiped, as patients move to other rooms. It happened when my first operating theatre exploded but not the second.

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -2414,3 +2414,23 @@ function Hospital:setCampaignData(campaign_data)
     self[key] = value
   end
 end
+
+--! Finds a random room with significantly long queues
+--!return room or nil The chosen room in the hospital or nil for no room busy enough
+function Hospital:getRandomBusyRoom()
+  local long_queue_rooms, active_rooms, total_queue = {}, 0, 0
+  for _, room in pairs(self.world.rooms) do
+    if room.is_active and room.door.queue then
+      total_queue = total_queue + #room.door.queue
+      active_rooms = active_rooms + 1
+      if #room.door.queue > 7 then
+        long_queue_rooms[#long_queue_rooms + 1] = room
+      end
+    end
+  end
+  if #long_queue_rooms == 0 then return end
+
+  local busy_threshold = 1.5 * total_queue / active_rooms
+  local chosen_room = long_queue_rooms[math.random(1, #long_queue_rooms)]
+  if #chosen_room.door.queue >= busy_threshold then return chosen_room end
+end
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -339,20 +339,8 @@ end
 --! Once a month the advisor may warn about long queues.
 --! Rooms requiring a doctor occasionally trigger the generic message
 function PlayerHospital:warnForLongQueues()
-  local queue_rooms, total_queue = {}, 0
-  for _, room in pairs(self.world.rooms) do
-    if #room.door.queue then
-      total_queue = total_queue + #room.door.queue
-    end
-    if #room.door.queue > 7 then
-      queue_rooms[#queue_rooms + 1] = room
-    end
-  end
-  if #queue_rooms == 0 or total_queue == 0 then return end
-
-  local busy_threshold = 1.5 * total_queue / #self.world.rooms
-  local chosen_room = queue_rooms[math.random(1, #queue_rooms)]
-  if busy_threshold > #chosen_room.door.queue then return end
+  local chosen_room = self:getRandomBusyRoom()
+  if not chosen_room then return end
   chosen_room = chosen_room.room_info
   -- Required staff that is not nurse is doctor, researcher, surgeon or psych
   if chosen_room.required_staff and not chosen_room.required_staff["Nurse"]


TheCycoONE:
I think this is wrong? self there is playerHospital and it doesn't have an is_active, so that always evaluates false?

tobylane:
Fixed and improved.

lewri:
Looks okay here

TheCycoONE:
This is for 0.65 (fixes regression) right?

tobylane:
Yes, should I change the base branch?

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -339,19 +339,20 @@ end
...
-    if #room.door.queue > 7 then
-      queue_rooms[#queue_rooms + 1] = room
+      active_rooms = active_rooms + 1
+      if #room.door.queue > 7 then
+        long_queue_rooms[#long_queue_rooms + 1] = room
+      end
     end
   end
-  if #queue_rooms == 0 or total_queue == 0 then return end
+  if #long_queue_rooms == 0 or total_queue == 0 then return end
```
Alberth289346, line 352: Not your fault, but if `#long_queue_rooms == 0` fails (ie it's larger than `0`), then `total_queue == 0` is also not going to hold, since a 'long queue room' has at least 8 people in the queue.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -339,19 +339,20 @@ end
...
 
-  local busy_threshold = 1.5 * total_queue / #self.world.rooms
-  local chosen_room = queue_rooms[math.random(1, #queue_rooms)]
+  local busy_threshold = 1.5 * total_queue / active_rooms
+  local chosen_room = long_queue_rooms[math.random(1, #long_queue_rooms)]
   if busy_threshold > #chosen_room.door.queue then return end
```
Alberth289346, line 356: So let's say we have 5 busy rooms, and by chance they all around 20 people in the queue.
So `busy_treshold` is 30, which is bigger than any of the queues, so we don't give advice.

Is that intended?

Alberth289346, line 356: There may not be rooms with a long queue, so `chosen_room` becomes `nil` (I hope).
In that case, `chosen_room.door.queue` will crash the program on dereferencing the `door` field from a `nil` value.

tobylane: If there aren't rooms with a long queue, then the function ends at line 352.

tobylane: Yes, #1807 has the intent. I think that situation is too rare to account for.

lewri: Chances are your busy threshold needs a ceiling of some kind. But I do get that it might be pointless your adviser stating the obvious.

tobylane:
Split up the select and advise parts following IRC feedback.

CorsixTH/Lua/hospitals/player_hospital.lua
Alberth289346: good point :)

Alberth289346: fair enough, let's keep it

TheCycoONE:
Keeping in mind this is a regression fix, I wouldn't have changed it, and won't suggest further changes for this PR.

CorsixTH/Lua/hospital.lua
```diff
@@ -2414,3 +2414,23 @@ function Hospital:setCampaignData(campaign_data)
     self[key] = value
   end
 end
+
+--! Finds a random room with significantly long queues
+--!return room or nil The chosen room in the hospital or nil for no room busy enough
+function Hospital:getRandomBusyRoom()
+  local long_queue_rooms, active_rooms, total_queue = {}, 0, 0
+  for _, room in pairs(self.world.rooms) do
+    if room.is_active and #room.door.queue then
```
TheCycoONE, line 2423: #room.door.queue in a boolean condition is not useful, 0 is true in lua. Perhaps you didn't mean to include the # here?

TheCycoONE (changes requested):
One problem that looks like a bug, not going to comment or suggest style changes for this PR.

CorsixTH/Lua/hospital.lua
tobylane: 0 is acceptable in this case, it's an empty queue but a room that should be counted. I included # because it's what I need to exist, and I don't know why `if room.door.queue and #room.door.queue then` had an error.

TheCycoONE: I'm pretty sure there's no situation where #foo returns false or nil in lua.

tobylane: Fixed.

CorsixTH/Lua/hospital.lua
```diff
@@ -2414,3 +2414,23 @@ function Hospital:setCampaignData(campaign_data)
...
+  for _, room in pairs(self.world.rooms) do
+    if room.is_active and room.door.queue then
+      total_queue = total_queue + #room.door.queue
+      active_rooms = active_rooms + 1
+      if #room.door.queue > 7 then
+        long_queue_rooms[#long_queue_rooms + 1] = room
+      end
+    end
+  end
+  if #long_queue_rooms == 0 then return end
```
lewri, line 2431: Confirming this is fine because long_queue_rooms is an array/list.

CorsixTH/Lua/hospital.lua
```diff
@@ -2414,3 +2414,23 @@ function Hospital:setCampaignData(campaign_data)
...
+
+--! Finds a random room with significantly long queues
+--!return room or nil The chosen room in the hospital or nil for no room busy enough
+function Hospital:getRandomBusyRoom()
+  local long_queue_rooms, active_rooms, total_queue = {}, 0, 0
+  for _, room in pairs(self.world.rooms) do
+    if room.is_active and room.door.queue then
+      total_queue = total_queue + #room.door.queue
+      active_rooms = active_rooms + 1
+      if #room.door.queue > 7 then
```
lewri, line 2426: The 7 could potentially be magic here... but I would be fine letting it slide (as it came from the previous patch)

lewri (approved):
All fine. But note for a different issue. The code here in player_hospital seems to cause the "place a doctor"[sic] message even though a doctor is in the room (happened with a GP Office).

```lua
if chosen_room.required_staff and not chosen_room.required_staff["Nurse"]
   and math.random(1, 3) &gt; 1 then
   local warn_msgs = {
      _A.warnings.queue_too_long_send_doctor:format(chosen_room.name),
      _A.staff_advice.need_doctors
      }
   self:giveAdvice(warn_msgs)
else
  self.world.ui.adviser:say(_A.warnings.queues_too_long)
end
```

CorsixTH/Lua/hospital.lua
TheCycoONE: Yes this is fine.

Issue #1930, opened 2021-06-13
[Bug] Significant aliasing in fullscreen with direct_zoom


flackr:
#### Describe the issue

With direct_zoom enabled, fullscreen mode is noticeably more aliased than without. See screenshots below.

#### Steps to Reproduce
1. Open the game

2. Press alt+enter to enter fullscreen

#### Save Game

N/A

#### Expected Behavior

Expect smooth consistent looking rendering.

#### System Information
**CorsixTH Version:** *0.65*

**Operating System:** *Windows 10*

**Theme Hospital Version:** *GOG.com*

#### Gamelog.txt

N/A

### Additional Info

Without direct_zoom:
![fullscreen-expected](https://user-images.githubusercontent.com/366761/121811282-e6a57900-cc31-11eb-9cbc-52770c22079a.png)

With direct_zoom:
![fullscreen](https://user-images.githubusercontent.com/366761/121811287-ec02c380-cc31-11eb-91f7-8bf14a5c9708.png)

flackr:
This is almost certainly caused by "nearest" SDL_HINT_RENDER_SCALE_QUALITY in https://github.com/CorsixTH/CorsixTH/pull/1921/commits/12160af00d5003e35e6767a74fe610e6134fecf7#diff-df053efb64fd3b9f46a7a2df00e9f910a02fa3ce9cbd364fdd031a7c6fa7137dR303

I suspect this would be fixed by using the "nearest" hint only for the game tiles and then switching back to "linear" afterwards.

PR #1931, opened 2021-06-13
Only apply nearest scaling to map textures for direct_zoom.

flackr:
*Mostly fixes aliased rendering issues in #1930*

- Applies SDL_HINT_RENDER_SCALE_QUALITY nearest only to map textures

Note:
- We could now also make direct_zoom a runtime configurable parameter as long as we reset the global_scale_factor and zoom_texture accordingly.

**Screenshot:**
titlescreen-linear-antialiasing

### DIFF
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -1007,13 +1007,13 @@ void level_map::draw(render_target* pCanvas, int iScreenX, int iScreenY,
   for (map_tile_iterator itrNode2(this, iScreenX, iScreenY, iWidth, iHeight);
        itrNode2; ++itrNode2) {
     if (itrNode2->flags.shadow_full) {
-      blocks->draw_sprite(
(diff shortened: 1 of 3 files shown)


CorsixTH/Src/th_map.cpp
```diff
@@ -996,7 +996,8 @@ void level_map::draw(render_target* pCanvas, int iScreenX, int iScreenY,
     blocks->draw_sprite(
         pCanvas, iBlock & 0xFF,
         itrNode1.tile_x_position_on_screen() + iCanvasX - 32,
-        itrNode1.tile_y_position_on_screen() + iCanvasY - iH + 32, iBlock >> 8);
+        itrNode1.tile_y_position_on_screen() + iCanvasY - iH + 32,
+        iBlock >> 8 | thdf_nearest);
```
Alberth289346, line 1000: I'd prefer some parentheses here (and at other places as well, as it was added several times)

flackr: Absolutely, makes sense! Done.

PR #1932, opened 2021-06-13
Draw floor shadows immediately after floor tiles

### LINKED ISSUE #1929: [Bug] Artifacts visible in shadows when zooming with direct_zoom
#### Describe the issue

With the direct_zoom feature enabled, there are pixels between tiles which are slightly darker than the rest of the shadow. 

#### Steps to Reproduce
1. Set `direct_zoom = true` in CorsixTH `config.txt`.

2. Start a new game.

3. Zoom in or out closely watching the shadows near northwest walls.

4. At certain zoom levels single pixels along seems between tiles will appear darker, as the translucent shadow rendering is likely overlapping by a single pixel. See screenshot below:

#### Save Game

This should be easily reproducible from a new game.

#### Expected Behavior

There shouldn't be any visible doubly shaded pixels.

#### System Information
**CorsixTH Version:** *0.65*

**Operating System:** *Linux*

#### Gamelog.txt

The Gamelog likely does not contain any useful information

### Additional Info

This is a picture of the issue:
shadow-artifacts

flackr:
*Fixes #1929*

This change draws each floor shadow immediately after the floor tile.

By drawing the floor shadows immediately after drawing the floor tile we ensure that no pixel is shaded by shadow twice, as if it would be it would first be covered by the adjacent floor tile if there will be overlap.

### DIFF
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -959,20 +959,19 @@ void level_map::draw(render_target* pCanvas, int iScreenX, int iScreenY,
                      int iWidth, int iHeight, int iCanvasX,
                      int iCanvasY) const {
   /*
-     The map is drawn in two passes, with each pass done one scanline at a
-     time (a scanline is a list of tiles with the same screen Y coordinate).
-     The first pass does floor tiles, as the entire floor needs to be painted
+     The map is drawn in a single pass, done one scanline at a time
+     (a scanline is a list of tiles with the same screen Y coordinate).
+     First the floor tile is drawn, as the entire floor needs to be painted
      below anything else (for example, see the walking north through a door
      animation, which needs to paint over the floor of the scanline below the
-     animation). On the second pass, walls and entities are drawn, with the
+     animation). Then the floor shadows, walls and entities are drawn, with the
      order controlled such that entities appear in the right order relative to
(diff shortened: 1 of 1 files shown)


mugmuggy:
I've got this same change in a playground code base, where I was looking at some of the rendering issues, but never took a look at shadows under zoom, or if I did I ignored it. Also using the inbuilt FPS counter, I was getting an ever so slight degradation (under extremes, zoomed out with 100s of entities) degradation. Maybe your other zoom changes improve on that anyway.

flackr:
@mugmuggy it's worth noting that the direct_zoom feature (which this change is intended to stabilize) is a significant performance improvement when zoomed out. On my system I see a 27% fps increase.

flackr:
Here's a side-by-side of the fixed shadows with `direct_zoom = true` (left is master, right is with this fix)
direct-zoom-shadows

flackr:
It occurs to me there's probably no need to separate the 1st pass at all if the floor shadows which should be below everything else were drawn as part of the second pass

TheCycoONE:
Yes, the tiles were only drawn first because there use to be an optimization in the old backends that made it quicker to draw that way, since that's gone it's no longer useful.

flackr:
Done, I've merged the two render passes and updated the comments.

PR #1934, opened 2021-06-13
Use intermediate screen-sized buffer for direct_zoom

### LINKED ISSUE #1930: [Bug] Significant aliasing in fullscreen with direct_zoom

flackr:
*In combination with #1931 this fixes #1930*

- Initialize and use an intermediate screen sized buffer when in fullscreen.

Note that this still has the advantage over indirect zoom that the buffer will always be screen-sized, rather than growing larger as you zoom out so it should still should fix the original crash issue that direct_zoom fixed with hopefully full parity in render quality.

**Screenshot:**
titlescreen-intermediate-buffer

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -389,6 +389,7 @@ bool render_target::set_scale_factor(double fScale, scaled_items eWhatToScale) {
     return false;
   } else if (eWhatToScale == scaled_items::all && direct_zoom) {
     global_scale_factor = fScale;
+    init_zoom_buffer(width, height);
     return true;
   } else if (eWhatToScale == scaled_items::all && supports_target_textures) {
     // Draw everything from now until the next scale to zoom_texture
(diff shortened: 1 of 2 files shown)


CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -389,6 +389,12 @@ bool render_target::set_scale_factor(double fScale, scaled_items eWhatToScale) {
     return false;
   } else if (eWhatToScale == scaled_items::all && direct_zoom) {
     global_scale_factor = fScale;
+    if ((SDL_GetWindowFlags(window) & SDL_WINDOW_FULLSCREEN_DESKTOP) ==
```
TheCycoONE, line 392: This function is actually called even more than once a frame, at least 2, maybe 3 times. It's called with a value of 1 before drawing UI elements, and then with a value of the current zoom level before drawing the map. That's why flush_zoom_buffer is called at the beginning of this function and the end of the frame.

TheCycoONE (commented):
Do you have an idea why this works? I think I found the same thing years ago (why the fScale == 1 condition is below the zoom instead of above), but I don't recall why.

TheCycoONE:
Haven't tried it yet but I wonder if SDL_WindowGetDisplayMode could be used to determine if the full screen window was running at native resolution.

CorsixTH/Src/th_gfx_sdl.cpp
flackr: > Do you have an idea why this works? I think I found the same thing years ago (why the fScale == 1 condition is below the zoom instead of above), but I don't recall why.

I haven't narrowed down exactly why the quality is so different. My leading guess is that drawing to an intermediate texure gives a pixel perfect rendering which is later upsampled as a whole whereas drawing directly is immediately upsampling each draw call smearing textures across pixels.

flackr: > This function is actually called even more than once a frame, at least 2, maybe 3 times. It's called with a value of 1 before drawing UI elements, and then with a value of the current zoom level before drawing the map. That's why flush_zoom_buffer is called at the beginning of this function and the end of the frame.

It should be easy to remove the intermediate flush calls and use a single texture with direct_zoom given that the texture is always screen size.

TheCycoONE (approved):
As discussed in IRC, any optimizations can come later.

flackr:
Note that this should be merged after #1931 , without which the intermediate texture would just use nearest quality when upscaled.

PR #1937, opened 2021-06-17
[Rdy] Training room speed adjustment

lewri:
- Allow handymen to move around at their regular speed in training rooms -- as per the original game.


TheCycoONE (approved):
LGTM

PR #1938, opened 2021-06-18
Implement serious radiation animation effect

### LINKED ISSUE #1659: Serious Radiation Graphics

flackr:
*Fixes #1659 *

This is a continuation of https://github.com/CorsixTH/CorsixTH/pull/1794 . I have addressed all of the comments on that PR.

- Adds a new animation_effect for applying jelly or radiation effects.
- Implements radiation effect using SDL_TextureColorMod to mask out green channel.
- Applies the effect to patients having serious radiation.


lewri:
Very nice :)

TheCycoONE:
Looks good.

CorsixTH/Lua/diseases/serious_radiation.lua
```diff
@@ -43,6 +43,7 @@ disease.initPatient = function(patient)
   patient:setLayer(1, math.random(0, 3) * 2)
   patient:setLayer(3, 0)
   patient:setLayer(4, 0)
+  patient.th:setEffect(AnimationEffect.Glowing)
 end
 -- TODO: visual radiation effect
```
lewri, line 48: This Todo line can go now as per the pr

CorsixTH/Lua/utility.lua
```diff
@@ -216,6 +216,12 @@ DrawFlags.ListBottom      = 2^11
 DrawFlags.BoundBoxHitTest = 2^12
 DrawFlags.Crop            = 2^13
 
+-- Keep in sync with animation_effect in th_gfx_common.h
+AnimationEffect = {}
+AnimationEffect.None = 0
+AnimationEffect.Glowing = 1
+AnimationEffect.Jelly = 2
```
lewri, line 223: Add an end of line comment e.g. not yet implemented. Just incase jelly isn't implemented straight away

lewri (commented):
Not sure if you need an afterLoad to apply effects to any existing patients.

flackr (commented):
> Not sure if you need an afterLoad to apply effects to any existing patients.

Yup, I sure do! Please have a look and let me know if what I've done makes sense. I've tested loading a save and it now reapplies the radiation effect.

CorsixTH/Lua/diseases/serious_radiation.lua
```diff
@@ -45,6 +45,9 @@ disease.initPatient = function(patient)
   patient:setLayer(4, 0)
   patient.th:setEffect(AnimationEffect.Glowing)
 end
+disease.afterLoad = function(patient)
+  patient.th:setEffect(AnimationEffect.Glowing)
```
flackr, line 49: I think this should check patient.cured, or patient.infected?

TheCycoONE:
That's a good point, the decontamination room needs to `setEffect(AnimationEffect.None)` on the patient after_use.

flackr:
> That's a good point, the decontamination room needs to `setEffect(AnimationEffect.None)` on the patient after_use.

Yep, that too. Maybe disease should have an update function which can check the !cured bit? (is that the right thing to check or is it infected or are there important differences between !cured and infected?)

flackr:
> > That's a good point, the decontamination room needs to `setEffect(AnimationEffect.None)` on the patient after_use.
> 
> Yep, that too. Maybe disease should have an update function which can check the !cured bit? (is that the right thing to check or is it infected or are there important differences between !cured and infected?)

I think that the latest patch should do this (calls updatePatient after cure is called), however I'm not actually sure how to test this. I tried all research in the debug menu but it didn't seem to unlock any treatment rooms.

flackr:
> I think that the latest patch should do this (calls updatePatient after cure is called), however I'm not actually sure how to test this. I tried all research in the debug menu but it didn't seem to unlock any treatment rooms.

Verified treatment works on the save file example attached to the issue. However I discovered another issue, while the patient is interacting with other sprites (doors, the decontamination room, etc) those other sprites get the same color mod. Should be a simple bug to fix.

TheCycoONE:
infected is different and has to do with epidemics.
cure doesn't quite match either, after the treatment room the patient is either cured or starts their dying action, but either way they no longer have the effect.

The patient has a treatDisease() function already which could call a similarly named function on their visual disease to remove visualizations when appropriate. I'm not sure that's always called at the right time though.

slack_tongue and pharmacy setType in after_use in the same tick that they call patient.treatDisease
dna_fixer and electrolysis setType before the treatment machine runs

TheCycoONE:
The effect should be persisted in the animation itself.
See th_gfx.cpp animation::persist and animation::depersist

The afterLoad in lua can approximate when to add the effect for old versions only based on the cured flag.

flackr:
> infected is different and has to do with epidemics.
> cure doesn't quite match either, after the treatment room the patient is either cured or starts their dying action, but either way they no longer have the effect.
> 
> The patient has a treatDisease() function already which could call a similarly named function on their visual disease to remove visualizations when appropriate. I'm not sure that's always called at the right time though.

There needs to be some persistence though so that we know what state to apply if we reload the game right? Or is the disease removed at that point?

> slack_tongue and pharmacy setType in after_use in the same tick that they call patient.treatDisease
> dna_fixer and electrolysis setType before the treatment machine runs

I think what's going wrong with the entities being interacted with getting the color mod applied is that this patch sets the effect on the whole animation. I suspect the animation may include things like doors being opened so that everything is layered correctly.

flackr:
> The effect should be persisted in the animation itself.
> See th_gfx.cpp animation::persist and animation::depersist

I think I need help with how to do this without breaking previous save games. Naively persisting / depersisting this new attribute throws an error when loading a previous save.

> The afterLoad in lua can approximate when to add the effect for old versions only based on the cured flag.

I've restored the afterLoad function to guess the state based on cured state. I'm not sure how to only apply this to old versions.

Out of curiosity, why does a patient who failed to be treated not continue to show the disease?

CorsixTH/Lua/diseases/serious_radiation.lua
```diff
@@ -43,8 +43,18 @@ disease.initPatient = function(patient)
   patient:setLayer(1, math.random(0, 3) * 2)
   patient:setLayer(3, 0)
   patient:setLayer(4, 0)
+  patient.th:setPatientEffect(AnimationEffect.Glowing)
+end
+disease.treatPatient = function(patient)
+  patient.th:setPatientEffect(AnimationEffect.None)
+end
+disease.afterLoad = function(patient)
```
lewri, line 51: This seems a little off in the way of doing an afterLoad that I've seen before. Especially in other files.
Not saying it's wrong in this case, just not what I normally see.

CorsixTH/Lua/diseases/serious_radiation.lua
```diff
@@ -43,13 +43,16 @@ disease.initPatient = function(patient)
...
   patient:setLayer(3, 0)
   patient:setLayer(4, 0)
-  patient.th:setEffect(AnimationEffect.Glowing)
+  patient.th:setPatientEffect(AnimationEffect.Glowing)
 end
-disease.updatePatient = function(patient)
+disease.treatPatient = function(patient)
+  patient.th:setPatientEffect(AnimationEffect.None)
+end
+disease.afterLoad = function(patient)
```
TheCycoONE, line 51: I don't have a good recommendation yet, but I don't like this.
afterLoad conventionally takes two parameters old and new version; that pattern applies everywhere else in CorsixTH.

Perhaps the problem is disease is suppose to be a collection of values and it's becoming increasingly a class with logic and side effects. Maybe diseases should just have an effect, and patient should apply it as appropriate...

TheCycoONE: That is closer to how Type is managed

flackr: Yeah, I didn't like it either. Thanks for explaining the parameters, I was searching the code to try to figure out they were.

I like your suggestion to just have an effect on the disease that the patient class applies, I've switched to this strategy now and tested / verified it's working.

TheCycoONE:
I can't really speak for the Theme Hospital developers but the implication is that a) there are variations of the diseases and by treating the wrong variation you will kill the person; and b) that sometimes doctor's botch the treatment (lack of skill, fatigue) and the patient dies.
In either case, the right room / equipment are used, we can assume they address the visual symptoms but messed something hidden up. The patient leaves, pays for their treatment, everything looks good - they take a few steps, and oops, dead.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -28,6 +28,7 @@ SOFTWARE.
 #include <cassert>
 #include <climits>
 #include <cstring>
+#include <iostream>
```
TheCycoONE, line 31: Leftover debugging?

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -307,7 +313,9 @@ bool render_target::create(const render_target_creation_params* pParams) {
   }
 
   Uint32 iRendererFlags =
-      (pParams->present_immediate ? 0 : SDL_RENDERER_PRESENTVSYNC);
+      (pParams->present_immediate
+           ? 0
+           : SDL_RENDERER_PRESENTVSYNC | SDL_RENDERER_ACCELERATED);
```
TheCycoONE, line 318: SDL_RENDERER_ACCELERATED should be the preferred default anyway. What did applying this flag here do?

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -1090,10 +1099,32 @@ void sprite_sheet::draw_sprite(render_target* pCanvas, size_t iSprite, int iX,
     }
   }
 
+  if (effect == animation_effect::glowing) {
+    // Let's convert SDL Ticks to Game Ticks
+    uint32_t tick = SDL_GetTicks() / usertick_period_ms;
```
TheCycoONE, line 1104: I expect (but haven't confirmed) that the desired behaviour would have the speed of the affect influenced by the game speed, that is according to the rate that animation.tick is called, rather than SDL_Ticks.

@mugmuggy do you happen to know off hand?

TheCycoONE (commented):
Nice work: open question about the speed, and a couple trivial things, looks really close.

CorsixTH/Src/th_gfx.cpp
flackr: Yup, from the original feature branch :-). Removed.

CorsixTH/Src/th_gfx_sdl.cpp
flackr: This was likely driver's experimentation to attempt to apply shaders. It is not necessary and does not affect this. I've removed it.

CorsixTH/Lua/app.lua
```diff
@@ -28,7 +28,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 156
+local SAVEGAME_VERSION = 157
```
TheCycoONE, line 31: Should be bumped before merged, master is now 157.

flackr: Done.

CorsixTH/Src/th_gfx_sdl.cpp
flackr: I've updated the code to track world ticks on animation_manager. I'm not entirely sure if I've done it correctly. It seems the lua AnimationManager is not actually hooked up to the C++ one, but instead the lua anims interface seems to be the place where we call through to the c++ animation_amanger.

In any event, the effect now seems to pause with the rest of the game.

CorsixTH/Src/th_gfx_sdl.h
```diff
@@ -527,7 +529,9 @@ class sprite_sheet {
       @param iFlags Flags to apply for drawing.
   */
   void draw_sprite(render_target* pCanvas, size_t iSprite, int iX, int iY,
-                   uint32_t iFlags);
+                   uint32_t iFlags,
+                   size_t game_ticks = 0u,
```
flackr, line 533: I'm not a fan of the default value here, however draw_sprite seems to have several call sites not from animation_manager which would complicate knowing the value of game_ticks. It is only used when you also have a non-none effect so maybe that's okay.

CorsixTH/Lua/diseases/jellyitis.lua
```diff
@@ -30,6 +30,7 @@ disease.cure_price = 1000
 disease.emergency_sound = "emerg014.wav"
 disease.emergency_number = 12
 disease.contagious = false
+disease.effect = AnimationEffect.Jelly
```
lewri, line 33: Nothing to change, but careful we're not creeping into the jellyitis pr

flackr: This was an intentional choice. This PR adds the animation patent_effect and persists it to save games, so I decided it would be nice to get both effects saved correctly. Otherwise in https://github.com/CorsixTH/CorsixTH/issues/1946 I'd have to add more complex savegame restoration logic to only restore radiation in 158 and jelly in 159. Having this here means that the game stores that the patient should use the jelly effect (and it is removed when treated), but it is not visualized yet.

CorsixTH/Src/th_lua_anims.cpp
```diff
@@ -182,6 +182,12 @@ int l_anims_set_marker(lua_State* L) {
   return 1;
 }
 
+int l_anims_tick(lua_State* L) {
+  animation_manager* pAnims = luaT_testuserdata<animation_manager>(L);
+  pAnims->tick();
+  return 1;
```
TheCycoONE, line 188: return 0. That's the number of return values pused to the lua stack.

flackr: Thanks! I thought it was some kind of success code or something, heh. Fixed.

TheCycoONE:
Looks good to me, did you have anything in mind for cleaning up the git history, or shall I just squish it all?

flackr:
Squishing it all seems fine.

TheCycoONE (approved):

lewri:
Ping @Alberth289346 if you are free to give this a once over?

CorsixTH/Lua/diseases/jellyitis.lua
lewri: Okay happy then :)

CorsixTH/Src/th_gfx_common.h
```diff
@@ -0,0 +1,39 @@
+/*
+Copyright (c) 2009-2013 Peter "Corsix" Cawley and Edvin "Lego3" Linge
```
lewri, line 2: If this is a new file in its entirety (can't see if it's sourced from somewhere else) Corsix and Edvin are not the people for this line. It is yourself and perhaps driverpt.

Ideally also add yourself to the license.txt file

flackr: Thanks, something like I've done in the latest commit?

CorsixTH/Src/th_gfx_common.h
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2009-2013 Peter "Corsix" Cawley and Edvin "Lego3" Linge
+Copyright (C) 2012-2013 Luis "Driver" Duarte and Robert "flackr" Flack
```
TheCycoONE, line 2: 2012-2013?

flackr: Ah, I got the mistaken idea from 2009-2013 in many files that the year here was not related to the files but instead the contributors.

PR #1940, opened 2021-06-18
Implement reset zoom hotkey.

flackr:
This creates the hotkey with default 0 to reset the current zoom as listed on https://github.com/CorsixTH/CorsixTH/wiki/Programming-Ideas#graphical-scaling

### DIFF
--- a/CorsixTH/Lua/config_finder.lua
+++ b/CorsixTH/Lua/config_finder.lua
@@ -573,6 +573,7 @@ local hotkeys_defaults = {
   ingame_zoom_in_more = {"shift", "="},
   ingame_zoom_out = "-",
   ingame_zoom_out_more = {"shift", "-"},
+  ingame_reset_zoom = "0",
   ingame_setTransparent = "x",
   ingame_toggleAdvisor = {"shift", "a"},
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/config_finder.lua
```diff
@@ -573,6 +573,7 @@ local hotkeys_defaults = {
   ingame_zoom_in_more = {"shift", "="},
   ingame_zoom_out = "-",
   ingame_zoom_out_more = {"shift", "-"},
+  ingame_reset_zoom = "0",
```
lewri, line 576: In regards to the other zoom controls I would think we should have it as ``shift + 0`` . Unless that's already taken.

lewri: Actually the two different zoom commands confused me. Maybe just 0 is fine.

flackr: Yeah, "=" and "-" without holding shift zoom in and out, so it made sense to me that 0 wouldn't require shift either.

TheCycoONE:
I was too quick on this, there's a regression on the hotkey dialog:
Screenshot 2021-06-18 220144

Issue #1941, opened 2021-06-19
Hotkeys don't fit in dialog


TheCycoONE:
https://github.com/CorsixTH/CorsixTH/issues/1940 added an additional hotkey,
unfortunately it caused the keys to reflow past the visible area in the dialog.

The dialog is already 640x480, but some keys are already on separate panels, so the reasonable choice would be to move one of the other sections to a new button.

Note: the keys only extend past the visible area if debug is enabled in the config. For most users the only issue is the odd reflow on game speed.

Suggest moving Game Speed Keys under a new button, between Panel Keys and Recall Position Keys.

![Screenshot 2021-06-18 220144](https://user-images.githubusercontent.com/823433/122627934-daca1480-d080-11eb-9630-4e81881dcd27.png)

PR #1942, opened 2021-06-19
New layout for hotkey settings

TheCycoONE:
Rework of Hotkey settings layout
Fixes: #1941

The UI is now significantly more data driven, making changes somewhat simpler, though the data is optimized for the current UI.

Buttons are on the left, options on the right, sized depending on the number of sections in the window.

screenshot0
screenshot0


Alberth289346:
My first idea was to make the window more 'tab-like', at the right one column of keys with descriptions, at the left buttons to select which keys to show. (And for those countries that read text in the opposite direction, switch left with right and vice-versa everywhere.)

lewri:
I'm also not overly fond of having the speed keeps floating all alone by themselves. I would +1 tabination, similar to how menus break down tabs into groups of relevance

TheCycoONE:
You guys and your scope creep.

I just extended the already existing button interface to fix the defect - but fine. I kinda like the categories on the left, options on the right idea.

We have zero RTL handling so that's out of scope.

TheCycoONE:
There - you people! :stuck_out_tongue:

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
```diff
@@ -764,328 +630,67 @@ function UIHotkeyAssign:close()
...
-    self.ui:addWindow(UIHotkeyAssign(self.ui, "menu"))
+  local current_pos_x = 1
+  for _, section in ipairs(key_options.sections) do
+    local current_pos_y = 1
+    self:addBevelPanel(self.panel_pos_table_x[current_pos_x], self.panel_pos_table_y[current_pos_y], panel_width*2, panel_height, col_caption):setLabel(section.title)
+    current_pos_y = current_pos_y + 1
+    for _, key in ipairs(section.keys) do
+      self:addBevelPanel(self.panel_pos_table_x[current_pos_x], self.panel_pos_table_y[current_pos_y], panel_width, panel_height, col_shadow, col_bg, col_bg) : setLabel(_S.hotkey_window[key])
+    self.hotkey_buttons[key] = self:addBevelPanel(self.panel_pos_table_x[current_pos_x]+panel_width, self.panel_pos_table_y[current_pos_y], panel_width, panel_height, col_hotkeybox, col_highlight, col_shadow)
+      :makeHotkeyBox(function() self:confirm_func(key) end, nil):setText( string.upper(array_join(app_hotkeys[key], "+")) )
```
Alberth289346, line 683: Wrong indentation, also 2nd line should have double indentation.

Also, if you add a local x and y for
``` lua
local x = self.panel_pos_table_x[current_pos_x]
local y = self.panel_pos_table_y[current_pos_y]
```
that would go a long way to get lines with a more readable length, and give space around the expression operators in it

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
```diff
@@ -764,328 +630,67 @@ function UIHotkeyAssign:close()
...
   hotkey_input(hotkey, self.hotkey_buttons, self.app)
 end
 
---! Assign panel keys window.
-class "UIHotkeyAssign_Panels" (UIResizable)
-
----@type UIHotkeyAssign_Panels
-local UIHotkeyAssign_Panels = _G["UIHotkeyAssign_Panels"]
+class "UIHotkeyAssignKeyPane" (Window)
+local UIHotkeyAssignKeyPane = _G["UIHotkeyAssignKeyPane"]
```
Alberth289346, line 646: This gets through lua doc and/or luacheck?

Afaik, the 4 lines to define a class are checked somewhere for their form

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
```diff
@@ -434,297 +410,187 @@ function UIHotkeyAssign:UIHotkeyAssign(ui, mode)
...
-  self.hotkey_buttons["ingame_toggleAdvisor"] = self:addBevelPanel(self.panel_pos_table_x[current_pos_x]+panel_width, self.panel_pos_table_y[current_pos_y], panel_width, panel_height, col_hotkeybox, col_highlight, col_shadow)
-    :makeHotkeyBox(function() self:confirm_func("ingame_toggleAdvisor") end, nil):setText( string.upper(array_join(ui.app.hotkeys["ingame_toggleAdvisor"], "+")) )
-  get_next_pos_y()
-  -- ingame_toggleInfo
-  self:addBevelPanel(self.panel_pos_table_x[current_pos_x], self.panel_pos_table_y[current_pos_y], panel_width, panel_height, col_shadow, col_bg, col_bg) : setLabel(_S.hotkey_window.ingame_toggleInfo)
-  self.hotkey_buttons["ingame_toggleInfo"] = self:addBevelPanel(self.panel_pos_table_x[current_pos_x]+panel_width, self.panel_pos_table_y[current_pos_y], panel_width, panel_height, col_hotkeybox, col_highlight, col_shadow)
-    :makeHotkeyBox(function() self:confirm_func("ingame_toggleInfo") end, nil):setText( string.upper(array_join(ui.app.hotkeys["ingame_toggleInfo"], "+")) )
-  get_next_pos_y()
+  self.key_windows = {}
+  for _, o in ipairs(self.key_options) do
```
Alberth289346, line 571: Single letter `o` is often confused with `0`, depending on the used font.
I'd suggest to rename to `opt` or `key_opt` or `key_option`.

Alberth289346:
The people are eternally grateful to you for your efforts in this matter.

Looks great!

lewri:
It is wonderful. Much love!

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
```diff
@@ -434,297 +410,187 @@ function UIHotkeyAssign:UIHotkeyAssign(ui, mode)
...
+        title = _S.hotkey_window.panel_debugKeys,
+        keys = {
+          "global_connectDebugger",
+          "global_showLuaConsole",
+          "global_runDebugScript",
+          "ingame_showCheatWindow",
+          "ingame_poopLog",
+          "ingame_poopStrings" }}}})
+  end
+
```
lewri, line 563: You've got indentation inconsistencies here. Some are 2, some are 4.

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
```diff
@@ -434,297 +410,187 @@ function UIHotkeyAssign:UIHotkeyAssign(ui, mode)
...
+    {
+      id = "misc",
+      title = _S.hotkey_window.panel_miscInGameKeys,
+      tooltip = _S.hotkey_window.panel_miscInGameKeys,
+      sections = {{
+          title = _S.hotkey_window.panel_miscInGameKeys,
+          keys = {
+            "ingame_rotateobject",
+            "ingame_patient_gohome",
+            "ingame_setTransparent" }}}},
```
lewri, line 479: imo this should be directly below "General In-Game Keys" for consistency.

TheCycoONE:
Alright, hopefully formatted a little better now.

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
TheCycoONE: Done

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
TheCycoONE: Fixed.

As far as I can tell nothing actually requires or even uses @type - but it's consistent.

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
TheCycoONE: Fixed

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
TheCycoONE: Convince me? I was thinking it might belong at the bottom or between Recall and Debug (all of the keys except Global are in-Game - but I didn't make up the categories)

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
TheCycoONE: Changed to key_opt_set

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
lewri: It could also be at the bottom sure.
The reason behind that is game keys and misc game keys are going to be at least somewhat related. Putting them next to each other would make more sense .

However, if you're treating them like a general miscellaneous category put them at the bottom.

lewri:
There's some kind of bug.

I imagined to click a button once to toggle, then it broke and wouldn't let me click any others.
I can only reliably click on Panel keys.

Saying that, I think the "clicked" effect needs to be a bit more obvious... I can't tell easy when I'm editing a button (can raise a new issue though for that if you want)

TheCycoONE:
Can you figure out what you did? Was there any error in the console or gamelog.txt?

You typically hit a situation where you can't click anything if there is an error in the buttonup handler, but it's possible there's another cause.

lewri:
I was click clicking the keys to set them. No errors.

lewri:
[screen-recording(2).zip](https://github.com/CorsixTH/CorsixTH/files/6775671/screen-recording.2.zip)

TheCycoONE:
Oh wow, that would be a problem

TheCycoONE:
And we're back - apparently we never use hidden windows and they still swallowed mouse clicks. No more.

Alberth289346:
Bummer, no more magic clicking at hidden windows :p 

Looks good to me!

lewri:
You have another bug:

1. Select a key to change
2. Navigate to a new panel
3. Press a key
4. Key is updated on the previous panel rather than cancelling the input.

TheCycoONE:
That sounds pre-existing but I'll address it.

lewri (approved):

lewri:
I can't break this anymore. So happy if you want to merge it.

PR #1947, opened 2021-06-20
Update check-for-updates file for 0.65

tobylane:
- Copy the legacy check-for-updates file to the new one
http://www.corsixth.com/check-for-updates


TheCycoONE:
Thanks. I meant to ask

PR #1954, opened 2021-06-23
Switch draw order of walls when north wall continues to next row.

### LINKED ISSUE #1953: [Bug] Sometimes 1px of walls slightly clip through intersecting walls
#### Describe the issue

Sometimes at certain zoom levels (with direct_zoom = true) 1px of the west-most edge of a wall facing SE can clip through the intersecting SW facing wall it adjoins.

#### Steps to Reproduce
1. Set `direct_zoom = true` in config.txt
 
2. Load the attached save.

3. Examine the walls connecting to the southwest edge of the hospital as you move the viewport around.

#### Save Game
[test.zip](https://github.com/CorsixTH/CorsixTH/files/6698115/test.zip)

#### Expected Behavior

Expect the interior walls not to show at all through the exterior.

#### System Information
**CorsixTH Version:** *0.6.5*

**Operating System:** *Linux*

#### Gamelog.txt
N/a

### Additional Info
wall-clipping

flackr:
This fixes #1953 by ensuring that the north wall face edge is drawn on
top of the next west wall face edge.

*Fixes #1953*

- Removes right-to-left pass drawing of north walls, drawing all walls in left-to-right order as it is impossible for north walls to overlap each other in scan order so this right-to-left ordering should be unnecessary.
- When the north wall adjoins with a wall in the next row (i.e. map x + 1), draw the west wall from the next tile first, so that the north wall edge will be on top of it to seamlessly join with the next row.

### DIFF
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -1017,12 +1017,50 @@ void level_map::draw(render_target* pCanvas, int iScreenX, int iScreenY,
       continue;
     }
 
-    for (map_scanline_iterator itrNode(
-             itrNode1, map_scanline_iterator_direction::backward, iCanvasX,
-             iCanvasY);
-         itrNode; ++itrNode) {
+    map_scanline_iterator itrNext(
+        itrNode1, map_scanline_iterator_direction::forward, iCanvasX, iCanvasY);
+
+    if (!bFirst) {
+      // since the scanline count from one THMapScanlineIterator to
+      // another can differ synchronization between the current iterator
+      // and the former one is needed
+      if (itrNext.x() < -64) {
+        ++itrNext;
+      }
+      while (formerIterator.x() < itrNext.x()) {
+        ++formerIterator;
+      }
+    }
+    bool bPreviousTileNeedsRedraw = false;
+    bool predrawNextWestWall = false;
+    for (; itrNext;) {
+      map_scanline_iterator itrNode = itrNext;
+      ++itrNext;
+      bool bNeedsRedraw = false;
       int iH;
-      unsigned int iBlock = itrNode->iBlock[1];
+      // If we drew the next west wall in the previous iteration, skip it in
+      // this iteration.
+      bool skipWestWall = predrawNextWestWall;
+      // If we have a next tile, we are not at the maximum x value so it is safe
+      // to get the adjacent tile.
+      const map_tile* adjacent =
+          itrNext ? itrNode.get_adjacent_tile() : nullptr;
+
+      // If the north wall continues in the next tile, we must draw it on top of
+      // the next west wall to ensure it is seamless so we draw the next west
+      // wall first.
+      predrawNextWestWall = adjacent && adjacent->iBlock[1];
+      unsigned int iBlock;
+      if (predrawNextWestWall) {
+        iBlock = itrNext->iBlock[2];
+        if (iBlock != 0 &&
+            blocks->get_sprite_size(iBlock & 0xFF, nullptr, &iH) && iH > 0) {
+          blocks->draw_sprite(pCanvas, iBlock & 0xFF, itrNext.x() - 32,
+                              itrNext.y() - iH + 32,
+                              (iBlock >> 8) | thdf_nearest);
+        }
+      }
+      iBlock = itrNode->iBlock[1];
       if (iBlock != 0 && blocks->get_sprite_size(iBlock & 0xFF, nullptr, &iH) &&
           iH > 0) {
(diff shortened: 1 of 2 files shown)


CorsixTH/Src/th_map.h
```diff
@@ -564,6 +564,7 @@ class map_scanline_iterator {
   inline int y() const { return y_relative_to_screen; }
   inline const map_tile* get_next_tile() { return tile + tile_step; }
   inline const map_tile* get_previous_tile() { return tile - tile_step; }
+  inline const map_tile* get_adjacent_tile() { return tile + 1; }
```
TheCycoONE, line 567: Why did you need this? `tile_step` is +1 or -1 depending on whether it's a forward or backward iterator, so `get_next_tile` or `get_previous_tile` are equivalent and more descriptive.

CorsixTH/Src/th_map.cpp
```diff
@@ -1017,12 +1017,50 @@ void level_map::draw(render_target* pCanvas, int iScreenX, int iScreenY,
...
+      bool bNeedsRedraw = false;
       int iH;
-      unsigned int iBlock = itrNode->iBlock[1];
+      // If we drew the next west wall in the previous iteration, skip it in
+      // this iteration.
+      bool skipWestWall = predrawNextWestWall;
+      // If we have a next tile, we are not at the maximum x value so it is safe
+      // to get the adjacent tile.
+      const map_tile* adjacent =
+          itrNext ? itrNode.get_adjacent_tile() : nullptr;
```
TheCycoONE, line 1047: Just itrNext.get_tile(), no reason to grab an adjacent one from itrNode.

CorsixTH/Src/th_map.cpp
```diff
@@ -1048,31 +1086,14 @@ void level_map::draw(render_target* pCanvas, int iScreenX, int iScreenY,
         pItem->draw_fn(pItem, pCanvas, itrNode.x(), itrNode.y());
```
TheCycoONE, line 1086: You're no longer drawing the earlyEntities on a previous reverse pass... I suspect there might be fallout, need to pay attention to interactions closely.

TheCycoONE (changes requested):
Iterator changes appear unnecessary.

I also suspect the backwards pass could be removed - but in the past when I've tried there have been lots of side effects. If it was intended, we'll need to spend some time watching.

CorsixTH/Src/th_map.h
flackr: Not true, tile_step is get_width() - 1 or -get_width() + 1, per initialization here https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Src/th_map.cpp#L1781 . forwards has value 2 and backwards 0 from this enum: https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Src/th_map.h#L440

The tile_step steps to x + 1, y - 1 in the fowards direction (i.e. horizontally to the right but at the same vertical height) or to x - 1, y + 1 (i.e. horizontally to the left at the same vertical height).

CorsixTH/Src/th_map.cpp
flackr: See response above, the scanline iterator steps to x + 1, y - 1, to maintain the same vertical position on screen. The tile I am interested in is the one at x + 1 to see if the wall on the current tile continues in that direction.

CorsixTH/Src/th_map.cpp
flackr: Random thought ... should we make a test suite? We could render some simple map test cases constructed to exercise various edge conditions. It would need some test image sprites so that it could render those into an image and compare to a reference screenshot.

This would help with confidently landing rendering changes. Maybe I'm overengineering this though :-).

TheCycoONE: It would help a lot; not a small ask though?

flackr: Yeah, it would be a large undertaking. I'm not suggesting we block this on it just maybe something to plan. It's challenging to have to worry about a bunch of cases you don't know about. A somewhat manual way of verifying might be to look at the bugs which were fixed by the change in draw order - if they have save files attached it could be an easy test.

flackr:
If you're nervous about this one, we could go with https://github.com/CorsixTH/CorsixTH/pull/1958 and/or I could put together a small PR to stretch the north walls by 1px (fixing it on non float rects too), since either of those fixes also fix the gap at the very left corner of the building that sometimes shows up currently. See the picture.

corner-gap

flackr:
Closing this for now since I think it may actually break the sample case in https://github.com/CorsixTH/CorsixTH/wiki/Rendering#rendering-of-a-scanline and the underlying issue was fixed by the FRect PR.

PR #1958, opened 2021-06-23
Use SDL_RenderCopyF on SDL 2.0.10 and newer for better scaling

### LINKED ISSUE #1953: [Bug] Sometimes 1px of walls slightly clip through intersecting walls

flackr:
*Fixes #1953*

- When using SDL 2.0.10 or newer, uses SDL_FRect and SDL_RenderCopyF to improve scaling precision.
- When using SDL 2.0.9 or older, uses SDL_Rect and the enclosing integer SDL_Rect for compatibility.

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -749,8 +749,8 @@ void render_target::draw(SDL_Texture* pTexture, const SDL_Rect* prcSrcRect,
   SDL_FRect scaledDstRect;
   getScaleRect(prcDstRect, global_scale_factor, &scaledDstRect);
   if (iSDLFlip != 0) {
-    SDL_RenderCopyExF(renderer, pTexture, prcSrcRect, &scaledDstRect, 0, nullptr,
-                     (SDL_RendererFlip)iSDLFlip);
+    SDL_RenderCopyExF(renderer, pTexture, prcSrcRect, &scaledDstRect, 0,
+                      nullptr, (SDL_RendererFlip)iSDLFlip);
   } else {
     SDL_RenderCopyF(renderer, pTexture, prcSrcRect, &scaledDstRect);
   }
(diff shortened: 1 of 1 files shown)


CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -119,6 +119,33 @@ void getEnclosingScaleRect(const SDL_Rect* rect, double scale_factor,
...
+  dst_rect->h = static_cast<float>(rect->h * scale_factor);
+}
+
+#else
+
+// Prior to SDL 2.0.10, fallback to using the enclosing integer SDL_Rect for
+// scaled rendering.
+#define SDL_FRect SDL_Rect
+#define SDL_RenderCopyF SDL_RenderCopy
+#define SDL_RenderCopyExF SDL_RenderCopyEx
```
flackr, line 140: Not sure if you'd prefer we make up names e.g. SDL_SCALED_RECT, SDL_RENDER_COPY_SCALED, SDL_RENDER_COPY_SCALED_EX to avoid the presumption that where you see SDL_FRect you always have a floating point rect rather than it depending on sdl version. Let me know I'm happy to change it.

TheCycoONE: This is fine. I doubt we'll be doing much more work with older SDL versions.

flackr:
I think it's a bit cleaner to have a single function definition (now documented) and just switch the body out based on the SDL version so I've done that and tested on both sides of the SDL version to make sure everything works correctly.

PR #1960, opened 2021-06-24
Make jellyitus effect intermittent.

flackr:
- Adds an effect offset per animation so that patient effects are not in sync.
- Only applies jellyitus for the first M out of every N frames.

### DIFF
--- a/CorsixTH/Src/th_gfx.cpp
+++ b/CorsixTH/Src/th_gfx.cpp
@@ -900,7 +900,8 @@ bool animation_manager::hit_test(size_t iFrame, const ::layers& oLayers, int iX,
 void animation_manager::draw_frame(render_target* pCanvas, size_t iFrame,
                                    const ::layers& oLayers, int iX, int iY,
                                    uint32_t iFlags,
-                                   animation_effect patient_effect) const {
+                                   animation_effect patient_effect,
+                                   size_t patient_effect_offset) const {
   if (iFrame >= frame_count) {
     return;
   }
@@ -942,6 +943,7 @@ void animation_manager::draw_frame(render_target* pCanvas, size_t iFrame,
     animation_effect render_effect =
         (oElement.layer > 0 || oElement.layer_id > 0) ? patient_effect
                                                       : animation_effect::none;
+    size_t effect_ticks = game_ticks + patient_effect_offset;
     if (iFlags & thdf_flip_horizontal) {
       int iWidth;
       int iHeight;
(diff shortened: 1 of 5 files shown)


lewri:
I think I also noticed the speed of effect is slower than original. But not sure how much slower

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -1101,14 +1101,15 @@ void sprite_sheet::draw_sprite(render_target* pCanvas, size_t iSprite, int iX,
...
+        sin(static_cast<double>(effect_ticks % 15) / 15.0 * 2 * pi) * 50);
     int err = SDL_SetTextureColorMod(pTexture, 0, 205 + currentVariation, 0);
     if (err < 0) {
       throw std::runtime_error(SDL_GetError());
     }
   }
 
-  if (effect == animation_effect::jelly) {
+  // TODO: Adjust jelly frequency and duration.
+  if (effect == animation_effect::jelly && effect_ticks % 200 < 30) {
```
TheCycoONE, line 1112: 200 and 30 are better off as constants. jelly_effect_period, and jelly_effect_duration perhaps.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -1332,6 +1334,7 @@ animation::animation()
   hit_test_fn = THAnimation_hit_test;
   is_multiple_frame_animation_fn = THAnimation_is_multiple_frame_animation;
   patient_effect = animation_effect::none;
+  patient_effect_offset = rand();
```
TheCycoONE, line 1337: Discussion point: should this be the system random number generator or the game random number generator (the later is intended to be reseeded where it left off for deterministic saves, though it doesn't quite work out in reality)

TheCycoONE:
If using rand, #include<cstdlib>

I think it's being pulled in indirectly/coincidentally.

CorsixTH/Src/th_gfx.cpp
flackr: I don't think there's value in having the complexity to try to keep the patient_effect_offset the same. Given that we also are not preserving game_ticks you would not reload to the exact same animation position, however as a random effect I don't think it's much of a concern. After a reload it would be hard to tell the difference between the patient reaching the end / start of a new jelly phase or the offset changing.

flackr:
> I think I also noticed the speed of effect is slower than original. But not sure how much slower

If you have any concrete measurements I am happy convert those to the constants in the code. Right now it completes a cycle in 30 game ticks (see (effect_ticks % 30) * 30). From the video it looks like I should make the effect run much longer before turning off. Does anyone have a convenient way to test this in the original?

I think I should be able to reproduce the exact size of the jiggle (currently +/-2px) and frequency of the waves (currently cycles every 14px) from a screenshot.

CorsixTH/Src/th_gfx_sdl.cpp
flackr: Done, also updated these to be similar to what i observed in the original.

flackr:
> If using rand, `#include<cstdlib>`
> 
> I think it's being pulled in indirectly/coincidentally.

Done.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -1332,6 +1334,7 @@ animation::animation()
   hit_test_fn = THAnimation_hit_test;
   is_multiple_frame_animation_fn = THAnimation_is_multiple_frame_animation;
   patient_effect = animation_effect::none;
```
TheCycoONE, line 1336: Several things here that probably belong in the initialization list - but none of them new in this PR. Just an observation.

PR #1968, opened 2021-06-25
[RDY] Stop cheat menu slide

### LINKED ISSUE #1647: Cheats Dialog/Menu can slide off screen
#### Describe the issue
If you move the screen with the cheats menu open, it can occasionally clip off the screen when it is closed and re-opened. Happens both in windowed and full-screen mode.

#### Steps to Reproduce
1. Open cheat menu. Scroll up the screen a bit

2. Close cheat menu, then re-open it.

3. If fails, repeat 1 + 2

#### Expected Behavior
Menu should retain its previous location when closed and reopened. Or at least re-centered to its default position.

#### System Information
**CorsixTH Version:** Trunk 139 (+ 0.64rc1)

**Operating System:** W10

### Additional Info
image

This might also happen with the debug patient dialog, but haven't tested it.

lewri:
*Fixes #1647*

- Moves the time we call ``setDefaultPosition`` on UICheats to account for dynamically added panels and buttons.
- This means that the window will always stay on screen when it recalls previous position.

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
+++ b/CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
@@ -70,7 +70,7 @@ function UICheats:UICheats(ui)
   self.modal_class = "cheats"
   self.esc_closes = true
   self.resizable = false
-  self:setDefaultPosition(0.2, 0.4)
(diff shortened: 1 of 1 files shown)


TheCycoONE (approved):
After setSize in particular but yes.

PR #1970, opened 2021-06-26
th_movie: Attempt to resample for surround sound

### LINKED ISSUE #1956: Intro movies incompatible with THX Spatial Audio
On a new headset I got it comes with THX Spatial Audio enabled. However, it looks like it gets globally enabled when on (not app specific).

When it is on CorsixTH's intro music becomes insanely distorted and no idea why. The MIDI music in the game is fine though.
Turning THX off and relaunching the game appears to fix it, but only until it is turned back on again.
Happening across 0.64 and 0.65

TheCycoONE:
*Fixes #1956*

https://wiki.libsdl.org/SDL_AudioSpec#remarks suggests that SDL2 can return 1, 2, 4, or 6 channels.

We're using https://www.libsdl.org/projects/SDL_mixer/docs/SDL_mixer.html#SEC15 where the documentation says it only returns 1 or 2, but I think it's lying.

### DIFF
--- a/CorsixTH/Src/th_movie.cpp
+++ b/CorsixTH/Src/th_movie.cpp
@@ -24,6 +24,8 @@ SOFTWARE.
 
 #include "config.h"
 
+#include <libavutil/channel_layout.h>
+
 #include "lua_sdl.h"
 #if defined(CORSIX_TH_USE_FFMPEG) && defined(CORSIX_TH_USE_SDL_MIXER)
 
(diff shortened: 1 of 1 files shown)


lewri (commented):
Working fine in current state.
Just note this warning generates that I already mentioned:
``WARN: unsupported channel layout 8. Please report issue.``

lewri (approved):

PR #1972, opened 2021-06-26
Only display dollar sign on local hospital patients

tobylane:
- Check the hospital before drawing the dollar.


mugmuggy:
One less thing to include in the AI Hospital. Was fun when I first started the AI Hospital one, I used normal patient functions, and turned off the limit camera and I had all these stuck patients and floating dollar signs.

PR #1973, opened 2021-06-26
Make disease discovery advice local player only

tobylane:
- Disease discovery fax and casebook updates are local player only.


mugmuggy:
And another for the AI Hospital

PR #1975, opened 2021-06-27
[RDY] Fullscreen Hotkey updates options button

### LINKED ISSUE #1955: [Bug] Fullscreen hotkey breaks settings toggle
#### Describe the issue
When using Alt+Enter to switch between fullscreen and windowed -- the Fullscreen setting won't update if the settings menu is open. As a result the button starts operating in reverse and the clicked/unclicked doesn't match up anymore.

Found while looking at #1637

#### Steps to Reproduce
1. Open settings menu

2. Press Alt+Enter to toggle fullscreen.

3. Button doesn't update -- pressing the Fullscreen button in settings then causes the behaviour to flip (thinks it is on when it's now in windowed mode)

#### Expected Behavior
Toggle the button correctly.

#### System Information
**CorsixTH Version:** 0.65 beta-2

#### Gamelog.txt

```
Paste gamelog.txt output here
```

### Additional Info
image

lewri:
*Fixes #1955*

- When using the fullscreen hotkey, the fullscreen button will now update if the Options menu was option.
- The solution feels a little janky, but happy to accept other ideas.

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/options.lua
+++ b/CorsixTH/Lua/dialogs/resizables/options.lua
@@ -292,12 +292,17 @@ function UIOptions:selectResolution(number)
   end
 end
 
-function UIOptions:buttonFullscreen(checked)
-  if not self.ui:toggleFullscreen() then
+function UIOptions:buttonFullscreen()
+  if not self.ui:toggleFullscreen(true) then
       local err = {_S.errors.unavailable_screen_size}
       self.ui:addWindow(UIInformation(self.ui, err))
       self.fullscreen_button:toggle()
   end
+  self.fullscreen_panel:setLabel(self.ui.app.fullscreen and _S.options_window.option_on or _S.options_window.option_off)
+end
+
+function UIOptions:updateFSButtonOnHotkey()
+  self.fullscreen_button:toggle()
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -292,15 +292,20 @@ function UIOptions:selectResolution(number)
   end
 end
 
-function UIOptions:buttonFullscreen(checked)
```
lewri, line 295: Just to note, ``checked`` was never used as far as I can see in the function.

CorsixTH/Lua/ui.lua
```diff
@@ -720,6 +723,11 @@ function UI:toggleFullscreen()
     -- Save new setting in config
     self.app.config.fullscreen = self.app.fullscreen
     self.app:saveConfig()
+    -- If options window is open, make sure the button updates on hotkey
+    if not from_uioptions and self:getWindow(UIOptions) then
```
lewri, line 727: The second param of the if is more a safeguard, but happy to remove it if not needed.

TheCycoONE: This is a bit of a violation again, you shouldn't care where a function was called from generally.

window:updateFSButtonOnHotkey() should be in the appropriate caller instead.

Issue #1935, opened 2021-06-14
Allow music control from main menu


lewri:
**What feature or enhancement would you like to add?**
In CorsixTH on first launch, music will play at full volume with no proper control to stop it until you start a game (only global audio). -- Plus 50% music vol is blazingly loud somehow
Instead, a toggle or setting should added to turn music on/off in main menu settings.

**How would this feature benefit the gameplay of CorsixTH?**
This would prevent blaring music for users on launch who don't want it.

**Was the feature in the original game?**
(Technically no, but music didn't happen in main menu on original)

**What version of CorsixTH are you using?**

0.65 beta-2

**Does this feature/enhancement relate to a current problem?**
No

**Additional information**
Add any other information or screenshots about the feature request here.

lewri:
As pointed out, it may be better to remove playing music from the menus.
I suppose you could also add a Jukebox to the main menu too

TheCycoONE:
There was some opposition on discord to that idea. I think the real takeaway here is to default the volume < 100%

lewri:
Thanks, probably set music vol to 10% or 20% then.

Though I'm not sure how you really update the config.txt code

TheCycoONE:
Update config_finder.lua to set the default. It doesn't affect people that have already launched the game.

lewri:
@tobylane does your jukebox change to add it to settings now resolve this? (I think it does enough -- and config finder sets 50% already)

tobylane:
Yes.

PR #1977, opened 2021-06-27
[NRDY] Lower the music *held*

lewri:
Resolves part of #1935

- Set default music to 0.2 instead of 0.5 for first launch.


lewri:
If 1978 is actually what caused the blaring in the first place this PR might not be needed.

TheCycoONE:
Let's fix 1978 and see where it takes us.

TheCycoONE:
This isn't actually [RDY] is it? I keep coming back to it, but iirc it doesn't solve the problem?

lewri:
Not really anything at the moment -- but there's nothing to do with it either.
Should change it to the rare [NRDY]

lewri:
On some additional testing, the adjustment is still needed. Just no benefit using it yet.

lewri:
Closing for now awaiting 1978

PR #1979, opened 2021-06-27
Use intermediate buffer for floating dollar sign spritesheet.

### LINKED ISSUE #1966: [Bug] Money icon's text gets distorted
#### Describe the issue
If not playing at normal zoom (100%), the money inside the dollar bill when someone pays for services gets distorted.
It is not related to direct zoom.

#### Steps to Reproduce
1. Have a patient use a room with some zoom

2. Watch the dollar bill get distorted.

#### Expected Behavior
We want to keep the text crisp regardlelss of zoom.

#### System Information
**CorsixTH Version:** Trunk (Appveyor 585)

#### Gamelog.txt

```
Paste gamelog.txt output here
```

### Additional Info
Normal zoom
image

Zoom in/out
image

flackr:
*Fixes #1966*

- Add a parameter, use_intermediate_buffer to spritesheet.
- When set, drawing the spritesheet creates an intermediate unscaled buffer to draw to before scaling the result to the screen.
- Sets this parameter for the floating dollar sign spritesheet which contains transparent bitmap character sprites which do not preserve text fidelity when scaled independently without their background.

Before patch:
money-master

After patch:
money-patch

We may be able to use the same technique to draw menu fonts nicely without creating a fullscreen zoom_texture as we do currently.


CorsixTH/Lua/world.lua
```diff
@@ -1903,6 +1903,7 @@ function World:newFloatingDollarSign(patient, amount)
   spritelist:setPosition(-17, -60)
   spritelist:setSpeed(0, -1):setLifetime(100)
   spritelist:setSheet(TheApp.gfx:loadSpriteTable("Data", "Money01V"))
+  spritelist:setUseIntermediateBuffer()
```
lewri, line 1906: Existing floating dollars likely would need an afterLoad to fix them. But this could be more trouble than it's worth

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -39,9 +39,12 @@ SOFTWARE.
 
 #include "th_map.h"
 
-#if !SDL_VERSION_ATLEAST(2, 0, 10)
+#if SDL_VERSION_ATLEAST(2, 0, 10)
```
lewri, line 42: Do we still see improvement in older SDL?

lewri (commented):
$100 well spent for a confused doctor :)

CorsixTH/Src/th_gfx_sdl.cpp
flackr: Yes, the same improvement. This change draws the floating dollar icon and text into it's native size before scaling it up / down to the screen. I.e. supposing its sprite size is 80x50, it draws the text and icon into a texture at that size and then scales the entire thing. This is the same idea as how https://github.com/CorsixTH/CorsixTH/pull/1934 preserved text quality for full screen but applied on the small scale of an individual sprite rather than the entire screen.

CorsixTH/Lua/world.lua
flackr: Right, and persist support which I'm not sure if we should do just yet. I'm open to opposing viewpoints, but IMO what we should be doing eventually is drawing TTF text which would negate the need for this bit. I'm not sure how CorsixTH supports loading old saves with persist / depersist and I'm worried about adding temporary baggage on the save file format.

lewri: Okay then. +1 to just dropping any afterload. The new code will be picked up on any following dollar sign generations.

TheCycoONE: I'm opposed to adding an afterLoad to fix such a slight transient visual artifact.

PR #1981, opened 2021-06-28
Add volume control buttons to jukebox

tobylane:
- Add music volume up and down buttons to the jukebox. The other buttons are moved closer together to fit this in.

Currently, I'm reusing the skip back/forward sprites. What's the way to add/draw new buttons and is it worth it for this function?


tobylane:
This is (half of) a solution without a problem. I'll shelve it until users want it.

PR #1985, opened 2021-06-30
Allow any disease to be delayed

### LINKED ISSUE #1982: Allow delay on all diseases
Theme Hospital only allowed a select number of diseases to be held back until x months. However, some diseases not listed (e.g. Golf Stones) may be desirable in custom levels to hold back x months.
From the way the .level configs look this should be doable.

This would provide more customisation in custom levels with disease control.

No

 0.65 beta-2

No

**Additional information**
Ping @lloyd-lejack

lewri:
*Fixes #1982*

- Non-visual diseases couldn't be delayed in the original. This PR updates this so any disease can be delayed in a custom level. 
- PR is still backwards compatible with 0.65 etc if a custom-level maker included the new variables in a ``.level`` -- they'll be not used and ignored

``.level`` configuration code. Insert below ``#visuals_available`` if used (or ``#non_visuals`` if not)
```sam
This is a new feature from the original game, by default all non-visual diseases are available immediately
in the original.
Defines which month a non-visual illness becomes available (if it is going to be available is defined
above in non-visuals). 0 means available immediately.

#non_visuals_available[0] 0 I_UNCOMMON_COLD
#non_visuals_available[1] 0 I_BROKEN_WIND
#non_visuals_available[2] 0 I_SPARE_RIBS
#non_visuals_available[3] 0 I_KIDNEY_BEANS
#non_visuals_available[4] 0 I_BROKEN_HEART
#non_visuals_available[5] 0 I_RUPTURED_NODULES
#non_visuals_available[6] 0 I_MULTIPLE_TV_PERSONALITIES
#non_visuals_available[7] 0 I_INFECTIOUS_LAUGHTER
#non_visuals_available[8] 0 I_CORRUGATED_ANKLES
#non_visuals_available[9] 0 I_CHRONIC_NOSEHAIR
#non_visuals_available[10] 0 I_3RD_DEGREE_SIDEBURNS
#non_visuals_available[11] 0 I_FAKE_BLOOD
#non_visuals_available[12] 0 I_GASTRIC_EJECTIONS
#non_visuals_available[13] 0 I_THE_SQUITS
#non_visuals_available[14] 0 I_IRON_LUNGS
#non_visuals_available[15] 0 I_SWEATY_PALMS
#non_visuals_available[16] 0 I_HEAPED_PILES
#non_visuals_available[17] 0 I_GUT_ROT
#non_visuals_available[18] 0 I_GOLF_STONES
#non_visuals_available[19] 0 I_UNEXPECTED_SWELLING
```

Ping @lloyd-lejack to test this out.


lewri:
I'm not sure if this warrants a config.txt option. In my opinion it does not -- we're not touching any characteristics of the original campaign. It is simply an add-on for custom levels.

lloyd-lejack:
@lewri I'll be doing several tests for this, probably during the next two weeks to see how it goes.

Alberth289346:
> I'm not sure if this warrants a config.txt option. In my opinion it does not -- we're not touching any characteristics of the original campaign. It is simply an add-on for custom levels.

I agree with not needing a user setting for this.

TheCycoONE (approved):
Looks solid, one line that should be impossible to reach that I think should error instead.

CorsixTH/Lua/world.lua
```diff
@@ -479,23 +479,28 @@ end
...
+  if disease.visuals_id then
+    -- level files can delay visuals to a given month
+    -- and / or until a given number of patients have arrived
+    local hold_visual_months = level_config.gbv.HoldVisualMonths
+    local hold_visual_peep_count = level_config.gbv.HoldVisualPeepCount
+
+    -- if the month is greater than either of these values then visuals will not appear in the game
+    if (hold_visual_months and hold_visual_months > self.game_date:monthOfGame()) or
+        (hold_visual_peep_count and hold_visual_peep_count > hospital.num_visitors) then
+      return false
```
lewri, line 492:
```suggestion
      return error("disease has neither visuals_id or non_visuals_id")
```

lewri: Actually, won't returning an error cause lua to think it returned true? Or will it still pick up it was false?

mugmuggy: L503 is what Cyco is referring to. As you elseif the non-visual there is a potential path that is unlikely to hit

lewri: Yep got the line wrong. Though the question remains

Alberth289346: `error` never returns, it crashes the program with an error.
So no "return", just `error("crash message reason")`

CorsixTH/Lua/world.lua
```diff
@@ -479,23 +479,28 @@ end
...
+  elseif disease.non_visuals_id then
+    -- Non-visual disease. Simply check the month it can appear from has passed
+    -- 0 means it can show up anytime.
+    return level_config.non_visuals_available[disease.non_visuals_id].Value < self.game_date:monthOfGame()
   end
-
-  -- The value against #visuals_available determines from which month a disease can appear.
-  -- 0 means it can show up anytime.
-  return level_config.visuals_available[disease.visuals_id].Value < self.game_date:monthOfGame()
+  return false
```
lewri, line 503:
```suggestion
  error("disease has neither visuals_id or non_visuals_id")
```

lewri:
Error added. Will wait on this one for LLoyd to finish testing it.

PR #1986, opened 2021-07-01
[RDY] Spawn changes

### LINKED ISSUE #649: First patient takes too long to appear
1. Start a new game
2. Open the hospital right away (green timer button) at 1st of Jan
3. Build a reception

In the original TH the first patient would come to the hospital around 11th of Jan
In CorsixTH it may take forever for the first patient to appear. Initially I thought the game was broken as no one was coming for two months.

I have comapred the date when the first patient appears in the lower left corner as shown on the screenshot (this is the furthest spot i can obsrerve in the original)

screenshot3

Original:
12 Jan; 11 Jan; 11 Jan; 11 Jan
CorsixTH:
21 Jan; 2 Feb; 8 Mar; 8 Feb; 28 Jan

9cc1254

Win7 x64, buit with MSVC2013CE, SDL 2.0.3, SDL_mixer 2.0.0, Lua 5.1.5, all defaults

Full level 1

mugmuggy:
*Partially Fixes #649*

- Laying the foundation for AI Hospitals, hence the AIHospitals change for hasStaffedDesk
- Hospital new functions to support reputation based allocation of patients to hospitals
- World - even though you don't have a staffed desk, TH still counts them towards the statistics, even though no patients are actually spawned.
- World - assign the disease to the patient, choose the hospital based on the patients disease and spawns the patient to the chosen hospital with the selected disease
- World - spawn hours align with the 0-49 rather than 1-50 so spawns actually occur the day they should
- World - onEndMonth removes unnecessary logic, though deprecates getReputationImpact (I have left that in for now maybe I should remove it) - popn config, for the initial start, the first value is used and then at the end of the month the value is used again. L1 6 spawns in Feb.
- World - update spawn dates removes the math.n_random/math.ceil possibility of returning a number < 1, more closely aligns with TH - mostly fixes #649
- World - new functions getNewPatientDisease (refactored from spawnPatient) - chooseHospitalForNewPatient, introduces AllocRand in to the mix
- Progress report tracks population changes in real time, also allows enables switching between the different hospitals - left unimplemented comment on the tooltip for now, as AIs aren't implemented

~~Bonus of this one, turn off limit camera and the patients for the AI Hospitals will be strolling around on the edges of the map~~


lewri:
>Bonus of this one, turn off limit camera and the patients for the AI Hospitals will be strolling around on the edges of the map

I feel sorry for those lost little lambs

(Seriously though, I'll try to look over this tomorrow evening)

CorsixTH/Lua/world.lua
```diff
@@ -2867,3 +2850,64 @@ function World:setCampaignData(campaign_data)
...
+    local reputation = hosp:getDiseaseReputation(disease.id)
+    value_proposition = hosp:getDiseaseReputationPriceFactor(disease.id)
+    value_proposition = value_proposition * rep_multiplier * reputation
+    if s > 1 then
+      value_proposition = value_proposition + hosp_patient_value[s - 1]
+    end
+    hosp_patient_value[s] = value_proposition
+  end
+
+  local threshold = math.random(0, math.floor(value_proposition - 1))
```
TheCycoONE, line 2907: Nice, though clever enough it deserves a comment (or the comment above expanded) i.e. to include that the hospital is chosen randomly with a chance proportional to it's relative value proposition.

TheCycoONE (commented):
Nice PR. Not sure `World - even though you don't have a staffed desk, TH still counts them towards the statistics, even though no patients are actually spawned.` is actually the right thing to do, but for fidelity I suppose it can stay.

I agree you should remove getRepuationImpact.

CorsixTH/Lua/world.lua
```diff
@@ -2867,3 +2850,64 @@ function World:setCampaignData(campaign_data)
...
+    if #usable_diseases == 0 then return nil end
+
+    disease = usable_diseases[math.random(1, #usable_diseases)]
+  end
+  return disease
+end
+
+--! Choose hospital for normal patient
+--!param disease (table) - the disease assigned to this patient
+--!return hospital  (Hospital)
```
lewri, line 2878: Accidental double space?

CorsixTH/Lua/world.lua
lewri: Seconding this... I think

lewri (commented):
Looks fine to me. Hoping the edge spawned patients will be hidden in another patch

CorsixTH/Lua/world.lua
mugmuggy: Yeah the hasStaffedDesk, they don't spawn, but they do count to the visitor count on the annual report but not on the charts/graphs ui, so maybe I should adjust that also. Though I think that will introduce another statistic. Looks like I could repurpose the World.total_population and Hospital.population variables to account for this.

CorsixTH/Lua/hospitals/ai_hospital.lua
```diff
@@ -46,6 +46,10 @@ function AIHospital:logTransaction()
   -- AI doesn't need a log of transactions, as it is only used for UI purposes
 end
 
+function AIHospital:hasStaffedDesk()
```
lewri, line 49: Possible future todo?:
Adjust this to only be true when the AI Hospital opens

mugmuggy: Indeed this will switch with the AI hospitals change to actually detect a receptionist has been hired and a desk exists again. This just gets it working before they are implemented

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -159,92 +157,97 @@ function UIProgressReport:drawMarkers(canvas, x, y)
...
-  local ly = 73
+  local ly = 73 * s
   for pnum, player in ipairs(world.hospitals) do
     local font = (pnum == self.selected) and self.red_font or self.normal_font
-    font:draw(canvas, player.name:upper(), x + 272, y + ly)
-    ly = ly + 25
-    total_spawns = total_spawns + math.max(1, player.population)
+    font:draw(canvas, player.name:upper(), x + 272 * s, y + ly)
+    ly = ly + 25 * s
+    total_spawns = total_spawns + math.max(1, player.population)
```
ARGAMX:
```suggestion
    total_spawns = total_spawns + math.max(1, player.population)
```

CorsixTH/Lua/world.lua
```diff
@@ -2927,4 +2738,9 @@ function World:chooseHospitalForPatient(disease)
       return self.hospitals[i]
     end
   end
+end  
+
```
ARGAMX:
```suggestion
end

```

CorsixTH/Lua/world.lua
```diff
@@ -2712,6 +2675,72 @@ function World:validateMap()
...
+  end
+
+  -- Check random value, against individual hospital thresholds
+  local threshold = math.random(0, math.floor(value_proposition - 1))
+  for i, hosp_value in ipairs(hosp_patient_value) do
+    if threshold < hosp_value then
+      return self.hospitals[i]
+    end
+  end
+end
```
ARGAMX:
```suggestion
```

ARGAMX (approved):
I think this PR needs to be merged to move forward with competing hospitals' functionality.

lewri:
The PR is probably a bit dusty in itself. It might be more productive and subject to better scrutiny  to pull it apart into smaller patches.

Read: Not an objection to get stuff merged from this.


Though, I'm still a little concerned about:
> Bonus of this one, turn off limit camera and the patients for the AI Hospitals will be strolling around on the edges of the map

It's not a great approach to render the entities.

ARGAMX (dismissed):
.

ARGAMX:
I tried this PR today.
I hadn’t done this before, but today I set it up for myself to test it.
After that I started the default campaign. Level 1.

Apparently, this PR was supposed to improve patient spawning.
But it seems to me that something went wrong.

Screenshot 2026-09-07 at 11 57 33

ARGAMX:
> > Bonus of this one, turn off limit camera and the patients for the AI Hospitals will be strolling around on the edges of the map
> 
> It's not a great approach to render the entities.

I couldn’t register this behavior. 
This doesn’t happen.
No one extraneous walks along the edges of the map.
Patients from other hospitals simply don’t appear on the map.
Possibly mugmuggy may have fixed this after he wrote it.
So everything is fine regarding this.

ARGAMX:
I also found some bugs. For example, patients never spawn on the first day of the month. Due to corner case.

CorsixTH/Lua/world.lua
```diff
@@ -1190,7 +1174,8 @@ function World:updateSpawnDates()
   local local_hospital = self:getLocalPlayerHospital()
 
   -- Decide on number of visitors
-  local no_of_spawns = self.spawn_rate * local_hospital.population
+  -- Use ceil so that at least one patient arrives (unless population = 0)
+  local no_of_spawns = math.ceil(self.spawn_rate * local_hospital.population)
```
ARGAMX, line 1179: Why we multiply `spawn_rate`, which is already increasing every month, by `hospital.population`?
`hospital.population` is the sum of all `hospital.num_visitors` 
>   self.population = 0 -- Count of total attempted spawned patients

With this multiplication, we get explosive growth in the number of visitors after a few months.

Screenshot 2026-09-07 at 11 57 33

I have no choice but to remove `hospital.population` from here.

CorsixTH/Lua/hospital.lua
```diff
@@ -132,9 +132,7 @@ function Hospital:Hospital(world, avail_rooms, name)
   self.num_vips = 0 -- used to check if it's the user's first vip
   self.percentage_cured = 0
   self.percentage_killed = 0
-  self.population = 0.25 -- TODO: Percentage showing how much of
-  -- the total population that goes to the player's hospital,
-  -- used for one of the goals. Change when competitors are there.
+  self.population = 0 -- Count of total attempted spawned patients
```
ARGAMX, line 135: ok, now I see where the confusion with `population` comes from.

ARGAMX: Judging by how `population` is described in the original game and how it works there, `population` is exactly what it was before it was reworked in this PR.
Moreover, using the same variable name requires an `afterLoad`.
I think I’ll have to rework it a bit back.

Screenshot 2026-09-15 at 15 36 29

ARGAMX:
It’s time to deal with this PR.
PR refreshed and ready for review. 
Please check it.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -203,12 +203,16 @@ function UIProgressReport:draw(canvas, x, y)
   local world    = self.ui.app.world
   local hospital = world.hospitals[self.selected]
 
+  -- for population calculation
+  local total_visitors = 0
+
   -- Names of the players playing
   local ly = 73 * s
-  for pnum, player in ipairs(world.hospitals) do
+  for pnum, player_hospital in ipairs(world.hospitals) do
```
tobylane: Is player hospital the right word here, it's all active hospitals including player and computer opponents?

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -245,8 +249,15 @@ function UIProgressReport:draw(canvas, x, y)
   self.normal_font:draw(canvas, _S.progress_report.header .. " " ..
       (world:date():year() + 1999), x + 227 * s, y + 40 * s, 400 * s, 0)
   self.small_font:draw(canvas, _S.progress_report.win_criteria:upper(), x + 263 * s, y + 172 * s)
+  -- population percent
+  -- TODO: `math.floor` may result in the total of all four hospitals not equal up to 100%.
+  print("# population", total_visitors, hospital.num_visitors_pm)
+  local population_percent = 100 / 4 -- default value: 25% for each hospital
```
tobylane: It may not be four, use #world.hospitals

CorsixTH/Lua/hospital.lua
```diff
@@ -122,7 +122,10 @@ function Hospital:Hospital(world, avail_rooms, name)
   }
   self:buildReceptionDesksCache()
 
-  self.num_visitors = 0 -- Counts patients only
+  self.spawn_attempts = 0 -- Count of total attempted spawned patients
+  self.num_visitors_tm = 0 -- Current month visitors count
```
tobylane, line 126: Expand tm and pm?

CorsixTH/Lua/world.lua
```diff
@@ -442,22 +444,15 @@ function World:spawnPatient(hospital)
...
-    for _, d in ipairs(self.available_diseases) do
-      if isDiseaseUsableForNewPatient(self, d, hospital) then
-        usable_diseases[#usable_diseases + 1] = d
-      end
+  -- Still need to count them even though you haven't got a staffed desk
+  -- and you also don't want to spawn them on the map
+  hospital.spawn_attempts = hospital.spawn_attempts + 1
+  if not hospital:hasReceptionDesk(true) then
+    if not hospital:isPlayerHospital() then
+      -- TODO: Count patients for AIHospital for now. But remove for AIHospital capable of building a reception desk.
```
tobylane: Not sure on this comment. What is the intended future change?

CorsixTH/Lua/world.lua
```diff
@@ -966,6 +961,7 @@ function World:onTick()
       -- TODO: Multiplayer support.
       local spawn_count = self.spawn_hours[self.game_date:hourOfDay() + i - 1]
       if spawn_count and self.hospitals[1].opened then
+        -- TODO: Should check target hospital, not hospital 1
```
tobylane, line 965: Is the potential spawn at the first hospital in turn order that has a staffed desk, or if the target hospital doesn't have a desk the world doesn't have that spawn?

tobylane (commented):
First pass review, play testing later.

CorsixTH/Lua/world.lua
ARGAMX: > Is the potential spawn at the first hospital in turn order that has a staffed desk, or if the target hospital doesn't have a desk the world doesn't have that spawn?

Both — no. 🙂
Spawn will only happen if the first hospital is open.
Other hospitals can’t open yet, and the reception desk can only exist in hospital 1.  
So for now, this can’t be fixed without further development in the “AI Hospitals” direction.

CorsixTH/Lua/hospital.lua
ARGAMX: > Expand tm and pm?

Some variables already use this name style.

```Lua
  self.num_vips_ty = 0
  self.pleased_vips_ty = 0
  self.num_cured_ty = 0
  self.not_cured_ty = 0
  self.num_visitors_ty = 0
  ```

One variable not use this name style.

```Lua
self.num_deaths_this_year = 0
```

At first, I wrote it in full. 
But in the end, I decided that for consistency, there should be a first version.

This is how `tm` came into being...

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
ARGAMX: > It may not be four, use #world.hospitals

Good find! Fixed.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
ARGAMX: > Is player hospital the right word here, it's all active hospitals including player and computer opponents?

Yes.
Initially, the word “player” was here.
But I thought it’s confusing.   
I renamed it to “hospital”.
But then the linter started complaining. 😀
I had to add something to make the error go away. 
After all these manipulations, I ended up with “player_hospital”. 
Apparently, the person who left “player” here had a problem with linter too. Now I understand this. 
But today I came up with a good idea — the linter is complaining because the name “hospital” is already taken.
I freed it up and used “hospital” here.
Now it should be better.

CorsixTH/Lua/world.lua
ARGAMX: > Not sure on this comment. What is the intended future change?

The count of patients is performed in `Hospital:addPatient(patient)`.

But for now, `addPatient` is not being called for AIHospitals. 

Because of this, the count didn’t work for them.
I had to add a count here so that counting would work for AIHospitals already today. 

And accordingly, I wrote this comment, which say that this call will need to be removed from here when AIHospitals are able to accept patients.

ARGAMX:
Today is a good day to Approve this or at least do a Review.

CorsixTH/Lua/date.lua
```diff
@@ -189,7 +200,7 @@ end
 
 --[[ Returns the hour of the day
 
-! Finds out what is an hour of this date starting on 0
+! Finds out what is an tick of this date starting on 0 up to 49
```
lewri: It is hour for this class' purpose. A day has 50 hours in this game. Defining one term and using another in the doc is not useful.

If truly bothersome it could be hour (tick)

CorsixTH/Lua/hospital.lua
```diff
@@ -1407,12 +1420,16 @@ end
 
 function Hospital:addPatient(patient)
   self.patients[#self.patients + 1] = patient
+  self:countNewPatient()
+  -- Decide if the patient belongs in an epidemic
```
lewri: Not you, you can ignore but it should be `to` and not `in`

CorsixTH/Lua/hospital.lua
```diff
@@ -2433,6 +2450,22 @@ end
...
+--! Return reputation value for disease or hospital
+--!param disease id (string) - name of the disease
+--!return reputation (int)
+function Hospital:getDiseaseReputation(disease)
+  return self.disease_casebook[disease].reputation or self.reputation
+end
+
+--! Calculate value to influence spawn chance based on the price relative to reputation
+--!param disease id (string) - name of the disease
+--!return (float) Multiplier to apply for spawn chance
```
lewri: Lua (or more, our usage) don't use float type. It is just number.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -126,7 +126,7 @@ function UIProgressReport:UIProgressReport(ui)
     self:addPanel(0, 265, 71 + (num - 1) * 25)
       :makeButton(0, 0, 147, 20, 9, btn_handler(num))
       :setTooltip(tooltip(num))
-      :enable(num == 1)
+      :enable(true)
```
lewri: This felt a bit more intuitive as `num == 1`. I think the idea here is that the num is the hospital selected in the report. However, it's still not implemented.

Otherwise this placeholder feels really clueless.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -245,8 +249,14 @@ function UIProgressReport:draw(canvas, x, y)
   self.normal_font:draw(canvas, _S.progress_report.header .. " " ..
       (world:date():year() + 1999), x + 227 * s, y + 40 * s, 400 * s, 0)
   self.small_font:draw(canvas, _S.progress_report.win_criteria:upper(), x + 263 * s, y + 172 * s)
+  -- population percent
+  local population_percent = 100 / #world.hospitals -- default value: 25% for each hospital
```
lewri: Comment assumes we always have 4 'players' (we do), but in reality we shouldn't make the assumption? Saying default is equal share per hospital/player is more flexible.

CorsixTH/Lua/hospital.lua
```diff
@@ -132,9 +135,6 @@ function Hospital:Hospital(world, avail_rooms, name)
   self.num_vips = 0 -- used to check if it's the user's first vip
   self.percentage_cured = 0
   self.percentage_killed = 0
-  self.population = 0.25 -- TODO: Percentage showing how much of
```
lewri, line 135: I'm not sure why it isn't useful to cache the population in each hospital. 
Well, that being said all hospitals must also make up to 1.

lewri: Another perhaps unrelated question is if reputation affects the actual population delivered across all hospitals. I.e. Could it be possible to not have 100% of the population in all hospitals if everyone's rep was low? Unless we're defining a population solely here as those in attendance rather than the fictional town.

CorsixTH/Lua/world.lua
```diff
@@ -3047,6 +3035,71 @@ function World:validateMap()
...
+    return self.hospitals[math.random(1, 4)]
+  end
+
+  -- From all hospitals randomly select a hospital based on the reputation
+  -- and cost of treatment of the patients disease
+  local value_proposition = 0
+  local hosp_patient_value = {}
+  -- calculate the threshold values for each hospital
+  for s, hosp in ipairs(self.hospitals) do
+    local rep_multiplier = self.map.level_config.gbv.AllocIndRep and self.map.level_config.gbv.AllocIndRep or 2
```
lewri: Value needs populating in base_config.lua (as well as the two above in this function). Potentially afterLoaded into map.lua too.

CorsixTH/Lua/hospital.lua
```diff
@@ -2433,6 +2450,22 @@ end
...
+function Hospital:getDiseaseReputation(disease)
+  return self.disease_casebook[disease].reputation or self.reputation
+end
+
+--! Calculate value to influence spawn chance based on the price relative to reputation
+--!param disease id (string) - name of the disease
+--!return (float) Multiplier to apply for spawn chance
+function Hospital:getDiseaseReputationPriceFactor(disease)
+  local reputation = self.disease_casebook[disease].reputation or self.reputation
+  local percentage = self.disease_casebook[disease].price
```
lewri, line 2465: Price == percentage? Is this one of those weird inconsistencies again?

lewri (commented):
The reason why I optimistically assigned this to myself months ago was to pull this apart into smaller PRs. As we don't live in mugmuggy's head, I didn't like the idea of second guessing a large patch.

In retrospect its not too bad.

------

Has this one been playtested well enough?

ARGAMX:
> The reason why I optimistically assigned this to myself months ago was to pull this apart into smaller PRs. As we don't live in mugmuggy's head, I didn't like the idea of second guessing a large patch.

I think I managed to get into mugmuggy head related this PR. 😁
I know why he opened this PR.
Once I understood that, I was able to refine it.

It’s obvious that he wanted to start implementing the competing hospitals. 
This PR - is very very first step in this way.
When I started working on competing hospitals functionality I came up with the idea to do the same things he did here.
So it was easier for me to push this PR through than to do the same things from scratch.

Actually, it’s not that big of a patch.

ARGAMX:
> Has this one been playtested well enough?

I played with it.
Overall, I liked it.
You should play too.

CorsixTH/Lua/hospital.lua
ARGAMX: > I'm not sure why it isn't useful to cache the population in each hospital. Well, that being said all hospitals must also make up to 1.

they are "caching" in `hospital.num_visitors_pm`

ARGAMX: > Another perhaps unrelated question is if reputation affects the actual population delivered across all hospitals. I.e. Could it be possible to not have 100% of the population in all hospitals if everyone's rep was low? Unless we're defining a population solely here as those in attendance rather than the fictional town.

Nope.

reputation affects the distribution
of patients, i.e. who gets how many patients. 
the distribution affects the population.

Let's say this month will be assigned 10 patients.
This is constant for month.
No matter what the reputation is, 10 people will come to hospitals anyway. 
The only question is which one.

So there will be no population less than 100%.
The population is always 100%. Even if no one came anywhere.

CorsixTH/Lua/world.lua
ARGAMX: > Value needs populating in base_config.lua (as well as the two above in this function). Potentially afterLoaded into map.lua too.

Can you please fix this, to make it better and push to this PR?
I haven’t touched this config part; it’s been here for 4 years in this state.
I don’t understand how the configs work at all.

CorsixTH/Lua/hospital.lua
ARGAMX: > Price == percentage? Is this one of those weird inconsistencies again?

Yes, this is not the price. This is a modifier for markup or discount.
Price is 1 mean 100%
Price is 1.20 mean 120%

CorsixTH/Lua/date.lua
ARGAMX: > It is hour for this class' purpose. A day has 50 hours in this game. Defining one term and using another in the doc is not useful.
> 
> If truly bothersome it could be hour (tick)

Well, yes, there’s actually some kind of problem with the abstraction.

There are 24 hours in a day. That’s obvious even to a schoolchild.
But here it’s being violated. So it’s not an hour. 
It’s kind of like if we called people dogs in code and dogs people — obviously, that would be confusing.

Ideally, this should be given some kind of special name.
It could be called corsixhour, or onefiftiethday or something like that.

Essentially, solving this problem isn’t this PR’s responsibility.
So, okay, I’ll put it back the way it was.

CorsixTH/Lua/hospital.lua
ARGAMX: > Not you, you can ignore but it should be `to` and not `in`

Changed 👌

CorsixTH/Lua/hospital.lua
ARGAMX: > Lua (or more, our usage) don't use float type. It is just number.

Fixed 👌

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
ARGAMX: > This felt a bit more intuitive as `num == 1`. I think the idea here is that the num is the hospital selected in the report. However, it's still not implemented.
> 
> Otherwise this placeholder feels really clueless.

Generally, yes, it’s too early to enable this switch. 
It allows you to select other hospitals. 
But it’s not useful yet, since the data for other hospitals isn’t ready yet. 
Only the population value.
I returned it to `:enable(num == 1)`.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
ARGAMX: > Comment assumes we always have 4 'players' (we do), but in reality we shouldn't make the assumption? Saying default is equal share per hospital/player is more flexible.

Updated the comment 👌

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -201,14 +201,18 @@ function UIProgressReport:draw(canvas, x, y)
...
+  -- for population calculation
+  local total_visitors = 0
 
   -- Names of the players playing
   local ly = 73 * s
-  for pnum, player in ipairs(world.hospitals) do
+  for pnum, hospital in ipairs(world.hospitals) do
     local font = (pnum == self.selected) and self.red_font or self.normal_font
-    font:draw(canvas, player.name:upper(), x + 272 * s, y + ly)
+    font:draw(canvas, hospital:upper(), x + 272 * s, y + ly)
```
tobylane:
```suggestion
    font:draw(canvas, hospital.name:upper(), x + 272 * s, y + ly)
```

tobylane (commented):
On the first campaign level my population share moved quickly, in June 12%, July 33%, August 42%. It was a good hospital, I won in October (still 42%), but is that rate of change expected?

ARGAMX:
> On the first campaign level my population share moved quickly, in June 12%, July 33%, August 42%. It was a good hospital, I won in October (still 42%), but is that rate of change expected?

Hmm, probably not.
mugmuggy set up a decent distribution system across the hospitals.
However, the other hospitals aren't treating anyone—that feature hasn't been implemented yet—so their reputation isn't growing. Yours, on the other hand, is.
As a result, the patient population is being drawn disproportionately toward you.
This creates an imbalance.
It seems the distribution needs to be scaled back for now so that it always allocates 25% to each.
Otherwise, you end up getting more patients simply for nothing.
I’ll set it at a fixed 25%.

CorsixTH/Lua/world.lua
```diff
@@ -3047,6 +3035,74 @@ function World:validateMap()
...
+  -- TODO: Until changes in competing hospitals' reputations are implemented,
+  -- use only the round-robin method. Otherwise, uncomment the other methods.
+
+  -- round robin if less than gbvAllocDelay
+  --if self.map.level_config.gbv.AllocDelay and self.game_date:monthOfGame() <= self.map.level_config.gbv.AllocDelay then
+  local total_spawn_attempts = 0
+  for _, hospital in ipairs(self.hospitals) do
+    total_spawn_attempts = total_spawn_attempts + hospital.spawn_attempts
+  end
+  return self.hospitals[total_spawn_attempts % 4 + 1]
```
tobylane: This can return 4 in a 3 hospital world.

ARGAMX: Good found 👍
Fixed

PR #1987, opened 2021-07-01
Use the watch to show tutorial progress and ability to skip

### LINKED ISSUE #1828: Make tutorial more or less restrictive
Either make the tutorial more restrictive (force click only buttons of tutorial) or less restrictive (allow tutorial exit) so a hospital can't get stuck with no patients. (noted from IRC chat report)

Will either force a defined path to complete the tutorial -- or allow escaping to continue playing.

Not sure.

 0.64

No

**Additional information**

tobylane:
*I believe this would fix #1828*

- Add tutorial as a watch type, which is pushed forward with tutorial progress
- Prevent epidemics happening during other watch events (to match emergencies)

todo: The tutorial watch looks identical to the following initial opening watch, it could be blue?. The watch progress doesn't yet take into account any tutorial steps the user skips, such as looking through available staff or invalid room/staff placement. The watch doesn't fill in segment by segment, how is that done?

### DIFF
--- a/CorsixTH/Lua/dialogs/adviser.lua
+++ b/CorsixTH/Lua/dialogs/adviser.lua
@@ -123,6 +123,7 @@ function UIAdviser:hide()
   self.th:setAnimation(self.ui.app.world.anims, 440)
   self.frame = 1
   self.number_frames = 4
+  self.ui:tutorialStep(1, 1, 2)
 end
 
 -- Makes the adviser say something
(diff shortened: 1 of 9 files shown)


lewri:
I wonder if we can superimpose tutorial on the watch at all... Or add an adviser message to point it out.

tobylane:
Added a tutorial start message (button is meant to be bubble).

If any steps were missed in a chapter the progress is pushed up to date. This does draw another watch near the end of the tutorial, I have no idea why. It's the step when you place a doctor, 4, 4?

lewri:
1,3 we start using a clock -- which I like -- but not sure why we go from all white background to all red.
image

PR #1988, opened 2021-07-02
Machine repair changes

tobylane:
- Move the announcing function to PlayerHospital
- Add the handymen check per the comment

### DIFF
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -20,10 +20,6 @@ SOFTWARE. --]]
 
 local TH = require("TH")
 
-corsixth.require("announcer")
-
-local AnnouncementPriority = _G["AnnouncementPriority"]
-
 --! An `Object` which needs occasional repair (to prevent explosion).
 class "Machine" (Object)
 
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -391,6 +391,19 @@ function PlayerHospital:adviseBoilerBreakdown(broken_heat)
   end
 end
 
+local AnnouncementPriority = _G["AnnouncementPriority"]
```
lewri, line 394: I would recommend always puttting this line at the top of the file before the class declaration. Also missing the ``require``

CorsixTH/Lua/entities/machine.lua
```diff
@@ -22,8 +22,6 @@ local TH = require("TH")
 
 corsixth.require("announcer")
```
lewri, line 23: As per other comment.
You don't need this line anymore if you're taking any announcements out of machine.

lewri (changes requested):
We can ignore the lua failure

lewri:
I still think for the sake of uniformity you should re-add those require and announcementpriority lines into Player Hospital for now.
We can make a separate PR to test removing them across the board if the hunch of them not being required is right.

tobylane:
Agreed. Ready.

lewri (approved):

PR #1997, opened 2021-07-09
Add cheats to ingame menu

tobylane:
**New cheats**
- Toggle earthquakes in the world
- Toggle epidemics in the hospital
- Set death counts to zero
- Set reputation to the maximum

Screen 2021-06-29 at 17 38 40

### DIFF
--- a/CorsixTH/Lua/cheats.lua
+++ b/CorsixTH/Lua/cheats.lua
@@ -37,8 +37,8 @@ function Cheats:Cheats(hospital)
     {name = "money",          func = self.cheatMoney},
     {name = "all_research",   func = self.cheatResearch},
     {name = "emergency",      func = self.cheatEmergency},
+    {name = "show_infected",  func = self.cheatShowInfected},
     {name = "epidemic",       func = self.cheatEpidemic},
-    {name = "toggle_infected", func = self.cheatToggleInfected},
     {name = "vip",            func = self.cheatVip},
     {name = "earthquake",     func = self.cheatEarthquake},
     {name = "create_patient", func = self.cheatPatient},
@@ -102,16 +102,11 @@ function Cheats:cheatEmergency()
   end
 end
 
---[[ Creates a new contagious patient in the hospital - potentially an epidemic]]
-function Cheats:cheatEpidemic()
-  self.hospital:spawnContagiousPatient()
-end
-
---[[ Before an epidemic has been revealed toggle the infected icons
-to easily distinguish the infected patients -- will toggle icons
+--[[ Before an epidemic has been revealed, show/hide the infected mood icons
+to easily distinguish the infected patients. This will show/hide icons
 for ALL future epidemics you cannot distinguish between epidemics
 by disease ]]--
-function Cheats:cheatToggleInfected()
+function Cheats:cheatShowInfected()
   local hosp = self.hospital
   if hosp.future_epidemics_pool and #hosp.future_epidemics_pool > 0 then
     for _, future_epidemic in ipairs(hosp.future_epidemics_pool) do
@@ -123,8 +118,13 @@ function Cheats:cheatToggleInfected()
       end
     end
   else
-    self.hospital.world:gameLog("Unable to toggle icons - no epidemics in progress that are not revealed")
+    return false, _S.misc.epidemic_no_icon_to_toggle
   end
+end
+
+--[[ Creates a new contagious patient in the hospital - potentially an epidemic]]
+function Cheats:cheatEpidemic()
+  self.hospital:spawnContagiousPatient()
 end
 
 function Cheats:cheatVip()
--- a/CorsixTH/Lua/cheats.lua
+++ b/CorsixTH/Lua/cheats.lua
@@ -39,8 +39,10 @@ function Cheats:Cheats(hospital)
     {name = "emergency",      func = self.cheatEmergency},
     {name = "show_infected",  func = self.cheatShowInfected},
     {name = "epidemic",       func = self.cheatEpidemic},
-    {name = "vip",            func = self.cheatVip},
+    {name = "toggle_epidemic", func = self.cheatToggleEpidemic},
     {name = "earthquake",     func = self.cheatEarthquake},
+    {name = "toggle_earthquake", func = self.cheatToggleEarthquake},
+    {name = "vip",            func = self.cheatVip},
     {name = "create_patient", func = self.cheatPatient},
     {name = "end_month",      func = self.cheatMonth},
     {name = "end_year",       func = self.cheatYear},
@@ -59,9 +61,11 @@ end
 --! Performs a cheat from the cheat_list (menu cheats)
 --!param num (integer) The cheat from the cheat_list called
 --!return true if cheat was successful, false otherwise
+--!return message (optional string) The text message of the cheat.
 function Cheats:performCheat(num)
-  local cheat_success = self.cheat_list[num].func(self) ~= false
-  return cheat_success and self.cheat_list[num].name ~= "lose_level"
+  local cheat_success, message = self.cheat_list[num].func(self)
+  cheat_success = cheat_success ~= false
+  return cheat_success and self.cheat_list[num].name ~= "lose_level", message
 end
 
 --! Updates the cheated status of the player, with a matching announcement
@@ -93,11 +97,10 @@ end
 
 function Cheats:cheatEmergency()
   local err = self.hospital:createEmergency()
-  local ui = self.hospital.world.ui
   if err == "undiscovered_disease" then
-    ui:addWindow(UIInformation(ui, {_S.misc.cant_treat_emergency}))
+    return false, _S.misc.cant_treat_emergency
   elseif err == "no_heliport" then
-    ui:addWindow(UIInformation(ui, {_S.misc.no_heliport}))
+    return false, _S.misc.no_heliport
   -- else 'err == nil', meaning success. The case doesn't need special handling
   end
 end
@@ -124,15 +127,43 @@ end
 
 --[[ Creates a new contagious patient in the hospital - potentially an epidemic]]
 function Cheats:cheatEpidemic()
-  self.hospital:spawnContagiousPatient()
+  if not self.hospital.epidemics_disabled then
+    return self.hospital:spawnContagiousPatient()
+  end
+end
+
+--! Toggles the possibility of epidemics
+function Cheats:cheatToggleEpidemic()
+  local hosp, msg = self.hospital
+  if hosp.epidemics_disabled then
+    msg = _S.misc.epidemics_on
+  else
+    msg = _S.misc.epidemics_off
+  end
+  hosp.epidemics_disabled = not hosp.epidemics_disabled
+  return true, msg
+end
+
+function Cheats:cheatEarthquake()
+  if not self.hospital.earthquakes_disabled then
+    self.hospital.world:createEarthquake()
+  end
+end
+
+--! Toggles the possibility of earthquakes
+function Cheats:cheatToggleEarthquake()
+  local world, msg = self.hospital.world
+  if world.earthquakes_disabled then
+    msg = _S.misc.earthquakes_on
+  else
+    msg = _S.misc.earthquakes_off
+  end
+  world.earthquakes_disabled = not world.earthquakes_disabled
+  return true, msg
 end
 
 function Cheats:cheatVip()
   self.hospital:createVip()
-end
-
-function Cheats:cheatEarthquake()
-  return self.hospital.world:createEarthquake()
 end
 
 function Cheats:cheatPatient()
--- a/CorsixTH/Lua/cheats.lua
+++ b/CorsixTH/Lua/cheats.lua
@@ -33,7 +33,6 @@ local Cheats = _G["Cheats"]
 function Cheats:Cheats(hospital)
   self.hospital = hospital
   -- Cheats to appear specifically in the cheats window
-  -- New cheats require an afterLoad when added
   self.cheat_list = {
     {name = "money",          func = self.cheatMoney},
     {name = "all_research",   func = self.cheatResearch},
--- a/CorsixTH/Lua/cheats.lua
+++ b/CorsixTH/Lua/cheats.lua
@@ -49,6 +49,8 @@ function Cheats:Cheats(hospital)
     {name = "win_level",      func = self.cheatWin},
     {name = "increase_prices", func = self.cheatIncreasePrices},
     {name = "decrease_prices", func = self.cheatDecreasePrices},
+    {name = "reset_death_count", func = self.cheatResetDeathCount},
+    {name = "max_reputation",  func = self.cheatMaxReputation},
   }
 
   self.active_cheats = {} -- Toggle cheat status
@@ -275,4 +277,13 @@ function Cheats:toggleCheat(name)
   if cheatWindow then
     cheatWindow:updateCheatedStatus()
   end
+end
+
+function Cheats:cheatResetDeathCount()
+  self.hospital:resetDeathCount()
+end
+
+function Cheats:cheatMaxReputation()
+  local hosp = self.hospital
+  hosp:unconditionalChangeReputation(hosp.reputation_max)
 end
--- a/CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
+++ b/CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
@@ -135,12 +135,7 @@ function UICheats:buttonBack()
 end
 
 function UICheats:afterLoad(old, new)
-  if old < 145 then
-    -- Window must be closed if open for compatibility
-    local cheatWindow = self.ui:getWindow(UICheats)
-    if cheatWindow then
-      cheatWindow:close()
-    end
-  end
+  -- Window must be closed for compatibility
+  self:close()
   UIResizable.afterLoad(self, old, new)
 end
--- a/CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
+++ b/CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
@@ -122,11 +122,13 @@ function UICheats:buttonClicked(num)
     --TODO: Prevent selectx.wav playing with this
     return self.ui:playSound("wrong2.wav")
   end
-  if self.cheats:performCheat(num) then
+  local success, message = self.cheats:performCheat(num)
+  if success then
     self.cheats.announceCheat(self.ui)
     self:updateCheatedStatus()
-  else
-    self.ui:addWindow(UIInformation(self.ui, {_S.information.cheat_not_possible}))
+  end
+  if message then
+    self.ui:addWindow(UIInformation(self.ui, {message or _S.information.cheat_not_possible}))
   end
 end
 
--- a/CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
+++ b/CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
@@ -63,7 +63,7 @@ function UICheats:UICheats(ui)
   self.cheats = ui.hospital.hosp_cheats
   self.cheat_list = ui.hospital.hosp_cheats.cheat_list
 
-  self:UIResizable(ui, 300, 200, col_bg)
+  self:UIResizable(ui, 300, 560, col_bg)
 
   self.default_button_sound = "selectx.wav"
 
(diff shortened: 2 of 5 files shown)


lewri (commented):
I think on/off may be better than toggle.
Having "toggle earthquake" and "create earthquake" together feels a bit confusing.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -496,6 +493,9 @@ function PlayerHospital:afterLoad(old, new)
   if old < 149 then
     self.win_declined = false -- Has not yet declined the level win fax
   end
+  if old < 159 then
+    self.hosp_cheats = Cheats(self)
```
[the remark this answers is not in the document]
lewri: @Alberth289346 pointed out in the past my implementation led to this drawback. So a forced reload each time could be okay.

Please also update ``function UICheats:afterLoad(old, new)`` incase the window is open.

tobylane:
Agreed, though "turn earthquakes on or off" may be too long. / isn't an available character.

Changed and rebased.

tobylane:
This puts earthquakes and epidemics first, in a toggle and do now section. However, the buttons don't connect correctly and I'm stuck. What's needed?

CorsixTH/Lua/cheats.lua
```diff
@@ -35,15 +35,13 @@ function Cheats:Cheats(hospital)
...
   -- New cheats require an afterLoad when added
   self.cheat_list = {
+    {name = "epidemic",       toggle = self.cheatToggleEpidemic,  func = self.cheatEpidemic, state = self.hospital.epidemics_disabled},
+    {name = "earthquake",     toggle = self.cheatToggleEarthquake, func = self.cheatEarthquake, state = self.hospital.earthquakes_disabled},
     {name = "money",          func = self.cheatMoney},
     {name = "all_research",   func = self.cheatResearch},
     {name = "emergency",      func = self.cheatEmergency},
-    {name = "epidemic",       func = self.cheatEpidemic},
-    {name = "toggle_epidemic", func = self.cheatToggleEpidemic},
     {name = "toggle_infected", func = self.cheatToggleInfected},
```
TheCycoONE, line 43: Now that toggle is being used in this context as an on/off switch, toggle_infected is - not a good name.

CorsixTH/Lua/cheats.lua
```diff
@@ -58,9 +56,12 @@ end
 
 --! Performs a cheat from the cheat_list
 --!param num (integer) The cheat from the cheat_list called
+--!param toggle (boolean) Whether the toggle function of the cheat should be used
 --!return true if cheat was successful, false otherwise
-function Cheats:performCheat(num)
-  local cheat_success = self.cheat_list[num].func(self) ~= false
+function Cheats:performCheat(num, toggle)
+print(num, toggle)
```
Alberth289346, line 62: Delete the debug code, or make it a more readable log line.

CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
```diff
@@ -84,14 +84,41 @@ function UICheats:UICheats(ui)
...
     return --[[persistable:cheats_button]] function(window)
       window:buttonClicked(num)
+      print("Clicked", num)
+    end
+  end
+
+  local function toggle_clicked(num, toggle)
+    return --[[persistable:cheats_toggle]] function(window)
+      window:buttonClicked(num, toggle)
+      print("Toggled", num)
```
Alberth289346, line 94: Either delete debug code or make it a nice readable log line

lewri:
One thing that occurred to me is the default resolution of 800x600 in the game. Does the cheats menu still fit fine at this resolution?

lewri:
Update: Still haven't actually checked but by going from the Screenshot the cheats menu is too tall for the minimum resolution 800x600

tobylane:
Fixed the issues mentioned but not the buttons yet. 

It's usable, but doesn't fit, in the smallest option of 480 high.
Screen 2021-09-27 at 14 45 04

TheCycoONE:
Looks like it would fit better with two columns of buttons. Not sure about all locales but there seems to be a lot of horizontal space.

CorsixTH/Lua/world.lua
```diff
@@ -1417,17 +1417,18 @@ end
...
-    self.next_earthquake.start_month = self.game_date:monthOfGame()
-    if self.next_earthquake.size == nil then
-      --forcefully make an earthquake if none left in level file
-      self.next_earthquake.size = math.random(1,6) -- above 6 seems disastrous
-      self.next_earthquake.remaining_damage = self.next_earthquake.size
-      self.next_earthquake.damage_timer = earthquake_damage_time
-      self.next_earthquake.warning_timer = earthquake_warning_period
-    end
+  -- Make sure an earthquake isn't already happening, or if they're turned off in cheats
+  if self.next_earthquake.active or self.earthquakes_disabled then
```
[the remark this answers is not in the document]
lewri: Out of scope, but for further extension in a future PR you could make the "create earthquake" (and epidemic) buttons grey out if we disable them.

tobylane:
I've dropped the toggle cheats for now.

Added a second column of buttons. I would like to have the longer labels over two lines, how can I have the // do that here, if that is the best way? One minor to do: odd numbered cheat with a long label needs to push the next row down.

CorsixTH/Lua/hospital.lua
```diff
@@ -2440,3 +2442,10 @@ end
 function Hospital:announceStaffLeave()
   -- Nothing to do, override in a derived class.
 end
+
+--! Reset the death counts (key losing conditions)
+function Hospital:resetDeathCount()
+  self.num_deaths = 0
+  self.num_deaths_this_year = 0
+  self.statistics.deaths = 0
+end
```
lewri, line 2451: Based on previous cheat code -- we usually only elect to use a function from another class if it is used routinely.
In this case, it's a method specific to the cheat itself.

lewri:
> Added a second column of buttons. I would like to have the longer labels over two lines, how can I have the // do that here, if that is the best way?

`Panel:addLabel` should technically wrap the line by itself. However, it looks like nothing in `cheats_dialog.lua` actually goes off to adjust the panel size if you end up with two rows of text.

You'd either have to find a way for the code to report back the total number of lines used (and adjust the button height appropriately); or maybe create an array of text lines in `_S` for the name and get the code to measure the array length to calculate height.

The best solution really is having the panel react to the text, rather than us having 'fixed' heights on the dialog. Looks like `sizeOf` in `graphics.lua` might be a good pointer.

lewri:
> Looks like it would fit better with two columns of buttons. Not sure about all locales but there seems to be a lot of horizontal space.

Out of scope, but you could also implement a scrollbar?

tobylane:
Rewrote the two columns part. The buttons keep the same width, so the Portuguese text fits without a linebreak.

Screen 2023-10-02 at 13 01 57

CorsixTH/Lua/languages/english.lua
```diff
@@ -264,7 +264,11 @@ adviser = {
     norest_off_cheat = "Phew! Looks like that buzz finally wore off. Your staff will now rest properly.",
   },
   staff_place_advice = {
-    not_enough_lecture_chairs = "Each student doctor needs a lecture chair to sit in!"
+    not_enough_lecture_chairs = "Each student doctor needs a lecture chair to sit in!",
+    epidemics_off = "Epidemics are disabled",
+    epidemics_on = "Epidemics are re-enabled",
+    earthquakes_off = "Earthquakes are disabled",
+    earthquakes_on = "Earthquakes are re-enabled",
```
lewri, line 271: There probably weren't meant for here

CorsixTH/Lua/cheats.lua
```diff
@@ -101,16 +105,28 @@ function Cheats:cheatEmergency()
...
 end
 
+--! Toggles the possibility of epidemics
+function Cheats:cheatToggleEpidemic()
+  local hosp = self.hospital
+  local ui = hosp.world.ui
+  if hosp.epidemics_disabled then
+    ui:addWindow(UIInformation(ui, {_S.adviser.cheats.epidemics_on}))
+  else
+    ui:addWindow(UIInformation(ui, {_S.adviser.cheats.epidemics_off}))
```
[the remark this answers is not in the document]
Alberth289346: Maybe also route the UI interactions through the hospital instance. I know you instantiate the cheats in the player hospital, but in this form `self.hospital` is not as generic as it looks.

If you reroute the calls through an hospital interface then cheat s will work for any hospital.

tobylane:
Added PlayerHospital:giveInfo for those dialogs, rather than persisted advisor strings.

Added Window.resize_redraw to trigger reopening some windows when the screen resolution changes.

CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
```diff
@@ -62,8 +62,13 @@ local col_cheated_yes = {
 function UICheats:UICheats(ui)
   self.cheats = ui.hospital.hosp_cheats
   self.cheat_list = ui.hospital.hosp_cheats.cheat_list
+  self.resize_redraw = UICheats
```
lewri, line 65: An interesting implementation

CorsixTH/Lua/window.lua
```diff
@@ -113,6 +113,11 @@ function Window:onChangeResolution()
   if self.x_original and self.y_original then
     self:setPosition(self.x_original, self.y_original)
   end
+  if self.resize_redraw then
+    -- Reopen dialog in the new resolution in case the need for columns changes
+    self:close()
+    self.ui:addWindow(self.resize_redraw(self.ui))
+  end
 end
```
lewri, line 121: An interesting implementation/2

lewri (approved):
There's a lot here. I gave it another skim and it all seems fine, but someone else may wish to check it over again.

CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
```diff
@@ -91,20 +97,37 @@ function UICheats:UICheats(ui)
...
+          :setLabel(_S.cheats_window.cheats[self.cheat_list[num].name])
+        self.item_buttons[num] = self.item_panels[num]:makeButton(0, 0, 260, 20, nil, button_clicked(num))
+          :setTooltip(_S.tooltip.cheats_window.cheats[self.cheat_list[num].name])
+      end
+      y = y + 20
+    else
+      self.item_panels[num] = self:addBevelPanel(20, y, 260, 20, col_bg)
+        :setLabel(_S.cheats_window.cheats[self.cheat_list[num].name])
+      self.item_buttons[num] = self.item_panels[num]:makeButton(0, 0, 260, 20, nil, button_clicked(num))
+        :setTooltip(_S.tooltip.cheats_window.cheats[self.cheat_list[num].name])
```
Alberth289346, line 119: You add this code 3 times, stick it in a local function and call that 3 times.

In that way, it's much simpler to understand what is unique in these calls and what is always the same. Also, if some value needs to be changed, you can do that with 1 change rather than 3 changes with all the risks that it becomes inconsistent.

CorsixTH/Lua/hospital.lua
```diff
@@ -1181,9 +1181,11 @@ function Hospital:spawnContagiousPatient()
       self:addToEpidemic(patient)
     else
       print("Cannot create epidemic - no contagious diseases available")
+      return false
     end
   else
     print("Cannot create epidemic - no staffed reception desk")
+    return false
```
Alberth289346, line 1188: These returns are in `Hospital:spawnContagiousPatient()` if I read the code correctly?

The function name doesn't read asif you should expect a value to be returned, also it has no `@return` documentation, so you shouldn't return a value here as far as I can see.  
If you want the function to end here (which is perhaps what you're aiming for), just `return` will do.

CorsixTH/Lua/hospital.lua
```diff
@@ -1232,7 +1234,7 @@ end
  appropriate epidemic if so.
  @param patient (Patient) patient to determine if contagious]]
 function Hospital:determineIfContagious(patient)
-  if patient.is_emergency or not patient.disease.contagious then
+  if patient.is_emergency or not patient.disease.contagious or self.epidemics_off then
```
Alberth289346, line 1237: I'd write your addition as first condition rather than as last condition, perhaps even as separate `if` (but that may be just me).

The new condition is a big "everything epidemics doesn't apply" switch. That is at a higher abstraction level than digging around in specific patient types and diseases to decide whether the patient may be contagious.

When reading the code the mind has to make abstraction level switches from details to high level and back to some other details. By doing high level conditions first, you ease the reader from higher level concerns to lower level concerns.   If you do this consistently, it also becomes a pattern. Higher level concerns are near the top of a function. That makes finding and checking easier, eventually.

The name of the variable is quite horrible (it has negative logic which gets very complicated very quickly in code), but that's not your fault and out of scope in this patch.

CorsixTH/Lua/hospital.lua
```diff
@@ -2365,6 +2367,13 @@ function Hospital:getRandomBusyRoom()
   if #chosen_room.door.queue >= busy_threshold then return chosen_room end
 end
 
+--! Reset the death counts (key losing conditions)
+function Hospital:resetDeathCount()
+  self.num_deaths = 0
+  self.num_deaths_this_year = 0
+  self.statistics.deaths = 0
+end
```
Alberth289346, line 2375: Good that you moved that to here.

Function name is perhaps a bit too neutral, hospitals normally don't reset death counts, so it's not a regular function you'd use in normal hospital computations.
Not sure if you want to fix that, adding some comment may also do the trick?

CorsixTH/Lua/hospital.lua
```diff
@@ -2456,3 +2465,7 @@ end
 --! Called when the vip is out of the hospital grounds
 function Hospital:makeVipEndFax(vip_rating, name, cash_reward, vip_message)
 end
+
+--! Give the player information with the UIInformation dialog
+function Hospital:giveInfo(message)
+end
```
Alberth289346, line 2471: `giveInfo` is highly generic, while the description is very specific.

For example, the function name doesn't exclude using speech to "give information".
The description is more about "how" rather  than "what does the function aim to do".

As for changing this, my guess is to avoid "how" (basically, the way to get such detailed information should be to read the actual code or comments in that code), and just tell the aim. Not sure if you want to exclude eg speech or other ways to give information to the player. In general there is nothing wrong with keeping possibilities that we currently don't have open. Who knows what the future will bring.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -761,6 +761,11 @@ function PlayerHospital:makeVipEndFax(vip_rating, name, cash_reward, vip_message
   self.world.ui.bottom_panel:queueMessage("report", message, nil, Date.hoursPerDay() * 20, 1)
 end
 
+--! Give the player information with the UIInformation dialog
+function PlayerHospital:giveInfo(message)
+  self.world.ui:addWindow(UIInformation(self.world.ui, { message }))
```
Alberth289346, line 766: I don't know what other Hospital code is doing, but `UIInformation` is a UI thing rather than a hospital thing, so this leaks UI details to hospital code.

Likely it happens all over the place, and is it not something you can easily fix. As such, it's probably better to keep this for now,

`world.ui` has similar problems, but that is even more wide-spread. No danger to run out of code to refactor thus :)

CorsixTH/Lua/world.lua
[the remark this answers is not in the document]
Alberth289346: Also here a detailed check before the big master-switch check.

Not sure if you should handle that case in this way. The function name says "please create me an earth quake". That means the caller of this function must be thinking that earthquakes are a relevant thing in this game.  
That directly contradicts with the `self.earthquakes_disabled` condition that with its negative logic says earthquakes don't happen at all.  
So it is highly questionable if anything should ever call this function if earthquakes are turned off. Basically you don't want to spend CPU cycles on code that does nothing in the game. The caller thus shouldn't be in code that even considers creating an earthquake.

Adding the `self.earthquakes_disabled` check prevents earthquakes from happening. However at the same time it also hides code that is wasting CPU cycles trying to create earthquakes when it shouldn't do that. You will never ever find out that such code is running at the wrong time.

Some of the better tactics here are
- assert that earth quakes are allowed here.
- find out what code calls this function, and fix that code to never create a quake.

Basically high level master switches should only be checked at the entry points, functions called from external code (that is, code not involved with earth quakes), so all code involved with the detail is simply never executed and no CPU cycles are wasted.

CorsixTH/Lua/hospitals/player_hospital.lua
lewri: Anything and everything seems to call ui in terms of needing to generate a message of some kind. Did this come from a misunderstanding of what I said about the adviser message previously? I was fine to give the cheats dialog ability to create messages, they just shouldn't use `_A` for something that always uses `_S`

CorsixTH/Lua/window.lua
```diff
@@ -113,6 +113,11 @@ function Window:onChangeResolution()
   if self.x_original and self.y_original then
     self:setPosition(self.x_original, self.y_original)
   end
+  if self.resize_redraw then
+    -- Reopen dialog in the new resolution in case the need for columns changes
+    self:close()
+    self.ui:addWindow(self.resize_redraw(self.ui))
```
Alberth289346, line 119: Big sledge hammer approach, move a road a bit by leveling the entire block of houses, and rebuilding them all.

It works, in a very crude way. For example, if you have multiple windows on the screen, this will break the layering order (what window is above which other windows) for the user. May not be a problem currently in the program.

Ignoring the big hammer, there are some parts not quite right here.

The first thing is `self.resize_draw`. You define it in the derived class, and use it in the base class. That is against object-oriented thinking, where the base class should never know anything about the derived class. (The other way around is allowed, although should be limited and/or run through functions to make code changes more easily). Here, the `Window` class should thus not know about the `resize_redraw` variable since the variable "lives" in a derived class.

If you want this approach (although I will argue below you probably don't), then `Window` should own `resize_redraw`. Derived classes may then access or change it. A minor secondary point is that there are better ways to retrieve the class (did you ever read the `class.lua` file comments?) from `self`. I'll leave that for another time, ask me if you really want to know.

The second thing is that the base class is in no position to truthfully re-create a new instance with all relevant derived class information preserved. It should NOT know about content of derived classes, so if the recreated window needs some of its information to be preserved in the next instance (eg scrollbar positions, buttons that are pushed or disabled, currently displayed data, etc), this approach will completely fail.

The only way to handle that is not do this here, and do it in the derived class instead. In other words, ask or tell the derived class to do this by means of a function call to it. The derived class knows what needs to be copied to the new instance. This also avoids breaking the construction process, in case a derived class has additional parameters in its constructor (beyond `self.ui`) that you now completely ignore.

Usually you'd create a new function for that. In this particular case, I am not sure that is needed. Is it an option to override `onChangeResolution` in the derived class instead?  Likely you'd want to call `Window.onChangeResolution(self)` there at some point to get the base class code executed.

CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua
```diff
@@ -62,8 +62,13 @@ local col_cheated_yes = {
 function UICheats:UICheats(ui)
   self.cheats = ui.hospital.hosp_cheats
   self.cheat_list = ui.hospital.hosp_cheats.cheat_list
+  self.resize_redraw = UICheats
 
-  self:UIResizable(ui, 300, 200, col_bg)
+  local width, header_indent, header_width, second_col = 300,  20, 260, false
+  if 180 + 20 * #self.cheat_list > ui.app.config.height then -- Cheats list will not fit in one column
+    width, header_indent, header_width, second_col     = 560, 100, 360, 280
```
Alberth289346, line 69: Shouldn't it set `second_col` ?

Alberth289346: Ah, my wrong, it does. It doesn't use `true`  though, that's somewhat confusing.

tobylane (commented):
I have a few fixes done, but the major todo is the window resizing.

Alberth289346:
> I have a few fixes done, but the major todo is the window resizing.

So what is the problem with that?
That is, what needs done? What would be the goal?

tobylane:
Screenshot 2024-01-10 at 18 08 49

Two columns for everyone, and maybe add to the todo list to try something better, like a scrollbar.

TheCycoONE:
Seems like there are conflicts before this can be merged again.

PR #2000, opened 2021-07-10
Document effect and effect_ticks parameters for patient effects.

flackr:
- Adds missing documentation of animation effect parameters.


lewri:
Looks good here... Then again is just documentation :)

PR #2009, opened 2021-07-12
[RDY] Backport Trophy Fix (0.65)

lewri:
*Fixes release branch of #2004*

**Describe what the proposed change does**
- Apply the correct config params for 0.65 release for the All Cures Trophy.

PR #2010, opened 2021-07-12
[RDY] Apply all cures trophy variables (permfix)

### LINKED ISSUE #2004: [Bug] AllCuresTrophy crashes existing saves
#### The issue
While playing the first hospital, having reached all my goals (10 cured patients and a rating above 200), time freezes on the 31st of December, right before I get the letter to move on to the second hospital. I can still build and place things, but the people are frozen, time doesn't progress and I don't get my letter to go to the next level. Moreover, if I quit out onto the main menu, my cursor freezes in place. And the only way to visibly move it is to click somewhere else.

#### Steps to Reproduce
1. Load the save file attached.

2. See that it's frozen in time.

3. Quit out onto the main menu and observe that the cursor is stuck.

#### Save Game
[CorsixTH save.zip](https://github.com/CorsixTH/CorsixTH/files/6796561/CorsixTH.save.zip)

#### Expected Behavior
I expected the letter to move on to the next hospital, or at least that the game would continue playing.

#### System Information
**CorsixTH Version:** 0.65

**Theme Hospital Version:** GOG.com

#### Gamelog.txt
Gamelog.txt comes within the .zip file included.

### Additional Info
I don't know if this is related, but my (local) account name on my laptop has a Romanian ș in it, and the Users folder has the Turkish ş in it. This has caused problems for me in the past (I still can't play OpenLoco on this account, and I've already run into countless issues with CorsixTH). It might also be my laptop that's bungling up something. I never had the same problems with CorsixTH on my older laptop (or any program not working because of my username), despite it having the same account name. Thought you might wanna know.

lewri:
*Fixes #2004*

- Applies the afterLoad needed for the Allcures trophy
- Deletes the hotfix variable caused by #2009


lewri:
@TheCycoONE should master also be getting updated to 0.65.1 in ``getVersion``?
Or should it have a separate entry independent of 0.65?

lewri:
Ready. I accidentally made the PR's commits a bit too dirty so will need squishing :)

CorsixTH/Lua/app.lua
```diff
@@ -1470,7 +1470,7 @@ function App:getVersion(version)
   if ver > 156 then
     return "Trunk"
   elseif ver > 138 then
-    return "v0.65"
+    return "v0.65.1"
```
TheCycoONE, line 1473: might be best to leave it - we don't know whether they had the hotfix or not.

CorsixTH/Lua/app.lua
```diff
@@ -28,7 +28,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
 
-local SAVEGAME_VERSION = 159
+local SAVEGAME_VERSION = 160
```
TheCycoONE, line 31: needs to bump again since hotkey was merged

lewri:
```suggestion
local SAVEGAME_VERSION = 161
```

CorsixTH/Lua/app.lua
lewri:
```suggestion
    return "v0.65"
```

CorsixTH/Lua/map.lua
```diff
@@ -802,4 +802,11 @@ function Map:afterLoad(old, new)
       self:setCellFlags(58, 72, {passable = false})
     end
   end
+  if old < 160 then
```
lewri, line 805:
```suggestion
  if old < 161 then
```

Issue #2014, opened 2021-07-14
Handymen trying to sweep outside litter


DS1992ds:
hey - an annoying glitch to look into - how people queue, on the most up to date version its literally one person in and one person out with a pause in the middle - no matter the room. I currently have 32 people queuing for the ward despite there being available beds haha. its the same for the staff room - there's also a glitch on the map I have made where handymen get stuck but I'm not sure if this can be fixed, plus I don't fancy rebuilding my hospital haha

tobylane:
Please can you upload the files as a zip - the save, map and level.

DS1992ds:
[MAssive.zip](https://github.com/CorsixTH/CorsixTH/files/6817534/MAssive.zip)
hope this helps

lewri:
```lua
Error in timer handler: 
Lua\humanoid_actions\sweep_floor.lua:63: attempt to index a nil value (local 'hospital')
stack traceback:
	Lua\humanoid_actions\sweep_floor.lua:63: in field '?'
	Lua\entities\humanoid.lua:527: in upvalue 'Humanoid_startAction'
	Lua\entities\humanoid.lua:627: in method 'finishAction'
	Lua\humanoid_actions\walk.lua:187: in local 'timer_function'
	Lua\entity.lua:216: in field 'tick'
	Lua\entities\humanoids\staff.lua:122: in method 'tick'
	Lua\world.lua:1081: in method 'onTick'
	Lua\app.lua:1134: in function <Lua\app.lua:1131>
	(...tail calls...)
	Lua\app.lua:1034: in function <Lua\app.lua:1029>
Recovering from error in timer handler...
Error in timer handler: 
Lua\humanoid_actions\sweep_floor.lua:63: attempt to index a nil value (local 'hospital')
stack traceback:
	Lua\humanoid_actions\sweep_floor.lua:63: in field '?'
	Lua\entities\humanoid.lua:527: in upvalue 'Humanoid_startAction'
	Lua\entities\humanoid.lua:627: in method 'finishAction'
	Lua\humanoid_actions\walk.lua:187: in local 'timer_function'
	Lua\entity.lua:216: in field 'tick'
	Lua\entities\humanoids\staff.lua:122: in method 'tick'
	Lua\world.lua:1081: in method 'onTick'
	Lua\app.lua:1134: in function <Lua\app.lua:1131>
	(...tail calls...)
	Lua\app.lua:1034: in function <Lua\app.lua:1029>
```

[...]

PR #2015, opened 2021-07-14
[RDY] Ask litter for hospital

lewri:
Fixes the bug in #2014 but not preventing it from happening (needs a separate PR/thinking)

- Fixes a missed change in #1768 since litter now knows about hospital.


lewri:
On reading other issues this fix would at least stop any crashing while we improve where people can litter. Assuming that makes sense @mugmuggy ?

CorsixTH/Lua/objects/litter.lua
```diff
@@ -159,6 +159,15 @@ function Litter:afterLoad(old, new)
   if old < 151 then
     self.hospital = self.world:getHospital(self.tile_x, self.tile_y)
   end
+  if old < 159 then
```
lewri, line 162:
```suggestion
  if old < 162 then
```

Issue #1915, opened 2021-05-31
Hint to people that there are other languages available, after a font selection


kparal:
**What feature or enhancement would you like to add?**
Currently languages not supported by the default font are way too hidden for most people (who don't read corsixth wiki etc). There's nothing in the GUI telling you there are some languages hidden:
![font selection](https://user-images.githubusercontent.com/755451/120196183-dba91c80-c21f-11eb-98ff-b97a2582f762.png)

I believe this could be easily improved. Ideas:
* Add a note into the tooltips (one above `Game Language`, the other shown in the screenshot). Something like:
  > Note: There are some languages hidden because the game font can't display them properly. Please select an appropriate font in the Folder menu and they will show up in this menu. Currently hidden languages are: Chinese, Czech, Russian, ...

  Create this tooltip dynamically to include the list of hidden languages, and display it only if there is no alternative font selected. If an alternate font is selected, replace it with this content:
  > Note: Some languages are displayed using an alternative font, because they are not supported by the default game font. If some characters are not displayed correctly, select a different font in the Folder menu.

* Add some small info/warning icon right next to the `Game Language` text, to make sure user moves their cursor over the icon and displays the tooltip.

* Alternatively, when the language drop-down is expanded, show a window with this hint next to the `Settings` window.

* Alternatively, add a new item into the language drop-down to the very end, called e.g. `More languages`. When the user clicks on it, ignore the selection (keep the previous language), but display a window/tooltip with the hint.

**What version of CorsixTH are you using?**
0.65-beta2, Flathub

tobylane:
Could we add information like this to the tips window to be more noticeable? In English only, add to the set of tips one of the two quotes above, depending on whether a font has been chosen. When a user sets a font in game, the current tip is set to the second message.

Alberth289346:
Why not have the full set of languages always, with the unavailable languages disabled? That's what is commonly used to list unavailable options. Could add a tool tip that the user needs to find an appropriate font to enable it.

Issue #1914, opened 2021-05-31
Font option is slightly confusing


kparal:
**What feature or enhancement would you like to add?**
In _Settings -> Folders -> Font_, it seems the game lets you select a different font. But it doesn't. If you select some font, "nothing happens". **Until** you select a different language, and not just any language, but a language that is not supported by the default font. This confused me when I first used this option. Concretely I thought:
1) Why has nothing happened? Do I need to restart the game? (Hint: Restarting the game doesn't change anything).
2) How can I go back "to the defaults" and take back the selected font? Even if "nothing happened", I don't want to take anything risks, how do I remove the selected font path? (Hint: You can't).

It would be more obvious if the option wasn't called `Font`, but `Font for alternative languages` or `Alternative font` or something similar, and the tooltip would be updated to describe this more clearly (i.e. the default font is used, and can't be changed, for languages supported by it, and the alternative font is used for any other languages).

**What version of CorsixTH are you using?**
0.65-beta2, Flathub

TheCycoONE:
If you were interested in submitting a PR, this is a change to: https://github.com/CorsixTH/CorsixTH/blob/6aea7c0a943cc62da5fd74b998a527286dac5cc9/CorsixTH/Lua/languages/english.lua#L488 and https://github.com/CorsixTH/CorsixTH/blob/6aea7c0a943cc62da5fd74b998a527286dac5cc9/CorsixTH/Lua/languages/english.lua#L503

As long as it fits in the current UI, it should be minor enough to sneak into 0.65

The behaviour of this option will likely change when #1901 is addressed, but it's a reasonable stopgap measure.

kparal:
Unfortunately my life balance doesn't allow me to become a code contributor at this moment :disappointed:

Patil2099:
I would love to help in solving this issue @TheCycoONE. Would you guide me a little in solving this?

lewri:
> I would love to help in solving this issue @TheCycoONE. Would you guide me a little in solving this?
I believe we just need to make those two English strings clearer for people. By current UI it just means making sure any change won't cause badly placed text.

You can test a change on those strings by editing the copy of English.lua in your install. (Found in Lua > Languages > english.lua)

mefistotelis:
I don't think it's just a matter of two strings.

My experience:
* Installed CorixTH
* Noticed there is no "polish" in the language list
* Found old fan translation of the original game, installed it on data folder
[...]

Issue #1901, opened 2021-05-26
Bundle international fonts and allow users to select all translations available


kparal:
I read on CorsixTH wiki that Czech translation should be available, but it isn't shown in the game menu. I started dig into the issue, tested Flatpak on Linux and also Windows (same behavior), and ultimately found related issues in #1213, #1219 and #823. When I look into source code, there are 20 different languages supported through translations. However, only 12 of them are available through the game menu. I think this is really unfortunate and harms the project a lot.

I started to experiment with font selection, and found it to be a **terrible** experience, definitely not fit for regular users. First, the file browser starts in `/`, and users have no idea where fonts are stored. I found some known fonts in `/usr/share/fonts`. I tried Dejavu Sans and Liberation Mono, both of which support Czech (and most world languages as well), and Linux distributions use them by default everywhere (including right now when writing this report). Both of them look terrible in CorsixTH, I have no idea why. They are almost unreadable. Dejavu is so broken that e.g. the letter "J" is never rendered, just missing, for some reason.

Furthermore, font experimentation is very tedious, because the font file selection doesn't remember the last selected directory, so you have to traverse the whole structure every time. On top of that, the _Folders_ menu crashes the game when entered (at least in Czech) [1], so you have to constantly switch back to English, change the font, then switch back to your language of choice.

I wonder if all of this is really necessary? How many games offer you to select an arbitrary font (and break the game in the process)? Wouldn't it be better to simply pick two, three open source fonts, which can display most or all world languages when combined, and bundle them into the game? The necessary font would be automatically used for a language that needs it. It would be easy for users, no custom fonts difficulties. It would be easy for developers, a consistent experience instead of random issues arising from random font selection. Power users might still be able to change the font, if you prefer, but just through the config file, the GUI option could be removed. I believe this is much better approach than asking users to deal with font issues, or depending on fonts installed in the system.

I believe this would allow CorsixTH to reach much wider audience. Currently almost half the translations are hidden behind a really difficult and fickle barrier.

Thanks for consideration.

[1]
```
[...]

PR #2025, opened 2021-07-23
Improve font file load

tobylane:
Look for a file at`unicode_font`location or the OS specific presets, or warn none of the above were found.

### DIFF
--- a/CorsixTH/Lua/graphics.lua
+++ b/CorsixTH/Lua/graphics.lua
@@ -120,23 +120,37 @@ end
 --! Tries to load the font file given in the config file as unicode_font.
 --! If it is not found it tries to find one in the operating system.
 function Graphics:loadFontFile()
+  local lfs = require("lfs")
+  local function check(path) return path and lfs.attributes(path, "mode") == "file" end
   -- Load the Unicode font, if there is one specified.
-  local font_file = self.app.config.unicode_font
-  if not font_file then
-    -- Try a font which commonly comes with the operating system.
-    local windir = os.getenv("WINDIR")
-    if windir and windir ~= "" then
-      font_file = windir .. pathsep .. "Fonts" .. pathsep .. "ARIALUNI.TTF"
-    elseif self.app.os == "macos" then
-      font_file = "/Library/Fonts/Arial Unicode.ttf"
-    else
-      font_file = "/usr/share/fonts/truetype/arphic/uming.ttc"
-    end
+  local config_path = self.app.config.unicode_font
+  -- Try a font that commonly comes with the operating system.
+  local os_path, font_file
+  local windir = os.getenv("WINDIR")
+  if windir and windir ~= "" then
+    os_path = windir .. pathsep .. "Fonts" .. pathsep .. "ARIALUNI.TTF"
+  elseif self.app.os == "macos" then
+    os_path = "/Library/Fonts/Arial Unicode.ttf"
+  else
+    os_path = "/usr/share/fonts/truetype/arphic/uming.ttc"
   end
-  font_file = font_file and io.open(font_file, "rb")
-  if font_file then
-    self.ttf_font_data = font_file:read"*a"
-    font_file:close()
+  if check(config_path) then font_file = config_path
+  elseif check(os_path) then
+    font_file = os_path
+    print("Configured unicode font not found, using " .. font_file .. " instead.")
+    print("This will be written to the config file.")
+  elseif config_path ~= nil then
+    print("Configured unicode font not found, no fallback available.")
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/graphics.lua
```diff
@@ -120,18 +120,25 @@ end
...
-    end
+  local config_path = self.app.config.unicode_font
+  -- Try a font which commonly comes with the operating system.
+  local os_path, font_file
+  local windir = os.getenv("WINDIR")
+  if windir and windir ~= "" then
+    os_path = windir .. pathsep .. "Fonts" .. pathsep .. "ARIALUNI.TTF"
+  elseif self.app.os == "macos" then
+    os_path = "/Library/Fonts/Arial Unicode.ttf"
+  else
```
lewri, line 134: Maybe be more exclusive here? We're assuming only Linux is left (even though it probably is)

CorsixTH/Lua/graphics.lua
```diff
@@ -120,18 +120,25 @@ end
...
+    os_path = "/usr/share/fonts/truetype/arphic/uming.ttc"
+  end
+  if check(config_path) then font_file = config_path
+  elseif check(os_path) then font_file = os_path
+  else
+    print("Could not find a Unicode font file")
```
lewri, line 140: We never explicitly told the user before about this. And not sure they need to. Maybe gamelog it instead.

CorsixTH/Lua/graphics.lua
```diff
@@ -120,18 +120,25 @@ end
...
-    local windir = os.getenv("WINDIR")
-    if windir and windir ~= "" then
-      font_file = windir .. pathsep .. "Fonts" .. pathsep .. "ARIALUNI.TTF"
-    elseif self.app.os == "macos" then
-      font_file = "/Library/Fonts/Arial Unicode.ttf"
-    else
-      font_file = "/usr/share/fonts/truetype/arphic/uming.ttc"
-    end
+  local config_path = self.app.config.unicode_font
+  -- Try a font which commonly comes with the operating system.
```
lewri, line 127: `which` or `that`?

CorsixTH/Lua/graphics.lua
tobylane: We generally assume that it'll follow the Filesystem Hierarchy Standard, and it's left for other platform developers to add something.

CorsixTH/Lua/graphics.lua
tobylane: Agreed, changed it to only message when there is a value to the config path. Has to be print, the gamelog isn't available yet. Should this be a language string? In a dialog?

lewri: Maybe now I see why you previously wanted gamelog in app. It could warrant a UIInformation then. And it would need a language string.

TheCycoONE: Falling back to the OS fonts isn't new, the new part is doing so even if a font is chosen in the config and it's not valid, and warning the user in that case if we fall through.

If we're tweaking the messaging for the player (who happens to have their console open), I would print when config_path is specified and not found regardless of whether we succeed in a fallback or not.

`Configured unicode font not found, using '....' instead.`
or
`Configured unicode font not found, no fallback available.`

tobylane: Done, plus saved the setting so that a warning isn't repeated.

CorsixTH/Lua/graphics.lua
```diff
@@ -120,23 +120,37 @@ end
...
+  else
+    os_path = "/usr/share/fonts/truetype/arphic/uming.ttc"
+  end
+  if check(config_path) then font_file = config_path
+  elseif check(os_path) then
+    font_file = os_path
+    print("Configured unicode font not found, using " .. font_file .. " instead.")
+    print("This will be written to the config file.")
+  elseif config_path ~= nil then
+    print("Configured unicode font not found, no fallback available.")
```
lewri, line 143: We still have the game font though, right?

tobylane: Yes, this is preparation for the player to pick a unicode language. That line could say that only the non-unicode languages are available?

lewri: I think we'll need to bear in mind #1914 and #1915 . How do you see this as a larger plan of attack of those? Open question.

I'm not against this pr though, don't get me wrong.

tobylane: Maybe this is a misstep towards the greater solution described in #1901

PR #2026, opened 2021-07-25
Announce VIP arrival in local hospital only

tobylane:
- Move vip:announce and the block that calls it to player hospital

### DIFF
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -45,6 +45,7 @@ function PlayerHospital:PlayerHospital(world, avail_rooms, name)
   }
 
   self.win_declined = false -- Has not yet declined the level win fax
+  self.announce_vip = 0 -- Number of spawned VIPs who need to be announced
 end
 
 --! Give advice to the player at the end of a day.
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -45,6 +45,7 @@ function PlayerHospital:PlayerHospital(world, avail_rooms, name)
   }
 
   self.win_declined = false -- Has not yet declined the level win fax
+  self.announce_vip = 0 -- Whether there is a VIP who needs to be announced
```
lewri, line 48: Predates you, but using 0 and -1 seems sloppy in practice.

lewri (approved):

CorsixTH/Lua/hospitals/player_hospital.lua
TheCycoONE: Because this wasn't a boolean, it's a counter, every time a VIP is added that should be announced it's incremented and when they are announced it's decremented; presumably in case multiple VIPs arrive on the same day?

TheCycoONE:
https://github.com/CorsixTH/CorsixTH/blob/6baf31a07447060db0ad80c345266ec2c132fadb/CorsixTH/Lua/world.lua#L569 is I suppose fine because that is always a PlayerHospital. That said we prefer a function over outside manipulation. `PlayerHospital:onSpawnVip` seems appropriate.

TheCycoONE (approved):
LGTM

CorsixTH/Lua/hospitals/player_hospital.lua
lewri: Yes it's a good point. Not that more than one should be arriving together

tobylane:
Added onSpawnVIP. VIPs can only overlap by cheating, and it safely fails to announce the second one. The change to boolean doesn't need an afterload to function, though there would briefly be a difference between hospitals with 0 and false.

TheCycoONE:
> Added onSpawnVIP. VIPs can only overlap by cheating, and it safely fails to announce the second one. The change to boolean doesn't need an afterload to function, though there would briefly be a difference between hospitals with 0 and false.

0 is true in lua, this is going to announce at the end of the first day for all saves...

tobylane:
Yup, runs the checks for relevant VIP too much. Fixed, and merged two ifs.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -45,6 +45,7 @@ function PlayerHospital:PlayerHospital(world, avail_rooms, name)
   }
 
   self.win_declined = false -- Has not yet declined the level win fax
+  self.announce_vip = false -- Whether there is a VIP who needs to be announced
```
TheCycoONE, line 48: You went back to numbers so this should too.

TheCycoONE (changes requested):

TheCycoONE (approved):

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -442,13 +443,39 @@ function PlayerHospital:announceRepair(room)
...
+  if self.announce_vip > 0 then
+    -- check if the VIP is in the building yet
+    for _, e in ipairs(self.world.entities) do
+      if e.humanoid_class == "VIP" and e.announced == false and
+          self:isInHospital(e.tile_x, e.tile_y) then
+        -- play VIP arrival sound and show tooltips
+        local ui = self.world.ui
+        -- there is also vip008 which announces a man from the ministry
+        ui:playRandomAnnouncement({ "vip001.wav", "vip002.wav", "vip003.wav",
+            "vip004.wav", "vip005.wav" }, AnnouncementPriority.High)
```
lewri, line 467: Some changes I made to the announcer drops duplicate announcements out the queue. But the realistic likelihood of that happening is very low. Yet, I see no issue to hold this pr up.

lewri (approved):

PR #2035, opened 2021-08-05
Update screenshot URLs in metainfo

rmnvgr:
- Update screenshot URLs in metainfo to the new images

**Do not merge until corsixth.com points to the new website**

### DIFF
--- a/CorsixTH/com.corsixth.corsixth.metainfo.xml
+++ b/CorsixTH/com.corsixth.corsixth.metainfo.xml
@@ -107,10 +107,10 @@
   </releases>
   <screenshots>
     <screenshot type="default">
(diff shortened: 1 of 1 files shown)


TheCycoONE:
Question - when corsixth.com moves there will be a disruption - does the github url always work? Perhaps we should switch to that domain now to avoid any disruption?

rmnvgr:
The GitHub URLs will always work indeed. Note that Flathub downloads a copy of the images to serve them itself, so it won't disrupt anything there. Not sure about other package front-ends though.

I can of course change to GitHub URLs.

TheCycoONE:
Site has been moved, we probably want this now

rmnvgr:
Nice! I checked that the URLs were correctly loading, this is ready to be merged.

PR #2037, opened 2021-08-07
Remove .0 from negatives in Annual Report

### LINKED ISSUE #2032: Awards are in the wrong numeric type
From #2031 regarding the awards screen: It may have been due to the "bad award" I got -60.0 reputation (also why not use ints for all awards maybe).

Reputation is always an integer, it shouldn't have a decimal after it.

lewri:
*Fixes #2032*

- Adds a helper function to make sure we represent awards with whole numbers only.

Only question is should it be in utility? If it is meant to be utility, it would need more documenting for use cases.

Edit: I don't think this calls for an afterload really. It would fix itself in the next annual report anyway.

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/annual_report.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/annual_report.lua
@@ -40,11 +40,6 @@ local function setActive(button, active)
   end
 end
 
--- Helper function to make sure all values are whole numbers
-local function toInt(number)
-  return math.floor(number)
-end
-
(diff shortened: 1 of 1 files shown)


TheCycoONE:
If we're going to add a toInt it needs to follow the semantics of math.tointeger and preferably be a polyfill that falls away in 5.3+. This doesn't - I suggest simply calling math.floor at point of use unless to want to go to that effort.

lewri:
Reverted  to math.floor

TheCycoONE:
Yep, looks ok. As I mentioned in chat, string:format is best.

PR #2050, opened 2021-09-16
Extra info on replacing machines

### LINKED ISSUE #2049: Add machine strength indicator on replacing machine prompt
Suggested in discord. Add the ability to see what strength a machine will have when buying a replacement.
This opens the question to needing a TH97 toggle down the line to try and manage QoL improvements that don't align with the original gameplay.

Will let players see if replacing a machine is going to be worth it (e.g. less repairs?)

No

 0.65.1

No

**Additional information**

lewri:
*Fixes #2049*

- Adds extra information to the replace machine dialog (current machine strength and new machine strength)
- Made into a config setting. But later down the line should likely form part of a QoL settings menu.
- Also moves the hospital spend money line into the `replaceMachine()` function in Machine (renamed from `machineReplaced()`.
- Game pauses on the confirmation dialog -- so a machine can't explode while replacing it.

image

### DIFF
--- a/CorsixTH/Lua/config_finder.lua
+++ b/CorsixTH/Lua/config_finder.lua
@@ -141,7 +141,8 @@ local config_defaults = {
   check_for_updates = true,
   room_information_dialogs = true,
   allow_blocking_off_areas = false,
-  direct_zoom = nil
+  direct_zoom = nil,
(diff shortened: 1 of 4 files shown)


lewri:
Have tested it in different font types and languages -- all works fine.

PR #2069, opened 2021-10-30
Remove IRC Notice

lewri:
Has now been 4/5 months since the migration. No issues reported about this and seems that almost everyone has moved to Discord or Matrix. Suggest to remove notice.


lewri:
The IRC on freenode has likely also been shut down now -- I was the last person to leave it ~90 days ago.

PR #2071, opened 2021-11-11
[RFC] No rest cheat

lewri:
- New cheat from fax to toggle fatigue off and on
- Also sets max speed for all staff
- Some doc
- Moves toggle cheats into their own set
- Removes cheat processing code from Fax
- Adjusts announceCheat to accomodate adviser text.
- active_cheats in hospital is now a set

Some code lines probably need tidying, any feedback appreciated
Code to test: 363749


CorsixTH/Lua/languages/english.lua
```diff
@@ -260,6 +260,8 @@ adviser = {
     th_cheat = "Congratulations, you have unlocked cheats!",
     roujin_on_cheat = "Roujin's challenge activated! Good luck in the coming months...",
     roujin_off_cheat = "Roujin's challenge deactivated. Everything will be back to normal soon.",
+    norest_on_cheat = "It appears you staff have overdosed on caffeine and no longer feel a need to rest.",
```
Alberth289346, line 263:
```suggestion
    norest_on_cheat = "It appears your staff have overdosed on caffeine and no longer feel a need to rest.",
```

CorsixTH/Lua/languages/english.lua
```diff
@@ -260,6 +260,8 @@ adviser = {
...
+    norest_off_cheat = "Phew! Looks like that buzz finally wore off. Your staff can now rest properly.",
```
Alberth289346, line 264:
```suggestion
    norest_off_cheat = "Phew! Looks like that buzz finally wore off. Your staff will now rest properly.",
```
?

CorsixTH/Lua/cheats.lua
```diff
@@ -169,3 +169,50 @@ function Cheats:cheatDecreasePrices()
...
+  end
+end
+
+--[[Cheats operated through Faxes go here]]
+--! Cheat that disables/enables Roujin's challenge (spawn rate cheat)
+function Cheats:cheatRoujin()
+  -- Roujin's challenge cheat
+  local hosp = self.hopsital
+  local ui = hosp.world.ui
+  if not hosp.spawn_rate_cheat then
```
Alberth289346, line 189: Instead of throwing lots of random names in the hospital name space, add a table with a useful name (eg `done_cheats`), and store this information in there.

Alberth289346: or `active_cheats` ?

CorsixTH/Lua/cheats.lua
```diff
@@ -169,3 +169,50 @@ function Cheats:cheatDecreasePrices()
...
+
+--! Cheat that disables/enables need to rest and set walking speed to maximum
+function Cheats:cheatNoRest()
+  local hosp = self.hospital
+  local ui = hosp.world.ui
+  if not hosp.no_rest_cheat then
+    for _, staff in ipairs(self.hospital.staff) do
+      if staff.attributes["fatigue"] then
+        staff:wake(staff.attributes["fatigue"])
+      end
```
Alberth289346, line 209: Here you compensate for the fatique, and in staff fatique update you skip updating.
Wouldn't it be simpler to perform this action there so you don't have it here, and fatique updates are more closely together?

CorsixTH/Lua/cheats.lua
```diff
@@ -169,3 +169,50 @@ function Cheats:cheatDecreasePrices()
...
+
+--! Cheat fax function
+function Cheats:cheatByFax()
+  local cheatWindow = self.hospital.world.ui:getWindow(UICheats)
+  self:announceCheat()
+  -- If a cheats window is open, make sure the UI is updated
+  if cheatWindow then
+    cheatWindow:updateCheatedStatus()
+  end
+end
```
Alberth289346, line 181: I don't quite get the name of the function, how does it cheat by fax? it just shows a cheat became active (or non-active perhaps)

CorsixTH/Lua/dialogs/fullscreen/fax.lua
```diff
@@ -263,18 +263,11 @@ function UIFax:validate()
...
-      self.ui.adviser:say(_A.cheats.roujin_off_cheat)
-      -- Clear the current month's spawns to give the player a break
-      self.ui.app.world.spawn_dates = {}
-      hosp.spawn_rate_cheat = nil
-    end
+    local cheat = self.ui.hospital.hosp_cheats
+    cheat:cheatRoujin()
+  elseif 185.5 < x and x < 185.6 then
+    local cheat = self.ui.hospital.hosp_cheats
+    cheat:cheatNoRest()
```
Alberth289346, line 270: I can see the use for a table with `x` value range and the corresponding function, but it may be too early.

Alberth289346 (commented):
quite simple to create this functionality, nice

CorsixTH/Lua/cheats.lua
Alberth289346: Also, simply assign `0.0` to the fatique rather than this complicated update, imho.

lewri: This one I'd want to double check as originally I did set fatigue directly to 0 (technically should be 0.00).
However, the `wake` function does two things:
```lua
function Staff:wake(amount)
  self:changeAttribute("fatigue", -amount) -- This one changes the fatigue value
  self:updateDynamicInfo() -- This one updates the UI
end
```

I could go with the route of mimicking this function manually in cheats, but I thought in this case it would be better to use the proper function (wake). I value the view on moving this all to inside `staff.lua`, and did also try this -- but it generates needless updates to the fatigue value. It seemed easier to just set it once on cheat activation, and the cause any execution of the `tire` function inside `staff.lua` to just exit.
Though I'm happy to move on this if there is a good example.

CorsixTH/Lua/dialogs/fullscreen/fax.lua
lewri: Likely good to bear in mind if the fax cheats extend more over time.

CorsixTH/Lua/cheats.lua
lewri: Still inside hospital I'm guessing? So `hospital.active_cheats` ?

CorsixTH/Lua/languages/english.lua
lewri: I'd also like to open the floor to opinions on using `overdosed` -- the term I feel shouldn't be used with jest personally but had nothing better to fit at the time. Maybe I'm thinking too hard on it :)

lewri:
```suggestion
    norest_on_cheat = "Oh no! It appears your staff have consumed too much caffeine and no longer feel a need to rest.",
```

CorsixTH/Lua/cheats.lua
lewri: Is the new name better?

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -34,6 +34,7 @@ function PlayerHospital:PlayerHospital(world, avail_rooms, name)
   self:Hospital(world, avail_rooms, name)
   -- The player hospital in single player can access the Cheat System should they wish to.
   self.hosp_cheats = Cheats(self)
+  self.active_cheats = {} -- records any active toggle-based cheats
```
lewri, line 37: @Alberth289346 I think this one might need to go into `hospital` rather than `playerhospital` and `aihospital` separately?

CorsixTH/Lua/languages/english.lua
```diff
@@ -260,6 +260,8 @@ adviser = {
     th_cheat = "Congratulations, you have unlocked cheats!",
     roujin_on_cheat = "Roujin's challenge activated! Good luck in the coming months...",
     roujin_off_cheat = "Roujin's challenge deactivated. Everything will be back to normal soon.",
+    norest_on_cheat = "Oh no! It appears your staff have consumed too much caffeine and no longer feel a need to rest.",
```
TheCycoONE, line 263: Oh no!? Oh yes! 😄

CorsixTH/Lua/cheats.lua
TheCycoONE: I think what you did is justified.

CorsixTH/Lua/cheats.lua
TheCycoONE: Not really - functions get awkward when you name them based on their context instead of their function, and I'm having a hard time classifying the two actions this function does while also excluding the act of enabling the cheat which is in the parent.

You could *cheat* and call it something like Cheats:_enableCheatCommon; you could write one function for enabling cheats that takes a cheat type and fields out to helper functions for special activation/deactivation logic.

As a loose example (not meant to copy):
```lua
function Cheats:toggleCheat(cheat)
  local hosp = self.hospital
  local ui = hosp.world.ui
  if not hosp.active_cheats[cheat] then
    ui.adviser:say(cheat_messages[cheat].enable)
    hosp.active_cheats[cheat] = true
    self:announceCheat()
    local cheatWindow = hosp.world.ui:getWindow(UICheats)
    if cheatWindow then
     cheatWindow:updateCheatedStatus()
    end
    cheat_fns[cheat].enable(self)
  else
    ui.advistor:say(cheat_messages[cheat].disable)
    cheat_fns[cheat].disable(self)
    hosp.active_cheats[cheat] = nil
  end
end
```

CorsixTH/Lua/languages/english.lua
lewri: I nearly put coffee, but that'd be a sin

lewri:
Sorry for the noise on that one. Have given another go at this now, just took a while.

CorsixTH/Lua/cheats.lua
```diff
@@ -60,11 +103,52 @@ function Cheats:performCheat(num)
...
+  local speech
+  if not self.hospital.active_cheats[name] then
+    cheat.enable(self)
+    speech = self._getAdviserMessage(name, true)
+    self:announceCheat(speech)
+  else
+    cheat.disable(self)
+    speech = self._getAdviserMessage(name, false)
+    self:announceCheat(speech)
+  end
```
lewri, line 135: Move repeated function call to 136
```suggestion
  else
    cheat.disable(self)
    speech = self._getAdviserMessage(name, false)
  end
  self:announceCheat(speech)
```

lewri:
Closed for 2095

PR #2080, opened 2021-12-01
Dynamic Info - Machine replacement strength

RetroGamerFan:
*Fixes #*
- Modified machine.lua so the replacement machine strength is set for the formatted strength string.
- english.lua modified so the strength string displays the additional info.
- research_department.lua - When machine strength is improved, any rooms with the same machine are instructed to update their dynamic info.

- When the mouse cursor hovers over a machine, the replacement machine strength is shown next to the machine strength.

image


CorsixTH/Lua/research_department.lua
```diff
@@ -478,6 +478,15 @@ function ResearchDepartment:improveMachine(machine)
     research_info.start_strength = research_info.start_strength
       + improve_rate
     research_info.strength_imp = research_info.strength_imp + 1
+    -- Update dynamic info for any machines that now have improved strength.
+    for _, room in ipairs(self.world.rooms) do
+      for obj in pairs(room.objects) do
+        if(obj.object_type.id == machine.id) then
+          obj:updateDynamicInfo(false)
+          break
```
lewri, line 486: Your `break` here ends the for loop, but what if there's multiple rooms that need this update? Maybe we have 2 tongue slicers, but here it would only update one.

lewri (changes requested):
I'm not convinced on how you've implemented it into the dynamic info.
Additionally, if you move forward with this I would recommend you use the config flag `new_machine_extra_info` to allow people to optionally turn this enhancement off. 

https://github.com/CorsixTH/CorsixTH/blob/ac268008795b4ac1cf8f2442d3a6e9f292d62d77/CorsixTH/Lua/config_finder.lua#L527-L532

.github/workflows/main.yml
```diff
@@ -0,0 +1,36 @@
+# This is a basic workflow to help you get started with Actions
```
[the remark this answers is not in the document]
RetroGamerFan: Github did it when I created a workflow to run the luachecks.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -358,7 +358,8 @@ function Machine:updateDynamicInfo(only_update)
   if self.strength then
     self:setDynamicInfo("text", {
       self.object_type.name,
-      _S.dynamic_info.object.strength:format(self.strength),
+      -- _S.dynamic_info.object.strength:format(self.strength),
+      _S.dynamic_info.object.strength:format(self.strength, self.world.ui.hospital.research.research_progress[self.object_type].start_strength),
```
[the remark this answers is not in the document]
RetroGamerFan: Just copied the line from earlier up in the same file for fetching the strength info.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -358,7 +358,8 @@ function Machine:updateDynamicInfo(only_update)
   if self.strength then
     self:setDynamicInfo("text", {
       self.object_type.name,
-      _S.dynamic_info.object.strength:format(self.strength),
+      -- _S.dynamic_info.object.strength:format(self.strength),
```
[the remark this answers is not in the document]
RetroGamerFan: Do you have a coding rules doc somewhere. Some people like to comment out older code for 1 or more released before deleting them?

CorsixTH/Lua/languages/english.lua
```diff
@@ -268,6 +268,7 @@ dynamic_info.staff.actions.heading_for = "Heading for %s"
 dynamic_info.staff.actions.fired = "Fired"
 dynamic_info.staff.actions.vaccine = "Vaccinating a patient"
 dynamic_info.patient.actions.epidemic_vaccinated = "I am no longer contagious"
+dynamic_info.object.strength = "Strength %d (Replacement %d)"
```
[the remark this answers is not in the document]
RetroGamerFan: It's called DynamicInfo in the code and the screenshot shows the info in the HUD info bar - if it has a batter name what is it?

CorsixTH/Lua/research_department.lua
RetroGamerFan: I've tested this on a hospital with multiple rooms of the same type. The break only breaks from the existing loop - i.e. no need to look for further objects in that room and move on to the next room (in the outer for loop).

RetroGamerFan:
Why would such a tiny change in terms of interface or gameplay need a config option?

RetroGamerFan:
I think the latest changes covers everything.

NOTE! Dynamic text is not updated when loading a game.

Also saw the same issue when switching language to French and loading a game.

I see this as an existing bug outside the scope of this change.

CorsixTH/Lua/entities/machine.lua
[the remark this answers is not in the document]
lewri: Yes, here you go:
https://github.com/CorsixTH/CorsixTH/wiki/Coding-Conventions
Github will automatically preserve code history so we don't have a need to record it directly in the codebase.

CorsixTH/Lua/research_department.lua
lewri: Oh yeas, ignore me :P

lewri:
For your query on why it should be a configurable option -- the ability to see ahead on what a new machine's strength will be was not part of the original game.

Effectively the related PR #2050 is a quality of life (QoL) update that serves a similar purpose to this PR -- it allows you to preview strength on a new machine before getting it. Thus, why it should come under the configuration option.
You'll also need code to handle if the configuration option is off (i.e. the old setting of showing just self.strength).

CorsixTH/Lua/languages/english.lua
```diff
@@ -269,6 +269,10 @@ dynamic_info.staff.actions.fired = "Fired"
 dynamic_info.staff.actions.vaccine = "Vaccinating a patient"
 dynamic_info.patient.actions.epidemic_vaccinated = "I am no longer contagious"
 
+if self.ui.app.config.new_machine_extra_info then
+  dynamic_info.object.strength = "Strength %d (Replacement %d)"
+end
+
```
lewri, line 275: Logic does not belong inside a language file, it'll need to be in `machine.lua`; this may mean you'll need two versions of the string, or break it down (since I believe `dynamic_info.object.strength` exists) so your new part of the string is 
`dynamic_info.object.strength_extra_info = " (Replacement %d)"` -- note the space.
On the text part, maybe you could have "(Upgradeable to %d)" instead? Or "(Replaceable to %d)"?

Then you can join them together
```lua
local strength_text = (_S.dynamic_info.object.strength):format(self.strength)
if hosp.ui.app.config.new_machine_extra_info then
  local start_strength = hosp.research.research_progress[self.object_type].start_strength
  strength_text = strength_text .. (_S.dynamic_info.object.strength_extra_info):format(start_strength)
end
self:setDynamicInfo("text", {
  self.object_type.name,
  strength_text,
  _S.dynamic_info.object.times_used:format(self.times_used),
})
```
Not that you have to use that... but if you do I may have made a booboo somewhere writing on the fly.

RetroGamerFan: Some languages might take issue with how strings are concatenated in the code. Perhaps in the future the strings in the language lua files can dictate which variable value is placed where inside the text to be formatted.

CorsixTH/Lua/entities/machine.lua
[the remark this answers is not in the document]
RetroGamerFan: I don't see an applicable rule for this. Is code deletion rather than commenting out the old code the norm on Github?

CorsixTH/Lua/research_department.lua
```diff
@@ -478,6 +478,17 @@ function ResearchDepartment:improveMachine(machine)
     research_info.start_strength = research_info.start_strength
       + improve_rate
     research_info.strength_imp = research_info.strength_imp + 1
+    if self.world.ui.app.config.new_machine_extra_info then
+      -- Update dynamic info for any machines that now have improved strength.
+      for _, room in ipairs(self.world.rooms) do
+        for obj in pairs(room.objects) do
+          if(obj.object_type.id == machine.id) then
```
lewri, line 485:
```suggestion
          if (obj.object_type.id == machine.id) then
```

lewri (commented):
Only that typo. I have no glaring issues personally otherwise. I imagine Alberth and Cyco will eventually pick this up.

CorsixTH/Lua/research_department.lua
RetroGamerFan: LOL - those brackets are a product of years of habit - fixed now. Cheers.

CorsixTH/Lua/research_department.lua
```diff
@@ -478,6 +478,17 @@ function ResearchDepartment:improveMachine(machine)
     research_info.start_strength = research_info.start_strength
       + improve_rate
     research_info.strength_imp = research_info.strength_imp + 1
+    if self.world.ui.app.config.new_machine_extra_info then
+      -- Update dynamic info for any machines that now have improved strength.
+      for _, room in ipairs(self.world.rooms) do
+        for obj in pairs(room.objects) do
```
Alberth289346, line 484: `pairs`  returns 2 values

CorsixTH/Lua/research_department.lua
```diff
@@ -478,6 +478,17 @@ function ResearchDepartment:improveMachine(machine)
     research_info.start_strength = research_info.start_strength
       + improve_rate
     research_info.strength_imp = research_info.strength_imp + 1
+    if self.world.ui.app.config.new_machine_extra_info then
+      -- Update dynamic info for any machines that now have improved strength.
+      for _, room in ipairs(self.world.rooms) do
```
Alberth289346, line 483: Make a new function `updateMachinesDynamicInfo` and call that function here.

In that way, other causes for updating dynamic info can re-use this code. In that way, we 're avoiding to duplicate code.

Alberth289346: Note that `self.world.rooms` is all rooms in the world, not just in the player hospital.

You may want to verify the hospital of the room to limit the machines that you touch.

CorsixTH/Lua/research_department.lua
RetroGamerFan: Fixed and also for the room loop changed "ipairs" to "pairs".

CorsixTH/Lua/research_department.lua
RetroGamerFan: Done and inside the room loop added "if room.hospital == self.hospital then"

CorsixTH/Lua/research_department.lua
```diff
@@ -499,6 +489,23 @@ function ResearchDepartment:improveMachine(machine)
   end
 end
 
+--[[ Updates the dynamic info for each matching machine found in the player's
+hospital rooms.
+!param machine The machine for determining matches.
+--]]
+function ResearchDepartment:updateMachinesDynamicInfo(machine)
```
Alberth289346, line 496: The parameter `machine` seems to suggest that you will change only one machine (namely the one you pass in), while in fact you change all machines of a given type (eg all inflation machines in the hospital).

To make this more clear, instead of passing in `machine` pass in `machine.id` in the call. Also change the parameter name to something ending in `id` (that is the core data on which you make decisions in the code). For example `machine_id`. Line 500 simply uses the name name (eg `machine_id`). Line 494 also changes a bit.

RetroGamerFan: Done.

CorsixTH/Lua/research_department.lua
```diff
@@ -486,6 +489,23 @@ function ResearchDepartment:improveMachine(machine)
   end
 end
 
+--[[ Updates the dynamic info for each matching machine found in the player's
+hospital rooms.
+!param machine_id The machine ID for determining matches.
+--]]
```
lewri, line 495: I don't think we usually wrap luadoc comments any more (ignoring the already existing ones)... but I could be wrong.

```suggestion
--! Updates the dynamic info for each matching machine found in the player hospital's rooms.
--!param machine_id The machine ID for determining matches.
```

RetroGamerFan: The other functions have wrapped text and I see wrapped code as well. Consider it a sperate issue IMO.

Alberth289346: With long enough explanations we do wrap lines afaik. But doc changes are indeed a different issue.

@lewri LGTM, any other things you want changed? If not seems mergeable to me.

Alberth289346: Probably squash it? 22 commits seems a bit too much.

RetroGamerFan: There's very few changes comparing the original files to the final versions. Would it be easier to cancel this pull request and make a new one?

lewri: No, don't worry. I do it all the time :upside_down_face:

TheCycoONE:
Verify, but I believe you want an updateDynamicInfo unconditionally in afterLoad in case the option is different from the save. It will correct on its own fairly quickly, but it's quirk that's better not to have.

RetroGamerFan:
"I believe you want an updateDynamicInfo unconditionally in afterLoad" - see comment "RetroGamerFan commented on 3 Dec 2021"

TheCycoONE:
> "I believe you want an updateDynamicInfo unconditionally in afterLoad" - see comment "RetroGamerFan commented on 3 Dec 2021"
> 
> Bug #2081

Looks like one more reason to add the line 😄

RetroGamerFan:
"Looks like one more reason to add the line" - you know best mate. I see nothing wrong with quick fixes for those that know what they are doing. 👍

lewri:
I'm assuming the change wanted on this is adding `self:updateDynamicInfo(true)` to `Machine:afterLoad`?

RetroGamerFan:
It's a separate issue if I remember - an existing bug you can experience by
changing the language before loading a saved game.

TheCycoONE:
It was but there was enough change around it here that I was looking for that extra line yes.

RetroGamerFan:
Do what you think is best.

TheCycoONE:
Not clear @RetroGamerFan, will you add the afterLoad or must I merge without it and open an issue?

RetroGamerFan:
It should already be open as a separate issue.

lewri:
In that case, do you want to fix #2081 in this pr as well by adding that line to the afterLoad function? Two birds one stone then.

I think what we're also saying is unfortunately we can't push the fix into your pr, it has to come from you. Or we leave #2081 open and don't fix it here (fwiw, it's a valid argument to not fix it here but I think a reasonable exception applies in this case).

It's up to you. We will merge with or without the additional fix but need to know your intention.

PR #2083, opened 2021-12-25
Deduplicate staff-type name translations.

Alberth289346:
Remove duplicates of translating staff type-names to human-readable names.

### DIFF
--- a/CorsixTH/Lua/staff_profile.lua
+++ b/CorsixTH/Lua/staff_profile.lua
@@ -256,3 +256,16 @@ end
 function StaffProfile:getFullName()
   return self.initial .. ". " .. self.name
 end
+
+--! Translate the staff class to its translated text.
+--!param staff_class Class of the staff to translate.
+--!return The translated name.
+function StaffProfile.translateStaffClass(staff_class)
+  local staffclass_to_string = {
+    Nurse        = _S.staff_title.nurse,
+    Doctor       = _S.staff_title.doctor,
+    Surgeon      = _S.staff_title.surgeon,
+    Psychiatrist = _S.staff_title.psychiatrist,
+  }
+  return staffclass_to_string[staff_class]
+end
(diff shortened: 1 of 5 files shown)


CorsixTH/Lua/staff_profile.lua
```diff
@@ -256,3 +256,16 @@ end
...
+--!return The translated name.
+function StaffProfile.translateStaffClass(staff_class)
+  local staffclass_to_string = {
+    Nurse        = _S.staff_title.nurse,
+    Doctor       = _S.staff_title.doctor,
+    Surgeon      = _S.staff_title.surgeon,
+    Psychiatrist = _S.staff_title.psychiatrist,
+  }
+  return staffclass_to_string[staff_class]
+end
```
lewri, line 271: Excuse me as I haven't seen this before, but what's the difference making your function `StaffProfile.translateStaffClass` rather than `StaffProfile:translateStaffClass`?

Alberth289346:
The `.` is the primitive form, it's like the table field selection (`t={a=1}; t.a` selects the `1` value associated with field `a`).

For function definitions:
- `function C.f(x)` thus means function `f` in class `C` taking 1 parameter named `x`.
- `function C:f(x)` is a shorthand notation for `function C.f(self, x)`.

For function calls (with `d = C()`):
- `p = d.f(1)` means `p = C.f(1)` (I think, not sure that the class instance is reduced to its class), but the entire `d` disappears so the better way to call this is through the class `p = C.f(1)`. This also makes very clear that data in `d` is not available.
- `p = d:f(1)` is a shorthand notation for `p = C.f(d, 1)`

In the translate function `self` is not used, so there is no need to include `self` in the call. That also eliminates the need to even have a `StaffProfile` instance available.

Alberth289346:
See also Chapter 16: https://www.lua.org/pil/16.html

lewri (approved):

PR #2084, opened 2021-12-31
Humanoid start action move

Alberth289346:
Make the `Humanoid_startAction` a normal `Humanoid:startAction` method.

Also, move empty action queue handling to its own method to reduce clutter in starting actions code.

### DIFF
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -460,69 +460,18 @@ function Humanoid:getCurrentMood()
   end
 end
 
+--! Start the next (always first) action in the queue.
 function Humanoid:startAction()
   local action = self.action_queue[1]
 
   -- Handle an empty action queue in some way instead of crashing.
   if not action then
-    -- if this is a patient that is going home, an empty
-    -- action queue is not a problem
-    if class.is(self, Patient) and self.going_home then
-      return
-    end
+    self:handleEmptyActionQueue() -- Inserts an action into the action queue.
 
-    ---- Empty action queue! ----
-    -- First find out if this humanoid is in a room.
-    local room = self:getRoom()
-    if room then
-      room:makeHumanoidLeave(self)
-    end
-    -- Is it a member of staff, grim or a patient?
-    if class.is(self, Staff) then
-      self:queueAction(MeanderAction())
-    elseif class.is(self,GrimReaper) then
-      self:queueAction(IdleAction())
-    else
-      self:queueAction(SeekReceptionAction())
-    end
-    -- Open the dialog of the humanoid.
-    local ui = self.world.ui
-    if class.is(self, Patient) then
-      ui:addWindow(UIPatient(ui, self))
-    elseif class.is(self, Staff) then
-      ui:addWindow(UIStaff(ui, self))
-    end
-
-    -- Tell the player what just happened.
-    self.world:gameLog("")
-    self.world:gameLog("Empty action queue!")
-    self.world:gameLog("Last action: " .. self.previous_action.name)
-    self.world:gameLog(debug.traceback())
-
-    ui:addWindow(UIConfirmDialog(ui, true,
-      "Sorry, a humanoid just had an empty action queue,"..
-      " which means that he or she didn't know what to do next."..
-      " Please consult the command window for more detailed information. "..
-      "A dialog with "..
-      "the offending humanoid has been opened. "..
-      "Would you like him/her to leave the hospital?",
-      --[[persistable:humanoid_leave_hospital]] function()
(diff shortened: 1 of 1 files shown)


lewri:
Will need to make a note I'm going to move line 633-648 string into english once this one's done.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -460,69 +460,18 @@ function Humanoid:getCurrentMood()
   end
 end
 
-local function Humanoid_startAction(self)
+--! Start the next (always first) action in the queue.
+function Humanoid:startAction()
```
lewri, line 464: Any perceived issues changing this from local to global?

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -642,6 +592,65 @@ function Humanoid:hasLeavingAction()
...
+  -- First find out if this humanoid is in a room.
+  local room = self:getRoom()
+  if room then
+    room:makeHumanoidLeave(self)
+  end
+
+  -- Give the humanoid an action to avoid crashing.
+  if class.is(self, Staff) then
+    self:queueAction(MeanderAction())
+  elseif class.is(self,GrimReaper) then
```
lewri, line 612:
```suggestion
  elseif class.is(self, GrimReaper) then
```
(sorry, was bothering me)

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -642,6 +592,65 @@ function Humanoid:hasLeavingAction()
...
+  self.world:gameLog(debug.traceback())
+
+  ui:addWindow(UIConfirmDialog(ui, true,
+    "Sorry, a humanoid just had an empty action queue,"..
+    " which means that he or she didn't know what to do next."..
+    " Please consult the command window for more detailed information. "..
+    "A dialog with "..
+    "the offending humanoid has been opened. "..
+    "Would you like him/her to leave the hospital?",
+    --[[persistable:humanoid_leave_hospital]] function()
```
lewri, line 639: Would this be game-breaking if someone saved the game on this error (before this PR) and reloaded with the new code? -- I don't think it should.

lewri (commented):
Nice to see this in its own function now.

CorsixTH/Lua/entities/humanoid.lua
Alberth289346: Don't call it outside this file?  :P 

Nothing currently knows about this method, so that should be fine. Not sure what needs fixing if any.

CorsixTH/Lua/entities/humanoid.lua
Alberth289346: I agree with you. Persistence doesn't care about scope afaik.

TheCycoONE: It's fne

TheCycoONE (commented):
I like the refactoring of the empty action queue code.

I'm not sure I understand the motivation for Humanoid_startAction -> Humanoid:startAction. Anything?

Alberth289346:
It made sense to stick it in the class as well, almost all other code around it is also in the class. It also felt like the responsibility of a humanoid to start actions. Also code is very close to the class, it accesses the humanoid instance variables directly. I am trying to limit external access to make usage more tractable. I agree it's not really needed though.

As a side note, our static Lua analyzer doesn't catch typos, you can read data from a class using a name that is never used elsewhere in all our Lua and it doesn't report it.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -460,69 +460,18 @@ function Humanoid:getCurrentMood()
...
   local action = self.action_queue[1]
 
   -- Handle an empty action queue in some way instead of crashing.
   if not action then
-    -- if this is a patient that is going home, an empty
-    -- action queue is not a problem
-    if class.is(self, Patient) and self.going_home then
-      return
-    end
+    self:handleEmptyActionQueue() -- Inserts an action into the action queue.
```
Alberth289346, line 469:
```suggestion
    self:_handleEmptyActionQueue() -- Inserts an action into the action queue.
```

TheCycoONE: Yep

CorsixTH/Lua/entities/humanoid.lua
Alberth289346: done

CorsixTH/Lua/entities/humanoid.lua
Alberth289346: made the empty action queue handler private

PR #2086, opened 2022-01-02
Use proper actions (hopefully) everywhere

Alberth289346:
At a few points, actions were created as tables with properties rather than by instantiating the required class.
This patch fixes that.

Note that 41b3d1dee77a9393713589ca9c9fd50389e255e2 makes a mess in patient.lua due to whitespace changes.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -1122,39 +1122,36 @@ end
 !param meander_before_resume (bool) Indicator to insert an additional action before resuming.
 ]]
 function Patient:interruptAndRequeueAction(current_action, queue_pos, meander_before_resume)
-  if current_action.name == "idle" or current_action.name == "walk" then
-    local requeue_action
-    if current_action.name == "idle" then
-      requeue_action = IdleAction():setMustHappen(current_action.must_happen)
-      requeue_action:setDirection(current_action.direction)
+  -- Rebuild the current action.
+  local requeue_action
+  if current_action.name == "idle" then
+    requeue_action = IdleAction():setMustHappen(current_action.must_happen)
+    requeue_action:setDirection(current_action.direction)
 
-      -- Go away a little before continuing with that important action.
-      if meander_before_resume then
-        self:queueAction(MeanderAction():setCount(1), queue_pos)
-      end
-    else
-      assert(current_action.name == "walk")
-      requeue_action = WalkAction(current_action.x, current_action.y)
-      requeue_action:setMustHappen(current_action.must_happen)
-
-      -- need to copy the reserve_on_resume, otherwise the new queued action will not
-      -- unreserve on interrupt
-      requeue_action.reserve_on_resume = current_action.reserve_on_resume
-      requeue_action:setIsentering(current_action.is_entering)
+    -- Go away a little before continuing with that important action.
+    if meander_before_resume then
+      self:queueAction(MeanderAction():setCount(1), queue_pos)
     end
-    self:queueAction(requeue_action, queue_pos)
+  elseif current_action.name == "walk" then
+    requeue_action = WalkAction(current_action.x, current_action.y)
+    requeue_action:setMustHappen(current_action.must_happen)
 
+    -- need to copy the reserve_on_resume, otherwise the new queued action will not
+    -- unreserve on interrupt
+    requeue_action.reserve_on_resume = current_action.reserve_on_resume
+    requeue_action:setIsentering(current_action.is_entering)
   else
     -- We were seeking a room, start that action from the beginning
     -- i.e. do not set the must_happen flag.
-    local action = SeekRoomAction(current_action.room_type)
-    if current_action.treatment_room then action:enableTreatmentRoom() end
+    requeue_action = SeekRoomAction(current_action.room_type)
+    if current_action.treatment_room then requeue_action:enableTreatmentRoom() end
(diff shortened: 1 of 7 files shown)


CorsixTH/Lua/queue.lua
```diff
@@ -313,16 +313,22 @@ function Queue:movePatient(index, new_index)
 end
 
 --! Called when reception desk is destroyed, or when a room is destroyed from a crashed machine.
-function Queue:rerouteAllPatients(action)
+--!param room_id Id of the room to reroute to, 'nil' for reception.
+function Queue:rerouteAllPatients(room_id)
   for _, humanoid in ipairs(self) do
     -- check by class type as staff/vips shouldn't get a SeekRoomAction
     if class.is(humanoid, Patient) then
       -- slight delay so the desk is really destroyed before rerouting
```
lewri, line 321: I just realised this comment is slightly poor in context. I can get what it means at least, so maybe nothing to do/scope.

CorsixTH/Lua/queue.lua
```diff
@@ -313,16 +313,22 @@ function Queue:movePatient(index, new_index)
 end
 
 --! Called when reception desk is destroyed, or when a room is destroyed from a crashed machine.
-function Queue:rerouteAllPatients(action)
+--!param room_id Id of the room to reroute to, 'nil' for reception.
+function Queue:rerouteAllPatients(room_id)
   for _, humanoid in ipairs(self) do
     -- check by class type as staff/vips shouldn't get a SeekRoomAction
     if class.is(humanoid, Patient) then
```
lewri, line 320: Not you, but note to self we will need to reroute the vip if he was going to reception, or his vip-ing doesn't begin (I've documented this somewhere else so no need for action here.)

CorsixTH/Lua/queue.lua
```diff
@@ -313,16 +313,22 @@ function Queue:movePatient(index, new_index)
...
-      for k, v in pairs(action) do clone[k] = v end
-      humanoid:queueAction(clone)
+
+      local action
+      if room_id then
+        action = SeekRoomAction(room_id)
+      else
+        action = SeekReceptionAction()
+      end
+      humanoid:queueAction(action)
```
lewri, line 330: Interesting. Does queueAction function have an assert to make sure action isn't blank? Sorry, reviewing on phone is hard :)

lewri (commented):
Nice stuff, I'd be interested giving this a whirl when I'm free

CorsixTH/Lua/queue.lua
Alberth289346: Out of scope here, but imho the solution is poor. Instead of delaying and hoping it works, find a replacement action to perform. (Eg by finding another available desk or lacking that, start idling.)

This sort of hack is also done with the receptionist when you grab the desk.

Basically it means an NPC cannot start looking for a new desk by itself. Instead a desk to aim for needs to be provided externally.

CorsixTH/Lua/queue.lua
Alberth289346: Nope.

lewri: Looks like there's a 'good enough' assert in `queueAction`
```lua
  assert(type(x) == "number", "Invalid value for parameter 'x'")
  assert(type(y) == "number", "Invalid value for parameter 'y'")
  assert(class.is(queue, Queue), "Invalid value for parameter 'queue'")
```

CorsixTH/Lua/queue.lua
lewri: #1681 was closed a while back, but that was my attempt at trying to fix some reception bugs. I could probably make it cleaner now, though.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -1122,39 +1122,36 @@ end
...
-      -- unreserve on interrupt
-      requeue_action.reserve_on_resume = current_action.reserve_on_resume
     end
+  elseif current_action.name == "walk" then
+    requeue_action = WalkAction(current_action.x, current_action.y)
+    requeue_action:setMustHappen(current_action.must_happen)
+
+    -- need to copy the reserve_on_resume, otherwise the new queued action will not
+    -- unreserve on interrupt
+    requeue_action.reserve_on_resume = current_action.reserve_on_resume
```
TheCycoONE, line 1141: Did you mean to leave this one?

TheCycoONE (approved):
Seems sane, there's still on direct field set in your PR, not sure if that's in scope.

CorsixTH/Lua/entities/humanoids/patient.lua
Alberth289346: As in not making a `WalkAction:setReserveOnResume`?   Yes, at least for now.


The action code has callbacks, interrupts, requeueing and a zillion flags like this one for customizing and steering behavior all over the place. To me, it says that it had the nice but naive idea of nice sunny weather sequences of animations, and then it turned out to be much less sunny causing a lot of the complexity.
I don't know all the cases and handling and work-arounds yet, since the structure is hard to read with all this
callback glue. Even if I know it, there is currently no common data storage in place, as functions don't share data except by parameter.

So for now, I just try to move code to a more OO-like solution starting by sticking things into or at least near the `HumanoidAction` derived classes, where #2087 is an easy case. I don't know if that works for everything, nor do I really know where it will lead to, but I see no other way to get forward.

Alberth289346:
This one seems done to me.

PR #2087, opened 2022-01-02
Knock door rewrite

Alberth289346:
Moving code into the KnockDoor class instead of having it in a number of functions.
As such, this also functions as a sort-of (simple) example for the path I have in mind.

Adding the 'humanoid' parameter to the actions should be doable after #2086 . Obviously it should go into the base class eventually.

While this action works, I currently don't know if all the other actions can be converted.

The 'start' callback can quite easily be removed after merge of #2084 .
The 'tick' callback should be removable too, since the humanoid acts as entry point for the timer of the animation.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/knock_door.lua
+++ b/CorsixTH/Lua/humanoid_actions/knock_door.lua
@@ -26,41 +26,70 @@ local KnockDoorAction = _G["KnockDoorAction"]
 --! Constructor for knocking on the door action.
 --!param door (Object) Door to knock on.
 --!param direction (string) Direction of facing.
-function KnockDoorAction:KnockDoorAction(door, direction)
+function KnockDoorAction:KnockDoorAction(humanoid, door, direction)
+  assert(class.is(humanoid, Humanoid), "Invalid value for parameter 'humanoid'")
   assert(class.is(door, Door), "Invalid value for parameter 'door'")
   assert(direction == "north" or direction == "south" or
       direction == "east" or direction == "west",
       "Invalid value for parameter 'direction'")
 
   self:HumanoidAction("knock_door")
+  self.humanoid = humanoid
   self.door = door -- Door to knock on.
   self.direction = direction -- Direction of facing.
+  self.state = nil -- Next step in the animation.
+end
+
+function KnockDoorAction:start()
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/humanoid_actions/knock_door.lua
```diff
@@ -26,41 +26,70 @@ local KnockDoorAction = _G["KnockDoorAction"]
...
+    local direction = self.direction
+    local anims = self.humanoid.door_anims
+    local door = self.door
+    self.must_happen = true
+    local anim = anims.knock_north
+    local flag_mirror = (direction == "west" or direction == "south") and 1 or 0
+    if direction == "east" or direction == "south" then
+      anim = anims.knock_east
+    end
+    self.humanoid:setAnimation(anim, flag_mirror)
```
TheCycoONE, line 71: Save compat issues I think. self.humanoid wasn't defined in the old action.

Alberth289346: Ah yes indeed, look for `KnockDoorAction` instances in all action queues of humanoids, and add `action.humanoid`

CorsixTH/Lua/humanoid_actions/knock_door.lua
```diff
@@ -26,41 +26,70 @@ local KnockDoorAction = _G["KnockDoorAction"]
...
+    return
   end
-  humanoid:setAnimation(anim, flag_mirror)
-  humanoid:setTilePositionSpeed(humanoid.tile_x, humanoid.tile_y)
-  humanoid:setTimer(humanoid.world:getAnimLength(anim), action_knock_door_tick)
-  humanoid.user_of = door
-  door:setUser(humanoid)
-  door.th:makeVisible()
+
+  error("Illegal state " .. tostring(self.state) .. " in KnockDoorAction:update")
```
lewri, line 92: Illegal state eh?

CorsixTH/Lua/humanoid_actions/knock_door.lua
Alberth289346: rebased the branch and fixed the issue by adding a 3rd commit

PR #2089, opened 2022-01-07
Set volume straight after music is set playing

### LINKED ISSUE #1978: [Bug] CorsixTH does not obey config volume settings
#### Describe the issue
At least in Windows, music volume in config has no effect on music volume in the game. It will always play at the same level on launch. You have to click a different volume in the game then go back to original to have the correct volume set.
NB: is affecting all 3 volume settings.

Found while setting #1977
#### Steps to Reproduce
1. Set config music_volume to 0.1 and save file.

2. Open game and play music. Change music volume in game to 0.2, then back to 0.1.

#### Expected Behavior
Obey sound volumes.

#### System Information
**CorsixTH Version:** 0.64, 0.65-beta2, Appveyor 588 (jellyitis)

#### Gamelog.txt

```
Paste gamelog.txt output here
```

### Additional Info
illustrate the problem.*

mugmuggy:
*Fixes #1978*

- Sets the volume after starting playing the track - SDL2_mixers midi volume function on windows uses a handle to the stream to set the volume which requires that Mix_PlayMusic is called to open that handle initially.

### DIFF
--- a/CorsixTH/Lua/audio.lua
+++ b/CorsixTH/Lua/audio.lua
@@ -617,8 +617,8 @@ function Audio:playBackgroundTrack(index)
       end)
       return
     end
-    SDL.audio.setMusicVolume(self.app.config.music_volume)
     assert(SDL.audio.playMusic(music))
+    SDL.audio.setMusicVolume(self.app.config.music_volume)
     self.background_music = music
 
     self:notifyJukebox()


lewri:
It looks like this fixes the first song playback. Unfortunately, however it isn't sent until music is played.
The volume of the opening music is still stuck sadly until you get to main menu (looks like video still not sending SDL volume update)

mugmuggy:
I think we'd need a fix for 524 to actually get that to happen. Windows is remembering the last volume set, which due to 524 is the volume of the music in game from the last time its run. What you really need is that full system volume range is used and then the ratios apply. I have a hacky patch in mind for sdl2_mixer for the 524 issue, I just need to see how it integrates with the full behaviour once integrated.

lewri (approved):
Happy to proceed with this, but should probably make a new issue (or adjust 524) about the specific intro video bug.

lewri:
@mugmuggy do you know if our changes to SDL2/Fluidsynth for Windows in 0.67 still require this PR?

lewri:
@TheCycoONE is this still required with Windows changes? I can't imagine it hurts still having this?

TheCycoONE:
My understanding was that fluidsynth fixed it but I don't mind merging.

PR #2093, opened 2022-01-16
Seek room action faxes refactoring

Alberth289346:
- Refactor construction of the guess-cure fax and no-treatment-room fax to their own functions.
- Extend Patient:updateMessage() with 'guess_cure' update as already promised by the documentation, and use it in research.

Ignoring whitespace changes is helpful for getting a smaller diff.

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -269,7 +269,7 @@ end
 --! Update the 'guess the cure' faxes of waiting patients.
 --!param disease_id Disease to update.
 function Hospital:updateGuessCureFaxes(disease_id)
-  for _, patient in ipairs(self.hospital.patients) do
+  for _, patient in ipairs(self.patients) do
     if patient.disease.id == disease_id then
       patient:updateMessage("guess_cure")
     end
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -266,6 +266,15 @@ function Hospital:isRoomDiscovered(room_id)
   return self.room_discoveries[room_id].is_discovered
 end
 
+--! Update the 'guess the cure' faxes of waiting patients.
+--!param disease_id Disease to update.
+function Hospital:updateGuessCureFaxes(disease_id)
+  for _, patient in ipairs(self.hospital.patients) do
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/hospital.lua
```diff
@@ -266,6 +266,16 @@ function Hospital:isRoomDiscovered(room_id)
   return self.room_discoveries[room_id].is_discovered
 end
 
+--! Update the 'guess the cure' faxes of waiting patients.
+--!param disease_id Disease to update.
+function Hospital:updateGuessCureFaxes(disease_id)
+  for _, patient in ipairs(self.patients) do
+    if patient.disease.id == disease_id then
+      patient:updateMessage("guess_cure")
```
lewri, line 274: You might need to check what happens if the fax is open that needs updating (e.g. all research cheat, or edit while paused).
I would guess currently the fax fails to update on ui until it's reopened

lewri: Refactor, probably means it of scope but you'd hope this pr doesn't make it worse.

Alberth289346: I assumed the existing code for "research" was working, the "guess-cure" uses the same code for the same window to update. So if it doesn't, it likely never worked in the first place.

To be honest, I have no idea how to fix it if that's the case, but yeah definitely here is not the right spot or time.

lewri: On reflection, I think the function name might be wrong, or maybe just needs tweaking.

If we discover a disease (as you use it `ResearchDepartment:discoverDisease`), I think you are specifically using this function for when the doctor didn't know what disease the patient had. So you're updating the guess at cure fax because you discovered what the disease was?

It could be a nag though. If you think the name is fine we'll stick with the name.

Alberth289346: I use function names that express what the function is doing. Context of when the function is used isn't relevant to the function (that exists outside the function only).

Alberth289346:
The big tables to construct faxes are gone. Don't know what happened but it's not my fault (maybe) :)  
As such, PR can be closed I think.

lewri:
Might've been toby moving stuff to local hospital. Sure

Issue #528, opened 2014-12-24
Can't open a save with a window open (saved with demo) with full game files


TheCycoONE:
_From [maartenp...@hotmail.nl](https://code.google.com/u/103280305005036074093/) on March 05, 2013 14:40:22_

What steps will reproduce the problem? 1.Save a game with a window (like Town map and research-screen) open, when using the full original game files.
2.Use the demo files.
3.Open the save (from step 1.)
4. You'll get an error. What is the expected output? What do you see instead? There should no error occur when doing this, opening that window should be canceled. 

This situation could occur when somebody (using the demo-files) opens a savegame from an another user, who was using the full game files.

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1525_

Alberth289346:
#279  looks related?

TheCycoONE:
Related but not a duplicate I think.

lewri:
Not specifically related but CorsixTH can't handle loading a demo save with full game files. The game just breaks.
In this case we probably need a behaviour to restrict opening saves in both directions.

Alberth289346:
Even though the actual graphics are the same, should we consider full game data files and demo data files as 2 different 'graphics' sets? (As in, "custom graphics" sets sold commercially.)

Or does the data include more than graphics? (eg behavior or difficulty settings, ie anything beyond just pictures)

OpenTTD has the notion "baseset" (as opposed to graphics extensions, eg new sets of trains, buses, industries, etc). "baseset" is the stuff you need by default, and besides several original commercial versions, there are also 2 or 3 fee versions.

PR #2096, opened 2022-01-23
Make full/demo incompatible with eachother

### LINKED ISSUE #279: Demo to full: Cannot access the Research window
Hi everyone,

I'm using the 0.30 version with Ubuntu 14.04 and the videos disabled.

The problem is that I cannot access the Research window even after building the Research Dept. The doctor keeps saying that I need to build it.

The odd thing is that the new conditions keeping being found and I can send patients to the room without a problem, two doctors are working there too.

Please tell me if I can provide any information that would be helpful.

lewri:
*Closes #279, #528* (As per the recommended action, not actually fixing the underlying bugs)

- Due to the potential for unexpected bugs in playing a demo game in the full version and vice-versa, each version of the game is locked off from being played in the other (yes, full game saves were loadable in demo!).


CorsixTH/Lua/languages/english.lua
```diff
@@ -702,6 +701,11 @@ errors = {
   could_not_find_first_campaign_level = "Could not find the first level of this campaign: %s",
   save_to_tmp = "The file at %s could not be used. The game has been saved to %s. Error: %s",
   dialog_empty_queue = "Sorry, a humanoid just had an empty action queue, which means that they didn't know what to do next. Please consult the command window for more detailed information. A dialog with the offending humanoid has been opened. Would you like them to leave the hospital?",
+  compatibility_error = {
+    new_in_old = "Sorry, this save was created with a newer version of CorsixTH and is not compatible. Please update to a more recent version.",
+    demo_in_full = "Sorry, you can't open a demo savegame with the full game files loaded. The update your TH Data folder setting.",
```
lewri, line 706: I've corrected this one now, just not pushed

CorsixTH/Lua/app.lua
```diff
@@ -1531,15 +1535,26 @@ end
 
 --! Function to check the loaded game is compatible with the program
 --!param save_version (num)
+--!param demo_flag (bool)
 --!return true if compatible, otherwise false
-function App:checkCompatibility(save_version)
+function App:checkCompatibility(save_version, demo_flag)
```
Alberth289346, line 1540: `demo_flag` isn't clear wrt what thing it represents, the graphics set or the save game.

Alberth289346:
I think one step bigger would be useful.

Since we're aiming for a free graphics set, we'll have eventually at least 3 sets, and these may not be compatible with each other.

Instead of adding a true/false boolean, add a thing named "graphics-set" or so, which currently thus has 2 possible values.
A string as value is probably the simplest solution.

This would also allow to differentiate between different full releases of the game, if we'd like or need to.

CorsixTH/Lua/app.lua
lewri: Specifically, neither. But as per your general comment I'll look into graphics set

CorsixTH/Lua/app.lua
```diff
@@ -1531,15 +1537,48 @@ end
...
+  end
+
+  UILoadGame:loadError(err)
+  return false
+end
+
+--! Compatibility function to work out the game's graphics set
+--!param state The savegame data
+function App.configureGraphicsSetFlag(state)
+  -- First eliminate all but the first campaign level (same in demo and full)
```
lewri, line 1566: We may need an additional comment here that we call the demo "Demo Town" if you load it from Single Scenario menu.

lewri:
Should be ready now, at least for cyco to look at.

CorsixTH/Lua/app.lua
```diff
@@ -1553,10 +1554,32 @@ function App:checkCompatibility(save_version, demo_flag)
...
+
   UILoadGame:loadError(err)
   return false
 end
 
+--! Compatibility function to work out the game's graphics set
+--!param state The savegame data
+function App:configureGraphicsSetFlag(state)
+  -- First eliminate all but the first campaign level (same in demo and full)
+  local self = state
```
Alberth289346, line 1567: I don't get this line. The function parameters are both `self` and `state`, and here you make the original `self` disappear and rename `state` to a new `self`?

Huh?

lewri: You might be looking at an old commit. The line should be now
`function App.configureGraphicsSetFlag(state)`

lewri: Though arguably, there is no speific need to have `local self = state`, but I thought it might be less confusing to read. Potentially wrong.

CorsixTH/Lua/app.lua
```diff
@@ -27,8 +27,9 @@ local runDebugger = corsixth.require("run_debugger")
 
 -- Increment each time a savegame break would occur
 -- and add compatibility code in afterLoad functions
+-- Recommended: Also replace/Update the summary comment
```
TheCycoONE, line 30: :+1:

CorsixTH/Lua/app.lua
```diff
@@ -1531,15 +1537,48 @@ end
...
+  local err
+
+  -- First check the graphics set matches with the game files
+  if (gfx_set == "demo" and not self.using_demo_files) then
+    err = _S.errors.compatibility_error.demo_in_full
+  elseif (gfx_set == "full" and self.using_demo_files) then
+    err = _S.errors.compatibility_error.full_in_demo
+
+  -- if that's all good, check the save and app version
+  elseif app_version >= save_version or self.config.debug then
```
TheCycoONE, line 1553: I don't love that a middle branch is special in this if condition. It's a nit through. Will merge anyway.

CorsixTH/Lua/app.lua
TheCycoONE: I don't see how that's relevant here.

CorsixTH/Lua/app.lua
TheCycoONE: I agree with @Alberth289346, this is very strange. If it was to be self then just make it App:configureGraphicsSetFlags however reading the use, the object passed in is not self, and you shouldn't pretend it is.

CorsixTH/Lua/app.lua
```diff
@@ -1531,15 +1537,48 @@ end
...
+
+--! Compatibility function to work out the game's graphics set
+--!param state The savegame data
+function App.configureGraphicsSetFlag(state)
+  -- First eliminate all but the first campaign level (same in demo and full)
+  local self = state
+  local toxicity = tostring(_S.level_names[1]:upper())
+  local level_name = tostring(self.map.level_name)
+  if level_name ~= toxicity then
+    self.world.gfx_set = "full"
```
TheCycoONE, line 1571: Why so many tests, if this fails it doesn't matter - remove the level name test if baldness is sufficient.

CorsixTH/Lua/persistance.lua
```diff
@@ -266,8 +266,10 @@ function LoadGame(data)
   --local status, res = xpcall(function()
   local objtable = MakePermanentObjectsTable(true)
   local state = assert(persist.load(data, objtable))
+  -- Graphics set types were defined in savegame version 166+.
+  if not state.world.gfx_set then TheApp.configureGraphicsSetFlag(state) end
```
TheCycoONE, line 270: Probably should be split onto 3 lines

TheCycoONE (changes requested):
A few changes, overall looks good.

CorsixTH/Lua/app.lua
lewri: Sadly I don't think checking just baldness is enough.
Other levels in the game use baldness, and that could be falsely marking levels if we don't hone in on a specific level.
Additionally, the demo is called "Demo Town" in the full game and includes baldness.

TheCycoONE: Then is the first check on the level name enough?

lewri: No. The demo level in the demo, and the first level in the main campaign are both called "ToxiCity" (or the name in the respective original languages).

CorsixTH/Lua/app.lua
lewri: You could have

```lua
if demo and full then
  err = "error"
else if full and demo then
  err = "error"
end

if not err then 
  if (appver >= savver or config.debug) then
    return true
  else
    err = "error"
  end
end
```

Though not hot on that either...

PR #2100, opened 2022-02-13
[RDY] Recalculate sound effect boundary on resolution change

mugmuggy:
*Fixes #None*

- If you start with a smaller windowed game and increase the size, the sound effects outside the original centered window size won't play. This just recalls the function to calculate the camera size again


lewri:
Wew, this must've been one of things you found by chance.

lewri (approved):

PR #2101, opened 2022-02-14
[RDY] Fired staff handling for multi_use objects

### LINKED ISSUE #1926: [Bug] Firing staff on multi-use object prevents them leaving
#### Describe the issue
See save in #1924 along with PR #1925 
If you fire a member of staff at the start of treating a patient with a multi-use object (e.g. Inflator) they will fail to leave after treatment finishes.

#### Steps to Reproduce
1. Have a patient heading to inflator.

2. As soon as patient's head is popped, fire the doctor.

3. Watch doctor stay in room with exit sign.

#### Expected Behavior
Staff should leave when fired.

#### System Information
**CorsixTH Version:** 0.65-beta 1 (present in beta 2 and 0.64 too)

**Operating System:** Windows 10 (and arch linux, cyco's build)

**Theme Hospital Version:** GOG (and CD on cyco's build)

#### Gamelog.txt

```
Paste gamelog.txt output here
```

### Additional Info
illustrate the problem.*

mugmuggy:
*Fixes #1926*

- As discussed in the issue, puts the workaround dealing with the action queue changed when multi use object ends

### DIFF
--- a/CorsixTH/Lua/rooms/fracture_clinic.lua
+++ b/CorsixTH/Lua/rooms/fracture_clinic.lua
@@ -65,6 +65,7 @@ function FractureRoom:commandEnteringPatient(patient)
     patient:setLayer(2, 0) -- Remove casts
     patient:setLayer(3, 0)
     patient:setLayer(4, 0)
+    -- if no other actions for staff member meander in room
     if #staff.action_queue == 1 then
       staff:setNextAction(MeanderAction())
     else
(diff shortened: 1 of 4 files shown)


Alberth289346:
Imo it would help if you not only add the solution, but also what problem it fixes. Action code is incredibly dense, I think we want less magic rather than more magic.

lewri:
NB: To note, operating table is multi-use object but a different behaviour (as discussed)

TheCycoONE:
Short comment satisfy your concern @Alberth289346?

Alberth289346:
Any comment would be a step forward.

The point is these things weren't found in the last X years, so they are less than obvious. Just saying firing a staff is a concern that must be addressed is already helpful.

mugmuggy:
@Alberth289346 that sort of comment? Its not just fired, but basically if they have been tasked with nothing else to do.

Alberth289346:
Ok, so the PR title isn't wide enough.

Don't think there is much room for further improvement then, is there?
I am fine with this.

lewri (approved):
Have tested, seems to be working as intended now.
Note on the PR name, but it shouldn't be a blocking factor imo.

Issue #1380, opened 2018-04-09
Music causes crash at end of track, Level 5+


andrewspode:
After level 5, I was getting periodic crashes. I noticed it was happening when a background music track stopped playing, even when paused.

Turning off the Jukebox, or using MP3 files instead solves the problem.

CorsixTH version 0.61, Windows 10 64-bit, sound output via HDMI.

I am using a Theme Hospital ISO and it is otherwise working fine.

_(side note - in the settings it's shown as "MP3's" when it should be "MP3s")_

I can't be certain if it happens at the end of **every** track - but this is the trigger point and it never happens mid tune.

wolfy1339:
What version of Windows 10?

unclespode:
10.0.16299 Build 16299, should be fully up to date.

No other audio issues with the machine. Is there a verbose output mode where I might be able to get any error messages? As far as I can see it's not generating any errors, just closing.

wolfy1339:
If you know how to compile it, you could try making a debug build, and running it through a debugger until it crashes, that will probably give much more information

unclespode:
I'm (rather selfishly) reluctant to spend a lot of time on it when there are two fairly easy "fixes" available (no music, or MP3).

I wanted to post it here, so anyone else with the same problem of frequent crashes might find this - rather than a plea for a fix. I appreciate you both responding.

I've just been running back with the original music on CMD to capture any errors and have yet to reproduce the error, which is frustrating considering its regularity yesterday.

mugmuggy:
I've seen something that might be related to this but haven't actually been able to get a build working again since vcpkg switch. However, for me I'v never been sitting there when its happened. I'd be running a hospital, walk away, come back (sometimes much later) on annual report screen so game is paused and there has been a crash. I believe it was audio related from stack trace I pulled once, and still haven't got around to the debug build since.

MarkL1961:
@TheCycoONE   Hi I think I have found the old issue from a couple of years back.  At that time I was using mp3s for the music so the screen shots there may give a clue of what the windows error was.
https://github.com/CorsixTH/CorsixTH/issues/504
I have since stopped (due to a change of computer and losing the copies I had) using the mp3s and I have not had similar errors since - may or may not be relevant

MarkL1961:
scrub that, the screenshots linked to googlecode

lewri:
Assuming resolved by #2108 but can reopen if not

PR #2108, opened 2022-02-24
[RDY] Fix buffer overflow of played_sound_callbacks_ids

mugmuggy:
*Fixes #* 

- Fix buffer overflow 
- Potentially addresses #1380, was also some older bugs (still bugs?) where lava hole spawns but not accessible by Grim, which would also accelerate the overflow as the lava hole stays spawned.

### DIFF
--- a/CorsixTH/Src/th_lua_sound.cpp
+++ b/CorsixTH/Src/th_lua_sound.cpp
@@ -20,6 +20,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
+#include <array>
 #include <cctype>
 #include <cstring>
 #include <map>
@@ -31,7 +32,7 @@ SOFTWARE.
 
 namespace {
 
-int played_sound_callback_ids[1000];
+std::array<int, 1000> played_sound_callback_ids;
 int played_sound_callback_index = 0;
 std::map<int, SDL_TimerID> map_sound_timers;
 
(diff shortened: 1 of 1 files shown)


lewri:
Have you been able to test any of the possible affected bugs? I don't think we've had any sight of them in years so likely they were fixed or very rare.

mugmuggy:
Grim issue was resolved in #1708. There was another also resolved which was resolved I believe, kicking dying patients as they were on the way to lava hole. I've got the comment on #1380 about a crash with audio. I do recall the function which failed and seen that same buffer overflow cause the same access violation that I seen, but if it addresses 1380 or not I don't know. I think we've since silenced (?) announcements whilst paused, which is where I believe my crash was in 1380, and it was a silent crash to desktop on latest builds.

Its easy enough to replicate, spawn a lava hole through debug console, and you can monitor Audio:playSound for the value of
```lua
self.unused_played_callback_id = self.unused_played_callback_id + 1
```
This value will exceed 1000, get to somewhere else and then overwrites sound_player eventually. The do it twice thing from discord also makes getting to 1000+ quicker

lewri:
> I think we've since silenced (?) announcements whilst paused,

Announcements are halted now unless they had priority critical (so cheats usage was made known)

CorsixTH/Src/th_lua_sound.cpp
```diff
@@ -31,7 +31,8 @@ SOFTWARE.
 
 namespace {
 
-int played_sound_callback_ids[1000];
+constexpr int max_callback_ids = 1000;
```
TheCycoONE, line 34: Prefer modernizing to std::array<int, 1000> played_sound_callback_ids;

CorsixTH/Src/th_lua_sound.cpp
```diff
@@ -230,7 +231,7 @@ int l_soundfx_play(lua_State* L) {
     if (!lua_isnil(L, 7))
       iPlayedCallbackDelay = static_cast<int>(luaL_checknumber(L, 7));
 
-    if (played_sound_callback_index == sizeof(played_sound_callback_ids))
+    if (played_sound_callback_index == max_callback_ids)
```
TheCycoONE, line 234: After making this a std::array, len can be used here.

TheCycoONE (changes requested):
We're moving to c++11 and standard container types where possible to prevent overflows. This is a good opportunity to do that instead.

TheCycoONE (approved):

PR #2115, opened 2022-03-12
Update German translation

### LINKED ISSUE #2054: [Bug] German language files is 3 years old
#### Describe the issue

The german language files is over 3 years old (compared to the english one). I believe there might be some differences (e.g. the difference is right now approx. 200 lines)
Do you have anyone working on it?
Or can/should I help you to update the german one?

Bests
Alex

Manuel-K:
Fixes #2054 

Currently a draft; it contains some comments that have to be removed.

### DIFF
--- a/CorsixTH/Lua/languages/german.lua
+++ b/CorsixTH/Lua/languages/german.lua
@@ -149,12 +149,12 @@ casebook = {
 }
 
 -- only drug_improved_1 is in OVERRIDE section of english.lua
--- ..._1 is used to prevent 'the the squits'
--- FIXME check how this behaves in the translated version
+-- ..._1 is used to prevent 'the the squits' AKA orig [4,30], which is simply "Durchfall"
+-- so the sane thing is to duplicate drug_improved
 adviser = {
   research = {
     drug_improved = "Die Effektivität Ihres %s-Medikaments wurde von Ihrer Forschungsabteilung verbessert.",
-    drug_improved_1 = "Das Medikament %s wurde von Ihrer Forschungsabteilung verbessert.",
+    drug_improved_1 = "Die Effektivität Ihres %s-Medikaments wurde von Ihrer Forschungsabteilung verbessert.",
   },
 }
 
--- a/CorsixTH/Lua/languages/german.lua
+++ b/CorsixTH/Lua/languages/german.lua
@@ -34,7 +34,7 @@ fax.vip_visit_query.choices.invite = "Lassen Sie dem V.I.P. eine offizielle Einl
 
 fax.vip_visit_query.choices.refuse = "Speisen Sie den V.I.P. mit einer Entschuldigung ab." -- text was missing
 
-fax.vip_visit_result.remarks.super[1] = "Was für ein Spitzen-Krankenhaus! Wenn ich das nächste Mal ernsthaft krank bin, will ich hier hin!" -- storming should not be translated with 'turbulent' here
+--fax.vip_visit_result.remarks.super[1] = "Was für ein Spitzen-Krankenhaus! Wenn ich das nächste Mal ernsthaft krank bin, will ich hier hin!" -- storming should not be translated with 'turbulent' here  --FIXME is supposedly unused
 
 fax.vip_visit_result.vip_remarked_name = "%s hat Ihr Krankenhaus besucht."
 fax.vip_visit_result.cash_grant = "Ihnen wurde ein Geldbonus von %d gewährt."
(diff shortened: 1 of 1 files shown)


lewri:
Nice work.
Of course, if you think a comment may be helpful to any future people looking at the German translation it could stay in (especially sometimes where I believe context is needed in)

Manuel-K:
There are still some things that I'm not happy about and it contains a lot of compromises to keep text short enough to fit the UI.

Manuel-K:
Is there a test savegame for the `drug_improved_1`/the the squids situation?

lewri:
I'm not sure how it works. I imagine we blanketed the fix on all languages.... Even though it's likely the problem doesn't exist in other languages.

Manuel-K:
It's simply Durchfall in the German version, so I've made `drug_improved_1 == drug_improved`.

Manuel-K:
So this might be complete, unless you introduce a new tooltip for #2121

CorsixTH/Lua/languages/german.lua
```diff
@@ -124,13 +122,52 @@ tooltip.staff_window.center_view = "Linksklick, um zur Person zu springen, Recht
...
+  diag_procedure = "diagnoseverfahren",
+}
+
+-- remove hyphen since it makes it use the wrong font. again, the case cannot be fixed
+casebook = {
+  deaths = "ablebensfälle", --the ä looks weird as well
+}
+
+-- only drug_improved_1 is in OVERRIDE section of english.lua
+-- ..._1 is used to prevent 'the the squits' AKA orig [4,30], which is simply "Durchfall"
```
lewri, line 152: So we do think this bug was only in English?

CorsixTH/Lua/languages/german.lua
```diff
@@ -286,67 +319,80 @@ adviser = {
 dynamic_info.patient.actions.no_gp_available = "Wartet darauf, dass Sie eine Allgemeinmedizin bauen"
 dynamic_info.staff.actions.heading_for = "Geht zu: %s"
 dynamic_info.staff.actions.fired = "Gefeuert"
+dynamic_info.staff.actions.vaccine = "Impft einen Patienten"
+dynamic_info.patient.actions.epidemic_vaccinated = "Ich bin nicht mehr ansteckend"
```
lewri, line 323: This is out of scope but a comment for a future PR -- these should probably be changed to be nested.

CorsixTH/Lua/languages/german.lua
```diff
@@ -124,13 +122,52 @@ tooltip.staff_window.center_view = "Linksklick, um zur Person zu springen, Recht
...
+-- remove invisible hyphens. unfortunately it is impossible to fix the wrong case, since
+-- the case determines the font
+policy = {
+  diag_termination = "diagnoseerstellung",
+  diag_procedure = "diagnoseverfahren",
+}
+
+-- remove hyphen since it makes it use the wrong font. again, the case cannot be fixed
+casebook = {
+  deaths = "ablebensfälle", --the ä looks weird as well
```
lewri, line 148: Is the `ä` in the game font generally weird; or does it look like a completely different font altogether?
I'd be concerned if the latter happened.

lewri (approved):
I'm fine with this, thank you for documenting the translations that would likely benefit from context (in your opinion).

I will try and fix the tooltip bug you mentioned.

CorsixTH/Lua/languages/german.lua
Manuel-K: The dots are so high that they might be outside of the atmosphere. And the kerning of fä is bad. The policy window has several cases of bad kerning as well: ver f ahren should be one word, verl assen as well.

CorsixTH/Lua/languages/german.lua
Manuel-K: I would have done that, but I tried to match english.lua to facilitate side-by-side diffing, so what wasn't nested there isn't nested here.

CorsixTH/Lua/languages/german.lua
Manuel-K: It seems so, although there might be other languages that have unfortunate disease names. The thing is – it seems like most translators didn't know that the second string is a specific workaround. I've looked at several other files and they had different strings.

Manuel-K:
I think this is a good as it will get – unless there are any strings that are expected to be introduced in the very near future.

Manuel-K:
Maybe squashing the commits would be a good idea to get rid of all the FIXME notices.

TheCycoONE:
Can you report the kerning bug. I have some idea what might cause it.

PR #2117, opened 2022-03-20
[WIP] Fix dynamic info updates during epidemics

### LINKED ISSUE #1173: Dynamic information showing incorrect information

mugmuggy:
*Fixes #1173*

- During epidemics patient info don't update correctly as per #1173
- Prevent patients becoming fed up and leaving when in a room, also mentioned in the same issue


Alberth289346:
Sorry, this should not have been merged !!

What is the best course of action now?

PR #2119, opened 2022-03-21
Revert "[WIP] Fix dynamic info updates during epidemics"

lewri:
Reverts CorsixTH/CorsixTH#2117


lewri:
I can only do limited things sadly, this is just github's option. If git offers you a way of doing it in a different manner, maybe go with that.

Alberth289346:
Your suggestion is a valid way out, and the idea had crossed my mind already before you made the PR. I just don't like doing backward steps so I avoid them.

Before I pressed "merge" I did read the patch, and its content looks sane to me. I don't know why it's still a draft, but I think it's a good step forward at least. So another way out is to keep what I added and go from there. I think (but not sure) what is needed is resolving the reason to keep the patch draft. I don't know what needs to be done for that though.

lewri:
Let's wait for @TheCycoONE then

TheCycoONE:
Yeah. Most of the pain will be on mugmuggy I'm afraid. He'll need to rebase his changes on master without losing them.

PR #2120, opened 2022-03-25
Fix Negation Errors

lewri:
Update to luacheck now includes warnings for negation.

This addresses those, but needs to be checked that none of these changes break anything.
@colinjmatt could you check that OSX behaves correctly with the change to `world.lua` for the `pathsep` check?

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff/doctor.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff/doctor.lua
@@ -235,7 +235,7 @@ function Doctor:setCrazy(crazy)
   else
     -- make doctor sane
     if self.is_crazy then
-      if not (self.layers[5] < 5) then
+      if (self.layers[5] >= 5) then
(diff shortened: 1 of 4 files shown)


lewri:
Appears to be working fine in Windows

CorsixTH/Lua/entities/humanoids/staff/doctor.lua
```diff
@@ -235,7 +235,7 @@ function Doctor:setCrazy(crazy)
   else
     -- make doctor sane
     if self.is_crazy then
-      if not (self.layers[5] < 5) then
+      if (self.layers[5] >= 5) then
```
Alberth289346, line 238:
```suggestion
      if self.layers[5] >= 5 then
```

Alberth289346:
Likely some other parentheses can be removed as well, but readability doesn't improve much imho.

PR #2125, opened 2022-04-06
Refresh dynamic info at load

### LINKED ISSUE #2081: Dynamic Info (info displayed in the HUD info bar) does not update when loading a game.
When saving a game, dynamic info is also saved. When loading the game, the dynamic info (for a room machine for example) may be displaying the wrong info because the language setting changed, or the "new_machine_extra_info" config file flag setting changed, etc...

Part of the load routine, should instruct any applicable patient, room, machine, etc.. to update it's dynamic text.

[Temp2.zip](https://github.com/CorsixTH/CorsixTH/files/7647450/Temp2.zip)

lewri:
*Fixes #2081*

- *New* Refactors `Patient` to use method `setDynamicInfoText`
- Refreshes all dynamic info at load in case there was a language code.
- Separate issue raised about some strings staying in the old language if the language is changed. #2126


CorsixTH/Lua/entities/machine.lua
```diff
@@ -415,6 +415,7 @@ function Machine:afterLoad(old, new)
       room.hospital:removeHandymanTask(taskIndex, "repairing")
     end
   end
+  self:updateDynamicInfo(true)
```
TheCycoONE, line 418: Ugly that the others take string and machine takes bool for the same method name now no?

TheCycoONE:
Would keep the text functions on humanoids and the calculating (bool functions) on objects with their existing distinct names i think. Otherwise fine.

CorsixTH/Lua/entities/machine.lua
lewri: I think it may be because the machine dynamic info is never passed a string, which is why it's not needed.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -585,19 +585,34 @@ function Staff:increaseWage(amount)
...
+--!param text (string) the string to append
 function Staff:setDynamicInfoText(text)
   self.dynamic_text = text
   self:updateDynamicInfo()
 end
 
-function Staff:updateDynamicInfo()
+--! Updates a staff member's dynamic info
+--!param dynamic_text (string) The string to append
+function Staff:updateDynamicInfo(dynamic_text)
```
lewri, line 598: @TheCycoONE is your request changing the param on this to `action_string`?

lewri, line 598:
```suggestion
function Staff:updateDynamicInfo(action_string)
```

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -585,19 +585,34 @@ function Staff:increaseWage(amount)
...
+  -- Retain the old text if only an update is wanted, i.e. no new string is supplied.
+  if dynamic_text == nil then
+    if self.dynamic_text then
+      dynamic_text = self.dynamic_text
+    else
+      dynamic_text = ""
+    end
+  else
+    self.dynamic_text = dynamic_text
+  end
```
lewri, line 608:
```suggestion
  if action_string == nil then
    if self.dynamic_text then
      action_string = self.dynamic_text
    else
      action_string = ""
    end
  else
    self.dynamic_text = action_string
  end
```

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -585,19 +585,34 @@ function Staff:increaseWage(amount)
...
   local fatigue_text = _S.dynamic_info.staff.tiredness
   if not self.attributes["fatigue"] then
     fatigue_text = nil
   end
   self:setDynamicInfo('text', {
     self.profile.profession,
-    self.dynamic_text and self.dynamic_text or "",
+    dynamic_text,
```
lewri, line 615:
```suggestion
    action_string,
```

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -585,19 +585,34 @@ function Staff:increaseWage(amount)
...
+--! Sets dynamic info text before the dynamic info update
+--!param text (string) the string to append
 function Staff:setDynamicInfoText(text)
   self.dynamic_text = text
   self:updateDynamicInfo()
 end
 
-function Staff:updateDynamicInfo()
+--! Updates a staff member's dynamic info
+--!param dynamic_text (string) The string to append
```
lewri, line 597:
```suggestion
--!param action_string (string) The string to append
```

TheCycoONE:
No, what I was saying is that the method that takes the action text string should be setDynamicText. updateDynamicText should not take a string, at most it should take a bool like the other places with an updateDynamicText method.

You split the method and changed all the callers to updateDynamicText, but you could have left them with setDynamicText, not put the action_text setting logic in the new updateDynamicText (which would have made it cleaner), and called updateDynamicText from setDynamicText.

That would make the change smaller, and result in clearer code I think?

CorsixTH/Lua/humanoid_actions/seek_reception.lua
```diff
@@ -73,7 +73,7 @@ local function action_seek_reception_start(action, humanoid)
     local x = best_desk.tile_x + orientation.use_position[1]
     local y = best_desk.tile_y + orientation.use_position[2]
     local dist = humanoid.world:getPathDistance(humanoid.tile_x, humanoid.tile_y, x, y)
-    humanoid:updateDynamicInfo(_S.dynamic_info.patient.actions.on_my_way_to
+    humanoid:setDynamicInfoText(_S.dynamic_info.patient.actions.on_my_way_to
```
lewri, line 76: This one threw an error after re-testing the game but it only happened once... Haven't been able to replicate

```bash
An error has occurred!
Running: The timer handler.
Lua\humanoid_actions\seek_reception.lua:76: method 'setDynamicInfoText' is not callable (a nil value)
stack traceback:
        Lua\humanoid_actions\seek_reception.lua:76: in field '?'
        Lua\entities\humanoid.lua:476: in method 'startAction'
        Lua\entities\humanoid.lua:577: in local 'timer_function'
        Lua\world.lua:1085: in method 'onTick'
        Lua\app.lua:1125: in function <Lua\app.lua:1122>
        (...tail calls...)
```

lewri: Found the issue. Fixing...

TheCycoONE:
Yes, I like it. Second question - does it work? (did you test it?)

lewri:
Yes. But I'll test it again today for you, as I know humanoid actions can be a pain.

lewri:
To the best of my ability, working fine. Unless you can spot any potential issues @Alberth289346 ?

lewri:
> > To the best of my ability, working fine. Unless you can spot any potential issues @Alberth289346 ?
> 
> Oh, you're looking for more work? :)

Not right now perhaps :) Just anything looming that might break this.

TheCycoONE:
Yes I agree that times_used should be separate and also that it is out of scope

PR #2131, opened 2022-04-12
[RDY] Move faxes to player hospital

tobylane:
- Move faxes relating to emergencies, VIPs and missing rooms to PlayerHospital
- Apply minor consistent style

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1127,16 +1127,8 @@ function Hospital:checkEmergencyOver()
   end
 end
 
--- Creates VIP and sends a FAX to query the user.
 function Hospital:createVip()
-  local vipName =  _S.vip_names[math.random(1,10)]
-  local message = {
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/hospital.lua
```diff
@@ -2440,3 +2380,19 @@ end
...
+
+function Hospital:makeNoDiagnosisRoomFax()
+  -- Nothing to do, override in a derived class.
+end
+
+function Hospital:makeEmergencyStartFax()
+  -- Nothing to do, override in a derived class.
+end
+
+function Hospital:makeEmergencyEndFax()
```
lewri, line 2396: I'm a bit concerned you don't have params here but you do in playerHospital. But it's probably me being sceptical and it's actually fine to do this.

lewri (commented):
Looks fine I think, my comment is probably dismissible.

CorsixTH/Lua/hospital.lua
tobylane: You're right, added to the new parts and the older stubs.

PR #2133, opened 2022-04-20
Change how staff find work

### LINKED ISSUE #1582: Doctors in GPs office don't look for work autonomously
#### Describe the issue
The doctors in GPs office will only answer calls when having dealt with a patient. A call from the call dispatcher can still send them to a room, but if they are idle in a room, they won't answer a call, as they don't meander to look for work themselves within the room.

#### Steps to Reproduce
A contrived example can be replicated by building a single GPs office and one diagnosis room and having one doctor. After a patient is dealt with, they may seek out the diagnosis room. The doctor in the GP office will leave the room (findWorkForStaff is called at the end of GPRoom:dealtWithPatient).

Pick up doctor and place back in GPs Office. The patient goes to diagnostic room, call dispatcher has the call in the queue, and Doctor won't answer the call.

#### Expected Behavior
Doctors in GPs Office, should answer calls if available. Replicating the example above in the original TH and the doctor does answer the call.

Something like the below, in the loop callback in GPRoom:doStaffUseCycle, prevents the loop repeating again and allows the doctor to find work.
```lua
  local gp_loop_callback = --[[persistable:gp_loop_callback]] function()
    desk_use_time = desk_use_time - 1
    if desk_use_time == 0 then
      if humanoid:isIdle() then
        self:findWorkForStaff()
        if not humanoid:isIdle() then
          return
        end
      end
```
The downside of the above is that the doctor might become a candidate, and thus leave the room, affecting the queue assignment, call assignment etc (like behaviour mentioned in #1290). Potentially mitigated by penalising the priority (CallsDispatcher.getPriorityForRoom) for a room by taking into account the doctor will leave the room, probably will result in another call for a doctor.

#### System Information
**CorsixTH Version:** replicated on 0.6 and latest trunk

### LINKED ISSUE #2060: [Bug] Free doctors do not see patients waiting in other rooms
#### Describe the issue
I've just downloaded 0.65 version and I can notice some behavior different from previous versions.
I have patients waiting in General Diagnosis and also for Psychiatric, but neither the surgeons leave the Operating Theatre nor the psychiatrist on the GP's office is going to attend the Psychiatric or General Diagnosis rooms.

There are no patients in the row for Operating Theatre nor GP's offices room and Staff Leave Rooms is enabled.

Why don't they leave the Operating Theatre and GP's office to go to rooms with patients waiting?

#### Steps to Reproduce
2. Check conditions described above

#### Save Game
[21-STAGE3.zip](https://github.com/CorsixTH/CorsixTH/files/7230240/21-STAGE3.zip)

#### Expected Behavior
Doctors should leave rooms when there is no patients in the row of that room but there are patients in others.

#### System Information
**CorsixTH Version:** 0.65

**Operating System:** Windows 10 21H1 Version (Compilation 19043.1237)

mugmuggy:
This is a proposal to address these issues. The fix has had some funky behaviour in scenarios where a Dr calls to the room, patient leaves for toilet, before call is completed, then the staff member leaves the room and heads to the next room, where the same thing occurs and ended up returning to the same first room.

There is also a change to prefer wandering staff but not sure how effective that is at the moment to get them, due to the weighting of fatigue beforehand. For when excess staff are employed and they leave staff room after rest and there is no call, they start wandering, but they'll have near 0 fatigue and all else being equal is a distance difference of 10 tiles to have the staff leave the room instead.

*Fixes #1582 
Fixes #2060*

- Penalise finding work if in a room over a doctor that is wandering
- Move the find work call outside of the dealtWithPatient process

### DIFF
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -518,6 +518,12 @@ function PlayerHospital:onEndDay()
     end
   end
 
+  for _, staff in ipairs(self.staff) do
+    if staff:isIdle() then
+      self.world.dispatcher:answerCall(staff)
+    end
+  end
+
   Hospital.onEndDay(self)
 end
 
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -216,20 +216,6 @@ function Room:dealtWithPatient(patient)
   if self.dealt_patient_callback then
     self.dealt_patient_callback(self.waiting_staff_member)
   end
-  -- The staff member(s) might be needed somewhere else.
-  self:findWorkForStaff()
-end
-
---! Checks if the room still needs the staff in it and otherwise
--- sends them away if they're needed somewhere else.
-function Room:findWorkForStaff()
-  -- If the staff member is idle we can send him/her somewhere else
-  for humanoid in pairs(self.humanoids) do
-    -- Don't check handymen
-    if class.is(humanoid, Staff) and humanoid.humanoid_class ~= "Handyman" and humanoid:isIdle() then
-      self.world.dispatcher:answerCall(humanoid)
-    end
-  end
 end
 
 local profile_attributes = {
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -722,6 +722,20 @@ function Room:roomFinished()
   self:calculateHappinessFactor()
 end
 
+--! Check if a room is actively required by patients
+--! A room is deemed in demand should any patient be in a state of using, or
+--! wanting to use the room.
+--!return (boolean) true if the room is currently needed
+function Room:isRoomInDemand()
+  local door = self.door
+  if self:getPatientCount() > 0 and not self:getPatient():isLeaving() then
+    return true
+  end
+  return (door.queue and door.queue:patientSize() > 0) or
+      (door.reserved_for and class.is(door.reserved_for, Patient)) or
+      (door.user and class.is(door.user, Patient) and door.user:getCurrentAction().is_entering)
+end
+
 --! Try to move a patient from the old room to the new room.
 --!param old_room (Room) Room that currently has the patient in the queue.
 --!param new_room (Room) Room that wants the patient in the queue.
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -137,7 +137,23 @@ function Room:createEnterAction(humanoid_entering, callback)
   return WalkAction(x, y):setIsEntering(true)
 end
 
---! Get a patient in the room.
+--! Get a table of all patients using the room
+--! A room usually only has one patient, however in the ward can have multiple
+--! patients at once.
+--!return patients (table) All patients actively using the room
+function Room:getPatients()
+  local patients = {}
+  -- Are there patients using the room?
+  if self:getPatientCount() == 0 then return patients end
+  for humanoid in pairs(self.humanoids) do
+    if class.is(humanoid, Patient) then
+      patients[#patients + 1] = humanoid
+    end
+  end
+  return patients
+end
+
+--! Get any patient in the room.
 --!return A patient (humanoid) if there is a patient, nil otherwise.
 function Room:getPatient()
   for humanoid in pairs(self.humanoids) do
@@ -722,18 +738,28 @@ function Room:roomFinished()
   self:calculateHappinessFactor()
 end
 
+--! Private function to check for any patients outside the door queueing,
+--! or about to come in
+--!return (boolean) true if any patient wants to come in
+function Room:_arePatientsWantingToEnter()
+  local door = self.door
+  return (door.queue and door.queue:patientSize() > 0) or
+      (door.reserved_for and class.is(door.reserved_for, Patient)) or
+      (door.user and class.is(door.user, Patient) and door.user:getCurrentAction().is_entering)
+end
+
 --! Check if a room is actively required by patients
 --! A room is deemed in demand should any patient be in a state of using, or
 --! wanting to use the room.
 --!return (boolean) true if the room is currently needed
 function Room:isRoomInDemand()
-  local door = self.door
-  if self:getPatientCount() > 0 and not self:getPatient():isLeaving() then
-    return true
+  if self:getPatientCount() > 0 then
+    -- Maybe a patient is using the room right now?
+    for _, patient in ipairs(self:getPatients()) do
+      if not patient:isLeaving() then return true end
+    end
   end
-  return (door.queue and door.queue:patientSize() > 0) or
-      (door.reserved_for and class.is(door.reserved_for, Patient)) or
-      (door.user and class.is(door.user, Patient) and door.user:getCurrentAction().is_entering)
+  return self:_arePatientsWantingToEnter()
 end
 
 --! Try to move a patient from the old room to the new room.
--- a/CorsixTH/Lua/rooms/gp.lua
+++ b/CorsixTH/Lua/rooms/gp.lua
@@ -161,8 +161,6 @@ function GPRoom:dealtWithPatient(patient)
   if self.staff_member then
     self:setStaffMembersAttribute("dealing_with_patient", false)
   end
-  -- Maybe the staff member can go somewhere else
-  self:findWorkForStaff()
 end
 
 function GPRoom:sendPatientToNextDiagnosisRoom(patient)
--- a/CorsixTH/Lua/calls_dispatcher.lua
+++ b/CorsixTH/Lua/calls_dispatcher.lua
@@ -514,6 +514,11 @@ function CallsDispatcher.getPriorityForRoom(room, attribute, staff)
   -- Prefer the tirer staff (such that less chance to have "resting synchronization issue")
   score = score - staff:getAttribute("fatigue") * 40 -- 40 is just a weighting scale
 
+  -- Prefer a wandering staff member over a staff member in a room
+  if not staff:getRoom() then
+    score = score - 50
+  end
+
   -- TODO: Assign doctor with higher ability
 
   -- Room requires specilitist trumps over normal rooms
--- a/CorsixTH/Lua/calls_dispatcher.lua
+++ b/CorsixTH/Lua/calls_dispatcher.lua
@@ -327,8 +327,8 @@ end
 function CallsDispatcher:answerCall(staff)
   local min_score = 2^30
   local min_call = nil
-  assert(not staff.on_call, "Staff should be idea before he can answer another call")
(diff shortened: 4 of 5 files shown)


CorsixTH/Lua/room.lua
```diff
@@ -202,20 +202,6 @@ function Room:dealtWithPatient(patient)
...
--- sends them away if they're needed somewhere else.
-function Room:findWorkForStaff()
-  -- If the staff member is idle we can send him/her somewhere else
-  for humanoid in pairs(self.humanoids) do
-    -- Don't check handymen
-    if class.is(humanoid, Staff) and humanoid.humanoid_class ~= "Handyman" and humanoid:isIdle() then
-      self.world.dispatcher:answerCall(humanoid)
-    end
-  end
 end
```
lewri, line 205: I still like he idea of having an immediate find work at the end of the consult. Is there a specific consequence if we left this in with the new code?

I suppose the worst of it is that the staff member takes a day to find new work.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -481,6 +481,12 @@ function PlayerHospital:onEndDay()
     end
   end
 
+  for _, staff in ipairs(self.staff) do
+    if staff:isIdle() then
+      self.world.dispatcher:answerCall(staff)
+    end
+  end
+
```
lewri, line 489: This looks like it is preventing handymen from responding to calls -- I'm guessing they're never considered idle.

I have checked with latest master and this bug isn't present there.

lewri (changes requested):
Handyman issue needs addressing

CorsixTH/Lua/room.lua
lewri: On reflection I think I prefer the balance you have chosen, so happy to discuss my above comment.

mugmuggy: I did toss up leaving it in as well, hence leaving it more as a proposal. The original has decisions and logic processed on part day ticks. I assume so some logic doesn't run every tick but also some more frequently than once per day. This probably sits in one of those in the original.

lewri: Actually doesn't this function only get called from GP Office?

lewri:
I would like to see this one make the next release, as it does offer quite a marked improvement to finding work.

CorsixTH/Lua/room.lua
lewri: As per above. Only GP Office used this function. Maybe that was part of the problem too.

lewri:
> The fix has had some funky behaviour in scenarios where a Dr calls to the room, patient leaves for toilet, before call is completed, then the staff member leaves the room and heads to the next room, where the same thing occurs and ended up returning to the same first room.

Is this significant to occur; or would it have a significant impact on gameplay if it happened?
Also any blocking factor regarding RFC? I'm happy with this as is.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -510,6 +510,12 @@ function PlayerHospital:onEndDay()
     end
   end
 
+  for _, staff in ipairs(self.staff) do
+    if staff:isIdle() then
+      self.world.dispatcher:answerCall(staff)
+    end
+  end
```
lewri, line 517: On reinspecting some code, we do use idle a fair bit for some actions as a wait (I found vaccinate and multi-use object). Would these likely get misinterpreted by this code?

Alberth289346: Just judging by the question, it seems to me that `isIdle` is ambiguous. So one step is to decide yes or no on your question and then document it with `isIdle`, possibly also improving the name of the function to better match its use.

Also, you may want to split its use between 2 or more functions with different names, so it becomes easier to get the right function. To assist the programmer, it's likely a good idea to mention the related functions in lua doc and/or put them close together, to improve the chance the right function gets found.

As for the functionality, in the calls dispatcher you added `not staff:getRoom()` as condition, which apparently means "a wandering staff member". While I'd prefer an explicit function for that notion, eg `hasNoWork` or so, that check may possibly be good here?

Alberth289346: If you add a new function for the latter, please also add a link to that in the lua doc of `getRoom` to avoid further spreading of the `getRoom` function for this purpose.

Another measure that can be done is to make the `getRoom` private, although that has much larger consequences (it should be called with `self` only), not sure if that is feasible.

lewri: `isIdle` is actually quite a hefty function to determine a true idle state now that I got round to reading it. It is definitely much more than the mere `IdleAction` I thought it was. It's documented inside nicely enough in the function, not for doc though.

lewri (approved):
@mugmuggy I finally got round to testing this. I'm happy enough with this as is to suggest merge. If you're able to sanity check the patch still looks compatible let me know and will schedule to merge :)

NB: A different issue but later on, these calls should probably be able to finish use on a desk early (e.g. Dr and Nurse will continue through a use desk action even though they picked up a call).

ARGAMX:
I'd like to merge this.

I found another issue related to this problem - "Wandering Doctor Not Answering Calls".
But this PR resolves the problem.

CorsixTH/Lua/room.lua
lewri: Actually those comments were wrong. All rooms called this.

lewri:
I'm tempted to keep the existing workfinding function for rooms ontop of this new function. I can't remember now if in TH staff responded immediately to calls after a patient was dealt with; or if there was a wait cycle.

lewri:
There is another glitch in this change. If the patient is about to enter (pops from queue) and the day changes, the staff member will look for work and answer a new call.

lewri:
Changes:
- Refactored two early returns in `staff:isIdle`
- Refactored the room checks for `staff:isIdle` into its own function in `Room:isRoomInDemand`
- Adds additional checks to said new function to ascertain if a room is staff member is truly no longer required in room

CorsixTH/Lua/room.lua
```diff
@@ -722,6 +722,20 @@ function Room:roomFinished()
   self:calculateHappinessFactor()
 end
 
+--! Check if a room is actively required by patients
+--! A room is deemed in demand should any patient be in a state of using, or
+--! wanting to use the room.
+--!return (boolean) true if the room is currently needed
+function Room:isRoomInDemand()
+  local door = self.door
+  if self:getPatientCount() > 0 and not self:getPatient():isLeaving() then
```
TheCycoONE, line 731: I think this is wrong for the ward if patient-centered > 1.

lewri: Good point. Likely this was broken before. Will also check the changes that let the nurse leave ward before fixing but I don't think they included being able to find new work.

Alberth289346: I closed this by accident

lewri:
Added handling for ward. I think this is done now.

lewri:
Okay, ready.

CorsixTH/Lua/rooms/gp.lua
```diff
@@ -161,8 +161,6 @@ function GPRoom:dealtWithPatient(patient)
   if self.staff_member then
     self:setStaffMembersAttribute("dealing_with_patient", false)
   end
-  -- Maybe the staff member can go somewhere else
-  self:findWorkForStaff()
```
ARGAMX, line 165: This was annoying call. 
I like we are removing it from here. 👍 
Consultants are constantly leaving out of GPs because of this.
This is a significant change making me want this PR. 👍

CorsixTH/Lua/room.lua
```diff
@@ -736,6 +738,24 @@ function Room:roomFinished()
...
 
+--! Check if a room is actively required by patients
+--! A room is deemed in demand should any patient be in a state of using, or
+--! wanting to use the room.
+--!return (boolean) true if the room is currently needed
+function Room:isRoomInDemand()
+  local door = self.door
+
+  if self:getPatientCount() > 0 then
+    -- All patients must be leaving who are inside the room
```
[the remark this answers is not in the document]
lewri: The comment on this line technically is outlining the process based on the function comment `Check if a room is actively required by patients`. You got me scratching my head over this one, probably because it's clear to me but seemingly not clear enough to someone else (nobody is being asked to leave). 

Your suggestion feels a bit bloated for what the code block it refers to is.

Alternatives:
`If a patient is still using a room, then it is still required`. 
`A room is still needed if a patient is using it`.

Other form of comment alternatives I've seen before in this code
`If a patient is still using the room, no need to check any further`.
`Maybe a patient is using the room right now?` <-- I probably prefer this one

CorsixTH/Lua/room.lua
```diff
@@ -736,6 +738,24 @@ function Room:roomFinished()
...
+    for _, patient in ipairs(self:getPatients()) do
+      if not patient:isLeaving() then return true end
+    end
+  end
+  return (door.queue and door.queue:patientSize() > 0) or
+      (door.reserved_for and class.is(door.reserved_for, Patient)) or
+      (door.user and class.is(door.user, Patient) and door.user:getCurrentAction().is_entering)
```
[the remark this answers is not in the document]
lewri: Can't think of nice function name (that doesn't sound like it was shoehorned) for it right now or where it should live (room? door? queue? just local?). I'll note it and think about it.

CorsixTH/Lua/room.lua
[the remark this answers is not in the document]
ARGAMX: > The comment on this line technically is outlining the process based on the function comment `Check if a room is actively required by patients`.
> 

Ah, now I see it.
But it was not obvious, because function comment didn't say anything about leaving the room.

> Alternatives: `If a patient is using the room, then it is still required`. `A room is still needed if a patient is using it`.
> 
> Other form of comment alternatives I've seen before in this code `If a patient is still using the room, no need to check any further`. `Maybe a patient is using the room right now?` <-- I probably prefer this one

Yeah, the both variants is better than current one. 👍

CorsixTH/Lua/room.lua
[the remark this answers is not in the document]
Alberth289346: `isRoomNeededByPatients` would be my first guess. Since the problem holds for all rooms I am guessing, the `Room` baseclass should be fine.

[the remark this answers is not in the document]
Alberth289346: A good comment at line 753 would work too, imho.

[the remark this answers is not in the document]
lewri: Function privatised, it's likely it would be useful in future in other spots

ARGAMX (approved):
LGTM 

btw 
Force Push sometimes making good things. 

But sometimes I think it brings to lost some context, when you trying to see a diff to commit which is not exist anymore...

Issue #1606, opened 2020-05-15
Doctors in Research Room waiting for imaginary patient


lewri:
#### Describe the issue
In the last two levels I've been playing I've noticed that sometimes doctors in the Research Room are in state 'waiting for patient' despite me not sending anyone for autopsy.
One happened when I dragged a researcher mid-consult from another room (but sometimes will go unnoticed) to the Research Room.

May be a separate issue but can also happen in the operating theatre sometimes (queue becomes empty but doctors get stuck in 'waiting for patient'.

To resolve you just have to pickup the doctor and drop them back in the room. 

#### Steps to Reproduce
1. Load the game file. Go to Research Room

2. Doctor Milwin is in 'waiting for patient' state

To try and reproduce from the start, you could try dragging out a Doctor (Research Skilled) mid-consult or while they have a queue into the Research Room. I don't have a savestate of that one though.

#### Save Game
https://www.dropbox.com/s/u6apj6662ukwbdg/waitingresearcher.sav?dl=0

#### Expected Behavior
Doctors shouldn't be waiting for 'imaginary patients'.

#### System Information
**CorsixTH Version:** 0.63

**Operating System:** Windows 10

**Theme Hospital Version:** GOG

#### Gamelog.txt
When selecting the doctor in state:

-----------------------------------
On call: no
Busy:   busy
Last room:      research@62,65
-----------------------------------
Clicked on:
<LocalisedString> Current value:humanoid (MILWIN) - class: Doctor
Warmth: 0.361   Happiness: 0.954   Fatigue: 0.409  Thirst: 0.000  Toilet_Need: 0.000   Health: 0.930
Actions: [use_object - desk   must_happen, meander]
Skills: (0.680)  Surgeon (0.000)  Psych (0.000)  Researcher (1.000)
-----------------------------------

### Additional Info
![image](https://user-images.githubusercontent.com/20030128/82080060-cf692480-96db-11ea-86bc-22cd8823c652.png)

mugmuggy:
This one is from Room:tryAdvanceQueue. If you have two staff at the front of the queue, say one of the researchers and a handyman leaving the room as an example, causing one to queue, it will trigger 'waiting for patient' on all staff in the room. You could check front is a patient, and maybe need to check they are actually entering to avoid setting it to waiting for patient.

PR #2135, opened 2022-04-21
[RDY] Fix dynamic info updates during epidemics

### LINKED ISSUE #1173: Dynamic information showing incorrect information

mugmuggy:
*Fixes #1173/#1606*

- During epidemics patient info don't update correctly as per #1173
- Prevent patients becoming fed up and leaving when in a room, also mentioned in the same issue
- Added a helper method for setting staff_wait, and wrapped it to prevent some invalid scenarios #1606 - method refers to staff_member and staff_member_set also to only operate on staff, seems a handy shortcut for the class.is Staff and not handyman check.

### DIFF
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -493,19 +493,37 @@ function Room:commandEnteringStaff(humanoid, already_initialized)
   end
 end
 
+
+
+--! Activates and deactivates the staff waiting for patient mood icon
+-- and dynamic info text
+--!param activate (bool) - true to activate, false or nil to deactivate
+function Room:_staffWaitToggle(activate)
+  local state = "deactivate"
+  local dynamic_text = ""
+
+  if activate then
+    dynamic_text = _S.dynamic_info.staff.actions.waiting_for_patient
+    state = "activate"
+  end
+
+  if not self.staff_member_set and self.staff_member then
+    self.staff_member:setMood("staff_wait", state)
+    self.staff_member:setDynamicInfoText(dynamic_text)
+  else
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/room.lua
```diff
@@ -493,19 +493,37 @@ function Room:commandEnteringStaff(humanoid, already_initialized)
...
+    state = "activate"
+  end
+
+  if not self.staff_member_set and self.staff_member then
+    self.staff_member:setMood("staff_wait", state)
+    self.staff_member:setDynamicInfoText(dynamic_text)
+  else
+    for staff_member in pairs(self.staff_member_set) do
+      staff_member:setMood("staff_wait", state)
+      staff_member:setDynamicInfoText(dynamic_text)
```
lewri, line 516: `staff_member:setDynamicInfoText(dynamic_text)` will have a clash with #2125.
Did you want this PR to go first, or to wait for the other PR?

lewri (commented):
LGTM. I tripped up trying to find how you handled handymen but saw it in the end.

CorsixTH/Lua/room.lua
```diff
@@ -493,19 +493,37 @@ function Room:commandEnteringStaff(humanoid, already_initialized)
   end
 end
 
+
+
```
lewri, line 497: Don't think these two newlines are required? But it's not blocking if we left them in imo.

lewri (approved):

PR #2136, opened 2022-04-21
Local hospital advice 2

tobylane:
- Add advice about adding to a full training room
- Remaining advice about the local hospital is made player only

### DIFF
--- a/CorsixTH/Lua/rooms/research.lua
+++ b/CorsixTH/Lua/rooms/research.lua
@@ -131,12 +131,12 @@ function ResearchRoom:roomFinished()
   -- Is this the first research department built?
   if not self.hospital.research_dep_built and not TheApp.using_demo_files then
     self.hospital.research_dep_built = true
-    self.world.ui.adviser:say(_A.information.initial_general_advice
-    .research_now_available)
+    self.hospital:giveAdvice({_A.information.initial_general_advice
+      .research_now_available})
   end
   -- Also check if it would be good to hire a researcher.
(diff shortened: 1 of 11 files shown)


lewri (approved):
I had to read the change to hospital.lua a few times but yes lgtm

CorsixTH/Lua/rooms/research.lua
```diff
@@ -131,12 +131,12 @@ function ResearchRoom:roomFinished()
   -- Is this the first research department built?
   if not self.hospital.research_dep_built and not TheApp.using_demo_files then
     self.hospital.research_dep_built = true
-    self.world.ui.adviser:say(_A.information.initial_general_advice
-    .research_now_available)
+    self.hospital:giveAdvice({ _A.information.initial_general_advice
+    .research_now_available })
```
Alberth289346, line 135: Add 2 indents, it's a continuation of the statement.

lewri:
> I don't like the added spaces inside the curly braces, but our code conventions allows it apparently, it's allowed.

I thought that a little weird too. But I'm not actually fussed by it.

tobylane:
We could change the coding convention?

lewri:
While not stated, the nature of brackets used in coding conventions suggest no spacing e.g.
https://github.com/CorsixTH/CorsixTH/wiki/Coding-Conventions#bracketless-function-calls

Ideally then imo, if we were to stick to the norm the spaces would have to go.

tobylane:
Agreed and done, and squashed the autopsy commit into the general one.

TheCycoONE:
Thanks

PR #2137, opened 2022-04-21
Minor fixes related to luacheck

tobylane:
- Move two helper functions to the one place they're used. 
- Add the staff initials cache to globals list
- Add the two call parent afterloads that are missing, I haven't seen this as a problem in game but the change makes it consistent with the others - https://grep.app/search?q=afterload&filter[repo][0]=CorsixTH/CorsixTH&filter[lang][0]=Lua&filter[path][0]=CorsixTH/Lua/

Each of these could be instead left as is with luacheck ignores

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -28,7 +28,7 @@ globals = { -- Globals
             "lfs", "list_to_set", "loadfile_envcall", "loadstring_envcall",
             "permanent", "print_table", "rangeMapLookup", "rnc",
             "strict_declare_global", "table_length", "unpermanent", "values",
-            "serialize","array_join","shallow_clone",
+            "serialize","array_join","shallow_clone","staff_initials_cache",
 
             -- Game classes
             "AIHospital", "AnimationManager", "AnimationEffect", "App", "Audio",
@@ -228,7 +228,6 @@ add_ignore("CorsixTH/Lua/rooms/ward.lua", "113")
 add_ignore("CorsixTH/Lua/rooms/ward.lua", "211")
 add_ignore("CorsixTH/Lua/rooms/ward.lua", "212")
 add_ignore("CorsixTH/Lua/sprite_viewer.lua", "212")
-add_ignore("CorsixTH/Lua/staff_profile.lua", "113")
 add_ignore("CorsixTH/Lua/strict.lua", "212")
 add_ignore("CorsixTH/Lua/strings.lua", "212")
 add_ignore("CorsixTH/Lua/strings.lua", "122")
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1209,6 +1209,25 @@ local function doorWallOffsetCalculations(x, y, wall)
...
+
+--! Check bit is set
+--!param value (int) value to check set bit of
+--!param bit (int) 0-base index of bit to check
+--!return (boolean) true if bit is set.
+local function hasBit(value, bit)
+  local p = 2 ^ bit
+  return value % (p + p) >= p
+end
+
```
Alberth289346, line 1230: Obviously, lua doesn't have proper bit operators so arguably squeezing 3 bits into an integer isn't very lua-like.

Another solution would be to convert the bit-stuff to an array of 3 booleans (invalid_tile uses bits 1, 2, and 3 apparently).

A second alternative could be to implement the bit functions in c++, but that would still keep the bit-stuff of invalid_tile.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1209,6 +1209,25 @@ local function doorWallOffsetCalculations(x, y, wall)
   return x, y, x_mod, y_mod, wall
 end
 
+-- this is a pseudo bitwise OR operation
```
Alberth289346, line 1212:
```suggestion
--! This is a pseudo bitwise OR operation.
```

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1209,6 +1209,25 @@ local function doorWallOffsetCalculations(x, y, wall)
...
+-- assumes value2 is always a power of 2 (limits carry errors in the addition)
+-- mimics the logic of hasBit with the addition if bit not set
```
Alberth289346, line 1214:
```suggestion
-- It assumes value2 is always a power of 2 (limits carry errors in the addition)
-- and mimics the logic of hasBit with the addition if bit is not set.
```

Alberth289346:
Since the base class `afterLoad` changes were apparently never performed for the derived classes, I wonder if they should be performed now, so their changes are (rather late) applied to the current objects.

tobylane:
> Since the base class `afterLoad` changes were apparently never performed for the derived classes, I wonder if they should be performed now, so their changes are (rather late) applied to the current objects.

Done, the one thing missed is about queues which I doubt would trouble research rooms much but let's be complete.

I could put your comments about bitOr/hasBit in the Programming Ideas wiki page?

lewri:
> I could put your comments about bitOr/hasBit in the Programming Ideas wiki page?

I think best this. Right now may not be the best time to change around bitOr/hasBit

TheCycoONE:
Not a fan of this PR; utility is a better place for the bit functions and I don't think luacheck justifies changing that.

I think running the afterLoad for object back to version 56 is not necessarily safe to do on newer saves either.

lewri:
> Not a fan of this PR; utility is a better place for the bit functions and I don't think luacheck justifies changing that.

I agree with this, however our coding conventions do say otherwise:
https://github.com/CorsixTH/CorsixTH/wiki/Coding-Conventions#scope
> __Scope__
> Variables or functions which do not need to be accessed outside of a given chunk (file) should be declared local in that chunk.``

So we can keep this case as an exception (and keep it in utility); or we alter the wording of scope convention.

TheCycoONE:
I think that's too literal of a reading

PR #2139, opened 2022-04-24
Hospitals store their tile object counts.

Alberth289346:
*Fixes that tile object counts make sense if you have more than one hospital.

- Instead of a world-centric tile object count administration, defer counting to the hospital(s) that own the object(s).
- Hospital now returns its own reception desks.
- Also fixes an edge problem of warning patients of a (very) nearby hospital on object placement.

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -109,6 +109,16 @@ function Hospital:Hospital(world, avail_rooms, name)
     heating_broke = false -- (bool) Whether the heating system is broken down currently.
   }
 
+  -- Number of tile objects of each category in the hospital.
+  self.tile_object_counts = {
+    extinguisher = 0,
+    radiator = 0,
+    plant = 0,
+    reception_desk = 0,
+    bench = 0,
+    general = 0,
+  }
+
   self.num_visitors = 0
   self.num_deaths = 0
   self.num_deaths_this_year = 0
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/hospital.lua
```diff
@@ -1024,11 +1059,10 @@ end
 --! Collect the reception desks in the hospital.
 --!return (list) The reception desks in the hospital.
 function Hospital:findReceptionDesks()
-  -- TODO Breaks in multiplayer mode.
```
lewri, line 1027: 👍

CorsixTH/Lua/hospital.lua
```diff
@@ -1686,9 +1735,11 @@ function Hospital:objectPlaced(entity, id)
     local w, h = self.world.map.th:size()
     for tx = math.max(1, entity.tile_x - notify_distance), math.min(w, entity.tile_x + notify_distance) do
       for ty = math.max(1, entity.tile_y - notify_distance), math.min(h, entity.tile_y + notify_distance) do
-        for _, patient in ipairs(self.world.entity_map:getHumanoidsAtCoordinate(tx, ty)) do
-          if class.is(patient, Patient) then
-            patient:notifyNewObject(id)
+        if self:isInHospital(tx, ty) then -- Only patients of this hospital should be warned.
```
TheCycoONE, line 1738: Probably ok -

PR #2140, opened 2022-04-24
Prevent epidemic infection outside hospital

### LINKED ISSUE #2129: Epidemic patient spawns outside the Hospital.
Got an epidemic, found the one patient, Marked him for nurse cure, and went on my way. Health inspector showed up 30 seconds later, declared I'd failed and fined me. during this time (while the inspector was stood at the desk) an infected patient wanders in and heads to reception.

I assume this is an extremely fringe case, as reloading the month gave me 2 different patients that were contagious, I believe the patient spawning outside the hospital and becoming infectious is what triggered the epidemic fail as nobody infectious left the hospital, cured or not during that time, so this other patient is my only other suspect.

Can't provide saves, as by the time it was over, and I worked out what had happened, the inspector was gone and the patient was no longer contagious.

lewri:
Infection outside the hospital could cause a player to unfairly lose an epidemic by infecting a kicked/fed up/too pricey patient outside the hospital grounds.

For sake of fairness as discussed, they should remain able to be infected while in the grounds.
*May fix #2129*

- Adds a check on the infected patient and the target patient to make sure both are inside hospital grounds to pass the infection.

### DIFF
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -144,6 +144,13 @@ function Epidemic:infectOtherPatients()
     -- Patient is not infectious.
     if patient.cured or patient.vaccinated then return false end
 
+    -- Don't allow infection outside the hospital grounds
+    -- Also check both patients to prevent infecting through outer walls
+    local ppx, ppy = patient.tile_x, patient.tile_y
+    if ppx and ppy and not self.hospital:isInHospital(ppx, ppy) then return false end
+    local opx, opy = other.tile_x, other.tile_y
+    if opx and opy and not self.hospital:isInHospital(opx, opy) then return false end
+
     -- 'other' is already infected or is going home.
     if other.infected or other.cured or other.vaccinated then return false end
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/epidemic.lua
```diff
@@ -144,6 +144,13 @@ function Epidemic:infectOtherPatients()
     -- Patient is not infectious.
     if patient.cured or patient.vaccinated then return false end
 
+    -- Don't allow infection outside the hospital grounds
+    -- Also check both patients to prevent infecting through outer walls
+    local ppx, ppy = patient.tile_x, patient.tile_y
+    if ppx and ppy and not self.hospital:isInHospital(ppx, ppy) then return false end
+    local opx, opy = other.tile_x, patient.tile_y
```
tobylane, line 151:
```` suggestion
    local opx, opy = other.tile_x, other.tile_y
````

lewri: Thanks, spotted that at the same time :D

TheCycoONE:
Seems reasonable

PR #2149, opened 2022-04-30
Luacheck for unused args

tobylane:
- Remove unused arguments, except for where they are there to match parent/child classes
- Inline ignore the deliberate cases with a note why
- Reorder the Hospital stubs into how they're used in PlayerHospital, with a header and luacheck ignore for the rest of the file


TheCycoONE:
Thanks.

PR #2152, opened 2022-05-04
[RDY] Luacheck inlining

tobylane:
- Rewrite a few ifs to avoid empty branches
- Add inline ignores for some types of warnings

These commits can be squashed, they're only this way for some clarity.


.luacheckrc
```diff
@@ -138,96 +138,29 @@ add_ignore("CorsixTH/Lua/rooms/operating_theatre.lua", "wait_for_ready")
 
 -- W111: setting non-standard global variable XYZ
 -- W113: accessing undefined variable XYZ
--- W121: setting read-only global variable XYZ
 -- W122: mutating read-only global variable XYZ
--- W211: unused variable XYZ
 -- W212: unused argument XYZ
--- W231: variable XYZ is never accessed
--- W542: empty if branch
```
lewri, line 146: I'd keep these comment lines incase they end up being needed in the future

CorsixTH/Lua/graphics.lua
```diff
@@ -430,7 +430,7 @@ function Graphics:loadFont(sprite_table, x_sep, y_sep, ...)
 
   local use_bitmap_font = true
   -- Force bitmap font for the moneybar (Font05V)
-  if not sprite_table:isVisible(46) or load_font == "Font05V" then -- uppercase M
+  if not sprite_table:isVisible(46) or load_font == "Font05V" then -- luacheck: ignore 542
```
lewri, line 433: Why delete the other comment?

CorsixTH/Lua/world.lua
```diff
@@ -667,6 +668,7 @@ function World:tickEarthquake()
         -- patient:falling()
       end
     end
+    ]]
```
lewri, line 671: Isn't it usually `]]--`?

CorsixTH/Lua/window.lua
```diff
@@ -1538,9 +1538,6 @@ top-left corner of the window.
 top-left corner of the window.
 ]]
 function Window:hitTest(x, y)
-  if x < 0 or y < 0 or (self.width and x >= self.width) or (self.height and y >= self.height) then
```
lewri, line 1541: This line must've been here for some reason no? As you haven't compensated for its deletion.

lewri (commented):
Nice cleaning

CorsixTH/Lua/world.lua
lewri: Sorry I meant `--]]`

CorsixTH/Lua/graphics.lua
tobylane: The next paragraph has the useful information.

CorsixTH/Lua/window.lua
tobylane: https://github.com/CorsixTH/CorsixTH/commit/cd96f856601aaf84108f1e364517872e98557b81#diff-44af6d1126177f5b59057ef070ae3e6f47ab2aee59de3669d753fa97db28695eL1235

There wasn't a reason given at the time, but maybe.

lewri: Hmm. So it's just an orphaned line? I wonder if that if ever actually succeeds

lewri: It constantly executes (fun!). Get rid of it :D

.luacheckrc
TheCycoONE: That's what luacheck's documentation is for - https://luacheck.readthedocs.io/en/stable/warnings.html

I agree with removal. Maybe link the document.

CorsixTH/Lua/world.lua
tobylane: Yes, adding that now.

CorsixTH/Lua/humanoid_actions/seek_reception.lua
```diff
@@ -46,7 +46,7 @@ local function action_seek_reception_start(action, humanoid)
 
   -- Go through all receptions desks.
   for _, desk in ipairs(humanoid.hospital:findReceptionDesks()) do
-    if (not desk.receptionist and not desk.reserved_for) then
```
TheCycoONE, line 49: This is a good candidate to rewrite

CorsixTH/Lua/world.lua
```diff
@@ -649,6 +649,7 @@ function World:tickEarthquake()
       self.next_earthquake.damage_timer = self.next_earthquake.damage_timer + earthquake_damage_time
     end
 
+    --[[
```
TheCycoONE, line 652: Just remove dead code, don't comment it out (source control is for recovering if desired)

lewri:
Just a thought, but if you're leaning towards luacheck inlining (actually gives the ignore context imo) maybe it would be noted in the .luacheck file that this method should be preferred

CorsixTH/Lua/dialogs/resizables/map_editor.lua
```diff
@@ -1109,9 +1109,6 @@ function UIMapEditor:onMouseDown(button, xpos, ypos)
       repaint = true
     end
 
-  elseif self.cursor.state == "right" then
```
lewri, line 1112: Verify the position of this `elseif` doesn't matter to the subsequent `elseif`

tobylane: I can't confirm it in game (can't trigger this clause if I reinstate it with a print), but yes, as it won't check the next elseif.

lewri: I've reread it again. Was probably late when I did.
The removal of this check is fine

CorsixTH/Lua/dialogs/resizables/map_editor.lua
```diff
@@ -1471,7 +1469,7 @@ function UIMapEditor:onMouseUp(button, x, y)
 
   local repaint = false
   if self.cursor.state == "disabled" then
-    -- Don't care about buttons.
+    return repaint -- Don't care about buttons, return early.
```
TheCycoONE, line 1472: Why add the return early comment? It's obvious from the code.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -1027,7 +1027,7 @@ function Patient:updateMessage(choice)
         -- no requirements missing
         enabled = false
       end
-    else -- if choice == "guess_cure" then
+    --else -- if choice == "guess_cure" then
```
TheCycoONE, line 1030: drop the -- after else if you're putting it before else. Also space.

We don't need both this and the TODO, so maybe drop this entirely and change the TODO to `TODO: implement choice == "guess_cure"`

PR #2153, opened 2022-05-04
Only remove fax messages relating to local humanoids

tobylane:
- Only remove faxes (ui) of local patients/staff.


tobylane:
It's a strange one, it belongs with the action and the patient, but as ui it should be put with the player parts, like the creation of these faxes has been.

Alberth289346:
Lewri:
I  saw this change, and I also considered it weird. However, I also know that our actions do all kinds of weird things like creating faxes as Toby said, so I can fully understand where this is coming from.

Since having it hidden in an action is probably less desirable than having it in the hospital, I'd consider the move an improvement. The fact it's weird suggests that it is probably not the final spot for this code.

lewri:
Cool.. All good with me then.

TheCycoONE:
yeah I get it, with all player specific things going through player_hospital vs hospital.

PR #2157, opened 2022-05-07
Update references to luasocket for their v3 release

tobylane:
- Update the url and dates of luasocket in licence.txt
- Use the latest release, rather than dev, of luasocket in macos_luarocks

A connected todo: Update vcpkg. Should both happen at the same time? The windows and macos setups already result in using different commits of luasocket


TheCycoONE:
Thanks. I expect I'll have to update the vcpkg upstream.

TheCycoONE:
https://github.com/microsoft/vcpkg/pull/24606

Issue #2011, opened 2021-07-13
[Bug] update check never sees a new version


TheCycoONE:
#### Describe the issue
The update check does not find new versions

#### Steps to Reproduce
1. Launch CorsixTH 0.64 (or any other version since 0.60)

2. Observe there are no new updates reported

#### Expected Behavior
Update check should notify the user of 0.65.

#### System Information
**CorsixTH Version:** *0.64*

**Operating System:** *Windows 10*

**Theme Hospital Version:** *Any*

### Additional Info
getVersion since 0.60 has returned a string like 'v0.60' matching the tag name. The update check takes 3 fields and concatenates them into a string <major>.<minor><revision> and then does a string comparison. This string is always less than the v<version> returned by getVersion so it is not considered an update.

Note: When this is fixed, we should also plan to support 3 part versions like 0.65.1.

tobylane:
In App:getVersion there's a letter v in the version _number_, which prevents the check in checkForUpdates from doing a number comparison. Is it better to take the v out in getVersion, or have it stripped in checkForUpdates?

0.65.1 isn't a number, so we would need a different comparison such as semantical versioning.

TheCycoONE:
Yes, what i was picturing was adding a version table with it's own metamethods for comparison and string conversion.

lewri:
Should this be moved to `P4 Low` now as an interim solution is now present

TheCycoONE:
Sure

PR #2158, opened 2022-05-07
Fix version compare on update

lewri:
- Current check for update was comparing vx.xx to x.xx, which always resolved that CorsixTH was up to date.
- Bonus points: 0.66 > 0.65.1 (even though our web site only reports 0.65?)

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1678,7 +1678,7 @@ function App:checkForUpdates()
   -- Default language to use for the changelog if no localised version is available
   local default_language = "en"
   local update_url = 'https://corsixth.com/CorsixTH/check-for-updates'
-  local current_version = self:getVersion()
+  local current_version = string.gsub(self:getVersion(), "v", "") -- drop the 'v'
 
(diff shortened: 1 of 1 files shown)


TheCycoONE:
Finally, but is it enough - it does a lexicographical comparison. I guess it's fine until we get to 0.100, or 10.

TheCycoONE:
I reported this as https://github.com/CorsixTH/CorsixTH/issues/2011

lewri:
> Finally, but is it enough - it does a lexicographical comparison. I guess it's fine until we get to 0.100, or 10.

Potentially. Need to check.
+1 about the semantics in that issue.

lewri:
> > Finally, but is it enough - it does a lexicographical comparison. I guess it's fine until we get to 0.100, or 10.
> 
> Potentially. Need to check. +1 about the semantics in that issue.

```
Checking for CorsixTH updates...
current version: 0.99 new version: 0.100
You are running the latest version of CorsixTH.
```

It breaks. So that needs accounting for (or we never plan to reach 0.100?)

TheCycoONE:
Not letting best be the enemy of better here, I'm merging this.

PR #2159, opened 2022-05-08
POC - Update with semantics

### LINKED ISSUE #2011: [Bug] update check never sees a new version

lewri:
*Fixes #2011*

- "A solution" to update method so that the revision param is used properly in future version.
- Changes all versions to x.yy.z format
- Would require a change on the `check-for-updates` URL side for this to work.

I believe cyco actually wanted a metatable solution, which I don't know how to do. So this is just to prove it is possible.
Have tested though, it will work on `0.66.0` > `0.66.1` and `0.66.1` > `0.67.0` for example.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1476,33 +1476,33 @@ function App:getVersion(version)
   if ver > 156 then
     return "Trunk"
   elseif ver > 138 then
-    return "v0.65"
+    return "v0.65.0" -- Some users had v0.65.1 but no version number was specified
   elseif ver > 134 then
-    return "v0.64"
+    return "v0.64.0"
   elseif ver > 127 then
-    return "v0.63"
+    return "v0.63.0"
   elseif ver > 122 then
-    return "v0.62"
+    return "v0.62.0"
   elseif ver > 111 then
-    return "v0.61"
+    return "v0.61.0"
   elseif ver > 105 then
-    return "v0.60"
+    return "v0.60.0"
   elseif ver > 91 then
-    return "0.50"
+    return "v0.50.0"
   elseif ver > 78 then
-    return "0.40"
+    return "v0.40.0"
   elseif ver > 72 then
-    return "0.30"
+    return "v0.30.0"
   elseif ver > 66 then
-    return "0.21"
+    return "v0.21.0"
   elseif ver > 54 then
-    return "0.20"
+    return "v0.20.0"
   elseif ver > 53 then
-    return "0.11"
+    return "v0.11.0"
   elseif ver > 51 then
-    return "0.10"
+    return "v0.10.0"
   elseif ver > 45 then
-    return "0.01"
+    return "v0.01.0"
   else
     return "Beta 8 or earlier"
   end
@@ -1671,6 +1671,11 @@ function App:afterLoad()
 end
 
 function App:checkForUpdates()
+  local function extractVersionNumericals(version)
+    local _, _, major, minor, rev = string.find(version, "(%d+).(%d+).(%d+)") -- luacheck: ignore 211
+    return tonumber(major), tonumber(minor), tonumber(rev)
+  end
+
   -- Only check for updates once per application launch
   if not self.check_for_updates or not self.config.check_for_updates then return end
   self.check_for_updates = false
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/app.lua
```diff
@@ -1671,14 +1671,19 @@ function App:afterLoad()
 end
 
 function App:checkForUpdates()
+  local function extractVersionNumericals(version)
+    local _, _, major, minor, rev = string.find(version, "(%d+).(%d+).(%d+)") -- luacheck: ignore 211
```
lewri, line 1675: Apparently string.find outputs 5 arguments? Lua documentation says we should use `_` for unwanted args, even though luacheck doesn't like it.

Alberth289346: That pattern does according to https://www.lua.org/manual/5.3/manual.html#6.4 

Don't know what luacheck is saying, but perhaps it barks at you trying assign 2 values to the same `_` variable in one assignment? That could be fixed by making the names different, eg `_start, _end, major, minor, rev`

CorsixTH/Lua/app.lua
```diff
@@ -1709,9 +1716,10 @@ function App:checkForUpdates()
 
   local update_table = loadstring_envcall(update_body, "@updatechecker"){}
   local changelog = update_table["changelog_" .. default_language]
-  local new_version = update_table["major"] .. '.' .. update_table["minor"] .. update_table["revision"]
+  local new_version = update_table["major"] .. '.' .. update_table["minor"] .. "." .. update_table["revision"]
+  local new_version_major, new_version_minor, new_version_revision = extractVersionNumericals(new_version)
 
-  if (new_version <= current_version) then
+  if (new_version_major <= current_version_major) and (new_version_minor <= current_version_minor) and (new_version_revision <= current_version_revision) then
```
Alberth289346, line 1722: So I have version `3.7.0` while we just released `4.1.0`. Phew, I am up-to-date.

Alberth289346: Hmm, maybe I am not, I am getting very confused between `new` and `current` here.

Never mind, it works as long as I am not running a future version I think.

lewri: Did I make an error somewhere? I really don't like how I wrote this if statement just in order to save space.
Realising we only run this function once per run -- perhaps more clarity over saving space is better

CorsixTH/Lua/app.lua
```diff
@@ -1476,33 +1476,33 @@ function App:getVersion(version)
...
+    return "v0.20.0"
   elseif ver > 53 then
-    return "0.11"
+    return "v0.11.0"
   elseif ver > 51 then
-    return "0.10"
+    return "v0.10.0"
   elseif ver > 45 then
-    return "0.01"
+    return "v0.01.0"
```
Alberth289346, line 1505: Wouldn't it be simpler to have a table with numbers like `{ .... {53, "v0.11.0"}, {51, "v0.10.0"}, {45, "v0.01.0"} {-1, "Beta 8".}` and a loop finding the last item that matches?

lewri: I can only guess the current system was made to failsafe savegame versions from matching an exact number.
So you'd need an error or fallback along with the for loop.

CorsixTH/Lua/app.lua
Alberth289346: I think it is correct. It's just a compact way to compute the result.

What I did was attempting to break it with the `7` being larger than the `1`, but that example is irrelevant since with normal version numbers you then also must have the `3` and the `4`, since I cannot have `4.7.0` if `4.1.0` is the newest release.
The `3` versus `4` comparison already decides that the version is obsolete and the `7` versus `1` isn't relevant anymore.

CorsixTH/Lua/app.lua
lewri: Actually don't we have `rangeMapLookup` for this case I just described?

Alberth289346: I think the first person started with an `if`, and then everybody added another `elseif` at the end without ever considering alternatives.

TheCycoONE: I don't think we really need it at all...

PR #2166, opened 2022-05-15
Refactor clipping on render_target to keep a stack of clips.

flackr:
**Adds a clip stack on render_target which allows for temporarily adding and removing clips.**
- Since we never need to get the clip anymore, imprecision resulting from the scale factor no longer accumulates.
- Simplifies calling code as it doesn't need to explicitly remember the old clip and restore it.
- Adds a convenience class which automatically restores old clip when it goes out of scope.

### DIFF
--- a/CorsixTH/Lua/dialogs/message.lua
+++ b/CorsixTH/Lua/dialogs/message.lua
@@ -84,10 +84,9 @@ function UIMessage:draw(canvas, x, y)
   if self.on_top then
     Window.draw(self, canvas, x, y)
   else
-    local x_, y_, w, h = canvas:getClip()
-    canvas:setClip(x_, y + self.stop_y, w, self.height, true)
+    canvas:pushClip(0, y + self.stop_y, canvas:getWidth(), self.height, true)
     Window.draw(self, canvas, x, y)
-    canvas:setClip(x_, y_, w, h)
+    canvas:popClip()
   end
 end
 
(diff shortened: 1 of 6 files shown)


flackr:
@TheCycoONE I'm not 100% confident about the lua change as it required adding the new width accessor. What's the best way to test messages?

TheCycoONE:
The clipping that was changed for UIMessage is while the fax icons scroll up before they scroll left on the bottom panel. It happens fairly quickly and is subtle so it can be hard to catch.

Starting from any save game, or start a new hospital and place the just a receptionist, desk, gp office, doctor, open the hospital and hold Z - a message will pop up sooner than later.

TheCycoONE:
The clang complaint https://github.com/CorsixTH/CorsixTH/runs/6442929540?check_suite_focus=true#step:11:84 looks legit; you removed the get_clip_rect that was populating rcOld before copying to rcNew.

flackr:
> The clang complaint https://github.com/CorsixTH/CorsixTH/runs/6442929540?check_suite_focus=true#step:11:84 looks legit; you removed the get_clip_rect that was populating rcOld before copying to rcNew.

Yep, that's a simple fix. Since the clip rects are always intersected we can just prepopulate y = 0, h = height.

flackr:
Verified the UIMessage seems to still be working too :-).

PR #2169, opened 2022-05-20
[RFC] Use win and loss criteria for level progress advice

tobylane:
- Compare world.goals to the current hospital and form a set of relevant advice. Give a random one to the user.

I'm not sure about the halfway to winning/losing messages, they come at some unsuitable times.
I only use original strings, which are somewhat inconsistent. Do we want to add more, eg "Increase your patient cure percentage by 9%"? There's no "You've cured enough patients" without "but the rest of your hospital isn't good enough". Lines 278 and 287 could have the 2 and 3 bumped up 1?
Currently it will print all of the relevant advice, for a nearly winning hospital it looks like this.

> \<LocalisedString> Current value:Okay, your reputation is good enough to win this level. Keep it above 800 and sort out any other problems to finish it.
\<LocalisedString> Current value:You've satisfied the financial criteria for this level. Now keep your balance above 750000, whilst making sure your hospital is running efficiently
\<LocalisedString> Current value:You're getting close to winning. Increase your hospital value by 19270.
\<LocalisedString> Current value:You're about halfway to winning this level.

### DIFF
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -50,7 +50,6 @@ end
 function PlayerHospital:dailyAdviceChecks()
   local current_date = self.world:date()
   local day = current_date:dayOfMonth()
-  local current_month = current_date:monthOfYear()
 
   -- Hold any advice back until the game has somewhat started.
   if current_date < Date(1, 5) then
@@ -69,13 +68,10 @@ function PlayerHospital:dailyAdviceChecks()
         }
         self:giveAdvice(cashlow_advice)
 
-      elseif self.balance < -2000 and current_month > 8 then
-        -- TODO: Ideally this should be linked to the lose criteria for balance.
-        self:giveAdvice({_A.warnings.bankruptcy_imminent})
-
       elseif self.balance > 6000 and self.loan > 0 then
         self:giveAdvice({_A.warnings.pay_back_loan})
       end
+      self:adviseGoalProgress()
     end
   end
 
@@ -214,6 +210,143 @@ function PlayerHospital:dailyAdviceChecks()
   if day == 28 then
     self.adviser_data.temperature_advice = false
   end
+end
+
+--! Give specific advice about progress to the win and lose goals given (of any group)
+--!param goals (table) Goals of the current level, including start values
+function PlayerHospital:adviseGoalSpecifics(goals)
+  local msgs = {}
+  -- Win criteria
+  --  1 Reputation
+  local win1 = goals["reputation"]["win_value"]
+  if win1 and goals["reputation"]["max_min_win"] == 1 then
+    if self.reputation >= win1 and self.reputation > goals["reputation"]["start"] then
+      table.insert(msgs, _A.level_progress.reputation_good_enough:format(win1))
+    elseif win1 > goals["reputation"]["start"] then
+      table.insert(msgs, _A.level_progress.improve_reputation:format(win1 - self.reputation))
+    end
+  end
+  --  2 Balance
+  local win2 = goals["balance"]["win_value"]
+  if win2 and goals["balance"]["max_min_win"] == 1 and self.balance >= win2 and
+      self.balance > goals["balance"]["start"] then
+    table.insert(msgs, _A.level_progress.financial_criteria_met:format(math.floor(win2)))
+  end
+  --  4 Patients cured count
+  local win4 = goals["num_cured"]["win_value"]
+  if win4 and goals["num_cured"]["max_min_win"] == 1 and self.num_cured >= win4 then
+    table.insert(msgs, _A.level_progress.cured_enough_patients)
+  end
+  --  6 Hospital value
+  local win6 = goals["value"]["win_value"]
+  if win6 and goals["value"]["max_min_win"] == 1 then
+    if self.value >= win6 then
+      table.insert(msgs, _A.level_progress.hospital_value_enough :format(math.floor(win6)))
+    elseif self.value > 0.9 * win6 then
+      table.insert(msgs, _A.level_progress.close_to_win_increase_value
+          :format(math.floor(win6 - self.value)))
+    end
+  end
+  local rep_lose_ratio, balance_lose_ratio = 0, 0
+  -- Lose criteria. Ratios 0-1 from start to lose values
+  --  1 Reputation
+  local lose1 = goals["reputation"]["lose_value"]
+  if lose1 and goals["reputation"]["max_min_lose"] == 0 then
+    rep_lose_ratio = (goals["reputation"]["start"] - self.reputation) /
+        (goals["reputation"]["start"] - lose1) or 0
+  end
+  --  2 Balance
+  local lose2 = goals["balance"]["lose_value"]
+  if lose2 and goals["balance"]["max_min_lose"] == 0 then
+    balance_lose_ratio = (goals["balance"]["start"] - self.balance) /
+        (goals["balance"]["start"] - lose2) or 0
+    if balance_lose_ratio > 0.8 then table.insert(msgs, _A.warnings.bankruptcy_imminent) end
+  end
+  --  5 Patients killed percentage
+  local lose5 = goals["percentage_killed"]["lose_value"]
+  if lose5 and goals["percentage_killed"]["max_min_lose"] == 1 and self.percentage_killed > 0.8 * lose5 then
+    table.insert(msgs, _A.level_progress.dont_kill_more_patients)
+    table.insert(msgs, _A.level_progress.nearly_lost)
+  end
+
+  -- Generic negative
+  if rep_lose_ratio > 0.8 or balance_lose_ratio > 0.8 then
+    table.insert(msgs, _A.level_progress.nearly_lost)
+  elseif rep_lose_ratio > 0.6 or balance_lose_ratio > 0.6 then
+    table.insert(msgs, _A.level_progress.three_quarters_lost)
+  elseif rep_lose_ratio > 0.4 or balance_lose_ratio > 0.4 then
+    table.insert(msgs, _A.level_progress.halfway_lost)
+  end
+  return msgs
+end
+
+--! Give advice about general progress to fulfilling all of the goals to win the level
+function PlayerHospital:adviseGoalProgress()
+-- Move to World:World?
+  self.world:determineWinningConditions()
+  local start, goals = self.world.map:getLevelStartState(), self.world.goals
+  if not goals then return end
+
+  local highest_group = 0
+  -- Find highest count group
+  for criteria, v in pairs(goals) do
+    if type(criteria) == "string" then
+      if v["group"] > highest_group then highest_group = v["group"] end
+    end
+  end
+-- End move to World:World?
+-- Combine with World:checkWinningConditions?
+  local criteria_count, win_criteria_met, lose_criteria_met, best_group, best_value = {}, {}, {}
+  -- Collect winning criteria count for each group
+  for g = 1, highest_group do
+    win_criteria_met[g], criteria_count[g] = 0, 0
+    for criteria, t in pairs(goals) do
+      if type(criteria) == "string" and t["group"] == g then
+        -- Win criteria
+        if t["win_value"] then
+          criteria_count[g] = criteria_count[g] + 1
+          if (t["max_min_win"] == 1 and self[criteria] > t["win_value"]) or
+              (t["max_min_win"] == 0 and self[criteria] < t["win_value"]) then
+            win_criteria_met[g] = win_criteria_met[g] + 1
+          end
+        end
+        -- Lose criteria
+        if t["lose_value"] then
+          t["start"] = start[criteria] or 0
+          if (t["max_min_lose"] == 1 and self[criteria] > t["lose_value"]) or
+              (t["max_min_lose"] == 0 and self[criteria] < t["lose_value"]) then
+            lose_criteria_met[criteria] = true -- lose criteria are typically in their own group
+          end
+        end
+      end
+    end
+
+    -- Is this the best performing group in the winning goals?
+    if not best_value or win_criteria_met[g] > best_value then
+      best_value = win_criteria_met[g]
+      best_group = g
+      --[[ If all win criteria in a group are met the level will be won, no advice to give
+      if win_criteria_met[g] > 0 and criteria_count[g] == win_criteria_met[g] then
+        return
+      end]]
+    end
+  end
+
+  local c = 0
+  for _, _ in pairs(lose_criteria_met) do c = c + 1 end
+-- End combine with World:checkWinningConditions
+  local msgs = self:adviseGoalSpecifics(goals)
+
+  -- Generic positive
+  if win_criteria_met[best_group] >= criteria_count[best_group] * 0.75 then
+    table.insert(msgs, _A.level_progress.three_quarters_won)
+  elseif win_criteria_met[best_group] >= criteria_count[best_group] * 0.5 then
+    table.insert(msgs, _A.level_progress.halfway_won)
+  end
+
+  if #msgs == 0 then return end
+for _, m in pairs(msgs) do print(m["text"]) end -- Remove
+  self:giveAdvice(msgs, 1)
 end
 
 --! Give advice to the player at the end of a month.
--- a/CorsixTH/Lua/map.lua
+++ b/CorsixTH/Lua/map.lua
@@ -738,6 +738,22 @@ function Map:getParcelTileCount(parcel)
   return self.parcelTileCounts[parcel] or 0
 end
 
+--! Get the starting values of balance and reputation for this level
+--!return start (table)
+function Map:getLevelStartState()
+  local level_config, level, start = self.level_config, self.level_number, {}
+  if level_config.towns and level_config.towns[level] then
+    start["balance"] = level_config.towns[level].StartCash
+    start["reputation"] = level_config.towns[level].StartRep
+  elseif level_config.town then
+    start["balance"] = level_config.town.StartCash
+    start["reputation"] = level_config.town.StartRep
+  end
+  if not start["balance"] then start["balance"] = 0 end
+  if not start["reputation"] then start["reputation"] = 500 end
+  return start
+end
+
 function Map:afterLoad(old, new)
   if old < 6 then
     self.parcelTileCounts = {}


CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -216,6 +212,109 @@ function PlayerHospital:dailyAdviceChecks()
...
+  -- Starting position (measure progress from here towards win/lose)
+  local level_config, level, start = self.world.map.level_config, self.world.map.level_number, {}
+  if level_config.towns and level_config.towns[level] then
+    start["balance"] = level_config.towns[level].StartCash
+    start["reputation"] = level_config.towns[level].StartRep
+  elseif level_config.town then
+    start["balance"] = level_config.town.StartCash
+    start["reputation"] = level_config.town.StartRep
+  end
+  if not start["balance"] then start["balance"] = 0 end
```
lewri, line 226: Confirm, but base_config indicates the default `StartCash` is 40000.
`StartRep` is also 500 in that config. 
So why do we need to declare a failure case value here for both lines?

lewri (commented):
While I imagine this gives a lot of benefit to the player, the code itself is very argument heavy.

I also can't fully remember if this is like the original. I thought `_A.level_progress.cured_enough_patients` was a trigger based message once it was hit.

Though, I'm not a fan of the repeated use of `_A.level_progress.halfway_won` and `nearly_won` for some checks. If you had a balance at ratio 0.95 but everything else was almost at lose condition the advice could be wrong.

CorsixTH/Lua/hospitals/player_hospital.lua
lewri: `StartRep` was only introduced in 0.65 -- probably why you have that failsafe then.

tobylane:
https://www.youtube.com/watch?v=Z7XCXq8j4iA is a lets play of TH97, where the "half way won" advice comes when two criteria are met and the third is 90% there. After they meet all three, it still says "Keep the value of your hospital above win, and attend to other issues", which I'm preventing in this change.

Removed all the repeats in the positive section, just one in the negative. Cleaned up, better reading of the level's goals.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -216,6 +212,107 @@ function PlayerHospital:dailyAdviceChecks()
...
+    balance = self.world.goals[2]["win_value"],
+    cured_p = self.world.goals[3]["win_value"], -- Patients cured percentage
+    cured_c = self.world.goals[4]["win_value"], -- Patients cured count
+    value = self.world.goals[6]["win_value"]
+  }
+  local lose = {
+    reputation = self.world.goals[1]["lose_value"],
+    balance = self.world.goals[2]["lose_value"],
+    killed_p = self.world.goals[5]["lose_value"] -- Patients killed percentage
+  }
```
lewri, line 243: So a possible issue here is if the criterion is not set in the level file. Your checks will cause `attempt to compare nil with number`.

I have tricked myself a few times on trying to verify this so please clarify:
In base config the win/lose criteria is set but not against any value (all zeros).
```
  win_criteria = {
    [0] = {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
    {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
    {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
    {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
    {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
    {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
  },
  lose_criteria = {
    [0] = {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
    {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
    {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
    {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
    {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
    {Criteria = 0, MaxMin = 0, Value = 0, Group = 0, Bound = 0},
  },
```

However, none of these actually set values against e.g. criteria 6 (hospital value).
If we look at `EASY07.SAM` from Theme Hospital, only 4 win criteria are set (not pasting here).

As such, it looks like this PR doesn't accommodate for a not set value on these items.

lewri (commented):
As a side note, it looks like the Wiki is wrong in Custom Levels about win/lose criteria. I'm pretty sure we check against cured number rather than cured percentage?

```
win_criteria and lose_criteria

Here winning and losing conditions are defined each table element is defined to specify the following possible crieria:

1 = Total reputation

2 = Total balance

3 = Percentage of all patients your hospital has treated

4 = Percentage people that have been cured

5 = Percentage people that have been killed

6 = Hospital value
```

tobylane:
What should the priority be in [Strings:setupAdviserMessage?](https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/strings.lua#L310) Some of the other commented out lines can be filled in too. [Original commit of that function.](https://github.com/CorsixTH/CorsixTH/commit/134b9f0e0895e6381ffbed8a1bc1148f5a36e636#diff-895ff01013999cb631028dcdf94b0a7b20c9fcb7c4f0a9af5145eac7ccb9c846)

lewri:
Those priorities seem a little peculiar don't you think? Though if you're asking for now, default priority is probably fine.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -216,6 +212,131 @@ function PlayerHospital:dailyAdviceChecks()
...
+    balance = criteria[2] and criteria[2]["lose_value"],
+    killed_p = criteria[5] and criteria[5]["lose_value"] -- Patients killed percentage
+  }
+
+  -- Win criteria
+  --  1 Reputation
+  if win["reputation"] then
+    if self.reputation >= win["reputation"] then
+      if self.reputation > start["reputation"] then
+        table.insert(msgs, _A.level_progress.reputation_good_enough:format(win["reputation"]))
```
Alberth289346, line 252: Do I read correctly that you need to have (in this case) more reputation than than the stat as well as what the win criteria says?  If so, why not take `max(win["reputation"], start["reputation"])`  as win criteria?

Alberth289346:
I am not sure what you're implementing, but if it is win/lose of the level file, this is incomplete as far as I can see.

The `Group` value seems ignored. I am also not sure what happens if there are several lines with the same `Criteria`, especially if the `MinMax` or the `Value` are different there (People may copy/paste lines, keep old lines and only modify a few of them, etc etc, be robust against such things.)

The `Group` may need some more explanation. As I read it, it allows for combined criteria (which you seem to assume too) to win, but it also allows for different ways to win.

A few examples to make things more clear.
``` lua
win_criteria = {
  {Criteria = 1, MaxMin = 1, Value = 600, Group = 0, Bound = 0}, -- rep > 600
  {Criteria = 2, MaxMin = 0, Value = 10, Group = 0, Bound = 0}, -- balance < 10
}
```
Since `Group` is the same, both conditions must hold to win.

With different group numbers you get multiple ways to win:
``` lua
win_criteria = {
  {Criteria = 1, MaxMin = 1, Value = 600, Group = 0, Bound = 0}, -- rep > 600
  {Criteria = 2, MaxMin = 0, Value = 10, Group = 1, Bound = 0}, -- balance < 10
  {Criteria =6, MaxMin = 1, Value = 10000, Group = 1, Bound = 0}, -- hospital value > 10000
}
```
Two different group values, so 2 ways to win. Either
- rep > 600 (all criteria lines with group 0), **or**
- balance < 10 and hospvalue > 10000 (all criteria lines with group 1).

(ie all lines with the same group value must hold to win)

tobylane:
I hadn't seen that, thanks.

The win criteria section is ok except for the checks that not all other criteria are met. Generic negative is ok (or not made worse). Generic positive checks would need grouping at the win criteria stage, then find the group nearest to winning to use in the current setup.

CorsixTH/Lua/hospitals/player_hospital.lua
tobylane: It felt strange to be told the hospital was good enough when I had lost reputation/balance since the level started. In the first level of the campaign, 400 is enough to win, but not enough to be congratulated.

Alberth289346:
In case you missed it, the loose criteria also has a group, so you get the same problem at the negative side.

My first intuition on tackling the problem was to track end-results for all groups in a temporary table, and then simply update the end-result as you move along all criteria that exist. That should cover duplicate criteria as well as contradicting criteria.

The other direction is to first collect everything of a single group, process those, and then move on to the next group. Not sure what is simpler, mostly a matter of taste I think.

There is also a `Bound` value, which seems to be a threshold for warning the user. I am not sure where that is displayed, I am guessing in the progress window and not as advisor message, but don't know.

Last but not least, you may want to try splitting the code into several functions. Your current solution is a quite long. If you move some steps to a separate function, things will be easier to follow.

Alberth289346:
> What should the priority be in [Strings:setupAdviserMessage?](https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/strings.lua#L310) Some of the other commented out lines can be filled in too. [Original commit of that function.](https://github.com/CorsixTH/CorsixTH/commit/134b9f0e0895e6381ffbed8a1bc1148f5a36e636#diff-895ff01013999cb631028dcdf94b0a7b20c9fcb7c4f0a9af5145eac7ccb9c846)

oohh! my first ever contribution to CTH :)

tobylane:
I'm wondering what assumptions I can make. 

- There is a fixed number that is enough to look at all groups that will be made
- The player only wants to know about the collective group of winning criteria they are closest to,
- but all losing criteria, as they are (often?) singular
- The MinMax value will always be predictable, that only lose-balance and lose-reputation are criteria to be under

Alberth289346:
> There is a fixed number that is enough to look at all groups that will be made

Well yes, with a finite number of entries there is a finite number of groups.

Not sure why this is important though. Track the group numbers that you have done somewhere and repeatedly search for an entry with a group value you haven't done yet. Alternatively, track the first entry you have not done, and do the group of that entry.
In both cases you eventually run out of entries to do.

> The player only wants to know about the collective group of winning criteria they are closest to,

> but all losing criteria, as they are (often?) singular

I don't see why you treat them differently. Also for winning conditions there could be a different goal that I haven't much progressed in but would be easy to achieve.

I also believe that the initial briefing should be the main source in explaining the criteria to the user. In other words, finding out what cliffs are in the way for concluding the level successfully, and how to get around them is part of game-play.
Giving the player too much information reduces the challenge of the level. The point of any game is to waste time and have fun doing it. Failing and having to try again is thus not a bad thing in itself imho.

Perhaps the simplest option is to treat "win" and "loose" equally, The "nearest to goal" only seems fine enough to me.

> The MinMax value will always be predictable, that only lose-balance and lose-reputation are criteria to be under

I don't see any advantage of such assumptions. Sure, 99% of the levels will be that way, but you might be able to make real weird challenging levels by doing something non-standard. (I mean, high reputation and hospital value, but being near bankruptcy for a year is not a challenge you think? Good reputation with at least 100 people dead is not a challenge?) Please don't build blockades against level authors by adding assumptions of a level aims for.

You want to give as much room to the levels as you can, so you get more diverse levels. If every level is the same because you cannot express sufficient different challenges, you get reduced game replay value.

An entry is basically a binary decision, either you reached the goal of the entry or not. For a percentage how far you are, you may want to use the starting amount so you have a scale. (I think that's what you're doing at least.)
The difference between winning and loosing is that initially you're on the "bad" side of the former and the "good" side of the latter. That's all as far as I can see.

I am not sure what you're doing with the initial amount in the code, but please keep in mind that if a level author thinks that is a relevant condition, it can easily be expressed in the criteria.

tobylane:
The group nearest to goals doesn't work for lose as they're one goal per group. I count the total but can't think what to do with it. Improved on the other points, wip, only testing on the campaign so far.

It's now split into three functions, but that can go a bit further. Could start and finding the highest group number be made at the level start? Could most of the adviseGoalProgress function, up to the GoalSpecifics call, be merged with World:checkWinningConditions?

Level 12 has the balance win goal in group 2 and all other win goals in group 1. This isn't a problem to the progress report or World:checkWinningConditions. Does L12 have a bug covered up by nothing in the game yet looking at groups?

lewri:
It might be worth getting some thoughts from @mugmuggy on this too to clarify anything vs the original?

Alberth289346:
> The group nearest to goals doesn't work for lose as they're one goal per group. I count the total but can't think what to do with it.

This took a while to understand, but you're perhaps computing percentage of reached goals? That would be a problem indeed in that case.

Can't you use the progress within a single entry as well (or instead)? If a lose criteria is more than 10 dead, then 1 dead is obviously better than 9 dead, isn't it? (For winning, 9 dead would be better I think, so that seems to suggest inverting the percentage??? (Don't know.)

Would it work if you take 0 as `0%` and the threshold value as `100%` and then scale accordingly? (With a special case if threshold is 0.)  Alternatively, you could take the start value as `0%`, but that's more difficult and not sure there is any benefit in doing that, but you may have other ideas about that.

I feel there will be a few niggles there, but I'd have to work out a case systematically (eg an entry for a value above and below the threshold, both for win and loose and for min and max) to see what makes sense. In theory the threshold can also be negative, it may be worthwhile to check that too.  A third change may happen if the current value can only move in one direction (eg number of dead people can only increase), it makes some parts unreachable but don't know its impact in how you report or measure progress there.

> Could start and finding the highest group number be made at the level start?

You can, I am not sure why you'd want that number though. The only thing that the level file says about `Group` is "entries with the same group value must be combined". There is no statement about the actual allowed values of the group numbers itself that supports assumptions like starting with 0 or being consecutively numbered. I mean, `Group` numbers 0, 1, -15, and 100000 would be 4 groups too, wouldn't it?  
(Try being more paranoid. Seeing a nicely consecutively numbered set of criteria as example doesn't mean all files must have that pattern. You can make such assumptions if and only if there is a rule somewhere that actually says any other numbering system is not allowed.)

There are 2 valid paths here.
- Derive additional information from the original data, or convert the information to a simpler form for processing once. Then decide about giving advice etc from the new form.
- Do both things at the same time. While deciding about advice inspect the original data to find what entries to process next in some way.

Either path is fine, it's a matter of how much time can you save in checking, against how much time does it take to design and construct a new form as separate step first.

In either path, you must process the original data and work out groups at some point in time. In a previous comment I already stated a few ways to handle such arbitrary different group numbers. We can discuss that further if you like.

> Could most of the adviseGoalProgress function, up to the GoalSpecifics call, be merged with World:checkWinningConditions?

Less code is always better. Also, one piece of code that decides is better than two pieces of code for consistency reasons. (Thus, YES please :) )

On the other hand, giving advice is a different activity than checking for the end of the level, so you want to avoid connecting these so tightly that you cannot do one without also doing the other.  
Checking a level end without having the advisor pop up should be possible. The other way around, we may want to run the advisor advice more often than checking the end of the level for some reason.

There are several paths here.
- One path is that both activities call the same function for computing and returning conclusions, and each activity then uses the returned results differently.
- Another path is that you compute and store the conclusions every X days, and both activities use the stored conclusions when they need to. (This could be complicated due to using 'old' results, so perhaps a less good path?)
- A third option is that you compute, draw conclusions, and use the results for both activities in one function, but the function allows skipping some of the uses.

> Does L12 have a bug covered up by nothing in the game yet looking at groups?

Ha and you;re asking me, someone that never got beyond level 3 of the original levels? 😃 

It's definitely possible, but I cannot give you an answer. As it's in our files, you can raise an issue about it so it can be discussed and decided.

tobylane:
adviseGoalSpecifics mostly uses the strings intended for campaign, so I've included checks for the min_max_win/lose.

> One path is that both activities call the same function for computing and returning conclusions, and each activity then uses the returned results differently.

I think it will be this, that World:checkWinningConditions will return only the winning group with the most met criteria, and possibly all of the lose reasons.

CorsixTH/Lua/map.lua
```diff
@@ -738,6 +738,22 @@ function Map:getParcelTileCount(parcel)
...
+--!return start (table)
+function Map:getLevelStartState()
+  local level_config, level, start = self.level_config, self.level_number, {}
+  if level_config.towns and level_config.towns[level] then
+    start["balance"] = level_config.towns[level].StartCash
+    start["reputation"] = level_config.towns[level].StartRep
+  elseif level_config.town then
+    start["balance"] = level_config.town.StartCash
+    start["reputation"] = level_config.town.StartRep
+  end
```
Alberth289346, line 751:
```suggestion
  local towns
  if level_config.towns then
    if level_config.towns[level] then
      towns = level_config.towns[level]
    else
      towns = level_config.towns
    end
    
    start.balance = towns.StartCash or 0
    start.reputation = StartRep or 500
  end
```
Instead of copy/pasting a long sequence of indexing operations each time, introduce a variable that keeps the intermediate table available.

Also, by definition `x["abc"]` for table `x` is equivalent to `x.abc` but it has much less clutter. The rectangular brackets are useful for indexing with a variable like `[level]` above, or when your data is not an identifier, eg `x.["text with spaces"]`.

Alberth289346:
Code has too much near or completely duplicated lines and clutter. Avoid copy-paste, use a function if you have (nearly) the same computation more than once. (Checking 10 lines of nearly identical code is much much more difficult than seeing a single function call with identical or non-identical parameter values.)

I also am missing the reverse sides of the minmax checks as well as the group selection/splitting code.

tobylane:
I'll put this on hold, and in a new PR change goals in world and progress report, then come back to this.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -216,6 +212,143 @@ function PlayerHospital:dailyAdviceChecks()
...
+    table.insert(msgs, _A.level_progress.three_quarters_won)
+  elseif win_criteria_met[best_group] >= criteria_count[best_group] * 0.5 then
+    table.insert(msgs, _A.level_progress.halfway_won)
+  end
+
+  if #msgs == 0 then return end
+for _, m in pairs(msgs) do print(m["text"]) end -- Remove
+  self:giveAdvice(msgs, 1)
+end
+
```
lewri, line 351: Might just be indenting problems here, but would double check something hasn't been left open.

PR #2055, opened 2021-09-20
Refactor VIP Remarks

lewri:
Old method could cause two versions of the same string on our translations (leading to false flags for translators).

This now realigns the original strings to how remarks are meant to be ordered, eliminating the need for retranslation.

PR #2172, opened 2022-05-22
Reinstate legacy strings for VIP

### LINKED ISSUE #2171: [Bug] 0.65 VIP results Fax halts load in 0.66
### Describe the issue

If a player in 0.65 has a VIP report fax waiting when loading the game in 0.66 it can not load due to a non-existing string "ordered_remarks".

In 0.64 > 0.66 this also happens but with "good" / "bad" etc.

### Steps to Reproduce

1. Start a hospital in 0.65 and invite a VIP. Wait until they leave and the report fax generates

2.  Save game without opening  the fax.

3. Try to open in 0.66

### Expected Behaviour

Game should load. However, I can't see a method for fixing these strings as the failure happens before afterLoad.

### Save Game

[0-65vip2.zip](https://github.com/CorsixTH/CorsixTH/files/8749767/0-65vip2.zip)

### CorsixTH Version

0.66 beta-1

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

```bash
Error while loading game: Non-existent string: ordered_remarks
```

### Additional Information

_No response_

lewri:
*Fixes #2171*

- Reinstates VIP results fax strings from 0.64 and 0.65 for compatibility.

### DIFF
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -916,3 +916,56 @@ transactions.remove_room = "Build: Remove destroyed room"
 
 options_window.change_resolution = "Change resolution"
 tooltip.options_window.change_resolution = "Change the window resolution to the dimensions entered on the left"
+
+--[[ Compatibility mapping for VIP result faxes in old saves (< 0.66). Using non-
+standard string formatting here, which should not be repeated in normal
+circumstances. This mapping will cause the legacy string to print in English but only
+for the relevant fax. These should be deleted on 2024 release. ]]--
+fax = {
+  vip_visit_result = {
+    ordered_remarks = {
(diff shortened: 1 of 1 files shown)


lewri:
(or partially revert #2055 , but that makes more mess imo across multiple files)

colinjmatt (commented):
This does the trick. The save that couldn't load now does.

TheCycoONE:
To me this is a last resort fix. I'd like to see if I can find another option that doesn't involve english.lua

TheCycoONE:
Thanks, I see we've done this before on a small scale. Every other approach I considered seemed hackier.

lewri:
I agree. A last resort. That's why I added the comment to remove it after x releases. Whether we actually remove it then or not will depend I guess.

PR #2183, opened 2022-06-05
Use raw pointer for line_elements

### LINKED ISSUE #2182: [Bug] Game crashes on loading game with Graphs open
### Describe the issue

If a game is saved with the graphs screen open, the game will freeze and crash with no error when loading the save.
Crash is at the point of depersisting objects.

Noting the dump file output from Cyco:
>the error is in move_to on the line_sequence
>fX and fY are origin
>```
>Unhandled exception thrown: write access violation.
>this->line_elements._Mypair._Myval2.**_Mylast** was 0xFFFFFFFF00004403.
>```

Not present in 0.65.1.

### Steps to Reproduce

1. Start a new game of any kind
2.  Open the graphs screen and save game.
3. Attempt to load the game.

### Expected Behaviour

Game should not crash.

### Save Game

### CorsixTH Version

0.66-beta 1

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

```bash
See above from Cyco
```

### Additional Information

Ping @TheCycoONE

TheCycoONE:
When depersisting the constructor isn't called so the vector didn't
work.

Use placement constructor to ensure the vector is constructed.

Fixes #2182

### DIFF
--- a/CorsixTH/Src/th_lua_gfx.cpp
+++ b/CorsixTH/Src/th_lua_gfx.cpp
@@ -825,6 +825,7 @@ int l_line_depersist(lua_State* L) {
   lua_insert(L, 1);
   lua_persist_reader* pReader =
       static_cast<lua_persist_reader*>(lua_touserdata(L, 1));
+  new (pLine) line_sequence();
   pLine->depersist(pReader);
   return 0;
 }


lewri:
Am I needed to verify this one?

TheCycoONE:
@lewri Albert and I could both reproduce, so I don't think so.

Alberth289346:
Confirmed as fix for the crash.

PR #2186, opened 2022-06-07
Announce fall only when player initiated

lewri:
- Adds the code so that players have a chance to push patients over when they right click on them.
- Code has been commented out until falling is more reliable. However, it seems to work quite well with #1848 
- Changes to announcing a fall only when the player caused it (as the adviser messages should be)
- Some updates to the luadoc comments.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -356,8 +356,9 @@ function Patient:canPeeOrPuke(current)
   return parcel ~= 0 and th:getPlotOwner(parcel) == self.hospital:getPlayerIndex()
 end
 
---! Animations for when there is an earth quake
-function Patient:falling()
+--! Falling animations for when there is an earth quake or the player is being very mean
+--!param player_init (bool) if true, player triggered the fall
+function Patient:falling(player_init)
   local current = self:getCurrentAction()
   current.keep_reserved = true
   if self.falling_anim and self:canPeeOrPuke(current) and self.has_fallen == 1 then
@@ -376,13 +377,14 @@ function Patient:falling()
     if math.random(1, 5) == 3 then
       self:shakeFist()
     end
-    self:fallingAnnounce()
+    if player_init then self:fallingAnnounce() end
     self:changeAttribute("happiness", -0.05) -- falling makes you very unhappy
   else
     return
   end
 end
 
+--! Announcements specific to a patient being pushed over by the player
 function Patient:fallingAnnounce()
   local msg = {
   (_A.warnings.falling_1),
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -665,7 +665,7 @@ function World:tickEarthquake()
         -- they visit reception. Some debugging needed here to get
         -- this working.
 
-        -- patient:falling(false)
+        patient:falling(false)
       end
     end
     --]]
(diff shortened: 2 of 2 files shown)


CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -72,6 +72,13 @@ function Patient:onClick(ui, button)
         ui:addWindow(UIPatient(ui, self))
       end
     end
+  --[[ Support for knocking over patients, currently disabled
+  elseif button == "right" and not self.world:isPaused() then
```
lewri, line 76: Possibly something better than `self.world:isPaused()`

lewri: Or maybe this check is in the wrong place?

lewri: I've decided to move this down into the next if statement, to make the parent if cleaner.

Alberth289346:
I am not a fan of having commented-out game code, let alone add more. I'd suggest to hold this until falling and standing up is reliable.

lewri:
Maybe just add the Boolean then to falling and leave it there?

Alberth289346:
> Maybe just add the Boolean then to falling and leave it there?

Still changing dead code, but fair enough. I'd like something "player" in the name, ie "pushed" doesn't say "player-initiated" to me, perhaps `player_pushed`? (bit long though).

Also, since you're extending the signature of a function, I'd expect a `false` argument somewhere.

----
Should be in another issue (ie not here), but you're using `(self.cured or self.dead or self.going_home)`. Wouldn't it be useful to make a function for querying this state in the patient and use that? (That is, avoid grabbing internal information from another object, ask it for its higher level state with a query function instead.)

Not sure about a good name of such a function though, I am thinking `isDone()` or `isFinished()` or `isFinishing()` but they all feel as not properly expressing the intended notion to me.

By doing that, you get a name that expresses what you are computing in that expression. That is, `pat:isDone()` versus `(pat.a and not pat.b or perhaps pat.c and otherwise pat.q * 15 < 27)` what would you like better?  
Another change that using a function causes is that, if the patient code gets extended and a new variable gets added for this state, then changing the `return` statement in that function is way simpler than finding all the `cured` `dead` and `going_home` uses throughout the many files, judge _correctly_ in each of the zillion cases whether the new variable should be added, and if so, add it.

lewri:
Just dropping the removed code so it isnt lost

```lua
  --[[ Support for pushing over patients, currently disabled
  elseif button == "right" then
    -- Attempt to push patient over
    if not self.world:isPaused() and not (self.cured or self.dead or self.going_home)
         and math.random(1, 3) == 2 then
      self:falling(true)
    end
  --]]
```

CorsixTH/Lua/world.lua
```diff
@@ -665,7 +665,7 @@ function World:tickEarthquake()
         -- they visit reception. Some debugging needed here to get
         -- this working.
 
-        patient:falling()
+        -- patient:falling(false)
```
lewri, line 668:
```suggestion
        patient:falling(false)
```

Alberth289346:
Looks ok

lewri:
> Maybe add a debug flag if you're extending falling so people can play with it. jpirie's comment is quite old. This way it's no longer dead code

Without #1848 it's just not stable enough. I might commit this change to @mugmuggy 's fork to consider adding it into his PR and close this one.

PR #2191, opened 2022-06-16
Add cli function for decoding RNC files.

Alberth289346:
Add main function for decoding RNC files at the cli.

Still needs a CMake wizard, and perhaps some other CI integration wizardry for the finishing touches.


lewri:
Was this one meant to be a branch inside corsixth rather than your fork? Not that it really matters...

TheCycoONE:
Updated this branch with CMake (and moved some stuff around)

CMakeLists.txt
```diff
@@ -139,6 +140,12 @@ if(BUILD_ANIMVIEW)
   message("Building AnimView")
   add_subdirectory(AnimView)
 endif()
+
+if(BUILD_TOOLS)
+	message("Building Tools")
+	add_subdirectory("tools")
```
Alberth289346, line 146:
```suggestion
  message("Building Tools")
  add_subdirectory("tools")
```
Replaced TAB by 2 sapces

Alberth289346:
Tried building it. After setting `BUILD_TOOLS`, it generated a working binary. I noticed the option is mentioned in the comments, but it doesn't have a line like
``` cmake
option(ENABLE_UNIT_TESTS "Enables Unit Testing Targets" OFF)
```
No big deal, cli commands are a niche anyway.

Seems fine thus.

TheCycoONE:
> Tried building it. After setting `BUILD_TOOLS`, it generated a working binary. I noticed the option is mentioned in the comments, but it doesn't have a line like
> 
> ```cmake
> option(ENABLE_UNIT_TESTS "Enables Unit Testing Targets" OFF)
> ```
> 
> No big deal, cli commands are a niche anyway.
> 
> Seems fine thus.

Seems more consistent to fix it.

PR #2194, opened 2022-06-30
Use config value for next VIP date

lewri:
- Places previously unused config value `gbv.MayorLaunch` into `_generateNextVipDate()` -- in theory allows custom levels to set their own next VIP frequency from the hard-coded default.
- Fixes a rare case where `math.n_random` may return a number <= 0 causing VIP spawns to stop. Instead, if this happens, the next VIP date will use the `mean` value.
- VIP spawns may end up happening more frequently now with this, as the original intended value was 150 days (not 180).
- Variance value remains hard-coded, but could be added as a custom config value in future.


CorsixTH/Lua/world.lua
```diff
@@ -1378,10 +1378,12 @@ end
 -- PRIVATE method to generate the next VIP date
 function World:_generateNextVipDate()
   -- Support standard values for mean and variance
-  local mean = 180
+  local mean = self.map.level_config.gbv.MayorLaunch
   local variance = 30
   -- How many days until next vip?
   local days = math.round(math.n_random(mean, variance))
+  -- Add a safety net so the next_vip_date can't end up <= 0
```
lewri, line 1385: This comment might be too literal to the code line.
Perhaps "The VIP is not a time traveller" for fun or remove completely?

Alberth289346: The real culprit is likely that a normal distribution is the wrong way to compute this, a Poisson distribution seems much more appropriate: https://en.wikipedia.org/wiki/Poisson_distribution

That trivially solved the problem, as the range of the latter is `[0,infinity)` rather than `(-infinity, infinity)` of the normal distribution.

Alberth289346: Another option is to use the binomial distribution: https://en.wikipedia.org/wiki/Binomial_distribution which has similar properties as Poisson.

lewri:
Waiting on #2198

CorsixTH/Lua/world.lua
```diff
@@ -1379,11 +1379,11 @@ end
 
 -- PRIVATE method to generate the next VIP date
 function World:_generateNextVipDate()
-  -- Support standard values for mean and variance
-  local mean = 180
-  local variance = 30
+  -- Support standard values for mean and bound
+  local mean, bound = self.map.level_config.gbv.MayorLaunch, 30
```
Alberth289346, line 1383: I think 10 days between VIP visits would be nice to keep the user on its toes :p 
Your bound may now cause VIPs traveling in time to earn more money in the past.

TheCycoONE:
Any reason you preferred the triangle distribution to the normal distribution that was there before?

Alberth289346:
Reason was finite maximum deviation from the mean so it never picks days before today. I first suggested Poisson that has a non-negative range, but Lewri wanted more control over variance.

lewri:
> Reason was finite maximum deviation from the mean so it never picks days before today. I first suggested Poisson that has a non-negative range, but Lewri wanted more control over variance.

Also that poisson doesxn't provide any variance control. So the output from Poisson can be anything +ve.

PR #2196, opened 2022-07-02
Add missing parentheses, fix 2 style problems.

Alberth289346:
Fixing style issues.

For people that love PEG, please fix `string_extensions.lua`. (EDIT: Reason: Operator priority between parameter-less function application and other operators is not defined. Also, argument processing of such functions is not consistent. Therefore I don't know what the code means.)

Perhaps alternatively, use native string matching. It exists in lua 5.1 .

### DIFF
--- a/CorsixTH/Lua/audio.lua
+++ b/CorsixTH/Lua/audio.lua
@@ -111,7 +111,7 @@ function Audio:init()
   end
   for file in _f, _s, _v do
     local info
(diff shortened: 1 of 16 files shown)


lewri:
Cool, no problems loading the game after this? (Not that I'd expect it to have problems)

Alberth289346:
Fixed remaining missing parentheses in `string_extensions.lua`, thanks thecycoone for the pointer.

Game loaded without any problem.

PR #2197, opened 2022-07-03
Add Poisson distribution function.

Alberth289346:
Add pseudo-random Poission distribution function.

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1259,21 +1259,36 @@ end
 -- TODO: Requires adjustment for AIHospital spawns; see PR 1986 for progress.
 function World:updateSpawnDates()
   local local_hospital = self:getLocalPlayerHospital()
-  -- Set dates when people arrive
-  local no_of_spawns = math.n_random(self.spawn_rate, 2)
-  -- Use ceil so that at least one patient arrives (unless population = 0)
-  no_of_spawns = math.ceil(no_of_spawns*self:getLocalPlayerHospital().population)
+
+  -- Decide on number of visitors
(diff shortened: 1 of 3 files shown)


Alberth289346:
Holding, until triangular distribution is in master, and perhaps until `random.c` has been rewritten.

CorsixTH/Lua/world.lua
```diff
@@ -1265,19 +1265,39 @@ end
...
+      local x = math.p_random(interval)
+      day = day + x
+      if day <= last_day then
+        local count = self.spawn_dates[day]
+        self.spawn_dates[day] = count and count + 1 or 1
+        force_arrival = false
+      end
+    end
+
+    -- Give the poor user a patient this month anyway.
```
lewri, line 1297: 🙃

lewri (commented):
Still inclined to push this forward once everything else is satisfied

Alberth289346:
Rebased, not sure the patient spawning is right in this way. I am currently thinking to generate patients for all hospitals, and then drop the patients for non-opened hospitals.

Alberth289346:
Just removed draft status, what needs to be done here?

lewri:
> Just removed draft status, what needs to be done here?

I left a comment a while ago. I guess I made a few spawn changes around this area but I don't think they'd have an impact?

You also mentioned something about `random.c`?

TheCycoONE:
It's fine. There was some earlier conversation about using the c++ std random which includes both mt and a variety of distributions.

lewri:
@Alberth289346 sounds like this can be merged, unless you have anything else you wanted to amend first.

Alberth289346:
Nope, this is done.

PR #2198, opened 2022-07-03
Add Triangular disitrbution function.

Alberth289346:
Add Triangular distribution function.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -198,6 +198,26 @@ function App:init()
   -- App initialisation 2nd goal: Load remaining systems and data in an appropriate order
 
   math.randomseed(os.time() + SDL.getTicks())
+
+  -- Add math.t_random globally.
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/app.lua
```diff
@@ -198,6 +198,26 @@ function App:init()
   -- App initialisation 2nd goal: Load remaining systems and data in an appropriate order
 
   math.randomseed(os.time() + SDL.getTicks())
+
+  -- Add math.t_random globally.
+  -- It generates pseudo random triangular distributed numbers in interval (a, b)
+  -- mostly around c between a and b.
+  strict_declare_global "math.t_random"
+  math.t_random = function(a, c, b)
```
lewri, line 206: `-- luacheck: ignore 142`?
This code does work. Just not sure why `math.t_random` kicks up a fuss on luacheck but not `math.n_random`...

lewri: Ignore this, thanks for finding out why Albert

lewri (approved):

PR #2199, opened 2022-07-03
Cap popn in base_config

lewri:
Current set-up meant popn could increase to a level no longer enjoyable for the player.
Cap may need to be revisited later once AIHospitals actually receive a % of the population proper.

Although almost all levels/customs set a popn value, we should still account for a case where it isn't.


Alberth289346:
LGTM

TheCycoONE (approved):

lewri:
@TheCycoONE based on previous discussion with Alberth I've changed the cap to be 30 (30 / 4 = 7.5 -- rounds to 8 per month at 500 rep). 
It turns out an outdated comment was deceptive in thinking the allocated population was 50%. So I've also removed that

PR #2201, opened 2022-07-05
Cleanup world

Alberth289346:
Move some data and code out of world.

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -1609,6 +1609,11 @@ function UIEditRoom:placeObject()
 end
 
 function UIEditRoom:afterLoad(old, new)
+  if old < 171 then
+    self.wall_types = TheApp.walls
+    self:initWallTypes()
+  end
+
   UIPlaceObjects.afterLoad(self, old, new)
   self:registerKeyHandlers()
 end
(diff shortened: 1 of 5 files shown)


Alberth289346:
editroom still needs an afterload for the new wall data

CorsixTH/Lua/world.lua
```diff
@@ -905,13 +836,13 @@ function World:previousSpeed()
   end
 end
 
--- Return if the selected speed the same as the current speed.
+--! Return true if the given speed the same as the current speed.
```
lewri, line 839: It's `!return` for luadoc isn't it?

lewri (commented):
Did you need to do something in edit room still btw, or is it not important to this pr?

Alberth289346:
Done and confirmed it works. All set now

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -90,6 +93,58 @@ function UIEditRoom:UIEditRoom(ui, room_type)
   self:registerKeyHandlers()
 end
 
+function UIEditRoom:initWallTypes()
```
TheCycoONE, line 96: One bad place to another... but it's only used here, so I accept.

CorsixTH/Lua/world.lua
TheCycoONE: Yes, but convention is to prefer the method summary to the return parameter description when doing both would be redundant. The change is good.

TheCycoONE (approved):

CorsixTH/Lua/world.lua
Alberth289346: @lewri: I just aimed at getting the line recognized in luadoc, as it seemed mostly complete although I agree not entirely according to luadoc.

lewri (approved):
Give or take my comment, looks fine.

PR #2203, opened 2022-07-06
Replace hospitals[1] in code

lewri:
*Fixes #*

- Replaces a use of `hospitals[1]` in World and Annual Report with `getLocalPlayerHospital`
- With the upcoming changes to spawns in #1986 I haven't touched the `hospitals[1]` in spawn.
- Also didn't touch the `hospitals[1]` for `initStaff` -- as changes need to be made to the called function to work on all hospitals.
- Added a note that the Fame Screen isn't currently accessible from Annual Report (though looks like someone started coding it years ago).

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -817,9 +817,10 @@ end
 --!param room (Room) The new room.
 function World:markRoomAsBuilt(room)
   room:roomFinished()
-  local diag_disease = self.hospitals[1].disease_casebook["diag_" .. room.room_info.id]
+  local hosp = self:getLocalPlayerHospital()
+  local diag_disease = hosp.disease_casebook["diag_" .. room.room_info.id]
   if diag_disease and not diag_disease.discovered then
-    self.hospitals[1].disease_casebook["diag_" .. room.room_info.id].discovered = true
+    hosp.disease_casebook["diag_" .. room.room_info.id].discovered = true
   end
   for _, entity in ipairs(self.entities) do
     if entity.notifyNewRoom then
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/world.lua
```diff
@@ -817,9 +817,10 @@ end
 --!param room (Room) The new room.
 function World:markRoomAsBuilt(room)
   room:roomFinished()
-  local diag_disease = self.hospitals[1].disease_casebook["diag_" .. room.room_info.id]
+  local hosp = self:getLocalPlayerHospital()
+  local diag_disease = hosp.disease_casebook["diag_" .. room.room_info.id]
   if diag_disease and not diag_disease.discovered then
-    self.hospitals[1].disease_casebook["diag_" .. room.room_info.id].discovered = true
+    hosp.disease_casebook["diag_" .. room.room_info.id].discovered = true
```
Alberth289346, line 823: For multi-player we would also like to get rid of "local player hospital". So here use the hospital of the room instead. In that way this function works for any hospital.  

"local hospital" is thus mostly about "the hospital being shown in the UI", and not so much about "the hospital that gets changed" since in multiplayer, all 4 hospitals get changed all the time.

Another thing here ( but definitely out of scope here!! ), is that world is doing hospital administrative updates on behalf of the finished room. So from somewhere you have a finished room, it tells world, world tells room, world updates hospital internals, and world notifies entities. Does that strike you as a logical path? (In particular in the light of "each object messes ONLY with its own internals").

lewri: This crossed my mind too. So I'll look into that.

Alberth289346:
LGTM

PR #2208, opened 2022-07-17
Obey GP's Diagnosis Room Choice

lewri:
Just a starting note -- I am not good with the action queue, so likely this change is a little cautious or not optimal?

- Fixes the issue where the GP chose a specific diagnosis room for the patient, then the action queue threw that out the window.
- Seek room action now obeys the GP's requested diagnosis room for the patient, if that room is available
- Some additional comments added

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/seek_room.lua
+++ b/CorsixTH/Lua/humanoid_actions/seek_room.lua
@@ -48,21 +48,29 @@ function SeekRoomAction:setDiagnosisRoom(room)
   return self
 end
 
+--! Finds a relevant diagnosis/treatment room(s) for a patient
+--! If diagnosis room, attempt to use GP's choice first, else select any other available
+--! room at random.
 local action_seek_room_find_room = permanent"action_seek_room_find_room"( function(action, humanoid)
   local room_type = action.room_type
   if action.diagnosis_room then
+    -- Attempt to use the GP's chosen diagnosis room first
+    if room_type then
+      local room = humanoid.world:findRoomNear(humanoid, room_type, nil, "advanced")
+      if room then return room end
+    end
+
     local tried_rooms = 0
     -- Make numbers for each available diagnosis room. A random index from this list will be chosen,
-    -- and then the corresponding room index is taken as next room. (The list decrease for each room
+    -- and then the corresponding room index is taken as next room. (The list decreases for each room
     -- missing)
     local available_rooms = {}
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -188,7 +198,7 @@ local function action_seek_room_start(action, humanoid)
     humanoid:finishAction()
     return
   end
-  -- Tries to find the room, if it is a diagnosis room, try to find any room in the diagnosis list.
+  -- Try to find the appropriate room for the patient
   if not humanoid.diagnosed or action.room_type == "research" then
     local room = action_seek_room_find_room(action, humanoid)
```
lewri, line 203: Think the `local` can be dropped from this line now? Just want someone to confirm.

lewri: Thanks github actions,for confirming

lewri: (I actually meant line 68)

CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -48,12 +48,22 @@ function SeekRoomAction:setDiagnosisRoom(room)
...
 end
 
+--! Finds a relevant diagnosis/treatment room(s) for a patient
+--! If diagnosis room, attempt to use GP's choice first, else select any other available
+--! room at random.
 local action_seek_room_find_room = permanent"action_seek_room_find_room"( function(action, humanoid)
   local room_type = action.room_type
   if action.diagnosis_room then
+    -- Attempt to use the GP's chosen diagnosis room first
+    local room
```
TheCycoONE, line 58: No need to expand room's scope like this, you can have two `local room` in the same function when the scope doesn't overlap.

lewri: You could try moving it to line 60 to avoid the warning perhaps?

Alberth289346: Correct.

`room_at_index` declaration could also be moved more inward.

lewri:
Scope issues fixed

Alberth289346:
Out of scope, but I wonder if the "available diagnosis rooms" of the patient gets extended when I build a new diagnosis room.

PR #2209, opened 2022-07-19
[WIP] Support subdirectories in customs

lewri:
Allows one level deep subdirectory organisation for Levels and Campaigns.

~**Not currently working:** Advancing through campaigns from a subdirectory. Working on fixing that.~ Fixed

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_game.lua
+++ b/CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_game.lua
@@ -18,6 +18,8 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
+local pathsep = package.config:sub(1, 1)
+
 --! Custom Game Window
 class "UICustomGame" (UIMenuList)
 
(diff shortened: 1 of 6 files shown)


Alberth289346:
Downloaded campaigns are unzipped?
Probably the same question arises for downloaded level files.

lewri:
Luacheck issues should be fixed now.

TheCycoONE:
I don't believe we need a licence file with every campaign, if the repo itself specifies that everything in it is CC BY 4.0. From the human readable blurb. "You must give [appropriate credit](https://creativecommons.org/licenses/by/4.0/#), provide a link to the license, and [indicate if changes were made](https://creativecommons.org/licenses/by/4.0/#). You may do so in any reasonable manner, but not in any way that suggests the licensor endorses you or your use."

TheCycoONE:
I wanted the level files to specify authorship both for themselves and for the maps the point to. I thought they might also indicate license but wasn't fixated on that.

TheCycoONE:
Sandbox is a little leaky in this implementation. I would except that if there was a campaign subdir e.g. fooCampaign, with foo.level, it would always be used, rather than some other foo.level in another directory / another campaign. Only if there is none in the same folder would it possibly look in the other locations.

lewri:
> Sandbox is a little leaky in this implementation. I would except that if there was a campaign subdir e.g. fooCampaign, with foo.level, it would always be used, rather than some other foo.level in another directory / another campaign. Only if there is none in the same folder would it possibly look in the other locations.

Could you elaborate on that a bit more please as it was a bit confusing.

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_game.lua
```diff
@@ -97,6 +78,37 @@ function UICustomGame:UICustomGame(ui)
...
+      intro = level_info.briefing,
+      deprecated_variable_used = level_info.deprecated_variable_used,
+    }
+  end
+end
+
+--! Compile metainfo for all of the levels in the given path. For better Levels
+--! folder organisation, allow subdirectories one level deep
+--!param dir (string) The top-level directory that should contain level files.
+function UICustomGame:_findLevels(dir)
```
TheCycoONE, line 101: This change allows levels one folder deep. That's what my comment was about, I wouldn't do that without considering the implications in it's own PR

lewri: Ok makes sense now.

lewri:
I've now yeeted the archaic use of subdir from the PR. It is now all done on `App:getAbsolutePathToLevelFile` -- which evidently has gone out of scope of its name now. I may update the name later.

I still need to address the issue of making campaigns to use the same subdirectory preferential.

lewri:
Would we be okay if the campaign's directory was stored inside `Map` to be called on level advance? It would then mean, that folder can be checked first for the next level data before going to others.

lewri:
Will probably close this one once I pick it up again given appveyor changes (easier than a rebase at this point)

PR #2212, opened 2022-08-06
Make couches fully passable

### LINKED ISSUE #2211: [Bug] Psychiatric couch tiles are not passable
### Describe the issue

After enabling the debug flag for "Allow blocking off areas" it is possible to build a Psychiatry room like this:
image

The room looks functional, but the game crashes when a patient enters the room because the doctor is unable to path to the comfy chair.
The tile in front of the long couch and the plant combine to make it impossible to reach that chair.

### Steps to Reproduce

2. Wait for the patient to enter the Psychiatry room

3. Observe crash when the doctor's pathfinding starts.

To prevent the bug from happening, simply move the plant.

### Expected Behaviour

The game doesn't crash. Instead, the doctor chooses a path that takes him over the tiles in front of the long couch.

### Save Game

[PsychiatryPathing.zip](https://github.com/CorsixTH/CorsixTH/files/9274687/PsychiatryPathing.zip)

### CorsixTH Version

Release 0.66

### Operating System

Windows 10

### Theme Hospital Version

CD

### Gamelog.txt

_No response_

### Additional Information

_No response_

lewri:
*Fixes #2211*

- Sets both tiles adjacent to couch as passable.
- Previously, the non-passable tile could have a small object put on it. If this is the case, pathfinding will still work as intended after this PR (doctor sees blocked path). However, the player will not be able to place an object on the tile again if they edit the room.

### DIFF
--- a/CorsixTH/Lua/objects/couch.lua
+++ b/CorsixTH/Lua/objects/couch.lua
@@ -54,11 +54,11 @@ object.usage_animations = copy_north_to_south {
 }
 object.orientations = {
   north = {
-    footprint = { {-1, -1, complete_cell = true}, {-1, 0, complete_cell = true}, {0, -1}, {0, 0, only_passable = true} },
+    footprint = { {-1, -1, complete_cell = true}, {-1, 0, complete_cell = true}, {0, -1, only_passable = true}, {0, 0, only_passable = true} },
     render_attach_position = {-1, 0},
   },
   east = {
-    footprint = { {-1, -1, complete_cell = true}, {-1, 0}, {0, -1, complete_cell = true}, {0, 0, only_passable = true} }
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/objects/couch.lua
```diff
@@ -54,11 +54,11 @@ object.usage_animations = copy_north_to_south {
 }
 object.orientations = {
   north = {
-    footprint = { {-1, -1, complete_cell = true}, {-1, 0, complete_cell = true}, {0, -1}, {0, 0, only_passable = true} },
+    footprint = { {-1, -1, complete_cell = true}, {-1, 0, complete_cell = true}, {0, -1, only_passable = true}, {0, 0, only_passable = true} },
```
lewri, line 57: Just a sidenote: an object may already exist on this now "only_passable" cell. This pr doesn't account for such a case (because nothing breaks with the object there, currently)

PR #2217, opened 2022-08-20
Show queueing patients when hovering the door

### LINKED ISSUE #2016: Show queued patients while hovering on door
While hovering over the door the same "queue"-symbol should be shown when hovering over the room itself.

When hovering over the door the infobox at the bottom shows the number of queued patients but not the "queued"-symbol, but while hovering over the room it is the other way around. So you either know how many are queued or know who is queued.

Yes

0.65

Hovering over room
Hovering over door
Hovering over door in Theme Hospital

SilPho:
*Fixes #2016*

When hovering the door to a room (assuming the door is on the north or west edge), show the queue icon above patients that are queueing for that room.

(First time submitting a PR to this repo, so hopefully I'm doing it right)

### DIFF
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -476,9 +476,15 @@ function GameUI:onCursorWorldPositionChange()
   if not overwindow and wx > 0 and wy > 0 and wx < self.app.map.width and wy < self.app.map.height then
     room = self.app.world:getRoom(wx, wy)
   end
-   -- Enable queue icons for patients when hovering the door as well
-  if entity and not room and entity.room then
-    room = entity.room
+  -- Find the room associated with the current entity (Usually only applies to doors)
+  if entity and not room then
+    if entity.room then
+      room = entity.room
+    end
+    -- Special case to catch the non-dominant side of a double-door
+    if not room and entity.object_type and entity.object_type.class == "SwingDoor" then
+      room = entity.master.room
+    end
   end
   if room ~= self.cursor_room then
     -- Unset queue mood for patients queueing the old room
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -480,7 +480,7 @@ function GameUI:onCursorWorldPositionChange()
   if entity and not room then
     if entity.room then
       room = entity.room
-    elseif entity.object_type and entity.object_type.class == "SwingDoor" then
+    elseif entity.object_type and entity.object_type.id == "swing_door_left" then
       -- Special case to catch the non-dominant side of a double-door
       room = entity.master.room
     end
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -476,6 +476,10 @@ function GameUI:onCursorWorldPositionChange()
   if not overwindow and wx > 0 and wy > 0 and wx < self.app.map.width and wy < self.app.map.height then
     room = self.app.world:getRoom(wx, wy)
   end
+   -- Enable queue icons for patients when hovering the door as well
+  if entity and not room and entity.room then
+    room = entity.room
+  end
   if room ~= self.cursor_room then
     -- Unset queue mood for patients queueing the old room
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/game_ui.lua
```diff
@@ -475,6 +475,10 @@ function GameUI:onCursorWorldPositionChange()
   local room
   if not overwindow and wx > 0 and wy > 0 and wx < self.app.map.width and wy < self.app.map.height then
     room = self.app.world:getRoom(wx, wy)
+  end
+   -- Enable queue icons for patients when hovering the door as well
```
lewri, line 479: Comment slightly off. Your fix is for hovering over the room.

lewri (commented):
It works, great! :)

lewri:
If you have time, add yourself to the `LICENSE.txt` file after the long list of authors around the top.

CorsixTH/Lua/game_ui.lua
SilPho: My change detects whether the entity you are hovering over has a "room" field attached to it, which doors do. However, I have noticed that only one half of a double-door has that field. So this fix doesn't work perfectly for Operating Theatres or Wards.
I could adjust the check to look for entity.room OR "entity.isMaster = =false", because that would catch the second half of a double door, but that feels like a really janky solution.
I'll keep poking for another, more definitive, way to detect if the entity we are hovering over is a door or not.

CorsixTH/Lua/game_ui.lua
```diff
@@ -476,6 +476,16 @@ function GameUI:onCursorWorldPositionChange()
   if not overwindow and wx > 0 and wy > 0 and wx < self.app.map.width and wy < self.app.map.height then
     room = self.app.world:getRoom(wx, wy)
   end
+  -- Find the room associated with the current entity (Usually only applies to doors)
+  if entity and not room then
+    if entity.room then
+      room = entity.room
+    end
+    -- Special case to catch the non-dominant side of a double-door
+    if not room and entity.object_type and entity.object_type.class == "SwingDoor" then
```
lewri, line 485: You can change this to an else if so you don't have to check `not room` again. Good thinking about the double door.

SilPho: Thanks, good spot. Comitted an update for that

lewri (commented):
LGTM

CorsixTH/Lua/game_ui.lua
```diff
@@ -476,6 +476,15 @@ function GameUI:onCursorWorldPositionChange()
   if not overwindow and wx > 0 and wy > 0 and wx < self.app.map.width and wy < self.app.map.height then
     room = self.app.world:getRoom(wx, wy)
   end
+  -- Find the room associated with the current entity (Usually only applies to doors)
+  if entity and not room then
+    if entity.room then
+      room = entity.room
+    elseif entity.object_type and entity.object_type.class == "SwingDoor" then
```
Alberth289346, line 483: If you check "object_type.id" you get "swing_door_left" (the slave door, in `objects/doors/swing_door_left.lua`) or you get "swing_door_right" (the master door, in `objects/doors/swing_door_right.lua`).

In the latter file is also the SwingDoor class, and it has code to automatically handle differences between slave and master doors, maybe you can use that (not sure).

lewri: Looks feasible, except for the first 'tick' where only swing_door_right exists to the class.

lewri: Any further thoughts on this one?

SilPho: Sorry, missed these comments. Yes, using the id sounds much more sensible than using a class name. We only need to check for swing_door_left as a special case because the right-hand door behaves like any other door in a single-door room.

lewri:
Appveyor failure isn't you, if you remove the `[rfc]` when finished then cyco can poke it with his stick (may also need him to manually merge this one)

TheCycoONE:
Working on rerunning the build

TheCycoONE:
ok, I have to fix this one too...

PR #2219, opened 2022-08-22
Changes to max_salary behaviour

lewri:
*Fixes #*

- Maximum salary converted to configuration item (debate over if the max value should be changed is a separate thing imo)
- Staff will not request a raise if already at maximum salary (i.e. never, previously we just hoped and prayed)
- Salary raise now uses the correct sound
- Bonus and salary raise buttons in Staff Management no longer play "selectx" and dedicated sound together.
- Staff Management now shows the maximum salary, if a salary raise would hit that limit
- Clicking on salary raise button when at maximum salary now results in the "wrong2" sound
- Happiness is now fixed at 1 when at maximum salary
- Some rewording of inline comments
- Bugfix: Now properly deactivates salary rise and resets happiness if staff member hit max salary.

There may be a bit of redundancy in here still. I suppose I was trying to add safety nets that may not be needed.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -29,7 +29,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 173 -- Make couches fully passable objects
+local SAVEGAME_VERSION = 174 -- Max salary improvements
 
 class "App"
 
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -29,7 +29,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 174 -- Max salary improvements
+local SAVEGAME_VERSION = 175 -- Max salary improvements
 
 class "App"
 
--- a/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
@@ -67,9 +67,9 @@ function UIStaffManagement:UIStaffManagement(ui)
   -- Other buttons
   self:addPanel(0, 12, 86):makeButton(0, 0, 31, 74, 2, self.scrollUp):setTooltip(_S.tooltip.staff_list.prev_person)
   self:addPanel(0, 12, 274):makeButton(0, 0, 31, 74, 3, self.scrollDown):setTooltip(_S.tooltip.staff_list.next_person)
-  self:addPanel(0, 319, 372):makeButton(0, 0, 112, 39, 7, self.payBonus):setTooltip(_S.tooltip.staff_list.bonus)
-  self:addPanel(0, 319, 418):makeButton(0, 0, 112, 39, 8, self.increaseSalary):setTooltip(_S.tooltip.staff_list.pay_rise)
-  self:addPanel(0, 438, 372):makeButton(0, 0, 45, 85, 9, self.fire):setTooltip(_S.tooltip.staff_list.sack)
+  self:addPanel(0, 319, 372):makeButton(0, 0, 112, 39, 7, self.payBonus):setTooltip(_S.tooltip.staff_list.bonus):setSound() -- custom sound
+  self:addPanel(0, 319, 418):makeButton(0, 0, 112, 39, 8, self.increaseSalary):setTooltip(_S.tooltip.staff_list.pay_rise):setSound() -- custom sound
+  self:addPanel(0, 438, 372):makeButton(0, 0, 45, 85, 9, self.fire):setTooltip(_S.tooltip.staff_list.sack):setSound() -- prevent two sound trigger
 
   -- "Arrow" to show title of doctors
   self.arrow = self:addPanel(12, 259, 397)
@@ -336,7 +336,8 @@ function UIStaffManagement:draw(canvas, x, y)
   self.portrait_back.visible = false
   -- If a staff member is selected, draw picture, skill etc
   if self.selected_staff then
-    local profile = self.staff_members[self.category][self.selected_staff].profile
+    local staff = self.staff_members[self.category][self.selected_staff]
+    local profile = staff.profile
     -- Draw the red rectangle TODO: Make a neater function in C?
     local red = canvas:mapRGB(221, 83, 0)
     local y_pos = self.selected_staff - (self.page - 1)*10
@@ -353,8 +354,10 @@ function UIStaffManagement:draw(canvas, x, y)
     profile:drawFace(canvas, x + 68, y + 377, self.face_parts)
 
     -- 10 % increase in salary or a bonus:
+    local max_salary = self.hospital.world.map.level_config.payroll.MaxSalary
+    local new_salary = math.min(math.floor(profile.wage * 1.1), max_salary)
     titles:draw(canvas, "$" .. math_floor(profile.wage*0.1), x + 377, y + 387, 45, 0)
-    titles:draw(canvas, "$" .. math_floor(profile.wage*0.1 + profile.wage), x + 377, y + 432, 45, 0)
+    titles:draw(canvas, "$" .. new_salary, x + 377, y + 432, 45, 0)
 
     -- Attention to detail
     local attention_bar_width = math_floor(profile.attention_to_detail * 40 + 0.5)
@@ -537,18 +540,25 @@ function UIStaffManagement:payBonus()
     staff:changeAttribute("happiness", 0.5)
     self.hospital:spendMoney(math_floor(staff.profile.wage*0.1), _S.transactions.personal_bonus)
     self.ui:playSound("cashreg.wav")
+    return
   end
+  self.ui:playSound("wrong2.wav")
 end
 
 function UIStaffManagement:increaseSalary()
   if self.selected_staff then
     local staff = self.staff_members[self.category][self.selected_staff]
-    staff:increaseWage(math_floor(staff.profile.wage*0.1))
+    if staff:increaseWage(math_floor(staff.profile.wage * 0.1)) then
+      self.ui:playSound("bonusal2.wav")
+      return
+    end
   end
+  self.ui:playSound("wrong2.wav")
 end
 
 function UIStaffManagement:fire()
   if self.selected_staff then
+    self.ui:playSound(self.default_button_sound)
     self.ui:addWindow(UIConfirmDialog(self.ui, false, _S.confirmation.sack_staff, --[[persistable:staff_management_confirm_sack]] function()
       local current_category = self.staff_members[self.category]
       current_category[self.selected_staff]:fire()
@@ -560,6 +570,8 @@ function UIStaffManagement:fire()
       -- Update the staff list
       self:updateStaffList(current_category[self.selected_staff])
     end)) -- End of confirmation dialog
+  else
+    self.ui:playSound("wrong2.wav")
   end
 end
 
@@ -569,6 +581,11 @@ function UIStaffManagement:close()
 end
 
 function UIStaffManagement:afterLoad(old, new)
-  UIFullscreen.afterLoad(self, old, new)
   self:registerKeyHandlers()
+
+  if old < 174 then
+    self:close()
+  end
+
+  UIFullscreen.afterLoad(self, old, new)
 end
--- a/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
@@ -67,9 +67,10 @@ function UIStaffManagement:UIStaffManagement(ui)
   -- Other buttons
   self:addPanel(0, 12, 86):makeButton(0, 0, 31, 74, 2, self.scrollUp):setTooltip(_S.tooltip.staff_list.prev_person)
   self:addPanel(0, 12, 274):makeButton(0, 0, 31, 74, 3, self.scrollDown):setTooltip(_S.tooltip.staff_list.next_person)
-  self:addPanel(0, 319, 372):makeButton(0, 0, 112, 39, 7, self.payBonus):setTooltip(_S.tooltip.staff_list.bonus):setSound() -- custom sound
-  self:addPanel(0, 319, 418):makeButton(0, 0, 112, 39, 8, self.increaseSalary):setTooltip(_S.tooltip.staff_list.pay_rise):setSound() -- custom sound
-  self:addPanel(0, 438, 372):makeButton(0, 0, 45, 85, 9, self.fire):setTooltip(_S.tooltip.staff_list.sack):setSound() -- prevent two sound trigger
+  -- Disable the default sounds on the the following three buttons as their sounds are implemented in the callback.
+  self:addPanel(0, 319, 372):makeButton(0, 0, 112, 39, 7, self.payBonus):setTooltip(_S.tooltip.staff_list.bonus):setSound()
+  self:addPanel(0, 319, 418):makeButton(0, 0, 112, 39, 8, self.increaseSalary):setTooltip(_S.tooltip.staff_list.pay_rise):setSound()
+  self:addPanel(0, 438, 372):makeButton(0, 0, 45, 85, 9, self.fire):setTooltip(_S.tooltip.staff_list.sack):setSound()
 
   -- "Arrow" to show title of doctors
   self.arrow = self:addPanel(12, 259, 397)
--- a/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
@@ -583,7 +583,7 @@ end
 function UIStaffManagement:afterLoad(old, new)
   self:registerKeyHandlers()
 
-  if old < 174 then
+  if old < 175 then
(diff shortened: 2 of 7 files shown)


CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -737,10 +737,17 @@ end
...
+  if attribute == "happiness" and self.humanoid_class then
+    if self.humanoid_class == "Receptionist" then
+      -- A receptionist is never unhappy
+      self.attributes[attribute] = 1
+      return true
+    elseif self.profile and self.profile.wage >= self.max_salary then
+      -- A maximum salaried staff member is never unhappy
+      self.attributes[attribute] = 1
+      return true
+    end
```
lewri, line 750: I'm happy to squash this down to a single `if` should that be desired. Though I didn't want to make the two "special cases" more unreadable by doing so.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -578,16 +583,20 @@ end
...
+  -- Are we already paid the maximum?
+  if self.profile.wage >= self.max_salary then return false end
+  if self.profile.wage + amount > self.max_salary then
+    -- Maximum salary hit. The staff member will never be unhappy again
+    self.profile.wage = self.max_salary
+  else
+    -- Apply new salary
+    self.profile.wage = self.profile.wage + amount
     self:changeAttribute("happiness", 0.99)
     self:setMood("pay_rise", "deactivate")
```
lewri, line 597: Wonder if we should always be running this line? I think this might've been written wrong.

lewri: Outcome: yes

lewri:
Sorry, kept finding more pitfalls then intended :(

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -559,9 +559,16 @@ end
...
+  -- Never request a raise if already at max salary
+  if self.profile.wage >= self.max_salary then
+    self.timer_until_raise = nil
+    return
+  end
   -- Check whether there is already a request for raise.
   if not self:isMoodActive("pay_rise") then
     local amount = math.floor(math.max(self.profile.wage * 1.1, (self.profile:getFairWage() + self.profile.wage) / 2) - self.profile.wage)
+    -- Don't ask over the salary cap
+    amount = self.profile.wage + amount < self.max_salary and amount or self.max_salary - self.profile.wage
```
Alberth289346, line 571:
```suggestion
    amount = math.min(self.profile.wage + amount, self.max_salary)
```

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -35,6 +35,7 @@ function Staff:Staff(...)
   self.parcelNr = 0
   self.leave_sounds = {}
   self.leave_priority = AnnouncementPriority.High
+  self.max_salary = 2000
```
Alberth289346, line 38: Question: Why is "wage" in `self.profile`, while max salary is in `self`?

ie is `self` the right spot?

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
```diff
@@ -334,9 +334,12 @@ function UIStaffManagement:draw(canvas, x, y)
   self.qualified_researcher.visible = false
   self.arrow.visible = false
   self.portrait_back.visible = false
+  self.bonus_button:setSound(self.default_button_sound)
+  self.salary_button:setSound(self.default_button_sound)
```
Alberth289346, line 338: In `draw`?  You want to set this all the time?

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
```diff
@@ -353,8 +356,12 @@ function UIStaffManagement:draw(canvas, x, y)
     profile:drawFace(canvas, x + 68, y + 377, self.face_parts)
 
     -- 10 % increase in salary or a bonus:
+    local new_salary = math.floor(profile.wage*1.1)
```
Alberth289346, line 359:
```suggestion
    local new_salary = math.min(math.floor(profile.wage * 1.1), self.max_salary)
```

Alberth289346: and drop the next line

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
```diff
@@ -543,7 +550,11 @@ end
 function UIStaffManagement:increaseSalary()
   if self.selected_staff then
     local staff = self.staff_members[self.category][self.selected_staff]
-    staff:increaseWage(math_floor(staff.profile.wage*0.1))
+    if staff:increaseWage(math_floor(staff.profile.wage*0.1)) then
```
Alberth289346, line 553:
```suggestion
    if staff:increaseWage(math_floor(staff.profile.wage * 0.1)) then
```

CorsixTH/Lua/entities/humanoids/staff.lua
Alberth289346: Shouldn't max salary depend on the job of the staff member?

Probably out of scope here but the config file also has same salary increase settings.

Alberth289346: btw I don't agree with the hard limit, it makes writing a level where eg doctors are paid huge sums of money impossible.

Not something that has to be done now though.

CorsixTH/Lua/entities/humanoid.lua
Alberth289346: You changed it in the good direction. You don't want to squeeze code if it makes understanding more difficult, eg here you would merge two independent reasons for being ultimately happy.

Having each case handled separately is much simpler to understand.

lewri:
Changes updated. I've left alone self.max_salary for the moment, as there's been some chatter about what and where it's true home is.

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
```diff
@@ -537,18 +539,27 @@ function UIStaffManagement:payBonus()
     staff:changeAttribute("happiness", 0.5)
     self.hospital:spendMoney(math_floor(staff.profile.wage*0.1), _S.transactions.personal_bonus)
     self.ui:playSound("cashreg.wav")
+  else
+    self.ui:playSound("wrong2.wav")
   end
 end
 
 function UIStaffManagement:increaseSalary()
   if self.selected_staff then
```
TheCycoONE, line 548: Function is growing - let's play wrong and exit early on no staff rather than indenting everything

CorsixTH/Lua/entities/humanoids/staff.lua
TheCycoONE: max_salary doesn't belong here. It is not an instant variable of staff and we have many staff so it will be copied around a lot. config would be better.

TheCycoONE (changes requested):
Where did the idea of max_salary come from. Can you link the issue or discussion?

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
lewri: So:
```
if not self.selected_staff then
  self.ui:playSound("wrong2.wav")
  return
end
```

CorsixTH/Lua/entities/humanoids/staff.lua
lewri: Will move to config

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
Alberth289346: yup

Alberth289346: Other option is to do the good weather case and return, then close all "if" statements, and just before the end of the function play the "wrong" sound. In that way you don't get multiple calls to the same "wong" sound at the end.

ie
```
func ...
  if self.selected staff
    ...
    if (staff: increaseWage(...) then
      play happy sound
      return
    end
  end

  play wrong sound
end
```

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -578,16 +585,23 @@ end
...
-  self.profile.wage = self.profile.wage + amount
-  self.world.ui:playSound("cashreg.wav")
-  if self.profile.wage > 2000 then -- What cap here?
-    self.profile.wage = 2000
-  else -- If the cap has been reached this member of staff won't get unhappy
-       -- ever again...
-    self:changeAttribute("happiness", 0.99)
-    self:setMood("pay_rise", "deactivate")
+  -- Are we already paid the maximum?
+  local success = true
```
Alberth289346, line 591: If you name it `wage_raised` it becomes more clear what `true` means.

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
lewri: I've used Alberth's version

CorsixTH/Lua/entities/humanoids/staff.lua
lewri: Forgot this one, it's added now.

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
TheCycoONE: A bit of it is that testing for a condition and reacting to it are separate concerns that ideally wouldn't be in the same function. Multiple `return false` are acceptable in a way that multiple playSound/return lines are not.

But thats a level of standard I don't expect.

Alberth289346:
> Where did the idea of max_salary come from. Can you link the issue or discussion?

``` text
commit 7c191d3de30b8d6e4c72421158e3dd6388987d26
Author: Edvin Linge <edvin.linge@gmail.com>
Date:   Wed Jun 22 18:38:44 2011 +0000

    Fix: Boiler breakdowns should only break boilers, not entire custom levels. Fixes issue 883.
    Fix: Don't allow bonus payments if balance is negative.
    Fix: Cap for staff salary: 2000. Fixes issue 884.
    
    Updated Norwegian translation.
```

lewri:
Make an oversight on this one -- wage increases were doubling a staff member's current wage; and then some more...

Fixed now :)

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -559,11 +558,21 @@ end
...
   -- Check whether there is already a request for raise.
   if not self:isMoodActive("pay_rise") then
     local amount = math.floor(math.max(self.profile.wage * 1.1, (self.profile:getFairWage() + self.profile.wage) / 2) - self.profile.wage)
+    -- Don't ask over the salary cap
+    if self.profile.wage + amount > max_salary then
+      amount = max_salary - self.profile.wage
+    end
     -- At least for now, staff are timid, and only ask for raises 1/5th of the time
-    if math.random(1, 5) ~= 1 or amount <= 0 then
+    if math.random(1, 5) == 1 or amount <= 0 then
```
lewri, line 575: Dang it, read the code wrong! I'll revert this line

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
```diff
@@ -67,9 +67,9 @@ function UIStaffManagement:UIStaffManagement(ui)
   -- Other buttons
   self:addPanel(0, 12, 86):makeButton(0, 0, 31, 74, 2, self.scrollUp):setTooltip(_S.tooltip.staff_list.prev_person)
   self:addPanel(0, 12, 274):makeButton(0, 0, 31, 74, 3, self.scrollDown):setTooltip(_S.tooltip.staff_list.next_person)
-  self:addPanel(0, 319, 372):makeButton(0, 0, 112, 39, 7, self.payBonus):setTooltip(_S.tooltip.staff_list.bonus)
-  self:addPanel(0, 319, 418):makeButton(0, 0, 112, 39, 8, self.increaseSalary):setTooltip(_S.tooltip.staff_list.pay_rise)
-  self:addPanel(0, 438, 372):makeButton(0, 0, 45, 85, 9, self.fire):setTooltip(_S.tooltip.staff_list.sack)
+  self:addPanel(0, 319, 372):makeButton(0, 0, 112, 39, 7, self.payBonus):setTooltip(_S.tooltip.staff_list.bonus):setSound() -- custom sound
```
TheCycoONE, line 70: Can you explain these sound comments a little more, they don't make sense to me and they aren't describing the whole line.

TheCycoONE: I see it. ... lets move and clarify the comments a little.

```
-- Disable the default sounds on the the following three buttons as their sounds are implemented in the callback.

File: CorsixTH/Lua/dialogs/fullscreen/staff_management.lua [hunk 12]
```diff
@@ -336,7 +336,8 @@ function UIStaffManagement:draw(canvas, x, y)
   self.portrait_back.visible = false
   -- If a staff member is selected, draw picture, skill etc
   if self.selected_staff then
-    local profile = self.staff_members[self.category][self.selected_staff].profile
+    local staff = self.staff_members[self.category][self.selected_staff]
```
TheCycoONE, line 339: This isn't you, but if you wanted to help the code a little, catching the do not proceed action would result in a lot less indenting.
```lua
if not self.selected_staff then
  self.ui.playSound("wrong2.wav")
  return
end

and then everything in the current condition goes left and the reader doesn't have to look very far to see that we fail and return early when there's no selected_staff and the meat of the method applies so selected_staff.

Comment (TheCycoONE) [reply CorsixTH/Lua/dialogs/fullscreen/staff_management.lua:339]: Actually, this is the draw method. What in the world are we doing playing sounds from the draw method?

File: CorsixTH/Lua/app.lua [hunk 13]
```diff
@@ -29,7 +29,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 173 -- Make couches fully passable objects
+local SAVEGAME_VERSION = 174 -- Max salary improvements
```
lewri, line 32:
```suggestion
local SAVEGAME_VERSION = 175 -- Max salary improvements
```

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
```diff
@@ -569,6 +581,11 @@ function UIStaffManagement:close()
 end
 
 function UIStaffManagement:afterLoad(old, new)
-  UIFullscreen.afterLoad(self, old, new)
   self:registerKeyHandlers()
+
+  if old < 174 then
```
lewri, line 586:
```suggestion
  if old < 175 then
```

CorsixTH/Lua/map.lua
```diff
@@ -835,4 +835,10 @@ function Map:afterLoad(old, new)
     {Value = 0}, -- I_UNEXPECTED_SWELLING
     }
   end
+  if old < 174 then
```
lewri, line 838:
```suggestion
  if old < 175 then
```

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
```diff
@@ -67,9 +67,9 @@ function UIStaffManagement:UIStaffManagement(ui)
   -- Other buttons
   self:addPanel(0, 12, 86):makeButton(0, 0, 31, 74, 2, self.scrollUp):setTooltip(_S.tooltip.staff_list.prev_person)
   self:addPanel(0, 12, 274):makeButton(0, 0, 31, 74, 3, self.scrollDown):setTooltip(_S.tooltip.staff_list.next_person)
-  self:addPanel(0, 319, 372):makeButton(0, 0, 112, 39, 7, self.payBonus):setTooltip(_S.tooltip.staff_list.bonus)
-  self:addPanel(0, 319, 418):makeButton(0, 0, 112, 39, 8, self.increaseSalary):setTooltip(_S.tooltip.staff_list.pay_rise)
-  self:addPanel(0, 438, 372):makeButton(0, 0, 45, 85, 9, self.fire):setTooltip(_S.tooltip.staff_list.sack)
+  self:addPanel(0, 319, 372):makeButton(0, 0, 112, 39, 7, self.payBonus):setTooltip(_S.tooltip.staff_list.bonus):setSound() -- custom sound
+  self:addPanel(0, 319, 418):makeButton(0, 0, 112, 39, 8, self.increaseSalary):setTooltip(_S.tooltip.staff_list.pay_rise):setSound() -- custom sound
+  self:addPanel(0, 438, 372):makeButton(0, 0, 45, 85, 9, self.fire):setTooltip(_S.tooltip.staff_list.sack):setSound() -- prevent two sound trigger
```
lewri, line 72:
```suggestion
  -- Disable the default sounds on the the following three buttons as their sounds are implemented in the callback.
  self:addPanel(0, 319, 372):makeButton(0, 0, 112, 39, 7, self.payBonus):setTooltip(_S.tooltip.staff_list.bonus):setSound()
  self:addPanel(0, 319, 418):makeButton(0, 0, 112, 39, 8, self.increaseSalary):setTooltip(_S.tooltip.staff_list.pay_rise):setSound()
  self:addPanel(0, 438, 372):makeButton(0, 0, 45, 85, 9, self.fire):setTooltip(_S.tooltip.staff_list.sack):setSound()
```

TheCycoONE (approved):

PR #2221, opened 2022-08-24
Modify pause text behaviour

### LINKED ISSUE #2220: Pause text missing on build_while_paused
** when you hit 'pause' the word 'pause' no longer shows in the bar at the bottom? - that makes it kind of hard to tell the difference between a crash and a pause at a glance.

Tigertye

_Originally posted by @Tigertye in https://github.com/CorsixTH/CorsixTH/discussions/2218_

lewri:
*Fixes #2220*
(I purposefully separated the two scenarios again, but can merge into one if should be desired)

- Restores "Pause" text when game is paused when "Build while paused" is enabled
- Pause text will be temporarily overridden if the cursor triggers dynamic text. Resets after a second or so once dynamic text no longer relevant.

Ping @Tigertye

### DIFF
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -218,10 +218,21 @@ function UIBottomPanel:drawReputationMeter(canvas, x_left, y)
   self.panel_sprites:draw(canvas, 36, x_left + math.floor(step * (self.ui.hospital.reputation - self.ui.hospital.reputation_min)), y)
 end
 
+--! Adds dynamic text to the bottom panel based on cursor position
+--!param canvas
+--!param x (num) coordinate
+--!param y (num) coordinate
 function UIBottomPanel:drawDynamicInfo(canvas, x, y)
-  if self.world:isCurrentSpeed("Pause") and not self.world.user_actions_allowed then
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -218,10 +218,21 @@ function UIBottomPanel:drawReputationMeter(canvas, x_left, y)
   self.panel_sprites:draw(canvas, 36, x_left + math.floor(step * (self.ui.hospital.reputation - self.ui.hospital.reputation_min)), y)
 end
 
+--! Adds dynamic text to the bottom panel based on cursor position
+--!param canvas
+--!param x (num) coordinate
+--!param y (num) coordinate
 function UIBottomPanel:drawDynamicInfo(canvas, x, y)
```
Alberth289346, line 225: This line and line 221 are now no longer fully correct, you also draw "paused" in some cases.

Alberth289346:
hm, too late :)

CorsixTH/Lua/dialogs/bottom_panel.lua
lewri: Guess you're right, Maybe just a line in the comments above probably would have sufficed.

PR #2224, opened 2022-08-28
Properly cap number of players while loading a map file.

### LINKED ISSUE #2223: [Bug] Map editor cannot load maps set to 4 players
### Describe the issue

When trying to load a map that has a player count of "4 players", the game will throw an error and prevent you from editing that map file:
`Error while loading game: Lua\game_ui.lua:79: Player index out of range: 1`

This doesn't seem to cause problems if the player count is set to 1, 2, or 3, nor does it occur if you change the player count away from 4 before saving the file.

This means the file is unusable, which could ruin hours of work spent on making a map.

### Steps to Reproduce

1. Open the map editor
2. Set the player count to 4 using the top menu
3. Save the map with any name you like
4. Immediately try to load that map file. You don't need to leave the editor

### Expected Behaviour

The map can be opened and edited as normal.

### Save Game

### CorsixTH Version

0.66

### Operating System

Windows 10

### Theme Hospital Version

Origin

### Gamelog.txt

_No response_

### Additional Information

_No response_

Alberth289346:
*Fixes #2223

Taking modulo doesn't work for counts.

### DIFF
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -409,10 +409,8 @@ bool level_map::load_from_th_file(const uint8_t* pData, size_t iDataLength,
   }
 
   player_count = pData[0];
-  if (player_count < 1)
-    player_count = 1;
-  if (player_count > max_player_count)
-    player_count = max_player_count;
+  if (player_count < 1) player_count = 1;
+  if (player_count > max_player_count) player_count = max_player_count;
 
   for (int i = 0; i < player_count; ++i) {
     read_tile_index(pData + camera_offset + (i * 2), initial_camera_x[i],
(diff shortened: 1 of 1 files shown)


lewri:
Works 👍

CorsixTH/Src/th_map.cpp
```diff
@@ -408,7 +408,10 @@ bool level_map::load_from_th_file(const uint8_t* pData, size_t iDataLength,
     return false;
   }
 
-  player_count = pData[0] % max_player_count;
+  player_count = pData[0];
+  if (player_count < 1) player_count = 1;
+  if (player_count > max_player_count) player_count = max_player_count;
```
TheCycoONE, line 413: Ah c++ 17 has clamp, but we're not so lucky.

PR #2232, opened 2022-09-05
Add OK to Filebrowser

### LINKED ISSUE #568: Suggest adding OK button to select save
_From [mark.l1...@hotmail.co.uk](https://code.google.com/u/117099726060447259318/) on August 19, 2013 04:27:56_

What steps will reproduce the problem? 1. Load a save
2. Go to select a save from the list
3. First it is highlighted and then it is not
4. You have to double click to select the save where as before it would highlight and select it one go

I suspect this is as a result of the recent changes to the directory browser.  I see no hardship here by having to double click now, but thought it worth mentioning.
Perhaps we should add an ok button on this screen too?

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1689_

### LINKED ISSUE #1913: File browsers should have an OK button

### LINKED ISSUE #2118: [Bug] Customize folders setting is inconsistent
#### Steps to Reproduce
1. *Open settings menu, then customize folders, then font, click on back.*

2. *Open settings menu, then customize folders, then on anything than font, click on cancel.*

3. *Observe that the fonts window has only one button while all others have to buttons and that the fonts button throws you back to settings while all the others let you stay in the customize folders dialog.*

Nothing serious, but it's very noticeable when you check a modified translation file.

lewri:
*Fixes #1913*
*Fixes #568*
*Fixes #2118*

- Adds OK button to file browsers, allowing single click of item, then pressing OK to perform the same as a double-click action
- Saving a game will also add the selected save to the textbox when an item is single clicked. OK works same as above

Removed all the `origin` stuff now from this. Just got to work on making the code better before ready.

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/file_browser.lua
+++ b/CorsixTH/Lua/dialogs/resizables/file_browser.lua
@@ -197,23 +197,47 @@ function UIFileBrowser:UIFileBrowser(ui, mode, title, vertical_size, root, show_
   -- Initialize the tree control
   self.control = FilteredTreeControl(root, 5, 35, h_size - 10, vertical_size, self.col_bg, self.col_scrollbar, true, show_dates)
     :setSelectCallback(--[[persistable:file_browser_select_callback]] function(node)
-      if node.is_valid_file and (lfs.attributes(node.path, "mode") ~= "directory") then
-        self:choiceMade(node.path)
-      end
+      if self:checkChoice(node) then self:choiceMade(node.path) end
+    end)
+    :setValueChangeCallback(--[[persistable:file_browser_textbox_callback]] function(node, label)
+      -- Update the save UI textbox, if it exists
+      if self:checkChoice(node) then self:updateTextbox(label) end
     end)
   self:addWindow(self.control)
 
-  -- Create the back button.
-  self:addBevelPanel((h_size - 160) / 2, 340, 160, 30, self.col_bg):setLabel(_S.menu_list_window.back)
-    :makeButton(0, 0, 160, 40, nil, self.buttonBack):setTooltip(_S.tooltip.menu_list_window.back)
+  -- Create the back and ok buttons.
+  local button_size = 135
+  local indent = math.floor((h_size - (2*button_size))/3)
+  self:addBevelPanel(indent, 340, button_size, 30, self.col_bg):setLabel(_S.menu_list_window.back)
+    :makeButton(0, 0, button_size, 40, nil, self.buttonBack):setTooltip(_S.tooltip.menu_list_window.back)
+
+  self:addBevelPanel(h_size - button_size - indent, 340, button_size, 30,
+  self.col_bg):setLabel(_S.menu_list_window.ok)
+    :makeButton(0, 0, button_size, 40, nil, (--[[persistable:filebrowser_ok_callback]] function()
+      if self.confirmName then
+        self:confirmName()
+      elseif self.control.selected_node then
+        local sel_node = self.control.selected_node
+        if self:checkChoice(sel_node) then self:choiceMade(sel_node.path) end
+      end
+    end)):setTooltip(_S.tooltip.menu_list_window.ok)
 end
 
--- Function stub for dialogs to override. This function is called each time a file is chosen.
+--! Function stub for dialogs to override. This function is called each time a file is chosen.
 --!param name (string) Name of the file chosen.
 function UIFileBrowser:choiceMade(name) -- luacheck: ignore 212 keep args from parent class
+end
+
+--! Function stub for dialogs with a textbox present. This will be called for 
+--! updating the textbox, override it for a proper implementation in the derived class.
+--!param label (string) Name of the file chosen
+function UIFileBrowser:updateTextbox(label) -- luacheck: ignore 212 keep args from parent class
+end
+
+function UIFileBrowser:checkChoice(node)
+  return node.is_valid_file and (lfs.attributes(node.path, "mode") ~= "directory")
 end
 
 function UIFileBrowser:buttonBack()
   self:close()
 end
-
--- a/CorsixTH/Lua/dialogs/resizables/file_browsers/save_game.lua
+++ b/CorsixTH/Lua/dialogs/resizables/file_browsers/save_game.lua
@@ -63,6 +63,14 @@ function UISaveGame:abortName()
   self.new_savegame_textbox.panel:setLabel(_S.save_game_window.new_save_game)
 end
 
+--! Updates the textbox to selected file
+--!param label (string) Selected file name
+function UISaveGame:updateTextbox(label)
+  local name = string.gsub(label, "%.sav$", "")
+  self.new_savegame_textbox.text = name
+  self.new_savegame_textbox.panel:setLabel(name)
+end
+
 --! Function called when textbox is confirmed (e.g. by pressing enter)
 function UISaveGame:confirmName()
   local filename = self.new_savegame_textbox.text
--- a/CorsixTH/Lua/dialogs/resizables/file_browsers/save_map.lua
+++ b/CorsixTH/Lua/dialogs/resizables/file_browsers/save_map.lua
@@ -63,6 +63,14 @@ function UISaveMap:abortName()
   self.new_map_textbox.panel:setLabel(_S.save_map_window.new_map)
 end
 
+--! Updates the textbox to selected file
+--!param label (string) Selected file name
+function UISaveMap:updateTextbox(label)
+  local name = string.gsub(label, "%.map$", "")
+  self.new_map_textbox.text = name
+  self.new_map_textbox.panel:setLabel(name)
+end
+
 --! Function called when textbox is confirmed (e.g. by pressing enter)
 function UISaveMap:confirmName()
   local filename = self.new_map_textbox.text
--- a/CorsixTH/Lua/dialogs/tree_ctrl.lua
+++ b/CorsixTH/Lua/dialogs/tree_ctrl.lua
@@ -599,8 +599,21 @@ function TreeControl:onMouseDown(button, x, y)
   return redraw
 end
 
+--! Function to handle (final) selection by user that needs to feed back data to
+--! another dialog.
+--!param callback (function) Code to execute on trigger
+--!return self
 function TreeControl:setSelectCallback(callback)
   self.select_callback = callback
+  return self
+end
+
+--! Function for where an action in the file tree needs to feed back data to another
+--! dialog. Its specific usage should be noted in the parent element
+--!param callback (function) Code to execute on trigger
+--!return self
+function TreeControl:setValueChangeCallback(callback)
+  self.val_change_callback = callback
   return self
 end
 
@@ -622,6 +635,9 @@ function TreeControl:onMouseUp(button, x, y)
       redraw = true
     else
       self.selected_node = node
+      if self.val_change_callback then
+        self.val_change_callback(node, node:getLabel())
+      end
       node:select()
       redraw = true
     end
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -399,12 +399,14 @@ tooltip.save_map_window = {
 menu_list_window = {
   name = "Name",
   save_date = "Modified",
+  ok = "OK",
   back = "Back",
 }
 
 tooltip.menu_list_window = {
   name = "Click here to sort the list by name",
   save_date = "Click here to sort the list by last modification date",
+  ok = "Confirm choice",
   back = "Close this window",
 }
 
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -408,7 +408,10 @@ bool level_map::load_from_th_file(const uint8_t* pData, size_t iDataLength,
     return false;
   }
 
-  player_count = pData[0] % max_player_count;
+  player_count = pData[0];
+  if (player_count < 1) player_count = 1;
+  if (player_count > max_player_count) player_count = max_player_count;
+
   for (int i = 0; i < player_count; ++i) {
     read_tile_index(pData + camera_offset + (i * 2), initial_camera_x[i],
                     initial_camera_y[i]);


CorsixTH/Lua/dialogs/resizables/file_browser.lua
```diff
@@ -197,22 +197,38 @@ function UIFileBrowser:UIFileBrowser(ui, mode, title, vertical_size, root, show_
...
+
+  self:addBevelPanel(h_size - button_size - indent, 340, button_size, 30,
+  self.col_bg):setLabel("OK")
+    :makeButton(0, 0, button_size, 40, nil, (--[[persistable:filebrowser_ok_callback]] function()
+      if (self.new_savegame_textbox and self.new_savegame_textbox.text ~= "") or (self.new_map_textbox and self.new_map_textbox.text ~= "") then
+        self:confirmName()
+      elseif self.control.selected_node then
+        self:checkChoice(self.control.selected_node)
+      end
+    end)):setTooltip("Open the selected file")
```
lewri, line 218: To be replaced with `_S` string

CorsixTH/Lua/dialogs/resizables/file_browser.lua
```diff
@@ -197,22 +197,38 @@ function UIFileBrowser:UIFileBrowser(ui, mode, title, vertical_size, root, show_
...
   -- Create the back button.
-  self:addBevelPanel((h_size - 160) / 2, 340, 160, 30, self.col_bg):setLabel(_S.menu_list_window.back)
-    :makeButton(0, 0, 160, 40, nil, self.buttonBack):setTooltip(_S.tooltip.menu_list_window.back)
+  local button_size = 135
+  local indent = math.floor((h_size - (2*button_size))/3)
+  self:addBevelPanel(indent, 340, button_size, 30, self.col_bg):setLabel(_S.menu_list_window.back)
+    :makeButton(0, 0, button_size, 40, nil, self.buttonBack):setTooltip(_S.tooltip.menu_list_window.back)
+
+  self:addBevelPanel(h_size - button_size - indent, 340, button_size, 30,
+  self.col_bg):setLabel("OK")
```
lewri, line 211: To be replaced with `_S` string

CorsixTH/Lua/dialogs/resizables/file_browser.lua
lewri: Also change to "Confirm choice" for uniformity in save/load pages

CorsixTH/Lua/dialogs/tree_ctrl.lua
```diff
@@ -622,6 +622,7 @@ function TreeControl:onMouseUp(button, x, y)
       redraw = true
     else
       self.selected_node = node
+      if self.parent then self.parent:updateTextbox(node:getLabel()) end
```
lewri, line 625: This one might be too leaky.

lewri: Yes it is :(

lewri:
Currently a bug in this one -- selecting the `Autosaves` directory on save causes a text output of `Autes`. Will look into it.

CorsixTH/Lua/dialogs/resizables/file_browser.lua
```diff
@@ -197,23 +197,38 @@ function UIFileBrowser:UIFileBrowser(ui, mode, title, vertical_size, root, show_
...
+  -- Create the back and ok buttons.
+  local button_size = 135
+  local indent = math.floor((h_size - (2*button_size))/3)
+  self:addBevelPanel(indent, 340, button_size, 30, self.col_bg):setLabel(_S.menu_list_window.back)
+    :makeButton(0, 0, button_size, 40, nil, self.buttonBack):setTooltip(_S.tooltip.menu_list_window.back)
+
+  self:addBevelPanel(h_size - button_size - indent, 340, button_size, 30,
+  self.col_bg):setLabel(_S.menu_list_window.ok)
+    :makeButton(0, 0, button_size, 40, nil, (--[[persistable:filebrowser_ok_callback]] function()
+      if (self.new_savegame_textbox and self.new_savegame_textbox.text ~= "") or (self.new_map_textbox and self.new_map_textbox.text ~= "") then
```
TheCycoONE, line 213: Ideally there would be a parameter of file_browser for `save` mode which could be checked. The `confirmName` function definition isn't in this file, so the textboxes don't otherwise leak. Empty name should (and is) handled inside the function. I don't want to see controls for specific dialogs referenced in this file.

You might do `if self.confirmName then self:confirmName() elseif ...`

CorsixTH/Lua/dialogs/tree_ctrl.lua
```diff
@@ -622,6 +622,10 @@ function TreeControl:onMouseUp(button, x, y)
       redraw = true
     else
       self.selected_node = node
+      if self.parent and
+          (self.parent.new_savegame_textbox or self.parent.new_map_textbox) then
```
TheCycoONE, line 626: no, too leaky. Notice nothing else in this widgit references parent.

Make a callback; try to clearly differentiate from select_callback; you may have to rename that one as well.

TheCycoONE (changes requested):
Current implementation is a bit spaghetti, basically you don't want widgits to be too aware of each other's internal details.

CorsixTH/Lua/dialogs/resizables/file_browsers/save_game.lua
```diff
@@ -63,6 +63,14 @@ function UISaveGame:abortName()
   self.new_savegame_textbox.panel:setLabel(_S.save_game_window.new_save_game)
 end
 
+--! Updates the textbox to selected file
+--!param node (table) Selected file
+function UISaveGame:updateTextbox(node)
+  local name = string.gsub(node,".sav","")
```
TheCycoONE, line 69: This is the source of your `Autoe` bug as you strip sav from the name. Either way you don't want to be doing this on directories or invalid nodes.

TheCycoONE:
Also don't forget to test the directory browser, some changes seem suspect for when the target is suppose to be a directory.

lewri:
PR updated now to remove as much external knowledge as possible.

CorsixTH/Lua/dialogs/resizables/file_browser.lua
```diff
@@ -197,23 +197,40 @@ function UIFileBrowser:UIFileBrowser(ui, mode, title, vertical_size, root, show_
   -- Initialize the tree control
   self.control = FilteredTreeControl(root, 5, 35, h_size - 10, vertical_size, self.col_bg, self.col_scrollbar, true, show_dates)
     :setSelectCallback(--[[persistable:file_browser_select_callback]] function(node)
-      if node.is_valid_file and (lfs.attributes(node.path, "mode") ~= "directory") then
-        self:choiceMade(node.path)
-      end
+      if self:checkChoice(node) then self:choiceMade(node.path) end
+    end)
+    :setTextboxCallback(--[[persistable:file_browser_textbox_callback]] function(node, label)
+      if self.updateTextbox and self:checkChoice(node) then self:updateTextbox(label) end
```
lewri, line 203: Another way to remove the first check is probably making an empty `updateTextbox` in `UIFileBrowser` that just returns false. But maybe this way is better?

Alberth289346: Just make the method always available so you can call it without checking first by adding `function UIFileBrowser:updateTextbox(label) error("Override me"); end` The derived classes also implement that method so at runtime the latter will be executed. `UIFileBrowser:updateTextbox()` is a placeholder, saying "This will be called for updating the textbox, override it for a proper implementation in the derived class."

Same with `confirmName` below as far as I can see. Perhaps there are more (didn't check).

If you think in classes and derived classes, it's normal that the base class provides hooks for use by the derived classes in the form of empty or basic methods that are intended to be changed.
It's the standard way to customize behavior of a class by inserting custom code in base class code.

The `error()` call in my example forces that a derived class must provide a new function for it. If you don't want to force overriding, eg when skipping the update is fine too, you can provide an empty function in the base class.  
The derived class then has the option (but not the obligation) to provide a new implementation.

Overriding methods is mostly the same as our anonymous `(function() blah blah end)` things, except classes give more structured use to such functions, as it always has a well-defined class+method name.

CorsixTH/Lua/dialogs/resizables/file_browser.lua
```diff
@@ -197,23 +197,46 @@ function UIFileBrowser:UIFileBrowser(ui, mode, title, vertical_size, root, show_
...
+  local button_size = 135
+  local indent = math.floor((h_size - (2*button_size))/3)
+  self:addBevelPanel(indent, 340, button_size, 30, self.col_bg):setLabel(_S.menu_list_window.back)
+    :makeButton(0, 0, button_size, 40, nil, self.buttonBack):setTooltip(_S.tooltip.menu_list_window.back)
+
+  self:addBevelPanel(h_size - button_size - indent, 340, button_size, 30,
+  self.col_bg):setLabel(_S.menu_list_window.ok)
+    :makeButton(0, 0, button_size, 40, nil, (--[[persistable:filebrowser_ok_callback]] function()
+      if self.confirmName then
+        self:confirmName()
```
lewri, line 217: Had to leave this one behind as is

CorsixTH/Lua/dialogs/resizables/file_browsers/save_game.lua
```diff
@@ -63,6 +63,14 @@ function UISaveGame:abortName()
   self.new_savegame_textbox.panel:setLabel(_S.save_game_window.new_save_game)
 end
 
+--! Updates the textbox to selected file
+--!param label (string) Selected file name
+function UISaveGame:updateTextbox(label)
+  local name = string.gsub(label,"%.sav","")
```
Alberth289346, line 69:
``` lua
> label = "my.save"
> string.gsub(fname,"%.sav","")
mye
```
Also, some white space after the commas looks better.

Alberth289346: Fix: `string.gsub(fname, "%.sav$", "")`

CorsixTH/Lua/dialogs/resizables/file_browsers/save_map.lua
```diff
@@ -63,6 +63,14 @@ function UISaveMap:abortName()
   self.new_map_textbox.panel:setLabel(_S.save_map_window.new_map)
 end
 
+--! Updates the textbox to selected file
+--!param label (string) Selected file name
+function UISaveMap:updateTextbox(label)
+  local name = string.gsub(label,"%.map","")
```
Alberth289346, line 69:
```suggestion
  local name = string.gsub(label, "%.map$", "")
```

CorsixTH/Lua/dialogs/tree_ctrl.lua
```diff
@@ -604,6 +604,11 @@ function TreeControl:setSelectCallback(callback)
   return self
 end
 
+function TreeControl:setTextboxCallback(callback)
```
Alberth289346, line 607: Needs luadoc

Also, the name is **very** implementation oriented (`updateTextBox` has that problem too).

That is, if I wanted to change a label instead of a textbox, or I wanted to write some data to a remote UI, the name doesn't make much sense any more while it's purpose is still valid. This is a sign that the abstraction level in the name is too low.

The way out of this is take a step back, and ask yourself: What is the purpose of this call, why do you make this call at all? Use that as input for finding a name.  
Such higher level goals in software are much more stable than current code patterns.

Higher level names also makes it easier to connect code to such higher level goals, which usually improves understanding.

Alberth289346: In all fairness, giving good names is hard to get right. It needs lots of practice and even then we make a mess sometimes.

CorsixTH/Lua/dialogs/resizables/file_browser.lua
Alberth289346: It seems 2 derived classes have `confirmName`?

Technically, if you stick the "elseif" part in the file browser base class, that gets executed for all other browsers. That seems equivalent to this solution to me, at least if it's feasible/desirable to move that code to a separate function.

Or am I missing something?

CorsixTH/Lua/dialogs/resizables/file_browsers/save_game.lua
lewri: This should be a non-issue, see screenshots.
image
image

However, I'll add it in anyway.

Alberth289346: Chances is 100% without `$`, a `gsub` does global substitution if you allow it.

CorsixTH/Lua/dialogs/tree_ctrl.lua
lewri: Maybe `TreeControl:changeValueCallback(callback)`?

CorsixTH/Lua/dialogs/resizables/file_browser.lua
```diff
@@ -197,23 +197,47 @@ function UIFileBrowser:UIFileBrowser(ui, mode, title, vertical_size, root, show_
...
 function UIFileBrowser:choiceMade(name) -- luacheck: ignore 212 keep args from parent class
 end
 
+--! Function stub for dialogs with a textbox present. This will be called for
+--! updating the textbox, override it for a proper implementation in the derived class.
+--!param label (string) Name of the file chosen
+function UIFileBrowser:updateTextbox(label) -- luacheck: ignore 212 keep args from parent class
+end
+
+function UIFileBrowser:checkChoice(node)
```
Alberth289346, line 237: This needs a luadoc to explain its purpose.

CorsixTH/Lua/dialogs/resizables/file_browsers/save_map.lua
```diff
@@ -63,6 +63,14 @@ function UISaveMap:abortName()
   self.new_map_textbox.panel:setLabel(_S.save_map_window.new_map)
 end
 
+--! Updates the textbox to selected file
+--!param label (string) Selected file name
+function UISaveMap:updateTextbox(label)
```
Alberth289346, line 68: `updateDisplayedFilename` or so?

Say what the intent is instead, the text-box is just the means to do it.

Compare with: Hammering a nail is not about hammering a nail, but about connecting some wood to each other for some purpose. The former is the means, the latter is the intent.

CorsixTH/Lua/dialogs/resizables/file_browser.lua
```diff
@@ -197,23 +197,47 @@ function UIFileBrowser:UIFileBrowser(ui, mode, title, vertical_size, root, show_
   -- Initialize the tree control
   self.control = FilteredTreeControl(root, 5, 35, h_size - 10, vertical_size, self.col_bg, self.col_scrollbar, true, show_dates)
     :setSelectCallback(--[[persistable:file_browser_select_callback]] function(node)
-      if node.is_valid_file and (lfs.attributes(node.path, "mode") ~= "directory") then
-        self:choiceMade(node.path)
-      end
+      if self:checkChoice(node) then self:choiceMade(node.path) end
+    end)
+    :setValueChangeCallback(--[[persistable:file_browser_textbox_callback]] function(node, label)
+      -- Update the save UI textbox, if it exists
```
Alberth289346, line 203: say intent, rather than means.

TheCycoONE:
looks ok to me. I wish the val_changed_ and select_ callbacks had clearer names and identical parameters, but that's a nit at this point.

Alberth289346:
I agree it's good enough for now.

PR #2234, opened 2022-09-08
Ensure handymen wait for patients to leave

### LINKED ISSUE #70: Machines can be repaired even if it is in use
If a patient is using a machine, a handyman can still repair this machine (like the scanner) and this gives a strange animation (of a not-working and a working machine in one.)

Of course the handyman should wait with repairing until the machine is not in use.

SilPho:
*Fixes #70 *

- Handymen waiting to repair a machine will now meander in the room if a patient is in the same room. This fixes the bug where the machine is repaired while it is in use.

This is the same behaviour that the original game has.

### DIFF
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -237,7 +237,8 @@ function Machine:createHandymanActions(handyman)
   end
 
   local meander_loop_callback = --[[persistable:handyman_meander_repair_loop_callback]] function()
-    if not self.user then
+    -- Wait until the machine is not in use and not about to be used
+    if not self.user and (not self:getRoom():getPatient() or self:getRoom():getPatient().isLeaving) then
       -- The machine is ready to be repaired.
       -- The following statement will finish the meander action in the handyman's
       -- action queue.
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -239,7 +239,7 @@ function Machine:createHandymanActions(handyman)
   local meander_loop_callback = --[[persistable:handyman_meander_repair_loop_callback]] function()
     -- Wait until the machine is not in use and not about to be used
     local patient = self:getRoom():getPatient()
-    if not self.user and (not patient or patient.isLeaving) then
+    if not self.user and (not patient or patient.isLeaving()) then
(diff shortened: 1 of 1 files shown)


lewri:
Does this one still allow the handyman to enter the room? I know currently the handyman can wait in the room if things are going outside of the realms of the bug.

SilPho:
Yes, this change means that they just wait inside the room for longer while the patient is being treated or diagnosed. Without this change the handyman will wait in the room if the machine is in use when he gets there, but if the patient is changing behind a screen or just hasn't quite made it to the machine yet. With the fix the handyman will wait until the patient, if any, is leaving.

Technically this would break in rooms with multiple patients, but none of those have repairable machines in them.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -237,7 +237,8 @@ function Machine:createHandymanActions(handyman)
   end
 
   local meander_loop_callback = --[[persistable:handyman_meander_repair_loop_callback]] function()
-    if not self.user then
+    -- Wait until the machine is not in use and not about to be used
+    if not self.user and (not self:getRoom():getPatient() or self:getRoom():getPatient().isLeaving) then
```
lewri, line 241: You call the two functions twice here. You'd probably be better making a `local patient = self:getRoom():getPatient()` then in the if:
` if not self.user and (not patient or patient.isLeaving) then`

SilPho: Thanks. Not sure why I didn't do that sooner

lewri (approved):
I'm ok with this

CorsixTH/Lua/entities/machine.lua
```diff
@@ -237,7 +237,9 @@ function Machine:createHandymanActions(handyman)
   end
 
   local meander_loop_callback = --[[persistable:handyman_meander_repair_loop_callback]] function()
-    if not self.user then
+    -- Wait until the machine is not in use and not about to be used
+    local patient = self:getRoom():getPatient()
+    if not self.user and (not patient or patient.isLeaving) then
```
lewri, line 242: My oversight, I think this one needs to be `patient:isLeaving()`

Calling `patient.isLeaving` just confirms that a function exists for the patient.

PR #2236, opened 2022-09-12
Refactor improveDrug

### LINKED ISSUE #230: Drug improvements seem to be slow
@Lego3 I know this has been mentioned before and I know we did a lot of research on this at the time, but even so drug improvements do seem to be a lot slower than they were in TH.
As an example I can have a large research room with four to six desks and only hire consultant researchers and the drug improvements take 4 - 5 years to get from 80% to 100% and that is with drug improvements taking most of the 100% you have to play with in the research screen.
Having looked into the code I see that there is something called gbv.ResearchPointsDivisor which does not exist in the original level files, but does in the base config and in research there is and "or 5".  So all the research points are being divided by this number.  In my custom game I reduced this to 3 in the hopes that this would speed up research, but it didn't - not noticeably anyway.  In my custom map I am half way through the fourth year and I still have five drugs at 95% and the starting point was 85%, this is with four consultant researchers and research on drugs at 85% for most of that time.  Research on actual machines seems to be quite fast by comparison even though that research percentage is low.  I don't know what the purpose of gbv.ResearchPointsDivisor was for and whether it is that that is slowing things down.  If you do get any spare time, would you be able to take a look to make sure there is not something that is making drug improvements slower please?

lewri:
*Fixes #230* (or at least should assist in it, but from my testing it seems like there may have been other changes over the years)

It's worth pointing out that we only select a new drug for improvement if we reduce the drug's cost. ~~This is probably wrong and happy to fix it if needed now (or will put in another PR)~~. Now added

- Reverts to an original-like drug improvement cycle as per @mugmuggy's findings (see 230)
- Doesn't try to fix anything that leads up to the `improveDrug` function.
- Uses proper minimum drug cost from level files
- Fixes chance of negative drug costs

Some things to note, not fixed here:
- Once all drug effectivenesses hit 100%, no more improvements can be made (this includes cost reductions)
- Changes to machine improvement cycle not implemented

### DIFF
--- a/CorsixTH/Lua/research_department.lua
+++ b/CorsixTH/Lua/research_department.lua
@@ -400,30 +400,71 @@ end
 !param drug The drug to improve, table taken from world.available_diseases
 --]]
 function ResearchDepartment:improveDrug(drug)
-  local research_info = self.research_progress[drug]
   local disease = self.hospital.disease_casebook[drug.disease.id]
+  local level_config = self.world.map.level_config
+  local min_drug_cost = level_config.gbv.MinDrugCost
 
-  -- Improving effectiveness and cost should alternate
-  if research_info.effect_imp > research_info.cost_imp then
-    -- Time to improve cost
-    disease.drug_cost = disease.drug_cost - 10
-    research_info.cost_imp = research_info.cost_imp + 1
-    if disease.cure_effectiveness == 100 then
+  -- Work out the improvement to make
+  local function decideImprovement()
+    local at_max_effectiveness = disease.cure_effectiveness >= 100
+    local at_min_cost = disease.drug_cost <= min_drug_cost
+    local improvement
+    if at_max_effectiveness and at_min_cost then -- Nothing left to improve
+      return false
+    end
+    if at_max_effectiveness then -- Only cost improvement remains
+      improvement = "cost"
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/research_department.lua
```diff
@@ -400,30 +400,69 @@ end
...
-    research_info.cost_imp = research_info.cost_imp + 1
-    if disease.cure_effectiveness == 100 then
+  -- Decrease the cost of the drug
+  -- Set to 10%, provides good balance vs number of steps on the effectiveness ladder
+  local function decreaseDrugCost()
+    local new_cost = math.max(min_drug_cost, math.floor(disease.drug_cost * 0.9))
+    disease.drug_cost = new_cost
+    if self.research_policy.drugs.current == drug then
+      self:nextResearch("drugs")
+    end
```
Alberth289346, line 434: I like these changes!

To keep this function cleanly implement "decrease costs" only, you can move these lines after the call at 462.

I agree it's weird and probably wrong, but I agree with you that it's better in a next patch.

lewri: This one has been done

Alberth289346:
LGTM

PR #2240, opened 2022-09-14
[RFC] Adjust meandering when assigned to a parcel

### LINKED ISSUE #2123: Make handyman stay in their plots
The handyman zoning doesn't seem to work properly.  When I play the "confined" scenario and set a handyman to zone 1, he keeps leaving zone 1.

SilPho:
*Fixes #2123*

I noticed during my own play that a handyman assigned to a single plot that doesn't have much corridor space will often wander off to other parts of the hospital. The reasons for this are discussed in #2123, and this PR hopes to fix it.

**Prevent handyment from wandering into other plots when assigned to a single plot**
- This change only adjusts the behaviour of the meander action, so it doesn't stop a handyman from visiting another when they need to go to a Staff Room, and it also doesn't break if the player moves the handyman around manually.
- If the handyman is not inside their assigned plot when the check is completed, the outcome is unaffected. This means a handyman in the wrong parcel will go to the correct one only when attending to a job (or finds their way back randomly or by player action). 
- In an ideal situation the pathfinder would take the assigned parcel number as input, so we wouldn't need to reject the response from it. That makes this PR more of a patchy workaround than a pure fix.

I copied the IdleAction counter behaviour from elsewhere in the meander.lua class, I'm open to adjusting that if the values need changing.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/meander.lua
+++ b/CorsixTH/Lua/humanoid_actions/meander.lua
@@ -68,6 +68,20 @@ local function meander_action_start(action, humanoid)
   local x, y = humanoid.world.pathfinder:findIdleTile(humanoid.tile_x,
       humanoid.tile_y, math.random(1, 24))
 
+  -- If the staff member is confined to one parcel, make sure they don't try to leave
+  if humanoid.parcelNr then
+    local th = TheApp.map.th
+    local current_plot = th:getCellFlags(humanoid.tile_x, humanoid.tile_y).parcelId
+    local target_plot = th:getCellFlags(x, y).parcelId
+    -- As long as they're in the right parcel, prevent them wandering outside of it
+    if (current_plot == humanoid.parcelNr and target_plot ~= humanoid.parcelNr) then
+      humanoid:queueAction(IdleAction():setCount(math.random(5, 40)))
+      humanoid:queueAction(MeanderAction())
+      humanoid:finishAction()
+      return
+    end
+  end
+
   if x == humanoid.tile_x and y == humanoid.tile_y then
     -- Nowhere to walk to - go idle instead, or go onto the next action
     if #humanoid.action_queue == 1 then


Alberth289346:
In general, don't copy code. Make it a function that you call from all spots the need it.

Copying may seem an easy way out, but it makes maintenance more difficult, since people typically change one copy but not all copies, so bugs found and fixed at one spot will re-appear at other spots.


As for the approach, what should be done is that `findIdleTile` must not look for tiles outside an assigned parcel, imho.

SilPho:
I fully agree with everything you said. I even called it out in the first post that ideally the pathfinder would know about the parcels, but I'm not confident enough to poke around the C++ depths. This PR is admittedly a patch for a relatively rare edge case.

And to be clear, the only line I copied was the length of the random idle time.
`humanoid:queueAction(IdleAction():setCount(math.random(5, 40)))`
And that's why this is an [RFC], because I'm not yet clear on whether this would be the correct course of action. If it is correct and something worth repeating then yes, it should be moved to a new function.

mugmuggy:
I would be inclined to set action.can_idle and invalidate the following if statement so it hits the procrastination logic at the end. Or achieve the same thing through a new flag, rather than having 2 lots of code doing the same thing differently.

lewri:
The issue also suggested allowing Handymen to leave the parcel/plot to rest; then return back after.

SilPho:
The PR in its current state does allow for handymen to leave the plot in order to rest because it only affects the idling action rather than all movement.
I do like mugmuggy's idea of using the same procrastination logic further down, but I would hesitate to make too many changes to this code when, as Alberth points out, the best fix would be to handle it in the pathfinder code. Once that happens, this entire PR becomes redundant and could be removed by taking out one simple if block.
If changing the pathfinder code is deemed too risky or too cumbersome then a more elegant Lua-only solution could be used.

Alberth289346:
I have a patch, not sure how to go from here.

https://github.com/CorsixTH/CorsixTH/compare/master...Alberth289346:CorsixTH:add-parcel-check-for-meandering

mugmuggy:
@Alberth289346 that patch mostly good. The default of -1, is invalid, as 0 is no assigned parcel. I haven't tested on the confined scenario, but it looks pretty good. As soon as assigned a plot, the next meander sends them that way.

lewri:
The other issue we found while looking at Albert's C++ changes were that we also idle inside staff rooms outside our plots. We should probably go with the C++ solution this time round

Alberth289346:
> @Alberth289346 that patch mostly good. The default of -1, is invalid, as 0 is no assigned parcel. I haven't tested on the confined scenario, but it looks pretty good. As soon as assigned a plot, the next meander sends them that way.

`-1` means "no parcel assigned". That is, any parcel is allowed as destination (the default behavior thus).

PR #2244, opened 2022-09-16
[RDY] Save confirm_button state when editing room

### LINKED ISSUE #2242: [Bug] Blueprints can be built out of bounds
### Describe the issue

It is possible to build a room with blueprints out of the bounds of the hospital walls.

Note: This only happens if you edit an already made room. A brand new room won't do this on first build.

ping @Xcl4m4t10n

### Steps to Reproduce

2.  Go to edit the GP office

3. Back out (X) until the blueprints phase to size the room. Drag it both in the hospital and out hospital bounds.
4. Press (X), then press (X) on the cancel warning.
5. The green tick to proceed to window/door phase is now allowed.

### Expected Behaviour

Blueprint should always be rejected when out of bounds

### Save Game

[roomoutofbounds.zip](https://github.com/CorsixTH/CorsixTH/files/9587267/roomoutofbounds.zip)

### CorsixTH Version

0.66

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

_No response_

### Additional Information

image
image

mugmuggy:
*Fixes #2242*

- Saves confirm_button state to restore when editing room

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -171,6 +171,7 @@ function UIEditRoom:cancel()
   if self.phase == "walls" then
     if self.paid then
       -- While the confirmation window is open, don't allow the player to click the confirm button.
+      local confirm_button_state = self.confirm_button.enabled
       self.confirm_button:enable(false)
       self.confirm_dialog_open = true
       -- Ask if the user really wish to sell this room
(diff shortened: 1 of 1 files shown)


lewri:
Probably too complicated for an afterLoad, but I'm guessing we just can't fix this if someone saved in a state with edit room blueprint phase open.

mugmuggy:
I *think* we'd have to store the state elsewhere as the upvalues for the persistable wouldn't have the variable being used. There are other issues around blueprint states being visible, so yeah its annoying if you are relying on it an autosave.

lewri (approved):
Fine with me given comments. We'd hope that any affected saves don't do a silly and actually build outside of the bounds.

PR #2245, opened 2022-09-17
Fix meander-code indentation (found by Lewri)

Alberth289346:
* Fixes an indentation problem found by @lewri 

- Makes code more readable.


lewri:
Thanks ;)

Issue #1991, opened 2021-07-05
Heatmaps applying to lakes


lewri:
Reported in the Discord (and I've seen it too) that for some reason the radiator heatmaps are getting applied to the lakes in the relevant menu. This shouldn't happen. Pretty sure it's come up somewhere between ``0.64`` and ``0.65``

lewri:
![image](https://user-images.githubusercontent.com/20030128/124458255-a8a8fa00-dd84-11eb-9543-ada0aa25fec9.png)

lewri:
Have confirmed this one goes back to 0.62 at least.

TheCycoONE:
I don't remember this ever not being an issue.

SilPho:
At first I thought this might be a problem with the town map just painting certain tiles incorrectly, but I think it might actually be something to do with the map files themselves, or at least whatever designates a certain tile as passable or not, since enabling the "flags" debug option shows something like this:
![Pond flags](https://user-images.githubusercontent.com/22282491/185787666-930b8cb8-813b-4b25-8539-ce401c8a52f0.png)
Those are the central ponds in Level 7 (Greenpool).

lewri:
In original game level 7, are the ponds still drawn on the town map?

lewri:
So I checked Level 1 in DOSBOX.
It looks like they shouldn't be drawn on the map at all...

![image](https://user-images.githubusercontent.com/20030128/185789799-9d0cfd7c-f5a2-4d72-85cf-0db262dd3ac7.png)

lewri:
And level 7:
![image](https://user-images.githubusercontent.com/20030128/185789901-7004f784-ce3d-4ba8-9b77-b7b36e5db451.png)


So in reality, we shouldn't be drawing these pond/water tiles on our maps at all.

Alberth289346:
A few other options are

- Make these fixes in Lua (so there is an afterload)
- Paint temperatures only for parcels owned by the selected player (current cpp code uses hard-coded 1 as "the player", but uses that only for showing buyable parcel, it seems).

ARGAMX:
[2834_Heatmaps_on_lakes_1.sav.zip](https://github.com/user-attachments/files/19929399/2834_Heatmaps_on_lakes_1.sav.zip)

[2834_Heatmaps_on_lakes_2.sav.zip](https://github.com/user-attachments/files/19929400/2834_Heatmaps_on_lakes_2.sav.zip)

v0.68.0 :

[...]

PR #2246, opened 2022-09-18
[WIP] Add parcel check for meandering

### LINKED ISSUE #2123: Make handyman stay in their plots

Alberth289346:
*Fixes #2123 

- Extend CPP idle tile finding to find tiles only in a provided parcel is requested
- Extend handyman meandering to use that functionality for finding idle tiles.

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1636,6 +1636,11 @@ function World:getPath(x, y, dest_x, dest_y)
   return self.pathfinder:findPath(x, y, dest_x, dest_y)
 end
 
+-- Find an tile for idling.
+-- !param x X coordinate of the queue position.
+-- !param y Y coordinate of the queue position.
+-- !param idx Number of standing persons in the queue before the person asking for an idle tile.
+-- !return Position of an idle tile if it exists.
 function World:getIdleTile(x, y, idx)
   local cache_idx = (y - 1) * self.map.width + x
   local cache = self.idle_cache[cache_idx]
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -116,7 +116,7 @@ function World:World(app)
 
   -- If set, do not create salary raise requests.
   self.debug_disable_salary_raise = self.free_build_mode
-  self.idle_cache = {}
+  self.idle_cache = {} -- Cached queue standing positions for all queues.
   -- List of which goal criterion means what, and what number the corresponding icon has.
   self.level_criteria = local_criteria_variable
   self.delayed_map_objects = {} -- Initial objects in the map for parcels without owner.
--- a/CorsixTH/Src/th_pathfind.cpp
+++ b/CorsixTH/Src/th_pathfind.cpp
@@ -271,7 +271,7 @@ bool idle_tile_finder::try_node(path_node* pNode, map_tile_flags flags,
 }
 
 bool idle_tile_finder::find_idle_tile(const level_map* pMap, int iStartX,
-                                      int iStartY, int iN) {
+                                      int iStartY, int iN, int parcelId) {
   if (pMap == nullptr) {
     pMap = parent->default_map;
   }
@@ -290,9 +290,15 @@ bool idle_tile_finder::find_idle_tile(const level_map* pMap, int iStartX,
 
   while (true) {
     pNode->visited = true;
-    map_tile_flags flags = pMap->get_tile_unchecked(pNode->x, pNode->y)->flags;
 
-    if (!flags.do_not_idle && flags.passable && flags.hospital) {
+    // Check whether pNode is a feasible destination.
+    const map_tile* node_tile = pMap->get_tile_unchecked(pNode->x, pNode->y);
+    map_tile_flags flags = node_tile->flags;
+    bool correct_parcel = parcelId < 0 || parcelId == node_tile->iParcelId;
+    if (!flags.do_not_idle && flags.passable && flags.hospital &&
+        correct_parcel) {
+      // Try to avoid re-using a queue-ing position of a humanoid earlier in the
+      // queue.
       if (iN == 0) {
         parent->destination = pNode;
         return true;
--- a/CorsixTH/Src/th_lua_map.cpp
+++ b/CorsixTH/Src/th_lua_map.cpp
@@ -940,9 +940,12 @@ int l_path_idle(lua_State* L) {
   pathfinder* pPathfinder = luaT_testuserdata<pathfinder>(L);
 
   bool found = pPathfinder->find_idle_tile(
-      nullptr, static_cast<int>(luaL_checkinteger(L, 2)) - 1,
-      static_cast<int>(luaL_checkinteger(L, 3)) - 1,
-      static_cast<int>(luaL_optinteger(L, 4, 0)));
+      nullptr,                                        // Default map.
+      static_cast<int>(luaL_checkinteger(L, 2)) - 1,  // X coordinate queue.
+      static_cast<int>(luaL_checkinteger(L, 3)) - 1,  // Y coordinate queue.
+      static_cast<int>(
+          luaL_optinteger(L, 4, 0)),  // Number of standing persons in queue.
+      static_cast<int>(luaL_optinteger(L, 5, -1)));  // Parcel number.
   if (!found) {
     return 0;
   }
--- a/CorsixTH/Src/th_pathfind.h
+++ b/CorsixTH/Src/th_pathfind.h
@@ -166,12 +166,22 @@ class idle_tile_finder : public abstract_pathfinder {
   bool try_node(path_node* pNode, map_tile_flags flags, path_node* pNeighbour,
                 travel_direction direction) override;
 
-  bool find_idle_tile(const level_map* pMap, int iStartX, int iStartY, int iN);
+  //| Find a tile for idling.
+  /*!
+      @param pMap Map to search.
+      @param iStartX X coordinate of the queue start position.
+      @param iStartY Y coordinate of the queue start position.
+      @param iN Number of persons standing in the queue before the humanoid.
+      @param parcelId Id of the parcel with valid destinations, use \c -1 for
+     any parcel.
+   */
+  bool find_idle_tile(const level_map* pMap, int iStartX, int iStartY, int iN,
+                      int parcelId);
 
   path_node* best_next_node;
   double best_distance;
-  int start_x;  ///< X coordinate of the start position.
-  int start_y;  ///< Y coordinate of the start position.
+  int start_x;  ///< X coordinate of the queue position.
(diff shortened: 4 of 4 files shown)


CorsixTH/Lua/humanoid_actions/meander.lua
```diff
@@ -65,8 +65,17 @@ local function meander_action_start(action, humanoid)
       humanoid:setDynamicInfoText(_S.dynamic_info.staff.actions.wandering)
     end
   end
+
+  -- Handymen may have an assigned parcel, but they are free to visit any staff
+  -- room to rest
+  local meander_parcel
+  if humanoid.humanoid_class == "Handyman" and humanoid.parcelNr > 0 and
```
SilPho, line 72: Handymen are the only staff that can have an assigned parcel, so do we need to check for the Handyman class? If we decided to allow other staff to have assigned parcels then we wouldn't need this check then either.

Alberth289346: The `Staff` baseclass has the `parcelNr` variable, so all staff have that variable. I don't know what they do with it at all, so the `Handyman` check is mostly about avoiding accidents.

Also I didn't consider what is needed for other staff to work within a parcel, eg can you build a room on 2 parcels? Thus, it seemed that a different patch would be better to solve those problems.

SilPho: I'm happy with that.

For future reference, I don't know how often it can be done, but it is possible to build a room across multiple parcels. Campaign level 9 (Eastville Hospital) has a corridor between parcels 4 and 8 that is wide enough for a 4-tile room to be placed. I don't think it changes anything for this patch though.

CorsixTH/Src/th_pathfind.cpp
```diff
@@ -290,9 +290,15 @@ bool idle_tile_finder::find_idle_tile(const level_map* pMap, int iStartX,
 
   while (true) {
     pNode->visited = true;
-    map_tile_flags flags = pMap->get_tile_unchecked(pNode->x, pNode->y)->flags;
 
-    if (!flags.do_not_idle && flags.passable && flags.hospital) {
+    // Check whether pNode is a feasible destination.
+    const map_tile* node_tile = pMap->get_tile_unchecked(pNode->x, pNode->y);
+    map_tile_flags flags = node_tile->flags;
+    bool correct_parcel = parcelId < 0 || parcelId == node_tile->iParcelId;
```
mugmuggy, line 297: This is where I meant the -1 and 0 'issue' is. I guess its the same reason, that the meander changes are checking for a parcelNr > 0, whereas I would have thought we would have passed it through by default.

CorsixTH/Lua/humanoid_actions/meander.lua
mugmuggy: Changing isResting to a not in room condition would handle the room on parcel borders condition also.

CorsixTH/Src/th_pathfind.cpp
Alberth289346: A long tale to say I should exclude parcel `0` :)

Thanks mugmuggy for spotting this.

CorsixTH/Lua/humanoid_actions/meander.lua
lewri: Can you elaborate on what you mean mugmuggy? If you mean checking if handyman is in a specific room type we did think about that. But it's not that best way iirc.

mugmuggy: I just mean 'not humanoid:getRoom()'. isResting is just a proxy for staff room, but if a handyman meanders in a room, which does happen from time to time, you may actually desire them to stay in the room, ie machine to be repaired is currently in use. Normally I don't think this occurs on purpose, it just happens, as most of the time there is a check if they are in a room, queues leave action followed by meander. I just can't find a reference for meander waiting for machine repair

lewri: Mugmuggy, function changed. I assume that's what was wanted?

Alberth289346: Isn't room `0` the corridor? If so, it won't work:
``` lua
> if 0 then print"yes" end
yes
```
ie `0` is not `false`.

Alberth289346: Ah, nevermind, it returns `nil` if outside a room.

mugmuggy: Suggested vs wanted. On the times they do meander whilst waiting for the machine to become free, they should meander just fine anywhere in the room. As for finding and allocating work, I assume the use tiles and the objects itself determine who gets called to the room. Its already possible to have multiple handymen arrive if there are multiple calls to the one room, just it may look weird when a handyman doesn't repair when room is built on a border.

TheCycoONE:
Bump the api version for this PR too

CorsixTH/Src/th_pathfind.cpp
```diff
@@ -290,9 +290,15 @@ bool idle_tile_finder::find_idle_tile(const level_map* pMap, int iStartX,
...
   while (true) {
     pNode->visited = true;
-    map_tile_flags flags = pMap->get_tile_unchecked(pNode->x, pNode->y)->flags;
 
-    if (!flags.do_not_idle && flags.passable && flags.hospital) {
+    // Check whether pNode is a feasible destination.
+    const map_tile* node_tile = pMap->get_tile_unchecked(pNode->x, pNode->y);
+    map_tile_flags flags = node_tile->flags;
+
+    bool correct_parcel = parcelId <= 0 || parcelId == node_tile->iParcelId;
```
TheCycoONE, line 298: Confused about whether we're suppose to be respecting parcel 0 or not.

Alberth289346: Originally, I thought parcel `0` was the first used parcel, so I picked `-1` for not having a parcel requirement. Afterwards, mugmuggy wanted parcel `0` out, as it turned out to be outside, which is not where staff idles. Changing `< 0` to `<= 0` was the least amount of work at that point, so that's why we have the current implementation.

We do rely on Lua doing sane things here, as we neither check existence of a parcel number in CPP nor do we check if the given parcel is owned by the hospital of the staff.

We seem to have 3 useful solutions.
- Keep it as-is, with the advantage that the "lack of parcel requirement" is very clearly not a parcel.
- Also allow parcel `0` requirement so if we ever want a path to outside and current methods are not sufficient, it can be computed (ie revert back to `< 0`).
- Use `0` as lack of parcel since it's unlikely we'll ever need it to navigate to the outdoor-parcel. It's somewhat fuzzy that parcel `0` then has two meanings depending on what code you are looking at.

I have no strong feelings about any of the solutions, although the 3rd one looks less nice to me.

mugmuggy: The parcel < 0 and parcel <= 0 thing came because we had code to not perform it when it was 0, if you remove that check you get incorrect behaviour as its false in certain circumstances (0 < 0)' || '0 == n (as n > 0 inside the hospital typically)):
Line 72 in meander.lua to protect the cpp call from scenarios that introduce that false.
```lua
if humanoid.humanoid_class == "Handyman" and humanoid.parcelNr > 0 and
```
if you ignore the room check on that if, the meander can pass the parcelNr as the last argument.
```lua
      @param parcelId Id of the parcel with valid destinations, use \c -1 for
     any parcel.
```
So we use 0 for any plot on the staff member and -1 in cpp (new) as Albert points out but just to correct Albert on the outside parcel, as I pointed out, hospital tiles outside the hospital walls as parcelNr is 0, pull the handyman outside when the parcelNr check is removed (https://github.com/CorsixTH/CorsixTH/issues/2014#issuecomment-887968306 for a reference of the map issue and also #1991 ). They just don't meander outside anywhere. Where the map has no external hospital tiles, and the check is removed and you place a handyman (say L2 map) it crashes as there is no valid destination tile.

TheCycoONE: AFAIK using 0 for not in hospital is problematic, many maps stretch the parcel numbers well outside of the bounds of the hospital, and doing so has a real world use - to make the cost of some parcels larger than what would normally be calculated from the hospital.

I don't think there's any valid reason now or ever to query for parcel 0 when path finding, and suggest changing the default to 0 in cpp?

CorsixTH/Lua/world.lua
```diff
@@ -1636,7 +1636,13 @@ function World:getPath(x, y, dest_x, dest_y)
...
 end
 
-function World:getIdleTile(x, y, idx)
+-- Find an tile for idling.
+-- !param x X coordinate of the queue position.
+-- !param y Y coordinate of the queue position.
+-- !param idx Randomization factor for the idle tile, returns Nth candidate tile.
+-- !param parcel_id Optional parcel of the indoor returned idling tile.
+-- !return Position of an idle tile if it exists.
+function World:getIdleTile(x, y, idx, parcel_id)
```
TheCycoONE, line 1645: It `might` even be better to do `parcel_id = parcel_id or 0` here to make it clear nil parcel_ids are not a real thing?

CorsixTH/Src/th_pathfind.cpp
Alberth289346: Parcel 0 it is.

I did wonder about modifying maps after loading to clean up errors, but never tried to implement it.

As for parcel areas outside the hospital, could be nice to have an outdoor ward :p  
A more real use I guess is to artificially increase the parcel cost, which is arguably better solved by adding such a parameter to the level file of course.

CorsixTH/Lua/world.lua
```diff
@@ -1647,7 +1653,7 @@ function World:getIdleTile(x, y, idx)
     self.idle_cache[cache_idx] = cache
   end
   if not cache.x[idx] then
-    local ix, iy = self.pathfinder:findIdleTile(x, y, idx)
+    local ix, iy = self.pathfinder:findIdleTile(x, y, idx, parcel_id or 0)
```
lewri, line 1656: Is this a standard practice to put the `x or y` in the function parameter? Just asking

Alberth289346: If a function `f` wants value `x+1` rather than `x`, you write `f(x + 1)`.

If a function `g` wants value `x` or `0` if `x` is `nil` or `false` I would write `g(x or 0)`

In other words, `+` and `or` are the same thing to me, they compute a value desired by the function.
Why do you think `or` is different?

TheCycoONE:
@Alberth289346 any desire to clean the commit history, or shall we squash and merge and call it a day? I don't have a personal preference.

Alberth289346:
Just squash it.

PR #2257, opened 2022-09-28
Remove destroyed patients from epidemics

### LINKED ISSUE #2256: [Bug] Serious Radiation epidemic with no serious radiation patients
### Describe the issue

Game gave me a serious radiation epidemic with no radiation patients to cure, saves "bug3", "bug2" and "bug1" are the progression of months following the issue, save "prebug" is the month before the issue, and "bugtrigger" has the issue triggered, but the epidemic hasn't been accepted yet.

### Steps to Reproduce

1. *Load save games attached to this ticket*

### Expected Behaviour

Give me patients to cure

### Save Game

[Saves.tar.gz](https://github.com/CorsixTH/CorsixTH/files/9660715/Saves.tar.gz)

### CorsixTH Version

0.66

### Operating System

Win10

### Theme Hospital Version

GOG.com

### Gamelog.txt

_No response_

### Additional Information

_No response_

TheCycoONE:
*Fixes #2256*

- Removes patients from the infected list of epidemics if they are despawned. Prevents future_epidemics from being triggered with no active contagious patients and makes result fair.

### DIFF
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -279,7 +279,7 @@ function Epidemic:checkPatientsForRemoval()
   for i = #self.infected_patients, 1, -1 do
     local infected_patient = self.infected_patients[i]
     if (not self.coverup_in_progress and infected_patient.going_home) or
-        infected_patient.dead then
+        infected_patient.dead or infected_patient.tile_x == nil then
       table.remove(self.infected_patients,i)
     end
   end


lewri:
I would suggest we use mugmuggy's alternative of marking the patient as dead at point of autopsy rather than checking a tile (probably is cleaner overall).

You could still do an afterLoad as an ad-hoc cleanup then for autopsied epidemic patients using an empty tile?

PR #2264, opened 2022-10-03
update language files - simplified and traditional Chineses

lokpro:
- add new texts
- add fax.vip_visit_result compatibility mapping
- revise old texts


lewri (approved):
Thanks for also tidying up some of the arrays

PR #2265, opened 2022-10-03
 Handyman and plant Monthly advice checks

lewri:
*Fixes #*

- Prevents 'handyman to water plants' message triggering on every placement of plant
- Now runs a monthly advice check for plants without a handyman
- Reorganised and documented the `dailyAdviceChecks` function to be in chronological order.
- Prevents no radiator check from constantly spamming the player.

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1680,7 +1680,7 @@ function Hospital:objectPlaced(entity, id)
   end
 
   if id == "plant" then
-    self:msgPlant()
+    self:advisePlants(true)
     return
   end
 
@@ -2381,7 +2381,7 @@ function Hospital:msgMultiReceptionDesks()
 end
 
 --! Give advice to the user about maintenance of plants.
-function Hospital:msgPlant()
+function Hospital:advisePlants()
 end
 
 --! Show the 'Gates to hell' animation.
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -47,114 +47,14 @@ function PlayerHospital:PlayerHospital(world, avail_rooms, name)
 end
 
 --! Give advice to the player at the end of a day.
+--! Note: For patient/staff warmth checks, advice is only on one or the other per month
 function PlayerHospital:dailyAdviceChecks()
   local current_date = self.world:date()
   local day = current_date:dayOfMonth()
-  local current_month = current_date:monthOfYear()
 
   -- Hold any advice back until the game has somewhat started.
   if current_date < Date(1, 5) then
     return
-  end
-
-  -- Check for advice on money.
-  -- This must occur after monthly maintenance and salary costs
-  -- or it may give invalid advice
-  if day == 1 then
-    if not self.world.free_build_mode then
-      if self.balance < 2000 and self.balance >= -500 then
-        local cashlow_advice = {
-          _A.warnings.money_low, _A.warnings.money_very_low_take_loan,
-          _A.warnings.cash_low_consider_loan,
-        }
-        self:giveAdvice(cashlow_advice)
-
-      elseif self.balance < -2000 and current_month > 8 then
-        -- TODO: Ideally this should be linked to the lose criteria for balance.
-        self:giveAdvice({_A.warnings.bankruptcy_imminent})
-
-      elseif self.balance > 6000 and self.loan > 0 then
-        self:giveAdvice({_A.warnings.pay_back_loan})
-      end
-    end
-  end
-
-  -- Warn about lack of a staff room.
-  if day == 3 and self:countRoomOfType("staff_room", 1) == 0 then
-    local staffroom_advice = {
-      _A.warnings.build_staffroom, _A.warnings.need_staffroom,
-      _A.warnings.staff_overworked, _A.warnings.staff_tired,
-    }
-    self:giveAdvice(staffroom_advice)
-  end
-
-  -- Warn about lack of toilets.
-  if day == 8 and self:countRoomOfType("toilets", 1) == 0 then
-    local toilet_advice = {
-      _A.warnings.need_toilets, _A.warnings.build_toilets,
-      _A.warnings.build_toilet_now,
-    }
-    self:giveAdvice(toilet_advice)
-  end
-
-  -- Make players more aware of the need for radiators
-  if self:countRadiators() == 0 then
-    self:giveAdvice({_A.information.initial_general_advice.place_radiators})
-  end
-
-  -- Verify patients well-being with respect to room temperature.
-  if day == 15 and not self.adviser_data.temperature_advice
-      and not self.heating.heating_broke then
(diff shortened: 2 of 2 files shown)


CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -97,9 +130,44 @@ function PlayerHospital:dailyAdviceChecks()
...
+        }
+        self:giveAdvice(bench_advice)
+
+      elseif ratio > 0.9 then
+        -- Praise having enough well placed seats about once a year.
+        local bench_advice = {
+          _A.praise.many_benches, _A.praise.plenty_of_benches,
+          _A.praise.few_have_to_stand,
+        }
+        self:giveAdvice(bench_advice, 1/12)
```
Alberth289346, line 168:
```suggestion
        self:giveAdvice(bench_advice, 1 / 12)
```

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -307,10 +327,12 @@ function PlayerHospital:msgMultiReceptionDesks()
 end
 
 --! Give advice to the user about maintenance of plants.
-function PlayerHospital:msgPlant()
+--!param placed (bool) If a plant was placed
+function PlayerHospital:msgPlant(placed)
```
Alberth289346, line 331: Function name doesn't cover what it actually does.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -125,6 +193,11 @@ function PlayerHospital:dailyAdviceChecks()
     end
   end
 
+  -- Warn player of no handymen if there are plants
+  if day == 18 and self:countPlants() > 0 then
+    self:msgPlant(false)
+  end
```
Alberth289346, line 199: Add a no-handyman check.

EDIT: I wrote the above line as the function name doesn't imply it takes handy men into account in the message.
change the function name instead.

The name of the function should cover what it does so one can judge correctness of a line with a function call without knowing the specifics of the function internals.

CorsixTH/Lua/hospital.lua
```diff
@@ -1680,7 +1680,7 @@ function Hospital:objectPlaced(entity, id)
   end
 
   if id == "plant" then
-    self:msgPlant()
+    self:msgPlant(true)
```
Alberth289346, line 1683: Arguably, you don't even need the argument.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -48,6 +48,20 @@ end
...
+    Day 3: Staff rooms
+    Day 5: Radiators
+    Day 8: Toilets
+    Day 10: Long queues
+    Day 12: Benches
+    Day 15: Heating (patients)
+    Day 18: Plants (No handymen)
+    Day 20: Heating (staff) NB: only patient or staff executes per month (if any)
+    Day 24: Drinks machines
+  --]]
```
Alberth289346, line 63: Busy little guy that advisor eh?

Did we ensure the safety of the advisor sufficiently against angry clueless users that got mad due to all its messages spam?

CorsixTH/Lua/hospitals/player_hospital.lua
lewri: I think there might be a few of these with spaces needed that happy to fix anyway

CorsixTH/Lua/hospitals/player_hospital.lua
lewri: Probably not me, but any suggestions? We have several of these functions with this naming scheme currently, so maybe it isn't worth changing just this one right now?

CorsixTH/Lua/hospitals/player_hospital.lua
lewri: We do not, but it seems if you play the game right they shouldn't all trigger.
You could argue some could be set to execute a % of the time using the `rand_frac` parameter on `giveAdvice`.

I'm not sure how consistent our amount of announcements compares to TH.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -307,10 +327,12 @@ function PlayerHospital:msgMultiReceptionDesks()
...
 
 --! Give advice to the user about maintenance of plants.
-function PlayerHospital:msgPlant()
+--!param placed (bool) If a plant was placed
+function PlayerHospital:msgPlant(placed)
   local num_handyman = self:countStaffOfCategory("Handyman", 1)
+  local plants = placed and self:countPlants() or 1
 
-  if num_handyman == 0 then
+  if num_handyman == 0 and plants == 1 then
```
Alberth289346, line 335: Instead of this cunning `1` plant trick, maybe

``` lua
if self:countStaffOfCategory("Handyman", 1) > 0 then return end

local num_plants = self:countPlants()
if num_plants == 0 or (placed and num_plants > 1) then return end
self:giveAdvice({_A.staff_advice.need_handyman_plants})
```

CorsixTH/Lua/hospitals/player_hospital.lua
TheCycoONE: advisePlantMaintenance or alternately warnPlantMaintenance

The argument can be assumePlants
```lua
local hasPlants = assumePlants or (self:countPlants() > 1)```

**Review by TheCycoONE — APPROVED** [review]

File: CorsixTH/Lua/hospitals/player_hospital.lua (hunk 2 above)
Comment (lewri) [reply CorsixTH/Lua/hospitals/player_hospital.lua:331]: I've gone with Alberth's suggested alternative and renamed it `advisePlants`

**Comment by lewri** [comment]
Comments incorperated

File: CorsixTH/Lua/hospitals/player_hospital.lua [hunk 7]
```diff
@@ -47,175 +47,234 @@ function PlayerHospital:PlayerHospital(world, avail_rooms, name)
...
   end
-
   -- Make players more aware of the need for radiators
-  if self:countRadiators() == 0 then
+  if day == 5 and self:countRadiators() == 0 then
     self:giveAdvice({_A.information.initial_general_advice.place_radiators})
   end
+  if day == 8 then
+    self:_adviseToilets()
+  end
```
TheCycoONE, line 86: cool. This reads much better. `elseif` probably isn't important except as a microoptimization but I would have expected it.

tobylane (approved):

CorsixTH/Lua/hospitals/player_hospital.lua
lewri: Finally done this one, sorry.

TheCycoONE:
@lewri did you plan to go any further with this before I merge it?

lewri:
> @lewri did you plan to go any further with this before I merge it?

If you want the `else if`changes made I'll do them, just didn't think it was blocking currently.

TheCycoONE:
It's not, just waiting on your intent

PR #2268, opened 2022-10-07
Add demo gameplay movie

lewri:
- Adds `ATTRACT.SMK` video to game when idling in main menu
- Idle currently ~30s
- A slightly placeholder too for when we're able to replace with our own movies (e.g. wolter's one as discussed on Discord/Matrix)

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/main_menu.lua
+++ b/CorsixTH/Lua/dialogs/resizables/main_menu.lua
@@ -96,6 +96,12 @@ function UIMainMenu:draw(canvas, x, y)
     ly = ly - 15
   end
   self.label_font:draw(canvas, _S.main_menu.version .. self.version_number, x + 5, ly, 190, 0, "right")
+  TheApp:idle()
+end
+
+function UIMainMenu:onMouseMove(x, y, dx, dy)
+  TheApp:resetIdle()
+  UIResizable.onMouseMove(self, x, y, dx, dy)
 end
 
 function UIMainMenu:buttonNewGame()
--- a/CorsixTH/Lua/movie_player.lua
+++ b/CorsixTH/Lua/movie_player.lua
@@ -96,12 +96,14 @@ function MoviePlayer:MoviePlayer(app, audio, video)
   self.lose_movies = {}
   self.advance_movies = {}
   self.intro_movie = nil
+  self.demo_movie = nil
   self.win_movie = nil
   self.can_skip = true
   self.wait_for_stop = false
   self.wait_for_over = false
 end
 
+--! Initialises the different movies used in the game
 function MoviePlayer:init()
   self.moviePlayer = TH.moviePlayer()
   self.moviePlayer:setRenderer(self.video)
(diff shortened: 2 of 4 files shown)


CorsixTH/Lua/dialogs/resizables/main_menu.lua
```diff
@@ -96,6 +96,12 @@ function UIMainMenu:draw(canvas, x, y)
     ly = ly - 15
   end
   self.label_font:draw(canvas, _S.main_menu.version .. self.version_number, x + 5, ly, 190, 0, "right")
+  TheApp:idle()
```
lewri, line 99: This could potentially be inside `App:onTick()`, but then that function call happens every tick the game is open.

CorsixTH/Lua/movie_player.lua
```diff
@@ -141,6 +144,10 @@ function MoviePlayer:playIntro(callback_after_movie)
   self:playMovie(self.intro_movie, false, true, callback_after_movie)
 end
 
+function MoviePlayer:playDemoMovie(callback_after_movie)
+  self:playMovie(self.demo_movie, false, true, callback_after_movie)
+end
+
```
lewri, line 150: Happy to also add some doc to this file, which I neglected on this commit.

TheCycoONE:
Looks good, I want to have a closer look at some of the idle interaction but I think this passes.

Alberth289346:
In a setup without movie-support in the program, enabling the intro movie makes the mouse cursor jumps to the top-left corner every ~30 seconds. It's not really there though, a slight move with the mouse makes it jump back to its previous position.

By the looks of it, it's starting the movie which ends immediately. Perhaps rebuilding the UI has that side-effect?

Alberth289346:
Global movie control setting is also not taken into account

lewri:
> Global movie control setting is not taken into account

This is handled inside `MoviePlayer:playMovie(...)` already thankfully

lewri:
> In a setup without movie-support in the program, enabling the intro movie makes the mouse cursor jumps to the top-left corner every ~30 seconds. It's not really there though, a slight move with the mouse makes it jump back to its previous position.

I think this is due to me calling `self:loadMainMenu()` at the end of the movie, since it resets the UI. Turns out we don't need it since the main menu remains open while the movie plays (not sure if intentional).

TheCycoONE:
Leaving the UIMainMenu and removing the line you added to rebuild it is indeed the right thing to do. You have the wrong mental model, UI elements are redrawn every frame. When there's a movie the movie is drawn instead.

tobylane:
Can the aspect ratio be set? The intro movie has bars on the side to make it 4:3, but the demo movie has bars on the top and bottom to make it near 2:1. and unreadable.
Screen 2022-10-09 at 13 51 55

lewri:
Good question.... Was that the original aspect of attract.smk... I always thought it was that bad

TheCycoONE:
attract.smk was recorded that poorly. It is possible to distort the aspect ratio to compensate but it won't fix the display artifacts. If anything it would probably be even worse.

lewri:
Also to note this patch is also a slight placeholder until we can replace it with our own/wolter's video. Currently, `h264` doesn't work in the code's current state (my preference, though to record it here, cyco's is `vp8/vp9`.
And tbh I think CorsixTH sometimes is a bit too glitchy in anims to remake something like attract.smk --- yet.

PR #2273, opened 2022-10-19
Init campaigns directory

lewri:
Segment of #2209 (aka the easy bit)

- Initialises the campaigns directory on launch and adds it as a config line.
- Encompasses part of the prepwork for CommunityContent.


lewri:
This line probably should've also come across:
https://github.com/CorsixTH/CorsixTH/blob/1ee471158613ed40537e49445907fd67d107412d/CorsixTH/Lua/app.lua#L648

CorsixTH/Lua/app.lua
```diff
@@ -396,6 +398,28 @@ function App:initUserLevelDir()
...
+    if not lfs.mkdir(self.user_campaign_dir) then
+      print("Notice: Campaign directory does not exist and could not be created.")
+      return false
+    end
+  end
+  if self.user_campaign_dir:sub(-1, -1) ~= pathsep then
+    self.user_campaign_dir = self.user_campaign_dir .. pathsep
+  end
+  return true
+end
```
Alberth289346, line 421: Don't copy/paste code!!

Refactor `App:initUserLevelDir` to a common part and a userLevel-specific part, and use the former also here.

CorsixTH/Lua/app.lua
```diff
@@ -616,9 +640,8 @@ end
 --!return (string, error) Returns the found absolute path, or nil if not found. Then
 --!       a second variable is returned with an error message.
 function App:getAbsolutePathToLevelFile(level)
-  local path = debug.getinfo(1, "S").source:sub(2, -12)
   -- First look in Campaigns. If not found there, fall back to Levels.
-  local list_of_possible_paths = { self.user_level_dir, path .. "Campaigns", self.level_dir }
+  local list_of_possible_paths = { self.user_campaign_dir, self.campaign_dir, self.user_level_dir, self.level_dir }
```
Alberth289346, line 644: Not your fault, but yuck.

"list_of" is type information that is already expressed at the right of the `=` sign. Also it's clear the right side that it's a list of paths, the name doesn't need to tell that too. The reader now has to check what happens with the variable to understand why it exists.

Instead, use it to express the reason of having the variable. Something like "paths_to_search" says what the intention of that data is. From that intention it's fully logical that you iterate over the values next.

CorsixTH/Lua/app.lua
```diff
@@ -118,7 +118,9 @@ function App:init()
   local good_install_folder, error_message = self:checkInstallFolder()
   self.good_install_folder = good_install_folder
   self.level_dir = debug.getinfo(1, "S").source:sub(2, -12) .. "Levels" .. pathsep
+  self.campaign_dir = debug.getinfo(1, "S").source:sub(2, -12) .. "Campaigns" .. pathsep
```
Alberth289346, line 121: Not your fault, but `debug.getinfo(1, "S").source:sub(2, -12)`?

What is debugging code doing here? Why isn't there a nice function-name for this madness?

CorsixTH/Lua/app.lua
lewri: Tbh yes that was me trying to just get it working ages ago. Then neglected to actually tidy it. I'll work on that.

Alberth289346:
Sorry, much comment is not about your changes directly, but cleaning up code that you touch is always very useful.

CorsixTH/Lua/app.lua
lewri: Unfortunately I can't answer. And I would assume as I can't I probably shouldn't try to guess?

lewri:
> Sorry, much comment is not about your changes directly, but cleaning up code that you touch is always very useful.

Fixed two of these, but not the `debug` one.

I added a TODO that I will put into a new PR once this one merges... just in case I forget to do it.

CorsixTH/Lua/app.lua
```diff
@@ -374,26 +375,38 @@ function App:init()
...
-    self.user_level_dir = self.user_level_dir .. pathsep
-  end
-  return true
+
+  self.user_level_dir = self.config.levels or
+      conf_path:match("^(.-)[^" .. pathsep .. "]*$") .. "Levels"
+  self.user_level_dir = setUserDir(self.user_level_dir, "Levels")
+  self.user_campaign_dir = self.config.campaigns or
+      conf_path:match("^(.-)[^" .. pathsep .. "]*$") .. "Campaigns"
+  self.user_campaign_dir = setUserDir(self.user_campaign_dir, "Campaigns")
```
lewri, line 409: Will condense more in a future PR

CorsixTH/Lua/app.lua
lewri: @tobylane answered in Discord. Toby I'm happy if you wanted to commit that change into this PR if you wanted?

tobylane: https://grep.app/search?q=Debug.getinfo&filter[repo][0]=CorsixTH/CorsixTH&filter[lang][0]=Lua&filter[path][0]=CorsixTH/

There's 19 uses of it. It can go here, but if you're ready before then it can go in a new PR.

lewri: There's one in `CorsixTH.lua` -- not much you can do about that one in app, unless you set it in that file.

lewri: If you're touching the entire codebase, I'd put it in a new PR

CorsixTH/Lua/app.lua
```diff
@@ -374,26 +375,38 @@ function App:init()
...
+    end
+    if dir:sub(-1, -1) ~= pathsep then
+     dir = dir .. pathsep
+    end
+    return dir
   end
-  if self.user_level_dir:sub(-1, -1) ~= pathsep then
-    self.user_level_dir = self.user_level_dir .. pathsep
-  end
-  return true
```
lewri, line 396: Should the `return true` had been left in?

CorsixTH/Lua/app.lua
```diff
@@ -374,25 +375,39 @@ function App:init()
...
+  --!param label (string) What folder is being set, for the error dialogue
+  --!return The fully qualified path
+  local function setUserDir(dir, label)
+    if dir:sub(-1, -1) == pathsep then
+      dir = dir:sub(1, -2)
     end
+    if lfs.attributes(dir, "mode") ~= "directory" then
+      if not lfs.mkdir(dir) then
+        print("Notice:" .. label .. "directory does not exist and could not be created.")
+        return false -- Force game exit
```
Alberth289346, line 395: This is not what you promised at line 387.

CorsixTH/Lua/app.lua
Alberth289346: Insert an empty line below 406 so it's easier to understand it's the same thing twice but with a different kind of directly.

Re-use of `self.usr_xyz_dir` variable is confusing, change the first assignment to a local variable.

If you stick the common parts in a function and supply the non-common parts as parameters, future PR would be pretty much done?

CorsixTH/Lua/app.lua
```diff
@@ -616,10 +631,9 @@ end
 --!return (string, error) Returns the found absolute path, or nil if not found. Then
 --!       a second variable is returned with an error message.
 function App:getAbsolutePathToLevelFile(level)
-  local path = debug.getinfo(1, "S").source:sub(2, -12)
   -- First look in Campaigns. If not found there, fall back to Levels.
-  local list_of_possible_paths = { self.user_level_dir, path .. "Campaigns", self.level_dir }
-  for _, parent_path in ipairs(list_of_possible_paths) do
+  local paths_to_search = { self.user_campaign_dir, self.campaign_dir, self.user_level_dir, self.level_dir }
```
Alberth289346, line 635:
```suggestion
  local paths_to_search = {
    self.user_campaign_dir,
    self.campaign_dir,
    self.user_level_dir,
    self.level_dir
  }
```
Maybe too many lines?

CorsixTH/Lua/config_finder.lua
```diff
@@ -414,12 +414,13 @@ unicode_font = nil -- [[X:\ThemeHospital\font.ttc]]
 savegames = nil -- [[X:\ThemeHospital\Saves]]
 
 -------------------------------------------------------------------------------
--- Levels. By default, the "Levels" directory alongside this config file will
--- be used for storing new maps / levels in. Should this not be suitable, then
--- uncomment the following line, and point it to a directory which exists and
--- is more suitable.
+-- Levels and Campaigns. By default, the "Levels" and "Campaigns" directory alongside
```
Alberth289346, line 417:
```suggestion
-- Levels and Campaigns. By default, the "Levels" and "Campaigns" directory next to
```
?

CorsixTH/Lua/config_finder.lua
```diff
@@ -414,12 +414,13 @@ unicode_font = nil -- [[X:\ThemeHospital\font.ttc]]
...
+-- this config file will be used for storing new maps / levels / campaigns in. Should
+-- this not be suitable, then uncomment the following lines, and point it to a directory
```
Alberth289346, line 419:
```suggestion
-- this config file will be used for storing new maps / levels / campaigns. If this is
-- not be suitable, then uncomment the following lines, and point it to a directory.
```
I am a little lost where the maps are going, as there are two directories and three things to store.

CorsixTH/Lua/app.lua
lewri: That was the plan for the future PR, yes -- mainly to go with https://github.com/CorsixTH/CorsixTH/pull/2273#issuecomment-1287763941

CorsixTH/Lua/config_finder.lua
lewri: I don't think I paid enough attention to this textual comment. It needs to say that created maps go into the Levels folder currently.

CorsixTH/Lua/app.lua
Alberth289346: I considered that for about 2 seconds too, but it's not our code style.

Also, if you don't see it when you have 1 word at each line, would you see it if you have multiple words at each line?
I think the benefits are doubtful, and not enough to break code style.

CorsixTH/Lua/app.lua
Alberth289346: ok, fair enough.

Alberth289346:
LGTM

CorsixTH/Lua/app.lua
```diff
@@ -616,10 +632,14 @@ end
...
 --!       a second variable is returned with an error message.
 function App:getAbsolutePathToLevelFile(level)
-  local path = debug.getinfo(1, "S").source:sub(2, -12)
   -- First look in Campaigns. If not found there, fall back to Levels.
-  local list_of_possible_paths = { self.user_level_dir, path .. "Campaigns", self.level_dir }
-  for _, parent_path in ipairs(list_of_possible_paths) do
+  local paths_to_search = {
+    self.user_campaign_dir,
+    self.campaign_dir,
+    self.user_level_dir,
```
TheCycoONE, line 639: The comment is misleading, the user_level_dir should be between the user_campaign_dir and the campaign_dir

lewri: Order changed :+1:

lewri: Though our comment should be something like: First check user's campaigns/level directories. If not found there, fall back to default directories.

lewri: Comment likely isn't needed at all now, will remove.

lewri:
@TheCycoONE you may need to fix the commit history of this one before merge, Github's website didn't do a proper rebase by the look of it.

PR #2276, opened 2022-10-23
Check for updates menu option

### LINKED ISSUE #2274: Toggle on/off check for CorsixTH updates

I would like the option to toggle on/off the checking for CorsixTH updates when the game starts, which displays in the log window. I'm not always around an internet connection and if I wanted to know if there was a new release, I would just check the GitHub page when I was online. 

It wouldn't benefit the gameplay, it is just an annoyance that CorsixTH checks for updates at every startup. A toggle on/off in the settings menu shouldn't be to hard to implement.. not all of us are online 24/7. 

No

0.66

_No response_

### Additional Information

_No response_

lewri:
*Fixes #2274*

- Adds toggle
- Adds alert when updates disabled on main menu

image
image

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/main_menu.lua
+++ b/CorsixTH/Lua/dialogs/resizables/main_menu.lua
@@ -48,9 +48,9 @@ function UIMainMenu:UIMainMenu(ui)
   }
   self.no_menu_entries = #menu_items
 
-  -- Work out the size of non-false items in array
-  -- param items (list) what is to be checked
-  -- return calculated size
+  --! Work out the size of non-false items in array
+  --!param items (list) what is to be checked
+  --!return calculated size
   local function computeSize(items)
     local size = 0
     for _, item in pairs(items) do
@@ -61,10 +61,10 @@ function UIMainMenu:UIMainMenu(ui)
 
   -- Work out the menu's height, giving extra space for the version information
   local line_height = 15
-  local end_of_menu_items_spacer = 20
+  local top_padding = 20
   local num_lines = computeSize({not ui.app.config.check_for_updates, ui.app.config.debug, TheApp:getVersion()}) -- see UIMainMenu:draw for how these are used
-  local bottom_text_height = (line_height * num_lines) + end_of_menu_items_spacer
-  self.height = ((menu_item_height + 10) * #menu_items) + bottom_text_height
+  local bottom_text_height = (line_height * num_lines)
+  self.height = top_padding + ((menu_item_height + 10) * #menu_items) + bottom_text_height
 
   self:UIResizable(ui, 200, self.height, col_bg)
 
@@ -108,7 +108,7 @@ function UIMainMenu:draw(canvas, x, y)
   UIResizable.draw(self, canvas, x, y)
   x, y = self.x + x, self.y + y
 
-  -- The following strings are declared in reverse order
+  -- The following strings are drawn in reverse order
   local ly = y + self.height - 15
   -- Move the version string up a bit if showing check for updates disabled warning.
   if not TheApp.config.check_for_updates then
--- a/CorsixTH/Lua/dialogs/resizables/main_menu.lua
+++ b/CorsixTH/Lua/dialogs/resizables/main_menu.lua
@@ -47,7 +47,26 @@ function UIMainMenu:UIMainMenu(ui)
     {_S.main_menu.exit,            self.buttonExit,           _S.tooltip.main_menu.exit}
   }
   self.no_menu_entries = #menu_items
-  self:UIResizable(ui, 200, (menu_item_height + 10) * (#menu_items + 1), col_bg)
+
+  -- Work out the size of non-false items in array
+  -- param items (list) what is to be checked
+  -- return calculated size
+  local function computeSize(items)
+    local size = 0
+    for _, item in pairs(items) do
+      size = item and (size + 1) or size
+    end
+    return size
+  end
+
+  -- Work out the menu's height, giving extra space for the version information
+  local line_height = 15
+  local end_of_menu_items_spacer = 20
+  local num_lines = computeSize({not ui.app.config.check_for_updates, ui.app.config.debug, TheApp:getVersion()}) -- see UIMainMenu:draw for how these are used
+  local bottom_text_height = (line_height * num_lines) + end_of_menu_items_spacer
+  self.height = ((menu_item_height + 10) * #menu_items) + bottom_text_height
+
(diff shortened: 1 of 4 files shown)


lokpro:
Hi @lewri , would you mind copying my updated lang files and add them to your PR, in order to simplify the log history. Don't need to be credited.
- https://github.com/lokpro/CorsixTH/blob/master/CorsixTH/Lua/languages/traditional_chinese.lua
- https://github.com/lokpro/CorsixTH/blob/master/CorsixTH/Lua/languages/simplified_chinese.lua

lewri:
Someone with git magic might be able to grab them and stick them in here. Otherwise I'll take a look.

Alberth289346:
Not sure it's useful to display an alert when not polling updates. What kind of useful information do you expect to get from it?

lewri:
> Not sure it's useful to display an alert when not polling updates. What kind of useful information do you expect to get from it?

It's to reminder the user the setting is off as per https://github.com/CorsixTH/CorsixTH/issues/2274#issuecomment-1285468951 . Rather than printing to console, which is used for Dev/Trunk versions, a general user should likely be informed via the UI.

Alberth289346:
> It's to reminder the user the setting is off as per [#2274 (comment)](https://github.com/CorsixTH/CorsixTH/issues/2274#issuecomment-1285468951)

Ah right, I wouldn't even consider that as an option, but fair enough.

CorsixTH/Lua/dialogs/resizables/main_menu.lua
```diff
@@ -47,7 +47,12 @@ function UIMainMenu:UIMainMenu(ui)
     {_S.main_menu.exit,            self.buttonExit,           _S.tooltip.main_menu.exit}
   }
   self.no_menu_entries = #menu_items
-  self:UIResizable(ui, 200, (menu_item_height + 10) * (#menu_items + 1), col_bg)
+
+  -- Add a bit more space to menu if updates are disabled
+  local height_modifier = ui.app.config.check_for_updates and 1 or 1.25
+  -- Work out the menu's height, giving extra space for the version information
+  self.menu_items_height = math.floor((menu_item_height + 10) * (#menu_items + height_modifier))
```
Alberth289346, line 54: Your modifier looks like magic.

The usual approach to obtain a height (or width) for a window is to sum the heights(widths) of everything inside the window, including any spacing.
That is what `(menu_item_height + 10) * (#menu_items + 1)` said to me at least (ie `height of a single button` * `number of buttons`).

The "+ 1" in the latter is a bit weird, but that may be added space at the very top and the bottom of the window. Alternatively, someone was lazy and thought that the version numbers were mostly as high as a button so " +1" button would do the trick (instead of the better " + height_of_version_number_lines" after the last closing parenthesis).

The multiplication factor that you added has the wrong intent. In the above meaning, you stated that the total button height increased, so the buttons should become bigger or spacing between them should become larger accordingly.  
That doesn't happen though, the button size computation is off now, it allocates more height for itself than it needs, and you use that created empty space to add a new line.

The fix here is to add height for the 3 text lines, and probably also reduce some of the button height, as the spacing below the bottom button is now less high than it originally was (you'd have to check the history of the "+1" though if you reallt want to know).

lewri: Quite an old commit by lego in an attempt to make it less magic by the looks of it:
https://github.com/CorsixTH/CorsixTH/commit/7bbe037373418749106731b7db5d7dde99e6d480

Are we trying to say a line height here is 15px? As that could be correct for the built in font, but not freetype.
There's also no logic written to get line height that I can see, so usually all our efforts in the ui are best guess.

We could go by a line being 15px just for the sake of setting the height. Window stuff isn't very friendly.
You could also just do a +15 at the end of the menu_items_height when the extra line is needed.

Nothing really screams out a good solution here.

>  also reduce some of the button height, as the spacing below the bottom button is now less high than it originally was

Not sure that's something have a real effect here.

Alberth289346: We have hard-coded sizes all over the place in gui code, not much harm to add a few more imho.

If you want to go on a crusade against that, you'd better engage in a full scale war by building a new way to specify window content.

Alberth289346: (Where we specify what it should show and the computer does the size and position computations.)

CorsixTH/Lua/languages/czech.lua
```diff
@@ -1143,6 +1143,7 @@ tooltip = {
 	shift_scrollspeed = "Nastavte rychlost rolování s shiftem v rozmezí 1 (nejpomalejší) až 10 (nejrychlejší). Výchozí hodnota je 4.",
 	zoom_speed = "Nastavte rychlost přibližování v rozmezí 10 (nejpomalejší) až 1000 (nejrychlejší). Výchozí hodnota je 80.",
 	cancel_zoomspeed = "Neměnit rychlost přibližování.",
+    check_for_updates = "Nastavuje, zda má hra při spuštění kontrolovat aktualizace.",
```
lewri, line 1146:
```suggestion
	cancel_zoomspeed = "Neměnit rychlost přibližování.",
        check_for_updates = "Nastavuje, zda má hra při spuštění kontrolovat aktualizace.",
```

CorsixTH/Lua/languages/czech.lua
```diff
@@ -1143,6 +1143,7 @@ tooltip = {
 	shift_scrollspeed = "Nastavte rychlost rolování s shiftem v rozmezí 1 (nejpomalejší) až 10 (nejrychlejší). Výchozí hodnota je 4.",
 	zoom_speed = "Nastavte rychlost přibližování v rozmezí 10 (nejpomalejší) až 1000 (nejrychlejší). Výchozí hodnota je 80.",
 	cancel_zoomspeed = "Neměnit rychlost přibližování.",
+        check_for_updates = "Nastavuje, zda má hra při spuštění kontrolovat aktualizace.",
     },
```
lewri, line 1147: There's a general alignment issue here caused by a random 4 space indent, and can't seem to resolve it my self. I can try and fix it in a separate PR rather than here?

Alberth289346: Elsewhere is a good idea I think, and you're fighting TABs versus spaces:

Tabs are visible in my editor, and it shows
Screenshot from 2022-10-26 08-07-16

Alberth289346:
Looks fine to me.

lewri:
Sorry for continuous updates on this one. With Albert's nod I think we're ready now

PR #2279, opened 2022-10-27
Fix #2181 Store palette 255 transparency

### LINKED ISSUE #2181: [Bug] Balance Sheet can load green
### Describe the issue

If loading a save where the balance sheet was open, the text becomes highlighted in green and will not go away until the application is restarted. It also continues as new items scroll bay on the balance sheet.

If a save is previously opened where the balance sheet **is not** up, then it is opened, the aforementioned save no longer gets the green.

### Steps to Reproduce

1. Start a new game

2. Navigate to the Bank Manager and open the balance sheet. Save the game.

3. Close CorsixTH and reopen. Load the save.

### Expected Behaviour

The text should not be tainted green.

### Save Game

### CorsixTH Version

0,66 beta 1

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

_No response_

### Additional Information

image

I have a suspicion it could be something to do with the highlight colour of the "load game" menu not being cleared the game is loaded.

TheCycoONE:
Use a flag to mark if the 255 entry in a palette is transparent. Store that flag so it will be applied again on load.

This removes the bug that required sprite sheets to be all fetched on load as well.

RFC because I'm not confident in the afterLoad change made in ui.lua - I suspect I'll have to move those into the individual dialogs that use them instead.

### DIFF
--- a/CorsixTH/Lua/graphics.lua
+++ b/CorsixTH/Lua/graphics.lua
@@ -57,8 +57,8 @@ local cursors_name = {
   banksummary = 44,
 }
 local cursors_palette = {
-  [36] = "bank01v.pal",
-  [44] = "stat01v.pal",
+  [36] = "Bank01V.pal",
+  [44] = "Stat01V.pal",
 }
 
 function Graphics:Graphics(app)
@@ -159,8 +159,7 @@ function Graphics:loadMainCursor(id)
     id = cursors_name[id]
   end
   if id > 20 then -- SPointer cursors
-    local cursor_palette = self:loadPalette("QData", cursors_palette[id])
-    cursor_palette:setEntry(255, 0xFF, 0x00, 0xFF) -- Make index 255 transparent
+    local cursor_palette = self:loadPalette("QData", cursors_palette[id], true)
     return self:loadCursor(self:loadSpriteTable("QData", "SPointer", false, cursor_palette), id - 20)
   else
     return self:loadCursor(self:loadSpriteTable("Data", "MPointer"), id)
@@ -228,9 +227,19 @@ local function makeGreyscaleGhost(pal)
   return table.concat(remap, "", 0, 255)
 end
 
-function Graphics:loadPalette(dir, name)
+--! Load a palette file
+--!param dir (string) The directory of the palette relative to the HOSPITAL directory
+--!param name (string) The name of the palette file
+--!param transparent_255 (boolean) Whether the 255th entry in the palette should be transparent
+--!return (palette, string) The palette and a string representing the palette converted to greyscale
+function Graphics:loadPalette(dir, name, transparent_255)
   name = name or "MPalette.dat"
   if self.cache.palette[name] then
+    local li = self.load_info[self.cache.palette[name]]
+    if li and li[5] ~= transparent_255 then
+      print("Warning: palette " .. name .. " requested with different flags than stored")
+    end
+
     return self.cache.palette[name],
       self.cache.palette_greyscale_ghost[name]
   end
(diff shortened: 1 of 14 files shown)


CorsixTH/Lua/graphics.lua
```diff
@@ -4,7 +4,7 @@ Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
 the Software without restriction, including without limitation the rights to
 use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
-of the Software, and to permit persons to whom the Software is furnished to do
+of the Softwase, and to permit persons to whom the Software is furnished to do
```
TheCycoONE, line 7: Revert accidental typo

CorsixTH/Lua/graphics.lua
```diff
@@ -228,19 +227,28 @@ local function makeGreyscaleGhost(pal)
...
+--!param transparent_255 (boolean) Whether the 255th entry in the palette should be transparent
+--!return (palette, string) The palette and a string representing the palette converted to greyscale
+function Graphics:loadPalette(dir, name, transparent_255)
   name = name or "MPalette.dat"
   if self.cache.palette[name] then
     return self.cache.palette[name],
       self.cache.palette_greyscale_ghost[name]
   end
 
+  print("load palette " .. (dir or "") .. ":" .. name .. " transparent_255" .. (transparent_255 and "yes" or "no"))
```
TheCycoONE, line 242: Remove debug line

CorsixTH/Lua/ui.lua
```diff
@@ -1120,6 +1078,58 @@ function UI:afterLoad(old, new)
   if old < 5 then
     self.editing_allowed = true
   end
+  if old < 176 then
+    if self.app.good_install_folder and not self.app.using_demo_files then
+      local gfx = self.app.gfx
+      local palette
+      -- load drug casebook sprite table
+      gfx.cache.palette["DrugN01V.pal"] = nil
```
TheCycoONE, line 1086: Probably needs to be split up into each dialog

TheCycoONE: I fixed the bug the comments are referring to in this PR, but I don't think the afterload in ui is enough to fix old saves. It wasn't for bank_manager. That's the main reason I don't consider this PR ready yet, and why I added my own comment.

CorsixTH/Lua/graphics.lua
```diff
@@ -254,7 +262,7 @@ function Graphics:loadGhost(dir, name, index)
   return cached:sub(index * 256 + 1, index * 256 + 256)
 end
 
-function Graphics:loadRaw(name, width, height, dir, paldir, pal)
+function Graphics:loadRaw(name, width, height, dir, paldir, pal, transparent_255)
```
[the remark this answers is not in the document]
TheCycoONE: I don't like this method signature at all. I want Graphics:loadRaw(dir, name, width, heigh, palette) - but I can't change it without breaking compatibility because of the load table.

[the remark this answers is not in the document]
TheCycoONE: That deserves a further explanation:

Loading the same palette more than once with different flags doesn't work as expected, they are cached by name only; so when I was simply adding the transparent_255 flag to loadPalette, it was still getting loaded without transparency because there was a loadRaw earlier - and the loadRaw didn't even reference a palette, it automatically loads a palette matching the name of the bitmap.

I'm unimpressed.

TheCycoONE:
I think this is ready now. I've fun some more testing, caught a bug, and added an additional early warning message when these things get mixed up.

PR #2282, opened 2022-10-29
Properly apply Roujin's challenge

lewri:
- Moves the spawn rate modifier to become a bonus rather than just a static pool.
- Adds todo note about how updateSpawnDates() is to be adjusted ( #1986 )
- This modifier will likely need to change location when spawn changes include the AIHospital proper.

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1254,16 +1254,19 @@ end
 
 -- Called when a month ends. Decides on which dates patients arrive
 -- during the coming month.
+-- TODO: Requires adjustment for AIHospital spawns; see PR 1986 for progress.
 function World:updateSpawnDates()
   local local_hospital = self:getLocalPlayerHospital()
   -- Set dates when people arrive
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/world.lua
```diff
@@ -1254,16 +1254,19 @@ end
...
 function World:updateSpawnDates()
   local local_hospital = self:getLocalPlayerHospital()
   -- Set dates when people arrive
   local no_of_spawns = math.n_random(self.spawn_rate, 2)
-  -- If Roujin's Challenge is on, override spawn rate
-  if local_hospital.hosp_cheats:isCheatActive("spawn_rate_cheat") then
-    no_of_spawns = 40
-  end
   -- Use ceil so that at least one patient arrives (unless population = 0)
   no_of_spawns = math.ceil(no_of_spawns*self:getLocalPlayerHospital().population)
```
lewri, line 1263: I might as well change this to use `local_hospital` too right?

TheCycoONE (approved):
LGTM

PR #2285, opened 2022-10-31
Refactor ReceptionDesk:tick()

lewri:
- Attempts to make the tick function easier to read at the cost of increased length.

### DIFF
--- a/CorsixTH/Lua/objects/reception_desk.lua
+++ b/CorsixTH/Lua/objects/reception_desk.lua
@@ -92,48 +92,77 @@ function ReceptionDesk:onClick(ui, button)
 end
 
 function ReceptionDesk:tick()
-  local queue_front = self.queue:front()
-  local reset_timer = true
-  if self.receptionist and queue_front then
-    if queue_front:getCurrentAction().name == "idle" then
-      self.queue_advance_timer = self.queue_advance_timer + 1
-      reset_timer = false
-      if self.queue_advance_timer >= 4 + Date.hoursPerDay() * (1.0 - self.receptionist.profile.skill) then
-        reset_timer = true
-        if queue_front.next_room_to_visit then
-          queue_front:queueAction(SeekRoomAction(queue_front.next_room_to_visit.room_info.id))
-        else
-          if class.is(queue_front, Inspector) then
-            local inspector = queue_front
-            if not inspector.going_home  then
-              local epidemic = self.hospital.epidemic
-              if epidemic then
-                -- The result of the epidemic may already by determined
-                -- i.e if an infected patient has left the hospital
-                if not epidemic.result_determined then
-                  epidemic:finishCoverUp()
-                end
-                epidemic:applyOutcome()
-                inspector:goHome()
-              end
-            end
-            -- VIP has his own list, don't add the gp office twice
-          elseif queue_front.humanoid_class ~= "VIP" then
-            if queue_front:agreesToPay("diag_gp") then
-              queue_front:queueAction(SeekRoomAction("gp"))
-            else
-              queue_front:goHome("over_priced", "diag_gp")
-            end
-          else
-            -- the VIP will realise that he is idle, and start going round rooms
-            queue_front:queueAction(IdleAction())
-            queue_front.waiting = 1
-          end
-        end
-        self.queue:pop()
-        self.queue.visitor_count = self.queue.visitor_count + 1
-        queue_front.has_passed_reception = true
+  --! Handle a patient at the reception desk.
+  --!param patient The patient at the front of queue.
+  --!param is_new (boolean) Have they just entered the hospital?
+  --!return action taken by patient
+  local function handlePatient(patient, is_new)
+    if not is_new then
+      -- Patient redirected to reception, send them on their way
+      local id = patient.next_room_to_visit.room_info.id
+      return patient:queueAction(SeekRoomAction(id))
+    end
+
+    --else: a new patient has arrived in the hospital
+    if patient:agreesToPay("diag_gp") then
+      return patient:queueAction(SeekRoomAction("gp"))
+    end
+    --Patient goes home, too costly
+    return patient:goHome("over_priced", "diag_gp")
+  end
+
+  -- Sends the VIP on their way.
+  local function handleVIP(vip)
+    vip:queueAction(IdleAction())
+    vip.waiting = 1
+  end
+
+  -- Let the Inspector conclude the epidemic, then send them on their way.
+  local function handleInspector(inspector)
+    if inspector.going_home then return end
+    local epidemic = self.hospital.epidemic
+    if epidemic then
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -92,48 +92,74 @@ function ReceptionDesk:onClick(ui, button)
...
-        queue_front.has_passed_reception = true
-      end
+    elseif queue_front.next_room_to_visit then
+      handlePatient(queue_front, false)
+    elseif class.is(queue_front, Inspector) then
+      handleInspector(queue_front)
+    elseif class.is(queue_front, Vip) then
+      handleVIP(queue_front)
+    else -- New patient
+      handlePatient(queue_front, true)
```
lewri, line 156: I've passed `queue_front` here into the respective functions, rather than declaring `queue_front` at the top, to make those functions more fitting to context.

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -92,48 +92,74 @@ function ReceptionDesk:onClick(ui, button)
 end
 
 function ReceptionDesk:tick()
+  --! Handle a patient at the reception desk.
+  --!param patient The patient at the front of queue.
+  --!param new (bool) Have they just entered the hospital?
+  local function handlePatient(patient, new)
```
Alberth289346, line 98:
```suggestion
  local function handlePatient(patient, is_new)
```
By prepending "is" or "has" it becomes more obvious the variable holds a boolean value.

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -92,48 +92,74 @@ function ReceptionDesk:onClick(ui, button)
...
+    if not new then
+      -- Patient redirected to reception, send them on their way
+      local id = patient.next_room_to_visit.room_info.id
+      return patient:queueAction(SeekRoomAction(id))
```
Alberth289346, line 102: Here a value is returned while in the other cases it is not. There is also no lua doc comment in the function docs.

I would suggest to make things consistent.

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -92,48 +92,74 @@ function ReceptionDesk:onClick(ui, button)
...
+    vip:queueAction(IdleAction())
+    vip.waiting = 1
+  end
+
+  -- Let the Inspector conclude the epidemic, then send them on their way.
+  local function handleInspector(inspector)
+    if inspector.going_home then return end
+    local epidemic = self.hospital.epidemic
+    if epidemic then
+      -- The result of the epidemic may already by determined
```
Alberth289346, line 124:
```suggestion
      -- The result of the epidemic may already be decided
```
?

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -92,48 +92,74 @@ function ReceptionDesk:onClick(ui, button)
...
+      epidemic:applyOutcome()
+      inspector:goHome()
+    end
+  end
+
   local queue_front = self.queue:front()
-  local reset_timer = true
-  if self.receptionist and queue_front then
+  --! Advance the front humanoid of the queue.
+  --! Returns true if ready to be dealt with.
```
Alberth289346, line 136:
```suggestion
  --!return (boolean) Returns true if ready to be dealt with, else false.
```
In this way ldoc understands that the second sentence is about the return value.

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -92,48 +92,74 @@ function ReceptionDesk:onClick(ui, button)
...
+  local function advanceQueue()
     if queue_front:getCurrentAction().name == "idle" then
       self.queue_advance_timer = self.queue_advance_timer + 1
+    end
+    return self.queue_advance_timer >= 4 + Date.hoursPerDay() * (1.0 - self.receptionist.profile.skill)
+  end
+
+  local reset_timer = true
```
Alberth289346, line 144:
```suggestion
  -- Start of the ReceptionDesk:tick function.
  local reset_timer = true
```
A little hint to start reading here :)
Not strictly needed though.

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -92,48 +92,74 @@ function ReceptionDesk:onClick(ui, button)
...
+    end
+  end
+
   local queue_front = self.queue:front()
-  local reset_timer = true
-  if self.receptionist and queue_front then
+  --! Advance the front humanoid of the queue.
+  --! Returns true if ready to be dealt with.
+  local function advanceQueue()
     if queue_front:getCurrentAction().name == "idle" then
```
Alberth289346, line 138: Pass queue front in here with a parameter as well?  Looks like a humanoid to me.

A next step is to move this query into a function in the humanoid class so it goes through an interface instead of poking around into the internals of a foreign object. The current style is however still common, so leaving it as-is, is fine too.

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -92,48 +92,74 @@ function ReceptionDesk:onClick(ui, button)
...
-            queue_front:queueAction(IdleAction())
-            queue_front.waiting = 1
-          end
-        end
-        self.queue:pop()
-        self.queue.visitor_count = self.queue.visitor_count + 1
-        queue_front.has_passed_reception = true
-      end
+    elseif queue_front.next_room_to_visit then
+      handlePatient(queue_front, false)
```
Alberth289346, line 150: There is no test here that it's a patient (unlike the next two cases). Also, the other case with a patient is at the bottom, which is a bit weird.

Assuming that `handlePatient` here is correct, maybe do a class check here as well, and move both patient cases next to each other (or even merge them with `handlePatient(queue_front, not queue_front.next_room_to_visit)`).

Renaming `queue_front` to `humanoid` or `front_humanoid` may make things a little more readable as well.

CorsixTH/Lua/objects/reception_desk.lua
Alberth289346: Very good!

Alberth289346:
Nice work!

CorsixTH/Lua/objects/reception_desk.lua
lewri: I'm happy to test for patient. But what should we do if we fail to identify the humanoid as a valid user of the desk? Should we error -- or attempt to handle the humanoid as if they were a patient?

CorsixTH/Lua/objects/reception_desk.lua
lewri: I'll leave this off for the moment, unless there's pushback to include it.

Alberth289346:
LGTM

PR #2286, opened 2022-11-01
Remove tabs from code

lewri:
*Fixes https://github.com/CorsixTH/CorsixTH/pull/2276#pullrequestreview-1155573601*

- Replaces tabs with 2 spaces as documented.


lewri:
@TheCycoONE can we add a unit test to github actions to check for tabs?

tobylane:
That was the purpose of this line, but it doesn't work as intended. https://github.com/CorsixTH/CorsixTH/blob/54c0adc37caa612bd9e154f5f56958cc87ec57e9/.github/workflows/Linux.yml#L104

I think this is a suitable replacement - IPRn are ignore binary files, recursive into directories, Perl regex and give line numbers. `grep -IRPn '\t' CorsixTH/Lua CorsixTH/CorsixTH.lua`.

lewri:
> That was the purpose of this line, but it doesn't work as intended.
> 
> https://github.com/CorsixTH/CorsixTH/blob/54c0adc37caa612bd9e154f5f56958cc87ec57e9/.github/workflows/Linux.yml#L104
> 
> I think this is a suitable replacement - IPRn are ignore binary files, recursive into directories, Perl regex and give line numbers. `grep -IRPn '\t' CorsixTH/Lua CorsixTH/CorsixTH.lua`.

Not fully convinced languages should be excluded. Feed bad practice anywhere and it'll creep to other areas :( .

tobylane:
Yes, as here you make the changes needed we can stop excluding it.

PR #2288, opened 2022-11-07
Set direct zoom to true as default

lewri:
- Since direct zoom is now enabled by default, the setting should be true by default too.


lewri:
This one just bugged out... I have redone it now

Issue #1992, opened 2021-07-06
[Bug] config should always use the default value when property is nil or missing from the file


TheCycoONE:
See #1963 for an example where this caused an issue.

A nil value for a property in the config file, or a missing value should ALWAYS be treated as if a default value had been specified.

It should preferably do this without clobbering the nil value when the config is saved again, because in many cases we'd like values that were not explicitly set to follow new defaults as the game is updated.

At the moment this is handled in a largely ad-hoc way at point of use and many times is missed altogether, a uniform mechanism in config itself is preferable, perhaps through the use of an `__index` metamethod.

lewri:
A value being `nil` itself in the config seems to cause breakages -- since it doesn't exist.
And setting nil inside config.txt causes just as much breakage.

Alberth289346:
It should be a text-file rather than lua-code so you can parse it and detect available entries even when empty (or nil).
While at it, switching to a common format may be beneficial.

lewri:
I think it's #2082 causing most of this bug on looking back. Hopefully fixing that also fixes this

lewri:
@TheCycoONE #2106 has fixed this if it's enough for you to consider this done.

lewri:
Resolved by #2106

PR #2289, opened 2022-11-07
Set direct zoom to true as default

lewri:
Attempting #2288 again

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -156,7 +156,7 @@ function App:init()
   if self.config.track_fps then
     modes[#modes + 1] = "present immediate"
   end
-  if self.config.direct_zoom == nil or self.config.direct_zoom then
+  if self.config.direct_zoom then
     modes[#modes + 1] = "direct zoom"
   end
   self.modes = modes
--- a/CorsixTH/Lua/config_finder.lua
+++ b/CorsixTH/Lua/config_finder.lua
@@ -142,7 +142,7 @@ local config_defaults = {
   check_for_updates = true,
   room_information_dialogs = true,
   allow_blocking_off_areas = false,
-  direct_zoom = nil,
+  direct_zoom = true,
   new_machine_extra_info = true,
   player_name = [[]],
 }


TheCycoONE:
I believe you shouldn't need to do that, we wanted nil to represent the default so we could interpret it either way.

TheCycoONE:
Ah I see... the problem with explicitly marking things... actually the problem with the way we do config, is that we can't differentiate between defaults and user preferences which makes it hard to evolve.

lewri:
If you set a value to nil in `config.txt`, the game attempts to overwrite the nil value on every load. 
Ideally you don't want nil for such a setting here, as it now unnecessarily tries to replace it... with nil.

The setting imo should be a real representation of the status of direct zoom, rather than assuming nil to be true. 

The other debug lines as nil probably aren't the best either, they might be best using a format like player name as `[[]]`

TheCycoONE:
The problem I was trying to solve for originally was that directZoom was going to be false for everyone until they opted in. When you install a new version of the game you keep your config so we would have needed to tell everyone to change it. We could have said in the update notes, but most people wouldn't notice.

Thoughts?

Alberth289346:
#1992

lewri:
> The problem I was trying to solve for originally was that directZoom was going to be false for everyone until they opted in. When you install a new version of the game you keep your config so we would have needed to tell everyone to change it. We could have said in the update notes, but most people wouldn't notice.
> 
> Thoughts?

It may not matter all in all. And as I've stepped away from this for over a month I've lost the moment of understanding.

Currently the direct_zoom config value gets interpreted as `nil == true` in 0.66. 
In 0.65 we had `nil == false`. In essence, we went from an opt-in model to an opt-out model. So unless the user specified a choice on that value they were given the default. But in all other boolean settings that impact a user generally have a literal `true` or `false`, which gives a definitive statement on what the value is.

If you implement this change post 0.66, you have no additional unexpected impact on user experience (ergo, <=0.65 to 0.67 will always end up enabling direct zoom if nil or not present). If the config value was `nil`, it will be automatically changed to `true` during `App:fixConfig()`-- thus representing our currently enforced default. If the user expressly set a `true` or `false` value, it is not changed at all.

As for #1992 there is a great point about experimental changes using `nil` from you Cyco and I respect it. However, direct zoom is now past this point and we should seek to correct its course now it is enabled by default (unless we're still planning to work it again). 
Perhaps because `nil` use for experimental stuff is not showing as reliably suitable, but that is a conversation to be had in that issue, not here (and the other issue of `nil` being changed to `nil`).

lewri:
Closing for now

PR #2291, opened 2022-11-12
[rfc] Create gamelog on launch if missing

### LINKED ISSUE #2170: [Bug] Gamelog.txt isn't generated (generate system info on load/new game)
### Describe the issue

The expected gamelog.txt file isn't being generated when starting the game. No errors appear in the UI or within the console for this.

### Steps to Reproduce

1. Run CorsixTH

2. Check for gamelog.txt where config.txt is (in this case, ~/.config/CorsixTH/)

3. File is not created

### Expected Behaviour

File should be present.

### Save Game

### CorsixTH Version

0.66-beta1

### Operating System

MacOS Monterey (Apple Silicon)

### Theme Hospital Version

CD

### Gamelog.txt

```bash
Not present
```

### Additional Information

_No response_

lewri:
*Fixes #2170*

- Creates the initial system information for the gamelog.txt on launch if the file is completely missing.

Happy to discuss whether we should be doing more than just checking it exists. I took the approach that we shouldn't try to overwrite the gamelog at every launch -- in case a user wants to report the logs and then accidentally deletes it because they restarted the game.


CorsixTH/Lua/app.lua
```diff
@@ -374,6 +377,25 @@ function App:init()
...
+  local config_path = self.command_line["config-file"] or ""
+  config_path = config_path:match("^(.-)[^" .. pathsep .. "]*$")
+  return config_path .. "gamelog.txt"
+end
+
+--! Checks and creates the gamelog file if it does not exist.
+function App:initGamelogFile()
+  local gamelog_path = self:getGamelogPath()
+  local gamelog = io.open(gamelog_path, "r")
+  if gamelog and gamelog:close() then return end
```
lewri, line 392: Should this be a generic utility function (maybe in future)?

We have `FileSystem:fileExists` but this is TH folder-centric, and not a global checker.

lewri: Maybe it's safer to do `if gamelog then return gamelog:close() end` -- though if this generates a problem it'll unintentionally crash the game

TheCycoONE:
Stealing my *Good First Issues* 😛

CorsixTH/Lua/app.lua
```diff
@@ -374,6 +377,25 @@ function App:init()
...
+  local config_path = self.command_line["config-file"] or ""
+  config_path = config_path:match("^(.-)[^" .. pathsep .. "]*$")
+  return config_path .. "gamelog.txt"
+end
+
+--! Checks and creates the gamelog file if it does not exist.
+function App:initGamelogFile()
+  local gamelog_path = self:getGamelogPath()
+  local gamelog = io.open(gamelog_path, "r")
+  if gamelog then return gamelog:close() end
```
TheCycoONE, line 392: Why do you `return gamelog:close()` - what is that value?

It's unused in this patch. I would have expected just: `gamelog:close()`

lewri: It was
`if gamelog and gamelog:close() then return end`. Then it was `if gamelong then return gamelog:close end`.
Maybe I'm trying to save space where I shouldn't.

TheCycoONE: That's logically a different statement too. Revert to what it was please.

TheCycoONE:
Tested?

lewri:
Yes all fine my side. Not sure if @tobylane or @colinjmatt want to test this change in a Mac environment (you'll need to delete the gamelog before launch).

tobylane:
It's too early for ui or _S, so `App:writeToFileOrTmp` won't work yet. It needs a hard coded English error string, eg the following, or something similar to this in `App:initGamelogFile`.

````lua
function App:writeToFileOrTmp(file, mode)
  local f, err = io.open(file, mode or "w")
  if err then
    local tmp_file = os.tmpname()
    f = io.open(tmp_file, mode or "w")
    if self.ui then self.ui:addWindow(UIInformation(self.ui,
        {_S.errors.save_to_tmp:format(file, tmp_file, err)}))
    else print("A file was written to", tmp_file, "because of the error", err)
    end
  end
  assert(f, "Error: cannot write to filesystem")
  return f
end
````

**Comment by lewri** [comment]
Okay so initGamelog needs to load after strings and ui? Though if that window is made before the movie plays would it present itself correctly?

**Comment by tobylane** [comment]
I'd like it as early as possible, eg line 150, after the os is set.

**Comment by lewri** [comment]
> I'd like it as early as possible, eg line 150, after the os is set.

I see , apololgies I didn't notice the edit you made to the code. I'll modify that that function then and mode to line 150 if you think it'll be fine from then.

**Comment by lewri** [comment]
@tobylane updated accordingly. Can you confirm that the concatenate will work? Or is it likely to crash if you concatenate?

**Review by tobylane — APPROVED** [review]

File: CorsixTH/Lua/app.lua [hunk 3]
```diff
@@ -949,8 +971,11 @@ function App:writeToFileOrTmp(file, mode)
   if err then
     local tmp_file = os.tmpname()
     f = io.open(tmp_file, mode or "w")
-    self.ui:addWindow(UIInformation(self.ui,
-      { _S.errors.save_to_tmp:format(file, tmp_file, err) }))
+    if self.ui then self.ui:addWindow(UIInformation(self.ui,
+        { _S.errors.save_to_tmp:format(file, tmp_file, err) }))
+    else
+      print("A file was written to " .. tmp_file .. " because of the error: " .. err)
```
TheCycoONE, line 977: This error is going to be pretty useless for the end user and us as written.

Maybe `"Attempt to write to " .. file .. " failed. File was written instead to temporary location " .. tmp_file .. " because of the error: " .. err`

Context errors would be even better, but they'd need to be handled outside of this function by returning the result. e.g.

```
Attempting to write configuration failed. Configuration was instead written to temporary location xxx. Your configuration changes will be lost on restart unless you copy that file to aaa
```
That is actionable and user friendly.

lewri: The `_S` version uses a different format but generally the same:
``save_to_tmp = "The file at %s could not be used. The game has been saved to %s. Error: %s",``
E.g. when you set the config file to read-only:
image
I agree with the rewrite of the print string.

(This is more specifically about the context error part here)
I'm not sure how much I agree with end-users going around digging in temp files and moving them though. As it still may not solve the root problem (more Windows and Mac users, I imagine Linux are probably more handy but that's quite a generalisation). Plus, the temp file is saved as `.0`, I'm pretty sure we'd get people just dropping the `.0` file without changing extension unless they're told to. The current solution does include the error. So many also include the error output of why x file failed to write too some point in the future.

lewri: In fact the `save_to_tmp` string isnt generalised enough currently.

lewri:
Any more blocks here @TheCycoONE ?

PR #2292, opened 2022-11-13
Add date to gamelog

lewri:
*Fixes #*

- Append date to top of gamelog output
- This may be more useful in future, if we ever wanted to 'append' new gamelog entries to the file rather than overwriting it (though this could lead to bloat)

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -392,8 +392,7 @@ function App:initGamelogFile()
   if gamelog then gamelog:close() return end
 
   local fi = self:writeToFileOrTmp(gamelog_path)
-  local gen_date, sysinfo = self:getSystemInfo()
-  fi:write(string.format("Gamelog generated on %s\n", gen_date))
+  local sysinfo = self:gamelogHeader()
   fi:write(sysinfo)
   fi:close()
 end
@@ -1852,10 +1851,10 @@ function App:isAudioEnabled()
 end
 
 --! Generate information about user's system and the program
---!return gen_date (string) the date and time information was generated
---!return (compiled .. running) (string) system and program info
-function App:getSystemInfo()
+--!return System and program info as a string
+function App:gamelogHeader()
   local gen_date = os.date("%Y-%m-%d %H:%M:%S")
+  gen_date = string.format("Gamelog generated on %s\n", gen_date)
   local compile_opts = TH.GetCompileOptions()
   local comp_details = {}
   for key, value in pairs(compile_opts) do
@@ -1867,7 +1866,7 @@ function App:getSystemInfo()
   local running = string.format("%s run with api version: %s, game version: %s, savegame version: %s\n",
       compile_opts.jit or _VERSION, tostring(corsixth.require("api_version")),
       self:getVersion(), tostring(SAVEGAME_VERSION))
-  return gen_date, compiled .. running
+  return (gen_date .. compiled .. running)
 end
 
 -- Do not remove, for savegame compatibility < r1891
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -392,7 +392,9 @@ function App:initGamelogFile()
   if gamelog then gamelog:close() return end
 
   local fi = self:writeToFileOrTmp(gamelog_path)
-  fi:write(self:getSystemInfo())
+  local gen_date, sysinfo = self:getSystemInfo()
+  fi:write(string.format("Gamelog generated on %s\n", gen_date))
+  fi:write(sysinfo)
   fi:close()
 end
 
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/app.lua
```diff
@@ -1827,15 +1827,16 @@ end
...
   local compile_opts = TH.GetCompileOptions()
   local comp_details = {}
+  local date = os.date("%Y-%m-%d %H:%M:%S") -- Use ISO 8601 standard
   for key, value in pairs(compile_opts) do
     table.insert(comp_details, key .. ": " .. tostring(value))
   end
   table.sort(comp_details)
-  local compiled = string.format("Compiled with %s\nSDL renderer: %s\n",
-    table.concat(comp_details, ", "), self.video:getRendererDetails())
+  local compiled = string.format("Gamelog generated on %s\nCompiled with %s\nSDL renderer: %s\n",
```
lewri, line 1835: Now potentially out of scope of the function.. maybe be isolated to the gamelog functions instead (world:dumpgamelog and app:initgamelog).

lewri:
Blocked currently by #2291 unless I integrate this PR into it.

tobylane:
I like it, I have a minor suggestion to separate the date line for clarity.

````lua
function App:getSystemInfo()
  local date = string.format("Gamelog generated on %s\n", os.date("%Y-%m-%d %H:%M:%S")) -- Use ISO 8601 standard
  local compile_opts = TH.GetCompileOptions()
  local comp_details = {}
  for key, value in pairs(compile_opts) do
    table.insert(comp_details, key .. ": " .. tostring(value))
  end
  table.sort(comp_details)
  local compiled = string.format("Compiled with %s\nSDL renderer: %s\n",
      table.concat(comp_details, ", "), self.video:getRendererDetails())
  local running = string.format("%s run with api version: %s, game version: %s, savegame version: %s\n",
      compile_opts.jit or _VERSION, tostring(corsixth.require("api_version")),
      self:getVersion(), tostring(SAVEGAME_VERSION))
  return (date .. compiled .. running)
end
````

**Comment by lewri** [comment]
I'm not sure if Date should be part of scope of system information though? Do you think it should?

> I like it, I have a minor suggestion to separate the date line for clarity.
> 
> ```lua
> function App:getSystemInfo()
>   local date = string.format("Gamelog generated on %s\n", os.date("%Y-%m-%d %H:%M:%S")) -- Use ISO 8601 standard
>   local compile_opts = TH.GetCompileOptions()
>   local comp_details = {}
>   for key, value in pairs(compile_opts) do
>     table.insert(comp_details, key .. ": " .. tostring(value))
>   end
>   table.sort(comp_details)
>   local compiled = string.format("Compiled with %s\nSDL renderer: %s\n",
>       table.concat(comp_details, ", "), self.video:getRendererDetails())
>   local running = string.format("%s run with api version: %s, game version: %s, savegame version: %s\n",
>       compile_opts.jit or _VERSION, tostring(corsixth.require("api_version")),
>       self:getVersion(), tostring(SAVEGAME_VERSION))
>   return (date .. compiled .. running)
> end
> ```

At the moment I've changed it to this, but haven't pushed it yet.

```lua
function App:getSystemInfo()
  local compile_opts = TH.GetCompileOptions()
  local comp_details = {}
  local gen_date = os.date("%Y-%m-%d %H:%M:%S") -- Use ISO 8601 standard
  for key, value in pairs(compile_opts) do
    table.insert(comp_details, key .. ": " .. tostring(value))
  end
  table.sort(comp_details)
  local compiled = string.format("Compiled with %s\nSDL renderer: %s\n",
      table.concat(comp_details, ", "), self.video:getRendererDetails())
  local running = string.format("%s run with api version: %s, game version: %s, savegame version: %s\n",
      compile_opts.jit or _VERSION, tostring(corsixth.require("api_version")),
      self:getVersion(), tostring(SAVEGAME_VERSION))
  return gen_date, compiled .. running
end
```

CorsixTH/Lua/app.lua
```diff
@@ -1827,15 +1827,16 @@ end
 function App:getSystemInfo()
   local compile_opts = TH.GetCompileOptions()
   local comp_details = {}
+  local date = os.date("%Y-%m-%d %H:%M:%S") -- Use ISO 8601 standard
```
TheCycoONE, line 1830: For what it's worth ISO 8501 requires a T not a ` `. Remove the misleading comment or fix the format. I have no strong preference.

lewri: removed comment

CorsixTH/Lua/app.lua
```diff
@@ -1849,19 +1851,23 @@ function App:isAudioEnabled()
...
   local compiled = string.format("Compiled with %s\nSDL renderer: %s\n",
-    table.concat(comp_details, ", "), self.video:getRendererDetails())
+      table.concat(comp_details, ", "), self.video:getRendererDetails())
   local running = string.format("%s run with api version: %s, game version: %s, savegame version: %s\n",
-    compile_opts.jit or _VERSION, tostring(corsixth.require("api_version")),
-    self:getVersion(), tostring(SAVEGAME_VERSION))
-  return (compiled .. running)
+      compile_opts.jit or _VERSION, tostring(corsixth.require("api_version")),
+      self:getVersion(), tostring(SAVEGAME_VERSION))
+  return gen_date, compiled .. running
```
TheCycoONE, line 1870: Not clear that we should be mixing conventions like this for this case.

What I mean is: getSystemInfo is currently concatenating the exact string to output in the game log. Now it's doing part of that construction and relying on the caller to do the rest. What set off alarms for me was that it was no longer providing it's primary function in the first return value, which is the normal convention.

I understand you're trying to abstract the function from it's use. I think we should do one of three things:
1. give up pretending this is an abstract function and call it `gameLogHeader()` and include `Gamelog generated on %s\n` in the returned string.
2. Include the date string in the returned string in such a way that it need not be specific to a game log e.g:
```
2022-12-07 08:59:00
Compiled with ...
```
3. Call os.date at both call sites where you're currently specifying `Gamelog generated on`.

3 feels like the worst option, the duplication you already have with the string is a good smell.

TheCycoONE (changes requested):

CorsixTH/Lua/app.lua
lewri: > 1. give up pretending this is an abstract function and call it `gameLogHeader()` and include `Gamelog generated on %s\n` in the returned string

Option 1 done

PR #2306, opened 2023-02-13
Account for draw scale in canvas size.

### LINKED ISSUE #2304: [Bug] Some objects disappear from map at edge before going off-screen
### Describe the issue

The seems to only be affecting objects that have a resulting usage animation where that object can be placed in all 4 directions. And in only one of their orientations (I think south-facing, but the skew of the game confuses me).
Currently noted objects: Sofa, Bed, Bench (north facing?)

This appears to have broken in 0.66, working in 0.65.
It also breaks with direct_zoom being enabled; when disabled it is fine.

### Steps to Reproduce

2. 

3.

### Expected Behaviour

Objects shouldn't be removed from view while still on screen

### Save Game

Included for an example with the benches; save a little abused but should work 'enough' in 0.66 and dev.
[debug queue overflow.zip](https://github.com/CorsixTH/CorsixTH/files/10669546/debug.queue.overflow.zip)

### CorsixTH Version

0.66

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

_No response_

### Additional Information

image
image

flackr:
When a draw scale is set, the caller is not supposed to need to account for the draw scale. This scales the width and height returned accordingly.

*Fixes #2304*

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -689,13 +689,13 @@ void render_target::pop_clip_rect() {
 int render_target::get_width() const {
   int w;
   SDL_RenderGetLogicalSize(renderer, &w, nullptr);
-  return std::ceil(w / draw_scale());
+  return static_cast<int>(std::ceil(w / draw_scale()));
 }
 
 int render_target::get_height() const {
   int h;
   SDL_RenderGetLogicalSize(renderer, nullptr, &h);
-  return std::ceil(h / draw_scale());
+  return static_cast<int>(std::ceil(h / draw_scale()));
 }
 
 void render_target::start_nonoverlapping_draws() {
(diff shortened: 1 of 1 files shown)


lewri (commented):
Are these relevant flackr?

```
'return': conversion from 'double' to 'int', possible loss of data
CorsixTH\Src\th_gfx_sdl.cpp
692

'return': conversion from 'double' to 'int', possible loss of data
CorsixTH\Src\th_gfx_sdl.cpp
698
```

Otherwise, working fine here.

flackr:
> Are these relevant flackr?
> 
> ```
> 'return': conversion from 'double' to 'int', possible loss of data
> CorsixTH\Src\th_gfx_sdl.cpp
> 692
> 
> 'return': conversion from 'double' to 'int', possible loss of data
> CorsixTH\Src\th_gfx_sdl.cpp
> 698
> ```
> 
> Otherwise, working fine here.

Doesn't hurt to make the cast to int explicit :-). Done in latest patch.

lewri (approved):

PR #2310, opened 2023-02-18
[RDY]Fix regression in #2086

mugmuggy:
*Fixes #2278, #2305*

**Describe what the proposed change does**
- Restores settings interrupted walk must_happen from saved_must_happen
- Adds afterLoad to fix instances

PR #2313, opened 2023-03-13
Fix: patient going home without despawn action

JoshKeegan:
Fixes a bug where a patient could be going home (`self.going_home`) but not have any action. 

The previous handling of an empty action queue treated this as ok, but it can cause a crash. This hapenned to me in the attached save game. If you open it, the crash will happen on the next patient to use the decontamination shower (there's no doctor in there, so you  will need to wait a minute or just move one to the room).

[crash shower.zip](https://github.com/CorsixTH/CorsixTH/files/10961267/crash.shower.zip)

### DIFF
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -594,9 +594,9 @@ end
 
 --! Handle an empty action queue in some way instead of crashing.
 function Humanoid:_handleEmptyActionQueue()
-  -- if this is a patient that is going home, an empty
-  -- action queue is not a problem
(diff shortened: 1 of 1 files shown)


lewri:
Firstly thank you for taking an interest in CorsixTH! :)
I assume you mean this patient here in your save:
image

We are aware of this one (and probably should've pinned the issue in hindsight -- I'll do that now). It is fixed by #2310 you'll be pleased to know :)
Any patients that unfortunately got trapped by this will now be properly able to be sent home by #2307

lewri:
Closing, see above comments

JoshKeegan:
Thanks for the quick response @lewri. I hadn't seen that issue :+1:

PR #2315, opened 2023-03-23
try to fix infinite loop when word wider than box Width

lokpro:
https://discord.com/channels/731272375949328486/854282144101040179/1086963524338712576

how to reproduce the bug:
-  use some font and select some game language(maybe other than English) which uses freetypefont.
   my case - mingliu.ttc, Traditional Chinese.
- display a long word "a wwwwwwwwwwwwwwwwwwwwww..." in tips window of home screen.
  ("xxxx..." does not trigger the bug. Need to be "x xxxx...".)
- infinite loop; freeze;

thanks.


lewri:
This PR causes line height to shift, but it's fine enough to resolve the issue by the look of it. Sorry never got around to reviewing this before merge.
image
image

PR #2324, opened 2023-05-04
Use ffmpeg 5.1 new channel layout api

TheCycoONE:
Cleans up deprecation notices


Alberth289346:
Looks ok, but never done anything "sound" so don't know what the details are doing.

PR #2329, opened 2023-05-09
Init Campaigns Directory (fixed)

lewri:
Pull of commits from #2273 due to broken tree.

Segment of #2209 (aka the easy bit)

- Initialises the campaigns directory on launch and adds it as a config line.
- Encompasses part of the prepwork for CommunityContent.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -398,40 +398,32 @@ function App:initGamelogFile()
   fi:close()
 end
 
---! Tries to initialize the user and built in level directories, returns true on
---! success and false on failure.
+--! Tries to initialize the user level and campaign directories
 -- TODO: Integrate other directory initialisations into this function
 function App:initUserDirectories()
   local conf_path = self.command_line["config-file"] or "config.txt"
 
-  --! Attempt to set the user's directory choice
-  --!param dir (path) The defined path of the folder by the user
-  --!param label (string) What folder is being set, for the error dialogue
-  --!return The fully qualified path
+  -- Attempt to set the user's directory choice
+  -- param dir (path) The defined path of the folder by the user
+  -- param label (string) What folder was being set if there was an error
+  -- return The fully qualified path
   local function setUserDir(dir, label)
-    if dir:sub(-1, -1) == pathsep then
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/app.lua
```diff
@@ -399,23 +400,38 @@ end
...
+  --!return The fully qualified path or 'false' if directory creation failed (leading
+  --! to game exit)
+  local function setUserDir(dir, label)
+    if dir:sub(-1, -1) == pathsep then
+      dir = dir:sub(1, -2)
     end
+    if lfs.attributes(dir, "mode") ~= "directory" then
+      if not lfs.mkdir(dir) then
+        print("Notice:" .. label .. "directory does not exist and could not be created.")
+        return false -- Force game exit
```
TheCycoONE, line 419: This use to return false from the outer function, now the field gets set to false instead.

The new comment about forcing the game to exit doesn't appear to be credible, or at least direct.

TheCycoONE:
Would like a little testing (and code cleanup) around the part I commented on. If the outer function is always going to return true, and the return value is never used maybe it shouldn't. If a false directory is overloaded to mean exit game... ug, but point that out in more than just an exit comment on the return false.

lewri:
So the existing behaviour never would stop `app:init()` even though in CorsixTH.lua it looks like it was intended to
https://github.com/CorsixTH/CorsixTH/blob/8c772e5290b495ae33d40ebef4673640a81b4670/CorsixTH/CorsixTH.lua#L143

I've done some testing and even if a directory fails to be set nothing really causes a game-breaking bug, except that some instances may occur where you can't save or load something (because the directory browser loads empty)
image
(note Maps doesn't even exist)

The PR is currently altering that behaviour. If set a failed directory to `""` it becomes root (i.e. C: in Windows). However, that isn't really desired either.

The question is if we want to truly stop loading CorsixTH on a failed directory set/create or not? If not, I can just make the code carry on regardless of its failure and it will just behave like before (empty folder, no load/saves)

lewri:
I've thrown out the returns and any attempts to crash the game for now -- since its original state didn't crash.

PR #2334, opened 2023-05-15
Fix dynamic info on wakecheat

### LINKED ISSUE #2319: [Bug] No fatigue cheat breaks dynamic info
### Describe the issue

After enabling the no fatigue cheat, staff members' dynamic info does not update periodically.

This is because the cheat skips the tire function contents that controls dynamic info updates.
https://github.com/CorsixTH/CorsixTH/blob/d2cc183bda2527798b6c00332c3133575a0a52fd/CorsixTH/Lua/entities/humanoids/staff.lua#L331-L339

Potential ways to resolve
- Run dynamic info updates on tick
- Run dynamic info update whenever its attributes get changed
- Run the dynamic info update tick in tire regardless of cheat

### Steps to Reproduce

1. Load any game

2. Enable the no fatigue cheat

3. Watch for any dynamic info changes that happen organically (e.g. in a training room learning specialisms), the info won't update once a doctor is granted a skill.

### Expected Behaviour

Dynamic info needs to update periodically to show new changes. 

Arguably, the current method of relying on `Staff:tire()` is not the most reliable method. 

### Save Game

### CorsixTH Version

0.66

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

_No response_

### Additional Information

_No response_

lewri:
*Fixes #2319*

*Fixes #*

- Quick fix to resolve no dynamic info updates to staff during nosleep cheat for 0.67. Need to later on look at points raised in https://github.com/CorsixTH/CorsixTH/issues/2319#issuecomment-1513297717

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -332,9 +332,9 @@ end
 -- so amounts here should be appropriately small comma values.
 function Staff:tire(amount)
   -- The no rest cheat overrides tiring effects
-  if self.hospital.hosp_cheats:isCheatActive("no_rest_cheat") then return end
-
-  self:changeAttribute("fatigue", amount)
+  if not self.hospital.hosp_cheats:isCheatActive("no_rest_cheat") then
+    self:changeAttribute("fatigue", amount)
+  end
   self:updateDynamicInfo()
 end
 


Alberth289346:
If you don't want to have the issue closed when merging, use "Addresses" rather than "Fixes".

lewri:
> If you don't want to have the issue closed when merging, use "Addresses" rather than "Fixes".

Do want it closed. I'll raise the comment as a new issue as it's a bigger picture thing.

TheCycoONE:
This is pretty ugly... I suppose if you ticket it immediately

TheCycoONE:
I'd suggest moving the update up to the parent instead

lewri:
> This is pretty ugly... I suppose if you ticket it immediately

Issue is up

PR #2337, opened 2023-05-17
Rewrite empty if branches

tobylane:
- The empty if branches are reversed. This is ignored by luacheck, it's an optional change. Map.lua is easier to read without whitespace changes - https://github.com/CorsixTH/CorsixTH/pull/2337/files?w=1

### DIFF
--- a/CorsixTH/Lua/map.lua
+++ b/CorsixTH/Lua/map.lua
@@ -620,6 +620,46 @@ function Map:setDebugText(x, y, msg, ...)
   self.debug_text[(y - 1) * self.width + x - 1] = text
 end
 
+function Map:_drawDebugTiles(canvas, x, y, xpos, ypos)
+  local xy = y * self.width + x
+  if self.debug_flags then
+    local flags = self.debug_flags[xy]
+    if flags.passable then
+      self.cell_outline:draw(canvas, 3, xpos, ypos)
+    end
+    if flags.hospital then
+      self.cell_outline:draw(canvas, 8, xpos, ypos)
+    end
+    if flags.buildable then
+      self.cell_outline:draw(canvas, 9, xpos, ypos)
+    end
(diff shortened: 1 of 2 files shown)


TheCycoONE:
I like the concept

CorsixTH/Lua/map.lua
```diff
@@ -660,49 +660,50 @@ function Map:draw(canvas, sx, sy, sw, sh, dx, dy)
...
-              end
-              if flags.travelWest and self.debug_flags[xy - 1].passable then
-                self.cell_outline:draw(canvas, 7, xpos, ypos)
-              end
-              if flags.thob ~= 0 then
-                self.debug_font:draw(canvas, "T"..flags.thob, xpos, ypos, 64, 16)
-              end
-              if flags.roomId ~= 0 then
-                self.debug_font:draw(canvas, "R"..flags.roomId, xpos, ypos + 16, 64, 16)
+          if screenX >= -32 then
```
TheCycoONE, line 663: How can we clean this up. I spent a little time looking at I believe to make it really clean you need a scanline iterator like the one in c++ that iterates over tiles diagonally.

what you wrote is:
```
if tile_is_past_left_edge_of_screen then
  if tile_is_before_right_edge_of_screen then
    draw_debug_tile
  else
    break
  end
end
```

It could be a little more manageable if the actual draw_debug_tile logic was moved to a separate function. I suggest doing that, then you aren't indenting nearly as much code.

TheCycoONE (commented):
The first change is definitely better. The second is less obviously an improvement because the function is incredibly messy either way. See my suggestion.

tobylane:
Screen 2023-09-09 at 10 21 48
This is as far as I've got. The flags coverage depends on what area of the map you're looking at. What's missing?

tobylane:
Refactored.a bit more, and removed my scanline attempt.

PR #2339, opened 2023-05-19
Luacheck warnings inlining

tobylane:
- Convert most of the luacheck file wide ignores to line specific
- Fix the other warnings


CorsixTH/Lua/dialogs/menu.lua
```diff
@@ -753,7 +753,7 @@ function UIMenuBar:makeGameMenu(app)
   end
   local function overlay(...)
     local args = {n = select('#', ...), ...}
-    return function(item, m)
+    return function(item, m) -- luacheck: ignore 212
```
lewri, line 756: It's not needed, but was this one meant to have the full inline context as the other 212 ignores have?

lewri (commented):
I'm guessing the functions where you replaced the first argument with `_` will be for compatibility reasons? 

Otherwise my read through the changes seems ok enough

Alberth289346:
I am not sure why parameter information gets destroyed just because the local derived method doesn't use the value.

For plain functions it makes sense, since there is exactly one implementation.

In a class system with derived classes there are several implementations of the function. Keeping the function signature consistent across the classes helps in identifying the functions imho.

CorsixTH/Lua/dialogs/menu.lua
tobylane: That function could be ..., I made this change as I understand it. It's local, no context.

tobylane:
Yes, Luacheck doesn't know of classes. I've not taken parameters out of any class functions, aiming for that consistency.

Alberth289346:
I don't exactly know about all functions, but my guess is that at least the interrupt functions and eg the doKey functions are all stored in the same spot and called in the same manner too.
So while technically not a method of a class, effectively they do have a fixed call signature. (That is, we might add them as methods rather than with ad-hoc callback functions if we would write that code today.)

The original author of a lot of the code preferred plain functions above class methods, or maybe at the time of writing there were no classes (I wasn't there at the time so I don't know).

TheCycoONE:
Why don't we suppress that check globally. It appears more noisy than helpful on our code.

tobylane:
Should the sprite viewer be treated as if it is a class?

Added the ignore. Removed the std because it was set to the default at the time, the default is now 5.1-4 plus JIT.

PR #2340, opened 2023-05-21
Add option to disable demo movie

lewri:
- As some people may find the demo movie playback annoying a config option has been added to disable just that element.


lewri:
Personally I like having the intro play, but not as much the demo, which is why this PR exists.
Alternatives to this:
- Expand the wording of the `play_intro` config option and assume user either wants both enabled or both disabled.
- Write the app.lua check as `if not self.config.play_intro or not self.config.play_demo`, assume intro disablers don't want demo either.

PR #2342, opened 2023-05-24
Reduce use of getLocalPlayerHospital

tobylane:
Replace hospital[1] with the UI's hospital when the UI is the focus of the action.
This was mentioned here https://github.com/CorsixTH/CorsixTH/pull/2203#discussion_r915476462

### DIFF
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -637,7 +637,7 @@ function Humanoid:_handleEmptyActionQueue()
       else
         -- Set these variables to increase the likelihood of the humanoid managing to get out of the hospital.
         self.going_home = false
-        self.hospital = self.world:getLocalPlayerHospital()
+        self.hospital = self.world:getHospital(self.tile_x, self.tile_y)
         self:goHome("kicked")
       end
     end,
--- a/CorsixTH/Lua/humanoid_actions/die.lua
+++ b/CorsixTH/Lua/humanoid_actions/die.lua
@@ -178,7 +178,7 @@ local action_die_tick_reaper; action_die_tick_reaper = permanent"action_die_tick
     grim_reaper:setNextAction(IdleSpawnAction(1660, point_dir):setCount(40))
 
     --Initialise the grim reaper:
-    grim_reaper:setHospital(humanoid.world:getLocalPlayerHospital())
+    grim_reaper:setHospital(humanoid.hospital or humanoid.world:getHospital(hole_x, hole_y))
     grim_reaper.lava_hole = lava_hole
     grim_reaper.lava_hole.orientation = holes_orientation
     grim_reaper.use_tile_x = grim_use_tile_x
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -825,8 +825,9 @@ function Room:crashRoom()
     humanoid.user_of = nil
     -- Make sure any emergency list is not messed up.
     -- Note that these humanoids might just have been kicked. (No hospital set)
+    local hosp = humanoid.hospital or self.hospital or self.world:getHospital(self.x, self.y)
     if humanoid.is_emergency then
-      table.remove(self.world:getLocalPlayerHospital().emergency_patients, humanoid.is_emergency)
+      table.remove(hosp.emergency_patients, humanoid.is_emergency)
     end
     humanoid:die()
     humanoid:despawn()
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -359,10 +359,10 @@ function Machine:updateDynamicInfo(only_update)
   end
   if self.strength then
     if self.world.ui.app.config.new_machine_extra_info then
-      local hosp = self.world:getLocalPlayerHospital()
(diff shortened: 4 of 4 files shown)


CorsixTH/Lua/dialogs/menu.lua
```diff
@@ -689,12 +689,12 @@ function UIMenuBar:makeGameMenu(app)
...
     end, "", function ()
-      if app.world:getLocalPlayerHospital().policies.grant_wage_increase == nil then
-        app.world:getLocalPlayerHospital().policies.grant_wage_increase = app.config.grant_wage_increase
+      if self.ui.hospital.policies.grant_wage_increase == nil then
+        self.ui.hospital.policies.grant_wage_increase = app.config.grant_wage_increase
       end
-      return app.world:getLocalPlayerHospital().policies.grant_wage_increase == grant
+      return self.ui.hospital.policies.grant_wage_increase == grant
     end
   end
```
lewri, line 699: Not you, but blimey that function is a rollercoaster

CorsixTH/Lua/entities/object.lua
```diff
@@ -721,7 +721,7 @@ end
 
 function Object:afterLoad(old, new)
   if old < 52 then
-    self.hospital = self.world:getLocalPlayerHospital()
+    self.hospital = TheApp.ui.hospital
```
lewri, line 724: This afterload is very old. Is it worth changing?

CorsixTH/Lua/humanoid_actions/die.lua
```diff
@@ -178,7 +178,7 @@ local action_die_tick_reaper; action_die_tick_reaper = permanent"action_die_tick
     grim_reaper:setNextAction(IdleSpawnAction(1660, point_dir):setCount(40))
 
     --Initialise the grim reaper:
-    grim_reaper:setHospital(humanoid.world:getLocalPlayerHospital())
+    grim_reaper:setHospital(humanoid.hospital)
```
lewri, line 181: You're now relying here on the humanoid having a hospital set (it should). Makes sense.

CorsixTH/Lua/objects/plant.lua
```diff
@@ -313,7 +313,7 @@ end
 
 function Plant:afterLoad(old, new)
   if old < 52 then
-    self.hospital = self.world:getLocalPlayerHospital()
+    self.hospital = TheApp.ui.hospital
```
lewri, line 316: Again, very old afterload

CorsixTH/Lua/room.lua
```diff
@@ -808,13 +808,14 @@ function Room:crashRoom()
     self.door.reserved_for = nil
   end
 
-  local remove_humanoid = function(humanoid)
+  local remove_humanoid = function(humanoid, room)
```
lewri, line 811: This function already knows about room (self) (see line 822). Not needed but preference would be to rewrite the top line as `local function removeHumanoid(humanoid) ...` but either method is correct.

lewri (commented):
I'm happy with UI classes calling hospital from the UI here. A bit twitchy about non-ui ones.

Also of note: https://github.com/CorsixTH/CorsixTH/discussions/1771
If we want to commit to using the UI to give us the local hospital, I'd suggest doing that in a more concrete way.

CorsixTH/Lua/entities/object.lua
lewri: (there's no harm changing these unless the ui didn't know of hospital at that point of history)

tobylane:
Thanks, I had missed 1771. I will see if I can use World:getHospital(camera position).

TheCycoONE:
Why?

tobylane:
I had AI hospitals in mind, before multiplayer. Though, I expect we will need to gatekeep parts of the UIPatient and UIStaff dialogs. The changes to the afterloads may be too much, especially if we say games with visible AI competitors must be new games.

tobylane:
Some of the previous changes have been taken out, and I rebased the rest.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -64,7 +64,7 @@ function Patient:onClick(ui, button)
     if self.message_callback then
       self:message_callback()
     else
-      local hospital = self.hospital or self.world:getLocalPlayerHospital()
+      local hospital = self.hospital or self.world:getHospital(self.tile_x, self.tile_y)
```
TheCycoONE, line 67: But you presumably can't or shouldn't be able to vaccinate patients at other player's hospitals.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -637,7 +637,7 @@ function Humanoid:_handleEmptyActionQueue()
       else
         -- Set these variables to increase the likelihood of the humanoid managing to get out of the hospital.
         self.going_home = false
-        self.hospital = self.world:getLocalPlayerHospital()
+        self.hospital = self.world:getHospital(self.tile_x, self.tile_y)
```
lewri, line 640: Bit nervous about this now because we're already in a dicey place with an empty action queue. If the humanoid isn't on a hospital tile this would just crash.
image

tobylane:
This shouldn't come before the multiplayer design, Lewri pointed out we can then improve the localplayerhospital function.

PR #2349, opened 2023-06-20
Fix casebook label for alien dna

tobylane:
On a level with Alien DNA, eg L12, this lets the drug casebook tell the player they need a researcher.


lewri (approved):
Fine with me

lewri:
If merged, changelog needs to include this as a bugfix

PR #2363, opened 2023-07-13
[rdy] Add EA App TH Directory

### LINKED ISSUE #2338: Origin is now EA app (Download version of Theme Hospital still usable with CorsixTH?)
In https://github.com/CorsixTH/CorsixTH/blob/master/README.md?plain=1#L23 one of the download version of the original game link to Origin. But the Origin link redirects to a different download page.

`https://www.origin.com/en-gb/store/buy/theme-hospital-origin/pc-download/base-game/standard-edition`
-> `HTTP/2 301`
-> `https://www.ea.com/games/library/pc-download`

That might be the case since October 2022 https://en.wikipedia.org/wiki/Origin_(service)

Is it still possible to use the download version of Theme Hospital (via the EA app) together with CorsixTH? Does someone has an EA account to check that?

The game is available here: https://www.ea.com/games/theme/theme-hospital

lewri:
*Fixes #2338*

- Ensures EA App installs are detected by the game
- I haven't removed the Origin folder for backwards compatibility with older sources


Alberth289346 (approved):
LGTM

Issue #2308, opened 2023-02-14
UIInformation box is loading pink


lewri:
Not sure if this relates to #2279 as it's not present in 0.66 but UIInformation when generated at load is pink.
![image](https://user-images.githubusercontent.com/20030128/218836222-9ad9a0c1-15bf-4125-941b-3f6a3d66a267.png)

Doesn't happen if the save fails to load (error on load)

Does happen if we generate a new UIInformation during gameplay (regardless of very_old_save message or not)
![image](https://user-images.githubusercontent.com/20030128/218837015-9a02460c-7a7c-479a-a999-ce109cb4a591.png)

Edit: Didn't notice before but the tooltip can also load yellow on that box
![image](https://user-images.githubusercontent.com/20030128/225140506-5a0cb6c0-bce7-4220-8643-0c294f6f8ea1.png)

lewri:
(Safe to say, I almost prefer the change in colour, though I imagine this isn't the intended behaviour or method of getting the colour)

lewri:
Another one has popped up when loading a test build from #2310 
![image](https://user-images.githubusercontent.com/20030128/219865643-859d23e3-8cbf-4b86-8f3e-4c84086ec136.png)

mugmuggy:
To add to that last example. If loading the demo one first up. The dialog appears fine. If load a full one, and quit using the FILE menu, then load demo, appears fine. If load a full one up and then go into the FILE menu, and load again the demo (don't quit straight load), dialog does not appear fine.

lewri:
When I loaded a trunk save that happened after 0.66 but before #2279 I get this console warning:
`Warning: palette PREF01V.PAL requested with different flags than stored`

It doesn't happen with a normal run from 0.66

ryanpenfold:
Recieved same issue, but with 0.67-beta1 Windows 11
![image](https://github.com/CorsixTH/CorsixTH/assets/110283510
[Autosave1.zip](https://github.com/CorsixTH/CorsixTH/files/12138408/Autosave1.zip)
/32397eea-3f8a-4536-af8f-79a11d4418c8)
``Warning: palette Award02V.pal requested with different flags than stored``
I've also attached the save file, if you just load it it reproduces above.

TheCycoONE:
```
Loading Town01V.pal with flag
stack traceback:
[...]

PR #2368, opened 2023-07-25
Fixes at least some cases of #2308

TheCycoONE:
loadRaw was not saving the palette transparency flag leading to some mismatches depending on load order.


lewri (commented):
Code looks fine to me, not able to test currently though.

lewri:
I can confirm now that at the least the annual report one is fixed (though never repro'd myself)

PR #2369, opened 2023-07-25
Add a timeout to the update checker

tobylane:
- If there's the false appearance of an internet connection (eg the phone cable is unplugged from the router), the app will hang for several seconds until it times out. This cuts that down to two seconds.


tobylane:
The error message from luasocket changes from "307" to "Invalid argument", but the rest of the message to the user tells them to check their internet connection.

TheCycoONE:
It would be better to suppress entirely. That can be a follow up.

PR #2374, opened 2023-07-31
Changed default config values

### LINKED ISSUE #2371: Default values on new installation

I would recommend setting the default for 'Fullscreen' to 'off' and 'Capture Mouse' to 'off'. Both settings can contribute to a less-than-ideal initial experience for new users.

When I first launched the game in fullscreen, my instinct was to switch it to windowed mode. However, the window appeared on the wrong monitor (I have three) and when I attempted to grab the title bar, the mouse got captured. As you can see, these settings can lead to some initial confusion and frustration.

These settings may not directly affect gameplay, but they could potentially cause annoyance to users. If a player is frustrated, they may be more inclined to just delete the game.

We wouldn't want that to happen, would we? :

Unsure

v.067-rc1

No

### Additional Information

_No response_

RednibCoding:
*Fixes #2371*

*Description of the proposed change:*
The following default config values in config_finder.lua have been modified:
- prevent_edge_scrolling: false -> true
- capture_mouse: true -> false

Additionally, in the Windows installer script Win32Script.nsi, the fullscreen value has been changed from true to false. This change will prevent the installer from launching in fullscreen mode.


CorsixTH/Lua/config_finder.lua
```diff
@@ -105,8 +105,8 @@ local config_defaults = {
   announcement_volume = 0.5,
   play_music = true,
   music_volume = 0.5,
-  prevent_edge_scrolling = false,
-  capture_mouse = true,
+  prevent_edge_scrolling = true,
+  capture_mouse = false,
```
lewri, line 109: Changing these settings no longer matches with the default config file for the windows installer.

lewri (commented):
I agree with changing the Win32Script.nsi line for 0.67; but I wouldn't push the other two line changes for this release.

Could you separate those two changes into separate PRs?

CorsixTH/Lua/config_finder.lua
RednibCoding: Where is the default config file for the Windows installer? I couldn't find any settings related to mouse scrolling or capturing.

lewri: https://github.com/CorsixTH/CorsixTH/blob/master/WindowsInstaller/config_template.txt

lewri (approved):
I'd be happy merging this one for 0.67 @TheCycoONE if you are

PR #2375, opened 2023-08-01
button for mouse panning can be toggled to be the middle or the right mouse button

### LINKED ISSUE #2372: Add ability to change mouse options for panning

In the current game settings, viewport panning is strictly set to the middle mouse button. However, a trend in modern games assigns this function to the right mouse button and uses the middle button for rotation.

To enhance user experience and increase the game's alignment with current standards, it would be beneficial to provide an option to customize these controls. This would allow players to choose between using the middle or right mouse button for viewport panning based on their personal preferences or habits formed by other games

Incorporating user-customizable controls, specifically for viewport panning, it could significantly enhance gameplay experience for several reasons:

1. Familiarity: Many modern games utilize the right mouse button for viewport panning and the middle button for rotation. Players familiar with this control scheme may find it more intuitive to use similar settings in CorsixTH.

2. Comfort: Not all mice are designed with equally accessible middle buttons. For some users, it might be physically more comfortable (me included) to use the right mouse button for frequent actions like viewport panning.

3. Personalization: Allowing users to customize their controls provides a level of personalization that can improve overall player engagement and satisfaction with the game. It demonstrates the developers' commitment to accommodating diverse player preferences and could increase the game's appeal to a broader audience.

While this feature might not directly impact the core gameplay mechanics, it can enhance the overall user experience by making the game more intuitive, comfortable, and personalized.

Unsure

v0.67-rc1

No

### Additional Information

_No response_

RednibCoding:
*Fixes #2372*

- Introduces a new option in the settings menu to toggle the mouse panning button between the middle and right mouse button.
- Includes the same option in the in-game options menu.
- Ensures that the selected option will be saved to the configuration file.
- default is off (middle mouse button)

image

image


lewri (dismissed):
The requested change part here is that this option needs adding into `config.txt` and `config_finder.lua` (default should be middle button imo for users coming over from older versions). The rest is not a necessity from my perspective but hopefully helpful as a review.

Also, we usually don't place a GUI configuration option in both the in-game menu and the options window -- one or the other should suffice. If we want all/more the configuration options in general accessible in-game that should be a different discussion

Good work!

RednibCoding:
I'm a bit unclear about how config.txt and config_finder.lua function exactly. I couldn't locate a config.txt file, so I'm wondering if it is generated automatically. Additionally, I'm unsure about how changing defaults in config_finder.lua affects the actual defaults. Does config_finder.lua only influence the values at the first launch of the game?

I also noticed that changes made in the in-game option menu do not seem to be written to the config. For instance, I tested the capture_mouse option, and after setting it through the in-game options menu, it gets reset upon restarting the game. Is this the intended behavior?

CorsixTH/Lua/game_ui.lua
```diff
@@ -541,6 +541,13 @@ function GameUI:onMouseMove(x, y, dx, dy)
     return false
   end
 
+  local mouse_drag_button = nil
+  if self.app.config.right_mouse_panning then
+    mouse_drag_button = self.buttons_down.mouse_right
+  else
+    mouse_drag_button = self.buttons_down.mouse_middle
+  end
+
```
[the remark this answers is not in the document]
RednibCoding: Thanks for the tip, but this will cause it to always return true when either the middle or right mouse button is pressed, when right_mouse_panning is set to true. However, it will never return true when right_mouse_panning is not set.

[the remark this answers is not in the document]
TheCycoONE: That's not true, and is higher prescedent.
e and r or m

```
erm =
000 0
100 0
010 0
110 1
001 1
...
```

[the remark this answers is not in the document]
TheCycoONE: Not strictly true what you just said but not accurate either if we don't want middle to pan.

CorsixTH/Lua/game_ui.lua
```diff
@@ -784,9 +785,23 @@ function GameUI:playAnnouncement(name, priority, played_callback, played_callbac
...
+  return mouse_scroll_button
+end
+
 function GameUI:onTick()
   local repaint = UI.onTick(self)
-  if not self.buttons_down.mouse_middle then
+
+  
+
+  if not self:_isMouseScrollButtonDown() then
```
lewri, line 804:
```suggestion
  if not self:_isMouseScrollButtonDown() then
```
Luacheck error is here, and we don't need three line spaces imo. Maybe one if thought necessary.

CorsixTH/Lua/game_ui.lua
```diff
@@ -822,7 +837,8 @@ function GameUI:onTick()
       -- If the middle mouse button is down, then the world is being dragged,
       -- and so the scroll direction due to the cursor being at the map edge
       -- should be opposite to normal to make it feel more natural.
-      if self.buttons_down.mouse_middle then
+      -- //...
```
lewri, line 840:
```suggestion

```

lewri (approved):

CorsixTH/Lua/game_ui.lua
```diff
@@ -792,18 +785,23 @@ function GameUI:playAnnouncement(name, priority, played_callback, played_callbac
   self.announcer:playAnnouncement(name, priority, played_callback, played_callback_delay)
 end
 
-function GameUI:onTick()
-  local repaint = UI.onTick(self)
-
-  -- get the mouse button for dragging the map
-  local mouse_drag_button = nil
+-- Get the configured mouse drag button
+function GameUI:_isMouseDragButtonDown()
```
Alberth289346: The function name suggests you obtain a boolean value whether the mouse drag button is down or not.
The documentation says you are getting the configured mouse drag button.

Please make both things say the same.

The idea of documentation is that it briefly describes what the function is doing so you don't have to dig in the code to reverse engineer that information. For this function it's not so bad, but if you have 300 line function (which happens quickly in lua), reading say 3 lines of text and the input/output relation is much quicker than reading 300 lines of details and then hopefully correctly deduce the higher level aim of the intention of the function.

CorsixTH/Lua/game_ui.lua
```diff
@@ -792,18 +785,23 @@ function GameUI:playAnnouncement(name, priority, played_callback, played_callbac
...
   else
     mouse_drag_button = self.buttons_down.mouse_middle
   end
+  return mouse_drag_button
+end
+
+function GameUI:onTick()
+  local repaint = UI.onTick(self)
+
+
```
Alberth289346: One empty line is sufficient (here even 0 empty lines would work)

CorsixTH/Lua/game_ui.lua
```diff
@@ -785,15 +785,15 @@ function GameUI:playAnnouncement(name, priority, played_callback, played_callbac
   self.announcer:playAnnouncement(name, priority, played_callback, played_callback_delay)
 end
 
--- Get the configured mouse drag button
+-- Check whether the configured mouse drag button is being held down (true) or not (false).
```
Alberth289346, line 788:
```suggestion
--! Check whether the configured mouse drag button is being held down (true) or not (false).
```
By starting with a`"--! ` sequence our luadoc tool picks it up and adds it to the generated htnl documentation.

Alberth289346:
Stuff seems fine to me.

lewri:
@RednibCoding I'm content with what you have for this PR, the luadoc suggestion is optional but I wouldn't hold pr back because of it (we can add them later in a cleanup PR).

lewri:
I'm pretty sure the Windows build failures are red herrings here. I'm happy with this as is.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -78,7 +78,7 @@ end
 
 
 function UIOptions:UIOptions(ui, mode)
-  self:UIResizable(ui, 320, 460, col_bg)
+  self:UIResizable(ui, 320, 500, col_bg)
```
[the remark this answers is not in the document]
lewri: #2402

PR #2388, opened 2023-08-18
Update 'check-for-updates' for CorsixTH 0.67

Alberth289346:
*Fixes the 'check-for-updates' file.

- Update the check file for CorsixTH 0.67-


lewri (approved):
Thanks!

Issue #2386, opened 2023-08-16
Ability to customise Jukebox playlist order


MandrakeHorse:
### What feature or enhancement would you like to add?

Currently, it appears that the Jukebox will play Theme Hospital's MIDIs or the user's custom MP3 files in a random order that's determined during startup. Originally, Theme Hospital would always play the soundtrack in the following order:
1. Atlantis
2. 6's and 7's
3. Fortune
4. Candyfloss
5. On The Mend
6. Doctor Frog
7. Night Shift
8. Steady Pulse

The HOSP/SOUND/MIDI directory contains a file called MIDI.TXT which lists all the XMI files and their corresponding track names in the above order, suggesting that this file might be used by Theme Hospital to define the Jukebox's playlist.

There should be a way for the player to specify the order in which songs will play in the Jukebox, such as:
A) providing a configuration file similar to MIDI.TXT that can be edited before launch
B) allowing the player to manually reorder tracks via the Jukebox's UI
C) sorting MP3 files alphabetically (so that the player can specify their order by renaming each MP3 with a numeric prefix)

### How would this feature benefit the gameplay of CorsixTH?

The player would be able to listen to music in their preferred song order.
CorsixTH would be able to replicate the original tracklist for Theme Hospital's soundtrack by default.
Once a playlist has been defined, a shuffle button could potentially be added to the Jukebox later.

### Was the feature in the original game?

Unsure

### What version of CorsixTH are you using?

0.67

### Does this feature/enhancement relate to a current problem?

No

### Additional Information

I asked about this on the CorsixTH Discord's #help channel, and was informed that no sorting appears to be applied to the user-defined music folder: https://github.com/CorsixTH/CorsixTH/blob/fc38c4f46913b7b91c0b1b53efd7c7f2764adb20/CorsixTH/Lua/audio.lua#L61-L182

lewri:
So I think scenario A and C has been done. Correct @tobylane ?
Scenario B needs some control ability added to the UI I imagine -- Nice is drag and drop; less nice is just up/down buttons.

MandrakeHorse:
I can confirm that scenarios A and C have been implemented in 0.68. 

By default, CorsixTH will use the original track order as defined in midi.txt:
![2024-10-14 00_42_43-CorsixTH](https://github.com/user-attachments/assets/b7bb44c1-971a-45c0-9f62-3c6743c637e8)

[...]

PR #2390, opened 2023-08-24
Sort background music by midi.txt or alphabetical order

tobylane:
- Use midi.txt as a playlist, music is played in that order
- If that file isn't present, then order alphabetically

I think the first is expected, the original did it (#2386). The second is slightly more surprising, but if the files aren't numbered, then it's no worse than the current random order.


lewri:
Just to check, we fall back to alphabetical if we use mp3?

Good to know if so, we can work on randomising list order by toggle in the jukebox later down the line I guess.

lewri (commented):
Fine with me (new enhancement was opened for shuffle)

PR #2391, opened 2023-08-24
Permit brackets in config values

tobylane:
- A config value with brackets will be saved, but can't be loaded. This looks for [] and uses [=[ ]=] to contain the whole value. eg A music folder ending [2023], or player name with [].

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -945,13 +945,8 @@ function App:saveConfig()
           -- replace the line, if needed
           handled_ids[identifier] = true
           if value ~= tostring(self.config[identifier]) then
-            local new_value = self.config[identifier]
-            if type(new_value) == "string" then
-              new_value = string.format("[[%s]]", new_value)
-            else
-              new_value = tostring(new_value)
-            end
-            lines[#lines] = string.format("%s = %s", identifier, new_value)
+            lines[#lines] = string.format("%s = %s", identifier,
+                serialize(self.config[identifier], { long_bracket_level_start = 1 } ))
           end
         end
       end
--- a/CorsixTH/Lua/utility.lua
+++ b/CorsixTH/Lua/utility.lua
@@ -311,8 +311,8 @@ function array_join(array, separator)
   return result
 end
 
-local function serialize_string(val)
-  local level = 0
+local function serialize_string(val, options)
+  local level = options and options.long_bracket_level_start or 0
   while string.find(val, ']' .. string.rep('=', level) .. ']') do
     level = level + 1
   end
@@ -391,12 +391,14 @@ end
 --!param options Option settings, table, 'detect_cycles' field boolean that
 --  ends recursion on a cycle, and 'max_depth' integer that ends recursion at the
 --  specified depth. By default initialized with "{detect_cycles = True}"
+--  'long_bracket_level_start' field integer that sets the starting long bracket level for escaping strings.
+--  If not set, level zero is used.
 --!param depth Recursion depth, should be omitted.
 --!param pt_reflist Seen nodes, should be omitted.
 --!return The seralized output.
 function serialize(val, options, depth, pt_reflist)
   if type(val) == "string" then
-    return serialize_string(val)
+    return serialize_string(val, options)
   elseif type(val) == "table" then
     return serialize_table(val, options, depth, pt_reflist)
   else


Alberth289346:
Try String `]=])[=[`.

While you can rewrite for specific forms that you have in mind easily, you cannot easily cover everything that is possible, at least not without writing a full Lua parser. Simple string matching is simply not powerful enough to catch everything.

For a robust solution there are in general two ways out
- Find a library that properly escapes strings that look like valid or invalid Lua. I don't know if it exists at all. If it does and is any good, it will beat any home-made solution.
- Give up on the idea that the config file must be a Lua-parseable file, use a well-defined text-oriented format instead.
  - INI format is not difficult if you handle or forbid some "weird" strings
  - a line-oriented format `name = value` (ie INI without heading). Properties file are a generalization of this.
  - Use a library to get  a "heavier" format from JSON to XML. Many formats exist, many are not very friendly for manual editing of the file.
  - I am sure I am not complete.

tobylane:
I considered a whitelist, but that would be impossible to internationalise. This fixes a real world problem, rather than everything that could be written to cause issues. Would the issue turn up whenever the string is brought, eg from an .ini, into Lua?

Alberth289346:
> Would the issue turn up whenever the string is brought, eg from an .ini, into Lua?

If you write a good INI parser, it wouldn't.

Longer explanation:  
At the disk, all files are just sequences of bytes. The file data itself has no idea what it represents, we however attach meaning to the bytes by considering files to be of some "kind", and loading different kinds of files into different programs. Eq a PDF file is loaded into a PDF viewer, a Lua file is loaded into a Lua interpreter, and so on.  

In the above, "loading" is more than just grabbing the bytes from the disk and putting them in memory. In addition the file is "parsed". A parser is basically a function that converts the byte sequence from the file into [[ assuming a Lua environment here ]] a collection of tables, strings, numbers, etc such that its data can be easily processed by other code.

Different parser functions have different rules that they use for the conversion process. For example a `.txt` file parser groups bytes between `\n` characters, converts each byte in a group to a character [[ assuming ASCII here ]] , merges all characters of a group into a string, and collects all strings into an array. The result of that parser is thus an array of string as you would expect when you think a file to be a text file.  
A PDF file parser or a Lua parser or a EXE parser all have different more complicated rules. As a result, while they all start from a sequence of bytes as input, their results are hugely different.

What we are doing with the config file is that in the past is that we defined that we want to use the Lua parser rules to convert the configuration file from the disk into data usable by the CTH program. That means for **creating a configuration file**, we must write bytes to the file such that they get parsed in the way we want using Lua parsing rules, when in the future that file is loaded again.  
For simple identifier-like names and values this is not difficult. Lua parser conversion rules regarding such characters are quite simple. However when you want to extend beyond such simple names/values, other rules in the Lua parser suddenly apply, and before you know it, the result given to you by the Lua parser are not strings and simple values any more (since the Lua parser is capable of producing many other tables/strings/number collections).  
If you want to avoid that from happening, you basically have to "work around" all Lua parsing rules, and due to the complexity of the Lua parser rules that runs out of hand pretty quickly. (To succeed in all cases, you have to take all Lua parsing rules into account while deciding what bytes to store in the file.)

The simpler alternative at that point is then to switch to a different/simpler rule set for parsing the file back into strings and values at the time of loading the file again. That is, use a different parser function (or write one, of course so you can have any rule set you want).
In other words, if you pick a different parser function that cannot produce a Lua function or a Lua class (like the Lua parser can), you obviously never going to get a function call or class from that different parser function while loading the file again.

The INI file parser rules are much simpler, and thus it's much easier to ensure that data you want to store in a file is properly read back again (Less or no rules in the parser that can produce "wrong" results while reading the file back again.)

The alternative of inventing your own parser rules is also an option (and can be fun), but the disadvantage of that is that nobody knows the rules of it but you, you cannot simply refer to some Internet document and say "this is the rule set that is used". Also, there are other programs that can parse INI files, by switching to that ruleset we can use all those programs without any effort from us.

TheCycoONE:
I think toml makes some sense here as a better defined ini with libraries for lua (https://github.com/LebJe/toml.lua) or c (https://github.com/cktan/tomlc99) and c++.

Alberth289346:
TOML is also considered to be a bad language as it's too smart for its own good, so for mere humans it generates a lot of gotchas. 

An examples that I read:  List of countries `[en, nl, no, ca]`, except `no` is a boolean and not a string.

Further reading:
https://github.com/madmurphy/libconfini/wiki/An-INI-critique-of-TOML 
https://hitchdev.com/strictyaml/why-not/toml/


Arguably we don't even really need ini files today, just a list of "key = value" lines is sufficient today. However, it's commonly known, and gives us sections that may be useful tomorrow. (ie trying to avoid changing the config file format again in the future).

I can also live with the properties file format though, with multi-level keys, eg
```
graphics.dir = /....
debug = false
audio.volume = 28
audio.music_dir = /...
```
but it's less intuitive to edit that manually I think.

Alberth289346:
I guess the idea is that the semantics of a value in the configuration should depend on its key rather than the value itself, so you can have
```
debug = no
country = no
```

TheCycoONE:
I think we can stick with lua with the bracket fix. No one has complained except us.

TheCycoONE:
The lua serializer routines in utility should be used though and should support arbitrary brackets - i believe it already does.

tobylane:
With https://github.com/Dynodzzo/Lua_INI_Parser/blob/master/LIP.lua it is possible to switch to ini, but it loses comments and order. To keep the current write config functions, while loading it with that parser, doesn't seem good.

This uses serialize, with a change to serialize_string.

Alberth289346:
First things first, I am not sure you get the suggestion by TheCycoONE. That suggests to keep the current implementation but use our `utility.serialize` function, since it's "Lua interpreter proof" apparently.

If you want to stick with the current format, that's a good idea. Depending on how you use it, it may cause some changes what you write tho the file. For example there is currently escape text written to the file iirc, and that should not be needed anymore.  
A downside of using `serialize` *could* be readability of the file, but I am not sure how bad that is for the relative simple values that we store. Probably the code needs a few other changes (ie delete the current escape magic), but no idea how much impact that has.

----
If you want to go the INI file direction instead, the link that you found looks like what I'd expect for an INI file reader/writer indeed.

It's pretty common that a random implementation doesn't exactly do what you want. There are 3 ways to solve that:
- Just adopt it as-is, and live with its limitations and features.
- embrace and modify / extend the code. Add the code in a commit to ensure the exact imported version is preserved, and then add commits to change it to your liking. Technically, by doing that it becomes an upstream dependency with modifications, but I am not sure if it's worth tracking upstream changes in this case.
- consider the found implementation as insufficient, and find another one or
- build your own. The link gives a good impression of how many lines you'll need, depending on "extra features" this may get larger or smaller.

Regardless of what path you pick, it's quite normal that more things will change in the code. Move one thing, and everything around needs some work too.

You currently seem very focused to preserve the current format and code as much as possible. It's a valid choice, but it's not the only possible goal and possibly also not the best goal.  I don't know if better alternatives exist here, but in my experience it pays to throw a somewhat larger fishing net on the problem and see what you catch. Ponder a bit about 
- advantages and disadvantages of the current solution. (Our current solution seemed a like good idea at the time it was implemented, but it may not be good anymore now.)
- advantages and disadvantages of alternative solutions (what other ways exist, what parts get covered then and what parts need another solution then? How complicated are the latter parts?)

In the worst case, you conclude that the current solution is indeed the best path. However with the above exercise, you get more informed knowledge why that is the case. Otherwise, you'll find a better way to implement the change.

> it is possible to switch to ini, but it loses comments and order. To keep the current write config functions, while loading it with that parser, doesn't seem good.

You seem to point out new problems that happen if we use the code in the found link, with respect to the current solution. That's a  good step.

As a little nudge, I'll give you a few next questions based on your findings.

- Is it true that INI has no comments? (The found implementation doesn't support them, but its author may have decided that they are not relevant for its application.)

  Try finding other implementations, or find a definition of the format. (The latter is better as it's the definitive source what the format supports).

  <details><summary>SPOILER</summary>  wikipedia says the `;` starts a line comment . </details>

- Is it bad that we lose comments in the file? Are they useful to keep? Are they needed there or can they get moved to elsewhere? If you want to keep them, should all comments be preserved or can we discard them and write new comments every time?

  <details><summary>SPOILER</summary> It seems to assume the user will manually edit the file. How many users will do that? Is the wiki a better place for describing the meaning of settings? </details>

- Is it bad that you lose order of settings? If so, how can that be avoided or eliminated, eg could sort work or would using the sections feature of the INI format b useful? What would it take to preserve the entire content of the INI file in a read+write cycle?

tobylane:
An ini that keeps comments and order is just a custom config that resembles ini, which seems like no progress at all.

Minor change to the option name. Ready.

Alberth289346:
> An ini that keeps comments and order is just a custom config that resembles ini, which seems like no progress at all.

Iirc the topic is not about avoiding a custom config file implementation, but about not having to outsmart  the Lua interpreter when you write the configuration file.

Alberth289346:
To extend a bit on comments in the config file:

- If you assume that all users perform configuration within the program, there is no need to have comments in the file at all. Practically nobody is going to read that file. An explanation at the wiki would suffice.
- If you assume that all users perform configuration in the file itself, there is no need to ever (over-)write the file from the program. Users can add comments if they want.

Because we want both solutions, the need to preserve comment exists.

CorsixTH/Lua/utility.lua
```diff
@@ -391,12 +391,13 @@ end
 --!param options Option settings, table, 'detect_cycles' field boolean that
 --  ends recursion on a cycle, and 'max_depth' integer that ends recursion at the
 --  specified depth. By default initialized with "{detect_cycles = True}"
+-- 'start_at_one' field boolean that starts the count in serialize_string at 1 not 0.
```
Alberth289346, line 394: - The documentation is wrong. The entire point of function documentation is to explain what effect parameter values have, *without* having to read the code of the function. This documentation however doesn't explain anything, it just says "see code for understanding what it does".

- The proposed `start_at_one` addition excludes by design any other value as start level. Why? What makes `1` the one and only value that we ever need?

CorsixTH/Lua/utility.lua
```diff
@@ -311,8 +311,8 @@ function array_join(array, separator)
   return result
 end
 
-local function serialize_string(val)
-  local level = 0
+local function serialize_string(val, options)
+  local level = options and options.start_at_one and 1 or 0
   while string.find(val, ']' .. string.rep('=', level) .. ']') do
     level = level + 1
   end
```
Alberth289346, line 318: Not your fault, bu repeatedly walking over a potentially long string is  a bad idea. The better solution is to catch all level in one search by using a regular expression.

CorsixTH/Lua/app.lua
```diff
@@ -945,13 +945,8 @@ function App:saveConfig()
...
           handled_ids[identifier] = true
           if value ~= tostring(self.config[identifier]) then
-            local new_value = self.config[identifier]
-            if type(new_value) == "string" then
-              new_value = string.format("[[%s]]", new_value)
-            else
-              new_value = tostring(new_value)
-            end
-            lines[#lines] = string.format("%s = %s", identifier, new_value)
+            local new_value, options = self.config[identifier], { start_at_one = true }
```
Alberth289346, line 948: Why did you keep `new_value`  as variable name? It never contains a new value after your changes.

CorsixTH/Lua/utility.lua
tobylane: https://www.lua.org/manual/5.3/manual.html#:~:text=define%20an%20opening-,long%20bracket,-of%20level%20n suggests the field name long_bracket_level_start or initial count.

Alberth289346: The page is more about the precise definition of such brackets (it's almost math). The reason they do that is because a few lines further they start discussing "brackets of another level", which doesn't make sense unless you first explain the level of a single long bracket.

So in my view that page is about the formal definition of Lua concepts. (What is a long bracket, how do you decide the level of such a bracket, what happens if you have several such brackets?)   
While i don't think it suggests exactly that name, but if you like the name, I am fine with it.

CorsixTH/Lua/utility.lua
```diff
@@ -391,12 +391,14 @@ end
 --!param options Option settings, table, 'detect_cycles' field boolean that
 --  ends recursion on a cycle, and 'max_depth' integer that ends recursion at the
 --  specified depth. By default initialized with "{detect_cycles = True}"
+-- 'long_bracket_level_start' field integer that sets the starting long bracket level
+--  if not set, zero is used.
```
Alberth289346, line 395:
```suggestion
--  'long_bracket_level_start' field integer that sets the starting long bracket level for escaping strings.
--  If not set, level zero is used.
```

Alberth289346 (approved):

lewri:
This one is stuck for merge for some reason. May need a rebase.
@TheCycoONE are you happy with the changes?

PR #2394, opened 2023-09-03
FIX Level advance is possible while owing a loan #2027

### LINKED ISSUE #2027: Level advance when owing a loan
On
corsix-th-0.64-1.fc32.x86_64

Even when owing a loan, I received an invitation to advance level. This doesn't happend on the original ThemeHospital game, which requires to pay all debts before this.

nMustaki:
*Fixes #2027*

Fix issue #2027 by ensuring loans are payed before winning is possible

- If the players has loans and the state is "win", force it to "nothing"

Disclaimer:
I cannot compile yet on my computer, so I couldn't test

PS: I remember when the project was started, and it's great to see it's still moving on :-)

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1463,16 +1463,24 @@ function World:checkWinningConditions(player_no)
     end
     if goal.win_value then
       local max_min = goal.max_min_win == 1 and 1 or -1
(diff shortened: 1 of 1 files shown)


lewri:
Thank you for your contributions (both PRs) :)

PR #2398, opened 2023-09-04
Look for a bundled Noto CJK font

### LINKED ISSUE #823: Automatically select unicode font from system if available
It would be a better experience for our international users if their system default font could be picked up and used as the unicode_font if it's available and a supported type. (e.g. true type). This way their preferred language option would be visible without navigating through the font selection screens.

### LINKED ISSUE #1219: Create Corsix's own unicode font
#### Let's create a font

Any font artists around here maybe? 

I'm trying to use Polish in Corsix, and I managed to use it with some ```.ttf``` fonts, but none of them look good to be honest. Maybe we could try, as community, to create our own unicode font, looking very similar to TH original font (kinda like openTTD did with graphics/sounds)?  Or maybe there is some free font that we could use right now? Arial doesn't work so well.

That would probably fix other issues: https://github.com/CorsixTH/CorsixTH/issues/1213 for example.

I know it's very general idea, but I'm new to this community, and I didn't see anyone mentioning creating own font before.

### LINKED ISSUE #1901: Bundle international fonts and allow users to select all translations available

tobylane:
- Look for a bundled Noto CJK font, or a packager chosen bundled font path or a system font path. The official Noto project doesn't release a suitable single file, but there are some third party builds that do. [satbyy](https://github.com/satbyy/go-noto-universal/releases/tag/v7.0) has [GoNotoKurrent-Bold.ttf](https://github.com/satbyy/go-noto-universal/releases/download/v7.0/GoNotoKurrent-Bold.ttf), I chose this because the licence on this project is clearer.
Other font options include [KlokanTechNotoSansCJK-Regular.otf from klokantech](https://github.com/klokantech/klokantech-gl-fonts) or [GoNotoKurrent-Regular.ttf from satbyy](https://github.com/satbyy/go-noto-universal).
- The new priority is user setting > bundled font > system font. Linux users from before this change will have an old user setting by default that is probably worse than Noto for CJK.

Korean preview
Screen 2023-09-04 at 20 56 39

Fixes #1901 fixes #823 fixes #1219

### DIFF
--- a/CorsixTH/Lua/graphics.lua
+++ b/CorsixTH/Lua/graphics.lua
@@ -117,38 +117,50 @@ function Graphics:Graphics(app)
 end
 
 --! Tries to load the font file given in the config file as unicode_font.
---! If it is not found it tries to find one in the operating system.
+--! If it is not found it tries to use the font file from the compile setting WITH_FONT,
+--!  or look for one in the CorsixTH folder, the grandparent of this file.
 function Graphics:loadFontFile()
   local lfs = require("lfs")
-  local function check(path) return path and lfs.attributes(path, "mode") == "file" end
-  -- Load the Unicode font, if there is one specified.
-  local config_path = self.app.config.unicode_font
-  -- Try a font that commonly comes with the operating system.
-  local os_path, font_file
-  local windir = os.getenv("WINDIR")
-  if windir and windir ~= "" then
-    os_path = windir .. pathsep .. "Fonts" .. pathsep .. "ARIALUNI.TTF"
-  elseif self.app.os == "macos" then
-    os_path = "/Library/Fonts/Arial Unicode.ttf"
-  else
-    os_path = "/usr/share/fonts/truetype/arphic/uming.ttc"
+  local function check(path) return lfs.attributes(path, "mode") == "file" end
+  local config_path, compile_path = self.app.config.unicode_font, TH.GetCompileOptions().font
+
+  local function getFontPath()
+    local config_err, compile_err = "", ""
+    if config_path then
+      if check(config_path) then return config_path
+      else config_err = "Configured font set but not found. Check path " .. config_path
+      end
+    end
+    if compile_path then
+      if check(compile_path) then return compile_path
+      else compile_err = " Compiled font path set but not found. Check path " .. compile_path
+      end
+    end
+    local path = self.app:getFullPath({})
+    for file in lfs.dir(path) do
+      for _, ext in pairs({"%.ttc$", "%.ttf$", "%.otc$", "%.otf$"}) do
+        if file:match(ext) then
+          return path .. file
+        end
+      end
+    end
+    return nil, config_err .. compile_err
   end
-  if check(config_path) then font_file = config_path
-  elseif check(os_path) then
-    font_file = os_path
-    print("Configured unicode font not found, using " .. font_file .. " instead.")
-    print("This will be written to the config file.")
-  elseif config_path ~= nil then
-    print("Configured unicode font not found, no fallback available.")
+  local font_file, err = getFontPath()
+  if not font_file then
+    print("Unicode font not found, no fallback available.", err)
     return
   end
-  local font = font_file and io.open(font_file, "rb")
+
+  local font = io.open(font_file, "rb")
   if font then
     self.ttf_font_data = font:read("*a")
     font:close()
-    if self.ttf_font_data and self.app.config.unicode_font ~= font_file then
+    if self.ttf_font_data and config_path ~= font_file then
       self.app.config.unicode_font = font_file
       self.app:saveConfig()
+      print("Configured unicode font not found, using " .. font_file .. " instead.")
+      print("This will be written to the config file.")
     end
   end
 end
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@
 # CorsixTH Features and behaviour
 #   - WITH_AUDIO      : Activate sound (yes)
 #   - WITH_FREETYPE2  : Active support for non-Latin script alphabets (yes)
+#   - WITH_FONT       : Path of a font that will be used if the user does not set one (empty, Arial on macOS)
 #   - WITH_MOVIES     : Activate movies (requires with_audio, FFmpeg) (yes)
 #   - WITH_UPDATE_CHECK : Activates support to check for new version on launch (requires libcurl) (yes)
 #   - USE_SOURCE_DATADIRS : Use the source directory for loading resources. Incompatible with the install target (no)
@@ -90,6 +91,15 @@ endif()
 option(SEARCH_LOCAL_DATADIRS
   "Search resources in the working directory and the program directory where the executable stores."
   ${SEARCH_LOCAL_DATADIRS_DEFAULT})
+
+if(WITH_FREETYPE2)
+  if(APPLE)
+    set(WITH_FONT_DEFAULT "/Library/Fonts/Arial Unicode.ttf")
+  else()
+    set(WITH_FONT_DEFAULT "")
+  endif()
+  set(WITH_FONT ${WITH_FONT_DEFAULT} CACHE FILEPATH "Use this font if one is not set in the game settings")
+endif()
 
 # Dependency management
 if(UNIX AND CMAKE_COMPILER_IS_GNU)
--- a/CorsixTH/CMakeLists.txt
+++ b/CorsixTH/CMakeLists.txt
@@ -67,6 +67,11 @@ else()
   set(CORSIX_TH_ARCH ${CMAKE_SYSTEM_PROCESSOR})
 endif()
 
+# Set default value of font file in compile_opts.font
+if(WITH_FONT)
+  set(CORSIX_TH_FONT ${WITH_FONT})
+endif()
+
 # Ensure config.h is picked up by cmake - moving this into subdir cmake files will
 # prevent it applying to the CorsixTH project
 set(CorsixTH_generated_src_dir ${CMAKE_BINARY_DIR}/CorsixTH/Src/)
--- a/CorsixTH/Lua/api_version.lua
+++ b/CorsixTH/Lua/api_version.lua
@@ -32,4 +32,4 @@ Note: This file compiles as both Lua and C++. */
 
 #endif /*]] --*/
 
-return 2690;
+return 2691;
--- a/CorsixTH/Src/config.h.in
+++ b/CorsixTH/Src/config.h.in
@@ -105,4 +105,7 @@ using std::uint64_t;
 /** Report system architecture **/
 #cmakedefine CORSIX_TH_ARCH "@CORSIX_TH_ARCH@"
 
+/** Set default value of font file **/
+#cmakedefine CORSIX_TH_FONT "@CORSIX_TH_FONT@"
+
 #endif  // CORSIX_TH_CONFIG_H_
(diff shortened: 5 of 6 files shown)


lewri (commented):
For the time being, I don't think we should remove the ability to set font manually. Instead, just let it sit as the default variable.

Also these changes probably break the font dialog.

TheCycoONE:
Prefer not to bundle in the repo without first debating/setting up git lfs. It's sufficient to make it a packaging requirement.

tobylane:
Users's setting > bundled font > OS font. This means if the user has opened CorsixTH on Linux before, the incomplete OS font will be preferred. Line 138 could only choose the setting if it isn't the path two lines above?

I tried to have it pick up the youngest ttf file in the folder, but FileTreeNode isn't available.

lewri (commented):
It's worth noting here, ARIALUNI.TTF was a Microsoft Office font (not Windows) and hasn't been included since Office 2016. This would explain why gfx:loadFontFile() never wrote a unicode font to the config.txt since that time.
https://learn.microsoft.com/en-us/typography/font-list/arial-unicode-ms

lewri:
> Users's setting > bundled font > OS font. This means if the user has opened CorsixTH on Linux before, the incomplete OS font will be preferred. Line 138 could only choose the setting if it isn't the path two lines above?
> 
> I tried to have it pick up the youngest ttf file in the folder, but FileTreeNode isn't available.

Why not make it an explicit setting?

tobylane:
I think Arial and Noto are the only pan-CJK fonts. [Google font search](https://fonts.google.com/?preview.text=%EB%B3%91%EC%9B%90%EC%9D%B4%20%EB%AC%B8%EC%9D%84%20%EC%97%B4%EC%97%88%EC%8A%B5%EB%8B%88%EB%8B%A4&preview.text_type=custom&sort=popularity&subset=korean&noto.script=Kore) for the first Korean string in the language file, where after the Noto fonts they are Korean or Korean+European only, not including Chinese and Japanese.

Could we have the non-free Arial package recommended or required by linux packages? Ask the packagers to set the other-os line to the location of the font that their distribution installs Arial to? eg https://packages.debian.org/buster/ttf-mscorefonts-installer

> Why not make it an explicit setting?

Ideally it isn't a user decision at all, depending on how close we can get to an answer for everyone. Strictly free-only (avoiding debian non-free) users who don't like Noto may be left out.

lewri:
We can't rely on Arial Unicode any more in Windows. In Mac, it's still okay (it is in the Font Library). I'd support cyco's suggestion to make it a packaging requirement. Worst case is that if a suitable unicode font is missing unicode languages are just hidden until a font is selected.

Alberth289346:
It should be possible to extract all code points from all strings in a language, and verify that they exist in the font.

When to do that, and what to do if indeed code points are missing are next choices.

tobylane:
Rebased, and put the two new variables at the top of the file for packagers to edit.

lewri:
I think part of the idea is to also change our make files to require the font too as an optional dependency. At the least, we'd need to add more directions to the make instructions say we want to use x specific noto bundle.

>  Remove references/settings about picking a font file. Do we want this option left in?
Did you put this back in now? If we want to encourage Noto you may need to add something perhaps to the config file and/or Wiki we're using Noto as the default, but if your language doesn't show properly for you you can specify another?
Do you know what specific languages we currently support that aren't covered by this bundle?

tobylane:
> Did you put this back in now?

I did, and have updated the title and post.

> If we want to encourage Noto you may need to add something perhaps to the config file and/or Wiki we're using Noto as the default, but if your language doesn't show properly for you you can specify another?

I can draft a wiki page. The config file, and perhaps the folder settings tooltip, could say something like leave empty for a font chosen for you? Should the folder settings run `Graphics:loadFontFile()` when the font setting is cleared to find what it can?

> Do you know what specific languages we currently support that aren't covered by this bundle?

So far as I've seen, all are supported by Noto CJK.

lewri (approved):
Once merged I would suggest adding to the 'How to Compile' wiki page. We can add (yet another :) ) packaging note on release too.

I appreciate this PR is now only a 'look for', but it definitely is going to need its supporting docs where appropriate.

TheCycoONE:
I feel like we should favour a system noto if that's what we're packaging, over the system arial if available.

This is starting to feel similar to the hospital lookup function.

lewri:
> I feel like we should favour a system noto if that's what we're packaging, over the system arial if available.
> 
> This is starting to feel similar to the hospital lookup function.

Arial unicode is heavily deprecated/removed in Windows. Though that isn't (currently) what this PR is about now.

tobylane:
[Debian](https://packages.debian.org/bookworm/fonts-noto-cjk) follows the official project in not combining the font files. We could have the language file give the suffix (JP, KR, SC, TC from the debian page) for the right file for each non-European language select, but I don't like the sound of that.

tobylane:
Strings are loaded after and depend on Graphics. Strings can load the font file suffix and make the switch later, but that means setting the font twice.

[NotoSansCJK-VF.otf.ttc](https://github.com/notofonts/noto-cjk/tree/main/Sans#variable-otcs) (33mb) is another good option, but [not packaged](https://packages.debian.org/search?searchon=contents&keywords=NotoSansCJK-VF.otf.ttc).

lewri:
@tobylane are #1901 and #823 resolved or partially resolved at all by this PR? If so, could you edit the summary so they point respectively?

tobylane:
Updated, also #1219.

CorsixTH/Lua/graphics.lua
```diff
@@ -18,8 +18,12 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
-local TH = require("TH")
+-- The location of a bundled font relative to this file
+local bundled_path = TheApp:getFullPath({"GoNotoKurrent-Bold.ttf"})
+-- The location of a globally installed font
+local installed_path = "/usr/share/fonts/truetype/arphic/uming.ttc"
```
lewri, line 24: Is this global? The path given is likely only *nix global. It seems little odd why you isolated this path though from the other OS paths.

tobylane: It's for *nix packagers to change, I can make that more clear here or in the wiki. It's likely that *nix packagers will put a font at a global path and Windows packagers will put a font at a local path.

CorsixTH/Lua/graphics.lua
```diff
@@ -18,8 +18,12 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
-local TH = require("TH")
+-- The location of a bundled font relative to this file
+local bundled_path = TheApp:getFullPath({"GoNotoKurrent-Bold.ttf"})
```
lewri, line 22: #2565  -- Regular may be better looking at screenshot comparison

TheCycoONE:
Why are we setting default fonts in two different places, one spot to Arial, and the other to Noto/uming.

tobylane:
The order of preference is user setting > bundled > globally installed. The bundled font, if it exists, should be Noto. The globally installed font is Arial on Windows, which is only supplied by Office; Arial on macOS is always installed; uming or other packager choice on anything else. Is there a better way to find the packagers choice?

[Microsoft Yahei](https://learn.microsoft.com/en-us/typography/font-list/microsoft-yahei) is the CJK font in Windows 11, but it doesn't have the Korean for Korean.

CorsixTH/Lua/graphics.lua
TheCycoONE: This is hard coded, how did you expect packagers to change it?

tobylane: In a distro specific patch. Debian does it for their game data path, https://salsa.debian.org/games-team/corsix-th/-/tree/debian/master/debian/patches?ref_type=heads

If the Regular weight is chosen, this path should be /usr/share/fonts/opentype/noto/NotoSansCJK-Regular.ttc for Debian and Ubuntu, leaving the different distros like Arch to change it to their path, /usr/share/fonts/noto-cjk/NotoSerifCJK-Regular.ttc.

tobylane: The distro patch could just edit the default config, and have the packager set path as the initial value. In that case, there's no need for an installed_path variable.

CorsixTH/Lua/graphics.lua
```diff
@@ -131,22 +135,23 @@ function Graphics:loadFontFile()
   elseif self.app.os == "macos" then
     os_path = "/Library/Fonts/Arial Unicode.ttf"
```
TheCycoONE, line 136: No, why would we make it not work on macos? - actually why are we still overriding font in graphics.lua at all?

CorsixTH/Lua/graphics.lua
TheCycoONE: I don't want distros to have to patch our code. A cmake variable would be appropriate.

TheCycoONE (changes requested):

tobylane:
The cmake WITH_FONT option shouldn't be an option, as those are boolean, so it's now a variable. The lua side has been reworked, no change in functionality. Rebased and ready.

PR #2400, opened 2023-09-08
[RFC] Fix array indexing out of bounds in parcel adjacency calculation

alanwoolley:
The adjacency matrix calculation, which is responsible for working out which plots are should be available for purchasing, looks to be off by an index due to the way the pointer is incremented prior to performing the test. This results in undefined behaviour when testing the edge of the map tiles. It's probably never been noticed because there's many tiles per parcel.

I've rewritten this to use some of the existing methods to get map tiles rather than pointer magic, which might be a tad slower but as this code is infrequently run, I think the trade-off for readability is worth it.

I've tested this change with all the original campaign levels, and they seem to be purchasable in the order I'd expect.

### DIFF
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -722,14 +722,16 @@ std::vector<std::pair<int, int>> level_map::set_parcel_owner(int iParcelId,
 namespace {
 
 void test_adj(bool* parcel_adjacency_matrix, int parcel_count,
-              const map_tile* original_node, int xy, ptrdiff_t delta) {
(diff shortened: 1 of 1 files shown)


lewri:
Appveyor was successful, their server was just having a bad day

alanwoolley:
Thanks @lewri. That's right, you should only be able to purchase plots that are not blocked by a plot you have not yet purchased, if that makes sense.

Alberth289346:
Took a while to understand how it was wrong, but yeah, line old 753 looks incorrect.

Even if the old code was not wrong, merging this patch is a good idea, since the new code is much simpler to understand.

TheCycoONE:
Sorry for the delay. I missed this one! Thanks.

PR #2416, opened 2023-09-27
Advice player of new machine cost

### LINKED ISSUE #2318: Can't see price of machine upgrade unless you can afford it

If you *can* afford a machine upgrade, clicking the cog icon on the machine's dialogue shows you the price when confirming whether you want to proceed with the upgrade:

Screenshot from 2023-04-17 21-17-03

If you can't afford it, there's no way to find out how much it costs. It just ignores you clicking the cog, but the announcer message shows "You don't have enough money in the bank to make that purchase":

Screenshot from 2023-04-17 21-17-53

It would be nice if it showed either on the cog button (maybe on hover), after clicking the cog (in a similar box to the confirm purchase one), or perhaps in the bottom section where it tells you the machine's strength:

Screenshot from 2023-04-11 23-46-11

Alternatively, a menu dedicated to machine upgrades would be handy - list all machines and their current vs potential strength, upgrade cost etc.

It's difficult not knowing how much an upgrade costs, and the cog button seemingly doing nothing is also frustrating. Showing the price would be useful to players. All other costs of rooms and items is given, regardless of affordability.

No

0.66

No

### Additional Information

_No response_

lewri:
*Fixes #2318*

- Tell the player the cost of replacing their machine via the adviser if they can't afford it.
- Also adds ability for adviser to be given multiple arguments for string.format

Using the more flexible `%[1-9]%` argument format so translators can rephrase the string if needed.

### DIFF
--- a/CorsixTH/Lua/strings.lua
+++ b/CorsixTH/Lua/strings.lua
@@ -331,10 +331,11 @@ function Strings:setupAdviserMessage(messages)
     --build_advice
     --cheats
   }
+  -- Passes adviser message through string.format process
   local formatFunc
-  formatFunc = function(format_self, arg)
+  formatFunc = function(format_self, ...)
     -- After 'format', it is not useful to have indexing magic anymore.
-    return { text = format_self.text:format(arg), priority = format_self.priority }
+    return { text = format_self.text:format(...), priority = format_self.priority }
   end
   local indexFunc
   indexFunc = function(index_self, field)
--- a/CorsixTH/Lua/dialogs/machine_dialog.lua
+++ b/CorsixTH/Lua/dialogs/machine_dialog.lua
@@ -108,7 +108,8 @@ function UIMachine:replaceMachine()
   local cost = hosp.research.research_progress[machine.object_type].cost
   if hosp.balance < cost then
     -- give visual warning that player doesn't have enough $ to buy
-    self.ui.adviser:say(_A.warnings.cannot_afford_2, false, true)
+    local advice = _A.warnings.cannot_afford_machine:format(cost, machine.object_type.name)
+    self.ui.adviser:say(advice, false, true)
     self.ui:playSound("wrong2.wav")
     return
   end
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -238,6 +238,7 @@ adviser = {
     another_desk = "You'll need to build another desk for that new receptionist.",
     cannot_afford = "You don't have enough money in the bank to hire that person!", -- I can't see anything like this in the original strings
     cannot_afford_2 = "You don't have enough money in the bank to make that purchase!",
+    cannot_afford_machine = "You need at least $%1% in the bank to afford a new %2%!",
     falling_1 = "Hey! That is not funny, watch where you click that mouse; someone could get hurt!",
     falling_2 = "Stop messing about, how would you like it?",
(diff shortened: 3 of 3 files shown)


Alberth289346:
> Using the more flexible %[1-9]% argument format so translators can rephrase the string if needed.

I am confused about this. Why are you adding numbered parameters in the original string, you don't seem to need them and they are not obligatory afaik either.

I would expect that strings that don't need them can omit them. If a translation does want to change the order of including the the arguments, *they* can introduce numbered parameters in their translation, or am I missing something?

CorsixTH/Lua/strings.lua
```diff
@@ -331,10 +331,11 @@ function Strings:setupAdviserMessage(messages)
     --build_advice
     --cheats
   }
+  -- Passes adviser message through string.format process
   local formatFunc
-  formatFunc = function(format_self, arg)
+  formatFunc = function(format_self, ...)
```
Alberth289346, line 336: Did you try changing the language with this change?

Afaik, the language strings are not plain text, they remember how they were constructed so after a language change they cab rebuild the displayed text by doing the same construction process again.

That construction process may assume there is one parameter for `format`, and this change would then break changing a language.

Alberth289346: Also, I'd merge 335 and 336, and have an empty line above 334  :)

(I am fine with the current  solution too, it's just  a suggestion.)

lewri:
> > Using the more flexible %[1-9]% argument format so translators can rephrase the string if needed.
> 
> I am confused about this. Why are you adding numbered parameters in the original string, you don't seem to need them and they are not obligatory afaik either.
> 
> I would expect that strings that don't need them can omit them. If a translation does want to change the order of including the the arguments, _they_ can introduce numbered parameters in their translation, or am I missing something?

The alternative is using a `%d` and an `%s` respectively, but these iterate over the arguments in order (`%d` requires a number as the argument). String extension adds the `%[1-9]%` to alleviate the problem to be indiscriminate:
https://github.com/CorsixTH/CorsixTH/blob/87858c2d5a4060a6039ca4110fb24403d17d4123/CorsixTH/Lua/string_extensions.lua#L105-L118

The alternative can be using `%s` twice but this still would iterate over the arguments, not allowing you to change argument order provided by `string.format` Of course, we don't have to be as nice as this PR intends to be, we could roll back to the old format.

> Did you try changing the language with this change?
> 
> Afaik, the language strings are not plain text, they remember how they were constructed so after a language change they can rebuild the displayed text by doing the same construction process again.
> 
> That construction process may assume there is one parameter for `format`, and this change would then break changing a language.

Language change is fine and didn't break.

Alberth289346:
> The alternative can be using %s twice but this still would iterate over the arguments, not allowing you to change argument order provided by string.format

I don't think we understand each other. Let's have an example.
- Let's assume we write in english.lua
  ```
  cannot_afford_machine = "You need at least $%d in the bank to afford a new %s!"
  ```
- I don't like that in my language, and want to have swapped parameters. I write:
  ```
  cannot_afford_machine = "Een nieuwe %2% kost %1% Euro hoor, voor minder kan het niet!"
   ````

As long as each translation indicates which argument goes where (either implicitly or explicitly), I would expect it to work.

As such, you're of course free to add explicit parameter numbers in English, but doing that wouldn't (or perhaps shouldn't) make a difference I think.

lewri:
To summarise from matrix/discord -- string to stay in number format and discussion opened about wider implementations.

Alberth289346 (approved):

PR #2427, opened 2023-09-30
RFC: Move sprite functions into virtual methods.

Alberth289346:
The `drawable` struct has
``` cpp
void (*draw_fn)(drawable* pSelf, render_target* pCanvas, int iDestX,  int iDestY);
bool (*hit_test_fn)(drawable* pSelf, int iDestX, int iDestY, int iTestX,  int iTestY);
bool (*is_multiple_frame_animation_fn)(drawable* pSelf);
```
which seemed like virtual methods to me so I hacked the code somewhat to see if it could fly.

For the very limited testing I have done it seems to work. A new game doesn't seem to have problems, a very old save crashes on load but iit does in `master` too, so probably too old.

As for the code change, I did a minimum change. More can be done (at the very least reformatting).

What do you think of such a change?

### DIFF
--- a/CorsixTH/Lua/api_version.lua
+++ b/CorsixTH/Lua/api_version.lua
@@ -32,4 +32,4 @@ Note: This file compiles as both Lua and C++. */
 
 #endif /*]] --*/
 
-return 2684;
+return 2685;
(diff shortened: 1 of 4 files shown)


lewri:
Corruption reported, even from 0.67
image

Game crashes soon after this and closes.

Alberth289346:
I can probably figure out where this goes wrong, but it's small fish. It also needs the big fish of the linked-list elimination and I currently fail to understand where that should go.
As such, it's unlikely to get merged any time soon.

Closing.

PR #1361, opened 2018-03-03
[WIP] Implementing AI Hospitals

mugmuggy:
I mentioned this branch in #967 but thought I would include it here for some more visibility, whilst I fine tune some more aspects. Happy to receive any comments but I do intend to update and tweak until I'm satisfied and make it RFC.

PR #2433, opened 2023-10-03
Implementing AI player's core functionality [WIP]

ageekhere:
The goal is to create AI opponents that players can play with similar to how theme hospital multiplayer works. The player should be able to switch views to see the AI's hospital and view their game play.

In this pull request I am starting to Implement how the AI will be able to interact with the game and a AI reference which will allow players to be able to code new AI opponents. 

ai.lua will be where the core AI functionality will be which will allow an AI to interact with the game. 

To use an example getHappiness() will be an AI function that when called will return the current Happiness of their hospital and return that value to a new AI script. All AI functionality will be in the ai.lua file and explained in the AI reference after which users can create their own AI opponents from. 

Keep in mind that more functions will be added in ai.lua as AI development goes on

I plan to also create an ai once the functionality has been added.

A few other things will also need to be done
1. Create a Single player "skirmish" mode where the player can select a map, a win condition, ai opponents (could be different ai files that users have created) etc.
2. Create maps for the Single player games (atm i think 1 map would be good a good start)
3. Be able to switch to view the AI's hospital to see them build (this is also needed for the ai scripter to see if the ai is doing what they want them to do).
4. Have only one pool of patients that each player has to compete with, things like drug price and reputation will affect how likely the patient will decide to pick your hospital over the others.
5. Sabotage other players

While this is a start a lot of things still need to be discussed ( https://github.com/CorsixTH/CorsixTH/discussions/2431 ) and worked out. I will need some help with the ai.lua as i am not sure on the correct way to get the data from the game.

### DIFF
--- a/CorsixTH/AI/AIreference.txt
+++ b/CorsixTH/AI/AIreference.txt
@@ -1,0 +1,102 @@
+//AI reference
+
+int getRoomCost(int pRoomID) //Get the room cost
+bool getRoomAvailability(int pRoomID) //Check to see if a room is Available
+
+//Creates a new room from pRoomID with the length(pLength) and width(pWidth) at location (pRoomLocation)
+//Setting door location at pDoorLoc
+//Setting windows at positions from pWindowArray
+//Setting items (pItemArray) at the position at pItemLocationArray with their rotation from pItemRotationArray
+int buildRoom(int pRoomID, int pLength, int pWidth, int pRoomLocation, int pDoorLoc, int pWindowArray, int pItemArray, int pItemLocationArray, int pItemRotationArray)
+
+bool deleteRoom(int pRoomID) //Delete a room
+int getRoomQue(int pRoomID) //Get the size of queue  for the room
+bool setRoomQue(int pRoomID) //set the room queue  size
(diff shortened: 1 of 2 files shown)


lewri:
Do we know the overlaps currently on this with #1361 ? I'm not sure if that PR may be a nice starting point to expect the AI behaviour.

tobylane:
There is no code here to review. As the PR lewri links to shows, the AI part of this is small. You could read that PR carefully and share with us what you think is the next step.

Have you used Lua or written computer opponents before?

Alberth289346:
I don't think one PR will do the trick, what you're proposing here is a lot of work, the amount of code that it touches looks way too big to manage as a single change, imho.

A much better tactic here is to repeatedly grab some (very) small part of it, implement that, for as far as makes sense at this time, and make new PRs for reviewing the changes (PRs are very much code-change oriented, you normally discuss technical points about the concrete  changes in the code rather than higher level notions of future changes.

Perhaps keep the discussion for higher level discussions and PRs for making small steps in the code?

lewri:
Closing for now until there is a clearer code path in #2431

PR #2435, opened 2023-10-05
Add vaccinate sound for epidemics

### LINKED ISSUE #2434: [Bug] Missing sound effect when marking patients for vaccination
### Describe the issue

There is a unique sound effect ("VACCIN.WAV") that plays whenever a contagious patient is marked for vaccination during an Epidemic in Theme Hospital, but this sound is not heard when marking contagious patients in CorsixTH.

### Steps to Reproduce

1. Load save game attached to this ticket (or play CorsixTH until an Epidemic starts)
2. Click on a contagious patient to mark them for vaccination (i.e. Contagious Disease icon changes to Waiting For Vaccination)

### Expected Behaviour

- [x] Whenever a contagious patient is marked for vaccination, VACCIN.WAV is played.

Theme Hospital footage which demonstrates this behaviour: https://youtu.be/_PeCGwKKoZs?t=480

### Save Game

[level6 epidemic.zip](https://github.com/CorsixTH/CorsixTH/files/12810236/level6.epidemic.zip)

### CorsixTH Version

0.67

### Operating System

Windows 10

### Theme Hospital Version

CD

### Gamelog.txt

_No response_

### Additional Information

If there are other missing sounds in CorsixTH, it might be worth creating a meta issue to keep track of them.

tobylane:
*Fixes #2434*

- Add the sound of marking a patient for vaccination.

### DIFF
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -304,10 +304,9 @@ function Epidemic:markForVaccination(patient)
     patient.marked_for_vaccination = true
     patient:setMood("epidemy4","deactivate")
     patient:setMood("epidemy2","activate")
+    patient.hospital:playSound("vaccin.wav")
   end
 end
-
-
 
 --[[ Counts the number of patients that have been infected that are still
 -- infected
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -2456,3 +2456,7 @@ end
 --! Called when the vip is out of the hospital grounds
 function Hospital:makeVipEndFax(vip_rating, name, cash_reward, vip_message)
 end
+
+--! Play a sound file
+function Hospital:playSound(sound)
+end
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -761,6 +761,13 @@ function PlayerHospital:makeVipEndFax(vip_rating, name, cash_reward, vip_message
   self.world.ui.bottom_panel:queueMessage("report", message, nil, Date.hoursPerDay() * 20, 1)
 end
 
+--! Play a sound file
+function PlayerHospital:playSound(sound)
+  if self.world.app.config.play_sounds then
+    self.world.app.audio:playSound(sound)
+  end
+end
+
 function PlayerHospital:afterLoad(old, new)
   if old < 145 then
     self.hosp_cheats = Cheats(self)


CorsixTH/Lua/epidemic.lua
```diff
@@ -304,11 +304,10 @@ function Epidemic:markForVaccination(patient)
     patient.marked_for_vaccination = true
     patient:setMood("epidemy4","deactivate")
     patient:setMood("epidemy2","activate")
+    self.world.ui:playSound("vaccin.wav")
```
Alberth289346, line 307: This can't be good.

Only the local player hospital may use the UI, while all hospitals may have epidemics.

tobylane: It's a function triggered by UI, the timer icon.

Alberth289346: It's still looks like mixing internal administrative changes with performing UI notifications to me.

Life gets much easier if you follow one way to do things for all occurrences. With this solution we get some UI going through the hospital interface and other UI actions taking shortcuts through other paths. If we ever want to change UI notifications, or wamnt to have epidemcis handling by non-visible hospitals, you'll have to check everywhere rather than just modify the hospital interface functions.

Unfortunately it not possible to enforce a policy.

lewri: I don't think having the sound play here is out of scope of other code.
Earthquakes in world request sound to be played from inside that code too.

Alberth289346: Earth quakes are a bit different because they affect the world rather than a single hospital.

I agree that a function for playing a vaccine sound is a bit overkill in the amount of code, although I would still do it because it preserves the idea that UI always runs through the hospital.   
Coding is difficult. By making things predictable in doing the same thing in the same way every time we waste less time on handling complicated edge cases that don't fit in the general pattern. Code in general eventually becomes easier to grasp (eg by following the above,  if you want to know or change sounds of a hospital, you *know* that looking in the hospital is enough).

In this case, having a dedicated function tor this sound would indeed work. Other options are writing a general `Hospital:playSound`  function (likely more sounds can be rerouted in that way), or alternatively, ask the hospital for permission. with eg `Hospital:isSoundEnabled()` although that would be a slightly different way of doing things  where one may need to consider if all the currently implemented rerouting should be changed as well (to preserve the "do the same thing in the same way all the time").

lewri:
I'll give in a bit to the sake of uniformity.

Following on Albert's suggestions, I would propose making `Hospital:playSound()` that passes direct through to `ui:playSound()` for hospital safeguarding.
To keep this in scope, I'd suggest making that function and only pointing the vaccine sound playback at it.

Then a new PR should be made to correct all other instances.

tobylane:
`patient:playSound("vaccin.wav")` is another option, if it's neat to do everything to the patient.

lewri (approved):

lewri (approved):

PR #2441, opened 2023-10-16
Patients may only litter indoors

### LINKED ISSUE #1563: Null Pointer Exception in Litter.setLitterType
#### Describe the issue
During hospital evacuation, with people throwing up in the streets, the nil value returned by World:getHospital (world.lua:2251) is not checked correctly before calling Hospital:addHandymanTask in Litter.setLitterType (litter.lua:90)

#### System Information
**CorsixTH Version:** 0.63

#### Gamelog.txt
Error in timer handler: 
Lua\objects\litter.lua:90: attempt to index a nil value (local 'hospital')
stack traceback:
	Lua\objects\litter.lua:90: in method 'setLitterType'
	Lua\humanoid_actions\vomit.lua:33: in local 'timer_function'
	Lua\world.lua:1042: in method 'onTick'
	Lua\app.lua:1111: in function <Lua\app.lua:1108>
	(...tail calls...)
Recovering from error in timer handler...

tobylane:
*Fixes #1563 *

- Check the patient is in the players hospital before they litter.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -883,6 +883,7 @@ function Patient:setTile(x, y)
     if self.litter_countdown == 0 then
       if x and not self:getRoom() and not self.world:getObjects(x, y) and
           self.world.map.th:getCellFlags(x, y).buildable and
+          self.hospital:isInHospital(x, y) and
           (not self.world:findObjectNear(self, "bin", 8) or math.random() < 0.05) then
         -- Drop some litter!
         local trash = math.random(1, 4)


lewri:
Will this still trip if the hospital flags are poorly set in a custom map?

tobylane:
It depends on how broken the broken map is. If the parcel flags go beyond the building, which was the case in at least one bug report, then this will prevent that litter being dropped.

lewri:
> It depends on how broken the broken map is. If the parcel flags go beyond the building, which was the case in at least one bug report, then this will prevent that litter being dropped.

Cool. Won't break it any less then.

Alberth289346:
Sometimes I wonder if we should fix maps with some analysis after loading.

lewri:
> Sometimes I wonder if we should fix maps with some analysis after loading.

We can always fix that one somewhere else. You happy with the change on this one?

Alberth289346:
I don't quite remember how litter was created, but one thing that could happen is that litter may get moved to just after an entrance (I don't know if it does here!!).

To understand, assume one street, and people walking in both directions throw litter every now and then. So at global level, every street tile gets mostly the same amount of litter.

Now If you add a zone halfway in the street where people must keep the litter until leaving the zone, you'll find that more litter is dropped at the tiles near to the zones. The reason for that is because people walking thought zone get time to "produce" litter, but they must delay dropping it. As soon as they leave the zone that can be done, and thus more people than average will do that in in those tiles.

Things depend on how litter gets "produced" here. Also, even if this effect happens, may beit's not so bad. It means having litter bins next to the entrance become important,

As for the change in the current patch, it only restricts behavior so should be fine.

PR #2446, opened 2023-11-05
Movies in ISOs

TheCycoONE:
If you are playing with an ISO hosp directory you can now have in game movies!

### DIFF
--- a/CorsixTH/Lua/filesystem.lua
+++ b/CorsixTH/Lua/filesystem.lua
@@ -94,6 +94,7 @@ function FileSystem:setRoot(physical_path)
   if self:isIso(physical_path) then
     self.provider = ISO_FS()
     self.provider:setPathSeparator(pathsep)
+    self.physical_path = physical_path
     return self.provider:setRoot(physical_path)
   end
 
@@ -217,6 +218,23 @@ function FileSystem:fileSize(virtual_path, ...)
     return nil, e
   end
   return lfs.attributes(s, "size")
+end
+
+--! Get a URI for the file.
+--
+-- The returned URI formats match those described by <https://ffmpeg.org/ffmpeg-protocols.html>.
+--
+--!param virtual_path (string) a path relative to the FileSystem root.
+--!param ... (string) the virtual_path may be split into separate arguments
+-- for each level in the filesystem.
+--!return (string) A uri representing the files location
+function FileSystem:fileUri(virtual_path, ...)
+  virtual_path = getFullPath(virtual_path, ...)
+  if self.provider then
+    local fileStart, fileEnd = self.provider:fileOffsets(virtual_path)
+    return 'subfile,,start,' .. fileStart .. ',end,' .. fileEnd .. ',,:' .. self.physical_path
+  end
+  return 'file:' .. virtual_path
 end
 
 -- If the file exists and we are not using a provider then return a FileSystem
--- a/CorsixTH/Src/iso_fs.cpp
+++ b/CorsixTH/Src/iso_fs.cpp
@@ -423,28 +423,27 @@ bool iso_filesystem::file_metadata_less(const file_metadata& lhs,
 }
 
 int iso_filesystem::find_hosp_directory(const uint8_t* pDirEnt,
-                                        int iDirEntsSize, int iLevel) {
+                                        const uint32_t dirEntsSize, int level) {
   // Sanity check
   // Apart from at the root level, directory record arrays must take up whole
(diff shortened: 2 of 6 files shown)


CorsixTH/Src/iso_fs.cpp
```diff
@@ -423,36 +423,35 @@ bool iso_filesystem::file_metadata_less(const file_metadata& lhs,
...
-  iso_directory_iterator dir_iter(pDirEnt, pDirEnt + iDirEntsSize);
-  iso_directory_iterator end_iter(pDirEnt + iDirEntsSize,
-                                  pDirEnt + iDirEntsSize);
+  iso_directory_iterator dir_iter(pDirEnt, pDirEnt + dirEntsSize);
+  iso_directory_iterator end_iter(pDirEnt + dirEntsSize, pDirEnt + dirEntsSize);
   for (; dir_iter != end_iter; ++dir_iter) {
     const iso_file_entry& ent = *dir_iter;
     if (ent.flags & def_directory) {
       // The names "\x00" and "\x01" are used for the current directory
       // the parent directory respectively. We only want to visit these
```
Alberth289346, line 443:
```suggestion
      // The names "\x00" and "\x01" are used for the current directory and
      // the parent directory respectively. We only want to visit these
```
Probably fits as new line 438 is longer.

CorsixTH/Src/th_lua_iso.cpp
```diff
@@ -80,11 +80,27 @@ int l_isofs_file_size(lua_State* L) {
...
+  iso_filesystem::file_handle iFile = pSelf->find_file(sFilename);
+  if (!iso_filesystem::is_handle_good(iFile)) {
+    lua_pushnil(L);
+    lua_pushnil(L);
+    lua_pushfstring(L, "Could not find \'%s\' in .iso image", sFilename);
+    return 3;
+  }
+  uint32_t start = pSelf->get_file_start(iFile);
+  lua_pushinteger(L, start);
+  lua_pushinteger(L, start + pSelf->get_file_size(iFile));
```
Alberth289346, line 95: It isn't documented, but `start` looks like an inclusive lower bound. (If `.sector` is 0 you get 0 as return value).
Assuming "size" is a proper size, the second value is thus an exclusive upper bound.

In Lua it is normal to use inclusive upper bounds. (and start at 1, but I can see that might be a bad idea).

As it's used for building a string, it's probably fine just a bit strange from a Lua point of view.

lewri:
I'll need to make an ISO copy of the game. Should I wait for you to go through Albert's comments before testing?

CorsixTH/Lua/filesystem.lua
```diff
@@ -219,6 +220,15 @@ function FileSystem:fileSize(virtual_path, ...)
   return lfs.attributes(s, "size")
 end
 
+function FileSystem:fileUri(virtual_path, ...)
+  virtual_path = getFullPath(virtual_path, ...)
+  if self.provider then
+    local fileStart, fileEnd = self.provider:fileOffsets(virtual_path)
+    return('subfile,,start,' .. fileStart .. ',end,' .. fileEnd .. ',,:' .. self.physical_path)
```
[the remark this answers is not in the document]
TheCycoONE: Yes we do, I'll fix

CorsixTH/Src/th_lua_iso.cpp
TheCycoONE: I'd rather cover this one with documentation instead of converting back and forth.

TheCycoONE:
> I'll need to make an ISO copy of the game. Should I wait for you to go through Albert's comments before testing?

I'm actually more concerned with regressions on the non-ISO version. I can cover the ISO if you prefer.

CorsixTH/Src/th_lua_iso.cpp
TheCycoONE: @Alberth289346 for what it's worth, lua doesn't extend this convention to file offsets. https://www.lua.org/manual/5.4/manual.html#pdf-file:seek

TheCycoONE:
Updated with suggestions and some additional documentation. Tested in Windows.

CorsixTH/Src/th_lua_iso.cpp
Alberth289346: Interesting :)

Alberth289346 (approved):
LGTM

lewri (approved):
Non-ISO movies still all working

Issue #1378, opened 2018-04-07
Add warning when hospital entrance cannot be reached


AlexUrbanAc:
**CorsixTH Version:** 0.61
[Empty Hoppalong hospital](https://github.com/CorsixTH/CorsixTH/files/1886490/Hoppalong.zip)
[Last autosave before error occured](https://github.com/CorsixTH/CorsixTH/files/1886491/Autosave2.zip)
[Another try also failed](https://github.com/CorsixTH/CorsixTH/files/1886499/Hoppalong.2.zip)

**Operating System:** Windows 10
**Theme Hospital Version:** GOG.com

#### Gamelog.txt
Created game with savegame version 122.
Savegame version is 122 (v0.61), originally it was 122 (v0.61)
Error in timer handler: 
D:\CorsixTH\Lua\humanoid_actions\seek_reception.lua:59: attempt to perform arithmetic on a boolean value (local 'this_score')
stack traceback:
        D:\CorsixTH\Lua\humanoid_actions\seek_reception.lua:59: in field '?'
        D:\CorsixTH\Lua\entities\humanoid.lua:578: in upvalue 'Humanoid_startAction'
        D:\CorsixTH\Lua\entities\humanoid.lua:670: in local 'timer_function'
        D:\CorsixTH\Lua\entity.lua:216: in method 'tick'
        D:\CorsixTH\Lua\world.lua:1051: in method 'onTick'
        D:\CorsixTH\Lua\app.lua:1008: in function <D:\CorsixTH\Lua\app.lua:1005>
        (...tail calls...)
        D:\CorsixTH\Lua\app.lua:909: in function <D:\CorsixTH\Lua\app.lua:904>

EDIT: The issue can be circumvented by buying lots of the hospital early. It seems both main entrances on the Hoppalong map are not reachable by pathfinding, and from the error message I guess getPathDistance returns false if no path is found.

MarkL1961:
Yes it is a flawed map, there has to be a route to the edge of the map from the start.

TheCycoONE:
I'm open to closing this, or making it a feature request on a path sanity check when starting a level.

MarkL1961:
Perhaps the sanity check should be in the map making, maybe on save.
Yes you would get a warning if you saved the map during making it, but if on save there was a warning that it would not be playable and the same with hele-pads it would save this from happening.

[...]

PR #2447, opened 2023-11-06
Add a warning to mapeditor when a map has no spawn path

tobylane:
Partially addresses #1378

- Checks the map has a valid path from the hospital to the first spawn point.
Should the spawn points be niled after this?

### DIFF
--- a/CorsixTH/Lua/dialogs/menu.lua
+++ b/CorsixTH/Lua/dialogs/menu.lua
@@ -546,8 +546,15 @@ function UIMenuBar:makeMapeditorMenu(app)
   local menu = UIMenu()
   local hotkeys = app.hotkeys
 
+  local function mapeditorSave(ui)
+    local message = app.world:validateMap()
+    if message then
+      ui:addWindow(UIInformation(ui, {message}))
+    end
+    ui:addWindow(UISaveMap(ui))
+  end
   menu:appendItem(_S.menu_file.load:format(hotkey_value_label("ingame_loadMenu", hotkeys)), function() self.ui:addWindow(UILoadMap(self.ui, "map")) end)
-    :appendItem(_S.menu_file.save:format(hotkey_value_label("ingame_saveMenu", hotkeys)), function() self.ui:addWindow(UISaveMap(self.ui)) end)
+    :appendItem(_S.menu_file.save:format(hotkey_value_label("ingame_saveMenu", hotkeys)), function() mapeditorSave(self.ui) end)
     :appendItem(_S.menu_file.quit:format(hotkey_value_label("ingame_quitLevel", hotkeys)), function() self.ui:quit() end)
   self:addMenu(_S.menu.file, menu)
 
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -911,7 +911,10 @@ map_editor_window = {
     heliport_3 = "Heliport 3",
     heliport_4 = "Heliport 4",
     paste = "Paste area",
-  }
+  },
+  checks = {
+    spawn_points_and_path = "Warning: Patients cannot reach the hospital. They need 'road' tiles or 'outside' grey tiles at the edge of the map and a path of these tiles to the hospital entrance.",
+  },
 }
 
 hotkeys_file_err = {
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -172,7 +172,7 @@ function World:World(app)
     self.object_id_by_thob[object_type.thob] = object_type.id
   end
   self:makeAvailableStaff(0)
-  self:calculateSpawnTiles()
+  self.spawn_points = self:calculateSpawnTiles()
 
   -- Next Events dates
   -- emergencies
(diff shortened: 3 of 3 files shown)


CorsixTH/Lua/dialogs/menu.lua
```diff
@@ -546,8 +546,14 @@ function UIMenuBar:makeMapeditorMenu(app)
   local menu = UIMenu()
   local hotkeys = app.hotkeys
 
+  local function check_before_save(ui)
```
Alberth289346, line 549: Function name code style is incorrect, lowercase with underscores is for variables.

Also, the name is too detailed. Don't try to add every little step that a function does in the name. Today there is only one additional step but in time more steps could be added. You'll soon run out of line length if you want to express them all, not to mention you need to constantly change the function name throughout the entire program (in theory, this case is simple).  
You have a several lines room in the Luadoc of the function for a more detailed description, or in the code itself in comments.

Also, the name looks too generic. There are several things that can be saved in the program, `save` can be applied on all those things yet the function doesn't handle all those things. It saves maps only.

CorsixTH/Lua/map.lua
```diff
@@ -735,6 +735,15 @@ function Map:getParcelTileCount(parcel)
   return self.parcelTileCounts[parcel] or 0
 end
 
+function Map:checkMapPathFinder()
```
Alberth289346, line 738: Why is this function in `Map`?  It doesn't even use any data from the map (besides `app`).

Looks more like a world function to me.

That also solves the problem that world may not exist when the function is called.

Alberth289346:
Luadoc is missing for the new functions.

Alberth289346:
Shouldn't you check this for all hospitals, or at least for every player defined in the map?

CorsixTH/Lua/world.lua
```diff
@@ -2880,3 +2880,14 @@ function World:setCampaignData(campaign_data)
   end
   self:getLocalPlayerHospital():setCampaignData(campaign_data.hospital)
 end
+
+--! Perform some checks on a map that a user is saving in the map editor
```
Alberth289346, line 2884: It's a nice generic function that performs some checks and produce a warning if a problem is found.

Providing context outside this function in the Luadoc  (like your "a user is saving in the map editor") is tricky here, as in the future you or someone else may find another good use for these checks (eg a quick sanity check after loading a map, so the user gets a warning that it is not a good idea to play this map).  
If that happens, the Luadoc starts spreading false information (or at least partially incorrect information).

The safer option here is not to explain context of use for the function unless it's absolutely needed.
For example, if you have a number of functions that are designed to work together, and they only make sense in that way, (ie they are useless in any other situation). then it's fine to mention the usage context, because that's the one and only  way that the function(s) makes sense.

The latter  example seldomly happens, as one of the goals is to have many functions that are re-usable for various purposes..

CorsixTH/Lua/languages/english.lua
```diff
@@ -911,7 +911,11 @@ map_editor_window = {
     heliport_3 = "Heliport 3",
     heliport_4 = "Heliport 4",
     paste = "Paste area",
-  }
+  },
+  checks = {
+    spawnpoints = "Warning: Spawn points could not be generated for this map.",
```
Alberth289346, line 916: Is a "spawn point" something that a user is supposed to know?

maybe "Warning: New patients cannot enter the map." ? (better connection to patients, but not entirely happy with the sentence).

CorsixTH/Lua/languages/english.lua
```diff
@@ -911,7 +911,11 @@ map_editor_window = {
...
+    pathfind = "Warning: A path could not be made between the spawn points and the hospital.",
```
Alberth289346, line 917: "A path" or or "No path" ? Makes a big difference :)

I don't quite buy "made" either , it would work if you changed all outside tiles to road, right?

Perhaps "Warning: No path exists from  the edge of the map to the hospital" ?

or "no path available" could work too.

Alberth289346:
Somewhat out of scope, but the `calculateSpawnPoints` name is not telling the truth. It doesn't just calculate, is also adds them to world.

A better name would be `setSpawnPoints`, which expresses that something will be assigned.

A different (and imho preferable solution) is to keep the name as-is, and make it not change world. Instead, it should return the computer spawn points to the caller. The only currently existing call can then assign the return value to a variable in the world, and you can use the computed spawn points for your warnings without affecting world data as a side-effect of checking.  
This latter solution also make your `validateMap` function truthful to its name. When you read "validate" one doesn't expect that some data somewhere gets silently modified or updated.


I do like your change, it's a nice feature.

CorsixTH/Lua/languages/english.lua
tobylane: > Warning: There are no tiles from the road page on the edges of the map. Patients need these to enter the map.

That directly gives them the remedy and reason.

tobylane:
It can only be one test, not two. The message to cover both could be "Warning: Patients cannot spawn and walk to the hospital on this map. They need tiles from the road page, or the three grey tiles from the outside page. These tiles need to be present somewhere on the edge of the map, and on a path to the hospital." Should the three grey tiles be highlighted?

Alberth289346:
Combining them seems a good idea to me.

>  "Warning: Patients cannot spawn and walk to the hospital on this map. They need tiles from the road page, or the three grey tiles from the outside page. These tiles need to be present somewhere on the edge of the map, and on a path to the hospital."

Don't know what you think of it, but it seems a bit long to me. Below I tried a shorter version. Please use the bits you like :)

Warning: Patients cannot reach the hospital. They need 'road' tiles or 'outside' grey tiles at the edge of the map and a path of these tiles to the hospital entrance.

Besides the phrasing of the message, patch looks good to me.

tobylane:
Thanks, I've used your line.

TheCycoONE:
> Shouldn't you check this for all hospitals, or at least for every player defined in the map?

Yes, but I think that's out of scope for this, and in scope for multiplayer.

PR #2448, opened 2023-11-08
Show languages as disabled when they aren't available

tobylane:
*Addresses #1915*

- If the user can't switch to the unicode languages, show them in the language select list with their English name in a disabled button. The tooltip tells them to set a font.

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/options.lua
+++ b/CorsixTH/Lua/dialogs/resizables/options.lua
@@ -237,12 +237,17 @@ function UIOptions:checkForAvailableLanguages()
   local langs, c = {}, 1
   for _, lang in pairs(app.strings.languages) do
     local font = app.strings:getFont(lang)
-    if app.gfx:hasLanguageFont(font) and app.strings.languages_english[lang] then
-      local eng_name = app.strings.languages_english[lang]
-      c = c + 1
+    local eng_name = app.strings.languages_english[lang]
+    c = c + 1
+    -- If freetype support and a unicode font setting are not present then
+    -- languages not supported by the builtin font are named in English and cannot be selected
+    if app.gfx:hasLanguageFont(font) then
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -237,12 +237,17 @@ function UIOptions:checkForAvailableLanguages()
...
-      local eng_name = app.strings.languages_english[lang]
-      c = c + 1
+    local eng_name = app.strings.languages_english[lang]
+    c = c + 1
+    -- If freetype support and a unicode font setting are not present then
+    -- languages not supported by the builtin font are named in English and cannot be selected
+    if app.gfx:hasLanguageFont(font) then
       font = font and app.gfx:loadLanguageFont(font, app.gfx:loadSpriteTable("QData", "Font01V"))
-      langs[#langs + 1] = { text = lang, font = font,
+      langs[#langs + 1] = { text = lang,     name = lang, font = font, disabled = false,
```
lewri, line 246: Why the odd spacing?

lewri (commented):
This one is just disabling the items, not trying to fix the UI issues correct?

CorsixTH/Lua/dialogs/resizables/options.lua
tobylane: It's a little easier to compare with the similar line below when they're aligned.

tobylane:
Yes, the UI issue of tooltips only appearing when on top of the parent window is still there.

lewri (approved):
Cool.

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE, line 246: Too much white space before name.

PR #2449, opened 2023-11-08
Refactor World:findRoomNear

lewri:
- Removes the generally undesired repeat until loop
- Adds doc

I'm not convinced the "nearest" declaration is actually needed really except for clarity to where function calls are.

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1863,7 +1863,7 @@ function World:findFreeObjectNearToUse(humanoid, object_type_name, which, curren
   return object, ox, oy
 end
 
---! Finds the most desirable room of a given room type
+--! Finds the most desirable room of a given room type for the provided humanoid.
 --!param humanoid The queried NPC (start point)
 --!param room_type_id (string) The room we are looking for
 --!param max_distance (number) The maximum distance away a room will be considered from the humanoid
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1863,34 +1863,43 @@ function World:findFreeObjectNearToUse(humanoid, object_type_name, which, curren
   return object, ox, oy
 end
 
-function World:findRoomNear(humanoid, room_type_id, distance, mode)
-  -- If mode == "nearest" (or nil), the nearest room is taken
-  -- If mode == "advanced", prefer a near room, but also few patients and fulfilled staff criteria
+--! Finds the most desirable room of a given room type
+--!param humanoid The queried NPC (start point)
+--!param room_type_id (string) The room we are looking for
+--!param max_distance (number) The maximum distance away a room will be considered from the humanoid
+--!param mode (string) "nearest"/nil returns the nearest room, while
+--! "advanced" prefers a near room but also takes into account queue size/staffing
+--!return The best room for the huamnoid or nil if no room exists within the maximum distance
+function World:findRoomNear(humanoid, room_type_id, max_distance, mode)
+  assert(type(room_type_id) == "string", "Room type id given is not a string!")
   local room
   local score
   if not mode then
     mode = "nearest" -- default mode
   end
-  if not distance then
-    distance = 2^30
+  if not max_distance then
+    max_distance = 2^30
   end
-  for _, r in pairs(self.rooms) do repeat
-    if r.built and (not room_type_id or r.room_info.id == room_type_id) and r.is_active then
+
+  for _, r in pairs(self.rooms) do
+    if r.built and r.room_info.id == room_type_id and r.is_active then
       local x, y = r:getEntranceXY(false)
       local d = self:getPathDistance(humanoid.tile_x, humanoid.tile_y, x, y)
-      if not d or d > distance then
-        break -- continue
-      end
-      local this_score = d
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/world.lua
```diff
@@ -1863,9 +1863,14 @@ function World:findFreeObjectNearToUse(humanoid, object_type_name, which, curren
   return object, ox, oy
 end
 
+--! Finds the most desirable room of a given room type
```
Alberth289346, line 1866: As the "return" doc says, you're looking for a room for a humanoid rather than just one of a given type.

I'd rather just have a position to be honest instead of a humanoid, but that breaks the API.

CorsixTH/Lua/world.lua
```diff
@@ -1863,9 +1863,14 @@ function World:findFreeObjectNearToUse(humanoid, object_type_name, which, curren
...
+--!param humanoid The queried NPC (start point)
+--!param room_type_id (string) The room we are looking for
```
Alberth289346, line 1868: Code says it can also be `nil` is that something we want to keep?

Either way it's useful if the docs and the code agree on allowed input :)

CorsixTH/Lua/world.lua
```diff
@@ -1863,9 +1863,14 @@ function World:findFreeObjectNearToUse(humanoid, object_type_name, which, curren
...
+--!param distance (number) The maximum distance away a room will be considered from the humanoid
```
Alberth289346, line 1869: Perhaps rename to `max_distance`?

CorsixTH/Lua/world.lua
```diff
@@ -1863,9 +1863,14 @@ function World:findFreeObjectNearToUse(humanoid, object_type_name, which, curren
...
+--!param mode (string) "nearest"/nil returns the nearest room
```
Alberth289346, line 1870:
```suggestion
--!param mode (string) "nearest"/nil returns the nearest room, while
```

CorsixTH/Lua/world.lua
```diff
@@ -1863,9 +1863,14 @@ function World:findFreeObjectNearToUse(humanoid, object_type_name, which, curren
...
+--! "advanced" prefers a near room but also takes into account queue size/staffing
+--!return The best room for the huamnoid
```
Alberth289346, line 1872: or `nil` if there is no room of the requested type within the maximum distance.

CorsixTH/Lua/world.lua
```diff
@@ -1874,23 +1879,27 @@ function World:findRoomNear(humanoid, room_type_id, distance, mode)
   if not distance then
     distance = 2^30
   end
-  for _, r in pairs(self.rooms) do repeat
-    if r.built and (not room_type_id or r.room_info.id == room_type_id) and r.is_active then
+
+  for _, r in pairs(self.rooms) do
+    if r.built and (not room_type_id or r.room_info.id == room_type_id) and
```
Alberth289346, line 1884: I don't know how often this room type check is done in the code, but would creating a `Room::isOfType(room_type)` wrapper function be useful?

Alberth289346:
lol, `repeat until true`

CorsixTH/Lua/world.lua
lewri: It does? The uses in the code don't seem to point to it ever using nil.
Do you mean the inference by `(not room_type_id or r.room_info.id == room_type_id)` ?

CorsixTH/Lua/world.lua
lewri: I assume you mean moving that bracketed query into the `room` code? I mean maybe... though as per my other comment, is `not room_type_id` some weird legacy safeguard?

lewri: `if room,room_info.id == x` is used ~40 times. Not sure I want to consolidate it right now.

CorsixTH/Lua/world.lua
Alberth289346: Yes, that's what I mean.

If it unwanted, I'd suggest to assert that the parameter is  a string.

CorsixTH/Lua/world.lua
Alberth289346: Fair enough.

The idea of such wrapper functions is that `r:isRoomOfType(room_type_id)` is more clear wrt its intention (even more if the `_id`suffix gets dropped). It also raises the abstraction level a bit. The spot where a room is storing its type is then more an internal `Room` thing instead of the 40 places now where the knowledge is sprinkled through the code.

CorsixTH/Lua/world.lua
lewri: I'll do some sanity runs and verify before updating the pr

Alberth289346: Otherwise you'd end up with any nearest room, which feels much less useful :)

CorsixTH/Lua/world.lua
Alberth289346:
```suggestion
--! Finds the most desirable room of a given room type for the provided humanoid.
```

Alberth289346 (approved):

PR #2450, opened 2023-11-11
Add explore button to file and folder browsers to open folder in a file manager

### LINKED ISSUE #369: Load/Save game window needs a delete button (confirm button done)
I will start working on these new buttons today.

tobylane:
*May fix #369*

-  Add an explore button to all relevant windows. Rather than add rename and delete buttons, this helps the user make those changes in their file manager. The window doesn't update with those changes.
- Add App.osOpen to open these paths, and urls from updates. This could be in utility.lua or App:init with math.random variants.
- The windows open command start [needs extra quotes.](https://superuser.com/questions/1656455/how-to-use-a-path-with-spaces-in-the-batch-file)

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1928,6 +1928,25 @@ function App:gamelogHeader()
   return (gen_date .. compiled .. running)
 end
 
+--! Open the given string (URL or folder path) in the operating system. Requires io.popen.
+--!param path (string)
+function App:osOpen(path)
+  if not path or not io.popen then return end
+  local cmd
+  if self.os == "windows" then
+    cmd = 'start "" "'
+  elseif self.os == "macos" then
+    cmd = 'open "'
+  else
+    cmd = 'xdg-open "'
+  end
+  local handle = io.popen(cmd .. path .. '" 2>&1')
+  local output = handle:read('*a'):gsub('[\n\r]', ' ')
+  if output:len() > 0 then
+    self.ui:addWindow(UIInformation(self.ui, {output}))
+  end
+end
+
 -- Do not remove, for savegame compatibility < r1891
 local app_confirm_quit_stub = --[[persistable:app_confirm_quit]] function()
 end
--- a/CorsixTH/Lua/dialogs/resizables/options.lua
+++ b/CorsixTH/Lua/dialogs/resizables/options.lua
@@ -383,18 +383,6 @@ function UIOptions:buttonJukebox()
   end
 end
 
-function UIOptions:buttonBrowseForTHInstall()
-  local function callback(path)
-    local app = TheApp
-    app.config.theme_hospital_install = path
-    app:saveConfig()
-    debug.getregistry()._RESTART = true
-    app.running = false
-  end
-  local browser = UIDirectoryBrowser(self.ui, self.mode, _S.options_window.new_th_directory, "InstallDirTreeNode", callback)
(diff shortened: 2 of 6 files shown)


TheCycoONE:
I don't think I like this. What's the motivation?

tobylane:
It lets the user manage the files without needing us to implement all the necessary specific buttons - rename, delete, maybe new folder, move into folder. It makes it easier to share save games in issues.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -349,18 +349,6 @@ function UIOptions:buttonHotkey()
...
-    local app = TheApp
-    app.config.theme_hospital_install = path
-    app:saveConfig()
-    debug.getregistry()._RESTART = true
-    app.running = false
-  end
-  local browser = UIDirectoryBrowser(self.ui, self.mode, _S.options_window.new_th_directory, "InstallDirTreeNode", callback)
-  self.ui:addWindow(browser)
-end
-
```
lewri, line 363: This is just a deletion for an obsolete function now yes? (I.e. it's now in `folder_settings.lua`)

CorsixTH/Lua/app.lua
```diff
@@ -1912,6 +1912,18 @@ function App:gamelogHeader()
...
 
+--! Open the given string (URL or folder path) in the operating system
+--!param path (string)
+function App:osOpen(path)
+  if self.os == "windows" then
+    os.execute('start "" "' .. path .. '"')
+  elseif self.os == "macos" then
+    os.execute('open "' .. path .. '"')
+  else
+    os.execute('xdg-open "' .. path .. '"')
```
lewri, line 1923: Would this in future benefit from an error catch?

lewri (commented):
External changes to the files/folders prompted by the game either need to be reflected in the game once completed; a message shown to the user to advise them to reload the file/folder browser after they finish doing whatever externally; or close the in-game browser if they select explore.
The third option forces a reload of the directory structure since they have to go back in, but could be problematic it if they use this during save game. 

Generally speaking though, I'm not sure this implementation is correct. I'm not against the idea of making easier access to these directories, but not in the method laid out. For Windows we dump a shortcut to the CorsixTH AppData folder in Start, though that only covers defaults. An option maybe in the folders menu you could have an "Open" and "Change/Modify" button against the directory path shown.

tobylane:
Screenshot 2024-01-12 at 18 26 06

Moved to folder settings. The error message for a missing folder comes localised from the OS, but this uses io.popen which is apparently not always available. I don't know if that's relevant here.

Alberth289346:
Why not add a "delete" (which seems the most basic one) , and a way to get the path as a past-able string, so you can go there with whatever you like to change the structure?

Games often don't provide complete manipulation of their saves

tobylane:
Lua needs a library to put things in the clipboard.

I've brought back the original commit, with a delete button in place of explore. It doesn't refresh the file browser, I think it needs to.
Screenshot 2024-02-11 at 10 58 35

lewri:
> Lua needs a library to put things in the clipboard.
> 
> I've brought back the original commit, with a delete button in place of explore. It doesn't refresh the file browser, I think it needs to. 

A function to reload the tree maybe? Or close and reopen that window?

tobylane:
I made a function to remove nodes from the tree of files, but it was hacky and flawed. Reducing this pr to the commit of the open button in folder settings.

Issue #2451, opened 2023-11-12
[Bug] Better Linux CI actions


nMustaki:
### Describe the issue

The Linux CI checks need some love.
For instance the `grep` is finding all lines with a 't' instead of all lines with a tab

It also seems the CI step is not working as intended since it does not report errors

### Steps to Reproduce

1. Open the PR

### Expected Behaviour

The CI step should detect tabs

### Save Game

Save game attached to report: Yes/No

### CorsixTH Version

0.66

### Operating System

Linux

### Theme Hospital Version

CD

PR #2452, opened 2023-11-12
#2451 Tab detection broken in CI

nMustaki:
In CI, the `grep` command is finding all lines with a 't' instead of all lines with a tab


tobylane:
'! grep -IrnP '\t' CorsixTH/Lua' is another solution. I have a bunch of fixes to that file to submit.

nMustaki:
@tobylane thanks ! If you want to forward them to me I'll take care of them

nMustaki:
Closing, the PR https://github.com/CorsixTH/CorsixTH/pull/2453 is way more complete :-)

PR #2457, opened 2023-11-17
Show level briefing before the tutorial instead of after

### LINKED ISSUE #2001: Level Briefing should show before tutorial
It doesn't make much sense to show the level 1 brief after the tutorial in the campaign. Instead, it should be at the start before the tutorial begins.

Makes the level ordering better and more consistent.

Can't remember

 0.65

No

**Additional information**

tobylane:
*Fixes #2001*

- Always show the level briefing on day 1
- Don't show the final message of the tutorial when playing level 1, as it is the level briefing already shown

### DIFF
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -1094,13 +1094,14 @@ tutorial_phases = {
         -- the real game uses three.
         local texts = TheApp.using_demo_files and {
           {_S.introduction_texts["level15"]},
-          {_S.introduction_texts["demo"]},
         } or {
           {_S.introduction_texts["level15"]},
           {_S.introduction_texts["level16"]},
           {_S.introduction_texts["level17"]},
-          {_S.introduction_texts["level1"]},
         }
+        if TheApp.world.map.level_number ~= 1 then
+          table.insert(texts, _S.introduction_texts["level1"])
+        end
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/game_ui.lua
```diff
@@ -1099,8 +1099,10 @@ tutorial_phases = {
           {_S.introduction_texts["level15"]},
           {_S.introduction_texts["level16"]},
           {_S.introduction_texts["level17"]},
-          {_S.introduction_texts["level1"]},
         }
+        if TheApp.world.map.level_number ~= 1 then
+          table.insert(texts, _S.introduction_texts["level1"])
+        end
```
lewri, line 1105: Should we also cover the demo's tutorial? Or does the demo one make more logical sense to come after?

tobylane: Yes, both of them before is better.

lewri:
I haven't been able to test this currently -- but can you confirm if moving the level 1/demo briefing to before the tutorial is triggering the initial opening countdown timer? Do we want the tutorial to be timed?

tobylane:
> can you confirm if moving the level 1/demo briefing to before the tutorial is triggering the initial opening countdown timer? 

It does, that needs fixing.

lewri (approved):
Tested. I'm happy.

PR #2461, opened 2023-11-22
[RFC] Add simple interface for serialize

lewri:
I use this function for my own debugging purposes but may be helpful for others. It provides a simplified interface to use `serialize` rather than needing to juggle options every time.

I can close this PR if we don't want this functionality in utility.

### DIFF
--- a/CorsixTH/Lua/utility.lua
+++ b/CorsixTH/Lua/utility.lua
@@ -404,6 +404,15 @@ function serialize(val, options, depth, pt_reflist)
   end
 end
 
+--! Simplified interface for serializing a value with option for depth
+-- See serialize for further explanation
+-- Output is dumped to the console
+--!param value A table or string to inspect
+--!param depth (num) Optional, defaults at 1
+function inspect(value, depth)
+  print(serialize(value, {detect_cycles = true, max_depth = depth or 1, pretty = true}))
+end
+
 -- Clones a table, but only the first level.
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/utility.lua
```diff
@@ -404,6 +404,15 @@ function serialize(val, options, depth, pt_reflist)
   end
 end
 
+--! Simplified interface for serializing a value with option for depth
+-- See serialize for further explanation
+--!param value A table or string to inspect
+--!param depth (num) Optional, defaults at 1
+--!return the serialized value as a console output
+function inspect(value, depth)
+  return print(serialize(value, {detect_cycles = true, max_depth = depth or 1, pretty = true}))
```
Alberth289346, line 413: `print` doesn't return anything, does it?

Alberth289346:
I'd use `dump`, but `inspect` is fine too imho.

CorsixTH/Lua/utility.lua
lewri: I suppose you're right. Can you suggest a better way to rewrite the return param doc?

lewri: Actually, return isn't needed and not sure why I put it in.

Alberth289346: Just don't write the `!return` line :)

PR #2468, opened 2023-12-29
Resize UIInformation

lewri:
- Makes UIInformation more uniform to the original game as our implementation has excessive width padding. Works without any impact on non-standard fonts/cjk. Note Chinese still has an oddity where it adds additional blank space to the right.
- Makes close button have a reserved zone on bottom to allow for better utilisation of the box width
- Reformats some of the code spacing (there's too many magic numbers going on currently for me to focus on improving that here)
- Before 
image

- After 
image

A future improvement would be to also utilise the empty space adjacent to the button with awareness for that button existing on wrap. Or, move the button outside the box to look like a little tab on the top right.

### DIFF
--- a/CorsixTH/Lua/dialogs/information.lua
+++ b/CorsixTH/Lua/dialogs/information.lua
@@ -57,10 +57,10 @@ function UIInformation:UIInformation(ui, text, use_built_in_font)
   -- Window size parameters
   self.text_width = 300
   self.spacing = {
-    l = 40,
-    r = 40,
-    t = 20,
-    b = 20,
+    l = 15,
+    r = 15,
+    t = 15,
+    b = 18 + 15, -- Size of close button + padding
   }
 
   self:onChangeLanguage()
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/dialogs/information.lua
```diff
@@ -57,10 +57,10 @@ function UIInformation:UIInformation(ui, text, use_built_in_font)
...
   self.text_width = 300
   self.spacing = {
-    l = 40,
-    r = 40,
-    t = 20,
-    b = 20,
+    l = 15,
+    r = 15,
+    t = 15,
+    b = 33, -- includes space for close button (18px + 15px padding)
```
Alberth289346, line 63:
```suggestion
    b = 18 + 15, -- Size of  close button + padding
```
Not sure if the comment is fully correct, but in this way you let the computer add the numbers, so comment and reality always match.

Alberth289346:
> A future improvement would be to also utilise the empty space adjacent to the button with awareness for that button existing on wrap. Or, move the button outside the box to look like a little tab on the top right.

I quite like what you did here. First you read the text, then you find the button. It makes sense to me to keep the button below the text. Wrapping like you propose will likely also work, as long as you keep some empty space around the button I am guessing (so it clearly stays a separate item in the window).

CorsixTH/Lua/dialogs/information.lua
lewri: Happy to go with that for now

lewri:
@Alberth289346 can you see anything more to address?

PR #2472, opened 2024-01-15
Set minimum window size

tobylane:
*Fixes #*

- Sets the minimum window size in SDL, to prevent resizing the window smaller than the lua limit.


lewri (commented):
I'm not against the idea, I'd prefer though if we do this we have a single source of truth for the minimum rather than duplicating the number in Lua too though (or vice versa).

TheCycoONE:
Lewri isn't wrong, would be nice to make these c++ constants and then bind them into lua. Way more effort than this PR though.

PR #2474, opened 2024-01-20
Patch Lua Bitmap Makefiles

lewri:
Work should bring these scripts up to Lua 5.4 compatibility while retaining Lua 5.1/JIT compatibility (not tested)

-- Removes `setfenv` call and rewrites spec file to work without it
-- Removes `module`, replaced with `require`
-- Fixes bug in `makeRaw` where bitmap files were not closed after execution

I've not attempted to completely untangle the methods Corsix used for these files so likely a lot of odd design choices still exist (and ones that fail our coding conventions) but I can confirm these mkfiles are running from CorsixTH.exe via debug script.

[**mkboostrap.lua**](https://github.com/CorsixTH/CorsixTH/blob/835bb9aad6d751749ff88e9449797547bf6424da/CorsixTH/Bitmap/mkbootstrap.lua) in Bitmap folder may need addressing with these changes, but I don't want to look at that one myself (if someone wanted to repatch it if needed they can add a commit.

I'd recommend turning off whitespace changes.

### DIFF
--- a/CorsixTH/Bitmap/example.spec
+++ b/CorsixTH/Bitmap/example.spec
@@ -1,0 +1,21 @@
+--[[ EXAMPLE SPEC FILE FOR MAKING SPRITESHEETS ]]
+-- Specify your sprites (BMP images) and output file locations
+-- It is recommended to leave palette and compression options at their default values
+local SPEC = {
+  sprites = { -- one bmp per line
+    "example1.bmp",
+    "example2.bmp",
+  },
+  -- Palette options
+  palette = "from bitmap",
+
+  -- Compression options
+  complex = true,
+  rnc = false,
+
+  -- Output filenames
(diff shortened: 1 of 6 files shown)


TheCycoONE:
LGTM, what kind of feedback were you looking for?

lewri:
> LGTM, what kind of feedback were you looking for?

I need to know how these changes affect mkboostrap -- as they amend c++ files it may be better if someone can test where changes are needed to be accomodated.

Also I'm not sure if how I've made the requires work is best practice -- it works, but is it right enough?

TheCycoONE:
No one has run mkbootstrap in a decade or more...

lewri:
> No one has run mkbootstrap in a decade or more...

No worry, I'll merge as is.

PR #2480, opened 2024-02-02
Tidy root CorsixTH.lua

tobylane:
- Remove the ipairs function as lua ipairs was never deprecated (corsix's blog says it was in a beta)
- Rewrite the lua version check, including mentioning 5.4
- Combine the two parts of start debugging. Was this separate for a reason?

### DIFF
--- a/CorsixTH/CorsixTH.lua
+++ b/CorsixTH/CorsixTH.lua
@@ -15,14 +15,6 @@ if debug.setcstacklimit then
   debug.setcstacklimit(30000)
 end
 
--- Parse script parameters:
-local run_debugger = false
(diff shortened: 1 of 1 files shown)


lewri:
Are you happy enough the change isn't going to have an impact with 5.1-5.3 and JIT? The only reason I could think it was written like it was was if JIT had some differences regarding `ipairs` and `table.unpack`. But I haven't got a resource to hand to check building with these versions.

tobylane:
Luajit and luajit-openresty are compatible with 5.1, if we need to know more they can be checked. 
> \> print(jit.version)
LuaJIT 2.1.1700206165

PR #2481, opened 2024-02-02
Add semantic version comparison for updates

### LINKED ISSUE #2011: [Bug] update check never sees a new version

tobylane:
*Fixes #2011*

- Add a semver function to compare three segment version numbers in the check for updates.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1808,7 +1808,7 @@ function App:checkForUpdates()
   -- Default language to use for the changelog if no localised version is available
   local default_language = "en"
   local update_url = 'https://corsixth.com/CorsixTH/check-for-updates'
-  local current_version = string.gsub(self:getVersion(), "v", "") -- drop the 'v'
+  local current_version = self:getVersion()
 
   -- Only URLs that match this list of trusted domains will be accepted.
   local trusted_domains = { 'corsixth.com', 'github.com', 'corsixth.github.io' }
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/app.lua
```diff
@@ -1839,16 +1839,32 @@ function App:checkForUpdates()
...
+    end
+    if current_minor > update_table.minor then return true
+    elseif current_minor < update_table.minor then return false
+    end
+    current_revision = tonumber(current_revision) or 0
+    if current_revision >= update_table.revision then return true
+    else return false
+    end
+  end
+  if semver() then
```
lewri, line 1861: Is the function name correct? Semver what you're using, but the comment needs to tell me you're doing a compare. Whereas, it's likely going to be better to communicate it in the function name.

You could also split this out into two functions. One to convert to semver and another to compare versions. It may also be possible then to make this a little less messy.

CorsixTH/Lua/app.lua
```diff
@@ -1808,7 +1808,7 @@ function App:checkForUpdates()
   -- Default language to use for the changelog if no localised version is available
   local default_language = "en"
   local update_url = 'https://corsixth.com/CorsixTH/check-for-updates'
-  local current_version = string.gsub(self:getVersion(), "v", "") -- drop the 'v'
+  local current_version = self:getVersion()
```
lewri, line 1811: Good. I didn't much like the workaround I did.

CorsixTH/Lua/app.lua
```diff
@@ -1839,16 +1839,32 @@ function App:checkForUpdates()
   end
 
   local update_table = loadstring_envcall(update_body, "@updatechecker") {}
+  update_table.revision = update_table.revision or 0
```
lewri, line 1842: Nice we have built in some wider considerations for how we treat revision. I assume if we go ahead on this path we're going to end up at `0.70.0` or `0.68.0`.

TheCycoONE:
I agree that the function name should be changed to compare_versions or something.

The algorithm would have been a bit more elegant if we had numbered the parts instead of naming them, you could use a gmatch and iterate through until you found a difference - but we've used the named parts in the updater for a long time so this will do. 👍 with the rename.

PR #2482, opened 2024-02-03
Make the tips window adapt to the space beside the main menu

### LINKED ISSUE #2200: Reposition tips window (where resolution allows?)
>Speaking of which, maybe it's just me, but I find the intro screen tips window to be easily missed. The intro screen looks good, but perhaps the tips window could be placed closer to the interface window..? 

>image

_Originally posted by @andersand in https://github.com/CorsixTH/CorsixTH/issues/2036#issuecomment-895206480_

tobylane:
*Fixes #2200*

- The width of the tips window is narrower in small resolutions, with stacked previous/next buttons. As it doesn't overlap with the main menu, is this an acceptable place to close and reopen windows without caring about window order?

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/tip_of_the_day.lua
+++ b/CorsixTH/Lua/dialogs/resizables/tip_of_the_day.lua
@@ -31,9 +31,13 @@ local col_bg = {
 }
 
 function UITipOfTheDay:UITipOfTheDay(ui)
-  self:UIResizable(ui, 380, 110, col_bg)
-
   local app = ui.app
+  -- If the application window is not wide enough,
+  --  the tips window is narrower and taller to fit beside the main menu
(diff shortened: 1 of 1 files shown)


TheCycoONE:
Could you add some screenshots to the PR to illustrate it

lewri:
> 800x600
> Screenshot 2024-02-08 at 13 08 19

Should the height not resize itself to the size of the tip?
If we're going to adapt to the resolution horizontally, it might as well also dynamically adjust its height too?

Alberth289346:
I'd just move the main menu to the left somewhat, the empty sky isn't very interesting anyway.

Not sure that is enough for the tooltip window  though

lewri (approved):
More unlabelled numbers, but that is generally how these windows currently roll.

Issue #2432, opened 2023-10-03
Endcondition check seems incorrectly implemented


kl-mel:
### Describe the issue

Hello, I don't know if this is a bug with the game but I'm one of the custom corsixth levels "Dethsdaw".

I have met all the requirements (above 900 reputation, 90% cured, hospital value, bank balance which is a target of 450,000 and I'm now at over 3 million haha) but the level never ends.

I am aware of the fact that if you hold down z to fast forward it won't register a game win, but that isn't the problem here.

### Steps to Reproduce

1. *Load save game attached to this ticket*

2. wait until the first of the month, where usually the level would end

3.

### Expected Behaviour

I expected the level to end as I have matched all the requirements.

### Save Game

Save game attached to report: yes
[dethsdaw2.sav.zip](https://github.com/CorsixTH/CorsixTH/files/12790900/dethsdaw2.sav.zip)

### CorsixTH Version

0.66

### Operating System

macOS monterey

### Theme Hospital Version

GOG.com

### Gamelog.txt

```bash
Compiled with api_version: 2682, arch: x86_64, audio: true, os: macos
SDL renderer: metal
Lua 5.4 run with api version: 2682, game version: v0.66, savegame version: 170
Created game with savegame version 170.
Savegame version is 170 (v0.66), originally it was 170 (v0.66).
Savegame version is 170 (v0.66), originally it was 170 (v0.66).
Savegame version is 170 (v0.66), originally it was 170 (v0.66).
Savegame version is 170 (v0.66), originally it was 170 (v0.66).
Warning: Trying to remove nonexistent room build callback (function: 0x60000344a5e0) from humanoid (table: 0x600008855d80).
Warning: Trying to remove nonexistent room remove callback (function: 0x600001defb10) from humanoid (table: 0x600008855d80).
Warning: Trying to remove nonexistent staff callback (function: 0x6000034481e0) from humanoid (table: 0x600008855d80).
Warning: Trying to resolve door deadlock at (39, 75)
Savegame version is 170 (v0.66), originally it was 170 (v0.66).
Savegame version is 170 (v0.66), originally it was 170 (v0.66).
Warning: Trying to remove nonexistent room build callback (function: 0x6000035dd500) from humanoid (table: 0x60000445cc80).
Warning: Trying to remove nonexistent room remove callback (function: 0x600001c666d0) from humanoid (table: 0x60000445cc80).
[...]

PR #2484, opened 2024-02-06
Add groups to win/lose criteria

### LINKED ISSUE #232: Progress Report shows incorrect information with loans
~~This was raised in the forum http://forums.corsixth.com/index.php/topic,5497.msg7713.html#new~~ Forum gone, pasted post below

> Hello,
> 
> I noticed that CorsixTH behaves different then the regular TH when it comes to the amount of your bank balance in regards to level failing targets. The difference is that loans are applied in CorsixTH
> 
> For example: On level 7 you fail if you are more then $1000 in debt.
> 
> So when I'm 500$ in debt, the Progress Report shows a red column thats 1/2 in height and the description reads: "Your bank account balance must be at least -1000. Currently ists -500".
> 
> I then take a loan of $5000 and therefore have $4500 in the bank which I can spend. Here's where it becomes different:
> 
> In the original TH: The Progress Report column  turns green again (with minimum height). The description still says I'm -500 but also tells me how much I need to win the level. Original TH apparently just cares about what you have in the bank to spend and I don't fail the level at the end of the year.
> 
> In CorsixTH: The Progress Report column remains red and 1/2 in height, telling me that I'm -500 and will loose at -1000. CorsixTH apparently calculates:
> 
> $4500 in the bank
> - $5000 in loans
> = -500 bank balance.
> 
> This difference makes CorsixTH so much harder then the original, because with level 7's -1000 requirement you are basically prohibited from keeping a loan untill the following year, and in the beginning when you have to expand your hospital that's (at least for me) almost impossible.
> 
> Hope you understand what I meant. Was this changed on purpose or is it a mix up?
> 
> Thanks!
> Hage

The red bar shows even if your balance is above the threshold for losing.  I think this is because on lines 204 - 207 we make a special case for the balance.
The special case should only apply to winning and losing, I think by having these lines for the bar is giving the wrong information.   For those that don't remember or know what the special case is, this was that it should be possible to get your bank balance above the losing threshold with loans as otherwise there is no point in having loans.  You can't though win with a loan. What shows in the status screen though is that you are still about to lose, whereas it should show that you are now safe from losing.
I have checked this against TH and what is said in the topic is right.

tobylane:
- Collect level goals into their groups
- Check each group for win and lose game endings
- Make the progress report dialog show the most relevant group

Fixes #232

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
@@ -28,7 +28,6 @@ function UIProgressReport:UIProgressReport(ui)
   self:UIFullscreen(ui)
 
   local world = self.ui.app.world
-  local hospital = ui.hospital
   local gfx   = ui.app.gfx
 
   if not pcall(function()
@@ -49,53 +48,34 @@ function UIProgressReport:UIProgressReport(ui)
 
   -- Selected hospital number
   self.selected = 1
+  local hospital = world.hospitals[self.selected]
 
+  -- Collect which criteria to show here, draw columns in draw
   -- Add the icons for the criteria
   local x = 263
-  local world_goals = world.goals
-  for _, tab in ipairs(world_goals) do
-    local crit_name = world.level_criteria[tab.criterion].name
-    local res_value = world_goals[crit_name].win_value
-    world_goals[crit_name].visible = true
-    -- Special case for money, subtract loans
-    local cur_value = hospital[crit_name]
-    if crit_name == "balance" then
-      cur_value = cur_value - hospital.loan
+  local crit_data = world.endconditions:generateReportTable(hospital)
+  for _, crit_table in ipairs(crit_data) do
+    crit_table.visible = true
+    local crit_name = crit_table.name
+    local res_value = crit_table.win_value
+    local cur_value = world.endconditions:getAttribute(hospital, crit_name)
+    if crit_table.lose_value then
+      crit_table.red = true
+      res_value = crit_table.lose_value
     end
-    if world_goals[crit_name].lose_value then
-      world_goals[crit_name].red = false
-
-      if cur_value < world_goals[crit_name].boundary then
-        world_goals[crit_name].red = true
-        res_value = world_goals[crit_name].lose_value
-        -- TODO: Make the ugly workaround for the special case "percentage_killed" better
-        if crit_name:find("killed") then
-          res_value = nil
-          world_goals[crit_name].visible = false
-        end
-      elseif not world_goals[crit_name].win_value then
-        world_goals[crit_name].visible = false
-      end
+    -- FIXME: res_value and cure_value are depersisted as floating points, using
+    -- string.format("%.0f", x) is not suitable due to %d (num) param in _S string
+    local tooltip
+    if crit_table.formats == 2 then
+      tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value), math.floor(cur_value))
+    else
+      tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value))
     end
-    -- Only five criteria can be there at once.
-    if crit_name:find("killed") and world.winning_goal_count > 5 then
-      res_value = nil
-      world_goals[crit_name].visible = false
-    end
-    if res_value then
-      -- FIXME: res_value and cure_value are depersisted as floating points, using
-      -- string.format("%.0f", x) is not suitable due to %d (num) param in _S string
-      local tooltip
-      if world.level_criteria[tab.criterion].formats == 2 then
-        tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value), math.floor(cur_value))
-      else
-        tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value))
-      end
-      self:addPanel(world.level_criteria[tab.criterion].icon, x, 240)
-      self:makeTooltip(tooltip, x, 180, x + 30, 180 + 90)
-      x = x + 30
-    end
+    self:addPanel(crit_table.icon, x, 240)
+    self:makeTooltip(tooltip, x, 180, x + 30, 180 + 90)
+    x = x + 30
   end
+  self.crit_data = crit_data
 
   self:addPanel(0, 606, 447):makeButton(0, 0, 26, 26, 8, self.close):setTooltip(_S.tooltip.status.close)
 
@@ -182,9 +162,8 @@ function UIProgressReport:draw(canvas, x, y)
   UIFullscreen.draw(self, canvas, x, y)
 
   x, y = self.x + x, self.y + y
-  local hospital = self.ui.hospital
-  local world    = hospital.world
-  local world_goals = world.goals
+  local world    = self.ui.app.world
+  local hospital = world.hospitals[self.selected]
 
   -- Names of the players playing
   local ly = 73
@@ -194,23 +173,19 @@ function UIProgressReport:draw(canvas, x, y)
     ly = ly + 25
   end
 
-  -- Draw the vertical bars for the winning conditions
+  -- Draw the vertical bars for the selected conditions
   local lx = 270
-  for _, tab in ipairs(world_goals) do
-    local crit_name = world.level_criteria[tab.criterion].name
-    if world_goals[crit_name].visible then
-      local sprite_offset = world_goals[crit_name].red and 2 or 0
-      local cur_value = hospital[crit_name]
-      -- Balance is special
-      if crit_name == "balance" then
-        cur_value = cur_value - hospital.loan
-      end
+  for _, crit_table in ipairs(self.crit_data) do
+    if crit_table.visible then
+      local sprite_offset = crit_table.red and 2 or 0
+      local crit_name = crit_table.name
+      local cur_value = world.endconditions:getAttribute(hospital, crit_name)
       local height
-      if world_goals[crit_name].red then
-        local lose = world_goals[crit_name].lose_value
-        height = 1 + 49 * (1 - ((cur_value - lose)/(world_goals[crit_name].boundary - lose)))
+      if crit_table.red then
+        local lose = crit_table.lose_value
+        height = 1 + 49 * (1 - ((cur_value - lose)/(crit_table.boundary - lose)))
       else
-        height = 1 + 49 * (cur_value/world_goals[crit_name].win_value)
+        height = 1 + 49 * (cur_value/crit_table.win_value)
       end
       if height > 50 then height = 50 end
       local result_y = 0
@@ -242,6 +217,10 @@ function UIProgressReport:afterLoad(old, new)
     self.normal_font = gfx:loadFont("QData", "Font100V", false, palette)
     self.small_font = gfx:loadFont("QData", "Font106V")
     self.panel_sprites = gfx:loadSpriteTable("QData", "Rep02V", true, palette)
+  end
+
+  if old < 188 then
+    self:close()
   end
 
   UIFullscreen.afterLoad(self, old, new)
--- a/CorsixTH/Lua/endconditions.lua
+++ b/CorsixTH/Lua/endconditions.lua
@@ -1,0 +1,235 @@
+--[[ Copyright (c) 2024 Toby "tobylane"
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+of the Software, and to permit persons to whom the Software is furnished to do
+so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE. --]]
+
+-- The "Winning and Losing Conditions" section of level files uses Criteria
+--  numbers from the order of this table.
+-- Icon and formats are used in the progress report dialog.
+local local_criteria_variable = {
+  {name = "reputation",       icon = 10, formats = 2},
+  {name = "balance",          icon = 11, formats = 2},
+  {name = "percentage_cured", icon = 12, formats = 2},
+  {name = "num_cured" ,       icon = 13, formats = 2},
+  {name = "percentage_killed",icon = 14, formats = 2},
+  {name = "value",            icon = 15, formats = 2}, -- Hospital value
+}
+
+-- A table of functions for fetching criteria values that cannot be measured
+--  directly from a hospital attribute of the same name.
+local get_custom_criteria = {
+  balance = function(hospital) return hospital.balance - hospital.loan end,
+}
+
+class "EndConditions"
+---@type EndConditions
+local EndConditions = _G["EndConditions"]
+
+--! Collect the end conditions for this level, if they exist
+--!param level_config (table) The map's level_config, containing criteria.
+--!param level_number (integer) The number of the map, if playing a campaign.
+--!param freebuild (boolean) The free build status of the world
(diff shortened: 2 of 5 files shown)


CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -28,7 +28,7 @@ function UIProgressReport:UIProgressReport(ui)
   self:UIFullscreen(ui)
 
   local world = self.ui.app.world
-  local hospital = ui.hospital
+  self.world  = world
```
TheCycoONE, line 31: not much reason for the local world above then right?

TheCycoONE (approved):

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,323 @@
...
+end
+
+
+class "WinCriteria"
+--@type WinCriteria
+
+--!param hospital (Hospital) The hospital of the tests.
+--!param table (table) the win_criteria table for one group and one criteria
+--!return If report is false, return the gap from hospital value to win value, and the win value.
+-- if true, return the gap from the hospital value to boundary, and the boundary.
```
lewri, line 279: On doc, this only will apply to the top function. I fully understand why you made this decision, but for our documentation system it leaves gaps for sure.

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,323 @@
...
+end
+
+function WinCriteria.value(hospital, table, report)
+  local target = report and table.boundary or table.win_value
+  local value = hospital.value
+  local max_min = table.max_min_win == 1 and 1 or -1
+  if (value - target) * max_min >= 0 then
+    return math.abs(value - target), target
+  end
+end
```
lewri, line 323: You're repeating a lot of the same execution here from a code perspective. If there is a common thread of actions that could be condensed into one local or private function then I'd recommend that. Then call that function each time.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -244,5 +227,9 @@ function UIProgressReport:afterLoad(old, new)
     self.panel_sprites = gfx:loadSpriteTable("QData", "Rep02V", true, palette)
   end
 
+  if old < 182 then
```
lewri, line 230: Needs a bump

CorsixTH/Lua/world.lua
```diff
@@ -2801,6 +2659,12 @@ function World:afterLoad(old, new)
     self:initWallTypes()
   end
 
+  if old < 182 then
```
lewri, line 2662: Need a bump

CorsixTH/Lua/world.lua
```diff
@@ -299,69 +288,11 @@ function World:toggleInformation()
...
-        world_goals[crit_name].max_min_lose = values.MaxMin
-        world_goals[crit_name].group = values.Group
-        world_goals[world_goals[crit_name].number].lose_value = values.Value
-        world_goals[world_goals[crit_name].number].boundary = values.Bound
-        world_goals[world_goals[crit_name].number].criterion = values.Criteria
-        world_goals[world_goals[crit_name].number].max_min_lose = values.MaxMin
-        world_goals[world_goals[crit_name].number].group = values.Group
-      end
-    end
+  if not (self.app.using_demo_files or self.free_build_mode) then
```
lewri, line 293: The demo definitely has end conditions does it not?
Edit: Looks like CorsixTH actually loads the base_config, which give no win/lose condition so this isn't wrong.
But a different issue for later is that the Demo has a defined endCondition in the readme:
```txt
You have to cure 20 patients in a year in this demo so look after your
patients well and provide for their every need.
```
And in the demo itself from DOSBOX
image

CorsixTH/Lua/endconditions.lua
Alberth289346: If you merge the function computations into one as suggested elsewhere, this doc can be attached to that function.

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,323 @@
...
+function WinCriteria.num_cured(hospital, table, report)
+  local target = report and table.boundary or table.win_value
+  local num_cured = hospital.num_cured
+  local max_min = table.max_min_win == 1 and 1 or -1
+  if (num_cured - target) * max_min >= 0 then
+    return math.abs(num_cured - target), target
+  end
+end
+
+function WinCriteria.value(hospital, table, report)
```
Alberth289346, line 316: You may want to rename this one to `hospitalValue`, since `value` is quite generic.

tobylane: It uses the hospital attribute name. I'll add a comment to clarify that.

tobylane:
Converted the identical set of tests into a generic test, with the ability to add specific tests later.

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,285 @@
...
+-- The generic test is used for tests on hospital attributes with the same name as the goal,
+--  eg value is hospital.value
+-- Other tests can be implemented differently.
+
+--!param hospital (Hospital) The hospital of the tests.
+--!param crit_name (string) The attribute of the hospital in this test.
+--!param table (table) the win_criteria table for one group and one criteria
+--!return If report is false, return the gap from hospital value to win value, and the win value.
+-- if true, return the gap from the hospital value to boundary, and the boundary.
+function WinCriteria.generic(hospital, crit_name, table, report)
```
Alberth289346, line 278: Does this work? I am missing the `local WinCriteria = _G["WinCriteria"]` line that makes the `WinCriteria` used here available.

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,285 @@
...
+--  eg value is hospital.value
+-- Other tests can be implemented differently.
+
+--!param hospital (Hospital) The hospital of the tests.
+--!param attribute (string) The attribute of the hospital in this test.
+--!param table (table) the lose_criteria table for one group and one criteria.
+-- Value is the hospital value.
+--!return If report is false, return the gap from hospital value to lose value, and the lose value
+-- if true, return the gap from the hospital value to boundary, and the boundary.
+function LoseCriteria.generic(hospital, attribute, table, report)
```
lewri, line 256: Generic is a worrisome function name. Why not `LoseCriteria:testCriterion`?

CorsixTH/Lua/endconditions.lua
tobylane: It does, even the EndCondition one is optional. I've added them.

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,302 @@
...
+--!param attribute (string) The attribute of the hospital in this test.
+--!param table (table) the lose_criteria table for one group and one criterion.
+-- Value is the hospital value.
+--!return If report is false, return the gap from hospital value to lose value, and the lose value
+-- if true, return the gap from the hospital value to boundary, and the boundary.
+function LoseCriteria.testCriterion(hospital, attribute, table, report)
+  local target = report and table.boundary or table.lose_value
+  local measure = hospital[attribute]
+  local max_min = table.max_min_lose == 1 and 1 or -1
+  if (measure - target) * max_min > 0 then
```
tobylane, line 261: This isn't the right comparison, it doesn't warn the user early enough in cases like #232.

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,244 @@
...
+  elseif met_count == total_count then
+    return "win"
+  end
+end
+
+-- Find the group of win conditions best met by the hospital.
+--!param hospital (Hospital) The hospital of the tests.
+--!return best (number) The number of the best group.
+function EndConditions:findBestWinGroup(hospital)
+assert(self.highest_group, "error")
```
tobylane, line 234: Remove this.

tobylane (commented):
The generated parts of EndConditions - win_goals, lose_goals and highest_group, aren't kept in savegames, what's needed?

Added a get function for when a criterion isn't an attribute of the hospital, and a make_getter for when it is. This will make it possible to add more variety of new criteria later. Why does make_getter need to be persistable?

tobylane:
Fixed the issues in my previous comment, and added `EndConditions:getAttribute` to make it easier to change the get attribute functions later.

Fixes #232, #2432

tobylane:
This puts the persisted data in world, a better fix.

Rebased, ready.

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,254 @@
+--[[ Copyright (c) 2023 Toby "tobylane"
```
Alberth289346, line 1: Wrong year?

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,254 @@
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE. --]]
+
+-- The order of this table relates to the "Winning and Losing Conditions" section
```
Alberth289346, line 21:
```suggestion
-- The order of this table is the same as the goals in the "Winning and Losing Conditions" section
```
I am guessing this is what you're aiming for.

"relates to" is too free, you should give enough information such that anyone can check that the table below is accurate and/or fix it if the table is incorrect for whatever reason.

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,254 @@
...
+          max_min_lose = values.MaxMin,
+          icon = local_criteria_variable[values.Criteria].icon,
+          formats = local_criteria_variable[values.Criteria].formats,
+          start = start[crit_name] or 0,
+        }
+        if values.Group > highest_group then highest_group = values.Group end
+      end
+    end
+  end
+  return self, win_goals, lose_goals, highest_group
```
Alberth289346, line 102: The class has no data (no `self.x` values). As such there is no need to instantiate the class.

Code will work fine if you treat it as a bunch of static functions.

As an example, consider `function EndConditions:checkWinGroup(hospital, win_table, findbest)`. Change it to `function EndConditions.checkWinGroup(hospital, win_table, findbest)` (colon becomes a dot).

As `self` doesn't exist without instantiating the class, you change calls to the function from eg `local reason = self:checkWinGroup(hospital, tbl)` to `local reason = EndConditions.checkWinGroup(hospital, tbl)` ( `self:` becomes `EndConditions.`). 

What happens is close to
``` lua
EndConditions = {}
EndConditions.checkWinGroup = function (hospital, win_table, findbest) ... end

local reason = EndConditions.checkWinGroup(hospital, tbl)
```
That is, just a bunch of functions stored in a table. Call such a function is pointing at it, and then append `(....)` to say "execute it", as usual with a function.

Alberth289346: The other direction of course is to have an instance of this class, and then you can store the data that you now return from the constructor. It then becomes available for all other methods so you don't need to drag that around in other variables.

tobylane: I did that at first, but it wasn't saved. Is it better to have it save in World, or find out why it can't save in EndConditions?

CorsixTH/Lua/world.lua
```diff
@@ -299,69 +288,11 @@ function World:toggleInformation()
...
-    end
-  end
-
-  -- Order the criteria (some icons in the progress report shouldn't be next to each other)
-  table.sort(world_goals, function(a,b) return a.criterion < b.criterion end)
-  self.goals = world_goals
-  self.winning_goal_count = winning_goal_count
+  if self.free_build_mode then return end
+  self.endconditions, self.win_goals, self.lose_goals, self.highest_group =
+      EndConditions:EndConditions(self.map.level_config, self.map.level_number)
```
Alberth289346, line 295: If doing the checks in EndConditions is the goal, why are you not storing the data needed for it inside that instance?

If you want to do it like you propose, at least wrap a table around all the data you put in world, so there is only one variable in the class containing the data.

We already have too many variables in world, please avoid making (many) more.

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,254 @@
...
+  for crit_name, crit_table in pairs(win_table) do
+    local max_min = crit_table.max_min_win == 1 and 1 or -1
+    if (self:getAttribute(hospital, crit_name) - crit_table.win_value) * max_min >= 0 then
+      met_count = met_count + 1
+    end
+    total_count = total_count + 1
+  end
+
+  if findbest then
+    if met_count > 0 and total_count > 0 then
```
Alberth289346, line 222: `total_count` is always at least `met_count`, so if the former holds, there is no need to test the latter.

You can add an assert if you are afraid this relation may get broken.

CorsixTH/Lua/endconditions.lua
```diff
@@ -0,0 +1,254 @@
...
+          criterion = values.Criteria,
+          max_min_lose = values.MaxMin,
+          icon = local_criteria_variable[values.Criteria].icon,
+          formats = local_criteria_variable[values.Criteria].formats,
+          start = start[crit_name] or 0,
+        }
+        if values.Group > highest_group then highest_group = values.Group end
+      end
+    end
+  end
```
Alberth289346, line 101: Looks like a copy of the win-criteria code just above it, except for the outer names `lose`, `lose_criteria`, and `lose_goals`.

If so, refactor this function to a private function and pass in the variables I just listed to the function, so you avoid copy/pasting the entire body.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -49,52 +48,35 @@ function UIProgressReport:UIProgressReport(ui)
...
-    if res_value then
       -- FIXME: res_value and cure_value are depersisted as floating points, using
       -- string.format("%.0f", x) is not suitable due to %d (num) param in _S string
       local tooltip
-      if world.level_criteria[tab.criterion].formats == 2 then
+      if crit_table.formats == 2 then
         tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value), math.floor(cur_value))
       else
         tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value))
       end
```
Alberth289346, line 74: Wouldn't `formats` be better named as `number_of_values` ?

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -49,52 +48,35 @@ function UIProgressReport:UIProgressReport(ui)
...
-        world_goals[crit_name].red = true
-        res_value = world_goals[crit_name].lose_value
-        -- TODO: Make the ugly workaround for the special case "percentage_killed" better
-        if crit_name:find("killed") then
-          res_value = nil
-          world_goals[crit_name].visible = false
-        end
-      elseif not world_goals[crit_name].win_value then
-        world_goals[crit_name].visible = false
+  if not world.free_build_mode then
```
Alberth289346, line 56: Can't you have empty win and lose goals instead?

That avoids having this check here.

CorsixTH/Lua/endconditions.lua
TheCycoONE: Look into why the class isn't persisting.

Alberth289346: Indeed, I see no reason why you couldn't store data in a class instance in world.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
tobylane: I have plans for other formats that are two values,
```lua
       elseif crit_table.formats == 3 then
         tooltip = tooltip:format(math.floor(res_value) / 100, math.floor(cur_value) / 100)
```

tobylane:
Addressed feedback, improved comments and rebased.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -49,53 +48,34 @@ function UIProgressReport:UIProgressReport(ui)
...
-  for _, tab in ipairs(world_goals) do
-    local crit_name = world.level_criteria[tab.criterion].name
-    local res_value = world_goals[crit_name].win_value
-    world_goals[crit_name].visible = true
-    -- Special case for money, subtract loans
-    local cur_value = hospital[crit_name]
-    if crit_name == "balance" then
-      cur_value = cur_value - hospital.loan
+  local crit_data = world.endconditions:generateReportTable(hospital)
+  for i, crit_table in ipairs(crit_data) do
```
Alberth289346, line 57: `i` seems unused, by convention we write that as `_`.

Alberth289346:
LGTM

Issue #2483, opened 2024-02-03
[Bug] Global Audio off makes ffmpeg throw errors in movies


lewri:
### Describe the issue

When global audio is disabled in CorsixTH ffmpeg is being told incorrect data about audio channels. It's likely we need to do something but a warning generated isn't correct.

### Steps to Reproduce

1. Start CorsixTH
2. Turn global audio to off 
3. Restart CorsixTH
4. See warning in console.

### Expected Behaviour

In CorsixTH 0.65 we just get a null chunk error.
In CorsixTH 0.66 the second error (in red) appears with the null chunk.

### Save Game

Save game attached to report: No

### CorsixTH Version

0.67

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Additional Information

![image](https://github.com/CorsixTH/CorsixTH/assets/20030128/dd57d3f8-abb6-4244-b8ca-2a083aeed857)

Also from a different PC:
WARN: unsupported channel layout 1717531493. Please report issue.[SWR @ 0000024B49457000] Output channel layout "1717531493 channels" is invalid or unsupported.
[SWR @ 0000024B49457000] Rematrix is needed between stereo and 1717531493 channels but there is not enough information to do it

TheCycoONE:
Can reproduce

PR #2487, opened 2024-02-08
Disable audio in movies when there are no channels

TheCycoONE:
Fixes: #2483


Alberth289346:
LGTM

lewri (approved):
Error resolved on both systems (even the billion channel one)

PR #2489, opened 2024-02-15
Fix shutdown order

TheCycoONE:
Also show error message if SDL_WaitEvent fails.

If SDL_Mixer is used by any destructor owned by lua, e.g. the movie player then it needs to still be available when lua is closed.

### DIFF
--- a/CorsixTH/Src/sdl_core.cpp
+++ b/CorsixTH/Src/sdl_core.cpp
@@ -140,7 +140,8 @@ int l_mainloop(lua_State* L) {
       SDL_AddTimer(usertick_period_ms, timer_frame_callback, nullptr);
   SDL_Event e;
 
-  while (SDL_WaitEvent(&e) != 0) {
+  int wait_error = 0;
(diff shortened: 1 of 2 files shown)


CorsixTH/Src/sdl_core.cpp
```diff
@@ -296,6 +297,10 @@ int l_mainloop(lua_State* L) {
     lua_gc(L, LUA_GCSTEP, 2);
   }
 
+  if (wait_error != 0) {
+    std::fprintf(stderr, "%s\n", SDL_GetError());
+  }
```
Alberth289346, line 302: A few lines above do more `PollEvent` (274, 293), does the error survive that?

Alberth289346: High level structure is not very visible imho, but otherwise LGTM

TheCycoONE: Yes, the complexity of this function is too high. Put it in the refactor list :+1:

PR #2490, opened 2024-02-16
Update luarocks script

### LINKED ISSUE #2213: macos_luarocks not parsing correctly?
When I fire off a build of CorsixTH, at some point in the process I see this:

```
-- fixup_bundle: done
Usage: luarocks install [-h] [--keep] [--force] [--force-fast]
       [--only-deps] [--no-doc] [--verify] [--check-lua-versions]
       [--deps-mode <mode>] [--no-manifest] [--pin] <rock> [<version>]

Error: malformed argument 'dylib (compatibility version 5.4'
```

As near as I can tell, `macos_luarocks` is parsing the `otool` output in order to ascertain the Lua version to pass to luarocks.

So this line

`lua="$(otool -L ../MacOS/CorsixTH | grep 'lua' | cut -d. -f4-5)"`

is parsing out this

`@executable_path/../Frameworks/liblua.dylib (compatibility version 5.4.0, current version 5.4.4)`

And coming up with this

`dylib (compatibility version 5.4`

when it wants to come up with just `5.4`

Probably wouldn't be hard to fix, except it's odd if I'm running into this and no one else is. Any idea what I'm doing wrong?

tobylane:
*Fixes #2213*

- Improve the macos_luarocks script. The normal use, as in cmake install, is the same, but the settings allow some choices for other setups. Detect lua version and openssl path better.


colinjmatt (approved):
Looks good to me and builds fine.

TheCycoONE:
seems like this needs a kick, not sure why github is stuck.

PR #2491, opened 2024-02-16
Add Animview, 32bit, pr checkout on demand to Windows Actions

tobylane:
- Add on demand builds, with options for 32bit, Animview builds, of a specific pr. Here's a 32bit build with Animview of 2489 (a recent PR with cpp changes) https://github.com/tobylane/CorsixTH/actions/runs/7934724743/job/21666299492
- Name the output
- Update old steps

### DIFF
--- a/.github/workflows/Windows.yml
+++ b/.github/workflows/Windows.yml
@@ -1,47 +1,101 @@
 ---
 name: Windows
 
-on:  # yamllint disable-line rule:trurhy
+on:
   push:
     branches-ignore:
       - 'gh-pages'
   pull_request:
+  workflow_dispatch:
+    inputs:
+      preset:
+        description: 'Select CMake preset'
+        default: 'win-x64-rel'
+        type: choice
+        options:
+          - win-dev
+          - win-x64-rel
+          - win-x86-rel
+      animview:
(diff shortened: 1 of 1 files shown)


lewri (commented):
Not against 32 bit being still feasible as long as it doesn't have any limitations for us in future

TheCycoONE:
This is going to conflict with my PR

TheCycoONE:
Much of it should be redone to use cmake presets and the vcpkg addon. Building all three of the presets i added plus installing doxigen and either lua/lfs or using the vcpkg version like i did with appveyor

TheCycoONE (changes requested):
Needs to be updated to take advantage of the cmake --prefix options that I added and the new vcpkg mechanics.

tobylane:
Added a configure-only step for win-dev. Lfs, etc are included. Do you want docs built on windows?

[x86 run](https://github.com/tobylane/CorsixTH/actions/runs/8400619622/job/23008097759), [recent win-dev run](https://github.com/tobylane/CorsixTH/actions/runs/8400445325).

TheCycoONE:
I don't think we need docs here, as we have them in the linux action. Merged.

PR #2492, opened 2024-02-16
Spawn a patient on open

lewri:
This aims to partially address the issue of #649 by forcing a patient to be given on open https://github.com/CorsixTH/CorsixTH/issues/649#issuecomment-1279949701 . It doesn't address the difference in wait of that issue but should at least make the first patient a little more consistent.

- Moves hospital modification out of `UIWatch`
- Spawns a patient on open of hospital (providing you built a reception desk, of course!)
- Gives the `AIHospital` a function path to open for future use.

RFC needed because one oversight here is that some players will open their hospital as soon as they start, evading this change. Should another consideration be made for this instance (e.g. if open without desk, spawn patient once a desk gets made the first time).

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -2365,6 +2365,16 @@ function Hospital:getRandomBusyRoom()
   if #chosen_room.door.queue >= busy_threshold then return chosen_room end
 end
 
+--! Open the hospital
+function Hospital:open()
+  self.opened = true
+end
+
+--! Check if the hospital is ready to accept patients
+function Hospital:canAcceptPatients()
+  return self.opened and self:hasStaffedDesk()
+end
+
 ---- Stubs section - these functions have nothing to do here, are overridden in a derived class.
 
 --! Give advice to the user about the need to buy the first reception desk.
--- a/CorsixTH/Lua/hospitals/player_hospital.lua
+++ b/CorsixTH/Lua/hospitals/player_hospital.lua
@@ -483,9 +483,15 @@ end
 
 --! Called at the end of each day.
 function PlayerHospital:onEndDay()
-  -- Advise the player.
-  if self:hasStaffedDesk() then
+  if self:canAcceptPatients() then
+    -- Are we providing advice today?
     self:dailyAdviceChecks()
+
+    -- If we haven't spawned our first patient yet for the player make it happen so
+    -- they don't wait too long
+    if self.num_visitors == 0 then
+      self.world:spawnPatient(self)
+    end
   end
 
   -- check if we still have to announce VIP visit
--- a/CorsixTH/Lua/dialogs/watch.lua
+++ b/CorsixTH/Lua/dialogs/watch.lua
@@ -101,7 +101,7 @@ function UIWatch:onCountdownEnd()
       end
     end
   elseif self.count_type == "initial_opening" then
-    self.ui.hospital.opened = true
+    self.hospital:open()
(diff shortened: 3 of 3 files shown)


Alberth289346:
Indeed, just spawning a patient looks like a too simple approach. I don't think it's fair to punish a user for pressing the watch button too early (a spawn will fail without a reception desk and receptionist, and thus no early patient will arrive).

To fix, the meaning of pressing the watch button should change. In the current proposal, pressing the button opens the hospital unconditionally. In other words, the user is literally opening the hospital with the button. If pressing the button is seen as an intent (to the user, it is ok to receive patients), a hospital can add its own additional conditions.   
This means the `open` value and the hospital conditions (or just the conditions if they include checking the open flag) must be monitored until everything is fulfilled.

Triggering on buying a reception desk is not going to work, since getting a path needs a reception desk AND a receptionist. I am sure it's possible to fix by adding more code for that check, but in my view the simpler solution is to check regularly whether the first patient can be sent, and if so, do that.  
The next problem is then to decide where to do that. Several options exist, my first guess would be the patient spawning code that handles timing of the next patient. That avoids special case code, and keeps patient spawning at a single spot.

lewri:
> In other words, the user is literally opening the hospital with the button.

Yes, that is the intent of that button as per the original was. So I don't think we should change the intent of the button at all.

I think what you're suggesting instead is a `tickDay` event for a patient 0 if the spawn on open fails. I don't really want it to become a callback, since those do get lost into the void. One thing that would be debatable is having an afterload to correct existing saves in the "open but not reception desk" state. By all means it can be done but I wouldn't be for it for the benefit and small window.

lewri:
Updated it to now run as an `endDay` check separate from the watch. Will only trigger once the hospital is opened and has a staffed desk.

CorsixTH/Lua/hospitals/player_hospital.lua
```diff
@@ -483,9 +483,15 @@ end
...
   if self:hasStaffedDesk() then
+    -- Are we providing advice today?
     self:dailyAdviceChecks()
+
+    -- Can we take our first patient?
+    if self.opened and self.num_visitors == 0 then
+      -- Give player a patient so they don't wait too long
+      self.world:spawnPatient(self)
+    end
   end
```
lewri, line 495: @tobylane if the daily advisor checks could instead be tested against the below would that work?
```lua
function Hospital:canAcceptPatients()
  return self:hasStaffedDesk() and self.opened
end
```
(currently we just statically wait 5 months, regardless if the player is in motion or not)

CorsixTH/Lua/hospital.lua
```diff
@@ -2365,6 +2365,16 @@ function Hospital:getRandomBusyRoom()
...
 end
 
+--! Open the hospital
+function Hospital:open()
+  self.opened = true
+end
+
+--! Check if the hospital is ready to accept patients
+function Hospital:canAcceptPatients()
+  return self:hasStaffedDesk() and self.opened
```
Alberth289346, line 2375:
```suggestion
  return self.opened and self:hasStaffedDesk()
```
Do the cheap check first to avoid wasting CPU cycles.

Alberth289346:
Condition in the daily check is a bit tricky, as there is no direct connection from world to hospital to set the number of visitors, but it does seem to exist through patient, by reading some code.

CorsixTH/Lua/hospital.lua
lewri: I agree, checking a single variable is better.

PR #2493, opened 2024-02-17
Vcpkg manifest

TheCycoONE:
Switch to a more modern and standard method of using vcpkg.

There are a number of advantages:
 - versions of individual packages can be more easily controlled
 - our own overlays can be maintained in their own separate repo
 - take advantage of vcpkg artifact caching on local machines.

Also introduce cmake presets. For now only windows presets are included, but we can extend the list to make common build options easier for all targets.

Note that this is a breaking change to windows developers using the existing USE_VCPKG_DEPS. When merged, the wiki page needs updated and something announced.

If you previously built using USE_VCPKG_DEPS you need to migrate
to the new build method.

1. Delete your CorsixTH/vcpkg directory.
2. Delete your build directory
3. Download and install vcpkg following the instructions at:
   https://learn.microsoft.com/en-us/vcpkg/get_started/get-started
4. Run `cmake --prefix dev`
5. Open the visual studio project for your new build directory at
   `./build/dev`


lewri:
Appveyor console logs are now higher than before, is that expected? (went from 1800 to 4600+ lines -- mostly from fast-forward lines)

TheCycoONE:
Doesn't bother me, but as it's not really related I can pull in quiet mode.

TheCycoONE:
Added migration instructions to the first commit message, and quieted the git pull in appveyor per Lewri's suggestion.

lewri:
> Note that this is a breaking change to windows developers using the existing USE_VCPKG_DEPS. When merged, the wiki page needs updated and something announced.

@TheCycoONE has How to Compile been updated with this info now?

lewri:
> > Note that this is a breaking change to windows developers using the existing USE_VCPKG_DEPS. When merged, the wiki page needs updated and something announced.
> 
> @TheCycoONE has How to Compile been updated with this info now?

Yes you did but I found out that one section got left behind near the requisites table.

PR #2494, opened 2024-02-17
Adjust game speed and ticks

### LINKED ISSUE #1494: Game speed varies from TH
#### Describe the issue

Current timer ticks are at 30ms intervals.

The original would most likely be using 18.2ms interrupts.

Running dosbox side by side and starting both the first level and actually not doing anything.

Dosbox gets to Oct whilst Corsix is around finishing June at normal speed.

Changing the SDL timer ticks to 18ms gets a pretty consistent value.

Likewise the higher speed (And then some more) Corsix gets to end of year when TH in dosbox gets to around end of July. (hours_per_tick at 3, means 3 ticks every 30ms with the current timer which is 2 days per second and the original looks to be 18.2 * 50 or just under a second per day.

At the slowest speed, Corsix is much faster than TH also around 13.5 sec rather than around 50-52 secs.

Changing the timer to 36ms we can approximate fairly well the 18.2 ms timer base.

```lua
local tick_rates = {
  ["Pause"]              = {0, 1},
  ["Slowest"]            = {1, 28}, -- 1 tick per s
  ["Slower"]             = {1, 6}, -- 1 tick per 216ms
  ["Normal"]             = {2, 3}, -- 2 ticks per 108ms = 1 tick per 54 ms
  ["Max speed"]          = {1, 1}, -- 1 tick per 36 ms
  ["And then some more"] = {2, 1}, -- 2 ticks per 36 ms = 1 tick 18ms
  ["Speed Up"]           = {4, 1},
}
```
The 'Speed up' would be nearly equivalent of the old And then some more, at a tick per 9ms instead of 10 and a bit slower than the 7.5ms it would currently result in.

The above values, result in approx seconds per game day of 
Slowest - 50
Slower - 10.8
Normal - 2.7
Max - 2
And then some more - 1
Speed up - 0.5

The current rate on slowest speed, without using 'build when paused' is a worse experience than the original when you use it to build something that is missing and need to do it quickly, and at 'And then some more' you have more time to respond to say a message.

tobylane:
*Fixes #1494*

- Add the new tick rates mugmuggy researched in #1494 
- Bump api and savegame versions, restore previous speed

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -94,7 +94,7 @@ function World:World(app)
   self.next_earthquake = { active = false }
 
   -- Time
-  self.hours_per_tick = 1
+  self.hours_per_tick = 2
   self.tick_rate = 3
   self.tick_timer = 0
   self.game_date = Date() -- Current date in the game.
@@ -878,11 +878,11 @@ end
 -- happens.
 local tick_rates = {
   ["Pause"]              = {0, 1},
-  ["Slowest"]            = {1, 9},
-  ["Slower"]             = {1, 5},
-  ["Normal"]             = {1, 3},
-  ["Max speed"]          = {1, 1},
-  ["And then some more"] = {3, 1},
+  ["Slowest"]            = {1, 28}, -- 1 tick per s
+  ["Slower"]             = {1, 6}, -- 1 tick per 216ms
+  ["Normal"]             = {2, 3}, -- 2 ticks per 108ms = 1 tick per 54 ms
+  ["Max speed"]          = {1, 1}, -- 1 tick per 36 ms
+  ["And then some more"] = {2, 1}, -- 2 ticks per 36 ms = 1 tick 18ms
   ["Speed Up"]           = {4, 1},
 }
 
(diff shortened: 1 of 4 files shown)


TheCycoONE:
looks good to me

CorsixTH/Lua/world.lua
```diff
@@ -878,11 +878,11 @@ end
 -- happens.
 local tick_rates = {
   ["Pause"]              = {0, 1},
-  ["Slowest"]            = {1, 9},
-  ["Slower"]             = {1, 5},
-  ["Normal"]             = {1, 3},
-  ["Max speed"]          = {1, 1},
-  ["And then some more"] = {3, 1},
+  ["Slowest"]            = {1, 28}, -- 1 tick per s
```
lewri, line 881: Not you but would you be able to reword the comment above? I assume "world click" is meant to be "world tick"?

Otherwise I'd like to see how it plays now with the new speeds.

lewri (approved):
Dates for normal speed are pretty much perfectly aligned now. No impact on anims.

This does seem to expose a new issue though where going from speed 1 to 3, it has to wait a whole `speed 1` tick (1s) before the speed change takes place. In reality of TH, the change is pretty instantaneous.

PR #2497, opened 2024-02-18
Choose Closest Spawn for First Patient

lewri:
Another PR to partially address the contents of #649 

- Runs a path comparison calculation to get the closest path between all spawn points and reception desks, then chooses the closest spawn point for the first player's patient.
- In level 1, this brings the time for the first patient to reach the hospital entrance ~19 days.

Small potential hole here is `_findClosestSpawnToDesk` can return false/nil here if it fails to find any paths. Should that be safeguarded?

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -474,6 +474,37 @@ local function isDiseaseUsableForNewPatient(self, disease, hospital)
   error("disease has neither visuals_id or non_visuals_id")
 end
 
+--! Private function to assess the shortest path between all spawn points and desks
+--!param staffed_desks (table) All staffed reception desks
+--!param spawns (table) All available spawn points
+--!return Optimal spawn point to the nearest reception desk or false on failure
+function World:_findClosestSpawnToDesk(staffed_desks, spawns)
+  local best_spawn_point, distance
+  local function testNewSpawnPath(desk, spawn)
+    local new_dist = self:getPathDistance(desk.tile_x, desk.tile_y, spawn.x, spawn.y)
+    if not new_dist then return end
+    if not distance or new_dist < distance then
+      distance = new_dist
+      best_spawn_point = spawn
+    end
+  end
+
+  for _, spawn in ipairs(spawns) do
+    for _, desk in ipairs(staffed_desks) do
+      testNewSpawnPath(desk, spawn)
+    end
+  end
+
+  if not best_spawn_point then
+    local msg =
+        "Warning: Path not found between any spawn point and staffed desk. " ..
+        "Humanoid will still likely find its own way."
+    self:gameLog(msg)
+    return false
+  end
+  return best_spawn_point
+end
+
 --! Spawn a patient from a spawn point for the given hospital.
 --!param hospital (Hospital) Hospital that the new patient should visit.
 --!return (Patient entity) The spawned patient, or 'nil' if no patient spawned.
@@ -511,7 +542,12 @@ function World:spawnPatient(hospital)
   end
 
   -- Construct patient.
-  local spawn_point = self.spawn_points[math.random(1, #self.spawn_points)]
+  local spawns = self.spawn_points
+  -- Let the first patient to the player use the shortest path to a reception desk
+  local spawn_point = hospital:isPlayerHospital() and not hospital:hadPatients() and
+      self:_findClosestSpawnToDesk(hospital:getStaffedDesks(), spawns) or
+      spawns[math.random(1, #spawns)]
+
   local patient = self:newEntity("Patient", 2)
   patient:setDisease(disease)
   patient:setNextAction(SpawnAction("spawn", spawn_point))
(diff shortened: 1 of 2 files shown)


Alberth289346:
> Small potential hole here is _findClosestSpawnToDesk can return false/nil here if it fails to find any paths. Should that be safeguarded?

Yes, in the sense that the program should not crash.  
Under the assumption that a spawn attempt will silently fail if there is no path, you can just return any of the spawn-points.

However, paths outside the hospital are not editable to a user, so perhaps the question is, why is such a level even being played?

CorsixTH/Lua/world.lua
```diff
@@ -511,7 +534,15 @@ function World:spawnPatient(hospital)
   end
 
   -- Construct patient.
-  local spawn_point = self.spawn_points[math.random(1, #self.spawn_points)]
+  local desks = hospital:findReceptionDesks()
+  local spawns = self.spawn_points
+  local spawn_point = spawns[math.random(1, #spawns)]
+
+  -- Let the first patient to the player use the smallest path to a reception desk
+  if hospital.num_visitors == 0 and class.is(hospital, PlayerHospital) then
```
Alberth289346, line 542: Condition is becoming more complicated. It looks like a question that you should ask the hospital. By doing that you can handle the class check in a much nicer way too.

lewri: It's an odd thing to think of how to make the function useful. But yeah we should really be asking the hospital for this so I'll think on it.

CorsixTH/Lua/world.lua
```diff
@@ -511,7 +534,15 @@ function World:spawnPatient(hospital)
...
+    spawn_point = self:_findClosestSpawnToDesk(desks, spawns)
```
lewri, line 543: So in theory this problem should _never happen_ since we run code at the start of  `World:spawnPatient` to check spawn points and a staffed desk. 
I'd probably like to just make this a gamelog warning with mitigation first now than throw a full on error. In my testing of this even if for some reason a valid path failed the patient still manages to find their way.

lewri: Gamelog warning added and changed for risk mitigation

Alberth289346 (commented):
Apparently just writing comments doesn't work if a review is requested :(

CorsixTH/Lua/world.lua
```diff
@@ -474,6 +474,36 @@ local function isDiseaseUsableForNewPatient(self, disease, hospital)
...
+    local new_dist = self:getPathDistance(desk.tile_x, desk.tile_y, spawn.x, spawn.y)
+    if not new_dist then return end
+    if not distance or new_dist < distance then
+      distance = new_dist
+      best_spawn_point = spawn
+    end
+  end
+
+  for _, spawn in ipairs(spawns) do
+    for _, desk in pairs(desks) do
```
Alberth289346, line 493: It's a list, right? 

Maybe also name it `staffed_desk` to express that property clearly.

CorsixTH/Lua/world.lua
```diff
@@ -511,7 +541,12 @@ function World:spawnPatient(hospital)
   end
 
   -- Construct patient.
-  local spawn_point = self.spawn_points[math.random(1, #self.spawn_points)]
+  local desks = hospital:getStaffedDesks()
+  local spawns = self.spawn_points
+  -- Let the first patient to the player use the shortest path to a reception desk
+  local spawn_point = hospital:isPlayerHospital() and not hospital:hadPatients() and
+      self:_findClosestSpawnToDesk(desks, spawns) or spawns[math.random(1, #spawns)]
```
Alberth289346, line 548: Because you wrote this `and-or` construct, you always need `desks`, even if the first patient was already dispatched.

If you change it to recognize that specific case, and only then ask for desks, the hospital doesn't need a staff member to do all the paperwork about staffed desks to the world, for each patient :)

CorsixTH/Lua/world.lua
lewri: Tbh it should use `ipairs`, I forgot to change that in my late-night nitpick commit

CorsixTH/Lua/world.lua
lewri: True, maybe I shouldn't have made it a local and instead just use the function call directly.

CorsixTH/Lua/world.lua
```diff
@@ -474,6 +474,29 @@ local function isDiseaseUsableForNewPatient(self, disease, hospital)
...
+    if not new_dist then return end
+    if not distance or new_dist < distance then
+      distance = new_dist
+      best_spawn_point = spawn
+    end
+  end
+
+  for _, spawn in ipairs(spawns) do
+    for _, desk in pairs(desks) do
+      testNewSpawnPath(desk, spawn)
```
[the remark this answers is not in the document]
lewri: I don't like excessive nesting, extraction is usually cleaner. Plus the early return on 485 was I thought nicer.

CorsixTH/Lua/world.lua
Alberth289346: That would work too indeed.

Alberth289346:
Looks fine to me

PR #2500, opened 2024-02-20
Update persistance.lua

eltociear:
- persistance -> persistence

### DIFF
--- a/CorsixTH/Lua/persistance.lua
+++ b/CorsixTH/Lua/persistance.lua
@@ -63,7 +63,7 @@ function unpermanent(name)
   saved_permanents[name] = nil
 end
 
-local --[[persistable:persistance_global_fetch]] function global_fetch(...)
+local --[[persistable:persistence_global_fetch]] function global_fetch(...)
   local val = _G
   for _, k in ipairs({...}) do
     val = val[k]
@@ -313,7 +313,7 @@ function LoadGame(data)
   TheApp.ui:onChangeResolution()
   -- Check if the blueish tone should be applied.
   -- Note: Blue filter control should be handled from world or ui, however when
-  -- loading a game we should let persistance do it.
+  -- loading a game we should let persistence do it.
   if not TheApp.ui:checkForMustPauseWindows() and TheApp.world:isUserActionProhibited() then
     TheApp.video:setBlueFilterActive(true)
   end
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/persistance.lua
```diff
@@ -63,7 +63,7 @@ function unpermanent(name)
   saved_permanents[name] = nil
 end
 
-local --[[persistable:persistance_global_fetch]] function global_fetch(...)
+local --[[persistable:persistence_global_fetch]] function global_fetch(...)
```
lewri, line 66: I agree that the spelling is indeed unfortunate (even the file is spelt wrong). But there is a greater impact for amending a persistable function's name. I would hazard we don't change it unless we really have to.

CorsixTH/Lua/persistance.lua
```diff
@@ -313,7 +313,7 @@ function LoadGame(data)
   TheApp.ui:onChangeResolution()
   -- Check if the blueish tone should be applied.
   -- Note: Blue filter control should be handled from world or ui, however when
-  -- loading a game we should let persistance do it.
+  -- loading a game we should let persistence do it.
```
lewri, line 316: Let's keep this though. No matter how small a change something ends up it's always appreciated.

lewri (changes requested):
Appveyor did succeed it just made a fuss. Please revert the persistable rename.

CorsixTH/Lua/persistance.lua
Alberth289346: Indeed, that name is used in old savegames, they will break if you change the name.

tobylane:
You could change more of these. All but the first and last from https://grep.app/search?q=persistance&filter[repo][0]=CorsixTH/CorsixTH are worth trying out.

lewri:
> You could change more of these. All but the first and last from [https://grep.app/search?q=persistance&filter[repo][0]=CorsixTH/CorsixTH](https://grep.app/search?q=persistance&filter%5Brepo%5D%5B0%5D=CorsixTH/CorsixTH) are worth trying out.

` corsixth.require("persistance")` also out with the `linux.yml` file, unless you wanted to change the filename.

CorsixTH/Lua/persistance.lua
eltociear: Thanks! Reverted this.

lewri (approved):

PR #2502, opened 2024-02-22
Add a link to the Snap in the install instructions in the README

anthonydillon:
- Added a link to Snapcraft as an option for installing in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ You will need the following:
    - Windows and macOS builds can be downloaded directly from [releases](https://github.com/CorsixTH/CorsixTH/releases).
    - Linux and BSD repositories use either corsixth or corsix-th names [packaged versions](https://repology.org/metapackage/corsixth).
    - A Flatpak for Linux users is available on [Flathub](https://flathub.org/apps/details/com.corsixth.corsixth).
-   - A Snap for Linux users is available on [Snapcraft](https://snapcraft.io/corsixth).
+   - A Snap for Linux users is available on [Snapcraft](https://snapcraft.io/corsixth) [(support page)](https://github.com/snapcrafters/corsixth).
 - We use graphics, sound and other data from the original game so one of the following is required:
    - Original game CD from eBay etc. or your dusty bookshelf:smile:
    - A download from [GOG.com](https://www.gog.com/game/theme_hospital) or [EA](https://www.ea.com/games/theme/theme-hospital)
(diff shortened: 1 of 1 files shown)


lewri:
Are you the maintainer for the snap?

lewri:
We usually only maintain our own builds but will attempt to help with things where a packager might get stuck e.g. flathub/docker

I've pinged @TheCycoONE for checking over though as I have no issue advertising the snap if he's fine with it.

Alberth289346:
It may be useful to make clear that we don't create, maintain or fix these products. For example by stating the owner and giving a contact point or bug report link, for example.

lewri:
That's true. The Android, flatpak, and snap versions all do have GitHub repos but we don't mention them at all.

README.md
```diff
@@ -18,6 +18,7 @@ You will need the following:
    - Windows and macOS builds can be downloaded directly from [releases](https://github.com/CorsixTH/CorsixTH/releases).
    - Linux and BSD repositories use either corsixth or corsix-th names [packaged versions](https://repology.org/metapackage/corsixth).
    - A Flatpak for Linux users is available on [Flathub](https://flathub.org/apps/details/com.corsixth.corsixth).
+   - A Snap for Linux users is available on [Snapcraft](https://snapcraft.io/corsixth).
```
lewri, line 21:
```suggestion
   - A Snap for Linux users is available on [Snapcraft](https://snapcraft.io/corsixth) [(support page)](https://github.com/snapcrafters/corsixth).
```

Maybe we could do this as @Alberth289346 suggests @anthonydillon .

Alberth289346: Trying to have a bit more distinction, please have a look at the 2 commits in https://github.com/Alberth289346/CorsixTH/commits/split-suppliers/

lewri:
I've been made aware we also have https://github.com/nbuechner/corsixth-snap/ 
@nbuechner is your version separate from Snapcraft?

README.md
anthonydillon: I have committed the suggestion but agree it would be more clear to provide the support link for all package managers, as the linked commit does.

lewri: > Trying to have a bit more distinction, please have a look at the 2 commits in https://github.com/Alberth289346/CorsixTH/commits/split-suppliers/

If you wanted to, we could open a pr new for your styling commits now :)

PR #2509, opened 2024-03-01
Delay destruction of intermediate textures.

### LINKED ISSUE #2301: [Bug] Black bars in fullscreen (NVIDIA)
### Describe the issue

Black bar or black flickering really strange bug inside the game all the time. 
image

### Steps to Reproduce

1. just start the game with the GoG Version

### Expected Behaviour

No Black Bars

### Save Game

### CorsixTH Version

0.66

### Operating System

Windows 11

### Theme Hospital Version

CD

### Gamelog.txt

```bash
```

flackr:
This is a possible fix for #2301. It may be that destroying the intermediate texture as soon as we issue a draw call for it is unsafe. This delays the destruction of the texture until the next frame is started.

*Maybe fixes #2301*

- Collects textures that we are finished with on the render target into a vector.
- Destroys those textures when we begin the next frame, i.e. they should be done with at this point.

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -420,7 +420,7 @@ class render_target::scoped_target_texture
     } else {
       target->draw(texture, nullptr, &rect, 0);
     }
-    SDL_DestroyTexture(texture);
+    target->intermediate_textures.push_back(texture);
   }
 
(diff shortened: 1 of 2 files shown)


lewri:
This does resolve the issue at least on the available test machine.

TheCycoONE:
LGTM. Thanks!

Issue #2512, opened 2024-03-02
[Bug] True 4k resolution prevents game launch


lewri:
### Describe the issue

Should be a known error but may bear some exposure now. If someone plays fullscreen on a true 4k screen (4096 x 2160) the game crashes. This would be because it goes beyond I assume the map bounds.

You can also see it at 3840x2160. The map loads but a section of the edge of the map becomes visible. If you move or zoom the game forcefully resnaps back to the limited view.

### Steps to Reproduce

1. Set resolution to fullscreen and 4096x2160
2. Attempt to load a game or new game

### Expected Behaviour

It's probably advisable that we keep the resolution limitation right now but the game needs to error nicer. Currently it just errors and locks everything up (I think it comes after the `pcall`). Instead the game should exit back to the main menu nicely with a `UIInformation` box stating the resolution is set too high. Alternatively, this could also be done more aggressively in the change resolution setting if someone tries to alter beyond supported values.

### Save Game

Save game attached to report: No

### CorsixTH Version

0.67`

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

```bash
A stack trace is included below, and the handler has been disconnected.
Lua\game_ui.lua:75: Screen size too large for the map
stack traceback:
        [C]: in function 'error'
        Lua\game_ui.lua:75: in local 'constructor'
        Lua\class.lua:74: in global 'GameUI'
        Lua\app.lua:713: in method 'loadLevel'
        Lua\dialogs/resizables/menu_list_dialogs\custom_game.lua:136: in local 'callback'
        Lua\window.lua:699: in method 'handleClick'
        Lua\window.lua:1658: in method 'onMouseUp'
        Lua\window.lua:1642: in field 'onMouseUp'
        Lua\ui.lua:877: in function <Lua\ui.lua:864>
        (...tail calls...)
        Lua\app.lua:1102: in function <Lua\app.lua:1097>
```

TheCycoONE:
We're getting to the point where we'll be forced to scale everything 2x.

lewri:
Crash now prevented by #2513 , until scaling changes are considered/implemented.

lewri:
Merged itno #1210

PR #2513, opened 2024-03-03
Protected Call all Level Loads

lewri:
Brought on by the discovery of #2512 

- Helps prevent a program crash where an error occurs during load.

This patch currently loses the `debug.stacktrace` call though. If we need that we'll need to use `xpcall` and some tricks to work around Lua 5.1. As a result, this currently only reports the actual line that failed.

NB: This affects all calls of  `App:loadLevel` but not the more direct `App:load` for existing saves


lewri:
We do a lot of localised `pcalls` I've missed here actually. I'll WIP this

lewri:
The crashes from preventing crashes are fixed now 😅 .

Can discuss if we want xpcall now or not for a `debug.stacktrace` or not -- some of the old code already pcalled so the error was lost anyway.

CorsixTH/Lua/dialogs/fullscreen/fax.lua
```diff
@@ -207,16 +207,17 @@ function UIFax:choice(choice_number)
     self.ui.hospital.player_salary = self.ui.hospital.salary_offer
     if tonumber(self.ui.app.world.map.level_number) then
       local next_level = self.ui.app.world.map.level_number + 1
-      self.ui.app:loadLevel(next_level, self.ui.app.map.difficulty)
-      self.ui.app.moviePlayer:playAdvanceMovie(next_level)
+      if self.ui.app:loadLevel(_S.errors.load_level_prefix, next_level, self.ui.app.map.difficulty) then
```
lewri, line 210: An open fax on save doesn't crash here 👍

CorsixTH/Lua/app.lua
```diff
@@ -671,10 +672,28 @@ function App:getAbsolutePathToLevelFile(level)
   return nil, "Level not found: " .. level
 end
 
--- Loads the specified level. If a string is passed it looks for the file with the same name
+--! Invokes a protected call of App:_loadLevel(...). See that function for more information.
+--! This function should always be called to catch errors and properly pass the
+--! error to the player
+--!param error_prefix (string) Prefixes the error relevant to what was loaded
+--! return (boolean) The outcome of the pcall
+function App:loadLevel(error_prefix, level, difficulty, level_name, level_file, level_intro, map_editor)
```
TheCycoONE, line 680: error_prefix should not be the first argument as it's an optional and not central to the function purpose.

lewri: I agree, I was being lazy.

lewri: Moved to the end of the function params, but this may have been just as bad.

CorsixTH/Lua/app.lua
```diff
@@ -1703,18 +1722,13 @@ function App:restart()
       self.ui:addWindow(UIInformation(self.ui, { _S.information.cannot_restart }))
       return
     end
-    local status, err = pcall(self.loadLevel, self, level, difficulty, name, file, intro)
-    if not status then
-      err = "Error while loading level: " .. err
-      print(err)
-      self.ui:addWindow(UIInformation(self.ui, { err }))
-    end
+    self:loadLevel(_S.errors.load_level_prefix, level, difficulty, name, file, intro)
```
lewri, line 1725: I missed one. I'll just push a quick pr to fix it in a bit

PR #2514, opened 2024-03-05
Fix pink briefing in old saves

### LINKED ISSUE #2308: UIInformation box is loading pink

TheCycoONE:
Fixes #2308


lewri (approved):
Copied change into my testing environment. This works.

PR #2516, opened 2024-03-05
Undo cheat function rename

### LINKED ISSUE #2515: New cheats menu PR has broken persistence
PR: #1997

The changes in this PR have broken persistence for existing saves of all calibres. I'm guessing that we might need an afterLoad somewhere that got missed.

image

tobylane:
*Fixes #2515*

- Revert the function name changes of [d81b32](https://github.com/CorsixTH/CorsixTH/pull/1997/commits/d81b32c69b4a6332ee071398674d6445eb9b1a37).

### DIFF
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -1000,3 +1000,10 @@ fax = {
     }
   }
 }
+
+cheats_window.cheats = {
+ toggle_infected = show_infected,
+}
+tooltip.cheats_window.cheats = {
+ toggle_infected = show_infected,
+}
--- a/CorsixTH/Lua/cheats.lua
+++ b/CorsixTH/Lua/cheats.lua
@@ -105,6 +105,8 @@ function Cheats:cheatEmergency()
   end
 end
 
+function Cheats:cheatToggleInfected() end -- Stub of the old name of the function below
+
 --[[ Before an epidemic has been revealed, show/hide the infected mood icons
 to easily distinguish the infected patients. This will show/hide icons
 for ALL future epidemics you cannot distinguish between epidemics


lewri:
I think cyco wants to have a stab at this first to see if there's alternative ways round this.
Though generally it's probably better practice to keep your function name if it fits and make the old one a stub if it has to stay.

TheCycoONE:
Yes, I suspect at the very least there's a transitional approach here where you stub in the old function calling the new function but only reference the new function in new code. Eventually all stored references to the old function go away. The problem is that table - which we probably shouldn't even be persisting. We can probably fix it so that it's not persisted going forward.

CorsixTH/Lua/languages/english.lua
```diff
@@ -828,6 +828,7 @@ cheats_window = {
     toggle_epidemic = "Toggle epidemics",
     epidemic = "Spawn contagious patient",
     show_infected = "Show or hide infected icons",
+    toggle_infected = "Show or hide infected icons",
```
lewri, line 831: Move this and 857 to the unused section. You might want to follow the comment block in 953-956 but I'll leave that up to you.
Do the strings actually need to exist to not crash? Only instance I could think where they might crash is during the afterload sequence if the `UICheats` window was open.

tobylane: I think they are needed in the moment before the window is closed.

> You might want to follow the comment block in 953-956

Good idea, done.

TheCycoONE:
Does the cheat list get refreshed on a new save (save, modify the table, load)

lewri:
> Does the cheat list get refreshed on a new save (save, modify the table, load)

`PlayerHospital` does on every load. I'm not sure if it's wise to leave `AIHospital` unmodified. If `cheats_list` is moved to a local then you probably don't also then need the `afterLoad` in future.
https://github.com/CorsixTH/CorsixTH/blob/537c2e13e26e46107af4f978b2f2082df57c3b65/CorsixTH/Lua/hospitals/player_hospital.lua#L777-L802

CorsixTH/Lua/languages/english.lua
```diff
@@ -1000,3 +1000,10 @@ fax = {
     }
   }
 }
+
+cheats_window.cheats = {
```
lewri, line 1004:
```suggestion
-- To be removed in 2026 release
cheats_window.cheats = {
```
(might be 2027)

PR #2518, opened 2024-03-06
Encourage level restarts on very_old_saves

lewri:
Proposal to give the option to the user to restart their level for very old saves compared to just information.
This method might be too assertive on the player so it can be dropped if it's going to be too far.

- Changes this UIInformation to a UIConfirmDIalog for old saves.
- Adapts `App:restart()` to take an optional message for the box

image

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1755,9 +1755,7 @@ function App:getReleaseString(savegame_version)
   local release = self:getReleaseData(savegame_version)
   local release_string = "v" .. release.major .. "." .. release.minor .. "." ..
       release.revision
-  if string.len(release.patch) > 0 then
-    release_string = release_string .. release.patch
-  end
+  release_string = release_string .. release.patch
   return release_string
 end
 
@@ -1773,9 +1771,10 @@ end
 --!return The step difference between release a and release b for release method
 --- or The raw savegame version difference for version method
 function App:compareVersions(version_a, version_b, method)
-  assert((type(version_a) == "table" or type(version_a) == "number") and
-      (type(version_b) == "table" or type(version_b) == "number"),
-      "Compare requires savegame version or the version table to function")
+  assert(type(version_a) == "table" or type(version_a) == "number",
+      "version_a requires savegame version or an entry from the version table to compare")
+  assert(type(version_b) == "table" or type(version_b) == "number",
+      "version_b requires savegame version or an entry from the version table to compare")
   assert(method == "release" or method == "version",
       "Not using a valid compare method")
 
@@ -1791,8 +1790,9 @@ function App:compareVersions(version_a, version_b, method)
           if step == 0 then step = 1 break end -- working from current development
           break
         end
-        step = (release.revision == 0 and string.len(release.patch) == 0)
-            and step - 1 or step
+        if release.revision == 0 and release.patch == "" then
+          step = step - 1
+        end
       end
       return step
     end
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -96,7 +96,7 @@ function App:init()
   print("")
   print("---------------------------------------------------------------")
   print("")
-  print("Welcome to CorsixTH " .. self:getVersion(self.savegame_version, true) .. "!")
+  print("Welcome to CorsixTH " .. self:getReleaseString() .. "!")
   print("")
   print("---------------------------------------------------------------")
   print("")
@@ -1693,7 +1693,7 @@ end
   All beta versions must have a savegame increment each time
   Each patch note must begin with a '-'
 --]]
-local version_table = {
+local release_table = {
   -- Format: major, minor, revision, patch (string), savegame_version
   {major = 0, minor = 0, revision = 8, patch = "", version = 0}, -- Beta 8 or below
   {major = 0, minor = 1, revision = 0, patch = "", version = 51},
@@ -1724,44 +1724,54 @@ end
 
 --! Requests data regarding a given savegame version
 --!param savegame_version (number) What to lookup, uses application version if blank
---!param concatenate (boolean) (optional) Whether the version should be converted to a string
---!return the matching version as a table, or as a string if concatenate is true.
--- If no versions match it returns the base version with the savegame version
--- attached e.g. 0.66.0-dev172
-function App:getVersion(savegame_version, concatenate)
+--!return the matching release table entry
+-- If no releases match it returns the base release with the savegame version
+-- inserted
+function App:getReleaseData(savegame_version)
   savegame_version = savegame_version or self:getCurrentVersion()
-  local result
-  for i = #version_table, 1, -1 do
-    local version = version_table[i]
-    if version.version == savegame_version then
-      result = version
+  local release_data
+  for i = #release_table, 1, -1 do
+    local release = release_table[i]
+    if release.version == savegame_version then
+      release_data = release
       break
-    elseif version.version - savegame_version < 0 then
+    elseif (release.version - savegame_version) < 0 then
       -- we're not on a release version
-      local devel_version = shallow_clone(version) -- prevent recursion
-      devel_version.patch = version.patch .. "-dev" .. savegame_version
-      result = devel_version
+      local develop = shallow_clone(release) -- prevent recursion
+      develop.version = savegame_version
+      develop.patch = release.patch .. "-dev" .. savegame_version
+      release_data = develop
       break
     end
   end
-  if concatenate and result then
-    if string.len(result.patch) > 0 then
-      result = "v" .. result.major .. "." .. result.minor .. "." ..
-          result.revision .. result.patch
-    else
-      result = "v" .. result.major .. "." .. result.minor .. "." .. result.revision
-    end
+  return release_data
+end
+
+--! Provides the release (or development) string for a given savegame version
+--!param savegame_version (number) The version to look up
+--!return A string in the format "v<major>.<minor>.<revision>[patch]"
+--! For development builds, the savegame version is appended as a patch e.g. 'dev213'
+function App:getReleaseString(savegame_version)
+  local release = self:getReleaseData(savegame_version)
+  local release_string = "v" .. release.major .. "." .. release.minor .. "." ..
+      release.revision
+  if string.len(release.patch) > 0 then
+    release_string = release_string .. release.patch
   end
-  return result
+  return release_string
 end
 
 --! Reports a difference between two versions based on requested methodology.
 --!param version_a (number or table) The first (usually newer) version to test
 --!param version_b (number or table) The second (usually older) version to test
 --!param method (string) What method to compare by
---- method(release) reports the difference between the matching releases in steps, revisions are not counted.
+--- method(release) reports the difference between the matching releases in steps,
+--- revisions are not counted.
+--- For development builds it will use the base release it started from when using
+--- the release method
 --- method(version) reports the difference between two savegame versions
---! if a match is detected as a development version it will target the release it was built upon, except for via version method.
+--!return The step difference between release a and release b for release method
+--- or The raw savegame version difference for version method
 function App:compareVersions(version_a, version_b, method)
   assert((type(version_a) == "table" or type(version_a) == "number") and
       (type(version_b) == "table" or type(version_b) == "number"),
@@ -1772,24 +1782,23 @@ function App:compareVersions(version_a, version_b, method)
   if method == "release" then
     local function countBackward(version_to_check)
       local step = 0
-      for i = #version_table, 1, -1 do
-        local version = version_table[i]
-        if version.version == version_to_check then
+      for i = #release_table, 1, -1 do
+        local release = release_table[i]
+        if release.version == version_to_check then
           break
-        elseif (version.version - version_to_check) < 0 then
+        elseif (release.version - version_to_check) < 0 then
           -- we're not on a release version
-          if step == 0 then break end -- working from current development
-          step = step - 1
+          if step == 0 then step = 1 break end -- working from current development
           break
         end
-        step = (version.revision == 0 and string.len(version.patch) == 0)
+        step = (release.revision == 0 and string.len(release.patch) == 0)
             and step - 1 or step
       end
       return step
     end
 
-    if type(version_a) == "number" then version_a = self:getVersion(version_a) end
-    if type(version_b) == "number" then version_b = self:getVersion(version_b) end
+    if type(version_a) == "number" then version_a = self:getReleaseData(version_a) end
+    if type(version_b) == "number" then version_b = self:getReleaseData(version_b) end
     return countBackward(version_a.version) - countBackward(version_b.version)
   end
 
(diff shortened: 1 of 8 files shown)


lewri:
If we wanted to go ahead then I can ask our translators for strings if needed as this is an important message.

TheCycoONE:
Good idea. I would leave off the first "this level"

A few comments:
new-20 is a bit arbitrary, sometimes we'd eat through more than that in a single version, maybe we can do something with the release table we're maintaining in TheApp (currently a giant if condition).

I'm not sure this shouldn't be built into the existing TheApp.checkCompatibilty handling (see that function and peristence.lua where it's called)

lewri:
> Good idea. I would leave off the first "this level"
> 
> A few comments: new-20 is a bit arbitrary, sometimes we'd eat through more than that in a single version, maybe we can do something with the release table we're maintaining in TheApp (currently a giant if condition).
> 
> I'm not sure this shouldn't be built into the existing TheApp.checkCompatibilty handling (see that function and peristence.lua where it's called)

Pushed the string change for now.
In regards to version control I've started #2527 as my planned changes will probably deserve their own PR separate to this.

lewri:
This won't be ready in time, but the `- 20` check is definitely going to show up for versions coming from `0.66`. I have more sophisticated checks in the works so up to us if to temporarily bump the check more or not.

CorsixTH/Lua/app.lua
```diff
@@ -1686,9 +1686,10 @@ function App:checkCompatibility(save_version, gfx_set)
 end
 
 --! Restarts the current level (offers confirmation window first)
-function App:restart()
+--!param message (string) Optional message to the player
+function App:restart(message)
   assert(self.map, "Trying to restart while no map is loaded.")
-  self.ui:addWindow(UIConfirmDialog(self.ui, false, _S.confirmation.restart_level,
+  self.ui:addWindow(UIConfirmDialog(self.ui, true, message or _S.confirmation.restart_level,
```
lewri, line 1692: Ideally we should still try to be persisting campaign data if it exists

lewri: Fixed in a later PR already in master

lewri:
I still need to add the Spanish and French translations I was given

CorsixTH/Lua/app.lua
```diff
@@ -1680,57 +1680,131 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+  -- There was also 0.65.1, not differentiated by version number
+  {major = 0, minor = 66, revision = 0, patch = "", date = {2022, 06, 26}, version = 170},
+  {major = 0, minor = 67, revision = 0, patch = "", date = {2023, 08, 12}, version = 180},
+  {major = 0, minor = 68, revision = 0, patch = "", date = {2024, 10, 13}, version = 194},
+  {major = 0, minor = 69, revision = 0, patch = "-beta-1", date = {2025, 06, 07}, version = 216}
+}
+
+function App:getCurrentVersion()
+  return self.savegame_version
+end
```
lewri, line 1723: This is used a total of 1 time and might in general be useless to have, unless we'd rather start using this instead of a direct call to the variable.

Also, needs a doc line as it behaves differently to `App:getVersion`

CorsixTH/Lua/app.lua
```diff
@@ -1680,57 +1680,131 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+        end
+        step = (version.revision == 0 and string.len(version.patch) == 0)
+            and step - 1 or step
+      end
+      return step
+    end
+    return countBackward(version_a.version) - countBackward(version_b.version)
+  end
+
+  if method == "version" then
```
lewri, line 1805: This method currently checks against the returned table entries, which might not actually match what version you were searching for. This needs fixing

CorsixTH/Lua/app.lua
```diff
@@ -1680,57 +1680,131 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+    return version_a.version - version_b.version
+  end
+
+  if method == "age" then
```
lewri: Like version, this compare will only look at the table entries, it doesn't consider a development build any different in release date to the actual base release version. This could probably be scrapped, along with the date entries in the table (unless they're nice to have for some other purpose).

CorsixTH/Lua/app.lua
```diff
@@ -1680,57 +1680,131 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+--! Requests data regarding a given savegame version
+--!param savegame_version (number) What to lookup, uses application version if blank
+--!param concatenate (boolean) (optional) Whether the version should be converted to a string
+--!return the matching version as a table, or as a string if concatenate is true.
+-- If no versions match it returns the base version with the savegame version
+-- attached e.g. 0.66.0-dev172
+function App:getVersion(savegame_version, concatenate)
+  savegame_version = savegame_version or self:getCurrentVersion()
+  local result
+  for i=#version_table, 1, -1 do
```
Alberth289346:
```suggestion
  for i = #version_table, 1, -1 do
```

CorsixTH/Lua/app.lua
```diff
@@ -1680,57 +1680,131 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+    local version = version_table[i]
```
Alberth289346: Variable name is confusing. Likely, there are more such names.

It may make sense to reserve "version" for something numeric, ie `216` like the savegame version. Stuff like `1.2.3-something` looks like a release_number to me (release_name feels could work too, but not sure).

The "version_table" is not a list of versions, it has much more data. Perhaps `release[s]_table`?
If you still agree at this point, then the `version` here would be a `release` I guess.

lewri: While on the subject of this, rename function maybe to `getReleaseData`?

CorsixTH/Lua/app.lua
```diff
@@ -1680,57 +1680,131 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+    version_a = self:getVersion(version_a)
+  end
+  if type(version_b) == "number" then
+    version_b = self:getVersion(version_b)
+  end
+
+  if method == "release" then
+    local function countBackward(version_to_check)
+      local step = 0
+      for i=#version_table, 1, -1 do
```
Alberth289346:
```suggestion
      for i = #version_table, 1, -1 do
```

CorsixTH/Lua/app.lua
```diff
@@ -1680,57 +1680,131 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+        local version = version_table[i]
+        if version.version == version_to_check then
+          break
+        elseif (version.version - version_to_check) < 0 then
```
Alberth289346: This line has brackets, line 1738 doesn't (brackets are not doing anything semantically, either line is good).

CorsixTH/Lua/app.lua
Alberth289346: It's somewhat a waste to delete information, even if not used I'd keep it.
Moving it to comment is an option, although it's harmless to have it in the table too.

CorsixTH/Lua/app.lua
Alberth289346: But it also produces a string every now and then.

Perhaps also make `getReleaseString`, which calls this function and then does the "concatenate" trick by itself?
The concatenate part has no influence on getting the data it seems, so concatenation is just a post-processing step.

Alberth289346: Getting one kind of result from a function is helpful in the sense that you can express more precisely what it does in the function name, eg `getReleaseData()` versus `getReleaseString()` or so.
That typically beats magic `true`/`false` parameters, since they don't say what they mean at the calling side.

CorsixTH/Lua/app.lua
```diff
@@ -1680,60 +1680,127 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+-- attached e.g. 0.66.0-dev172
+function App:getVersion(savegame_version, concatenate)
+  savegame_version = savegame_version or self:getCurrentVersion()
+  local result
+  for i = #version_table, 1, -1 do
+    local version = version_table[i]
+    if version.version == savegame_version then
+      result = version
+      break
+    elseif version.version - savegame_version < 0 then
```
lewri: Add brackets

CorsixTH/Lua/app.lua
```diff
@@ -1680,60 +1680,127 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+      -- we're not on a release version
+      local devel_version = shallow_clone(version) -- prevent recursion
+      devel_version.patch = version.patch .. "-dev" .. savegame_version
+      result = devel_version
```
lewri: The result is slightly wrong. It would be better to also overwrite `devel_version.version` with the requested version number.

CorsixTH/Lua/app.lua
```diff
@@ -1680,60 +1680,136 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+  if method == "release" then
+    local function countBackward(version_to_check)
+      local step = 0
+      for i = #release_table, 1, -1 do
+        local release = release_table[i]
+        if release.version == version_to_check then
+          break
+        elseif (release.version - version_to_check) < 0 then
+          -- we're not on a release version
+          if step == 0 then step = 1 break end -- working from current development
```
lewri, line 1790: Small caveat -- this will prompt a restart level for e.g. 0.67.0 saves as soon as you start developing ontop of 0.69.0

lewri:
This should be ready for final review now. I can put the date values of the table back if we really want them.

CorsixTH/Lua/app.lua
```diff
@@ -1680,60 +1680,136 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+  for i = #release_table, 1, -1 do
+    local release = release_table[i]
+    if release.version == savegame_version then
+      release_data = release
+      break
+    elseif (release.version - savegame_version) < 0 then
+      -- we're not on a release version
+      local develop = shallow_clone(release) -- prevent recursion
+      develop.version = savegame_version
+      develop.patch = release.patch .. "-dev" .. savegame_version
```
TheCycoONE, line 1742: I don't think we should keep the release.patch here, just set it to -dev? or -dev(savegame_version)

CorsixTH/Lua/app.lua
```diff
@@ -1680,60 +1680,136 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+      for i = #release_table, 1, -1 do
+        local release = release_table[i]
+        if release.version == version_to_check then
+          break
+        elseif (release.version - version_to_check) < 0 then
+          -- we're not on a release version
+          if step == 0 then step = 1 break end -- working from current development
+          break
+        end
+        step = (release.revision == 0 and string.len(release.patch) == 0)
```
TheCycoONE: Returning a negative number is weird.

TheCycoONE (approved):

CorsixTH/Lua/app.lua
```diff
@@ -1680,60 +1680,136 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+--!param savegame_version (number) The version to look up
+--!return A string in the format "v<major>.<minor>.<revision>[patch]"
+--! For development builds, the savegame version is appended as a patch e.g. 'dev213'
+function App:getReleaseString(savegame_version)
+  local release = self:getReleaseData(savegame_version)
+  local release_string = "v" .. release.major .. "." .. release.minor .. "." ..
+      release.revision
+  if string.len(release.patch) > 0 then
+    release_string = release_string .. release.patch
+  end
```
Alberth289346: If the patch-part is empty, you can also concatenate it. That avoids the special case.

CorsixTH/Lua/app.lua
```diff
@@ -1680,60 +1680,136 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+--- revisions are not counted.
+--- For development builds it will use the base release it started from when using
+--- the release method
+--- method(version) reports the difference between two savegame versions
+--!return The step difference between release a and release b for release method
+--- or The raw savegame version difference for version method
+function App:compareVersions(version_a, version_b, method)
+  assert((type(version_a) == "table" or type(version_a) == "number") and
+      (type(version_b) == "table" or type(version_b) == "number"),
+      "Compare requires savegame version or the version table to function")
```
Alberth289346:
```suggestion
      "version_b requires savegame version or an entry from the version table to compare")
```
Likely you don't want the entire table as parameter :)

If you split the check (one check for parameter `version_a` and one check for parameter `version_b`), it's simpler to understand what is wrong precisely.
Alternatively, add the values to the error string, so we can read what's broken.

CorsixTH/Lua/app.lua
```diff
@@ -1680,60 +1680,136 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+        local release = release_table[i]
+        if release.version == version_to_check then
+          break
+        elseif (release.version - version_to_check) < 0 then
+          -- we're not on a release version
+          if step == 0 then step = 1 break end -- working from current development
+          break
+        end
+        step = (release.revision == 0 and string.len(release.patch) == 0)
+            and step - 1 or step
```
TheCycoONE:
```suggestion
        if release.revision == 0 and release.patch == "" then
          step = step - 1
        end
```

TheCycoONE: The equality check is cheaper and more straight forward, I think the if condition is clearer here.

CorsixTH/Lua/app.lua
lewri: As long as we pertain "" yes

```suggestion
  release_string = release_string .. release.patch
```

CorsixTH/Lua/app.lua
```diff
@@ -1680,60 +1680,136 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+--- method(release) reports the difference between the matching releases in steps,
+--- revisions are not counted.
+--- For development builds it will use the base release it started from when using
+--- the release method
+--- method(version) reports the difference between two savegame versions
+--!return The step difference between release a and release b for release method
+--- or The raw savegame version difference for version method
+function App:compareVersions(version_a, version_b, method)
+  assert((type(version_a) == "table" or type(version_a) == "number") and
+      (type(version_b) == "table" or type(version_b) == "number"),
```
lewri:
```suggestion
  assert(type(version_a) == "table" or type(version_a) == "number",
      "version_a requires savegame version or an entry from the version table to compare")
  assert(type(version_b) == "table" or type(version_b) == "number",
```

CorsixTH/Lua/app.lua
lewri: Whole assert edited above, had to amend your suggestion just so it preserved your co-author

CorsixTH/Lua/app.lua
lewri: If you start preserving betas then you probably do want it,

CorsixTH/Lua/app.lua
```diff
@@ -1680,60 +1680,136 @@ function App:loadLuaFolder(dir, no_results, append_to)
...
+  {major = 0, minor = 61, revision = 0, patch = "", version = 122},
+  {major = 0, minor = 62, revision = 0, patch = "", version = 127},
+  {major = 0, minor = 63, revision = 0, patch = "", version = 134},
+  {major = 0, minor = 64, revision = 0, patch = "", version = 138},
+  {major = 0, minor = 65, revision = 0, patch = "", version = 156},
+  -- There was also 0.65.1, not differentiated by version number
+  {major = 0, minor = 66, revision = 0, patch = "", version = 170},
+  {major = 0, minor = 67, revision = 0, patch = "", version = 180},
+  {major = 0, minor = 68, revision = 0, patch = "", version = 194},
+  {major = 0, minor = 69, revision = 0, patch = "-beta-1", version = 216}
```
lewri, line 1717: If we want to reduce the ever growing list here filling up `app.lua`, if there's a good way to stick this table in its own file and then assign it back to release_table I'm open to suggestions.
`release_history.lua`?

TheCycoONE: sure but we can do that any time

PR #2528, opened 2024-03-14
Refactor staff happiness calculation

tobylane:
- The calculation for happiness from windows and space was done per staff per day, now it is done on room finish and once after load.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -29,7 +29,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 185 -- Fix missing staff initial
+local SAVEGAME_VERSION = 186 -- Refactor staff happiness calculation
 
 class "App"
 
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -102,20 +102,7 @@ function Staff:tickDay()
 
   local room = self:getRoom()
   if room then
-    -- It always makes you happy to see the outdoors (or windows to anywhere)
-    local count = room:countWindows()
-    if room.room_info.id == "staff_room" then -- Pleased another bit
-      count = count * 2
-    end
-    if count > 0 then
-      -- More windows help but in smaller increments
-      self:changeAttribute("happiness", math.round(math.log(count)) / 1000)
-    end
-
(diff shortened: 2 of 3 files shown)


CorsixTH/Lua/app.lua
```diff
@@ -29,7 +29,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 183 -- Adjust game speeds
+local SAVEGAME_VERSION = 185 -- Refactor staff happiness calculation
```
lewri, line 32: Should this be 184?

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -101,22 +101,9 @@ function Staff:tickDay()
...
-    end
-
-    -- Extra space in the room you are in adds to your happiness
-    local extraspace = (room.width * room.height) / (room.room_info.minimum_size * room.room_info.minimum_size)
-    -- Greater space helps but in smaller increments
-    self:changeAttribute("happiness", math.round(math.log(extraspace)) / 1000)
+  local happiness_factor = self:getRoom() and self:getRoom().happiness_factor
+  if happiness_factor and happiness_factor > 0 then
+    self:changeAttribute("happiness", happiness_factor)
   end
```
lewri, line 107: Why would we need to check happiness is above 0? It would be more open if the possibility for a room to make a staff member unhappy remained.

tobylane: Interesting idea! Fixed both.

CorsixTH/Lua/room.lua
```diff
@@ -1028,6 +1029,9 @@ function Room:afterLoad(old, new)
     -- no longer using this so empty it
     self.humanoids_enroute = {}
   end
+  if old and old < 185 then
```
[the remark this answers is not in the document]
tobylane: https://github.com/CorsixTH/CorsixTH/commit/07892845f6928898442c3d8fb1dd3b5c923f5416

I don't see a reason why it was added, and this new afterload works with this format removed, so I've taken it out of the function.

lewri:
Bump savegame version and I will merge 👍

CorsixTH/Lua/room.lua
```diff
@@ -1028,6 +1029,9 @@ function Room:afterLoad(old, new)
     -- no longer using this so empty it
     self.humanoids_enroute = {}
   end
+  if old < 185 then
```
lewri, line 1032: Need to bump this one too 😅

Issue #1924, opened 2021-06-12
[Bug] Attempt to getServiceQuality of staff with nil fatigue


TheCycoONE:
#### Describe the issue
Error confirmation dialog popped up while navigating through staff list. See gamelog output.

#### Steps to Reproduce
1. Load the before save (or start a new game)
2. When a patient is in the inflation room, fire the doctor treating the patient
3. When patient is treated, game errors

#### Save Game
After the issue occurred (doctor's appear to be stuck):
[save.zip](https://github.com/CorsixTH/CorsixTH/files/6642550/save.zip)

#### System Information
**CorsixTH Version:** *3a0cd420* (appoximately 0.65 + zoom patch)

**Operating System:** *Arch Linux*

**Theme Hospital Version:** *CD*

#### Gamelog.txt
```
Created game with savegame version 155.
Error in callback handler: 
/home/aimee/CorsixTH/CorsixTH/Lua/audio.lua:577: Could not load music file 'Sound/Midi/DOCTOR.XMI' (Couldn't open /etc/timidity++/freepats.cfg)
stack traceback:
	[C]: in function 'coroutine.yield'
	/home/aimee/CorsixTH/CorsixTH/Lua/app.lua:1031: in function </home/aimee/CorsixTH/CorsixTH/Lua/app.lua:1026>
Warning: Trying to remove nonexistent room build callback (function: 0x55c4d4c0e820) from humanoid (table: 0x55c4d50dfd50).
Warning: Trying to remove nonexistent room remove callback (function: 0x55c4d428edf0) from humanoid (table: 0x55c4d50dfd50).
Warning: Trying to remove nonexistent staff callback (function: 0x55c4d4b9acf0) from humanoid (table: 0x55c4d50dfd50).
Warning: Trying to remove nonexistent room build callback (function: 0x55c4d40f2cd0) from humanoid (table: 0x55c4d26b18f0).
Warning: Trying to remove nonexistent room remove callback (function: 0x55c4d49ac480) from humanoid (table: 0x55c4d26b18f0).
Warning: Trying to remove nonexistent staff callback (function: 0x55c4d4f619b0) from humanoid (table: 0x55c4d26b18f0).
Warning: Trying to remove nonexistent room build callback (function: 0x55c4d3d712f0) from humanoid (table: 0x55c4d4f70c90).
Warning: Trying to remove nonexistent room remove callback (function: 0x55c4d3356a20) from humanoid (table: 0x55c4d4f70c90).
[...]

PR #2529, opened 2024-03-15
Add a function to get humanoid attribute values

tobylane:
- ~~Staff fatigue can be nil, [see mugmuggy's comment](https://github.com/CorsixTH/CorsixTH/issues/1924#issuecomment-860139913). Check the staff member has fatigue before accessing it, or use zero in its place.~~
- Add a function to get a humanoid's attribute, or the given default value, or 0.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -43,29 +43,28 @@ function Staff:tickDay()
   if not Humanoid.tickDay(self) then
     return false
   end
+  if self.humanoid_class == "Receptionist" then return true end
   -- Pay too low  --> unhappy
   -- Pay too high -->   happy
   local fair_wage = self.profile:getFairWage()
   local wage = self.profile.wage
   self:changeAttribute("happiness", 0.05 * (wage - fair_wage) / (fair_wage ~= 0 and fair_wage or 1))
+
   -- if you overwork your Dr's then there is a chance that they can go crazy
   -- when this happens, find him and get him to rest straight away
-  if self.attributes["fatigue"] then
-    if self.attributes["fatigue"] < 0.7 then
-      if self:isResting() then
-        self:setMood("tired", "deactivate")
-        self:changeAttribute("happiness", 0.006)
-      end
-    end
-  -- Working when you should be taking a break will make you unhappy
-    if self.attributes["fatigue"] >= self.hospital.policies["goto_staffroom"] then
-      self:changeAttribute("happiness", -0.02)
-    end
-  -- You will also start to become unhappy as you become tired
-    if self.attributes["fatigue"] >= 0.5 then
-      self:changeAttribute("happiness", -0.01)
-    end
+  if self:getAttribute("fatigue") < 0.7 and self:isResting() then
+    self:setMood("tired", "deactivate")
+    self:changeAttribute("happiness", 0.006)
   end
+  -- Working when you should be taking a break will make you unhappy
+  if self:getAttribute("fatigue") >= self.hospital.policies["goto_staffroom"] then
+    self:changeAttribute("happiness", -0.02)
+  end
+  -- You will also start to become unhappy as you become tired
+  if self:getAttribute("fatigue") >= 0.5 then
+    self:changeAttribute("happiness", -0.01)
+  end
+
   -- It is nice to see plants, but dead plants make you unhappy
   self.world:findObjectNear(self, "plant", 2, function(x, y)
     local plant = self.world:getObject(x, y, "plant")
@@ -141,7 +140,7 @@ function Staff:tick()
   end
 
   -- Make staff members request a raise if they are very unhappy
-  if not self.world.debug_disable_salary_raise and self.attributes["happiness"] < 0.1 then
+  if not self.world.debug_disable_salary_raise and self:getAttribute("happiness") < 0.1 then
     if not self.timer_until_raise then
       self.timer_until_raise = 200
     end
@@ -358,10 +357,10 @@ function Staff:updateSpeed()
     level = 1
   elseif self.hospital and self.hospital.hosp_cheats:isCheatActive("no_rest_cheat") then
     level = 3 -- Cheat makes everyone speedy
-  elseif self.attributes["fatigue"] then
-    if self.attributes["fatigue"] >= 0.8 then
+  elseif self.humanoid_class ~= "Receptionist" then
+    if self:getAttribute("fatigue") >= 0.8 then
       level = level - 2
-    elseif self.attributes["fatigue"] >= 0.7 then
+    elseif self:getAttribute("fatigue") >= 0.7 then
       level = level - 1
     end
   end
@@ -381,12 +380,12 @@ end
 -- and go to the StaffRoom if it is.
 function Staff:checkIfNeedRest()
   -- Only when the staff member is very tired should the icon emerge. Unhappiness will also escalate
-  if self.attributes["fatigue"] >= 0.7 then
+  if self:getAttribute("fatigue") >= 0.7 then
     self:setMood("tired", "activate")
     self:changeAttribute("happiness", -0.0002)
   end
   -- If above the policy threshold, go to the staff room.
-  if self.attributes["fatigue"] >= self.hospital.policies["goto_staffroom"] and
+  if self:getAttribute("fatigue") >= self.hospital.policies["goto_staffroom"] and
       not class.is(self:getRoom(), StaffRoom) then
     -- The staff will get unhappy if there is no staffroom to rest in.
     if self.waiting_for_staffroom then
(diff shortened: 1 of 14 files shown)


lewri (commented):
We're mainly protecting against fired staff right in flight? Do these instances make logical sense to protect (i.e. should this check never reach this stage on a fired staff member)?

tobylane:
Some of them don't make sense, but it's protection from staff completing their actions before they leave.

CorsixTH/Lua/humanoid_actions/use_staffroom.lua
```diff
@@ -114,7 +114,7 @@ local function use_staffroom_action_start(action, humanoid)
   local loop_callback_use = --[[persistable:use_staffroom_action_loop_callback]] function(obj_action)
     humanoid:wake(relaxation[action.target_type])
     -- if staff is no longer fatigued, make them leave the staff room
-    if humanoid.attributes["fatigue"] == 0 then
+    if humanoid.attributes.fatigue and humanoid.attributes.fatigue == 0 then
```
TheCycoONE, line 117: Why change this line?

tobylane: To check fatigue exists before comparing it with 0.

TheCycoONE: The result is the same, nil ~= 0

tobylane: Reverted

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -626,7 +627,7 @@ function Staff:updateDynamicInfo()
     dynamic_text,
     fatigue_text,
   })
-  self:setDynamicInfo('progress', self.attributes["fatigue"])
+  self:setDynamicInfo('progress', self.attributes.fatigue or 0)
```
lewri, line 630: This is not suitable as it creates a tiredness bar on a receptionist
image

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -380,6 +380,7 @@ end
 -- Check if fatigue is over a certain level (decided by the hospital policy),
 -- and go to the StaffRoom if it is.
 function Staff:checkIfNeedRest()
+  if not self.attributes.fatigue then return end
```
lewri, line 383: Have you sufficiently checked a fired staff member can still leave the room if their tiredness was over the need to rest threshold?

lewri (changes requested):

CorsixTH/Lua/entities/humanoids/staff.lua
lewri: (You can likely either add a case to ignore receptionists, or the old check may be fine as is. See `Entity:setDynamicInfo`)

CorsixTH/Lua/entities/humanoids/staff.lua
tobylane: Yes, they're on their way out of the world either way.

CorsixTH/Lua/entities/humanoids/staff.lua
tobylane: Reverted this too.

lewri (approved):

TheCycoONE:
Does anything depend on fatigue being nil?

tobylane:
The receptionist doesn't have a tiredness bar because their fatigue is started at nil and never changed.

Alberth289346:
Did you consider adding a method to the humanoids that returns fatigue as a number?

----
There is lots of code that directly accesses or modifies data owned by an class instance such as a humanoid. That direct access or changing should also includes checking special cases and/or normalizing the value if needed, at every single spot.

When you look at it from one such spot in the code, it is the shortest path to a solution, and as such a reasonable solution.  
However, if you consider what happens by looking at all those spots in the code at the same time, you'll find that it introduces a lot of duplication (all spots that access fatigue must handle the special case, so all spots have "if nil then .." code in some way).  
It also introduces a risk of errors (some developers may not be aware of the special case and don't handle it. With limited testing, this inevitably introduces errors in the code that can be hidden for years.

By introducing methods to query and update a value, the balance shifts. Introducing a method takes some additional effort, but all uses of the method benefit from its existence:
- It creates THE spot for getting some value or changing some data. All code can just do `hum:getFatigue()` or so. No need to go hunting through the code how to get some information, just look at the methods of the object. The method can have documentation telling what it does and expects. The method can also "prepare" the value it returns, so the code requesting the value doesn't have to do that.
- It reduces risks of missing special cases etc. The method handles all weird cases, local code asking for the data can just rely on what method promises about its return value.
- It reduces duplicated code, since that method already did all the conversion and checks. Code using the value gets shorter and less convoluted.
- By calling a method you are informing the owner of the data about the interest in the value or it becomes aware of the change. It may do checks or initiate other actions all by itself, so the code that does the change doesn't have to (again leading to less code duplication, less risk of forgetting to inform the object, etc).

----
Of course this doesn't only hold for `fatigue`, these considerations apply for every case where you have multiple spots with code doing the same steps with the same data (even if the ultimate goal of the spots are unique. Both reading fatigue and updating fatigue need to get the data and convert it to a useful value before they can do what they aim to do).

tobylane:
This accesses all humanoid attributes through Humanoid:getAttributes. A side effect is that receptionists now have a service quality.

CorsixTH/Lua/entities/humanoids/vip.lua
```diff
@@ -265,7 +265,7 @@ function Vip:setVIPRating()
   if count_staff > 1 then
     -- Loop through staff tiredness, if any above verytired, break loop
     for _, staff in ipairs(self.hospital.staff) do
-      if staff.attributes["fatigue"] ~= nil and staff.attributes["fatigue"] >= 0.7 then
+      if staff:getAttribute("fatigue") >= 0.7 then
```
Alberth289346, line 268: Should not be done here, but isn't `verytired` a level config value?

CorsixTH/Lua/entities/humanoids/staff/doctor.lua
```diff
@@ -42,16 +42,14 @@ function Doctor:tickDay()
...
-    if self.attributes["fatigue"] < 0.7 then
-      if self:isResting() then
-        self:setCrazy(false)
-      end
-    else
-      -- doctor can go crazy if they're too tired
-      if math.random(1, 300) == 1 then
-        self:setCrazy(true)
-      end
+  if self:getAttribute("fatigue") < 0.7 then
```
Alberth289346, line 45: Value is here too.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -381,12 +380,12 @@ end
 -- and go to the StaffRoom if it is.
 function Staff:checkIfNeedRest()
   -- Only when the staff member is very tired should the icon emerge. Unhappiness will also escalate
-  if self.attributes["fatigue"] >= 0.7 then
+  if self:getAttribute("fatigue") >= 0.7 then
```
Alberth289346, line 383: And another one.

CorsixTH/Lua/entities/humanoids/vip.lua
lewri: `gbv.VeryTired` indeed exists, requires division by 100 (or multiplying fatigue by 100) to use here.
Also if you want to switch to this in future (I would encourage it), please add a default to `base_config.lua` because it's currently missing.

lewri:
> A side effect is that receptionists now have a service quality.

Should they?

Receptionists currently only have their performance based on skill alone.
https://github.com/CorsixTH/CorsixTH/blob/6d223bcb0485d265085db764cd4f2881d3305c39/CorsixTH/Lua/objects/reception_desk.lua#L135-L143

Potentially you could make `Receptionist:getServiceQuality()` and just return `self.profile.skill`.

CorsixTH/Lua/entities/humanoids/vip.lua
lewri: Essentially you was a boolean returning function probably e.g.
```lua
function Staff:isVeryTired()
  return staff:getAttribute("fatigue") >= 0.7
end
```
(or replace 0.7 with the gbv.VeryTired -- but you could push that as out of scope)

I reference back to https://github.com/CorsixTH/CorsixTH/pull/2529#issuecomment-2016873106 though regarding receptionist handling.

TheCycoONE: > `gbv.VeryTired` indeed exists, requires division by 100 (or multiplying fatigue by 100) to use here. Also if you want to switch to this in future (I would encourage it), please add a default to `base_config.lua` because it's currently missing.

When someone gets to it, multiplying is the better option.

tobylane:
The only place the receptionists service quality is seen is in the new `Humanoid:toString`, though the function you suggest would be good if we expanded what receptionists do. Ready.

lewri:
Referenced config value and boolean implementation in a new issue.

PR #2539, opened 2024-03-28
Add localised text to level and campaign descriptions and winning text

tobylane:
- Collect language codes for languages, two letters long for all but Chinese.
- Use localised strings in the level files for the briefing and debriefing (single scenario description and end of level winning text), and in campaign files for description and winning text.
- Add French translations to the example level and campaign in differing ways.


lewri:
Chinese is difficult under `ISO-639` standards. 
What is documented is `zh_hans` (simplified) and `zh_hant` (traditional). In reality, it's more common to have `zh_cn` (simplified) and `zh_tw` (traditional).

The language files for these don't seem to align to either of those. That is a different issue though.
But my point here is make sure there's accompanying documentation made for this new feature please.

PR #2540, opened 2024-03-29
Add an option to use the advance level movie in custom campaigns

tobylane:
- Add the option 'movie' to custom campaigns. This plays the board game advance level movie, with the correct level number for that campaign (the movie doesn't play if there are more than 12 levels).
- Add this option to the example campaign.

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/fax.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/fax.lua
@@ -212,13 +212,18 @@ function UIFax:choice(choice_number)
         self.ui.app.moviePlayer:playAdvanceMovie(next_level)
       end
     else
-      for i, level in ipairs(self.ui.app.world.campaign_info.levels) do
+      local campaign_info = self.ui.app.world.campaign_info
+      for i, level in ipairs(campaign_info.levels) do
         if self.ui.app.world.map.level_number == level then
-          local next_level = self.ui.app.world.campaign_info.levels[i + 1]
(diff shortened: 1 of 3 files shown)


lewri:
I'd rather we made the movie for campaigns start x away from the final level then on the final level go to the last place on the board. 
With campaigns >12 levels, your code doesn't seem to support that it won't play. Unless you mean after Level 12 no movie plays.

tobylane:
Good idea, done that and rebased.

CorsixTH/Lua/dialogs/fullscreen/fax.lua
```diff
@@ -212,13 +212,18 @@ function UIFax:choice(choice_number)
...
+      for i, level in ipairs(campaign_info.levels) do
         if self.ui.app.world.map.level_number == level then
-          local next_level = self.ui.app.world.campaign_info.levels[i + 1]
+          local next_level = campaign_info.levels[i + 1]
           local level_info, _ = self.ui.app:readLevelFile(next_level)
           if level_info then
             self.ui.app:loadLevel(next_level, nil, level_info.name,
                 level_info.map_file, level_info.briefing, nil, _S.errors.load_level_prefix)
+            if campaign_info.movie then
+              local n = math.max(1, 12 - #campaign_info.levels + i)
```
lewri, line 224: Cool. I'm happier having level 1 board anim repeat until we start moving. Perhaps add a comment to the example file and/or wiki documentation to say movie isn't recommended for campaigns >12.

PR #2541, opened 2024-03-30
Remove queued StaffRise messages on sack/die

### LINKED ISSUE #1642: Staff payrise requests can lapse
#### Describe the issue
If your bottom UI is full of faxes etc. the staff rise requests get delayed. As a result a staff member can leave before the request appears. However, instead of the request being cancelled, it appears once there is a space on the bottom (even though the staff member is gone).

#### Steps to Reproduce

2. You'll find there are two staff rise requests in the bottom UI. They are for two staff already left.

3. However, if you continue playing, the two staff currently in the hospital will also do it, because the bottom UI is full.

#### Save Game
https://www.dropbox.com/s/zkfv0pxoqtr3m3c/delayed-staff-rise.sav?dl=0

#### System Information
**CorsixTH Version:** Trunk / 0.64 rc1

**Operating System:** W10

**Theme Hospital Version:** GOG.

lewri:
*Fixes #1642*

- Now correctly removes a sacked staff member's queued raise request if the message icon is not shown on the bottom panel
- Also for death

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -183,8 +183,8 @@ function Staff:checkIfWaitedTooLong()
           break
         end
       end
-      --If the hospital policy is set to automatically grant wage increases, grant the requested raise
-      --instead of firing the staff member
+      -- If the hospital policy is set to automatically grant wage increases, grant
+      -- the requested raise instead of firing the staff member
       if self.hospital.policies.grant_wage_increase then
         local amount = math.floor(math.max(self.profile.wage * 1.1, (self.profile:getFairWage(self.world) + self.profile.wage) / 2) - self.profile.wage)
         self.quitting_in = nil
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -193,10 +193,11 @@ function Staff:checkIfWaitedTooLong()
         return
       end
 
-      -- Plays the sack sound, but maybe it's good that you hear a staff member leaving?
+      -- else: The staff member is sacked
       if staff_rise_window then
         staff_rise_window:fireStaff()
       else
+        self.hospital:removeMessage(self) -- close any queued wage rise requests
```
Alberth289346, line 200: Wouldn't it be better to move this line inside the `fire` function below? It already does more cleanup.

lewri: I put it there originally. Then put it here because this function leans on the scope of handling the staff rise window. We also call it here for granting the raise by policy L191)
Happy to move it back if you think it's better.

Alberth289346: Oh, don't know, it seemed a general thing that would be needed no matter how a staff got fired, but if you think here is better, that's ok.

I don't know enough of the details of the code to give a definitive answer.

lewri: Yeah. The die function also really should look for this if the room explodes.

Issue #2370, opened 2023-07-31
Add Option to Automatically Mute and Pause Game When Window Loses Focus


RednibCoding:
### What feature or enhancement would you like to add?

Add an option to toggle the following: the game should be automatically paused and muted when the game window loses focus, and unpaused and unmuted when it regains focus.

Create two distinct options: one for muting the game and another for pausing it.
This way, players can seamlessly continue running the game in the background while attending to other tasks on the computer, ensuring it remains muted.

### How would this feature benefit the gameplay of CorsixTH?

When playing the game in windowed mode, it becomes inconvenient having to manually pause and mute the game whenever the window loses focus, especially when you quickly switch to other tasks on your computer. With this new feature, players can enjoy a seamless experience as the game will automatically pause and mute when needed and resume without hassle when it regains focus.

### Was the feature in the original game?

Unsure

### What version of CorsixTH are you using?

CorsixTH 0.67 Release Candidate 1

### Does this feature/enhancement relate to a current problem?

No

lewri:
I can relate, when I added the demo video functionality to the game sometimes I'd forget I left it running and it would suddenly go into breakdance mode 😅 
It's a +1 from m

lewri:
Now #2542 is implemented this relies on me getting #2254 also being resolved to work properly (both this and that issue rely on the same fix for proper functionality)

PR #2542, opened 2024-03-30
Add foundation code for using window active status

lewri:
- Adds some groundwork functionality to tackle #2370 
- Being idle on main menu when game window is no longer is in focus will no longer trigger the demo movie

### DIFF
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -934,8 +934,22 @@ end
 
 local UpdateCursorPosition = TH.cursor.setPosition
 
---! Called when the mouse enters or leaves the game window.
+--! Called when focus changes on game window.
+--!param gain (number) 1 for in-focus, 0 for out-of-focus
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/ui.lua
```diff
@@ -934,8 +934,22 @@ end
...
+
+--! Stores the game window active status
+--!param state (boolean) true for in-focus, 0 for out-of-focus
+function UI:updateWindowActiveStatus(state)
+  self.app.window_active_status = state
+end
+
+--! Gets the game window active status
+--!return true for in-focus, false for out-of-focus
+function UI:getWindowActiveStatus()
```
Alberth289346, line 951: Give that the second function is `get`, making the first function `set` makes sense. In particular, as it is just assigning the given value rather than "merging" it with old data which "update" somewhat suggests.

lewri: Done

Alberth289346 (approved):

PR #2543, opened 2024-03-31
Add more music file extensions

tobylane:
Expand the list of music file extensions that will be accepted for the jukebox. Sourced from the list https://github.com/libsdl-org/SDL_mixer/blob/release-2.8.x/src/music.c#L680 and https://github.com/libsdl-org/SDL_mixer/blob/release-2.8.x/src/music.c#L561 which ends with the comment that everything else gets given to modplug/xmp. GME is excluded as debian and vcpkg don't include it.
Sample music files
https://getsamplefiles.com/sample-audio-files
https://www.stanislavs.org/OldPages/mody_collection/
https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/AIFF/Samples.html only uncompressed

### DIFF
--- a/CorsixTH/Lua/audio.lua
+++ b/CorsixTH/Lua/audio.lua
@@ -100,8 +100,11 @@ function Audio:init()
     - Uses titles from MIDI.TXT if found, else the filename.
   --]]
   local midi_txt -- File name of midi.txt file, if any.
-  local waveform = { MP3=true, OGG=true, WAV=true, AIFF=true, VOC=true, FLAC=true }
-  local instructional = { MID=true, MOD=true, XM=true, XMI=true }
+  local waveform = list_to_set({"OGG", "OPUS", "FLAC", "WV", "WAV", "WAVE",
+      "MPG", "MPEG", "MP3", "MAD", "AIFF", "AIFC", "AIF"})
+  local instructional = list_to_set({"MID", "MIDI", "KAR", "669", "AMF", "AMS", "DBM",
+      "DSM", "FAR", "GDM", "IT", "MED", "MDL", "MOD", "MOL", "MTM", "NST", "OKT", "PTM",
+      "S3M", "STM", "ULT", "UMX", "WOW", "XM"})
 
   local _f, _s, _v
   if music_dir then


CorsixTH/Lua/audio.lua
```diff
@@ -100,8 +100,11 @@ function Audio:init()
     - Uses titles from MIDI.TXT if found, else the filename.
   --]]
   local midi_txt -- File name of midi.txt file, if any.
-  local waveform = { MP3=true, OGG=true, WAV=true, AIFF=true, VOC=true, FLAC=true }
-  local instructional = { MID=true, MOD=true, XM=true, XMI=true }
+  local waveform = list_to_set({"OGG", "OPUS", "FLAC", "WV", "WAV", "WAVE",
+      "MPG", "MPEG", "MP3", "MAD", "AIFF", "AIFC", "AIF"})
+  local instructional = list_to_set({"MID", "MIDI", "KAR", "669", "AMF", "AMS", "DBM",
+      "DSM", "FAR", "GDM", "IT", "MED", "MDL", "MOD", "MOL", "MTM", "NST", "OKT", "PTM",
+      "S3M", "STM", "ULT", "UMX", "WOW", "XM"})
```
TheCycoONE, line 107: Missing XMI.

CorsixTH/Lua/audio.lua
```diff
@@ -100,8 +100,11 @@ function Audio:init()
     - Uses titles from MIDI.TXT if found, else the filename.
   --]]
   local midi_txt -- File name of midi.txt file, if any.
-  local waveform = { MP3=true, OGG=true, WAV=true, AIFF=true, VOC=true, FLAC=true }
-  local instructional = { MID=true, MOD=true, XM=true, XMI=true }
+  local waveform = list_to_set({"OGG", "OPUS", "FLAC", "WV", "WAV", "WAVE",
```
TheCycoONE, line 103: Are you sure they dropped VOC?

tobylane: The codec is in 2.8, but it's not in the music.c detection.

I can't get voc files, or the xmi from CD or demo TH to play in my build or Schnapple's, the xmi error "Unrecognized file type (not AIFF or AIFC)" seems like it's detecting a different magic word. Is it successful on Windows or Linux?

lewri: For VOC:
```bash
Lua\audio.lua:583: Could not load music file 'C:\temp\audio\sample3.voc' (ModPlug_Load failed)
stack traceback:
        [C]: in function 'coroutine.yield'
        Lua\app.lua:1102: in function <Lua\app.lua:1097>
```

XMI (non TH directory)
```bash
Lua\audio.lua:583: Could not load music file 'C:\temp\audio\CANDY.XMI' (Unrecognized file type (not AIFF or AIFC))
stack traceback:
        [C]: in function 'coroutine.yield'
        Lua\app.lua:1102: in function <Lua\app.lua:1097>
```

TheCycoONE:
Ok, we can drop both but as far as I'm concerned xmi is a bug. We have code to support it.

lewri:
NB for new changelog: Support droppoed for VOC (Creative Voice) files.

CorsixTH/Lua/audio.lua
lewri: Afaik this has been megred with XMI still missin gfrom the list?

TheCycoONE:
We didn't really though. We were on the same mixer so it's probably been broken for awhile. Never heard of anyone using it.

PR #2545, opened 2024-04-02
Add confirmation window to map editor restart

tobylane:
- Add confirm window to the restart map editor menu option
- Add mapeditor specific message to quit confirm window
- Move map editor save function to GameUI (used in a future PR).

### DIFF
--- a/CorsixTH/Lua/dialogs/menu.lua
+++ b/CorsixTH/Lua/dialogs/menu.lua
@@ -677,17 +677,10 @@ function UIMenuBar:makeMapeditorMenu(app)
   local menu = UIMenu()
   local hotkeys = app.hotkeys
 
-  local function mapeditorSave(ui)
-    local message = app.world:validateMap()
-    if message then
-      ui:addWindow(UIInformation(ui, {message}))
-    end
-    ui:addWindow(UISaveMap(ui))
-  end
   menu:appendItem(_S.menu_file.load:format(hotkey_value_label("ingame_loadMenu", hotkeys)), function() self.ui:addWindow(UILoadMap(self.ui, "map")) end)
-    :appendItem(_S.menu_file.save:format(hotkey_value_label("ingame_saveMenu", hotkeys)), function() mapeditorSave(self.ui) end)
-    :appendItem(_S.menu_file.restart:format(hotkey_value_label("ingame_restartLevel", hotkeys)), function() app:mapEdit() end)
-    :appendItem(_S.menu_file.quit:format(hotkey_value_label("ingame_quitLevel", hotkeys)), function() self.ui:quit() end)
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/dialogs/menu.lua
```diff
@@ -677,17 +677,10 @@ function UIMenuBar:makeMapeditorMenu(app)
...
-    if message then
-      ui:addWindow(UIInformation(ui, {message}))
-    end
-    ui:addWindow(UISaveMap(ui))
-  end
   menu:appendItem(_S.menu_file.load:format(hotkey_value_label("ingame_loadMenu", hotkeys)), function() self.ui:addWindow(UILoadMap(self.ui, "map")) end)
-    :appendItem(_S.menu_file.save:format(hotkey_value_label("ingame_saveMenu", hotkeys)), function() mapeditorSave(self.ui) end)
-    :appendItem(_S.menu_file.restart:format(hotkey_value_label("ingame_restartLevel", hotkeys)), function() app:mapEdit() end)
-    :appendItem(_S.menu_file.quit:format(hotkey_value_label("ingame_quitLevel", hotkeys)), function() self.ui:quit() end)
+    :appendItem(_S.menu_file.save:format(hotkey_value_label("ingame_saveMenu", hotkeys)), function() self.ui:saveMapeditor() end)
```
lewri, line 681: Shouldn't validation be able to be handled by the UISaveMap class? Or more, should it belong there instead?

tobylane: Done that, though it may break a design rule about logic in a UI class.

lewri: @TheCycoONE can you advise? There's not a clear-cut answer I think as to where the call and code should exist.

tobylane:
https://github.com/tobylane/CorsixTH/commit/c47001007f9a03cd929cbc4af64694d42c6777ea

This is the related commit, which also needs to check the map and open the save map dialog.

CorsixTH/Lua/dialogs/menu.lua
Alberth289346: Would the `Map` class  work?

It seems to mostly handle the loaded map, although there is also other stuff in there such as loading level config, it seems.

Alternatively, something `LoadSave`, although it may be too early for that.

CorsixTH/Lua/game_ui.lua
```diff
@@ -1311,3 +1313,8 @@ end
 function GameUI:showMenuBar()
   self.menu_bar:appear()
 end
+
+function GameUI:restartMapeditor()
```
TheCycoONE, line 1317: restartMapEditor no?

TheCycoONE:
I have no problem with validate being called by the UI. Living in world can be addressed later if desired.

mapeditor vs MapEditor is something to sort out perhaps.

lewri:
@tobylane anything to address still?

tobylane:
No, ready to merge.

PR #2551, opened 2024-04-08
Fix check for audio in menus

### LINKED ISSUE #2549: [Bug] Empty music folder disables audio control
### Describe the issue

If you have a custom music folder set in CorsixTH config the game will load and work with sound but all control over audio gets disabled.

### Steps to Reproduce

1. Point the `audio_music` config value at an empty folder

2. Open a game

3. Navigate to Options and see no audio control.

### Expected Behaviour

Audio control should still be presented to the user. Having 'Music' and 'Jukebox' removed from the menu is fine though.

### Save Game

### CorsixTH Version

0.67

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

```bash
Notice: Audio system loaded, but found no background tracks
```

### Additional Information

image

tobylane:
*Fixes #2549*

- Check for audio when adding audio menu items.

### DIFF
--- a/CorsixTH/Lua/dialogs/menu.lua
+++ b/CorsixTH/Lua/dialogs/menu.lua
@@ -547,7 +547,7 @@ end
 -- game sounds controls are added.
 local function audio_options(menu, game)
   local app = TheApp
-  if not app.audio.has_bg_music then return end
+  if not app:isAudioEnabled() then return end
 
   local function vol(level, setting)
     if setting == "music" then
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/menu.lua
```diff
@@ -547,7 +547,7 @@ end
 -- game sounds controls are added.
 local function audio_options(menu, game)
   local app = TheApp
-  if not app.audio.has_bg_music then return end
+  if not app:isAudioEnabled() then return end
```
lewri, line 550: This is fine as a fix. But it makes music options available again that there isn't much point in controlling/using. Ideally per #2549 a better alternative is any music option should really be disabled instead if there is no songs.

(A separate issue is that movies are also tied to music volume.)

lewri (approved):

PR #2552, opened 2024-04-12
Add descriptive audio error messages

### LINKED ISSUE #2520: [Bug] No soundfont crashes callback handler on launch
### Describe the issue

When loading CorsixTH without an included, detected, or configured soundfont file the callback handler immediately crashes. Gameplay appears to not stop but I'm not sure of the knock-on effects (probably significant).

### Steps to Reproduce

1. Load game without the included SoundFont file, detected file, or configured file.
2. Attempt to play music of MIDI files from Jukebox (or if already enabled, ignore).
3.  Notice the crash in the console.

### Expected Behaviour

Proper handling of lack of soundfont needs to be handled correctly. At a minimum, the handler crash requires suppression before next release.

### Save Game

### CorsixTH Version

Compiled from master

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

```bash
An error has occurred!
Running: The callback handler.
C:/dev/CorsixTH/CorsixTH/Lua\audio.lua:590: Could not load music file 'Sound\Midi\CANDY.XMI' (No SoundFonts have been requested)
stack traceback:
        [C]: in function 'coroutine.yield'
        C:/dev/CorsixTH/CorsixTH/Lua\app.lua:1097: in function <C:/dev/CorsixTH/CorsixTH/Lua\app.lua:1092>
```

### Additional Information

_No response_

tobylane:
*Fixes #2520*

- Change some SDL Mixer error messages to more helpful text, "Required soundfont is not found, please download one. A suitable soundfont is linked from the CorsixTH wiki." and the below message for known errors, and a catch all for anything else.
- Skip the playlist item and uncheck its box.

Screenshot 2024-04-12 at 17 26 26

### DIFF
--- a/CorsixTH/Lua/audio.lua
+++ b/CorsixTH/Lua/audio.lua
@@ -587,11 +587,24 @@ function Audio:playBackgroundTrack(index)
       -- Someone might want to stop the player from
       -- starting to play once it's loaded though.
       self.load_music = true
-      SDL.audio.loadMusicAsync(data, function(music_data, e)
+      SDL.audio.loadMusicAsync(data, function(music_data, err)
 
         if music_data == nil then
-          error("Could not load music file \'" .. (info.filename_music or info.filename) .. "\'" ..
-              (e and " (" .. e .. ")" or ""))
+          info.enabled = false
+          local name, msg = (info.filename_music or info.filename)
+          if not self.warned then -- Warn once per session
+            self.app.ui:addWindow(UIInformation(self.app.ui, {_S.errors.music}))
+          end
+          self.warned = true
+          if err == "No SoundFonts have been requested" then
+            msg = "Required soundfont is not found, please download one. A suitable soundfont is linked from the CorsixTH wiki."
+          elseif err == "XMP: Unrecognized file format" or err == "ModPlug_Load failed" then
+            msg = "Music format not supported for file " .. name
+          else
+            msg = "Could not load music file " .. name .. ". Error: " .. err
+          end
+          if self.app.world then self.app.world:gameLog(msg) end
+          if not self.app.world or not self.app.config.debug then print(msg) end
         else
           info.music = music_data
           -- Do we still want it to play?
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -737,6 +737,7 @@ errors = {
     demo_in_full = "Sorry, you can't open a demo savegame with the full game files loaded. Please update your TH Data folder setting.",
     full_in_demo = "Sorry, you can't open a full game save with the demo files loaded. Please update your TH Data folder setting.",
   },
+  music = "There are playback issues with one or more files in your music directory. Problematic files will be disabled in the jukebox. See the console window for more information.",
 }
 
 warnings = {


lewri (commented):
This could be potentially spammy if jukebox is active on load with the no soundfont one. In general, if a load of files fail in succession it could be spammy.

I could suggest maybe one UIInformation for that session e.g. "There are playback issues with one or more files in your music directory. Problematic files have been disabled in the jukebox. See the console window for more information"?

tobylane:
Is it OK to be spammy in the console? This will tell the user once per session in game, and once per every play attempt of an unplayable file in console/game log.

lewri:
It shouldn't tell them more than once if the song gets disabled no? I suppose soundfont is an issue though as it will still spam for every track. Imo these messages are definitely a necessity, but the soundfont issue might need additional handling outside the scope of this pr.

The console is less likely to cause an issue to the player than putting a graphical information prompt every time.

tobylane:
After this disables the song, the user can enable and try playing it, which causes another console message.

lewri:
> After this disables the song, the user can enable and try playing it, which causes another console message.

I thought that too. The Window class has a way to disable a button. But it's out of scope possibly here.

tobylane:
Added the libmodplug error to the same check as the libxmp error, as these two libraries do the same catch all job for SDL mixer.

TheCycoONE:
This feels fragile but I don't have a better approach.

CorsixTH/Lua/audio.lua
```diff
@@ -587,11 +587,24 @@ function Audio:playBackgroundTrack(index)
...
+          info.enabled = false
+          local name, msg = (info.filename_music or info.filename)
+          if not self.warned then -- Warn once per session
+            self.app.ui:addWindow(UIInformation(self.app.ui, {_S.errors.music}))
+          end
+          self.warned = true
+          if err == "No SoundFonts have been requested" then
+            msg = "Required soundfont is not found, please download one. A suitable soundfont is linked from the CorsixTH wiki."
+          elseif err == "XMP: Unrecognized file format" or err == "ModPlug_Load failed" then
+            msg = "Music format not supported for file " .. name
```
lewri, line 602: Are these interpretations always true?

lewri (commented):
I'm more supportive in this in the fact we no longer cause a crash of the entire application because a song couldn't be handled.

Future idea: you could let the Jukebox's UI disable the button for the session so the user can no longer click it (or gives some UIInformation feedback).

CorsixTH/Lua/audio.lua
tobylane: Don't know. The catch-all way they are used in SDL Mixer means we can't tell some error situations apart.

lewri: In light of that it could be too much to infer the error is a single truth. I'm not precious and happy to keep it as is though.
Does `name` contain the file extension?

tobylane: It could give the original error, plus "This usually means" and these new messages. Or instead of this change, add the details to the wiki.  
Yes, full path.

lewri: I don't think it's worth the confusion on reflection. Full path is good enough.

lewri (approved):

PR #2558, opened 2024-04-19
Fix custom campaign scrollbar

tobylane:
Add support for long custom campaign descriptions, requiring a scrollbar. I think this was intended in the PR that makes the custom campaign dialog, which added this function in the single scenario dialog I copied. https://github.com/CorsixTH/CorsixTH/pull/676/files#diff-7cb59940abdd2a76ef1e4a7ef1a43bed82b0381d550c30229453f02e6230bb04R99
I tested this by repeating a line in the ChrizmanTV campaign description.


TheCycoONE:
Good catch!

PR #2560, opened 2024-04-21
Switch to libcurl for update check

TheCycoONE:
Removes dependence on undermaintained luasocket and luasec libraries.

As an additional benefit, update checks can now be disabled at build time for distros that would prefer not to show it.


TheCycoONE:
@lewri i know it builds. You can set getVersion to return v0.5 in app.lua and try it out.

lewri (changes requested):
Appears something's up
image

Tried with both 91 (0.40) and 156 (0.65) as test values.

I'm not sure if this PR is at fault or #2481

TheCycoONE:
I mentioned that in the room. The old scheme used just one digit of the version. You need to return v0.5. This will correct itself after the next release.

lewri:
> I mentioned that in the room. The old scheme used just one digit of the version. You need to return v0.5. This will correct itself after the next release.

Right let me try again

lewri (commented):
First: Yes this can see updates.

There may have been a reason why @tobylane  previously did not force HTTPS when working on updates previously. It may have just been the update table, but there might be a bigger picture.

Out of scope, but the 'Check for Updates' options menu item should be diabled if built without the update checker. If it is indeed out of scope of this can a new issue be raised to have the behaviour changed?

.github/workflows/Linux.yml
```diff
@@ -68,11 +69,11 @@ jobs:
 
           sudo luarocks --lua-version 5.4 install luafilesystem
 
-      - name: Install CMake 3.10
+      - name: Install CMake 3.14
```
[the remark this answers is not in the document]
TheCycoONE: Changes to the FindCURL module that I'm using.

TheCycoONE:
> Out of scope, but the 'Check for Updates' options menu item should be diabled if built without the update checker. If it is indeed out of scope of this can a new issue be raised to have the behaviour changed?

I'll do this now.

CorsixTH/Lua/dialogs/resizables/main_menu.lua
```diff
@@ -111,7 +111,7 @@ function UIMainMenu:draw(canvas, x, y)
   -- The following strings are drawn in reverse order
   local ly = y + self.height - 15
   -- Move the version string up a bit if showing check for updates disabled warning.
-  if not TheApp.config.check_for_updates then
+  if TheApp:isUpdateCheckDisabledByConfig() then
```
lewri, line 114: Good catch but my concern was also here
image

TheCycoONE: Ah... I'll take care of tgat one too.

TheCycoONE:
Something like this. Reflowing everything might be even better, but at that point I think we are stretching the scope a bit.

lewri (approved):
All okay here

PR #2563, opened 2024-04-26
Linux/MacOS vcpkg support

TheCycoONE:
Allow vcpkg to be used to build for Linux and MacOS.

To make the process simpler and more contained, when luarocks is not used, lua C modules are linked to the executable rather than loaded by lua.


CMakeLists.txt
```diff
@@ -102,6 +105,14 @@ else()
   message("Note: FreeType2 is disabled")
 endif()
 
+if(WITH_UPDATE_CHECK)
+  message("Note: Update check is enabled (default)")
+else()
+  message("Note: Update check is disabled")
+endif()
+
+set(CORSIX_TH_LINK_LUA_MODULES ON)
```
tobylane, line 114:
```suggestion
if(VCPKG_TARGET_TRIPLET AND NOT WITH_LUAROCKS)
  set(CORSIX_TH_LINK_LUA_MODULES ON)
else()
  set(CORSIX_TH_LINK_LUA_MODULES OFF)
endif()
```
This works on Mac and Linux. Windows needs a vcpkg sha bump to confirm.

tobylane:
https://github.com/TheCycoONE/CorsixTH/blob/linux-vcpkg/.github/workflows/Linux.yml#L121

CopyVcpkgLua is here too.

TheCycoONE:
I need some of your other changes to refix windows. Like the sdl find_package line

tobylane (approved):

PR #2572, opened 2024-05-21
Add Actions for Windows Installer

tobylane:
- Builds a Windows installer from successful x64 and x86 Windows Actions runs, chosen by their run id from the url.
 example https://github.com/tobylane/CorsixTH/actions/runs/9158967469 is made of two runs from March.

### DIFF
--- a/.github/workflows/Windows.yml
+++ b/.github/workflows/Windows.yml
@@ -29,7 +29,6 @@ jobs:
     runs-on: windows-2022
     env:
       PRESET: ${{inputs.preset || 'win-x64-rel'}}
-      VCPKG_DEFAULT_TRIPLET: ${{inputs.preset == 'win-x86-rel' && 'x86-windows' || 'x64-windows-release'}}
       ANIMVIEW: ${{inputs.animview && 'ON' || 'OFF'}}
       NAME: CorsixTH${{inputs.animview && '_and_AnimView' || ''}}${{inputs.preset == 'win-x86-rel' && '_x86' || ''}}
     defaults:
@@ -64,10 +63,10 @@ jobs:
         uses: lukka/run-cmake@v10
         with:
           configurePreset: ${{env.PRESET}}
-          configurePresetAdditionalArgs: "['-DBUILD_ANIMVIEW=${{env.ANIMVIEW}}']"
-          buildPreset: ${{env.PRESET}}
-          buildPresetAdditionalArgs: "['--verbose', '--target install',
+          configurePresetAdditionalArgs: "['-DBUILD_ANIMVIEW=${{env.ANIMVIEW}}',
             '-DCMAKE_INSTALL_PREFIX=build/${{env.PRESET}}/CorsixTH/RelWithDebInfo/']"
+          buildPreset: ${{env.PRESET}}
+          buildPresetAdditionalArgs: "['--verbose', '--target install']"
           testPreset: ${{env.PRESET}}
 
       - name: Download soundfont
@@ -79,17 +78,15 @@ jobs:
       - name: Copy data files for archive
         if: inputs.PRESET != 'win-dev'
         run: |
-          cp -R CorsixTH/Lua build/${{env.PRESET}}/CorsixTH/RelWithDebInfo/Lua
-          cp -R CorsixTH/Bitmap build/${{env.PRESET}}/CorsixTH/RelWithDebInfo/Bitmap
-          cp -R CorsixTH/Levels build/${{env.PRESET}}/CorsixTH/RelWithDebInfo/Levels
-          cp -R CorsixTH/Campaigns build/${{env.PRESET}}/CorsixTH/RelWithDebInfo/Campaigns
-          cp CorsixTH/CorsixTH.lua build/${{env.PRESET}}/CorsixTH/RelWithDebInfo/
+          cp -r build/${{env.PRESET}}/. .
+          cp -Ru CorsixTH/RelWithDebInfo/CorsixTH/. CorsixTH/RelWithDebInfo/
+          rm -rf CorsixTH/RelWithDebInfo/CorsixTH/
           if [ "${{inputs.animview}}" == "true" ]; then
             mkdir -p artifact
-            mv build/${{env.PRESET}}/AnimView/RelWithDebInfo artifact/AnimView
-            mv build/${{env.PRESET}}/CorsixTH/RelWithDebInfo artifact/CorsixTH
+            mv AnimView/RelWithDebInfo artifact/AnimView
+            mv CorsixTH/RelWithDebInfo artifact/CorsixTH
           else
-            mv build/${{env.PRESET}}/CorsixTH/RelWithDebInfo artifact
+            mv CorsixTH/RelWithDebInfo artifact
           fi
           ls -R artifact
 
--- a/.github/workflows/WindowsInstaller.yml
+++ b/.github/workflows/WindowsInstaller.yml
@@ -1,0 +1,47 @@
+---
+name: Windows installer
+
+on:
+  workflow_dispatch:
+    inputs:
+      x86:
+        description: 'x86 run ID (url part)'
+        required: true
+        type: number
+      x64:
+        description: 'x64 run ID (url part)'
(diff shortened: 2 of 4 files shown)


lewri (changes requested):
Microsoft Visual C++ 2015 - 2022 Redistributable appear to be missing from this installer (msvc140 and vcruntime140)

image

It installs and launches though, up to that point

TheCycoONE:
I copied the files from my VS install manually

TheCycoONE:
https://github.com/GitCommons/cpp-redist-nsis/blob/main/installer.nsh or something like it with the right version might be useful

lewri:
Apparently the way of dropping them in the install folder isn't recommended. Microsoft encourages bundling the redist exe by the looks of things to install it systemwide (if needed). https://learn.microsoft.com/en-us/cpp/windows/redistributing-visual-cpp-files?view=msvc-170

However, I'm more than happy to continue supporting drag and drop of the required dlls instead to save space. Upon checking our current Appveyor and Github Actions builds though we don't have those bundle in the redist dlls (I assume the x86/x64/ARM versions are all different in retrospect). I'm not sure if the installer could run a check for the redists itself and download the installer itself if missing (I think I've seen some installers do that before).

TheCycoONE:
Some do, if it was a wix/msi installer Microsoft ships a merge module to integrate with it, as we use nsis that's not an option. Supposedly there are different runtime dlls for different locales but I haven't heard of ours not working.

tobylane:
Didn't mean to test in this branch, excuse the runs.

https://github.com/tobylane/CorsixTH/actions/runs/9257678748 has an installer using cpp-redist-nsis. Does that work?

WindowsInstaller/cpp-redist.nsh
```diff
@@ -0,0 +1,43 @@
...
+
+!include LogicLib.nsh
+
+Section "customInit"
+  Var /GLOBAL VCRedistDownload
+  Var /GLOBAL isInstalled
+  ${If} ${RunningX64}
+    ;check H-KEY registry
+    ReadRegDWORD $isInstalled HKLM "SOFTWARE\Wow6432Node\Microsoft\VisualStudio\14.0\VC\Runtimes\x64" "Installed"
+    StrCpy $VCRedistDownload "http://aka.ms/vs/17/release/vc_redist.x64.exe"
```
lewri, line 18: Use https?

WindowsInstaller/cpp-redist.nsh
```diff
@@ -0,0 +1,43 @@
...
+  ${EndIf}
+
+
+  ${If} $isInstalled != "1"
+    MessageBox MB_YESNO "NOTE: This application requires$\r$\n\
+      'Microsoft Visual C++ Redistributable'$\r$\n\
+      to function properly.$\r$\n$\r$\n\
+      Download and install now?" /SD IDYES IDNO VSRedistInstalled
+
+    ;if no goto executed, install vcredist
```
lewri, line 31: Is this meant to say `else install vcredist`?

WindowsInstaller/cpp-redist.nsh
```diff
@@ -0,0 +1,43 @@
...
+    StrCpy $VCRedistDownload "http://aka.ms/vs/17/release/vc_redist.x64.exe"
+  ${Else}
+    ReadRegDWORD $isInstalled HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x86" "Installed"
+    StrCpy $VCRedistDownload "http://aka.ms/vs/17/release/vc_redist.x86.exe"
+  ${EndIf}
+
+
+  ${If} $isInstalled != "1"
+    MessageBox MB_YESNO "NOTE: This application requires$\r$\n\
+      'Microsoft Visual C++ Redistributable'$\r$\n\
```
lewri, line 27: 2015-2022?

lewri (commented):
This needs a fallback operation for the vcredist installer failing to download.
Also I tried with a network connection and it still skipped the vcredist and carried on installing (leading to missing dll errors again)

WindowsInstaller/cpp-redist.nsh
tobylane: There's a limitation/unreliability in the download command and https.

WindowsInstaller/cpp-redist.nsh
tobylane: Kind of. The goto is at the end, a different flow to if else.

.github/workflows/Windows.yml
```diff
@@ -28,10 +28,10 @@ jobs:
   Windows:
     runs-on: windows-2022
     env:
-      PRESET: ${{inputs.preset || 'win-x64-rel'}}
-      VCPKG_DEFAULT_TRIPLET: ${{inputs.preset == 'win-x86-rel' && 'x86-windows' || 'x64-windows-release'}}
+      PRESET: ${{inputs.preset || 'win-x64-static'}}
+      VCPKG_DEFAULT_TRIPLET: ${{inputs.preset == 'win-x86-static' && 'x86-windows' || 'x64-windows-staticease'}}
```
lewri, line 32:
```suggestion
      VCPKG_DEFAULT_TRIPLET: ${{inputs.preset == 'win-x86-static' && 'x86-windows' || 'x64-windows-static-release'}}
```
I assume you meant that?

CMakePresets.json
```diff
@@ -47,6 +47,30 @@
...
+    {
+      "name": "win-x64-static",
+      "binaryDir": "${sourceDir}/build/win-x64-static",
+      "generator": "Visual Studio 17 2022",
+      "architecture": "x64",
+      "cacheVariables": {
+        "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "VCPKG_TARGET_TRIPLET": "x64-windows-static",
+        "ENABLE_UNIT_TESTS": "ON",
+        "DCMAKE_MSVC_RUNTIME_LIBRARY": "MultiThreaded"
```
TheCycoONE, line 60: D should not be there

WindowsInstaller/LanguageStrings.nsh
```diff
@@ -27,6 +27,7 @@ LangString options_title      ${LANG_DANISH} "Yderligere indstillinger"
 LangString save_in_appdata    ${LANG_DANISH} "Gemme spil data og indstillinger i %APPDATA% (Anbefales)"
 LangString remove_saves       ${LANG_DANISH} "Vil du gerne beholde dine gemte spil?"
 LangString overwrite_install  ${LANG_DANISH} "Det ser ud til at du forsøger at installere over en eksisterende installation af CorsixTH. Eventuelle ændringer mellem filerne kan gå tabt. Ønsker du at fortsætte?"
+LangString cpp_redist         ${LANG_DANISH} "NOTE: This application requires$\r$\n'Microsoft Visual C++ Redistributable'$\r$\nto function properly.$\r$\n$\r$\nDownload and install now?"
```
TheCycoONE, line 30: If we get /MT compiling properly we don't need the redistributable anymore.

CMakePresets.json
TheCycoONE: https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html

TheCycoONE: You also need to bump CMake's minimum version to 3.15 to use this flag.

tobylane:
This uses [InstallRequiredSystemLibraries](https://cmake.org/cmake/help/latest/module/InstallRequiredSystemLibraries.html) to collect the dlls.

Installer made from this: https://github.com/tobylane/CorsixTH/actions/runs/9921377052

lewri:
The latest installer functions as intended for me

lewri (commented):
Are we now sticking with CMake 3.14?

tobylane:
> Are we now sticking with CMake 3.14?

Yes, this cmake function is 3.10+.

TheCycoONE:
Perfect, thanks @tobylane !

PR #2574, opened 2024-06-15
Use static mt builds for release on windows

TheCycoONE:
Removes the need for shipping the runtime

Screenshot_20240615_233000

### DIFF
--- a/.github/workflows/Linux.yml
+++ b/.github/workflows/Linux.yml
@@ -69,11 +69,11 @@ jobs:
 
           sudo luarocks --lua-version 5.4 install luafilesystem
 
-      - name: Install CMake 3.14
+      - name: Install CMake 3.15
         if: matrix.cmake
         run: |
-          # Install CMake version 3.14, the oldest CorsixTH-supported version, which does not support Lua 5.4
-          curl -L https://github.com/Kitware/CMake/releases/download/v3.14.6/cmake-3.14.6-Linux-x86_64.sh -o cmake.sh
+          # Install CMake version 3.15, the oldest CorsixTH-supported version, which does not support Lua 5.4
+          curl -L https://github.com/Kitware/CMake/releases/download/v3.15.6/cmake-3.15.6-Linux-x86_64.sh -o cmake.sh
           sudo bash cmake.sh --prefix=/usr/local/ --exclude-subdir --skip-license
           cmake --version
       - name: Create CorsixTH${{ matrix.animview }} makefiles with ${{ matrix.lua }}
--- a/.github/workflows/Windows.yml
+++ b/.github/workflows/Windows.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Setup vcpkg
         uses: lukka/run-vcpkg@v11
         with:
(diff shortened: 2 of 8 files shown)


lewri:
Warning in windows action if ne needs addressing in this PR (think it's come from this PR)
```
Windows: build/win-x64-rel/vcpkg_installed/x64-windows-static-release/share/libffi/libffiConfig.cmake#L2
find_package(libffi) is deprecated.
```

tobylane:
x86 run https://github.com/CorsixTH/CorsixTH/actions/runs/9534791148/job/26279717337#step:7:1980 has the error

>LINK : fatal error LNK1104: cannot open file 'pcre2-8d.lib' [D:\a\CorsixTH\CorsixTH\build\win-x86-rel\CorsixTH\CorsixTH.vcxproj]

lewri (commented):
I'm happy once the deprecation warning is decided if it requires action or not. 
I was writing this before toby spotted the x86 issue -- so that one should probably get looked at too.

Just as a point:
Windows build now increases from 52MB to 132MB (uncompressed) with these changes. Are we comfortable with such a level of inflation of size? If the redistributable download was able to work via Toby's route that would be a nice option but this one is of course more stable.

I would like to see the option for a non-static build option for testing purposes locally rather than the bloat (I'm not sure if win-dev would be the solution to this)

TheCycoONE:
> x86 run https://github.com/CorsixTH/CorsixTH/actions/runs/9534791148/job/26279717337#step:7:1980 has the error
> 
> > LINK : fatal error LNK1104: cannot open file 'pcre2-8d.lib' [D:\a\CorsixTH\CorsixTH\build\win-x86-rel\CorsixTH\CorsixTH.vcxproj]

Looks like fluidsynth again. Seems that package is pretty broken.

TheCycoONE:
The size I can speak to.

The pdb is much bigger because it includes the debug info for all the packages. We will not be including it in the installer, but it's useful to have for ourselves for crash reports, the old pdb was minimal and didn't include symbols for the libraries.

The release (without debug info) exe is 14MB on my vm, the release with debug info from the pipeline is 24MB

The dlls that are no longer shipped were ~18MB, plus 1.7M for the runtime. The 0.67 release was 23M of dlls including the runtime. Now by doing a full static compile we can add things like link time optimization which should improve the program; I haven't played with them at all yet.

It might be worth putting the pdb in a separate artifact, but I'm not overly concerned; I think people that use these builds can benefit from having it.

The deprecation notice is within a vcpkg package, and they will handle it at some point; I'm not concerned.

The continuing issues with fluidsynths dependencies are a major concern; I can't move ahead in it's current state.

TheCycoONE:
https://github.com/FluidSynth/fluidsynth/issues/1337 and https://github.com/microsoft/vcpkg/issues/36139 for tracking the fluidsynth related build errors.

lewri:
Do we know if this is planned to come in for the next version or will it be postponed? Given out usual turnaround timeline it may be a bit close to go for such a change.

TheCycoONE:
It's not coming as long as it's not supported, and there's no timeline for a fix from vcpkg or fluidsynth. I think we look for another way to ship.

lewri:
As toby's method is working now are we still continuing with static?

PR #2578, opened 2024-06-23
End condtions and progress report dialog improvements

tobylane:
- These improvements are preparation for adding new criteria, and won't have any impact yet. The additional commits in my [addcriteria](https://github.com/CorsixTH/CorsixTH/compare/master...tobylane:CorsixTH:addcriteria) branch use the first, third and fifth commits here.

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
@@ -73,6 +73,9 @@ function UIProgressReport:UIProgressReport(ui)
     local tooltip
     if crit_table.formats == 2 then
       tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value), math.floor(cur_value))
+    elseif crit_table.formats == 3 then
+      tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value) / 1000,
+          math.floor(cur_value) / 1000)
     else
       tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value))
     end
--- a/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
@@ -71,13 +71,19 @@ function UIProgressReport:UIProgressReport(ui)
     -- FIXME: res_value and cure_value are depersisted as floating points, using
     -- string.format("%.0f", x) is not suitable due to %d (num) param in _S string
     local tooltip
+    if crit_table.two_tooltips then
+      local direction = crit_table.win_value and "over" or "under"
+      tooltip = _S.tooltip.status[direction][crit_name]
+    else
+      tooltip = _S.tooltip.status[crit_name]
+    end
     if crit_table.formats == 2 then
-      tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value), math.floor(cur_value))
+      tooltip = tooltip:format(math.floor(res_value), math.floor(cur_value))
     elseif crit_table.formats == 3 then
-      tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value) / 1000,
+      tooltip = tooltip:format(math.floor(res_value) / 1000,
           math.floor(cur_value) / 1000)
     else
-      tooltip = _S.tooltip.status[crit_name]:format(math.floor(res_value))
+      tooltip = tooltip:format(math.floor(res_value))
     end
     if not crit_table.icon_file then -- Icons from QData/Rep02V
       self:addPanel(crit_table.icon, x, 240)
--- a/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
@@ -228,7 +228,7 @@ function UIProgressReport:draw(canvas, x, y)
 end
 
 function UIProgressReport:afterLoad(old, new)
-  if old < 188 then
+  if old < 206 then
     self:close()
   end
 
--- a/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
@@ -37,7 +37,12 @@ function UIProgressReport:UIProgressReport(ui)
     self.red_font = gfx:loadFont("QData", "Font101V", false, palette)
     self.normal_font = gfx:loadFont("QData", "Font100V", false, palette)
     self.small_font = gfx:loadFont("QData", "Font106V")
-    self.panel_sprites = gfx:loadSpriteTable("QData", "Rep02V", true, palette)
+    -- Load all sprite tables needed for all goal icons
+    self.panel_sprites_table = {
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/languages/english.lua
```diff
@@ -945,6 +945,26 @@ hotkeys_file_err = {
...
+  under = {
+    reputation = tooltip.status.reputation,
+    balance = tooltip.status.balance,
+    percentage_cured = tooltip.status.percentage_cured,
+    num_cured = tooltip.status.num_cured,
+    percentage_killed = tooltip.status.percentage_killed,
+    value = tooltip.status.value,
+  },
+}
+
```
lewri, line 967: This is probably ill advised as a permanent fixture. The only reason I used remappings below in unused was because they were bridges to be removed in future release. Is there another way to achieve what you're after?

CorsixTH/Lua/endconditions.lua
```diff
@@ -19,7 +19,8 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
 -- The "Winning and Losing Conditions" section of level files uses Criteria
---  numbers from the order of this table.
+--  numbers from the order of this table. Any Criteria numbers higher than
+--  the size of this table will be silently ignored.
```
lewri, line 23: Ideally should be aligned the same or two space indent

CorsixTH/Lua/endconditions.lua
```diff
@@ -124,7 +126,7 @@ end
 -- lose criteria with the smallest gap between current value and boundary,
 -- then fill up to five with win criteria in the best group.
 function EndConditions:generateReportTable(hospital)
-  local count, lose_table, report_table, tmp_table = 0, {}, {}, {}
+  local count, lose_table, tmp_table, crit_set, report_table = 0, {}, {}, {}, {}
```
lewri, line 129: Is it necessary to change the order?

lewri (commented):
Need to close the progress report on this change.

CorsixTH/Lua/languages/english.lua
lewri: Ah yeah, this was why:
`This mapping will cause the legacy string to print in English`

tobylane: I'm trying to offer translators and criteria-writers the ability to write two strings instead of one, while keeping all the original strings in their place. Copying them in breaches copyright, rewriting them is a large step. Maybe an allowlist, so that only new criteria that have an under/over pair will try to use them.

Additionally, we now see the English tooltips are worded more for lose (under) but I don't want to apply that rule to other languages. Adding the new text for English, but in an uninherited way is not possible?

lewri: I get the premise -- I support that. I just don't think this should be the method.
We have previously allowed lenience for typos or clarifications of a string.

Another issue is below, I think you just missed one though.
```lua
    local direction = crit_table.win_value and "over" or "under"
```
`tooltip.status.percentage_killed` is in reverse to the direction of others (win is under rather than over). I think you can actually omit that tooltip from this, it does look succinctly clear enough.

I think the stemmed ambiguity we have (#2579) is only `tooltip.status.balance` really because it can go negative. The others are always `>=0` and are more logical to understand. However, I'm open if I'm missing something.

tobylane: Undone this, now only criteria that have the field `two_tooltips` set to true will look for over/under strings. Example in the [addcriteria branch.](https://github.com/CorsixTH/CorsixTH/compare/master...tobylane:CorsixTH:addcriteria)

CorsixTH/Lua/endconditions.lua
tobylane: They're now in the order they're used.

CorsixTH/Lua/endconditions.lua
```diff
@@ -124,7 +127,7 @@ end
 -- lose criteria with the smallest gap between current value and boundary,
 -- then fill up to five with win criteria in the best group.
 function EndConditions:generateReportTable(hospital)
-  local count, lose_table, report_table, tmp_table = 0, {}, {}, {}
+  local count, lose_table, tmp_table, report_table, crit_set = 0, {}, {}, {}, {}
```
lewri, line 130: `crit_set` Variable may benefit from being given unshortened name here imo

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -66,12 +70,23 @@ function UIProgressReport:UIProgressReport(ui)
     -- FIXME: res_value and cure_value are depersisted as floating points, using
     -- string.format("%.0f", x) is not suitable due to %d (num) param in _S string
```
lewri, line 71: The fixme could now be moved to above line 79

CorsixTH/Lua/endconditions.lua
```diff
@@ -143,16 +146,21 @@ function EndConditions:generateReportTable(hospital)
   for _, crit_table in pairs(tmp_table) do
     table.insert(report_table, crit_table)
     count = count + 1
-    if count == 5 then break end
+  end
+  -- If there are more than five, keep the five criteria closest to meeting lose conditions
+  if count > 5 then
+    table.sort(report_table, function(a,b) return a.gap > b.gap end)
+    for n = 6, #report_table do report_table[n] = nil end
```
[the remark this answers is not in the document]
TheCycoONE: This is reasonably standard; what do you think would be broken?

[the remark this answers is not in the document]
lewri: You're probably right. I think I was confusing a different scenario that method causes a problem with.

lewri:
Generally I'm okay with this as is, my latest comments shouldn't be blockers here.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -37,7 +37,11 @@ function UIProgressReport:UIProgressReport(ui)
     self.red_font = gfx:loadFont("QData", "Font101V", false, palette)
     self.normal_font = gfx:loadFont("QData", "Font100V", false, palette)
     self.small_font = gfx:loadFont("QData", "Font106V")
-    self.panel_sprites = gfx:loadSpriteTable("QData", "Rep02V", true, palette)
+    self.panel_sprites_table = {
```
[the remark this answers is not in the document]
tobylane: Added, though I will write a wiki article for adding new criteria which expands on this.

lewri (commented):
Afterload bump for this one now.
Requesting @TheCycoONE to re-review and merge if happy.

CorsixTH/Lua/endconditions.lua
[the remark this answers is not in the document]
TheCycoONE: I see what you mean, but in lua the end condition is only called once and it is always fine to set a list entry to nil

TheCycoONE:
Merge after the release

TheCycoONE:
Need to resolve conflicts and bump save version, then merge.

tobylane:
Rebased and bumped.

PR #2584, opened 2024-07-07
Add Load/save buttons & save warning message

### LINKED ISSUE #2576: Make save/load UI more unique

Pulled from the CorsixTH Blog:

>can you make different design of save- and load-game screens? Both menu items are located very close to each other, so the player sometimes mistakenly chooses the wrong item and cannot immediately notice it, since the screens are very similar. Even adding two icons “Save” and “Load” of different colors will be enough.

Multiple levels of differentiation usually can assist someone who may have difficulty discerning the two. I do admit our save/load UI elements are very samey.

Yes

0.67

No

### Additional Information

Technically there was a different in the original, since the original save UI was a menu item. The load screen was independent as well as being another menu item.

lewri:
*Fixes #2576*

- Adds an optional parameter to `UIFileBrowser` to set custom text for the ok/submit button
- Introduces this to Save/Load Game/Map
- Optional additional commit to add a message when attempting to save without selecting/setting a filename.

I added some doc to the `UIFileBrowser` main function but level blank 2 parameters not 100% sure what they do.
image
image

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/file_browsers/load_game.lua
+++ b/CorsixTH/Lua/dialogs/resizables/file_browsers/load_game.lua
@@ -27,7 +27,7 @@ local UILoadGame = _G["UILoadGame"]
 function UILoadGame:UILoadGame(ui, mode)
   local treenode = FilteredFileTreeNode(ui.app.savegame_dir, ".sav")
   treenode.label = "Saves"
-  self:UIFileBrowser(ui, mode, _S.load_game_window.caption:format(".sav"), 295, treenode, true)
+  self:UIFileBrowser(ui, mode, _S.load_game_window.caption:format(".sav"), 295, treenode, true, _S.load_game_window.load_button)
   -- The most probable preference of sorting is by date - what you played last
   -- is the thing you want to play soon again.
   self.control:sortByDate()
--- a/CorsixTH/Lua/dialogs/resizables/file_browsers/load_game.lua
+++ b/CorsixTH/Lua/dialogs/resizables/file_browsers/load_game.lua
@@ -27,7 +27,8 @@ local UILoadGame = _G["UILoadGame"]
 function UILoadGame:UILoadGame(ui, mode)
   local treenode = FilteredFileTreeNode(ui.app.savegame_dir, ".sav")
   treenode.label = "Saves"
-  self:UIFileBrowser(ui, mode, _S.load_game_window.caption:format(".sav"), 295, treenode, true, _S.load_game_window.load_button)
+  self:UIFileBrowser(ui, mode, _S.load_game_window.caption:format(".sav"), 295,
+      treenode, true, _S.load_game_window.load_button)
   -- The most probable preference of sorting is by date - what you played last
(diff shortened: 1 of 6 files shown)


CorsixTH/Lua/dialogs/resizables/file_browsers/load_game.lua
```diff
@@ -27,7 +27,7 @@ local UILoadGame = _G["UILoadGame"]
 function UILoadGame:UILoadGame(ui, mode)
   local treenode = FilteredFileTreeNode(ui.app.savegame_dir, ".sav")
   treenode.label = "Saves"
-  self:UIFileBrowser(ui, mode, _S.load_game_window.caption:format(".sav"), 295, treenode, true)
+  self:UIFileBrowser(ui, mode, _S.load_game_window.caption:format(".sav"), 295, treenode, true, _S.load_game_window.load_button)
```
Alberth289346, line 30:
```suggestion
  self:UIFileBrowser(ui, mode, _S.load_game_window.caption:format(".sav"), 295,
      treenode, true, _S.load_game_window.load_button)
```
Split the line like below?

CorsixTH/Lua/dialogs/resizables/file_browsers/load_map.lua
```diff
@@ -28,7 +28,8 @@ function UILoadMap:UILoadMap(ui, mode)
   local path = ui.app.user_level_dir
   local treenode = FilteredFileTreeNode(path, ".map")
   treenode.label = "Maps"
-  self:UIFileBrowser(ui, mode, _S.load_game_window.caption:format(".map"), 295, treenode, true)
```
Alberth289346, line 31: `load_game_window` name for loading a map?

Save uses `save_map_window`, so that seems ok.

CorsixTH/Lua/dialogs/resizables/file_browsers/save_game.lua
```diff
@@ -75,6 +76,7 @@ function UISaveGame:confirmName()
   local filename = self.new_savegame_textbox.text
   local app = self.ui.app
   if filename == "" then
+    self.ui:addWindow(UIInformation(self.ui, { _S.save_game_window.missing_filename }))
```
Alberth289346, line 79: I'd write the table without spaces inside the curly braces (also at line 84). Not sure what the convention is.

Also not too important, I am fine if you prefer this.

CorsixTH/Lua/dialogs/resizables/file_browsers/load_map.lua
Alberth289346: Ah, in #2593. Fair enough.

CorsixTH/Lua/dialogs/resizables/file_browsers/save_game.lua
lewri: Convention appears to be to use spaces (at least according to `app.lua`) inside, I probably just followed the example.
I'll commit the other line break suggestion though you put in elsewhere.

Alberth289346 (approved):

Issue #2469, opened 2023-12-30
[Bug] Right mouse panning can cause glitches with object placement


lewri:
### Describe the issue

Following on from #2375 -- we now need to address a small issue where enabling Right Mouse panning causes some little bugs on object placement:
1. When panning, the object will jitter around a little.
2. When you finish panning, the object will rotate 1 time onMouseUp.

The jitter I could live with for longer, the unintentional rotate is a little bit of a pain.

### Steps to Reproduce

1. Start a game on a CorsixTH build on or after #2375 
2. Set mouse panning to right mouse
3. Go into an object placement task and use panning to move around the view.

### Expected Behaviour

We should avoid jittering of objects during a pan -- we also want to avoid unintentional rotation.

### Save Game

Save game attached to report: No

### CorsixTH Version

version with 2375 implemented (based on 0.67/master)

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

lewri:
Wonder if we could make a secondary rotation map to a keybind instead. Thus disabling right click rotation when it's used for panning.

Or we swap right button with middle button in addition too to support users who have a middle button on the mouse.

lewri:
> Wonder if we could make a secondary rotation to a keybind instead. Thus disabling right click rotation when it's used for panning.

This exists my bad: `ingame_rotateobject = "space",`

lewri:
A new one: object pickup happens too if you try to pan while hovered on a selectable entity.

lewri:
I'm going to submit a PR to remove this option from the UI and label it as experimental in the config. Right now I keep thinking more things it breaks than benefits. Eventual fix is probably going to need to be a key remap for right mouse panning (or make it disable certain right more operations).

PR #2588, opened 2024-07-10
Mark Right Mouse Scrolling Experimental

lewri:
*See #2469*

-  Adds to the config.txt that the setting is experimental
- Removes the setting from the Options menu (preserves the code in blockquote)

@tobylane let me know if a button should be moved after this, the LHS column has a gap at the bottom, RHS does not.

### DIFF
--- a/CorsixTH/Lua/config_finder.lua
+++ b/CorsixTH/Lua/config_finder.lua
@@ -288,6 +288,8 @@ local string_01 = [=[
 -------------------------------------------------------------------------------
 -- Right Mouse Scrolling: By default, it is disabled (right_mouse_scrolling = false).
 -- This means that the default scrolling method is pressing the middle mouse button.
+-- Please note this an Experimental Feature and may interfere with other right mouse
+-- operations. Report bugs for this on Github Issue 2469.
 --]=] .. '\n' ..
 'right_mouse_scrolling = ' .. tostring(config_values.right_mouse_scrolling) .. '\n' .. [=[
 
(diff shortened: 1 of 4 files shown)


Alberth289346:
Since it interferes in unintended ways, taking out the functionality is likely a good idea.

Commenting out code is mostly a good idea if there is a plan to have a final decision about its fate, to avoid having it linger in there forever.  
An alternative to commenting out code is to simply delete it, and note down the git revision containing it somewhere findable. Technically the latter isn't even necessary, as git allows you to search for revisions containing some given text or  text-pattern.  
A second alternative is of course to delete it and just to write it again if the need arises. It likely needs to change anyway, since the current implementation isn't doing what it's supposed to do.

lewri:
> Commenting out code is mostly a good idea if there is a plan to have a final decision about its fate, to avoid having it linger in there forever.

I would optimistically hope to uncomment the code again once the bugs get remedied.

TheCycoONE:
It can be restored from git, remove it.

tobylane:
It looks a bit odd, so yes do move a button across (not language). Leave a space before the lower row to prevent them overlapping this new button on the left side.

```lua
  self.zoomspeed_button = self.zoomspeed_panel : makeToggleButton(0, 0, BTN_WIDTH, BTN_HEIGHT, nil, self.buttonZoomSpeed) : setTooltip(_S.tooltip.options_window.zoom_speed)

  self:_getOptionYPos() -- Increment the y position once more while there are fewer options in the second column
  local lower_row_y_pos = self:_getOptionYPos()
  -- "Customise" button
```

lewri:
I have moved global audio across to the left now as it made more sense than breaking up the zoom/shift buttons.

PR #2596, opened 2024-08-04
Copyrights 2024

lewri:
Refresh copyright list
I also inserted spaces between the commas but I can revert that if copyright years aren't meant to be spaced.


TheCycoONE:
Thanks

PR #2599, opened 2024-08-07
Add Actions to create a draft release

tobylane:
- Add Actions that takes the recent additions to the changelog and post them as a draft release.
[My repo's Actions run](https://github.com/tobylane/CorsixTH/actions/runs/10288532931) and [the release](https://github.com/tobylane/CorsixTH/releases/tag/0.67.9beta).
This uses [ncipollo/release-action](https://github.com/marketplace/actions/create-release). It can only mark prereleases or create discussions on full releases, not drafts, so these remain as manual steps.

### DIFF
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -1,0 +1,38 @@
+---
+name: Create draft Github release
+on:
+  push:
+    tags:
+    - '*'
+
+jobs:
+  release:
+    name: Prepare release
+    if: github.repository == 'CorsixTH/CorsixTH'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        repository: CorsixTH/CorsixTH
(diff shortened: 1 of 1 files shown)


TheCycoONE:
This requires changing the workflow every release, it would be better if from tag was a variable, it would be even better if it could be automatically determined.

`fromver=$(git tag -l --sort=-creatordate | head -n 1)` might work though it will be from the last prerelease. You could probably get fancier with grep before the head if you only cared about particular patterns.

tobylane:
Improved to fetch the relevant changes for the upcoming/current release. This [new release](https://github.com/tobylane/CorsixTH/releases/tag/v0.67.9beta) ends with the diff of the older changelog from a codespell update, which could happen in the future and needs manually deleting. [Actions run.](https://github.com/tobylane/CorsixTH/actions/runs/11519824944/job/32069738434)

tobylane:
The motivation for this is to ease creating releases, but I don't think it does. The draft release this makes still needs assets, so whoever is uploading them could make the release first.

PR #2605, opened 2024-08-24
soda price can now configurable in the level file

SlightPersimmon1:
Allows soda price to be configurable in the level file in order to expand level creation capabilities.


tobylane:
This is a good idea. I've added an afterload, and rebased the branch.

TheCycoONE (approved):
Looks good, will need to be documented, maybe add to the example configuration.

PR #2610, opened 2024-09-07
Prep for next prerelease

lewri:
Awaiting decision on if this will be Beta 2 or RC 1 (and a suitable period of test time)

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -29,7 +29,7 @@ local runDebugger = corsixth.require("run_debugger")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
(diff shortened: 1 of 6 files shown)


lewri:
@TheCycoONE are we going with Beta 2, and halting features after that? The features for Beta 1 probably need a short window to be set in.

TheCycoONE:
RC1 is when I intend feature freeze.

lewri:
> RC1 is when I intend feature freeze. I intend to go straight to it. We don't usually see much stabilization anyway

So format this one to RC1 year?

lewri:
Updated to be RC1

TheCycoONE:
```
appstreamcli validate --pedantic CorsixTH/com.corsixth.corsixth.metainfo.xml
✔ Validation was successful.
```

PR #2615, opened 2024-09-09
Allow switching between staff by using staff name label

Matroftt:
Screenshot 2026-06-15 at 14 18 39

- One more way to switch between staff, similar to switching between machinery
- Press RMB on name in staff dialog to switch


TheCycoONE:
Going to hold this one for the next release.

Thanks.

CorsixTH/Lua/dialogs/staff_dialog.lua
```diff
@@ -260,7 +260,8 @@ function UIStaff:onMouseUp(button, x, y)
   end
   local repaint = Window.onMouseUp(self, button, x, y)
   -- Test for hit within the view circle
-  if button == "right" and is_in_view_circle(x, y, self.staff.profile.humanoid_class == "Handyman") then
+  if button == "right" and is_in_view_circle(x, y, self.staff.profile.humanoid_class == "Handyman")
+     or x > 18 and x < 170 and y > 19 and y < 42 then
```
TheCycoONE, line 264: Could use an explanation. That's a lot of magic numbers.

Matroftt: These are coordinates of name label

TheCycoONE: Yes

lewri: Or a table to label them instead (but could be overkill, and you'd change the inequalities to `>=` etc)
e.g. (change name to what is relevant)
```lua
local name_box = {
  x_min = 19
  x_max = 169
  y_min = 20
  y_max = 41
}
```

Though are these name box's coordinates not always specified in the main function `UIStaff:UIStaff` (when we draw we typically specify specific pixels or an `x+number` increment), rather than repeat them two times?

Matroftt: > Or a table to label them instead (but could be overkill, and you'd change the inequalities to `>=` etc) e.g. (change name to what is relevant)
> 
> ```lua
> local name_box = {
>   x_min = 19
>   x_max = 169
>   y_min = 20
>   y_max = 41
> }
> ```
> 
> Though are these name box's coordinates not always specified in the main function `UIStaff:UIStaff` (when we draw we typically specify specific pixels or an `x+number` increment), rather than repeat them two times?

Could use the similar table to refactor machine menu later, because it does checking for the click same as my pull request

Alberth289346:
Is it possible to query the coordinates from the label widget in the UI itself? That would be more future-proof.

tobylane:
The numbers are in self.tooltip_regions[1].

Matroftt:
> self.tooltip_regions[1].

How to find width and height value using it? I've only managed to get x and y

tobylane:
The r (right) and b (bottom) fields are in that table, they are x/y + width/height.

Matroftt:
> The r (right) and b (bottom) fields are in that table, they are x/y + width/height.

thanks tobylane

CorsixTH/Lua/dialogs/staff_dialog.lua
```diff
@@ -277,6 +285,11 @@ function UIStaff:onMouseUp(button, x, y)
       local staff = ui.hospital.staff[i]
       if staff.profile.humanoid_class == self.staff.profile.humanoid_class then
         ui:addWindow(UIStaff(ui, staff))
+        if hit_namebox then
+          local sx, sy = ui.app.map:WorldToScreen(staff.tile_x, staff.tile_y)
+          local dx, dy = staff.th:getPosition()
+          ui:scrollMapTo(sx + dx, sy + dy)
+        end
```
lewri, line 292: If you wanted to go one step further, you could pull this out into a local function inside `UIStaff:onMouseUp`; possibly just before the `if button = right` line

```lua
local function scrollToMapPosition(staff)
  if not hit_namebox then return end
  local sx, sy = ui.app.map:WorldToScreen(staff.tile_x, staff.tile_y)
  local dx, dy = staff.th:getPosition()
  ui:scrollMapTo(sx + dx, sy + dy)
end
```
(I'm pretty sure inserting the function at the point I said means it is aware of `local ui` and `local hit_namebox`)

Then:
```lua
scrollToMapPosition(staff)
```

lewri (approved):
While this isn't a stopper the behaviour seems a little peculiar between the two modes from a UX perspective:

**In view circle**
Left click - _scrolls to staff member (can hold down to follow)_
Right click - _changes staff member, does not scroll_

**In name label**
Left click - _nothing_
Right click - _changes staff member and scrolls to position._

PR #2616, opened 2024-09-15
Free build improvements in town map

Matroftt:
-  If free build is enabled, town map displays 0$ heating bills and 0$ parcel cost.

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
@@ -189,11 +189,10 @@ function UITownMap:draw(canvas, x, y)
 
   -- Heating costs
   local heating_costs = math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)
-  if app.config.free_build_mode then
-    self.info_font:draw(canvas, "0",  x + 100, y + 355)
-  else
-    self.info_font:draw(canvas, ("%8i"):format(heating_costs),  x + 100, y + 355)
+  if app.config.free_build_mode then  -- Everything is 0$ in free build mode
+    heating_costs = 0
   end
+  self.info_font:draw(canvas, ("%8i"):format(heating_costs),  x + 100, y + 355)
 
   -- Draw balance with temporary offset in unicode languages
   local offset_x, offset_y = 0, 0
@@ -307,11 +306,9 @@ function UITownMap:draw(canvas, x, y)
       local owner_num = map.th:getPlotOwner(self.hover_plot)
       if owner_num == 0 then
         owner = _S.town_map.for_sale
-	-- Everything is 0$ in free build
-	if app.config.free_build_mode then
+	price = "$" .. map:getParcelPrice(self.hover_plot)
+	if app.config.free_build_mode then  -- Everything is 0$ in free build mode
 	  price = "$0"
-	else
-          price = "$" .. map:getParcelPrice(self.hover_plot)
 	end
       else
         owner = world.hospitals[owner_num].name
--- a/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
@@ -305,7 +305,7 @@ function UITownMap:draw(canvas, x, y)
       if owner_num == 0 then
         owner = _S.town_map.for_sale
 	price = app.config.free_build_mode and "$0"  -- Everything is 0$ in free build mode
-  		or ("$" .. map:getParcelPrice(self.hover_plot))
+		or ("$" .. map:getParcelPrice(self.hover_plot))
       else
         owner = world.hospitals[owner_num].name
       end
--- a/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
@@ -188,8 +188,8 @@ function UITownMap:draw(canvas, x, y)
   self.info_font:draw(canvas, radiators,    x +  95, y + 265)
 
   -- Heating costs
-  local heating_costs = app.config.free_build_mode and 0
-      or math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)
+  local heating_costs = app.config.free_build_mode and 0 or
+      math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)
   self.info_font:draw(canvas, ("%8i"):format(heating_costs),  x + 100, y + 355)
 
   -- Draw balance with temporary offset in unicode languages
@@ -304,8 +304,8 @@ function UITownMap:draw(canvas, x, y)
       local owner_num = map.th:getPlotOwner(self.hover_plot)
       if owner_num == 0 then
         owner = _S.town_map.for_sale
-	price = app.config.free_build_mode and "$0"
-		or ("$" .. map:getParcelPrice(self.hover_plot))
+        price = app.config.free_build_mode and "$0" or
+            ("$" .. map:getParcelPrice(self.hover_plot))
       else
         owner = world.hospitals[owner_num].name
       end
--- a/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
@@ -303,7 +303,12 @@ function UITownMap:draw(canvas, x, y)
       local owner_num = map.th:getPlotOwner(self.hover_plot)
       if owner_num == 0 then
         owner = _S.town_map.for_sale
-        price = "$" .. map:getParcelPrice(self.hover_plot)
+		-- Everything is 0$ in free build
+		if app.config.free_build_mode then
+		  price = "$0"
+		else
+          price = "$" .. map:getParcelPrice(self.hover_plot)
+		end
       else
         owner = world.hospitals[owner_num].name
       end
--- a/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
@@ -188,7 +188,7 @@ function UITownMap:draw(canvas, x, y)
   self.info_font:draw(canvas, radiators,    x +  95, y + 265)
 
   -- Heating costs
-  local heating_costs = app.config.free_build_mode and 0  -- Everything is 0$ in free build mode
+  local heating_costs = app.config.free_build_mode and 0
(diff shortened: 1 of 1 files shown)


TheCycoONE:
Seems to me it would be more straight forward to set heating_costs to 0 and keep the display logic untouched.

CorsixTH/Lua/dialogs/fullscreen/town_map.lua
```diff
@@ -303,7 +306,10 @@ function UITownMap:draw(canvas, x, y)
       local owner_num = map.th:getPlotOwner(self.hover_plot)
       if owner_num == 0 then
         owner = _S.town_map.for_sale
-        price = "$" .. map:getParcelPrice(self.hover_plot)
+	price = "$" .. map:getParcelPrice(self.hover_plot)
+	if app.config.free_build_mode then  -- Everything is 0$ in free build mode
+	  price = "$0"
+	end
```
Alberth289346, line 312: As a general rule, try to avoid useless computations. A computer is fast but not infinitely fast, so use its time wisely.

Here you first construct a string with the actual price, which you then completely discard in case the free-build mode is active.

The better tactic is first to split on the case and then do the computation. An if/else or an and/or would work here.

If splitting is not easily possible and you want  something like you have here, first perform the cheaper case, and override that result with the more expensive case if necessary. In that way, if the override is needed you're throwing away less computation effort.

Alberth289346:
To explain the idea of TheCycoONE a bit: If you only change what is displayed, then internally, the game works like a non-free build game. Costs etc accumulate, but the user never sees them because it displays `0` everywhere. 

Since the internal logic however still works with eg the accumulated costs, the game may react "weird" because it's making decisions based on information that is invisible to the user.  
If you instead implement free build as really free build (ie everything has cost 0), such problems are much easier to avoid.

CorsixTH/Lua/dialogs/fullscreen/town_map.lua
Matroftt: Variant 1
```
price = "0$"
if not app.config.free_build_mode then
  price = "$" .. map:getParcelPrice(self.hover_plot) 
end
```
Variant 2
```
if app.config.free_build_mode then
  price = "0$"
else
  price = "$" .. map:getParcelPrice(self.hover_plot) 
end
```

Which variant is better?

Alberth289346: The 2nd variant has the advantage that it very explicit expresses the fact there are 2 independent cases.  
The 1st variant makes that more difficult, as a reader to have to verify that the word `price` above the `if` is the same as the word `price` within the `if`.

Obviously, here it's not very difficult, but image assigning to `something[self:computeIndex(123, 45, 65)][28].field.subfield` instead. Then comparing all the numbers and words is much more difficult to do manually. A typo is easily over-looked as humans are sloppy readers.

Alberth289346: In such a case the and/or approach is better, since you don't have to write the long assigned name twice. Alternatively, use a simple variable like `price` :p  and afterwards copy the value to the long name, ie

``` lua
local price = ....
something[self:computeIndex(123, 45, 65)][28].field.subfield = price
```
One additional line of code, but it avoids writing the assigned field twice.

CorsixTH/Lua/dialogs/fullscreen/town_map.lua
```diff
@@ -188,10 +188,8 @@ function UITownMap:draw(canvas, x, y)
   self.info_font:draw(canvas, radiators,    x +  95, y + 265)
 
   -- Heating costs
-  local heating_costs = math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)
-  if app.config.free_build_mode then  -- Everything is 0$ in free build mode
-    heating_costs = 0
-  end
+  local heating_costs = app.config.free_build_mode and 0  -- Everything is 0$ in free build mode
+			or math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)
```
Alberth289346, line 192: Second line uses TABs rather than spaces.  
Perhaps you should spend some time teaching your editor not to use them, to avoid having to fix it afterwards every time. Alternatively you can look for a more sane editor.  
Making your life more easy is always a good idea :)

Continuation lines (line 192 here) use a double indent (ie 4 spaces) from the left. That is, it should look like
```
local ...
    or ...
```

I am not sure about the comment in the middle of the `and/or` sequence. A reader may think the line is finished at the end of line 191.  Would it make sense to you to merge that information into the comment at line 190?   It's fine to rephrase the text.

The old code didn't have proper spacing around the multiplication operators. You can fix that as well while changing the code (there should be a space at both sides of a `*`).  
The old code has a few too many parentheses, eliminating them can also be done.  
These improvements are optional, but leaving the place a little cleaner than you found it is generally a good idea.

A more complicated change is that `math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)` here is used to display  (and only display) the heating costs. That means that elsewhere in the program the same formula must exist as well to actually compute the costs and deduct it from the hospital balance. Obviously, both spots should have the exact same formula or the displayed cost here is lying to the user.  
In programming, a general objective is to have everything exactly once. In other words, this formula should exist at one spot only, and be used both for displaying and for updating the hospital balance.
In that way the displayed amount is always the same as the deducted amount. That becomes very useful if we ever need to change the formula.  
When you feel up for it, you could try fixing that. Perhaps in a next patch? (it's not quite a "free build improvement in town map", the topic of this patch)

Always feel free to start a discussion in the chat. It's easy to get overwhelmed by all the things, a little guidance what to do can be very useful. We all started that way, and all advice is free :)

CorsixTH/Lua/dialogs/fullscreen/town_map.lua
```diff
@@ -306,10 +304,8 @@ function UITownMap:draw(canvas, x, y)
       local owner_num = map.th:getPlotOwner(self.hover_plot)
       if owner_num == 0 then
         owner = _S.town_map.for_sale
-	price = "$" .. map:getParcelPrice(self.hover_plot)
-	if app.config.free_build_mode then  -- Everything is 0$ in free build mode
-	  price = "$0"
-	end
+	price = app.config.free_build_mode and "$0"  -- Everything is 0$ in free build mode
+  		or ("$" .. map:getParcelPrice(self.hover_plot
```
Alberth289346: This line doesn't work. It has two `(` parentheses` and zero `)` parentheses. They should be balanced.

A good way to find these kind of problems is by running the program after you changed it. It won't start at all with problems like the above. If it does run, it's a good time to check that your change is working correctly.

Except for the parentheses, the line also has a TAB and incorrect indenting.

The comment is here as well. Like above, in the middle of a statement is somewhat a bad spot.  
I don't know how you think about the comment, but it seems to me that it states a quite known fact, and thus doesn't have much value (it does say anything unexpected or special that only applies here) .

If you feel the comment has value enough to keep, I'd suggest to make a new comment above line 307 that should then cover the entire `price = ...` line rather than just the free-build bit.  
If you feel the comment isn't worth keeping, you can delete it.

Alberth289346:
Wrt your TAB fight, I once configured my editor to show TABs:

image

The light-grey `>-------` is a TAB character in the text. It helped a lot in setting up my editor correctly at the time (you can actually see what it's doing :) ). Today it's helping me in finding these problems in files that I get from others.

No doubt you use a different editor, but perhaps it has an option as well to show you the white space. Maybe that is helpful for you as well.

CorsixTH/Lua/dialogs/fullscreen/town_map.lua
Matroftt: I have some issues with loading file into github from notepad++, it breaks spacing so i just copy stuff manually, and it sometimes generate that kind of problems.

About the comment - Sure, I'll delete it now

Matroftt:
Ready to merge i guess

Alberth289346:
> Ready to merge i guess

Not quite, there are still TABs:
image

----
(press LMB and drag over the area, a TAB will highlight like above).

Also, indenting wasn't fixed:

Continuation lines (line 192 here) use a double indent (ie 4 spaces) from the left. That is, it should look like
``` lua
local ...
    or ...
```
that holds for lines 192 and 308

CorsixTH/Lua/dialogs/fullscreen/town_map.lua
```diff
@@ -303,7 +304,8 @@ function UITownMap:draw(canvas, x, y)
       local owner_num = map.th:getPlotOwner(self.hover_plot)
       if owner_num == 0 then
         owner = _S.town_map.for_sale
-        price = "$" .. map:getParcelPrice(self.hover_plot)
+	price = app.config.free_build_mode and "$0"
+		or ("$" .. map:getParcelPrice(self.hover_plot))
       else
```
lewri, line 309:
```suggestion
        price = app.config.free_build_mode and "$0" or
            ("$" .. map:getParcelPrice(self.hover_plot))
      else
```

CorsixTH/Lua/dialogs/fullscreen/town_map.lua
```diff
@@ -188,7 +188,8 @@ function UITownMap:draw(canvas, x, y)
   self.info_font:draw(canvas, radiators,    x +  95, y + 265)
 
   -- Heating costs
-  local heating_costs = math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)
+  local heating_costs = app.config.free_build_mode and 0
+			or math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)
```
lewri, line 192:
```suggestion
      or math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)
```

lewri (changes requested):

CorsixTH/Lua/dialogs/fullscreen/town_map.lua
```diff
@@ -188,7 +188,8 @@ function UITownMap:draw(canvas, x, y)
   self.info_font:draw(canvas, radiators,    x +  95, y + 265)
 
   -- Heating costs
-  local heating_costs = math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)
+  local heating_costs = app.config.free_build_mode and 0
+      or math.floor(((hospital.heating.radiator_heat *10)* radiators)* 7.5)
```
lewri, line 192: Move the `or` up a line

TheCycoONE: Can we do this for him; I hate to discourage a new dev too much with nits.

PR #2626, opened 2024-09-21
Update for 0.68.0 RC2

lewri:
For next RC.

Note the bugfix change `Added a safeguard to prevent a crash of games previously using the old cheat system` isn't in the main changelog because on release it will be fall under the bugfix `Active cheats will now persist across saves` instead.

### DIFF
--- a/CorsixTH/com.corsixth.corsixth.metainfo.xml
+++ b/CorsixTH/com.corsixth.corsixth.metainfo.xml
@@ -20,6 +20,22 @@
   </provides>
   <launchable type="desktop-id">com.corsixth.corsixth.desktop</launchable>
   <releases>
+    <release version="0.68.0~rc2" date="2024-09-25" type="development">
+      <description>
+        <p>Translations:</p>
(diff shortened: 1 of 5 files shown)


TheCycoONE:
You'll squash the version list down for release?

CorsixTH/com.corsixth.corsixth.metainfo.xml
```diff
@@ -20,6 +20,21 @@
...
+        <p>Translations:</p>
+        <ul>
+          <li>Dutch translation has been updated</li>
+        </ul>
+        <p>Bug Fixes:</p>
+        <ul>
+          <li>Added a safeguard to prevent a crash of games previously using the
+              old cheat system</li>
+          <li>Patients waiting for a player decision no longer lose their mood icon
+              on vomit/pee</li>
```
lewri, line 34: Does the slash need to be `&sol;`?

lewri:
Updated for 2621, all done.

PR #2629, opened 2024-09-22
Generalize whitespace check to catch TABs in source files

Alberth289346:
- Renamed and adapted the trailing_whitespace script.
  Eventually pretty much everything was changed. If you read each commit, steps are much less large.
- Fixed Russion and Ukrainan language files.
  If those files are desired to be added before this PR, a cherry-pick of commit 1abd7cf should be cleanly mergeable with this patch.
- Found TABs in generated `SpriteEncoder/{scanner,parser}.cpp` files.
  To fix, I added CLI processing, and they are now listed as excluded patterns in the YML file.


scripts/check_whitespace.py
```diff
@@ -0,0 +1,124 @@
...
+                        offending_files.append(path)
+
+                except UnicodeDecodeError:
+                    print(f"ERROR: File {path} has Unicode errors.")
+                    found_errors = True
+
+    # Report files with bad whitespace.
+    print('Checked {} files'.format(count))
+    if offending_files:
+        print('Found files with bad whitespace:')
```
lewri, line 113: Is there a chance it could also print the line(s) it was found on/what the whitespace issue was?

Alberth289346: It's feasible of course.

I did consider it a bit, but didn't know what to do if you have a lot of these problems. I feel there isn't much point in listing all, for eg the previous version of the two changed language files.

The kind of problem could probably be useful.

lewri: Yeah, Luacheck's copy does output the offending issue and the line:column as far as I remember. At the very least I agree the kind of issue will be very useful to especially someone new to our code.

TheCycoONE: The file is enough to find it, I wouldn't print the lines.

TheCycoONE:
Suggest merging post release.

Alberth289346:
Extended the program to also report what is bad. An example:
```
Checked 7 files

Found files with bad whitespace:
File "~/dev/corsix-th/corsix/scripts/./tab_spc.lua" has both at least one tab character and trailing whitespace
File "~/dev/corsix-th/corsix/scripts/./tab.lua" has at least one tab character
File "~/dev/corsix-th/corsix/scripts/./spc.lua" has trailing whitespace
```

PR #2641, opened 2024-09-26
improved save_to_tmp string

### LINKED ISSUE #2294: save_to_tmp string is too specific
```lua
save_to_tmp = "The file at %s could not be used. The game has been saved to %s. Error: %s",
```

This string is used for when we run into a file saving issue and need to save to a temporary location. However, the string currently references `game`, whereas it could be other files that run into issues. As such below, the context doesn't fit:
image

There are two routes to resolve this:
1. Amend the string to be more generalised **OR**
2. Create a set of contextual based strings based on the kind of file it had issues with (e.g. a save, config file, gamelog file). A supporting function should assist other code in grabbing the right error text (rather than individual strings in each function).

Pulled from #2291

MandrakeHorse:
*Fixes #2294*

This pull request replaces the current `save_to_tmp` string with a more general message. I have reproduced the error by setting my configuration file to read-only and can confirm that the updated string will appear.
2024-09-26 23_22_29-CorsixTH

### DIFF
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -1,4 +1,4 @@
---[[ Copyright (c) 2010-2020 Manuel "Roujin" Wolf, Edvin "Lego3" Linge, lewri et al.
+--[[ Copyright (c) 2010-2024 Manuel "Roujin" Wolf, Edvin "Lego3" Linge, MandrakeHorse, lewri et al.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/languages/english.lua
```diff
@@ -1,4 +1,4 @@
---[[ Copyright (c) 2010-2020 Manuel "Roujin" Wolf, Edvin "Lego3" Linge, lewri et al.
+--[[ Copyright (c) 2010-2024 Manuel "Roujin" Wolf, Edvin "Lego3" Linge, MandrakeHorse, lewri et al.
```
lewri, line 1: Not you but this copyright has taken a unique turn to our other normal formats. I'll try to remember to fix it up in future.

lewri (approved):

TheCycoONE:
This will probably require corrections from the translators.

PR #2646, opened 2024-09-30
0.68.0 Release

lewri:
Prereleases squashed at request.
Hopefully the XML changelog is mirrored correctly. Date set to the 1st if the version plans to be tagged early.


lewri:
@TheCycoONE can we added `This version comes with two experimental features you can help us with by giving feedback on! These will be off by default.` to the release text again at the top?

TheCycoONE:
Yes I'll copy all of the past release notes as far as makes sense.

Issue #2254, opened 2022-09-24
mustPause windows can unpause a user-initiated pause


lewri:
This issue is more a note for myself:

1. If a game is paused by the user, with "build while paused" is on, you can open a mustPause window (e.g. Fax) [normal]
2. However, on exiting the window, the game will unpause itself. As such, it doesn't open the original pause. [unintended behaviour]

Really the game should obey the user's pause. I still need some time to decide how best to manage this though.

TheCycoONE:
I think this is fairly straight forward by separating pause invoked by the user vs pause invoked by a dialog instead of trying to maintain a single state.

lewri:
To be honest I was going to go down the route of making an entire new class for gamespeed. Though it sounds like you see something simpler.

Alberth289346:
Make a list of all game-speed input, and decide how the input must be combined to derive speed of the game (on paper or in a text file). Once you understand that, it is probably easy to see whether a full class is desired.

ARGAMX:
Related #2254

ARGAMX:
When a new machine is purchased to replace the old one, the pause is also removed. (it wasn't obvious for me)

<img width="189" height="202" alt="Image" src="https://github.com/user-attachments/assets/7b2dd785-7423-49b5-98e7-3347b5ddc55d" />

PR #2674, opened 2024-10-16
#2625 Fix: Auto satisfying salary request don't close window

### LINKED ISSUE #2625: [Bug] Automatically satisfying salary request does not close the salary request window.
### Describe the issue

Screenshot 2024-09-20 at 21 51 30

_
Set 'WAGE REQUEST' options to 'GRANT'. When employee requesting salary we get "salary request fax icon". If press the "salary request fax icon" then salary request window will appear. And if we will not take any action in it then salary increase request will be made (grant) automatically. But window will still appear on the screen and there is no way to close it or hide. If we press "rise" button then salary will increase one more time but there no longer any need for this. If we press "dismissal/sacking" button, then employee will be fired.

### Steps to Reproduce

1. Set 'WAGE REQUEST' options to 'GRANT'.
2. In Debug menu check that the option 'DISABLE SALARY RISE' is disabled.
3. Wait for any employee request salary increase.
4. Press the "salary request fax icon" for salary request window appear.
5. Do not take any action in this window.
6. Wait around 2 weeks game time.
7. Salary request will satisfy (grant) automatically.
8. That salary request window still appear on the screen.

[salary_request_situation.sav.zip](https://github.com/user-attachments/files/17083312/salary_request_situation.sav.zip)

### Expected Behaviour

Salary request window should close on automatically resolve.

### CorsixTH Version

0.68.0 Beta 1

### Operating System

MacOS Sonoma 14.5

ARGAMX:
*Fixes #2625*

-
- Fixes the issue: Automatically satisfying salary request does not close the salary request window.
- Also fixed that on Automatically satisfying salary request there was no 'increaseWage' sound.
-

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -182,7 +182,7 @@ function Staff:checkIfWaitedTooLongForRaise()
           staff_rise_window:increaseSalary() -- close window and raise
         else
           self:increaseWage(rise_amount)
-        end  
+        end
       -- else: The staff member is sacked
       else
         if staff_rise_window then
@@ -190,7 +190,7 @@ function Staff:checkIfWaitedTooLongForRaise()
         else
           self:fire()
         end
-      end 
+      end
     end
   end
 end
@@ -593,13 +593,11 @@ function Staff:increaseWage(amount)
   local wage_raised = true
   if self.profile.wage >= max_salary then
     wage_raised = false -- Already at max salary
-  else 
-    local new_wage = self.profile.wage
+  else
+    local new_wage = self.profile.wage + amount
     if self.profile.wage + amount > max_salary then
       -- Maximum salary hit. The staff member will never be unhappy again
       new_wage = max_salary
-    else
-      new_wage = self.profile.wage + amount
     end
     -- Apply new salary
     self.profile.wage = new_wage
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -117,7 +117,7 @@ function Staff:tick()
 
   -- check if we need to use the staff room and go there if so
   self:checkIfNeedRest()
-  -- check if staff has been waiting too long for a raise and fire if so
+  -- check if staff has been waiting too long for a raise
   self:checkIfWaitedTooLongForRaise()
 
   -- Decide whether the staff member should be tiring and tire them
@@ -162,7 +162,6 @@ function Staff:checkIfWaitedTooLongForRaise()
     if is_waiting_time_is_up then
       local rise_windows = self.world.ui:getWindows(UIStaffRise)
       local staff_rise_window = nil
-      local rise_amount = self.profile:getRiseAmount()
       self.quitting_in = nil
       self.hospital:removeMessage(self)
 
@@ -181,6 +180,7 @@ function Staff:checkIfWaitedTooLongForRaise()
         if staff_rise_window then -- if rise window open
           staff_rise_window:increaseSalary() -- close window and raise
         else
+          local rise_amount = self.profile:getRiseAmount()
           self:increaseWage(rise_amount)
         end
       -- else: The staff member is sacked
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -118,7 +118,7 @@ function Staff:tick()
   -- check if we need to use the staff room and go there if so
   self:checkIfNeedRest()
   -- check if staff has been waiting too long for a raise and fire if so
-  self:checkIfWaitedTooLong()
+  self:checkIfWaitedTooLongForRaise()
 
   -- Decide whether the staff member should be tiring and tire them
   if self:isTiring() then
@@ -154,37 +154,43 @@ function Staff:tick()
   self:updateSpeed()
 end
 
-function Staff:checkIfWaitedTooLong()
+function Staff:checkIfWaitedTooLongForRaise()
   if self.quitting_in then
     self.quitting_in = self.quitting_in - 1
-    if self.quitting_in < 0 then
+
+    local is_waiting_time_is_up = self.quitting_in < 0
+    if is_waiting_time_is_up then
       local rise_windows = self.world.ui:getWindows(UIStaffRise)
       local staff_rise_window = nil
+      local rise_amount = self.profile:getRiseAmount()
+      self.quitting_in = nil
+      self.hospital:removeMessage(self)
 
-      -- We go through all "requesting rise" windows open, to see if we need
-      -- to close them when the person is fired.
+      -- We go through all "requesting rise" windows open
+      -- to close one of them if open when request resolved.
       for i = 1, #rise_windows do
         if rise_windows[i].staff == self then
           staff_rise_window = rise_windows[i]
           break
         end
       end
+
       -- If the hospital policy is set to automatically grant wage increases, grant
       -- the requested raise instead of firing the staff member
       if self.hospital.policies.grant_wage_increase then
-        local amount = math.floor(math.max(self.profile.wage * 1.1, (self.profile:getFairWage(self.world) + self.profile.wage) / 2) - self.profile.wage)
-        self.quitting_in = nil
-        self.hospital:removeMessage(self)
-        self:increaseWage(amount)
-        return
-      end
-
+        if staff_rise_window then -- if rise window open
+          staff_rise_window:increaseSalary() -- close window and raise
+        else
+          self:increaseWage(rise_amount)
+        end  
       -- else: The staff member is sacked
-      if staff_rise_window then
-        staff_rise_window:fireStaff()
-        return
-      end
-      self:fire()
+      else
+        if staff_rise_window then
+          staff_rise_window:fireStaff() -- close window and fire
+        else
+          self:fire()
+        end
+      end 
     end
   end
 end
@@ -571,7 +577,7 @@ function Staff:requestRaise()
     end
     -- The staff member can now successfully ask for a raise
     self.hospital:makeRaiseRequest(amount, self)
-    self.quitting_in = 25*30 -- Time until the staff members quits anyway
+    self.quitting_in = 25*30 + math.random(50) -- Time until the staff members quits anyway
     self:setMood("pay_rise", "activate")
   end
 end
@@ -587,12 +593,17 @@ function Staff:increaseWage(amount)
   local wage_raised = true
   if self.profile.wage >= max_salary then
     wage_raised = false -- Already at max salary
-  elseif self.profile.wage + amount > max_salary then
-    -- Maximum salary hit. The staff member will never be unhappy again
-    self.profile.wage = max_salary
-  else
+  else 
+    local new_wage = self.profile.wage
+    if self.profile.wage + amount > max_salary then
+      -- Maximum salary hit. The staff member will never be unhappy again
+      new_wage = max_salary
+    else
+      new_wage = self.profile.wage + amount
+    end
     -- Apply new salary
-    self.profile.wage = self.profile.wage + amount
+    self.profile.wage = new_wage
+    self.world.ui:playSound("bonusal2.wav")
   end
   -- Reset
   self:setMood("pay_rise", "deactivate")
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -155,41 +155,40 @@ function Staff:tick()
 end
 
 function Staff:checkIfWaitedTooLongForRaise()
-  if self.quitting_in then
-    self.quitting_in = self.quitting_in - 1
+  if self.quitting_in >= 0 then return end
+  self.quitting_in = self.quitting_in - 1
 
-    local is_waiting_time_is_up = self.quitting_in < 0
-    if is_waiting_time_is_up then
-      local rise_windows = self.world.ui:getWindows(UIStaffRise)
-      local staff_rise_window = nil
-      self.quitting_in = nil
-      self.hospital:removeMessage(self)
+  local is_waiting_time_is_up = self.quitting_in < 0
+  if is_waiting_time_is_up then
+    local rise_windows = self.world.ui:getWindows(UIStaffRise)
+    local staff_rise_window = nil
+    self.quitting_in = nil
+    self.hospital:removeMessage(self)
 
-      -- We go through all "requesting rise" windows open
-      -- to close one of them if open when request resolved.
-      for i = 1, #rise_windows do
-        if rise_windows[i].staff == self then
-          staff_rise_window = rise_windows[i]
-          break
-        end
+    -- We go through all "requesting rise" windows open
+    -- to close one of them if open when request resolved.
+    for i = 1, #rise_windows do
+      if rise_windows[i].staff == self then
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -154,37 +154,43 @@ function Staff:tick()
...
+function Staff:checkIfWaitedTooLongForRaise()
   if self.quitting_in then
     self.quitting_in = self.quitting_in - 1
-    if self.quitting_in < 0 then
+
+    local is_waiting_time_is_up = self.quitting_in < 0
+    if is_waiting_time_is_up then
       local rise_windows = self.world.ui:getWindows(UIStaffRise)
       local staff_rise_window = nil
+      local rise_amount = self.profile:getRiseAmount()
```
lewri, line 165: Calculate this instead at the time you need it (after line 180)

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -154,37 +154,43 @@ function Staff:tick()
...
 end
 
-function Staff:checkIfWaitedTooLong()
+function Staff:checkIfWaitedTooLongForRaise()
   if self.quitting_in then
     self.quitting_in = self.quitting_in - 1
-    if self.quitting_in < 0 then
+
+    local is_waiting_time_is_up = self.quitting_in < 0
+    if is_waiting_time_is_up then
```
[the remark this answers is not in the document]
lewri: We can also return early to remove one nesting.
```lua
if not self.quitting_in < 0 then return end
--[[or]]
if self.quitting_in >= 0 then return end
```

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -587,12 +593,15 @@ function Staff:increaseWage(amount)
...
   else
+    local new_wage = self.profile.wage + amount
+    if self.profile.wage + amount > max_salary then
+      -- Maximum salary hit. The staff member will never be unhappy again
+      new_wage = max_salary
+    end
     -- Apply new salary
-    self.profile.wage = self.profile.wage + amount
+    self.profile.wage = new_wage
+    self.world.ui:playSound("bonusal2.wav")
```
lewri, line 604: Potentially, we may not want to sound this if a wage request is auto granted. Thoughts?

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -118,7 +118,7 @@ function Staff:tick()
   -- check if we need to use the staff room and go there if so
   self:checkIfNeedRest()
   -- check if staff has been waiting too long for a raise and fire if so
```
lewri, line 120: Not you, but this comment is inaccurate (not expecting you to address it)

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -154,37 +154,43 @@ function Staff:tick()
...
     self.quitting_in = self.quitting_in - 1
-    if self.quitting_in < 0 then
+
+    local is_waiting_time_is_up = self.quitting_in < 0
+    if is_waiting_time_is_up then
       local rise_windows = self.world.ui:getWindows(UIStaffRise)
       local staff_rise_window = nil
+      local rise_amount = self.profile:getRiseAmount()
+      self.quitting_in = nil
+      self.hospital:removeMessage(self)
```
lewri, line 167: Should be processed at the very end (you can also remove it from `Staff:fire()` if you do this, I think.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -154,37 +154,43 @@ function Staff:tick()
...
-      if staff_rise_window then
-        staff_rise_window:fireStaff()
-        return
+      else
+        if staff_rise_window then
+          staff_rise_window:fireStaff() -- close window and fire
+        else
+          self:fire()
+        end
       end
```
lewri, line 193: Having to change behaviour based on the window being open is bothersome.

I think this requires a refactor across `UIStaffRise` and `Staff:checkIfWaitedTooLongForRaise()`. Ideally, the code here should be able to handle both scenarios completely itself, closing the window at the end if it exists.

1. Identify if a window exists and remember it
2. Setup the parts to be always cleared `self.message_callback = nil` and `self.quitting_in = nil`
3. Run the raise/fire determination and payment (the auto wage grant request)
4. Close the window if one was found `staff_rise_window:close()`

Another point to bear in mind is that `UIStaffRise` stores a raise amount, but here we're disregarding that and calculating that again. It may be beneficial instead to store the rise_amount in `Staff` instead. This would require an `afterLoad` in `UIStaffRise` though akin to
```lua
if old < xxx
  self.staff.rise_amount = self.rise_amount
end
```
Alternatively, we can just ask the queued message (but would require more code to go find it if it's not open).

lewri (commented):
I fully appreciate providing a working fix here, but I do think this will be a good one we can work on doing better :)

Sidenote: Generally speaking, `UIStaffRise` is a `mustPause` window so should be pausing the game anyway (nonetheless, #2254 exists). It's better to safeguard the erroneous behaviour than not.

CorsixTH/Lua/entities/humanoids/staff.lua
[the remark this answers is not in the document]
ARGAMX: > The local isn't needed here imo, `self.quitting_in < 0` is self-explanatory. I wonder if it should be `<` or `<=`

It was probably pretty clear, but I thought it could be made even clearer.
`is_waiting_time_is_up` is super obvious. 
I think used `<` and not `<=` because it want to stop on `0` which was when we just entered this function call. It seems more logical than stopping at 1.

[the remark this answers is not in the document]
ARGAMX: > We can also return early to remove one nesting.
> 
> ```lua
> if not self.quitting_in < 0 then return end
> --[[or]]
> if self.quitting_in >= 0 then return end
> ```

Sorry, didn't get that.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -571,7 +577,7 @@ function Staff:requestRaise()
     end
     -- The staff member can now successfully ask for a raise
     self.hospital:makeRaiseRequest(amount, self)
-    self.quitting_in = 25*30 -- Time until the staff members quits anyway
+    self.quitting_in = 25*30 + math.random(50) -- Time until the staff members quits anyway
```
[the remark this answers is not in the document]
ARGAMX: > I'm not against randomness (I adore it) but is it needed here?

May be. But while testing sometimes a group of several doctors would resign together in one tick. It looks very strange. I don't see how %6 random here makes this worse. 🙂

CorsixTH/Lua/entities/humanoids/staff.lua
ARGAMX: > Calculate this instead at the time you need it (after line 180)

Agree. Moved.

CorsixTH/Lua/entities/humanoids/staff.lua
ARGAMX: > Potentially, we may not want to sound this if a wage request is auto granted. Thoughts?

Sorry, I don't see any reason why we could ever want to silent this action. 
It was more looks like a bug and lack of consistency:

1. Take for example fire. playSound("sack.wav") is placed inside Staff:fire(). And looks like it doesn't consider the potential to not sound. 
2. Also It looks illogical that the sound of dismissal is inside the Staff functions and granting rise is not.  When calling Staff:fire() you need to think about sound. And when calling Staff:increaseWage() you need to think about sound. Looks like a lack of consistency.
3. So everywhere in the code when there were calls of increaseWage() there was a call of playSound("bonusal2.wav"). (Except that 'checkIfWaitedTooLongForRaise'). Now ("sack.wav") and ("bonusal2.wav") calls for staff only in one place.
4. I never saw in the original game both things without sound.

So, taking these arguments into account the idea of granting raises can happen without sound sounds strange. At least now. If we will need it for some reason, I believe it resolvable in both cases.

CorsixTH/Lua/entities/humanoids/staff.lua
ARGAMX: > Not you, but this comment is inaccurate (not expecting you to address it)

Agree. Fixed.

CorsixTH/Lua/entities/humanoids/staff.lua
ARGAMX: > Should be processed at the very end (you can also remove it from `Staff:fire()` if you do this, I think. Edit: Ah no you can't remove it from `Staff:fire()` as you can fire in multiple ways.

Yeap.

CorsixTH/Lua/entities/humanoids/staff.lua
ARGAMX: I totally agree with you that the change behaviour based on the window being open is bothersome. But it was there before me and when I tried to refactor that I come to that refactoring all this will potentially bring more bugs than fix them. And I decided just to stop on this. 

So I would like to leave refactoring out of this task by now. 
Although in general I like your thoughts.

CorsixTH/Lua/entities/humanoids/staff.lua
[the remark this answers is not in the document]
lewri: By flipping your if condition, you can exit the function early because we no longer need to do anything more.

CorsixTH/Lua/staff_profile.lua
```diff
@@ -251,6 +251,11 @@ function StaffProfile:getFairWage()
   return math.max(math.floor(wage), level_config.staff[conf_id[self.humanoid_class]].MinSalary)
 end
 
+function StaffProfile:getRiseAmount()
+  local rise_amount = math.floor(math.max(self.wage * 1.1, (self:getFairWage(self.world) + self.wage) / 2) - self.wage)
+  return rise_amount
```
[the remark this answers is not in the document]
ARGAMX: > Use return only

Ok. Thanks. I will stick to this.

CorsixTH/Lua/entities/humanoids/staff.lua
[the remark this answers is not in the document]
ARGAMX: > By flipping your if condition, you can exit the function early because we no longer need to do anything more.

Oh, yeah, I got it now. Thanks. It was not mine, it was there before. But yes, I agree. Changed nesting level.

CorsixTH/Lua/entities/humanoids/staff.lua
lewri: I can sort the refactor after this pr, but might have to tweak some parts in this patch.

ARGAMX: > I can sort the refactor after this pr, but might have to tweak some parts in this patch.

Sounds great.

CorsixTH/Lua/entities/humanoids/staff.lua
lewri: I agree that it's slightly illogical to sound one and not the other. My thought path was if you turned auto-grant wage requests on you wouldn't really care when one is given out. However, I'm fine with either unless there's opinion elsewhere.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -154,37 +154,42 @@ function Staff:tick()
...
-        return
+  local is_waiting_time_is_up = self.quitting_in < 0
+  if is_waiting_time_is_up then
+    local rise_windows = self.world.ui:getWindows(UIStaffRise)
+    local staff_rise_window = nil
+    self.quitting_in = nil
+    self.hospital:removeMessage(self)
+
+    -- We go through all "requesting rise" windows open
+    -- to close one of them if open when request resolved.
```
lewri, line 169: How about `Check if any open "requesting rise" windows refer to this staff member.`

We comment later down the function what we're going to do with the window, so we don't need intent stated twice.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -154,37 +154,42 @@ function Staff:tick()
...
+    -- the requested raise instead of firing the staff member
+    if self.hospital.policies.grant_wage_increase then
+      if staff_rise_window then -- if rise window open
+        staff_rise_window:increaseSalary() -- close window and raise
+      else
+        local rise_amount = self.profile:getRiseAmount()
+        self:increaseWage(rise_amount)
+      end
+    -- else: The staff member is sacked
+    else
```
lewri, line 187:
```suggestion
    else -- The staff member is sacked
```
Heh, I added this when the else was implicit, now it's not.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -154,37 +154,42 @@ function Staff:tick()
...
-        self.quitting_in = nil
-        self.hospital:removeMessage(self)
-        self:increaseWage(amount)
-        return
+  local is_waiting_time_is_up = self.quitting_in < 0
+  if is_waiting_time_is_up then
+    local rise_windows = self.world.ui:getWindows(UIStaffRise)
+    local staff_rise_window = nil
+    self.quitting_in = nil
+    self.hospital:removeMessage(self)
```
lewri, line 166: This line specifically should still be moved down to the very very bottom of the function. There is currently no harm running this function call more than once but we shouldn't give instructions to destroy the window while we _may_ need it if it's open.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -154,37 +154,42 @@ function Staff:tick()
...
 
-function Staff:checkIfWaitedTooLong()
-  if self.quitting_in then
-    self.quitting_in = self.quitting_in - 1
-    if self.quitting_in < 0 then
-      local rise_windows = self.world.ui:getWindows(UIStaffRise)
-      local staff_rise_window = nil
+function Staff:checkIfWaitedTooLongForRaise()
+  if self.quitting_in >= 0 then return end
+  self.quitting_in = self.quitting_in - 1
```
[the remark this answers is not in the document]
ARGAMX: Didn't get the idea here.
But If you mean to swap them

Then
The thing is that `:checkIfWaitedTooLong` is called every tick

`function Staff:tick() 
...
self:checkIfWaitedTooLong()
...
end`

So
`if self.quitting_in >= 0`
Plays here the guard/gate/constraint role for internal block in this flow.

There no any guard/gate/constraint in `:tick()` for this flow.

And if decrement will be outside of this internal block, then that mean it will works every tick.

But I think we don't want it.

We want decrement (timer) only work when deadline is setup.

I think we've messed something up here. 😅

And we should back 
  `if self.quitting_in then`
As a guard/gate/constraint.
Not
  `self.quitting_in >= 0`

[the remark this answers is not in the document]
lewri: Apologies. To put it more simply.

Line 158 runs a check if `quitting_in` is over 0 we exit early (nothing else happens in this function).
Line 159 is the decrement of `quitting_in` by 1.

In essence, by having 158 first, `self.quitting_in` never decreases, because your if statement will always be true.

CorsixTH/Lua/entities/humanoids/staff.lua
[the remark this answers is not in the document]
ARGAMX: > Maybe `math.random(0, 50)` if you want the bound to start from no random uptick (but keeping as is is cool too).

Ok. Applied.

lewri:
To be refactored post merge.

Issue #2675, opened 2024-10-17
[Bug] CorsixTH Campaign Map Four Corners can't be cleared despite fulfilling winning conditions


xyressa:
### Describe the issue

I was addicted enough to try out the custom ChrizmanTV campaign that came with the Corsix download, and currently am stuck on the Four Corners level. The briefing instructions at the start of the game says I need (1) Cure 600 people (2) Treat 75% of them (3) Have $200K (4) Value worth $300,000 and (5) 750 Reputation to win. But in the attached 'Charts' screenshots, I have clearly fulfilled all the win conditions including one new one that says to have killed <25%, but the level still hasn't cleared for me.

Is there a bug with this?

(I definitely cleared the value one as I bought a new plot and built a gigantic psych room with like, 100 skeletons.)

![FourCorners_map](https://github.com/user-attachments/assets/d65bee57-1c55-444c-b459-ec99a67c55ec)
![FourCorners_charts](https://github.com/user-attachments/assets/97801985-ddd2-42f3-8604-63cc8bbc15a3)
![FourCorners_briefing](https://github.com/user-attachments/assets/81245102-e067-4390-9c4d-b0a574091dbd)

### Steps to Reproduce

Fulfilled the winning conditions for the level 'Four Corners' in the ChizmanTV custom campaign but the stage does not clear

### Expected Behaviour

Level should auto-clear as the winning conditions are met.

### Have you enabled any current CorsixTH Experimental Features?

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

Save game attached to report: Yes
[FourCorners.zip](https://github.com/user-attachments/files/17406975/FourCorners.zip)

### CorsixTH Version

v0.68.0 RC 2

### Operating System

Windows 10 Home

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml
Gamelog generated on 2024-10-05 14:31:49
Compiled with api_version: 2687, arch: x64, audio: true, os: windows, update_check: true
SDL renderer: direct3d
Lua 5.4 run with api version: 2687, game version: v0.68.0 RC 2, savegame version: 193
Created game with savegame version 193.
Warning: Trying to resolve door deadlock at (65, 90)
Error in timer handler: 
Lua\calls_dispatcher.lua:436: call to be executed is still assigned
stack traceback:
	[C]: in function 'assert'
	Lua\calls_dispatcher.lua:436: in method 'executeCall'
[...]

PR #2676, opened 2024-10-17
Improvements for ChrizmanTV Campaign Levels

Matroftt:
* Relates to #2675 

- Makes HospitalValue criteria less for specific levels
- Some levels warn player about 10% or more death percentage, not 0%
- Fixes Upper Pukington flags


lewri:
Your edits to the level file appear to have suspiciously affected every line with whitespace changes of some kind. Are you able to check that?

Matroftt:
> Your edits to the level file appear to have suspiciously affected every line with whitespace changes of some kind. Are you able to check that?

Yes, they are showing as new for some reason, even if i paste unchanged .level files (royalhospital.level is an example)
I thought that it was a problem of encoding - but no, I tried unicode, utf8, ansi, iso-8859-1 but they all are still showing new strings in github
Sorry for creating commit mess, i made a big mistake while testing encodings

lewri:
In that case it's probably fine. I'll leave it to @TheCycoONE just to double check (could well be some strange github bug)

TheCycoONE:
No they're actually different. Will need to inspect.

lewri:
> No they're actually different. Will need to inspect.

Git bash isn't giving me many clues, it believes they could be the same:
```yaml
$ file dethsdaw.level "dethsdaw (2).level"
dethsdaw.level:     ASCII text, with very long lines (456), with CRLF line terminators [original file]
dethsdaw (2).level: ASCII text, with very long lines (456), with CRLF line terminators [PR file]
```

ARGAMX:
> Levels are still Line ending broken. @TheCycoONE are you able to push a fixed commit via your git so we can merge?

I will help to @Matroftt with this.

tobylane:
Sorry, that was meant to be a fix but it wasn't so I rolled back.

tobylane:
Fixed the whitespace, and a strange apostrophe in wishinwell.level. The original changes, minus whitespace, are here https://github.com/CorsixTH/CorsixTH/compare/master...Matroftt:fb3e34c93fb75c711e892f4f6b68871c529c5ef3?w=1

Matroftt:
Thanks Toby

PR #2677, opened 2024-10-17
#2627 fix: 'free_build_mode' param affecting campaign maps.

### LINKED ISSUE #2627: [Bug] 'free_build_mode' param affecting campaign maps.
### Describe the issue

'free_build_mode' apply to any of the campaign maps. But it should not.

'config.txt' says, that 'free_build_mode' does not apply to any of the campaign maps.

Screenshot 2024-09-21 at 22 47 53

Also it looks very strange for me that enabling option 'Free Build' on 'Single Scenario' menu affecting anything outside of that screen. As keeping 'free_build_mode' in 'config.txt' too.
It seems more reasonable to me to select this value only at level start and not keeping it in game configs. So it can only affect this Single Scenario start.
Or maybe there is some meaning to keep 'free_build_mode' in 'config.txt' which I don't understand?

### Steps to Reproduce

1. Run game. 
2. In Main Menu select 'Single Scenario'.
3. Enable option 'Free Build'.
4. Press 'Back' to back to main menu.
5. In Main Menu select 'Custom Campaign'.
6. Select any campaign, press 'Start'.
7. Observe that you have $0 at the start and you in 'Free Build' mode.

### Expected Behaviour

1. 'free_build_mode' should not apply to any of the campaign maps.

### CorsixTH Version

0.68.0 Beta 1

### Operating System

MacOS Sonoma 14.5

ARGAMX:
*Fixes #2627*

-
- Only single fix for this:
- `'free_build_mode' should not apply to any of the campaign maps.`
- without refactoring and extracting 'free_build_mode' from config.txt
- which will be good to implement, but for now only bugfix.
-

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -744,15 +744,12 @@ function App:_loadLevel(level, difficulty, level_name, level_file, level_intro, 
   self.map:setBlocks(self.gfx:loadSpriteTable("Data", "VBlk-0"))
   self.map:setDebugFont(self.gfx:loadFont("QData", "Font01V"))
 
-  -- In Free Build mode?
-  local free_build_mode
-  local isMainCampaign = tonumber(new_map.level_number)
-  local isCustomCampaign = campaign_info ~= nil
-  if isMainCampaign or isCustomCampaign then
-    free_build_mode = false -- disable free_build_mode for any Campaign
-  else
-    free_build_mode = self.config.free_build_mode
+  local function determineFreeBuildMode()
+    local is_main_campaign = tonumber(new_map.level_number)
+    local is_custom_campaign = campaign_info ~= nil
+    return self.config.free_build_mode and not (is_main_campaign or is_custom_campaign)
   end
+  local free_build_mode = determineFreeBuildMode()
 
   -- Load world
   self.world = World(self, free_build_mode)
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -716,7 +716,7 @@ end
 --! Loads the specified level. If a string is passed it looks for the file with the same name
 -- in the "Levels" folder of CorsixTH, if it is a number it tries to load that level from
 -- the original game.
---!param campaign_info (object) is information about campaign in case loading level 
+--!param campaign_info (object) is information about campaign in case loading level
 -- is a part of a campaign. if not part of a campaign then campaign_info is nil.
 function App:_loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor, campaign_info)
   if self.world then
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -749,10 +749,9 @@ function App:_loadLevel(level, difficulty, level_name, level_file, level_intro, 
     local is_custom_campaign = campaign_info ~= nil
     return self.config.free_build_mode and not (is_main_campaign or is_custom_campaign)
   end
-  local free_build_mode = determineFreeBuildMode()
 
   -- Load world
-  self.world = World(self, free_build_mode)
+  self.world = World(self, determineFreeBuildMode())
   self.world:createMapObjects(map_objects)
 
   -- Enable / disable SoundEffects
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -716,6 +716,8 @@ end
 --! Loads the specified level. If a string is passed it looks for the file with the same name
 -- in the "Levels" folder of CorsixTH, if it is a number it tries to load that level from
 -- the original game.
+--!param campaign_info (object) is information about campaign in case loading level 
+-- is a part of a campaign. if not part of a campaign then campaign_info is nil.
 function App:_loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor, campaign_info)
   if self.world then
     self:worldExited()
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -699,6 +699,8 @@ end
 --! This function should always be called to catch errors and properly pass the
 --! error to the player
 --!param error_prefix (string) (Optional) Prefixes the error relevant to what was loaded
+--!param campaign_info (object) (Optional) is information about campaign in case loading level
+-- is a part of a campaign. if not part of a campaign then campaign_info is nil.
 --! return (boolean) The outcome of the pcall
 function App:loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor, error_prefix, campaign_info)
   local status, err = pcall(self._loadLevel, self, level, difficulty, level_name,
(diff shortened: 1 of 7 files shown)


TheCycoONE:
Not sure I agree. Was there a discussion on this?

lewri:
The issue here trying to be fixed in #2627 is that you can set free build on a custom campaign. However, you may (sometimes can't) advance through the campaign when it is on. I provided some options in https://github.com/CorsixTH/CorsixTH/issues/2627#issuecomment-2376527748 without giving any attachment to what one could be the most reasonable.

ARGAMX:
Thanks @lewri  

I believe that 'free_build_mode' cannot be applied to campaign maps:

1. 'config.txt' clearly says it shouldn't.
2. there was checks in code on that. even for custom campaigns. but the problem that they was executing not in a proper moment. so code looks like it shouldn't too. but it was work in other way due to bug.
3. I don't understand how player can progress thourg campaing with 'free_build_mode' enabled. it literally prevents it.
4. in UI there no any sign that custom campaign can run in 'free_build_mode'. it is only custom scenario option according to UI. 

So this fix only pretend to be a bugfix, not improvements or enhancement.

CorsixTH/Lua/world.lua
```diff
@@ -1227,7 +1221,8 @@ function World:winGame(player_no)
     local bonus_rate = math.random(4,9)
     local with_bonus = self.ui.hospital.cheated and 0 or (self.ui.hospital.player_salary * bonus_rate) / 100
     self.ui.hospital.salary_offer = math.floor(self.ui.hospital.player_salary + with_bonus)
-    if type(self.map.level_number) == "number" or self.campaign_info then
+    local isCampaign = type(self.map.level_number) == "number" or self.campaign_info
+    if isCampaign then
```
lewri, line 1225: I'd revert this. Or make a new function for it before `World:getCampaignData()`

```lua
--! Returns whether the level being played is part of a campaign or not
function World:isCampaign()
  return type(self.map.level_number) == "number" or self.campaign_info
end
```

Then
```suggestion
    if self:isCampaign() then

**Review by lewri — COMMENTED** [review]
There's potential here that to patch this properly it may be considered larger than just a bugfix. So even though I've provided some review comments it could be the fixing PR's scope may need widening. Or may need to go back to the issue itself to determine a more detailed approach.

I'm satisfied enough though that this would at least align with the UI presented one setting up a game.

**Comment by ARGAMX** [comment]
> There's potential here that to patch this properly it may be considered larger than just a bugfix. 

Totally agree.

File: CorsixTH/Lua/app.lua [hunk 2]
```diff
@@ -744,8 +744,18 @@ function App:_loadLevel(level, difficulty, level_name, level_file, level_intro,
...
 
+  -- In Free Build mode?
+  local free_build_mode
+  local isMainCampaign = tonumber(new_map.level_number)
+  local isCustomCampaign = campaign_info ~= nil
+  if isMainCampaign or isCustomCampaign then
+    free_build_mode = false -- disable free_build_mode for any Campaign
+  else
+    free_build_mode = self.config.free_build_mode
+  end
```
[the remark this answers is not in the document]
ARGAMX: > This could fit nicer in a local function.

I agree. 

> NB: locals almost always use lower snake case; except constants (and then then sometimes we still don't lol).

Oh, true, my miss.

CorsixTH/Lua/world.lua
ARGAMX: > Or make a new function for it before `World:getCampaignData()`

Thanks. Applied.

CorsixTH/Lua/app.lua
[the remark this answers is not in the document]
ARGAMX: > This could fit nicer in a local function.

Thanks. Applied.

CorsixTH/Lua/app.lua
```diff
@@ -744,8 +744,15 @@ function App:_loadLevel(level, difficulty, level_name, level_file, level_intro,
...
+  local function determineFreeBuildMode()
+    local is_main_campaign = tonumber(new_map.level_number)
+    local is_custom_campaign = campaign_info ~= nil
+    return self.config.free_build_mode and not (is_main_campaign or is_custom_campaign)
+  end
+  local free_build_mode = determineFreeBuildMode()
+
   -- Load world
-  self.world = World(self)
+  self.world = World(self, free_build_mode)
```
lewri, line 755: Make the function call directly here to be cleaner, as I made sure the function name was explanatory.

lewri (commented):
@TheCycoONE please advise on whether this patch with the discussions around the pr and the issue would be favourable to apply? In a sense behaviour is now logical to the custom campaign UI's layout (plus original campaign can never be free build).

--

The only element I could lastly suggest if this could be unexpected to players (mainly as a suggestion to Cyco to potentially filling the gap) is to add a string element somewhere in `Custom Campaign`'s UI to let them know Free Build was disabled. E.g.
`This campaign will be started without free build enabled`
You could go one step further potentially, only showing this string if `free_build` is true in the config.txt

CorsixTH/Lua/app.lua
ARGAMX: > Make the function call directly here to be cleaner, as I made sure the function name was explanatory.

Ok. Applied.

lewri:
> Guys, what's stopping us from merging this? 🙂

Likely the below. I won't be the one to merge this.

> @TheCycoONE please advise on whether this patch with the discussions around the pr and the issue would be favourable to apply? In a sense behaviour is now logical to the custom campaign UI's layout (plus original campaign can never be free build).

I've left it to him to provide input on this. I also gave him a suggested resolution to fulfil reservations.

CorsixTH/Lua/app.lua
```diff
@@ -716,7 +716,7 @@ end
 --! Loads the specified level. If a string is passed it looks for the file with the same name
 -- in the "Levels" folder of CorsixTH, if it is a number it tries to load that level from
 -- the original game.
-function App:_loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor)
+function App:_loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor, campaign_info)
```
TheCycoONE, line 719: Can we add some parameter documentation.

IMO the number of parameters here is getting silly.

ARGAMX: I agree that number of parameters here is getting silly. 💯

I have added campaign_info parameter documentation here.

About other params I can't tell more that their names says.

lewri: The doc will be a mirror of `function:App:loadLevel(...)` bar the error paramter. I'm not sure it requires documenting twice but c'est la vie :)

CorsixTH/Lua/app.lua
```diff
@@ -700,9 +700,9 @@ end
 --! error to the player
 --!param error_prefix (string) (Optional) Prefixes the error relevant to what was loaded
 --! return (boolean) The outcome of the pcall
-function App:loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor, error_prefix)
+function App:loadLevel(level, difficulty, level_name, level_file, level_intro, map_editor, error_prefix, campaign_info)
```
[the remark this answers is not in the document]
lewri: Actually... is it optional?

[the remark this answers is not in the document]
ARGAMX: True. 👍
Yeap, optional.
I have added `campaign_info` parameter documentation here.

PR #2686, opened 2024-10-19
Rename canPeePukeOrFall

Alberth289346:
*Fixes a sub-optimal function name.

- Functionally it does nothing.
- It renames the `Patient:canPeePukeOrFall` method to `Patient:atFullyEmptyTile`
- It add some lua docs and converts some comment to lua docs.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -158,7 +158,7 @@ function Patient:setDiagnosed()
   self:updateDynamicInfo()
 end
 
--- Modifies the diagnosis progress of a patient.
+--! Modifies the diagnosis progress of a patient.
 -- incrementValue can be either positive or negative.
 function Patient:modifyDiagnosisProgress(incrementValue)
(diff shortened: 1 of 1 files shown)


lewri (approved):
Function name indeed better :)
It does make me question though if this check should belong in patient, but that could well be a different discussion.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -204,7 +204,7 @@ function Patient:completeDiagnosticStep(room)
   self:modifyDiagnosisProgress(diagnosis_base + (diagnosis_bonus * multiplier))
 end
 
--- Sets the hospital for the patient - additionally removing them from a
+--! Sets the hospital for the patient - additionally removing them from a
 -- hospital if they already belong to one. For player hospitals, patients who
 -- are not debug or emergency patients are made to seek a reception desk.
```
lewri, line 209: I think every line needs a `!`? I was never too sure though.

lewri: Nope, I'm wrong, sorry :)

PR #2687, opened 2024-10-19
Increase buy object limit

Matroftt:
- Increases buy object limit from 99 to 100, to match original TH behavior

### DIFF
--- a/CorsixTH/Lua/dialogs/furnish_corridor.lua
+++ b/CorsixTH/Lua/dialogs/furnish_corridor.lua
@@ -136,8 +136,8 @@ function UIFurnishCorridor:purchaseItem(index, quantity)
   quantity = quantity + o.qty
   if quantity < o.min_qty then
     quantity = o.min_qty
-  elseif quantity > 99 then
-    quantity = 99
+  elseif quantity > 100 then
+    quantity = 100
   end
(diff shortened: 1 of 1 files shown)


lewri:
I'm not sure if it was 100? In TH we stop being able to purchase over 99
image

However, we can support it easy so all good

TheCycoONE:
Want to revert then. Where did 100 come from @Matroftt

MandrakeHorse:
I can confirm that Theme Hospital prevents you from buying more than 99 of the same item at once. We could theoretically increase the upper limit in CorsixTH (perhaps even to 999), but aside from the alignment issue mentioned above, I'm not sure if there's a valid use case for doing so. Most players aren't going to buy more than 99 of the same item unless they're messing around or trying to reach the level's Hospital Value goal.

Incidentally, when I checked the behaviour in CorsixTH 0.68, I noticed that if the player tries to exceed the maximum number of objects, the Adviser appears and incorrectly tells you that "you don't have enough money in the bank to make that purchase". I'll create a separate issue for this bug.

Matroftt:
Weird, i remember it being 100. Have to check stuff next time, sorry about that

MandrakeHorse:
Reverted by #2692

PR #2688, opened 2024-10-20
Smooth out gameplay at normal speed

### LINKED ISSUE #2679: [Bug] 'Normal' game speed looks choppy
### Describe the issue

'Normal' game speed looks choppy and 'Maximum speed' looks like actual 'Normal' speed.

Windows 10 x64. 
GoG game files.
Nvidia 3080 gpu
AMD x5900 cpu
16 gb ram.

### Steps to Reproduce

1.  Compare between those speeds.

https://github.com/user-attachments/assets/9347d5a7-eb6a-47b7-99df-d8ad31f7570a

### Expected Behaviour

Normal speed should at least look like "maximum speed" and I suppose "maximum speed" should be faster.

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

### CorsixTH Version

0.68

### Operating System

Windows 10 x64

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml
Gamelog generated on 2023-11-11 00:59:47
Compiled with api_version: 2684, arch: x64, audio: true, os: windows
SDL renderer: direct3d
Lua 5.4 run with api version: 2684, game version: v0.67, savegame version: 180
```

### Additional Information

This is all it has inside the gamelog.txt, and seems old due to the date it has.
No tampering done and it's the only gamelog.txt that exists.

TheCycoONE:
Double the number of tick events so that we can do one movement per tick instead of two at normal speed.

Fixes #2679 but may have side effects. Please play test.

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -67,7 +67,7 @@ function World:World(app)
   self.entity_map = EntityMap(self.map)
 
   -- Time
-  self.hours_per_tick = 2
+  self.hours_per_tick = 1
   self.tick_rate = 3
   self.tick_timer = 0
   self.game_date = Date() -- Current date in the game.
@@ -718,12 +718,12 @@ end
 -- happens.
 local tick_rates = {
   ["Pause"]              = {0, 1},
-  ["Slowest"]            = {1, 28}, -- 1 tick per s
-  ["Slower"]             = {1, 6}, -- 1 tick per 216ms
-  ["Normal"]             = {2, 3}, -- 2 ticks per 108ms = 1 tick per 54 ms
-  ["Max speed"]          = {1, 1}, -- 1 tick per 36 ms
-  ["And then some more"] = {2, 1}, -- 2 ticks per 36 ms = 1 tick 18ms
-  ["Speed Up"]           = {4, 1},
+  ["Slowest"]            = {1, 56}, -- 1 hour per s
+  ["Slower"]             = {1, 12}, -- 1 hour per 216ms
+  ["Normal"]             = {1, 3}, -- 1 hour per 54 ms
+  ["Max speed"]          = {1, 2}, -- 1 hour per 36 ms
+  ["And then some more"] = {1, 1}, -- 1 hour per 18ms
+  ["Speed Up"]           = {8, 1},
 }
 
 function World:speedUp()
@@ -837,8 +837,8 @@ local outside_temperatures = {
 
 local start_date = Date()
 
---! World ticks are translated to game ticks (or hours) depending on the
--- current speed of the game. There are 50 hours in a TH day.
+--! World ticks (every 18ms) are translated to game ticks (or hours) depending
+-- on the current speed of the game. There are 50 hours in a TH day.
 function World:onTick()
   if self.map.level_number == "MAP EDITOR" then return end
 
@@ -2461,22 +2461,6 @@ function World:afterLoad(old, new)
     self.wall_types = self.app.walls
     self:initWallTypes()
   end
-  if old < 183 then -- Adjust game speed and tick rates
-    local old_tick_rates = {
-      ["Pause"]              = {0, 1},
-      ["Slowest"]            = {1, 9},
-      ["Slower"]             = {1, 5},
-      ["Normal"]             = {1, 3},
-      ["Max speed"]          = {1, 1},
-      ["And then some more"] = {3, 1},
-    }
-    for name, rate in pairs(old_tick_rates) do
-      if rate[1] == self.hours_per_tick and rate[2] == self.tick_rate then
-        self:setSpeed(name)
-        break
-      end
-    end
-  end
   if old < 185 then
     -- Fix any missing initial in staff names
     self:updateInitialsCache()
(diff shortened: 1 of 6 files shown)


TheCycoONE:
In particular I think there's a risk that the earthquake effect might be negatively impacted.

lewri:
The game speed changes 0.67 > 0.68.0 do already affect the earthquake shake. It's likely that needs addressing.

Also nice double CI checks :)

lewri:
#2494 used an API version bump. Does this one need it too?

TheCycoONE:
Yes it does.

TheCycoONE:
Testing this myself a bit, earthquakes seem ok; but the demo video on the menu screen is being most annoying.

TheCycoONE:
Opening a new issue for the intro video issue, it's unrelated to my change.

CorsixTH/Lua/world.lua
```diff
@@ -718,12 +718,12 @@ end
...
-  ["Normal"]             = {2, 3}, -- 2 ticks per 108ms = 1 tick per 54 ms
-  ["Max speed"]          = {1, 1}, -- 1 tick per 36 ms
-  ["And then some more"] = {2, 1}, -- 2 ticks per 36 ms = 1 tick 18ms
-  ["Speed Up"]           = {4, 1},
+  ["Slowest"]            = {1, 56}, -- 1 tick per s
+  ["Slower"]             = {1, 12}, -- 1 tick per 216ms
+  ["Normal"]             = {1, 3}, -- 1 ticks per 54 ms
+  ["Max speed"]          = {1, 2}, -- 1 tick per 36 ms
+  ["And then some more"] = {1, 1}, -- 1 tick 18ms
+  ["Speed Up"]           = {8, 1},
```
lewri, line 726: We also need an afterload here.

lewri (commented):
This PR also fixes issues with entity sounds not always playing (I never raised an issue for it though)

However, with this change the animation for the Autopsy machine's idle state goes to fast

TH:
[screen-recording(5).webm](https://github.com/user-attachments/assets/ec15e269-c042-4df4-bd6f-7852ef977094)

This PR:
[screen-recording(6).webm](https://github.com/user-attachments/assets/d11615bb-6d2d-40e1-bf5e-7a9616f1d685)

TheCycoONE:
Thanks @lewri 

I'll look at autopsy as that sounds like it wasn't written properly.

What about scroll speed.

lewri:
> What about scroll speed.

Regarding scroll speed, this PR roughly doubles its rate (it's definitely smoother though, which is probably a good thing).

TheCycoONE:
Fixed the scrolling speed and the save, still looking into animation speed. Shouldn't it have depended on game speed?

lewri:
Sounds like handyman sweep audio may also be playing back weirdly.

Animations are indeed dependent on game speed. The autopsy's idle state is a 2 frame cycle.

CorsixTH/Lua/world.lua
```diff
@@ -67,8 +67,8 @@ function World:World(app)
   self.entity_map = EntityMap(self.map)
 
   -- Time
-  self.hours_per_tick = 2
-  self.tick_rate = 3
+  self.hours_per_tick = 1
+  self.tick_rate = 6
```
tobylane, line 71:
```suggestion
  self.tick_rate = 3
```
This matches the Normal speed set below.

lewri (commented):
I still can't currently work out what's causing the autopsy's idle animation to run too fast, nor the sweep anim's sound sounding like it's playing over itself.

TheCycoONE:
Autopsy idle state wasn't right before, I don't think it was ever right because in AnimView it goes too fast too, it looks like it had special handling in the game to slow it down, similar to slow motion people. I'm considering that out of scope for this PR, but ok for it's own issue.

I think I liked the keyboard scroll speed better faster - open to opinions there.

lewri (approved):

TheCycoONE:
@lewri I updated the afterLoad code; I think this is better?

CorsixTH/Lua/world.lua
```diff
@@ -2462,14 +2462,6 @@ function World:afterLoad(old, new)
...
-      ["Pause"]              = {0, 1},
-      ["Slowest"]            = {1, 9},
-      ["Slower"]             = {1, 5},
-      ["Normal"]             = {1, 3},
-      ["Max speed"]          = {1, 1},
-      ["And then some more"] = {3, 1},
-    }
     for name, rate in pairs(old_tick_rates) do
       if rate[1] == self.hours_per_tick and rate[2] == self.tick_rate then
         self:setSpeed(name)
```
lewri, line 2467: Need to fully wipe out this old if moving it down.

CorsixTH/Lua/world.lua
```diff
@@ -2502,6 +2494,40 @@ function World:afterLoad(old, new)
...
+    local found_speed = false
+    for name, rate in pairs(old_tick_rates) do
+      if rate[1] == self.hours_per_tick and rate[2] == self.tick_rate then
+        found_speed = true
+        self:setSpeed(name)
+        break
+      end
+      if not found_speed then
+        self:setSpeed("Normal")
+      end
```
lewri, line 2528: I assume this block is meant to be outside the for loop? It's currently inside which means that it executes the first iteration on the loop.

PR #2703, opened 2024-10-30
Make VIP walk slow

lewri:
- VIPs walk slowly, like in TH
- Inspectors walk at normal speed, so unaffected by this

I also found that the `self.speed` variable on `humanoid` is pretty useless to have the value as `slow`. But not planning to look at that.

This PR should be followed by a future one to suggest changing walk speed where patients and vips are off map from the spawn point.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/walk.lua
+++ b/CorsixTH/Lua/humanoid_actions/walk.lua
@@ -111,6 +111,10 @@ local flag_flip_h = 1
 local navigateDoor
 
 local function action_walk_raw(humanoid, x1, y1, x2, y2, map, timer_fn)
+  -- The variables below must always make up factor*quantity = 8 or the
+  -- animation glitches
+  -- Factor must also be able to multiply by 2 to become an integer
+  -- Quantity must always be an integer
   local factor = 1
   local quantity = 8
   if humanoid.speed and humanoid.speed == "fast" then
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/humanoid_actions/walk.lua
```diff
@@ -111,6 +111,10 @@ local flag_flip_h = 1
 local navigateDoor
 
 local function action_walk_raw(humanoid, x1, y1, x2, y2, map, timer_fn)
+  -- lewri: for some reason these must always make up factor*quantity = 8 or the
```
lewri, line 114: I think this might be me more asking why. I can trim it to state the fact only instead if needed.

Alberth289346: My guess is that a tile walk animation has some `F` frames, and in those frames the animated NPC must move from center of the source tile to center of the destination tile, which is some `X` pixels horizontally and some `Y` pixels vertically.

Apparently this results in the stated numbers.

Alberth289346: I think theoretically you don't need to track the NPC position within a tile.
As far as I can see, you should be able to derive that from eg the position of the head sprite.
I tried that a few times, but so far no luck getting a useful result.

PR #2706, opened 2024-11-01
Use date and time in log filenames

### LINKED ISSUE #2665: Make gamelogs unique

Currently when a game crashes we overwrite the gamelog file, which could destroy error trace data before it is reported back to here.

A suggestion here is to generate a tagged gamelog.txt file with the savegame name and date/time in the name to preserve the error entry.

Additionally, a method would need to exist to remove old logs from within the game.

While having a single gamelog file makes it simpler for uses, it is potentially destructive in nature. Multiple will allow it to be pinpointed and preserved.

No

0.68.0

No

### Additional Information

_No response_

tobylane:
*Fixes #2665*

- The gamelog is written to a file, in a new logs subfolder, named for the current date, time. At this point, older logs are deleted.

Flaws
- No UI yet

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -392,10 +392,25 @@ end
 --! Create the gamelog, using the launch time in the filename, and write the system information.
 function App:initGamelogFile()
   self.gamelog_path = self.user_log_dir .. os.date("%y-%m-%d--%H-%M-%S--gamelog.txt", os.time(os.date("!*t")))
-  local fi = self:writeToFileOrTmp(self.gamelog_path)
+  local fi, success = self:writeToFileOrTmp(self.gamelog_path)
   local sysinfo = self:gamelogHeader()
   fi:write(sysinfo)
   fi:close()
+  if success then self:trimLogs() end -- Only trim logs if logs folder is writable
+end
+
+--! Trims the logs folder of old game logs down to ten files.
+function App:trimLogs()
+  local log_retention, log_table = 11, {}
+  for node in lfs.dir(self.user_log_dir) do
+    local file = self.user_log_dir .. pathsep .. node
+    if node:sub(-12) == "-gamelog.txt" then
+      table.insert(log_table, file)
+    end
+  end
+  table.sort(log_table,
+      function(a, b) return lfs.attributes(a, "modification") > lfs.attributes(b, "modification") end)
+  for i=log_retention, #log_table do os.remove(log_table[i]) end
 end
 
 --! Tries to initialize the user level and campaign directories
@@ -1011,7 +1026,7 @@ function App:saveConfig()
 end
 
 --! Tries to open the given file or a file in OS's temp dir.
--- Returns the file handler
+-- Returns the file handler, and true if it was written to the intended file.
 --!param file The full path of the intended file
 --!param mode The mode in which the file is opened, defaults to write
 function App:writeToFileOrTmp(file, mode)
@@ -1026,7 +1041,7 @@ function App:writeToFileOrTmp(file, mode)
     end
   end
   assert(f, "Error: cannot write to filesystem")
-  return f
+  return f, not err
 end
 
 function App:fixHotkeys()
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/app.lua
```diff
@@ -389,24 +389,38 @@ function App:init()
...
+  local gamelog_path = self:getGamelogPath(save_name)
   local gamelog = io.open(gamelog_path, "r")
   if gamelog then gamelog:close() return end
 
   local fi = self:writeToFileOrTmp(gamelog_path)
   local sysinfo = self:gamelogHeader()
   fi:write(sysinfo)
   fi:close()
+  self:trimLogs()
+end
```
lewri, line 410: If we're making individual logs we no longer need to init a file imo.
We should still init sysinfo and trim though.

CorsixTH/Lua/app.lua
```diff
@@ -1696,6 +1713,7 @@ function App:load(filepath)
   if self.world then
     self:worldExited()
   end
+  self.save_name = filepath:match(".*" .. pathsep .. "(.+)%.sav$")
```
lewri, line 1716: When player saves a game, probably useful to refresh self.save_name too.

Does `Continue Game` from Main Menu also check the autosaves folder? ... I don't think it does but if it does it could be an issue.

CorsixTH/Lua/app.lua
```diff
@@ -389,24 +389,38 @@ function App:init()
...
+end
+
+--! Trims the logs folder of old game logs down to ten files.
+function App:trimLogs()
+  local log_rentention, log_table = 11, {}
+  for node in lfs.dir(self.user_log_dir) do
+    local file = self.user_log_dir .. pathsep .. node
+    if lfs.attributes(file, "mode") == "file" then
+      table.insert(log_table, file)
+    end
```
lewri, line 419: Probably not an issue, but we're deleting any file indiscriminately in this folder > 1mo.

CorsixTH/Lua/app.lua
```diff
@@ -389,24 +389,38 @@ function App:init()
...
---! Works out the intended location of the gamelog file.
+--! Works out the intended location of the gamelog file based on the time and save game name.
 --!return full path gamelog should exist at
-function App:getGamelogPath()
-  local config_path = self.command_line["config-file"] or ""
-  config_path = config_path:match("^(.-)[^" .. pathsep .. "]*$")
-  return config_path .. "gamelog.txt"
+function App:getGamelogPath(save_name)
+  return self.user_log_dir .. pathsep .. os.date("%y-%m-%d--%H-%M-%S--") ..
+      (save_name or 'unknown') .. ".txt"
```
lewri, line 396: `New Game`?

lewri (commented):
Some other thoughts about this are out of scope, which if I remember I'll add them to the issue.

CorsixTH/Lua/app.lua
lewri: Unless you're trying to establish we just have write permission, toby?

CorsixTH/Lua/app.lua
tobylane: It does, so continuing a named save will be logged in files named for the autosave. That is possible to fix for the future, this save_name can be in world, but it could be confusing for existing games. I'm thinking about taking the save name out of the filename, so at least old gamelogs aren't overwritten.

TheCycoONE:
Why don't we just use the date and exclude the save name?

TheCycoONE:
I'm working on a bigger effort here to use spdlog with it's built in rolling log appender, and capturing all our engine errors/messages in the same files.

Maybe close this PR?

TheCycoONE:
Digging deeper, my approach doesn't solve this problem at all. Reopening this PR. Sorry @tobylane

tobylane:
The gamelog path is set at launch. With the time in the filename, it's taken out of the gamelog header. Old gamelogs are only trimmed if the new one was successfully written to the intended path, rather than tmp.

lewri:
That would indicate this is using UTC time, but needs a new blank issue opened for it if it should be targeting local device time.

PR #2719, opened 2024-11-11
Update brazilian_portuguese.lua

altiereslima:


MandrakeHorse:
For reference, the original string for VIP Remark 15 is "I'm shocked. You can't call that a hospital! I'm off for a pint." so this is a more accurate translation.

PR #2727, opened 2024-11-27
Set minimum possible machine strength after repair to 2

ARGAMX:
-
- ... as a result of Discord discussion...
- We noticed that in the vanilla TH machines have a minimum possible value for the strength, which we did not follow in CTX. This value 2. This PR fixes that for CTX.
-

### DIFF
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -290,6 +290,7 @@ end
 function Machine:machineRepaired(room)
   room.needs_repair = nil
   self:reduceStrengthOnRepair(room)
+  self.times_used = 0
   self:setRepairing(nil)
   setSmoke(self, false)
   local taskIndex = self.hospital:getIndexOfTask(self.tile_x, self.tile_y, "repairing")
@@ -297,7 +298,6 @@ function Machine:machineRepaired(room)
 end
 
 --! Calculates if machine strength should be reduced as a result of repair
--- Also resets 'times_used' value
 --!param room (object) machine room
 function Machine:reduceStrengthOnRepair(room)
   local minimum_possible_strength = 2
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/entities/machine.lua
```diff
@@ -289,6 +289,19 @@ end
...
+  local taskIndex = self.hospital:getIndexOfTask(self.tile_x, self.tile_y, "repairing")
+  self.hospital:removeHandymanTask(taskIndex, "repairing")
+end
+
+--! Calculates if machine strength should be reduced as a result of repair
+-- Also resets 'times_used' value
+--!param room (object) machine room
+function Machine:reduceStrengthOnRepair(room)
+  local minimum_possible_strength = 2
+  if self.strength <= minimum_possible_strength then return end
```
lewri, line 304: Early return doesn't reset times used (very important :) ) 

Two options:
Add it to early return if block e.g.
```lua
  if self.strength <= minimum_possible_strength then
    self.times_used = 0
    return
  end
```
Or move `self.times_used = 0` at line 315 into the calling function `Machine:machineRepaired(room)` to happen after `self:reduceStrengthOnRepair(room)` at line 292.

ARGAMX: > Early return doesn't reset times used (very important :) )
> 
> Two options: Add it to early return if block e.g.
> 
> ```lua
>   if self.strength <= minimum_possible_strength then
>     self.times_used = 0
>     return
>   end
> ```
> 
> Or move `self.times_used = 0` at line 315 into the calling function `Machine:machineRepaired(room)` to happen after `self:reduceStrengthOnRepair(room)` at line 292.

Oh, my bad. 
That's really important. 
Should testing more carefully by me instead of being lazy. 
Thanks. 👍

PR #2735, opened 2024-12-09
Add level file overlays

tobylane:
- Level file overlay lets a level writer copy and extend an existing campaign or custom level. They can choose either a difficulty and level number (hard 1) or a custom level in any of the relevant folders (avatar). I think overlay is the right word, as nothing is overwritten by configs loaded later, but I'm not sure. It only deals with the lines starting #, not %. To extend rather than overwrite, additional win/lose conditions should use higher numbers, eg starting at 10.

This custom level file is the first campaign level, with harder win conditions. To overlay the avatar custom level, use `#overlay avatar`.

```
---------------------- General Information -------------------------
%Name = "Prestige - Bronze L1"
%MapFile = "LEVEL.L1"

%LevelBriefing = "Beaten the original campaign and looking for something harder? We've added more goals.

You must get your reputation to 800, and your average staff and patient happiness above 70%."

%LevelDebriefing = "Nice one!"

#overlay.difficulty.level_number = hard 1

------------------ Winning and Losing Conditions -------------------
1 Total reputation
2 Balance total
3 Percentage people your hospital has handled
4 Count of cured patients
5 Percentage people have been killed
6 Hospital value
7 Staff happiness
8 Patient happiness

#win_criteria[10].Criteria.MaxMin.Value.Group.Bound	1	1	800	1	625
#win_criteria[11].Criteria.MaxMin.Value.Group.Bound	7	1	70	1	50
#win_criteria[12].Criteria.MaxMin.Value.Group.Bound	8	1	70	1	50

#lose_criteria[10].Criteria.MaxMin.Value.Group.Bound	7	1	20 4 50
#lose_criteria[11].Criteria.MaxMin.Value.Group.Bound	8	1	20 5	50
```

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -635,15 +635,15 @@ function App:loadCampaign(campaign_file)
       campaign_info.winning_text_table)
 
   if self:loadLevel(level_info.path, nil, level_info.name,
-      level_info.map_file, level_info.briefing, nil, _S.errors.load_level_prefix, campaign_info) then
+      level_info.map_file, level_info.briefing, nil, _S.errors.load_level_prefix, campaign_info) and self.world then
     -- The new world needs to know which campaign to continue on.
     self.world.campaign_info = campaign_info
-  end
 
-  -- Play the level advance movie from a position where this campaign will end at 12
-  if campaign_info.movie then
-    local n = math.max(1, 12 - #campaign_info.levels)
-    self.moviePlayer:playAdvanceMovie(n)
+    -- Play the level advance movie from a position where this campaign will end at 12
+    if campaign_info.movie then
+      local n = math.max(1, 12 - #campaign_info.levels)
+      self.moviePlayer:playAdvanceMovie(n)
+    end
   end
 end
 


lewri:
We already do have `base_config` setup in CorsixTH to plug gaps in custom files. Could this parameter not be named to follow that idea and call it the same; as a thought?

TheCycoONE:
I'm a little afraid of overlays of overlays - not supported.

What is the motivation for this feature?

tobylane:
> Could this parameter not be named to follow that idea and call it the same; as a thought?

The custom level with the overlay line is the base config in that case, with other base configs on top of it. I'm using overlay as the reverse, what the base config is called by its parent.

> I'm a little afraid of overlays of overlays - not supported.

Not intentionally supported, but not disallowed either. I can test, and prevent dependency loops.

> What is the motivation for this feature?

The level file I included is complete, but without this feature I'd have to duplicate almost all of hard01.sam. This level idea is intentionally derivative of the original, copywritten content. More generally, someone could create a detailed first level of a campaign, then more levels with minor variations or different maps.

Alberth289346:
How does it handle getting a missing level file that you overlay?
For files provided with Corsix this should be somewhat work (see below when not), since they "always" exist (modulo renaming!), but "custom level"??

There is also the risk of having more than one level file around. People distribute level files, and adapt them based on play-tests. So depending on when you downloaded a level file, it may not have the same content.
A user may also have several of them, if they actively tested the level.

Ie how do you handle overlaying the file that was supposed to be overlay-ed? (Just the name is probably not enough for .)

tobylane:
> How does it handle getting a missing level file that you overlay?
> 
> For files provided with Corsix this should be somewhat work (see below when not), since they "always" exist (modulo renaming!), but "custom level"??
> 

There is a specific error message on line 438. It continues, with only a printed error, but maybe this one is worth showing in game. It could be user error, where the others are only level editor error. 

> There is also the risk of having more than one level file around. People distribute level files, and adapt them based on play-tests. So depending on when you downloaded a level file, it may not have the same content.
> 
> A user may also have several of them, if they actively tested the level.
> 

True. I use the existing functions for that,  'App:getAbsolutePathToLevelFile' uses ipairs so someone sharing a a pair of dependent levels should specify where they go to take precedence.

> Ie how do you handle overlaying the file that was supposed to be overlay-ed? (Just the name is probably not enough for .)

We don't have any versioning, yet.

tobylane:
The existing display error part of app:loadlevel didn't work, but with that fixed, it seems the intended way to show errors. All the errors must be shown on the level select screen, not after load when the introduction is shown. The existing loadmapconfig error is raised from printed to shown to the user. Should all these strings be localised?

Message shown for loading a level that overlays another level with an error
Screenshot 2024-12-21 at 14 57 04
Message shown for a level that overlays a campaign difficulty that doesn't exist
Screenshot 2024-12-21 at 10 18 48

Alberth289346:
> Message shown for loading a level that overlays another level with an error

It's good that an error is shown. There does seem to be room for improvement though:

- More detail than "somewhere in the hard difficulty there is a problem" would be most useful imho.
- Is it useful for someone not developing code, maps, or levels to even see these errors?
  My first thought is "no", refusing to load the level is likely more useful then.

I ca live with doing this somewhen in the future though.

Alberth289346:
As for scope of the overlay feature, perhaps start with only allowing overlaying the original TH levels?

That is a valid usecase, everybody has those files, and they don't change in time.
Hopefully there is also exact 1 version of each original file, but not sure of that.

tobylane:
I believe they are the same, the English and French TH folders I have checked have identical Full00.sam files.

Taken out the overlay custom level, reduced to difficulty and level number. The error messages give the right values. These should only be seen by level editors, who won't need to see the terminal.

CorsixTH/Lua/map.lua
```diff
@@ -402,9 +404,33 @@ function Map:loadMapConfig(filename, config, custom)
...
+        -- Finally overlay additional CorsixTH config per level
+        local level_path = self.app:getFullPath({"Levels", "original" .. level_number .. ".level"})
+        _, config = self:loadMapConfig(level_path, config, true)
+      else return "No difficulty and level number given in overlay field of custom level"
+      end
+    end
     return nil, config
   else
-    return "Error: Could not find the configuration file, only 'Base Config' will be loaded for this level.", config
+    return "Error: Could not find the configuration file.", config
```
lewri, line 433: Are we no longer trying to load base_config as a final fallback Toby?

tobylane: Yes, where we attempt to load Corsixth changes to the campaign is the only time we want to ignore failure. Anywhere else, it’s a fatal error that the user (a level writer) should see.

Alberth289346: Add a comment to explain that?

Exceptions to a general rule raise questions about correctness, unless there is clear evidence it's intentional.

tobylane: There is one in the loadMapConfig doc, now adding comments to the calls where failure is ignored.

This now warns the user if a campaign level file is missing, which is only needed as we don't check all the files at start. It is meant for end users, and could be a language string. It's being strict, where we may have previously papered over with the base config.

tobylane:
Localised the error messages in map loading.

CorsixTH/Lua/map.lua
```diff
@@ -216,8 +216,12 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
         level_no = "0" .. level
       end
       -- Override with the specific configuration for this level
-      _, result = self:loadMapConfig(difficulty .. level_no .. ".SAM", base_config)
-      -- Finally load additional CorsixTH config per level
+      errors, result = self:loadMapConfig(difficulty .. level_no .. ".SAM", base_config)
+      if errors then
+        return nil, "Warning: Could not find campaign level " .. difficulty .. level_no ..
```
Alberth289346: No `.SAM` ?

A further improvement would be to use a separate variable for the filename, like you do at line 225 below. Doing that basically reduces the chance to create an inconsistency between what you load and what you report.

CorsixTH/Lua/map.lua
```diff
@@ -422,7 +426,7 @@ function Map:loadMapConfig(filename, config, custom)
         -- Overlay with the specific configuration for this level
         errors, config = self:loadMapConfig(difficulty .. level_number .. ".SAM", config)
         if errors then return "Overlay level number must be 1-12. Current value is " .. level_number end
-        -- Finally overlay additional CorsixTH config per level
+        -- Finally overlay additional CorsixTH config per level. As these files may not exist, failure is permitted.
         local level_path = self.app:getFullPath({"Levels", "original" .. level_number .. ".level"})
         _, config = self:loadMapConfig(level_path, config, true)
```
Alberth289346, line 425: Not your fault, but blindly assuming that a load works is a bad idea.

The loaded data gets used I assume, which means that all kinds of things can go wrong if the loaded data is not correct or the load fails. As it's not caught here, it will cause trouble elsewhere at a spot that has nothing to do with the cause of the error. It will take quite some time before it is tracked back to here.

It's fine to assume the load does not fail often enough to spend time on error reporting. However do check the result. The simplest is an `assert`.
If that triggers, you saved yourself a week of debugging. If it happens more frequently, something more convenient should probably done, but that's a few steps ahead currently.

As you didn't introduce this, I am fine with not fixing it here.

tobylane:
Refactored more, so that files are only used when they should exist, and failure is meaningful.

tobylane:
Rebased and ready.

CorsixTH/Lua/map.lua
```diff
@@ -279,6 +264,46 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
...
+-- A set of the TH campaign levels with additional CorsixTH config files, found in CorsixTH/Levels.
+local additional_config = list_to_set({"05", "07", 11, 12})
+
+--! Load further level configurations for the main campaign.
+--!param difficulty (string)
+--!param level_number (integer)
+--!param config (table) The level config created so far
+--!return error (string) Localised error message if error happens
+--!return config (table) Level config, if loaded successfully
+function Map:_loadCampaignLevel(difficulty, level_number, config)
```
lewri, line 276: Could we make this function name even more obvious it's not for custom campaigns?
`_loadOriginalCampaignLevel`?

tobylane: Agreed, done.

tobylane:
Added a commit to ensure map loading errors are shown, rather than the movie. Rebased.

CorsixTH/Lua/map.lua
```diff
@@ -169,30 +169,22 @@ the original game levels are considered.
 has been loaded.
 ]]
 function Map:load(level, difficulty, level_name, map_file, level_intro, map_editor)
-  local objects, _
+  local objects, _, result
```
Alberth289346, line 172: Please avoid introducing new variables miles before you use them.

Introduce it just above line 207 (that seems the first use of it).

CorsixTH/Lua/dialogs/resizables/new_game.lua
```diff
@@ -196,17 +196,18 @@ function UINewGame:buttonStart()
 end
 
 function UINewGame:startGame(difficulty)
-  if self.ui.app:loadLevel(1, difficulty, nil, nil, nil, nil,
-      _S.errors.load_level_prefix, nil) then
-    self.ui.app.moviePlayer:playAdvanceMovie(1)
+  local app = self.ui.app
```
Alberth289346, line 199: Why did you add this? `TheApp` is a defined global name for `app`.

I know we have code that pulls the value from many random spots, but there really isn't need for that.

CorsixTH/Lua/map.lua
```diff
@@ -252,10 +238,9 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
     else
       return nil, errors
     end
-    assert(base_config, "No base config has been loaded!")
     errors, result = self:loadMapConfig(level, base_config, true)
     if errors then
-      print(errors)
+      return nil, errors
```
Alberth289346, line 243: Add a Luadoc line stating that you do this.

Alberth289346:
Seems ok

tobylane:
Reinstated loading of CorsixTH overlay on campaign level 8.

CorsixTH/Lua/map.lua
```diff
@@ -207,22 +201,19 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
...
       end
-      -- Override with the specific configuration for this level
-      _, result = self:loadMapConfig(difficulty .. level_no .. ".SAM", base_config)
-      -- Finally load additional CorsixTH config per level
-      local level_path = self.app:getFullPath({"Levels", "original" .. level_no .. ".level"})
-      _, result = self:loadMapConfig(level_path, result, true)
       self.level_config = result
     end
   elseif map_editor then
+    local _
```
lewri, line 216: Do we usually have to declare the underscore for an unwanted value?
I suppose it could potentially leak into the global scope otherwise.

TheCycoONE: Yes we have to

TheCycoONE:
Are you going to add an example @tobylane 

We'll also need to note somewhere or maybe catch when it's not compatible with the demo.

tobylane:
I plan to make a campaign starting with the example in the top comment, and submit it to the community content repo. The Custom Levels wiki page will have a paragraph on this feature, linking to that campaign. Should there be an example here too, a second example town?

The demo can't run custom levels.

tobylane:
https://github.com/CorsixTH/CommunityContent/pull/6

This is a campaign of examples of how this feature is used.

PR #2737, opened 2024-12-11
Only capture mouse in game/mapeditor

### LINKED ISSUE #2716: Only capture mouse in game
Currently we capture mouse for the whole session of the program. However, it would likely be better to limit capture to be active only while in-game.

A user may need to check something in the file system before launching a game, which would mean needing to disable mouse capture first if it's on.

To build upon #2715 / #2711

tobylane:
*Fixes #2716*

- Sets the mouse capture to the user setting when entering/creating a world, and always release the mouse on returning to the main menu.


tobylane:
Should the options dialog tooltip for capture mouse include this change, so people don't expect it to work straight away?

CorsixTH/Lua/world.lua
```diff
@@ -171,6 +171,8 @@ function World:World(app)
   }
 
   self:gameLog("Created game with savegame version " .. self.savegame_version .. ".")
+
+  self.app:setCaptureMouse()
```
lewri, line 175: I'd suggest you enforce this via `GameUI` rather than `World`

lewri (commented):
I'm not sure if there was discussion if map editor should have mouse capture?

CorsixTH/Lua/app.lua
```diff
@@ -1837,6 +1838,7 @@ function App:afterLoad()
   self.map:afterLoad(old, new)
   self.ui:afterLoad(old, new)
   self.world:afterLoad(old, new)
+  self:setCaptureMouse()
```
lewri, line 1841: Same as the world issue. I'm pretty sure `GameUI` should be capable of this lifting as a single point of control.

lewri:
> Should the options dialog tooltip for capture mouse include this change, so people don't expect it to work straight away?

It definitely needs to be a changelog note. I'm tempted to suggest we start adding some changelog info to releases on first launch of the new version to put it even more visibly

TheCycoONE:
The map editor should IMO for the same reason as the game.

TheCycoONE:
> > Should the options dialog tooltip for capture mouse include this change, so people don't expect it to work straight away?
> 
> It definitely needs to be a changelog note. I'm tempted to suggest we start adding some changelog info to releases on first launch of the new version to put it even more visibly.
> 
> Extending a tooltip is very trivial to do though anyway.

I don't think it's that drastic.

TheCycoONE:
As for the tooltip it wouldn't hurt to add "in game"

tobylane:
Moved to GameUI.

I didn't look first, the options tooltip is already "Click to toggle capturing the cursor in the game window", which could be specific enough?

lewri:
> Moved to GameUI.
> 
> I didn't look first, the options tooltip is already "Click to toggle capturing the cursor in the game window", which could be specific enough?

Change to "while in game" is more nicely specific.

tobylane:
Done, and noted the need for a forum post.

Issue #1454, opened 2018-09-03
Missing vanilla mechanics


bobucles:
#### Describe the issue
1. Pawns do not evacuate during "place a door" stage of blueprint. This happens in vanilla even though the "zzz" isn't shown.

2. Epidemic clock does not instantly end when an epidemic is successfully eradicated. This is the leading cause of an early clock ending, at least in a good hospital.

#### System Information
CorsixTH Version: 0.62
Windows 10*
Theme Hospital Version: GOG.com

TheCycoONE:
Can you split this into two issues so we can track each please.

TheCycoONE:
I'm making this ticket about your first issue.

ARGAMX:
Can be closed because
second issue moved to more specialised ticket too
#2738

lewri:
Is the first issue actually fixed by another PR?

ARGAMX:
> Is the first issue actually fixed by another PR?

oops, indeed...
I was read "making ticket" as creating new one. 
Now I realised that this was just ticket renaming and that's all.

I can record a video for the first problem and create more specified ticket.
Should I do?

lewri:
This one doesn't seem to actually be a big problem overall. In TH, humanoids were 'hinted' at doors phase to start getting off the tiles. In CorsixTH we lack the hinting and instead have an explicit kick off after doors.

It could be simple enough to call `EditRoom:clearArea()` without the lockdown and cursor change when we go to enter the doors phase. Not tested though.

ARGAMX:
I think the point of this task is to start expelling humanoids from the construction site before placing the door, not after.
That is, when the blueprints change their transparency to less transparent, the humanoids should already be given the command to leave the construction site.
This is exactly what happens in the original TH.
In CorsixTH, humanoids are currently expelled only after placing the door and attempting to click the checkbox.
This means TH starts the construction site clearing process earlier than CorsixTH.

Moreover, humanoids do not leave the construction site via the optimal route, but I dedicated #3505 to this certain issue.

PR #2740, opened 2024-12-12
#2738 stop epidemic timer on epidemic successfully eradicated

### LINKED ISSUE #2738: [Bug] Epidemic. Timer doesn't stop when an epidemic is successfully eradicated
### Describe the issue

In CTH Epidemic clock does not instantly end when an epidemic is successfully eradicated.
In vanilla TH it does.

epidemic

vanilla Theme Hospital footage which demonstrates this behaviour: 
https://youtu.be/_PeCGwKKoZs?t=471

### Steps to Reproduce

2. Observe how last epidemic patient will be cured in surgery.
3. Observe that epidemic clock doesn't stop.

### Expected Behaviour

Epidemic clock should instantly end when an epidemic is successfully eradicated.

### Save Game

[epidemic_succesfully_eridicated_surgery.sav.zip](https://github.com/user-attachments/files/18102175/epidemic_succesfully_eridicated_surgery.sav.zip)

### CorsixTH Version

0.65.0 (still relevant for 0.68.0)

### Operating System

MacOS Sonoma 14.5

ARGAMX:
*Fixes #2738, #1454*

- Fixes bug when epidemic Timer doesn't stop when an epidemic is successfully eradicated.
-

### DIFF
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -271,7 +271,7 @@ end
 --[[ Check for conditions that no any uncured infected patients
 left in hospital. If so then epidemic must end earlier than the
 length of the timer.]]
-function Epidemic:checkThereAreNoAnyInfectedPatientsLeft()
+function Epidemic:checkNoInfectedPatients()
   if self.result_determined then return end
 
   if (self:countInfectedPatients() == 0) then
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -262,12 +262,15 @@ function Epidemic:checkNoInfectedPlayerHasLeft()
         px and py and not self.hospital:isInHospital(px,py) then
       -- Patient escaped from the hospital, discovery is inevitable.
       self.result_determined = true
-      if not self.inspector then
-        self:spawnInspector()
-      end
       self:finishCoverUp()
       return
     end
+  end
+
+  local all_infected_cured = self:countInfectedPatients() == 0 
+  if all_infected_cured then
+    self:finishCoverUp()
+    return
   end
 end
 
@@ -410,6 +413,10 @@ end
 later (@see applyOutcome) ]]
 function Epidemic:finishCoverUp()
   self.result_determined = true
+
+  if not self.inspector then
+    self:spawnInspector()
+  end
 
   self.timer:close()
 
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -267,7 +267,7 @@ function Epidemic:checkNoInfectedPlayerHasLeft()
     end
   end
 
-  local all_infected_cured = self:countInfectedPatients() == 0 
+  local all_infected_cured = self:countInfectedPatients() == 0
   if all_infected_cured then
     self:finishCoverUp()
     return
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -277,7 +277,6 @@ function Epidemic:checkThereAreNoAnyInfectedPatientsLeft()
   local no_any_infected_patients_left = self:countInfectedPatients() == 0
   if no_any_infected_patients_left then
     self:finishCoverUp()
-    return
   end
 end
 
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -110,9 +110,9 @@ function Epidemic:tick()
   if self.coverup_in_progress then
     if not self.result_determined then
       self:checkNoInfectedPlayerHasLeft()
+      self:checkNoInfectedPatients()
       self:markedPatientsCallForVaccination()
       self:showAppropriateAdviceMessages()
-      self:checkThereAreNoAnyInfectedPatientsLeft()
     end
     self:tryAnnounceInspector()
   end
@@ -274,7 +274,7 @@ length of the timer.]]
 function Epidemic:checkNoInfectedPatients()
   if self.result_determined then return end
 
-  if (self:countInfectedPatients() == 0) then
+  if self:countInfectedPatients() == 0 then
     self:finishCoverUp()
   end
 end
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -268,6 +268,9 @@ function Epidemic:checkNoInfectedPlayerHasLeft()
   end
 end
 
+--[[ Check for conditions that no any uncured infected patients
+left in hospital. If so then epidemic must end earlier than the
+length of the timer.]]
 function Epidemic:checkThereAreNoAnyInfectedPatientsLeft()
   if self.result_determined then return end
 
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/epidemic.lua
```diff
@@ -262,13 +262,16 @@ function Epidemic:checkNoInfectedPlayerHasLeft()
...
       self:finishCoverUp()
       return
     end
   end
+
+  local all_infected_cured = self:countInfectedPatients() == 0
+  if all_infected_cured then
+    self:finishCoverUp()
+    return
+  end
```
lewri, line 274: Function this is inserted in does not fit its current purpose.

The function name either needs updating; or this check could go elsewhere. Judging how the the tick function is set up, you'd make it a function in itself. First option better this time round though.

The name `all_infected_cured` is also potentially not a true reflection of the check you're running. We don't know that every patient got cured, one might have died instead.

ARGAMX: > Function this is inserted in does not fit its current purpose.
> 
> The function name either needs updating; or this check could go elsewhere. Judging how the the tick function is set up, you'd make it a function in itself. First option better this time round though.
> 
> The name `all_infected_cured` is also potentially not a true reflection of the check you're running. We don't know that every patient got cured, one might have died instead.

Yeap, agree.
I have moved new check to separated function.

Also I agree that `all_infected_cured` not reflecting the thing it contains. I have renamed it to `no_any_infected_patients_left`.

Also from my point of view we should rename function `checkNoInfectedPlayerHasLeft` to something more understandable. For example `checkIsAnyInfectedPatientLeftTheHospitalAndRevealedEpidemic`.

What do you think about this?

Alberth289346: > `checkNoInfectedPlayerHasLeft`

Players also get infected?  :O

In general, the name of the function indicates what you can expect to find inside.

The function name here is quite specific. For a larger function this is fine, as you mostly need all space for implementing the functionality. (Eg if you need 300 lines of code for that check, it's unlikely to be a good spot for doing something else as well.)

For a short function (which seems to be the case here), it is often more useful to give a slightly more general name. In that way, you sort-of pave the path for adding more checks or doing more work in that function in the future.
For example, if you use "isEpidemicContained" or "isEpidemicUnderControl" for example, then basically the function name suggests it covers any action that must be done when the epidemic "has finished". It could for example call off all nurses that are doing vaccination patrol in the corridor, etc.

The suggested function name is too narrow in scope for the latter action ("hasAnyPatientLeft" does not intuitively suggest actions wrt nurses), so if nurses need to be called off somewhen in the future development, it must be done elsewhere or the name of the function must be adapted.

CorsixTH/Lua/epidemic.lua
```diff
@@ -261,16 +262,22 @@ function Epidemic:checkNoInfectedPlayerHasLeft()
...
-      if not self.inspector then
-        self:spawnInspector()
-      end
       self:finishCoverUp()
       return
     end
   end
 end
 
+function Epidemic:checkThereAreNoAnyInfectedPatientsLeft()
```
Alberth289346, line 274:
```suggestion
function Epidemic:checkNoInfectedPatients()
```

Function needs some Lua documentation.

CorsixTH/Lua/epidemic.lua
```diff
@@ -261,16 +262,22 @@ function Epidemic:checkNoInfectedPlayerHasLeft()
...
+  if self.result_determined then return end
+
+  local no_any_infected_patients_left = self:countInfectedPatients() == 0
+  if no_any_infected_patients_left then
```
Alberth289346, line 278: You could move the `self:countInfectedPatients() == 0` check inside the `if`

Having zero infected patient implies already they don't exist (which is what I read in the variable name).

CorsixTH/Lua/epidemic.lua
```diff
@@ -411,6 +418,10 @@ later (@see applyOutcome) ]]
 function Epidemic:finishCoverUp()
   self.result_determined = true
 
+  if not self.inspector then
+    self:spawnInspector()
+  end
```
Alberth289346, line 425: I never played an epidemic, but an inspector always arrives, even if no infected uncured patients left the hospital?

If not, this looks like changed behavior

ARGAMX: > I never played an epidemic, but an inspector always arrives, even if no infected uncured patients left the hospital?
> 
> If not, this looks like changed behavior

Yeap, he should always arrive after CoverUp happened.
The only scenario when he won’t come is if you confess the epidemic and pay the fine straightaway, instead of doing CoverUp.
Currently, for some reason, the code was written so strangely, as if he could not always arrive after CoverUp.

CorsixTH/Lua/epidemic.lua
ARGAMX: > Function needs some Lua documentation.

Added as for other functions in this file

CorsixTH/Lua/epidemic.lua
ARGAMX: change to `if (self:countInfectedPatients() == 0) then`

CorsixTH/Lua/epidemic.lua
```diff
@@ -261,16 +262,23 @@ function Epidemic:checkNoInfectedPlayerHasLeft()
...
   end
 end
 
+--[[ Check for conditions that no any uncured infected patients
+left in hospital. If so then epidemic must end earlier than the
+length of the timer.]]
+function Epidemic:checkNoInfectedPatients()
+  if self.result_determined then return end
+
+  if (self:countInfectedPatients() == 0) then
```
lewri, line 277: We can drop the brackets too :)

CorsixTH/Lua/epidemic.lua
```diff
@@ -112,6 +112,7 @@ function Epidemic:tick()
       self:checkNoInfectedPlayerHasLeft()
       self:markedPatientsCallForVaccination()
       self:showAppropriateAdviceMessages()
+      self:checkThereAreNoAnyInfectedPatientsLeft()
```
lewri, line 115: Function name got a little garbed here, where's it referencing?

CorsixTH/Lua/epidemic.lua
```diff
@@ -261,16 +262,23 @@ function Epidemic:checkNoInfectedPlayerHasLeft()
...
       self:finishCoverUp()
       return
     end
   end
 end
 
+--[[ Check for conditions that no any uncured infected patients
+left in hospital. If so then epidemic must end earlier than the
+length of the timer.]]
+function Epidemic:checkNoInfectedPatients()
```
lewri, line 274: I'm fine with this, trying to add clarity to this function name any more likely will just make things worse.

CorsixTH/Lua/epidemic.lua
lewri, line 115: Potentially sanity check the order of these, I'm sure it's still fine.

lewri (commented):
Good refactor ontop of the fix 👍

CorsixTH/Lua/epidemic.lua
ARGAMX: > Potentially sanity check the order of these, I'm sure it's still fine.

I changed the order, but honestly I don't think it matters here. but thanks for the tip.

ARGAMX: > Function name got a little garbed here, where's it referencing?

Oh, true. thanks.

CorsixTH/Lua/epidemic.lua
ARGAMX: Dropped 👌

lewri (commented):
LGTM

PR #2748, opened 2024-12-21
Calculate Epidemic outcome on inspector arrive

ARGAMX:

In CTH 0.68.0 currently epidemic process not aligned to TH epidemic.

Discrepancies:

- In CTH currently epidemic and infecting process can stop on `finishCoverUp`, but in TH it stops only at `inspectorArrvied`. 
- In CTH currently epidemic outcome can be calculated on `finishCoverUp`, but in TH it calculates only at `inspectorArrvied`. 
- In CTH currently indicating epidemic patients can be ended on `finishCoverUp`, but in TH it indicating until `inspectorArrvied`. 

This change align these things.

<details>
<summary>Epidemic flow in Theme Hospital</summary>

	Game generates one pathogenic patient -> first phase.

1. Latent phase

	➖Clock hidden
	➖Timer not ticking
	➖Player can't mark patients
	➖Nurses can't vaccinate patients
	➖Infected don't have icon
	➕Infected can infect
	➖Inspector spawned
	➖Result not determined 

	Infected diagnosed. User selects "Attempt to cure..." in fax -> next phase.
	Infected diagnosed. User selects "Declare epidemic..." in fax -> 🏁 Epidemic cleared.
	All Infected leaved without diagnosis -> 🏁 Epidemic cleared.

2. CoverUp phase

	➕Clock visible
	➕Timer ticking
	➕Player can mark patients
	➕Nurses can vaccinate patients
	➕Infected have icon
	➕Infected can infect
	➖Inspector spawned
	➖Result not determined 

	No any Infected in hospital -> next phase. (❌ CorsixTH discrepancy. PR #2740)
	One of infected go outside -> next phase.
	Timer is up -> next phase.

3. Inspection mode

	➕Clock visible (❌ CorsixTH discrepancy)
	➖Timer not ticking
	➖Player can't mark patients (Why?)
	➕Nurses can vaccinate patients
	➕Infected have icon (❌ CorsixTH discrepancy. This PR  #2748)
	➕Infected can infect (❌ CorsixTH discrepancy. This PR  #2748)
	➕Inspector spawned
	➖Result not determined (❌ CorsixTH discrepancy. This PR  #2748)

	Inspector arrived to reception desk. 🏁 Epidemic cleared.

Inspector going home:

	➖Clock hidden
	➖Timer not ticking
	➖Player can't mark patients
	➖Nurses can't vaccinate patients
	➖Infected don't have icon as no any infected
	➖Infected can't infect as no any infected
	➖Inspector go home
	✔️Result determined 
</details>


CorsixTH/Lua/epidemic.lua
```diff
@@ -695,8 +705,10 @@ function Epidemic:hasNoInfectedPatients()
 end
 
 function Epidemic:tryAnnounceInspector()
+  if not self.coverup_in_progress or not self:inspectorSpawned() then return end
+
   local inspector = self.inspector
-  if inspector and not inspector.has_been_announced and
+  if inspector.has_been_announced and
```
lewri, line 711: Line 711 and 714 don't make sense together. I think you mean `if not inspector.has_been_announced`?

lewri (commented):
Lua documentation formats differ from C++ ones:
https://github.com/CorsixTH/CorsixTH/wiki/CodeDocumentationStyle

(I didn't realise our version was bespoke, thought it was [LDoc](https://lunarmodules.github.io/ldoc/manual/manual.md.html))

This is also getting confusing with #2740. Wait for that PR to resolve because you'll need to rebase changes after that one merges.

CorsixTH/Lua/epidemic.lua
ARGAMX: > Line 711 and 714 don't make sense together. I think you mean `if not inspector.has_been_announced`?

Nice catch. thanks! Fixed

ARGAMX:
> This is also getting confusing with #2740. Wait for that PR to resolve because you'll need to rebase changes after that one merges. (I stopped reviewing this early)

Yeap. I am already want make yet another one more pull request on the top of this one. So there will be 3 or more related to epidemics I think.
And each one is blocked by the previous one.
There are a lot of things that need to be align with vanilla TH in epidemics in my opinion.
And unfortunately, these changes do not intersect with each by meaning, but they do intersect in code.
I have played TH epidemics and CTH epidemics and there a totally different.
At the same time I want move further and to switch on testing "training overhaul", analyze stuff happiness logic / rise requests discrepancies and overcharging discrepancies. 🤪

CorsixTH/Lua/epidemic.lua
```diff
@@ -101,25 +97,16 @@ end
...
-    self:tryAnnounceInspector()
-  end
+  self:infectOtherPatients()
+  self:checkIfReadyToReveal()
+  self:showAppropriateAdviceMessages()
+  self:tryAnnounceInspector()
+  self:markedPatientsCallForVaccination()
+  self:checkInfectedLeftHospital()
+  self:checkNoInfectedPatients()
   self:checkPatientsForRemoval()
```
lewri, line 107: I somewhat kind of miss the pre-function checks in tick a little. 

Moreso because some of your other functions now repeat `if not self.coverup_in_progress or self:inspectorSpawned() then return end`.

Think you also repeat another check too now. If you wanted to make sure those functions aren't executing out of cycle you should use an `assert` (though asserts (rightly) throw the game into error).

lewri (commented):
I still agree with the changes here for the time the Inspector's outcome gets generated.

CorsixTH/Lua/epidemic.lua
ARGAMX: > I somewhat kind of miss the pre-function checks in tick a little.
> 
> Moreso because some of your other functions now repeat `if not self.coverup_in_progress or self:inspectorSpawned() then return end`.
> 
> Think you also repeat another check too now. If you wanted to make sure those functions aren't executing out of cycle you should use an `assert` (though asserts (rightly) throw the game into error).

I think this should be fine to call these function outside of tick cycle.
I saw in many other places here similar checks were inside functions.
And here they was outside for unknown reasons for me.
And it seemed logical to keep them inside, since you can call these functions in other places too without worrying for conditions.
In that case you don't need to think should you also wrap that function call inside checks or not.
And it is unclear why you need to describe these checks for calling this function several times everywhere, if you can describe check inside the function itself once.
Also, several levels of nesting of conditions was seemed ugly here. 
And these nested checks in that tick function were not very convenient to maintain during work with epidemics.
Now these checks is as part of functions itself.
Yes, some checks are now repeated, that's true.
But I believe these check execution is incredibly cheap. 
I try to write code for people to read and maintain, not for computers in order to save a trillionth of a second. 🙂
So I don't understand why we should miss them here. 🙂 🙂 🙂

lewri: In Lua, code execution is generally quite cheap I agree. Though it's certainly still not a habit to foster.

I've spent some time thinking about this more and there is discomfort in this approach.

As an example. A parent function `P` calls processes `A`, `B`, and `C`. For process `A`, `B` and `C`, they all rely on the same prerequisite of `x` to be `true`. So we ask each process to individually check that same prerequisite inside their code. That's 3 separate checks of the same thing.

Now lets say another programmer makes a change to the prerequisite for `A`, `B` and `C` and also rewrites part of process `B`. Now that programmer updates the prerequisite for `B` but isn't aware that `A` and `C` also rely on the same thing; they weren't looking there so had no idea. Their code doesn't execute properly, and it takes longer to go back and work out what went wrong. 

Now to avoid this, for the parent function `P`, we could instead have a single point of truth by checking this prerequisite only once as a nested `if` on `A`, `B`, and `C`. The nest in this case is pretty simple because we're asking a question, and if it is true then we call `A`, `B`, and `C`. We're not trying to execute additional code outside of these calls either so the nest does not get any more complicated.

So I'm going to advocate putting the checks back in tick. However, I'm open to some middle ground: put your check into its own function to go back to.

```lua
--! Private function to check if a cover up is in progress
function Epidemic:_isCoverUpActive()
  return self.coverup_in_progress and not self:inspectorSpawned()`
end
```
You'll have to account for this in the updated calls:
`
if not self:_isCoverUpActive() then return end
`

However, there is also some logic here that is not you (e.g. why is `coverup_in_progress` not set to `false` on `Epidemic:finishCoverup`?)

ARGAMX: Lewri Thank you for such detailed examples. 😊

> In Lua, code execution is generally quite cheap I agree. Though it's certainly still not a habit to foster.
> 
> I've spent some time thinking about this more and there is discomfort in this approach.

✌️
I agree that this is not a perfect approach. But I see even more flaws in the others.

> 
> As an example. A parent function `P` calls processes `A`, `B`, and `C`. For process `A`, `B` and `C`, they all rely on the same prerequisite of `x` to be `true`. So we ask each process to individually check that same prerequisite inside their code. That's 3 separate checks of the same thing.

I believe this is normal to ask each process to individually check his own prerequisite inside their code, even if other process have the same prerequisite. With proper abstraction and encapsulation of code, this is happens. According to the rules of abstraction, methods should not know about each other internal and somehow depend on what is written in another method internal. 🙂

👌
But Okay. Let's say code will check not 10 times but 1. 1 and 10 is constants. What we will achieve by optimisation this? This is optimisation on constants. "penny wise and pound foolish". In terms of algorithm optimization, this is hardly a significant optimization.
So I just don't understand what we're fighting for here. Maybe the code duplication bothers us? 🙂

Also, as you reasonably suggested: A parent function `P` calls child processes `A`, `B`, and `C`. I like this example. 👍
Each processes has his own prerequisite. Prerequisite in `A`, `B`, and `C` can be identical. But may be not. 
But in the way you suggest, if one child will change his prerequisite, this mean that you have come to parent `P` and modificative it - but as far as I know this is not how to write code according to the rules of writing good and maintainable code...

If you have the long chain of such dependencies: `P` calls `A`, `A` calls `B`, `B` calls `X`, `Y`, `Z`, where `P` is on the top layer on the app and `X`, `Y`, `Z` in the bottom, then by this philosophy you will be tempted to move prerequisite to from `X`, `Y`, `Z` to `P`. But if prerequisite of `X`, `Y`, `Z` will change, you will need to modificative top layer because of change on the bottom layer - which looks strange, especially when different developers works on separated levels. 🙂

So I believe this argument is unconvincing. On the contrary, it confirms that this should not be done this way. 🙂

> 
> Now lets say another programmer makes a change to the prerequisite for `A`, `B` and `C` and also rewrites part of process `B`. Now that programmer updates the prerequisite for `B` but isn't aware that `A` and `C` also rely on the same thing; they weren't looking there so had no idea. Their code doesn't execute properly, and it takes longer to go back and work out what went wrong.

Understand. 👍

But what if these prerequisites were different? Then this rule of writing code would not be applied? It turns out that generalities depend on particulars, and not vice versa? 🙂

I think it is important to understand here that these are internal prerequisites of these functions, not external ones. Even though they were originally coded from the outside in the body function tick, this does not mean that it was done wisely.

These childs depend on the same prerequisite by coincidence. They might not depend on the same prerequisite. These are their internal prerequisite and it should have been written inside, as we always do in other functions. But you propose to move away from this way by designing what is beyond them. The tick method should not be connected with internal kitchen of these functions. Tick shouldn't know the epidemics logic.🙂

So I believe Programmer doesn't need to know or be aware of that until he will open B in in intention to modify it. 🙂

> 
> Now to avoid this, for the parent function `P`, we could instead have a single point of truth by checking this prerequisite only once as a nested `if` on `A`, `B`, and `C`. The nest in this case is pretty simple because we're asking a question, and if it is true then we call `A`, `B`, and `C`. We're not trying to execute additional code outside of these calls either so the nest does not get any more complicated.

I see.. 🙄

But what if we want to call some of these functions each in another separate place? Will we wrap each call with checks outside as was in tick method? But what if a dev doesn't know that these functions have conditions that need to be checked before call outside? It turns out that in order to call them, he needs to have additional not obvious knowledge. Maybe we shouldn't assume that he should know this? 🙂

> 
> However, I'm open to some middle ground: put your check into its own function to go back to.
> 
> ```lua
> --! Private function to check if a cover up is in progress
> function Epidemic:_isCoverUpActive()
>   return self.coverup_in_progress and not self:inspectorSpawned()
> end
> ```

Wow
Okey. Good. Yes, I can do that.
...
That's very wise and constructively proposal in this discussion. Thank you. 🤝
YOU ARE GREAT. 🤗

I did it 👍

> 
> However, there is also some logic here that is not you (e.g. why is `coverup_in_progress` not set to `false` on `Epidemic:finishCoverup`?)

Yes, that's a reasonable point. I noticed that too. But I can't set to false it as Its still need to be true in `tryAnnounceInspector`.
Here in this class it was not customary to reset the flags back. But I didn't refract this for now.

Alberth289346: Not being able to do the logical thing usually means there is something wrong somewhere in the logic or in the meaning of variables. Finding out what is wrong can be non-trivial though.

If you cannot figure it out, at least document why the logical thing is not where you'd expect it. Such a comment avoids future programmers from thinking it's a simple mistake that should be corrected, and makes clear something funky is going on that they should be aware of.

For completeness also add a comment at the spot where the coverup does get completely  finished, again telling the reason why it's there and not earlier.

ARGAMX: > Not being able to do the logical thing usually means there is something wrong somewhere in the logic or in the meaning of variables. Finding out what is wrong can be non-trivial though.
> 
> If you cannot figure it out, at least document why the logical thing is not where you'd expect it. Such a comment avoids future programmers from thinking it's a simple mistake that should be corrected, and makes clear something funky is going on that they should be aware of.
> 
> For completeness also add a comment at the spot where the coverup does get completely finished, again telling the reason why it's there and not earlier.

Sounds reasonable and I generally agree.

The problem here is that this variable does not mean what it is named.

It was not added by me, so I can't just rename it.

And to fix this, I need to make changes that, in my opinion, are outside the scope of what this PR was supposed to fix.

I have another PR on the epidemic, which does not fix any bugs, only refactoring. I have an idea to try to fix it there. But it's not for 100%...

But I can't create next PR until this one is uploaded. They intersects.

Regarding adding a comment - a reasonable idea. 

But in the comment of the declaration of this variable there is a correct description of what this variable actually means. So I have nothing to add in terms of commentary.

Alberth289346: > The problem here is that this variable does not mean what it is named.
It was not added by me, so I can't just rename it.

Sure you can, why not?
As long as it's local in a function, renaming is not much of a problem.
Lots of variables have bad names in this project. Too much copy/paste similar code without adapting it properly.

Chances are that the author has long ago left.

ARGAMX: > Sure you can, why not? As long as it's local in a function, renaming is not much of a problem. Lots of variables have bad names in this project. Too much copy/paste similar code without adapting it properly.
> 
> Chances are that the author has long ago left.

Nope, unfortunately `coverup_in_progress` is not local and is used in 3 more outside classes that I did not touch in this PR. 

To be honest, I also don’t like that part of the epidemic’s functionality is spread across other classes. But in this PR I just wanted to correct the inconsistencies in the logic related to the original game. nothing more.

Also this will require an afterLoad which is also not supposed within the scope of this PR.

I will try to do this, but in next PR and not today. 🙂

lewri (approved):

PR #2752, opened 2024-12-27
Cleanup/"fix" the UseObjectAction

Alberth289346:
- Adds comment, some lua docs, and a few empty lines for better readability.
- Replaces some `DrawFlags` magic constants.
- Fixes the logic for required minimal use (does not affect the only case that uses it, since the dna fixer has 1 frame (ie any number of frames is correct), and `20 % 10 == 0` so nothing is actually added).
  To understand how it's broken, the old code was:
  ``` diff
  - local length = anim_length * humanoid.world:getAnimLength(anim)
  - if action.min_length and phase == 0 and action.min_length > length then
  -   -- A certain length is desired.
  -   -- Even it out so that an integer number of animation sequences are done.
  -   length = action.min_length + action.min_length % length
  - end
   ```
  - `anim_length` is the number of animations defined with the object.
  - `length` is the number of frames for `anim_length` animations.
  - In the `in_use` phase, if the action requires more frames: Take the number of frames for the action + some fraction of it.
  - This is only an "integer number of animation sequences" if the `action.min_length` is that as well. In general however, such an implicit connection is undesired. It may break as soon as the animation is changed.
  - However, if the `action.min_length` has the property above, then why does it lengthen the sequence at all, given that `action.min_length > length` already holds?

### DIFF
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -227,7 +227,7 @@ function Machine:createHandymanActions(handyman)
 
   local repair_action = UseObjectAction(self):setProlongedUsage(false)
       :setAfterUse(repair_after_use)
-  repair_action.min_length = 20
+  repair_action.min_length = 20 -- Minimum number of frames needed for the action.
 
   if handyman_room and handyman_room ~= this_room then
     handyman:setNextAction(handyman_room:createLeaveAction())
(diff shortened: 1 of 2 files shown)


Alberth289346:
lpeg download error: `error: https://www.inf.puc-rio.br/~roberto/lpeg/lpeg-1.1.0.tar.gz: WinHttpSendRequest failed with exit code 12002`

lewri (approved):

lewri:
@TheCycoONE is this one safe to merge?

TheCycoONE:
Yes

PR #2753, opened 2024-12-28
Fix for: Deactivate epidemics cheat doesn't work.

### LINKED ISSUE #2718: Deactivate epidemics cheat doesn't work
Has only happened since update to 0.68 on 14th October.
Will click 'Deactivate epidemics', a box pops up telling me so and the receptionist announces that I'm cheating.  But I still get epidemics.
cheat

ARGAMX:
*Fixes #2718*

Does not intersect with PR #2748

- Fixed a bug due to which disabling epidemics cheat doesn't work.
- Also added cancelation for any already preparing epidemics.
- Also added cancelation for any already ongoing epidemic:
- - Turn vaccination mode off if enabled.
- - Remove epidemic timer if visible.
- - Send inspector home if spawned.
- - Clear epidemic patients.
- - Remove `epidemic discovered fax` from `bottom_panel` if fax pending to open.
- Fax `owner` now can be not only `humanoid`, but any object, as `epidemic`.


tobylane (approved):
The check removed in `Cheats:cheatEpidemic` could be remade as below, but it's not important. Looks good, plays right.

```lua
function Hospital:spawnContagiousPatient()
  if self.epidemics_disabled then return false, _S.misc.epidemics_off end
```

ARGAMX:
> The check removed in `Cheats:cheatEpidemic` could be remade as below, but it's not important. Looks good, plays right.
> 
> ```lua
> function Hospital:spawnContagiousPatient()
>   if self.epidemics_disabled then return false, _S.misc.epidemics_off end
> ```

I thought about it, but I have not found any reasonable benefits for this condition. 

If add this condition then it's possible that the user press that button and tries to generate an infectious patient, but it doesn't generates. Then it looks like the cheat don't work. (bug)

It seems like if the user turns off epidemics, it means that he doesn't want the game to generate epidemics itself. Not that the game should block his ability to create an epidemic himself. (For what?)

CorsixTH/Lua/hospital.lua
```diff
@@ -75,6 +75,8 @@ function Hospital:Hospital(world, avail_rooms, name)
   -- epidemics. If epidemic_limit = 1 then only one epidemic can exist at a
   -- time either in the futures pool or as a current epidemic.
   self.concurrent_epidemic_limit = level_config.gbv.EpidemicConcurrentLimit or 1
+  -- For Cheat - flag determines possibility of epidemics
+  self.epidemics_disabled = false
```
lewri, line 79: As a general note, this declaration isn't needed; since nil also equals false for a boolean check. We also don't pre-assign a value to `hospital.cheated` or `world.earthquake.disabled`.
Old saves will still use nil anyway.

CorsixTH/Lua/cheats.lua
```diff
@@ -129,18 +129,17 @@ end
 
 --[[ Creates a new contagious patient in the hospital - potentially an epidemic]]
 function Cheats:cheatEpidemic()
-  if not self.hospital.epidemics_disabled then
-    return self.hospital:spawnContagiousPatient()
-  end
+  return self.hospital:spawnContagiousPatient()
```
lewri, line 132: We're now generating a contagious patient even when we set epidemics to be disabled? Ah no, please don't do that.

Toby's suggestion passes the failure reason across to the player as a UIInformation dialog. They do receive feedback:
https://github.com/CorsixTH/CorsixTH/blob/b2fe51272b46d43911cb1886e8d4a5c086205e5c/CorsixTH/Lua/dialogs/resizables/cheats_dialog.lua#L130-L144
As an example of what the player will see:
image

lewri (changes requested):

CorsixTH/Lua/hospital.lua
ARGAMX: > As a general note, this declaration isn't needed; since nil also equals false for a boolean check. We also don't pre-assign a value to `hospital.cheated` or `world.earthquake.disabled`. Old saves will still use nil anyway. Read: Not important -- just a nit

True. 👌
Yes, I understand.
But I wanted to add a comment to this flag. What it is and what it is for.
So I had to add an declaration here.

CorsixTH/Lua/cheats.lua
ARGAMX: Oh, Lewri thank you very much.
The screenshot was very helpful.
I really didn't understand Toby suggesting.
This is really much better. 👍

lewri (approved):

PR #2754, opened 2024-12-31
Add soundfont setting to config

### LINKED ISSUE #2747: Setting to use custom soundfont when using built-in midi playback

It would be nice to have the ability to point Corsix to a custom SF2 soundfont to use for MIDI playback via the settings menu or config file, now that playing on actual MIDI devices isn't supported.

Since it looks like the game is already using an external soundfont file, allowing the user to specify their own soundfont could be a fairly easy way to bring back some of that sound customizability. I for one _heavily_ dislike how some of the base game's soundtrack sounds through FluidR3 because of the wonky mixing on certain instruments.

No

0.68.0

No

### Additional Information

_No response_

lewri:
*Fixes #2747*

- SoundFont is now an addressable configuration option with specified caveats. Wiki update to follow.

I'm pretty sure this change won't apply to existing config files because the setting is nil, however. It could be false but that would break how we format folder settings.

### DIFF
--- a/CorsixTH/Lua/config_finder.lua
+++ b/CorsixTH/Lua/config_finder.lua
@@ -468,6 +468,15 @@ screenshots = nil -- [[X:\ThemeHospital\Screenshots]]
 --
 audio_music = nil -- [[X:\ThemeHospital\Music]]
 
+-------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Alberth289346:
LGTM

PR #2755, opened 2025-01-08
Fix advice when exceeding maximum buy object limit

### LINKED ISSUE #2689: [Bug] Incorrect adviser message when exceeding maximum buy object limit
### Describe the issue

When buying objects, if the player attempts to exceed the maximum number of objects in a single purchase (99 items), the Adviser will claim that "You don't have enough money in the bank to make that purchase!" regardless of the player's actual bank balance.
2024-10-20 14_44_38-CorsixTH

### Steps to Reproduce

1. Start a new level (so that you have plenty of money).
2. Click the Furnish Corridor button (or start building any room).
3. Add 99 plants (Ctrl+left click will add 10 at a time).
4. Try to increase the number of plants to 100.

### Expected Behaviour

The Adviser does not say "you don't have enough money in the bank to make that purchase!" (Since the player has selected the maximum limit of objects, their bank balance is irrelevant.)

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

### CorsixTH Version

0.68

### Operating System

Windows 10

### Theme Hospital Version

CD

### Gamelog.txt

_No response_

### Additional Information

Found while testing #2687

FMMazur:
*Fixes #2689*

- Advise if can't buy more object;
- Add new localization string in English and Brazilian Portuguese.


CorsixTH/Lua/dialogs/furnish_corridor.lua
```diff
@@ -137,7 +140,7 @@ function UIFurnishCorridor:purchaseItem(index, quantity)
   if quantity < o.min_qty then
     quantity = o.min_qty
   elseif quantity > 99 then
-    quantity = 99
```
ARGAMX, line 140: With that change if player hold `ctrl` or `shift` while press `+` he can't reach 99 units.

FMMazur: thx

MandrakeHorse:
I tried testing your changes (by copying furnish_corridor.lua into my CorsixTH\Lua\dialogs folder) and have found a bug. When I click the + icon for an item, the Total increases, but the number of items selected does not.
image

FMMazur:
@MandrakeHorse Did you try it before or after the last commit? I tried to replicate as you said, but couldn't do it.

FMMazur:
@ARGAMX Solved, the problem was that i was checking why the value would become a float when subtracting from the current object quantity, and when commiting I also removed the original code, sorry about that.

CorsixTH/Lua/dialogs/furnish_corridor.lua
```diff
@@ -139,11 +147,14 @@ function UIFurnishCorridor:purchaseItem(index, quantity)
...
+
+  local bought = quantity - o.qty
   local hospital = self.ui.hospital
-  if hospital.balance >= self.total_price + quantity * hospital:getObjectBuildCost(o.object.id) or is_negative_quantity then
-    o.qty = o.qty + quantity
-    self.total_price = self.total_price + quantity * hospital:getObjectBuildCost(o.object.id)
+  local total_price = self.total_price + bought * hospital:getObjectBuildCost(o.object.id)
+
+  if hospital.balance >= total_price or is_negative_quantity then
+    o.qty = math.tointeger(o.qty + bought)
```
lewri, line 156: This math function isn't supported by CorsixTH. We use the Lua 5.1 base
https://www.codecademy.com/resources/docs/lua/mathematical-library/tointeger

However, I'm not really sure how this number is not remaining an integer.

lewri (changes requested):
Thanks very much for taking a crack at this :)

I need to look further into the quantity issue from the first commit, unless someone else can see what's going on.

However, because you're editing a persistable it'd be advisable to add an afterload to close this window and bump savegame version. I can contribute a commit for the afterload if you're not sure how.

FMMazur:
I"ll gladly accept your help, as I'm new to the project and need to learn more about it

MandrakeHorse:
@FMMazur No worries! I've retested using your latest versions of `english.lua` and `furnish_corridor.lua`, and can confirm that the Adviser will pop up with a different message if you attempt to order more than 99 items, even if you can afford it. He'll also continue to pop up with his "You don't have enough money in the bank to make that purchase!" message if you can't afford to buy another item.

Some feedback:
- The grammar of the Adviser's message is slightly incorrect, since it's referring to each Item in the singular rather than the plural (e.g. "Radiator" instead of "Radiators", "Plant" instead of "Plants" and so forth). Normally, you would simply add an "s" to the end, except we'd also need to account for the plural of "Bench" (i.e. "Benches"), plus we'd have to consider translating this string into different languages.
- However, the easiest solution is to change the string for `cannot_buy` to `"You can't buy more of that item!"`, which avoids the problem entirely.
2025-01-09 20_55_45-CorsixTH

CorsixTH/Lua/dialogs/furnish_corridor.lua
FMMazur: My bad, the issue is that I have lua 5.4 installed, and the CMake configuration doesn't guarantee nor checks if the lua version is the 5.1, i think it would be better if we forced the download of the correct lib, or at least show some error

lewri:
@FMMazur I've run the code here. The `math.tointeger()` isn't needed, bit weird how Mandrake and Argamx got the same issue though...

Also we don't need an afterload 👍

FMMazur:
@lewri Strange, i will try to change my version to 5.1 and test, but without the `math.integer` the number becomes a float when removing an item, at least on 5.4

lewri:
> @lewri Strange, i will try to change my version to 5.1 and test, but without the `math.integer` the number becomes a float when removing an item, at least on 5.4

Do you get floats if you add `print("bought", bought)` below your declaration at line 151; and `print("o.qty", o.qty)` below your line 157 where you update o.qty?

CorsixTH supports LuaJIT too, which is also part of why we are stuck (mostly) with 5.1.

FMMazur:
As you can see, for some reason when you load a save this float situation happens, but if you try a new game it doesn't maybe we can use math_floor to solve it. @lewri 


https://github.com/user-attachments/assets/d38cb8a3-583a-411d-8920-3247cb1d9e98

lewri:
@FMMazur some good news, the float thing is not a bug you created. I'll see if I can wrangle it to behave.

lewri:
@FMMazur do you mind me pushing a commit into this PR to fix?

FMMazur:
@lewri no problem

ARGAMX:
> @ARGAMX @MandrakeHorse please feel free to retest the PR again if you wish.

I retested and now it works great for me. 👍

FMMazur:
@lewri if it's ready to go, i can remove the RFC

lewri:
Looks fine to me.

MandrakeHorse:
Here's what the new adviser message looks like while playing!
2025-01-12 17_31_55-CorsixTH

FMMazur:
@lewri I think it's all good now

PR #2756, opened 2025-01-09
Include the users campaign dir in looking for campaigns

### LINKED ISSUE #2742: [Bug] `campaigns` setting in config file does not work.
### Describe the issue

Config-file fragment:
image

`campaigns` variable is not set, the text claims it will look next to the config file (I checked with `config_finder.lua` and the text is still valid).

* Note the `~/.config/..` path at the top.
* I get the same results when I set the `campaigns` configuration file variable to the `Campaign` directory. (ie setting or not setting it makes no difference, it seems.)

----
``` sh
$ cd ~/.config/CorsixTH 
$ ls Campaigns/mazes 
'maze1 EASY.level'   maze1.map		 maze2.level  'maze3 EASY.level'   Maze3.map
 maze1.level	    'maze2 EASY.level'	 maze2.map     maze3.level	   mazes.campaign
```

Note there is a `mazes` campaign next to the config file.

----
Added some debugging prints, see further down for the exact place. Added the class and function name to the output to give a rough idea.

Started the application, main menu appears, debug output:
``` text
App:getFullPath([[Levels]]
App:getFullPath([[Campaigns]]
APP:init campaign_dir = /newhome/alberth/dev/corsix-th/corsix_alberth_fork/CorsixTH/Campaigns/
App:setUserDir user_campaign_dir = /newhome/alberth/.config/CorsixTH/Campaigns
App:setUserDir user_campaign_dir = /newhome/alberth/.config/CorsixTH/Campaigns/
App:getFullPath({[[Lua]],[[languages]]}
App:getFullPath({[[Lua]],[[walls/]]}
App:getFullPath({[[Lua]],[[objects/]]}
App:getFullPath({[[Lua]],[[objects/machines/]]}
App:getFullPath({[[Lua]],[[objects/doors/]]}
App:getFullPath({[[Lua]],[[rooms/]]}
App:getFullPath({[[Lua]],[[humanoid_actions/]]}
App:getFullPath({[[Lua]],[[diseases/]]}
App:getFullPath({[[Lua]],[[diagnosis/]]}
App:getFullPath({[[Lua]],[[dialogs/]]}
App:getFullPath({[[Lua]],[[dialogs/fullscreen/]]}
App:getFullPath({[[Lua]],[[dialogs/resizables/]]}
App:getFullPath({[[Lua]],[[dialogs/resizables/menu_list_dialogs/]]}
App:getFullPath({[[Lua]],[[dialogs/resizables/file_browsers/]]}
```
Path does seem to get picked up in app as `user_campaign_dir`.

Open the custom campaign window, debug output:
``` text
App:getFullPath([[Campaigns]]
custom_campaign.lua - looking for .campaign file at: /newhome/alberth/dev/corsix-th/corsix_alberth_fork/CorsixTH/Campaigns/
App:getFullPath({[[Campaigns]],[[ChrizmanTV.campaign]]}
App:readCampaignFile: path=/newhome/alberth/dev/corsix-th/corsix_alberth_fork/CorsixTH/Campaigns/ChrizmanTV.campaign
App:getFullPath({[[Campaigns]],[[example.campaign]]}
App:readCampaignFile: path=/newhome/alberth/dev/corsix-th/corsix_alberth_fork/CorsixTH/Campaigns/example.campaign
```
It seems that it doesn't search in the `~/.config/...` campaigns
The custom campaign window also only shows the two found campaigns.

----

``` diff
diff --git a/CorsixTH/Lua/app.lua b/CorsixTH/Lua/app.lua
index 4848b880..79e96e21 100644
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -84,6 +84,7 @@ end
 --!param trailing_slash (boolean) Whether the path needs to end with a local path separator
 --!return fullpath (string) The OS dependent full path
 function App:getFullPath(folders, trailing_slash)
+  print("App:getFullPath(" .. tostring(serialize(folders)))
   if type(folders) ~= "table" then folders = {folders} end
   local ending = trailing_slash and pathsep or ""
   return debug.getinfo(1, "S").source:sub(2, -12) .. table.concat(folders, pathsep) .. ending
@@ -123,6 +124,7 @@ function App:init()
   self.good_install_folder = good_install_folder
   self.level_dir = self:getFullPath("Levels", true)
   self.campaign_dir = self:getFullPath("Campaigns", true)
+  print("APP:init campaign_dir = " .. self.campaign_dir)
   self:initUserDirectories()
   self:initSavegameDir()
   self:initScreenshotsDir()
@@ -428,10 +430,15 @@ function App:initUserDirectories()
 
   self.user_level_dir = self.config.levels or
       conf_path:match("^(.-)[^" .. pathsep .. "]*$") .. "Levels"
+
   self.user_level_dir = setUserDir(self.user_level_dir, "User Levels")
   self.user_campaign_dir = self.config.campaigns or
       conf_path:match("^(.-)[^" .. pathsep .. "]*$") .. "Campaigns"
+  print("App:setUserDir user_campaign_dir = " .. self.user_campaign_dir)
+
   self.user_campaign_dir = setUserDir(self.user_campaign_dir, "User Campaigns")
+  print("App:setUserDir user_campaign_dir = " .. self.user_campaign_dir)
+
   self.user_log_dir = self.config.logs or
       conf_path:match("^(.-)[^" .. pathsep .. "]*$") .. "Logs"
   self.user_log_dir = setUserDir(self.user_log_dir, "Gamelogs")
@@ -620,6 +627,7 @@ end
 --!return (table) Definitions found in the campaign file.
 function App:readCampaignFile(campaign_file)
   local path = self:getFullPath({"Campaigns", campaign_file})
+  print("App:readCampaignFile: path=" .. path)
   local chunk, err = loadfile_envcall(path)
   if not chunk then
     return nil, "Error loading " .. path .. ":\n" .. tostring(err)
@@ -684,8 +692,11 @@ function App:getAbsolutePathToLevelFile(level)
     self.campaign_dir,
     self.level_dir,
   }
+  print("App:getAbsolutePathToLevelFile: user_campaign_dir=" .. self.user_campaign_dir)
+  print("App:getAbsolutePathToLevelFile: campaign_dir=" .. self.campaign_dir)
   for _, parent_path in ipairs(paths_to_search) do
     local check_path = parent_path .. pathsep .. level
+    print("App:getAbsolutePathToLevelFile " .. check_path)
     local file, _ = io.open(check_path, "rb")
     if file then
       file:close()
diff --git a/CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua b/CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
index affa24ce..2ed86f50 100644
--- a/CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
+++ b/CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
@@ -38,6 +38,7 @@ local details_width = 280
 local function createCampaignList(path)
   local campaigns = {}
 
+  print("custom_campaign.lua - looking for .campaign file at: " .. path)
   for file in lfs.dir(path) do
     if file:match("%.campaign$") then
       local campaign_info, err = TheApp:readCampaignFile(file)
```

### Steps to Reproduce

1.  Have a `Campaign` directory with a campaign in it, next to the config file.

2. Start the game, and open the custom campaign window.

3. Only find the standard campaigns provided with CTH listed in the window.

### Expected Behaviour

It finds the mazes campaign or prints an error if something is wrong with it.

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

### CorsixTH Version

101a4de3f602aca55240feba01871e1382f74fbc

### Operating System

Linux Mint 22

### Theme Hospital Version

CD

### Gamelog.txt

```yaml
---------------------------------------------------------------

Welcome to CorsixTH Trunk!

---------------------------------------------------------------

This window will display useful information if an error occurs.

---------------------------------------------------------------
```

### Additional Information

_No response_

tobylane:
*Fixes #2742*

- Adds a new function to look for a campaign file in the users campaign folder, as well as the Corsixth folder, copied from the getAbsolutePathToLevelFile function above it. 
- Add the users campaign folder to the search in the custom campaigns dialog.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -582,7 +582,7 @@ function App:loadCampaign(campaign_file)
     return
   end
 
-  level_info, errors = self:readLevelFile(campaign_info.levels[1])
+  level_info, errors = self:readLevelFile(campaign_info.levels[1], campaign_info.folder)
   if not level_info then
     self.ui:addWindow(UIInformation(self.ui, { _S.errors.could_not_find_first_campaign_level:format(errors) }))
     return
@@ -600,7 +600,7 @@ function App:loadCampaign(campaign_file)
   campaign_info.winning_text = self.strings:getLocalisedText(campaign_info.winning_text,
       campaign_info.winning_text_table)
 
-  if self:loadLevel(campaign_info.levels[1], nil, level_info.name,
+  if self:loadLevel(level_info.path, nil, level_info.name,
       level_info.map_file, level_info.briefing, nil, _S.errors.load_level_prefix, campaign_info) then
     -- The new world needs to know which campaign to continue on.
     self.world.campaign_info = campaign_info
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -572,7 +572,7 @@ end
 
 --! Loads the first level of the specified campaign and prepares the world
 --! to be able to progress through that campaign.
---!param campaign_file (string) Name of a CorsixTH Campaign definition Lua file.
+--!param campaign_file (string) Full path of a CorsixTH Campaign definition Lua file.
 function App:loadCampaign(campaign_file)
   local campaign_info, level_info, errors, _
 
@@ -615,18 +615,12 @@ end
 
 --! Reads the given file name as a Lua chunk from the Campaign folders
 --! A correct campaign definition contains "name", "description", "levels", and "winning_text".
---!param campaign_file (string) Name of the file to read.
+--!param campaign_file (string) Full path of the file to read.
 --!return (table) Definitions found in the campaign file.
 function App:readCampaignFile(campaign_file)
-  local search_paths = {
-    self.user_campaign_dir,
-    self.campaign_dir,
-  }
-  local path = self:findFileInDirs(search_paths, campaign_file)
-  if not path then return nil, "Campaign not found: " .. campaign_file end
-  local chunk, err = loadfile_envcall(path)
+  local chunk, err = loadfile_envcall(campaign_file)
   if not chunk then
-    return nil, "Error loading " .. path .. ":\n" .. tostring(err)
+    return nil, "Error loading " .. campaign_file .. ":\n" .. tostring(err)
   else
     local result = {}
     chunk(result)
@@ -683,9 +677,9 @@ function App:readLevelFile(level)
 end
 
 --! Searches in the given folders for the given filename
---!param search_paths (table) Set of folders to look through
+--!param search_paths (table) List of folders (strings) to look through
 --!param filename (string) Filename to search for.
---!return (string, error) Returns the found absolute path, or nil if not found
+--!return path (string) The full path of the found file
 function App:findFileInDirs(search_paths, filename)
   for _, parent_path in ipairs(search_paths) do
     local check_path = parent_path .. pathsep .. filename
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -630,16 +630,24 @@ end
 
 --! Opens the given file name and returns all Level definitions in a table.
 --! Values in the returned table: "path", "level_file", "name", "map_file", "briefing", and "end_praise".
---!param level (string) Name of the file to read.
+--!param level (string) Full path or name of the file to read.
+--!param campaign_dir (string) Folder of the parent campaign file
 --!return (table) Level info found in the file.
-function App:readLevelFile(level)
-  local search_paths = {
-    self.user_campaign_dir,
-    self.user_level_dir,
-    self.campaign_dir,
-    self.level_dir,
-  }
-  local filename = self:findFileInDirs(search_paths, level)
+function App:readLevelFile(level, campaign_dir)
+  local filename, level_dir
+  if level:match(pathsep) then -- Full path given
+    filename = level
+    level_dir = level:match("(.*)" .. pathsep)
+  else
+    local search_paths = {
+      campaign_dir or "",
+      self.user_campaign_dir,
+      self.user_level_dir,
+      self.campaign_dir,
+      self.level_dir,
+    }
+    filename = self:findFileInDirs(search_paths, level)
+  end
   local file, err = io.open(filename and filename or "")
   if not file then
     return nil, "Could not open the specified level file (" .. level .. "): " .. err
@@ -651,15 +659,35 @@ function App:readLevelFile(level)
   level_info.path = filename
   level_info.level_file = level
   level_info.name = contents:match("%Name ?= ?\"(.-)\"") or "Unknown name"
-  level_info.map_file = contents:match("%MapFile ?= ?\"(.-)\"")
-  if not level_info.map_file then
+  local map_file = contents:match("%MapFile ?= ?\"(.-)\"")
+  if not map_file then
     -- The old way of defining the Map File has been deprecated, but a warning is enough.
-    level_info.map_file = contents:match("%LevelFile ?= ?\"(.-)\"")
-    if level_info.map_file then
+    map_file = contents:match("%LevelFile ?= ?\"(.-)\"")
+    if map_file then
       print("\nWarning: The level '" .. level_info.name .. "' contains a deprecated variable definition in the level file." ..
         "'%LevelFile' has been renamed to '%MapFile'. Please advise the map creator to update the level.\n")
     end
     level_info.deprecated_variable_used = true
+  end
+  -- Find the full path of the map file, unless it is a CorsixTH additional config or a Theme Hospital map
+  if filename:match(self.level_dir .. "original%d%d%.level") then
+    level_info.map_file = map_file
+  elseif map_file then
+    if map_file:lower():match("^level") then
+      level_info.map_file = map_file
+    else
+      local search_paths = {
+        campaign_dir or "",
+        level_dir or "",
+        self.user_campaign_dir,
+        self.user_level_dir,
+        self.campaign_dir,
+        self.level_dir,
+      }
+      level_info.map_file = self:findFileInDirs(search_paths, map_file)
+    end
+  else
+    return nil, _S.errors.map_file_missing:format(map_file)
   end
 
   -- Pick a localised set of briefings, if available
@@ -682,7 +710,7 @@ end
 --!return path (string) The full path of the found file
 function App:findFileInDirs(search_paths, filename)
   for _, parent_path in ipairs(search_paths) do
-    local check_path = parent_path .. pathsep .. filename
+    local check_path = parent_path .. filename
     local file, _ = io.open(check_path, "rb")
     if file then
       file:close()
@@ -1571,27 +1599,21 @@ function App:readDataFile(dir, filename)
 end
 
 --! Get a level file.
---!param filename (string) Name of the level file.
+--!param path (string) Full path of the level file, or a Theme Hospital level filename
 --!return If the file could be found, the data of the file, else a
 --        tuple 'nil', and an error description
-function App:readMapDataFile(filename)
-  -- First look in the original install directory, if not found there
-  -- look in the CorsixTH directories "Levels" and "Campaigns".
-  local data = self.fs:readContents("Levels" .. pathsep .. filename)
-  if not data then
-    local search_paths = {
-      self.user_campaign_dir,
-      self.user_level_dir,
-      self.campaign_dir,
-      self.level_dir,
-    }
-    local absolute_path = self:findFileInDirs(search_paths, filename)
-    if absolute_path then
-      local file = io.open(absolute_path, "rb")
-      if file then
-        data = file:read("*a")
-        file:close()
-      end
+function App:readMapDataFile(path)
+  local data
+  if path:match(pathsep) then
+    local file = io.open(path, "rb")
+    if file then
+      data = file:read("*a")
+      file:close()
+    end
+  else
+    data = self.fs:readContents("Levels" .. pathsep .. path)
+    if not data then
+      return nil, _S.errors.missing_th_data_file:format(path)
     end
   end
   if data then
@@ -1600,7 +1622,7 @@ function App:readMapDataFile(filename)
     end
   else
     -- Could not find the file
-    return nil, _S.errors.map_file_missing:format(filename)
+    return nil, _S.errors.map_file_missing:format(path)
   end
   return data
 end
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -613,12 +613,17 @@ function App:loadCampaign(campaign_file)
   end
 end
 
---! Reads the given file name as a Lua chunk from the Campaigns folder in the CorsixTH install directory.
+--! Reads the given file name as a Lua chunk from the Campaign folders
 --! A correct campaign definition contains "name", "description", "levels", and "winning_text".
 --!param campaign_file (string) Name of the file to read.
 --!return (table) Definitions found in the campaign file.
 function App:readCampaignFile(campaign_file)
-  local path = self:getFullPath({"Campaigns", campaign_file})
+  local search_paths = {
+    self.user_campaign_dir,
+    self.campaign_dir,
+  }
+  local path = self:findFileInDirs(search_paths, campaign_file)
+  if not path then return nil, "Campaign not found: " .. campaign_file end
   local chunk, err = loadfile_envcall(path)
   if not chunk then
     return nil, "Error loading " .. path .. ":\n" .. tostring(err)
@@ -634,7 +639,13 @@ end
 --!param level (string) Name of the file to read.
 --!return (table) Level info found in the file.
 function App:readLevelFile(level)
-  local filename = self:getAbsolutePathToLevelFile(level)
+  local search_paths = {
+    self.user_campaign_dir,
+    self.user_level_dir,
+    self.campaign_dir,
+    self.level_dir,
+  }
+  local filename = self:findFileInDirs(search_paths, level)
   local file, err = io.open(filename and filename or "")
   if not file then
     return nil, "Could not open the specified level file (" .. level .. "): " .. err
@@ -671,27 +682,19 @@ function App:readLevelFile(level)
   return level_info
 end
 
---! Searches for the given level file in the "Campaigns" and "Levels" folder of the
---! CorsixTH install directory.
---!param level (string) Filename to search for.
---!return (string, error) Returns the found absolute path, or nil if not found. Then
---!       a second variable is returned with an error message.
-function App:getAbsolutePathToLevelFile(level)
-  local paths_to_search = {
-    self.user_campaign_dir,
-    self.user_level_dir,
-    self.campaign_dir,
-    self.level_dir,
-  }
-  for _, parent_path in ipairs(paths_to_search) do
-    local check_path = parent_path .. pathsep .. level
+--! Searches in the given folders for the given filename
+--!param search_paths (table) Set of folders to look through
+--!param filename (string) Filename to search for.
+--!return (string, error) Returns the found absolute path, or nil if not found
+function App:findFileInDirs(search_paths, filename)
+  for _, parent_path in ipairs(search_paths) do
+    local check_path = parent_path .. pathsep .. filename
     local file, _ = io.open(check_path, "rb")
     if file then
       file:close()
       return check_path
     end
   end
-  return nil, "Level not found: " .. level
 end
 
 --! Invokes a protected call of App:_loadLevel(...). See that function for more information.
@@ -1582,7 +1585,13 @@ function App:readMapDataFile(filename)
   -- look in the CorsixTH directories "Levels" and "Campaigns".
   local data = self.fs:readContents("Levels" .. pathsep .. filename)
   if not data then
-    local absolute_path = self:getAbsolutePathToLevelFile(filename)
+    local search_paths = {
+      self.user_campaign_dir,
+      self.user_level_dir,
+      self.campaign_dir,
+      self.level_dir,
+    }
+    local absolute_path = self:findFileInDirs(search_paths, filename)
     if absolute_path then
       local file = io.open(absolute_path, "rb")
       if file then
--- a/CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
+++ b/CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
@@ -40,7 +40,8 @@ local function createCampaignList(paths_table)
   for _, folder in pairs(paths_table) do
     for file in lfs.dir(folder) do
       if file:match("%.campaign$") then
-        local campaign_info, err = TheApp:readCampaignFile(file)
+        local full_path = folder .. file
+        local campaign_info, err = TheApp:readCampaignFile(full_path)
         if not campaign_info then
           print(err)
         else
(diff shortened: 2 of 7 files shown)


CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
```diff
@@ -33,40 +33,41 @@ local col_scrollbar = {
...
+  table.sort(campaigns, function(a,b) return a.name < b.name end)
   return campaigns
 end
 
 function UICustomCampaign:UICustomCampaign(ui)
   self.label_font = TheApp.gfx:loadFont("QData", "Font01V")
 
-  local path = TheApp:getFullPath("Campaigns", true)
-  local campaigns = createCampaignList(path)
+  local campaigns = createCampaignList({ui.app.campaign_dir, ui.app.user_campaign_dir})
```
Alberth289346, line 70: I wonder if first trying the user campaign directory would be better. A user may have a modified campaign file that they want to try. In the current order they would need to modify the game files rather than their own user files.

At some systems (eg the Unix file system hierarchy), the former may be stored in a read-only (for the user) systemwide directory like `/usr`.

CorsixTH/Lua/app.lua
```diff
@@ -695,6 +695,27 @@ function App:getAbsolutePathToLevelFile(level)
...
+    self.campaign_dir,
+  }
+  for _, parent_path in ipairs(paths_to_search) do
+    local check_path = parent_path .. pathsep .. campaign
+    local file, _ = io.open(check_path, "rb")
+    if file then
+      file:close()
+      return check_path
+    end
+  end
```
Alberth289346, line 715: Copy-paste is oh so convenient, but pretty much always a bad idea. De-duplicating such code afterwards is horribly complicated (try it a few times :) ).

Instead refactor the checking code to its own (private) function, and use it form both spots.

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
Alberth289346: Sorry, order is not relevant, and you do that in trying to load them.

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
```diff
@@ -33,40 +33,41 @@ local col_scrollbar = {
...
+                 campaign_info.description_table)
+            }
+          else
+            print("Warning: Loaded campaign that had no levels specified")
+          end
         end
       end
     end
   end
+  table.sort(campaigns, function(a,b) return a.name < b.name end)
```
Alberth289346, line 63: We get duplicate names if the same campaign exists in both the user and the game directories.

tobylane: As this is the name from reading the file, it's already possible from two files in the same folder. Should they be given suffixes, like (2)?

tobylane: Numbered suffixes don't help tell them apart, adding the filename to the description of clashes doesn't look good.

Alberth289346:
Tried this patch. It fails because I have a `mazes` directory in the `Campaigns` directory. Inside the `mazes` directory are the campaign files.

After copying those into the `Campaigns` directory, the campaign is found and started.

How to organize campaign files is a different problem from scanning the `Campaigns` directory, and as such, the patches solves the reported problem.

This problem should then probably be considered in a next issue, since I am pretty likely not the only person that wants to organize their data differently compared with throwing it all in one directory.

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
Alberth289346: For now, ignore all campaigns with the same name except the first one, dump a log entry with the details, and give a window that says some campaigns were not added due to having duplicated names?

Alberth289346:
Apparently, @lewri has previous art with respect to directories in #2209

tobylane:
Refactored the two almost identical functions. Added an information message and detail in console for duplicate campaign names. The UIInformation window is set to be on top, but is below the campaigns dialog.

CorsixTH/Lua/map.lua
```diff
@@ -252,7 +252,14 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
...
     assert(base_config, "No base config has been loaded!")
-    errors, result = self:loadMapConfig(self.app:getAbsolutePathToLevelFile(level), base_config, true)
+    local search_paths = {
+      self.app.user_campaign_dir,
+      self.app.user_level_dir,
+      self.app.campaign_dir,
+      self.app.level_dir,
+    }
+    local filename = self.app:findFileInDirs(search_paths, level)
+    errors, result = self:loadMapConfig(filename, base_config, true)
```
Alberth289346, line 262: You can load a config from a different spot than the map file. That makes it difficult to predict what files are going to be used.

In my view, if you load a campaign from `self.app.campaign_dir` the map and the config should also come from there.

Not sure what should happen if some files are missing, but things should be predictable for a user and level designer.

Also, preferably it should be stable under adding other files (but not sure this is achievable). Eg if a level file was released and it didn't have config, we used file X. If I then add a new version of the level and that one does have a config, then the new version should use that config, but the old one perhaps not, since the entire experience of the level could change.

tobylane:
Moved the warning to the dialog, where it's visible. Rebased.

I added and removed another commit, but in writing and rebasing that, I don't like it. I think we should be passing around the full path of the file the user picks, rather than just the name then try to find the right one later.

Alberth289346:
> I added and removed another commit, but in writing and rebasing that, I don't like it. I think we should be passing around the full path of the file the user picks, rather than just the name then try to find the right one later.

I agree that passing around a full path makes life much convenient and less error prone, go for it !

You could also see it as having 2 needs. One need is the program finding the file again that you mentioned above. The other need is showing the user what they get, and a full path may not the best information for that.  
Of course the latter can be expanded with a teaser text, some pictures, etc etc :p

I can live without the user part, but it may be useful not to make that impossible to add.

tobylane:
Full paths are used where the file will be used straight away, filenames and folder of parent file are used where the file might be opened later, eg on another computer. All subfolder (one deep) are checked for campaign files, but only the subfolder with the campaign file will be searched for the level file, and same with the level file dir and map file. I'm intending that a whole campaign can be in its own folder, and can use files outside of it. A map reused in a second campaign would be duplicated in the second folder.

CorsixTH/Lua/app.lua
```diff
@@ -631,10 +630,23 @@ end
...
-function App:readLevelFile(level)
-  local filename = self:getAbsolutePathToLevelFile(level)
+function App:readLevelFile(level, preferred_dir)
+  local filename
+  local search_paths = {
+    self.user_campaign_dir,
+    self.user_level_dir,
+    self.campaign_dir,
+    self.level_dir,
+  }
```
Alberth289346, line 643: Try to keep local variables as close to first use as possible.

It avoids accidentally using non-initialized variables, and makes it easier to find the declaration.

Alberth289346: Here, initializing `search_paths` just above 648 has the advantage that in some runs it never even creates the variable.

CorsixTH/Lua/app.lua
```diff
@@ -647,6 +659,7 @@ function App:readLevelFile(level)
   level_info.level_file = level
   level_info.name = contents:match("%Name ?= ?\"(.-)\"") or "Unknown name"
   level_info.map_file = contents:match("%MapFile ?= ?\"(.-)\"")
+  local map_file_backup = level_info.map_file
```
Alberth289346, line 662: If you need to make a backup, you may be doing things in the wrong order or you may have made a premature "jump" to the conclusion (as in you computed "the" answer, but for some reason it's not always "the" answer, and you need to rescue some values).

A different possibility is that the variable has in a sense "several" values that do not belong together just yet. (ie the values were merged too soon).

Here, it looks like you try too soon to create the `level_info` triplet values (`level_file`, `name`, and `map_file`). If you use separate variables for them, and then after everything is decided, you construct the triplet, I think the problem disappears.

To some extent this may also happen because we tend to compute everything in one function even if some parts are just different attempts to compute the same thing. Splitting it into more smaller functions may kind of "automatically" solve such problems, because bottom-level stuff is then done inside the child functions, and here you only worry about the higher level coordination of what function to call when.  
Heaps to gain there in our code :)

CorsixTH/Lua/app.lua
```diff
@@ -671,27 +691,21 @@ function App:readLevelFile(level)
...
-  }
-  for _, parent_path in ipairs(paths_to_search) do
-    local check_path = parent_path .. pathsep .. level
+--! Searches in the given folders for the given filename
+--!param search_paths (table) List of folders (strings) to look through
+--!param filename (string) Filename to search for.
+--!param preferred_dir (string) Path that should be checked first
+--!return path (string) The full path of the found file
+function App:findFileInDirs(search_paths, filename, preferred_dir)
+  if preferred_dir then table.insert(search_paths, 1, preferred_dir) end
```
Alberth289346, line 700: This modifies `search_paths`, which may not be expected by the caller (and the Lua doc also doesn't warn about that).

This function is doing 2 different things. It decides order of the paths to search (for one special case only), and it does the search.

The method name only claims it does a search. So there are 2 ways here to solve it.
1. Do what you already do here, but without messing up `search_paths`.
2. Move setting up the order of the search paths out of this function.
3. Rename the function to something wider (but imho that is not a good idea, deciding order and doing searching are different things).

It's a weird that you have search-paths in the order of visiting, but than apparently you need to fix the first search. That points at the search path construction code which is thus apparently doing something wrong.
In such cases, fix the code that is broken rather than adding some additional fixing code for special cases. If you do that long enough, nobody knows anymore what is supposed to happen where.

Code that is doing some job should completely do that job, so other code doesn't need to bother about it.

CorsixTH/Lua/app.lua
```diff
@@ -671,27 +691,21 @@ function App:readLevelFile(level)
   return level_info
```
Alberth289346, line 691: Not your fault probably, but if you look at this code you'll notice that 662 to 665 and 667 to 670 do the exact same thing, except for different pieces of text.

By encapsulating the 4 lines into a function and giving it `contents` and `"Briefing"` resp `"Debriefing"` you can move the 10 lines of details out of this function (for as far as I can see now), and replace it by 2 calls to the new function.

By doing that more, you get higher-level functions that handle bigger steps in the code without getting several hundreds of lines long. Reading and understanding such functions is much easier too, as you never even see all the little details that are hidden in the called functions.

CorsixTH/Lua/app.lua
```diff
@@ -1597,7 +1611,7 @@ function App:readMapDataFile(filename)
     end
   else
     -- Could not find the file
-    return nil, _S.errors.map_file_missing:format(filename)
+    return nil, _S.errors.map_file_missing:format(path)
   end
   return data
```
Alberth289346, line 1616: In general, looking down for the next matching `elseif`, `else`, or the `end` line is not trivial, especially if they code in-between is long.
For this reason prefer handling the shorter alternative first. An alternative that ends the function is also a good option, since it avoid handling multiple cases at the same time and/or doing the same test more than once (that doesn't quite happen here at first sight).

At a higher level, several lines look like they are generic. Eg the "test presence of a file" lines and "read the data lines if the file exists", and the "decompress if compressed" lines are used elsewhere too in the exact same way, and are thus good candidates to move out of this function.  
Not sure if "read a level file and throw an error if missing" is common too, but it could be.  
Mostly this is again "don't do everything in one function, move lower level details out of the way".

Not sure if you should do that now. I can live with not doing this.

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
```diff
@@ -33,42 +33,63 @@ local col_scrollbar = {
...
-
-  for file in lfs.dir(path) do
-    if file:match("%.campaign$") then
-      local campaign_info, err = TheApp:readCampaignFile(file)
-      if not campaign_info then
-        print(err)
+--!param paths_table (array) File system paths to search.
+--!return (array) The found levels, with some basic information about each level, in alphabetical order.
+local function createCampaignList(paths_table)
+  local campaigns, unique_names, duplicates = {}, {}
```
Alberth289346, line 39: Left side has 3 variables, right side has 2 values.
As it is now, it can be easily seen as a missing 3rd value.

If it is not, split the variable declarations so it is clear that it's not a mistake.

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
```diff
@@ -33,42 +33,63 @@ local col_scrollbar = {
...
+    local full_path = folder .. file
+    local campaign_info, err = TheApp:readCampaignFile(full_path)
+    if not campaign_info then
+      print(err)
+    else
+      local name = campaign_info.name
+      if unique_names[name] then
+        print("Custom campaign error: duplicate campaign name in file " .. file ..
+            ". Check the folders " .. table.concat(paths_table, ", "))
+        duplicates = true
```
Alberth289346, line 51: Avoid using and setting variables that exist outside the function. It makes understanding and debugging a mess. It also makes it impossible to move this local function out of this code without major headaches and/or re-writing.

In my view, Lua allowing this is a major wrong design decision.  
It may "simplify" small scripts, but at our scale the benefit of less input/output parameters is a non-problem compared to difficulties in understanding code structure, and what variables gets used and assigned where.

Alberth289346: One way to avoid these problems is to only have outer-level functions. Use a `_` prefix to indicate private functions then.

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
```diff
@@ -33,42 +33,63 @@ local col_scrollbar = {
...
-local function createCampaignList(path)
-  local campaigns = {}
-
-  for file in lfs.dir(path) do
-    if file:match("%.campaign$") then
-      local campaign_info, err = TheApp:readCampaignFile(file)
-      if not campaign_info then
-        print(err)
+--!param paths_table (array) File system paths to search.
+--!return (array) The found levels, with some basic information about each level, in alphabetical order.
```
Alberth289346, line 37: It also returns a duplicates flag

Technically the description at line 35 isn't complete, it also checks for validity of the campaign and reports errors to the log

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
```diff
@@ -80,6 +101,13 @@ function UICustomCampaign:UICustomCampaign(ui)
 
   self.description_offset = 0
 
+  -- Warn about hidden duplicate campaigns
+  if duplicates then
```
Alberth289346, line 105: You recognize duplicates while reading, and then report it here.

If you read and return every valid campaign you find, and here check the results for duplicates, you separate duplicate handling from reading campaigns.

Alberth289346: Eg extending to "3 similarly named campaigns X were hidden" becomes simpler then, as doesn't need to be done inter-mixed with reading and validating.

That is, doing one thing at a time often makes extending functionality tomorrow simpler.

CorsixTH/Lua/world.lua
```diff
@@ -1305,15 +1305,17 @@ function World:getCampaignWinningText(player_no)
...
-    local level_info = TheApp:readLevelFile(self.map.level_number)
     text[1] = _S.letter.dear_player:format(self.hospitals[player_no].name)
     if has_next then
-      text[2] = level_info.end_praise and level_info.end_praise:format(next_level_name) or _S.letter.campaign_level_completed:format(next_level_name)
-      text[3] = ""
+      if next_level_info.end_praise then
+        text[2] = next_level_info.end_praise:format(next_level_name)
+      else
+        text[2] = _S.letter.campaign_level_completed:format(next_level_name)
+      end
```
Alberth289346, line 1314: Not your fault, but `next_level_name` isn't initalized:

image


Keeping the declaration close to first use makes this much simpler to detect.

CorsixTH/Lua/app.lua
tobylane: Redone, with stricter error checking

CorsixTH/Lua/app.lua
tobylane: Chose 2, especially as there's now one case of adding two paths.

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/custom_campaign.lua
tobylane: Changed to private class functions, and class variables.

CorsixTH/Lua/world.lua
tobylane: It's declared out of the loop, and set inside in on L1302.

tobylane (commented):
Thanks, made changes. Should I avoid this overwriting of previous commits?

CorsixTH/Lua/world.lua
Alberth289346: Ah ok, my bad

Alberth289346:
> Thanks, made changes. Should I avoid this overwriting of previous commits?

In general, yes. Don't use it unless there is no other way.

The idea of open source is that everybody shares and extends code with each other.

For example, I can create a patch on top of your patch (even while you aren't done yet with your patch) to fix some problem or add new functionality.
I can share my patch with you and you can merge it with your own patch.  
Alternatively, if your patch gets merged before I am finished, I can submit my patch afterwards.

All the above assumes that your patch and my patch are compatible. They have a common starting git revision, a common history.  
Under that condition, git allows either of us merging an external patch into our own patch.  
If you force-push, there is no common history anymore, and merging other code becomes impossible. That is mostly very bad news for the person(s) **not** doing the force-push.

Of course, all this is only a problem when there is a need to merge code. If you are the only user of a repository there is never a problem. Here, someone could create a patch on top of your branch and force-psuh would give trouble, but the likelihood of that happening is not big. If however we would force-push in `master`, then everybody writing patches for CorsixTH would be affected.

Thus, the history problem for your patches is mostly a theoretical problem until you start co-operating with someone to write code with more than one person in the same branch without eg a review process like we have here. 
On the other hand, if you want to do that at some point in time, it doesn't hurt to get used to the idea that history shouldn't be destroyed if it is avoidable.

Alberth289346:
I think the patch is ok

lewri (commented):
I appreciate Albert has been helping with this one from the start.
I also like that our campaign load priorities the folder it came from.

A few queries:
1. This PR appears to also provide a visual change; can you screen grab it into the PR?
2. Given how https://github.com/CorsixTH/CommunityContent is currently designed: Levels are also bundled into their own encapsulated folder. It's a shame we aren't allowing Levels to be checked one deep too; but general it's also out of scope of what you're aiming to achieve.

PR #2763, opened 2025-01-19
Instant speed changes

### LINKED ISSUE #2758: [Bug] Delay on changing game speed from speed 1
### Describe the issue

At minimum game speed, pressing keys `1`, `2`, `3`, `4`, `5` and `Z` does not lead to an instant result, but occurs with a delay.
The same effect at speed 2, but not such noticeable.
Looks like game waiting next tick or current tick end to handle speed change.

### Steps to Reproduce

1. *Load any your save game*
2. Press `1` to set speed to minimum.
3. Try rapidly pressing and releasing `Z` key.
4. Observe that nothing happening.
5. Try rapidly pressing `1` and `5` keys in turn.
6. Observe that in most attempts game speed will never switch to 5. But for 1 always.

### Expected Behaviour

Result on pressing `1`, `2`, `3`, `4`, `5` and `Z` keys should be instantly, as in vanilla TH.
No waiting for next tick.

Perhaps a force move to next tick on speeds 1 and 2 at press `2`, `3`, `4`, `5` or `Z` would be the solution.

### CorsixTH Version

0.68.0

### Operating System

MacOS Sonoma 14.5

TheCycoONE:
Avoid waiting for next game hour to adjust to speed changes.

Fixes #2758

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -732,8 +732,8 @@ end
 
 --! Return true if the given speed the same as the current speed.
 function World:isCurrentSpeed(speed)
-  local numerator, denominator = unpack(tick_rates[speed])
-  return self.hours_per_tick == numerator and self.tick_rate == denominator
+  local hours_per_tick, tick_rate = unpack(tick_rates[speed])
+  return self.hours_per_tick == hours_per_tick and self.tick_rate == tick_rate
 end
 
 --! Return the name of the current speed, relating to a key in tick_rates.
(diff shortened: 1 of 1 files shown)


Alberth289346:
Approach seems fine to me.

Optionally:  
Method could use cleanup of the names, it has `current` values, `old` values, and just `values`.
In reality however there are just 2 sets of values. It would help reading the code if each set of variables has a distinct common prefix.

ARGAMX:
Wow! Speed ​​switches instantly now! Cool! Great job! 🔥

But I noticed that when loading some savegames, but not all, the following error occurs:

Screenshot 2025-01-19 at 13 43 34

Here is an example of a save that I can't load after applying the change to the current master:

[speed_switch_test.sav.zip](https://github.com/user-attachments/files/18468536/speed_switch_test.sav.zip)

ARGAMX:
Also, maybe this is not directly related to this PR and task, but I would like to draw attention to the fact that if the X key was held down during saving (especially autosaving) to speed up the game speed, then when loading such a save, this speed will be keeped. Which looks a little strange after loading and you want to immediately slow down the speed of that chaos. This became especially relevant after #2688, in which the X key speed seemed to about double compared to what it was before In 0.68.0.

Here is such save:

[Autosave5_ultraspeed.sav.zip](https://github.com/user-attachments/files/18468630/Autosave5_ultraspeed.sav.zip)

TheCycoONE:
hmm... are those all older saves? I would like one if you can provide it here.

TheCycoONE:
Try again, I addressed the naming and the speed key on load issue. It's not clear how this could cause errors on load but I added some paranoia handling of missing globals as well.

lewri (commented):
I'm not sure of the increased use of CamelCase but logically it all seems fine to me too.

ARGAMX:
> Try again, I addressed the naming and the speed key on load issue. It's not clear how this could cause errors on load but I added some paranoia handling of missing globals as well.

I just test again and seems both points resolved.
All savegames loading successfully now.
Speed key now reseting after load.
Thanks! 👍

Works good to me.

TheCycoONE:
Well the lua variable conventions don't match the C++ ones at all do they... fixed.

PR #2767, opened 2025-01-23
Improve cheat feedback

lewri:
- Fixes a bug where a failed cheat didn't display any feedback to the player, if no custom message was provided.
- Fixes a bug where the `no_heliport` message for createEmergency never happened (typo); and caused the player to be marked a cheater when cheat was unsuccessful.

@MandrakeHorse this change now means the string `information.cheat_not_possible` appears; and given its wording I'm not sure if it needs revision: `Cannot use that cheat on this level. You even fail to cheat, not that funny huh?`

(NB: there is also `information.level_lost.cheat` that does a similar thing: `This was your choice or did you select the wrong button? So you can't even cheat correctly, not that funny huh?`)


MandrakeHorse:
If you're taking suggestions, we could replace these with something less aggressive towards the player:
`information.cheat_not_possible = "You cannot use that cheat on this level."`
`information.level_lost.cheat = "Hope you didn't click the Lose Level button by accident!"`

TheCycoONE:
I agree with adjusting the tone of the messages. I'm sure they are meant to be funny but it doesn't match the game humour.

lewri:
> information.cheat_not_possible = "You cannot use that cheat on this level."

My concern with this one is sometimes it might just be the cheat can't be used at the current time. However, I'm not strongly opposed enough to it to want to revise it.

lewri:
Strings pushed and discord's translation section informed, thanks @MandrakeHorse 

@Alberth289346 if you desire feel free to adjust the Dutch ontop or in its own pr.

Alberth289346:
> @Alberth289346 if you desire feel free to adjust the Dutch ontop or in its own pr.

Added #2769 for that.

I changed the phrasing to better match the other messages nearby to:
`You have (hopefully not by accident) used the "Lose level" option.`

PR #2771, opened 2025-01-28
Replace calculateSpawnTiles repeat loop

lewri:
- Changes repeat loop to a while loop

In general I think the `return` definition might be wrong in the function overall but not confident.

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -327,14 +327,14 @@ function World:calculateSpawnTiles()
     local xs = {}
     local ys = {}
     local x, y = edge.origin[1], edge.origin[2]
-    repeat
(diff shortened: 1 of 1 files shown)


lewri:
> I am not sure why you changed it, but it LGTM.

It's a good point, though I get the general sense from the wiki we should try and push away from `repeat until ...` loops as much as we can. This usually means changing to a `for` or a `while` I would generally guess.

Alberth289346:
> I get the general sense from the wiki we should try and push away from `repeat until ...` loops

I haven't read that in a while, but in my mind it's mostly about using `break` which then turns into `continue` due to a surrounding context with a repeat.

However, I can see the advantage of reducing the diversity of loop constructs. It's makes the code more predictable, and thus easier to understand.

PR #2591, opened 2024-07-28
updated Italian translation, correct spelling

SebastianoPistore:
updated Italian translation, correct spelling

PR #2780, opened 2025-02-13
Update italian.lua

SebastianoPistore:
update italian translation


TheCycoONE:
That was explicitly called out as an unsupported character, did you test it?

lewri:
Worst case, this should just become E as a standard letter.
https://github.com/CorsixTH/CorsixTH/pull/2591#issuecomment-2254635309
https://github.com/CorsixTH/CorsixTH/wiki/Localization

lewri:
I've checked this in the white font and the fax font

image
image

Closing this, will reopen if we want to use `E` instead of `E'`

PR #2781, opened 2025-02-20
Epidemic refactoring. `coverup_in_progress` rename.

ARGAMX:
- `coverup_in_progress` renamed to `coverup_selected` as this flag indicates the fact that the coverup was ever started for this epidemic
- `patient:setMood` calls moved from `epidemic.lua` and `vaccinate.lua` to `patient.lua`
- due to `hospital.lua` improper `afterLoad` call, afterLoad was always crash for savegames with `SAVEGAME_VERSION` less than `106`
- `SAVEGAME_VERSION` bump

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -44,7 +44,8 @@ function Patient:Patient(...)
   self.vaccinated = false
   -- Has the patient been sent to the wrong room and needs redirecting
   self.needs_redirecting = false
-  self.attempted_to_infect= false
+  -- Is under attempt to be infected
+  self.attempted_to_infect = false
   -- Is the patient about to be vaccinated?
   self.vaccination_candidate = false
   -- Has the patient passed reception?
@@ -990,7 +991,7 @@ function Patient:updateDynamicInfo()
   end
   -- Set the centre line of dynamic info based on contagiousness, if appropriate
   local epidemic = self.hospital and self.hospital.epidemic
-  if epidemic and self.infected and epidemic.coverup_in_progress then
+  if epidemic and self.infected and epidemic.coverup_selected then
     if self.vaccinated then
       self:setDynamicInfo('text',
         {action_string, _S.dynamic_info.patient.actions.epidemic_vaccinated, info})
@@ -1059,6 +1060,18 @@ function Patient:updateMessage(choice)
   end
 end
 
+--[[ Show patient infected status ]]
+function Patient:setInfectedStatus()
+  self:setMood("epidemy4","activate")
+end
+
+--[[ Show patient as ready for vaccination status ]]
+function Patient:setToReadyForVaccinationStatus()
+  self:setMood("epidemy4","deactivate")
+  self:setMood("epidemy2","activate")
+  self.marked_for_vaccination = true
+end
+
 --[[ If the patient is not a vaccination candidate then
   give them the arrow icon and candidate status ]]
 function Patient:giveVaccinationCandidateStatus()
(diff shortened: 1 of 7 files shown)


CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -44,7 +44,8 @@ function Patient:Patient(...)
   self.vaccinated = false
   -- Has the patient been sent to the wrong room and needs redirecting
   self.needs_redirecting = false
-  self.attempted_to_infect= false
+  -- Is under attempt to be infected
```
TheCycoONE, line 47: This comment doesn't read quite right, I don't follow what it means.

CorsixTH/Lua/epidemic.lua
```diff
@@ -125,44 +126,42 @@ function Epidemic:infectOtherPatients()
...
   corridors or in the same room - don't infect through walls.
-  @param patient (Patient) already infected patient
-  @param other (Patient) target to possibly infect
+  @param infector (Patient) already infected patient who want to infect
+  @param victim (Patient) target to possibly infect
   @return true if patient can infect other, false otherwise (boolean) ]]
-  local function canInfectOther(patient, other)
-    -- Patient is not infectious.
-    if patient.cured or patient.vaccinated then return false end
+  local function canInfectOther(infector, victim)
```
TheCycoONE, line 132: Good naming change, helps a lot with clarity.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -1059,6 +1060,18 @@ function Patient:updateMessage(choice)
   end
 end
 
+--[[ Show patient infected status ]]
+function Patient:setInfectedStatus()
+  self:setMood("epidemy4","activate")
```
TheCycoONE, line 1065: Not clear why some of these apply paranoia to remove all other epi statuses and others don't. Should we be consistent?

TheCycoONE:
The title refactor `coverup_in_progress` I don't see as particularly problematic, but some of the other changes here, particularly victim/infector are helpful enough to be worth it.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: Ok, so

This flag indicates is currently someone infected trying to infect THIS victim. 
In our currently logic infection is process which last some certain gametime.
This flag help to prevent the situation when several infectors trying to infect the same victim.

So this is not infector flag. This is victim flag.
(potential_victim.attempted_to_infect = true)

But from the title one might think the opposite.

Maybe my English is not good enough and I couldn't reflect this in the comment. But I tried to convey exactly this through word 'under'.

Or maybe the problem now is that there seems to be a contradiction between the comment and the name of the flag?..

So, what shall we do here?

ARGAMX:
> The title refactor `coverup_in_progress` I don't see as particularly problematic

👍
The thing is, we have discussed the `coverup_in_progress` flag with Lewri and Albert in #2748 that the name of that flag doesn't quite reflect what it means.

Logically, when the coverup ends, we need to turn off this flag back to `false`.
But we can't do this, because it will break the epidemic logic.

Because in fact, this flag means whether we chose coverup at some point for this epidemic, not really is coverup happening at the moment or not.

Therefore, it is easier to rename the flag a little for clarity than to rewrite the logic. 👍

CorsixTH/Lua/entities/humanoids/patient.lua
TheCycoONE: Add this comment to the code. Comments don't have to be short.

TheCycoONE: Ok, the comment still adds clarity. Attack may not be the right word, but I don't have a suggestion.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: > Not clear why some of these apply paranoia to remove all other epi statuses and others don't. Should we be consistent?

Yeap.
I don't know why someone apply paranoia to remove all other epi statuses here.
Maybe there is no need paranoia here. 
But I am not sure on it. Who knows, maybe it really working for some cases.
It was here and I didn't touch it. Just moved it to `patient.lua`.

Overall, consistent sounds reasonable.
I added paranoia for each infected status icon transition.
Now it should be paranoidal consistent. 🙂

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: > Ok, the comment still adds clarity. Attack may not be the right word, but I don't have a suggestion.

`attempted_to_infect` ->`under_infection_attempt` ?

Ok, comment will be added anyway.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -44,7 +44,10 @@ function Patient:Patient(...)
   self.vaccinated = false
   -- Has the patient been sent to the wrong room and needs redirecting
   self.needs_redirecting = false
-  self.attempted_to_infect= false
+  -- Is under infection attempt
+  -- Indicates is currently some infected patient trying to infect this patient
+  -- Helps to prevent the situation when several infectors trying to infect the same victim
```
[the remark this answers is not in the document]
ARGAMX: > Helps to; or prevents fullstop?

Improved it a bit. It should probably be clearer now.

lewri (approved):

TheCycoONE:
Please bump the save version, then I think this is ready to merge.

ARGAMX:
> Please bump the save version, then I think this is ready to merge.

Done 👌

Issue #2779, opened 2025-02-13
[Bug] Unreadable low contrast text in fax for unicode languages


ARGAMX:
### Describe the issue

Casebook and fax has unreadable low contrast text for unicode languages.
Also date text has lost some contrast too.

<img width="912" alt="Image" src="https://github.com/user-attachments/assets/8140935b-9ae7-4948-8f38-b9935efa422b" />
<img width="912" alt="Image" src="https://github.com/user-attachments/assets/fedd9c57-b3ef-4ff7-b1dc-70bbc11a3c64" />

### Steps to Reproduce

1. Select any unicode language in Settings (Rus/Ukr/Pol/Cze/Hung)
2. *Start new game*
3. Open Casebook (F5).
4. Observe problem.

### Expected Behaviour

Text should be contrast as in 0.68.0

<img width="927" alt="Image" src="https://github.com/user-attachments/assets/083196b9-4412-4c30-b5ed-48585b8154a0" />

### CorsixTH Version

7fa317ce trunk master

### Operating System

MacOS Sonoma 14.5

lewri:
Confirmed regression of #2745 . @Alberth289346 any obvious reasons why the behaviour may have changed?


The issue appears to be somewhat related to fonts in dark colours e.g. Player names in the Popn/Win-lose goals; Casebook; Date; Faxes.

![Image](https://github.com/user-attachments/assets/f01c75c6-6ec5-4162-a402-649ea6e48960)

UIInformation is unaffected, maybe as it's pure black?

Alberth289346:
Don't think there is any alpha, that didn't exist in the 90's in home computers afaik.
The current renderer may use it though.

Alberth289346:
Patch:
``` diff
index bfcb2182..cacbfd04 100644
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -1200,6 +1200,9 @@ bool sprite_sheet::get_sprite_average_colour(size_t iSprite,
   const sprite* pSprite = sprites + iSprite;
   int iCountTotal = 0;
   int iUsageCounts[256] = {0};
+  int seen[256] = {0};
+  printf("==============================\n");
+  printf("Started analyzing text sprite.\n");
   for (long i = 0; i < pSprite->width * pSprite->height; ++i) {
     uint8_t cPalIndex = pSprite->data[i];
[...]

Issue #2744, opened 2024-12-16
[Bug] The pause label is almost invisible in Unicode languages.


ARGAMX:
### Describe the issue

The pause label is almost invisible in Unicode languages.

How it looks on non Unicode languages:
<img width="276" alt="pause_nonunicode" src="https://github.com/user-attachments/assets/748c79cf-4d61-4ebb-a600-02e06a09de50" />

How it looks on Unicode languages:
<img width="276" alt="pause_unicode" src="https://github.com/user-attachments/assets/299a3434-940a-4649-a1ed-8e5eaa7f495a" />

How game looks overall in pause mode in Unicode languages with `build on paused` enabled:
<img width="456" alt="Screenshot 2024-12-16 at 11 52 53_small" src="https://github.com/user-attachments/assets/5381e5b0-9eb0-4281-be18-4b912d5a06f7" />

Moreover, if you hover cursor over the doctor, patient, door, machine then even this barely legible label will disappear.

### Steps to Reproduce

1. *Select any Unicode language in settings (Ukrainian, Russian, Polish, Czech)*
2. Load savegame or start new game.
3. Press `P` to enable pause.
4. Observer that The pause label is almost invisible.

### Expected Behaviour

The pause label should be clear visible as for non Unicode languages.

### CorsixTH Version

0.68.0

### Operating System

MacOS Sonoma 14.5

lewri:
Font style in comparison looks rather awful too

Alberth289346:
Would need more than one font, and depend on what text you render

Alberth289346:
Not yet fixed, re-opened due to typo fix in #2782

PR #2783, opened 2025-02-23
Add method for sprite visibility test

Alberth289346:
Create separate method for testing if a sprite is visible at all.

Addresses #2779

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.h
+++ b/CorsixTH/Src/th_gfx_sdl.h
@@ -647,7 +647,7 @@ class sprite_sheet {
 
     //! Height of the sprite.
     int height;
-  }* sprites;
+  } * sprites;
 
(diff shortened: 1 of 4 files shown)


Alberth289346:
Actually, addresses #2744

CorsixTH/Src/th_gfx_sdl.h
```diff
@@ -562,6 +562,13 @@ class sprite_sheet {
   */
   bool get_sprite_average_colour(size_t iSprite, argb_colour* pColour) const;
 
+  //! Return whether the given sprite show any pixel when displayed.
```
Alberth289346, line 565:
```suggestion
  //! Return whether the given sprite shows any pixel when displayed.
```

TheCycoONE (approved):

Issue #2421, opened 2023-09-29
Adjust frequency of random announcements


MandrakeHorse:
### What feature or enhancement would you like to add?

One of Theme Hospital's memorable features are the humorous announcements that are randomly heard while playing ("Patients are reminded not to die in the corridors" etc). However, in the current version of CorsixTH (0.67), these announcements are very rarely heard during normal gameplay. 

The code which determines when the next random announcement should play is as follows (thanks @lewri!):
```
--! Private function. Sets the new time (in ticks) when a random announcement should be played.
function Announcer:_setRandomAnnouncementTarget()
  -- Note that random announcement are measured in ticks.
  -- This ensures that on fast game speeds random announcements aren't
  -- spammed, or on lower game speeds random announcements are never played.
  -- Every tick is 30ms, so 2000 ticks is 1 minute.
  self.random_announcement_ticks_target = math.random(8000, 12000)
end
```
This suggests that random announcements will only play after 4-6 minutes of inactivity, which is almost guaranteed to never occur once the hospital has been opened. We should reduce the number of ticks so that a smaller period of inactivity must pass before a random announcement will be played (although not too small so that they aren't spammed on faster game speeds or otherwise play too frequently).

### How would this feature benefit the gameplay of CorsixTH?

The random announcements would play more often during regular gameplay, thus provoking mirth and a sense of nostalgia.

### Was the feature in the original game?

Yes

### What version of CorsixTH are you using?

0.67

### Does this feature/enhancement relate to a current problem?

No

lewri:
If you use game speed in a calculation you can then scale the announcement delay to be roughly the same then

PR #2785, opened 2025-02-23
Adjust random announcement target

lewri:
*Fixes #*

- Tick changes in 0.68.0 didn't adjust the random announcement target value; so recalculated here
- Hopefully I can still eventually get around to improving this better WRT #2421

### DIFF
--- a/CorsixTH/Lua/announcer.lua
+++ b/CorsixTH/Lua/announcer.lua
@@ -226,8 +226,11 @@ function Announcer:_setRandomAnnouncementTarget()
   -- Note that random announcement are measured in ticks.
   -- This ensures that on fast game speeds random announcements aren't
   -- spammed, or on lower game speeds random announcements are never played.
-  -- Every tick is 30ms, so 2000 ticks is 1 minute.
-  self.random_announcement_ticks_target = math.random(8000, 12000)
+
+  -- Every tick is 18ms, so ~3333 ticks is 1 minute at normal speed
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/announcer.lua
```diff
@@ -226,8 +226,11 @@ function Announcer:_setRandomAnnouncementTarget()
   -- Note that random announcement are measured in ticks.
   -- This ensures that on fast game speeds random announcements aren't
   -- spammed, or on lower game speeds random announcements are never played.
-  -- Every tick is 30ms, so 2000 ticks is 1 minute.
-  self.random_announcement_ticks_target = math.random(8000, 12000)
+
+  -- Every tick is 18ms, so ~3333 ticks is 1 minute at normal speed
+  local tick_chunk = 3333
+  -- Set a target of 4 - 6 minutes (chunks)
+  self.random_announcement_ticks_target = math.random(tick_chunk * 4, tick_chunk * 6)
```
Alberth289346, line 233: The better approximation would be an exponential distribution.

https://en.wikipedia.org/wiki/Exponential_distribution

Alberth289346:
Patch LGTM in itself.

CorsixTH/Lua/announcer.lua
lewri: I don't like how it calculates already to be fair; but that's for #2421 to worry about

PR #2801, opened 2025-03-24
SDL3 port

TheCycoONE:
So far the basic game functionality works (tested in Linux, there's some Windows specific code that wasn't exercised.)

Movies, audio, and music have not yet been ported.

Some cleanup would be nice.

Linux pipelines will need to be fixed somehow (vcpkg?)

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -10,6 +10,7 @@
         "ENABLE_UNIT_TESTS": "ON",
         "BUILD_ANIMVIEW": "ON",
         "BUILD_TOOLS": "ON",
+        "WITH_AUDIO": "OFF",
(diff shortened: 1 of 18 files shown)


lewri:
Font render is a bit funky when fullscreen is non-native resolution.
SDL3
image

SDL2
image

(also includes money bar/date as if it was blurred then poorly sharpened)

Pointer also picked up an artifact once but haven't been able to replicate again.

TheCycoONE:
Thanks. I'm not totally surprised given all the rendering changes. The port is on hold and maybe some rendering options will be changed by the time they port sdl_mixer.

Alberth289346:
Code looks ok at first sight, nothing special or weird.

TheCycoONE:
This rebase is painful and a half - probably will start over instead.

PR #2802, opened 2025-03-24
Add interpeter path for portable linux installs

lewri:
*Fixes #*

- Add standard interpreter path when setting a custom install directory and standard `/usr` prefix for Linux

This is needed while working on building for AppImage.


TheCycoONE:
Approved.

PR #2803, opened 2025-03-24
Object/Machine updateDynamicInfo should not increase object usage

### LINKED ISSUE #2156: Object/Machine updateDynamicInfo should not increase object usage
`xxx.times_used` and `xxx.total_usage` should be updated at the point an object is used, rather than within the `updateDynamicInfo` function.
https://github.com/CorsixTH/CorsixTH/blob/01440cedbfcc74e6b4eb4d42d1069ce1975b5c7b/CorsixTH/Lua/entities/machine.lua#L353-L374
https://github.com/CorsixTH/CorsixTH/blob/01440cedbfcc74e6b4eb4d42d1069ce1975b5c7b/CorsixTH/Lua/entities/object.lua#L267-L273

For machine, `machineUsed` exists where this action can be moved to.
For object, this would need creating (currently `updateDynamicInfo` for objects is called from `humanoid_actions/use_object.lua`

ARGAMX:
*Fixes #2156*

- Object/Machine `updateDynamicInfo()` now not anymore responsible for increase object/machine usage. `incrementUsedCount()` now responsible for this.

Also some accompanying refactoring:

- Checking objects is a machine or not everywhere in manner `if object.strength then` replaced for more obvious
```
if object:isMachine() then
```

- Checking handyman existent or non-existent task in manner
```
local index = self.hospital:getIndexOfTask(self.tile_x, self.tile_y, taskType)
if index ~= -1 then
  self.hospital:removeHandymanTask(index, taskType)
...
```
replaced for more obvious
```
local task_index = self.hospital:getIndexOfTask(self.tile_x, self.tile_y, taskType)
local task_exist = task_index ~= -1
if task_exist then
  self.hospital:removeHandymanTask(task_index, taskType)
...
```

- Refactored some bad smelling values declaration like
`self.times_used = -1 -- Incremented in the call on the next line`
and
`self.total_usage = -1 -- Incremented in the constructor of Object.`
to
`self.xxx = 0`

- Big function `Machine:machineUsed` devided to `Machine:checkIsNeedToCallHandymanForRepair` and `Machine:calculateIsMachineMustExplode` and  `Machine:explodeMachine`.


CorsixTH/Lua/earthquake.lua
```diff
@@ -136,8 +136,9 @@ function Earthquake:tick()
 
   for _, room in pairs(self.world.rooms) do
     for object, _ in pairs(room.objects) do
-      if object.strength then
-        object:machineUsed(room)
+      local object_is_machine = object.strength
```
TheCycoONE, line 139: This would be better as a method of object.

ARGAMX: > This would be better as a method of object.

Wise.
Changed accordingly.

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -505,9 +505,9 @@ function UIPlaceObjects:placeObject(dont_close_if_empty)
     end
     -- Some objects (e.g. the plant) uses this flag to avoid doing stupid things when picked up.
     real_obj.picked_up = false
-    -- Machines may have smoke, recalculate it to ensure the animation is in the correct state
-    if real_obj.strength then
-      real_obj:calculateSmoke(room)
+    if real_obj:isMachine() then
+    -- Machines may have some state like smoke and etc. Update it.
```
Alberth289346, line 509: Please indent comment lines just like normal statement lines.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
   end
 end
 
+--! Call on machine earthquake impact.
+--!param room (object) machine room
+function Machine:earthquakeImpact(room)
```
Alberth289346, line 101: The first line is wrong in two ways.

- It repeats the function name almost literally. There is never a need to write the same thing twice. Eg
```
local j = 1  -- Assign 1 to j
```
Your audience are programmers, they can read and understand `local j = 1`. Don't repeat what the code is already telling. Instead explain the higher level reason. Why is j assigned, and/or why is it 1?

As you also don't want to explain the obvious, eg if 1 is the only obvious good value here, don't explain anything. On the other hand, if it is an unexpected value but it is correct for a complicated reason, then do explain.

The code explains what it does. The comments should explain why that is the correct thing to do.

- The text should be readable without the reader knowing all the magic words. I read "Impact", but I don't understand what that means.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+  self:machineUsed(room)
+end
+
 --! Call on machine use. Handles crashing the machine & queueing repairs
+--!param room (object) machine room
+--!return (bool) is room exploding after this last use
```
Alberth289346, line 107:
```suggestion
--!return (bool) is room exploding after this use
```
It is only the `last` use if the room is indeed exploding.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+  local must_explode = self:calculateIsMachineMustExplode(room, remaining_use_count)
+  if must_explode then
+    -- Room failed to be saved, it must be explode
+    self:explodeMachine(room)
     return true
-  -- Else if urgent repair needed or room didn't explode
-  elseif threshold < 4 then
+  else
+    -- Else if urgent repair needed or room didn't explode
+    self:checkIsNeedToCallHandymanForRepair(room, remaining_use_count, repair_task_index)
```
Alberth289346, line 126: If the code jumped to `else` from `if must_explode`, then yes, the room obviously didn't explode. No need to state in comment what you can quite easily read in the code.

I am not sure where "urgent repair" is coming from.

My first intuition in writing a comment here would be something like "Check if it needs maintenance."
You want to foremost describe what the next code line is doing. How you got here is not that important, usually. The reader should read the code and comments above here for that.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+    -- Update whether smoke gets displayed for this machine (and if so, how much)
+    self:calculateSmoke(room)
```
Alberth289346, line 128: The comment doesn't add much relative to the function name. I think it can do without comment here.

Don't know if you created `calculateSmoke`, but given your comment, `updateSmoke` may also be a good name.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+  end
+end
+
+--! Call on machine used. After machine used increment use values accordingly.
```
Alberth289346, line 132:
```suggestion
--! Call after use of the machine.
```
The function name already says use counts are incremented.

The goal of comment is to explain **why** the code is correct in doing what it does. Avoid repeating what the code is already saying.

Another use of comment is tell the high level steps of a longer process. For example if you need to 5 things in sequence, but each thing takes a page of code, then writing a comment line what step is being done above a page of code is helpful in guiding the reader to the code block that they are looking for.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+function Machine:incrementUsedCount()
+  self.times_used = self.times_used + 1
+  self.total_usage = self.total_usage + 1
+end
+
+--! Call on machine use. Handles crashing the machine & queueing repairs
+--!param room (object) machine room
+--!param remaining_use_count (int) machine remaining use count
+--!param repair_task_index (int) task id of repairing this machine handyman job
+function Machine:checkIsNeedToCallHandymanForRepair(room, remaining_use_count, repair_task_index)
```
Alberth289346, line 142: I don't see any machine crashing handling here, as claimed by line 138.
It just handles handyman tasks for repair.

The function name claims to "check". To me that means it returns a check result and doesn't actually do anything. This function not only checks, it also actively handles calling a handyman. The verb "check" is incomplete thus.

I find the `remaining_use_count` parameter a bit weird. This is the `Machine` class, why doesn't the machine already know this?

Also not sure about the repair_task_index. Can't you find that out? The machine knows its hospital as far as I know. It also knows its position, I hope.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+  local repair_task_dont_exist = repair_task_index == -1
```
Alberth289346, line 143:
```suggestion
  local repair_task_does_not_exist = repair_task_index == -1
```

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+  if remaining_use_count < 4 then
     -- If the job of repairing the machine isn't queued, queue it now (higher priority)
-    if taskIndex == -1 then
+    if repair_task_dont_exist then
```
Alberth289346, line 146: The task index is used twice, which makes creating a variable for it better.

An alternative here is a comment after the original line `-- There is no repair task.`. That clarifies what the code is doing without creating a new variable.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
       local call = self.world.dispatcher:callForRepair(self, true, false, true)
       self.hospital:addHandymanTask(self, "repairing", 2, self.tile_x, self.tile_y, call)
       self.hospital:announceRepair(room)
     else -- Otherwise the task is already queued. Increase the priority to above that of machines with at least 4 uses left
```
Alberth289346, line 150:
```suggestion
    else
      -- Otherwise the task is already queued. Increase the priority to above that of machines with at least 4 uses left
```

Not your fault, but it's nicer if comment stays together.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
        -- Upgrades task from low (1) priority to high (2) priority
        -- This does not lock the room, as happens when the task call starts at high priority
-      if self.hospital:getHandymanTaskPriority(taskIndex, "repairing") == 1 then
-        self.hospital:modifyHandymanTaskPriority(taskIndex, 2, "repairing")
+      if self.hospital:getHandymanTaskPriority(repair_task_index, "repairing") == 1 then
+        self.hospital:modifyHandymanTaskPriority(repair_task_index, 2, "repairing")
```
Alberth289346, line 154: Not your fault, but the `1` and `2` (probably priorities) are magic here.

I don't know if there is a `Task` class, but if there is, you can create a nice name for these.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
         self.hospital:announceRepair(room)
       end
     end
   -- Else if repair is needed, but not urgently
```
Alberth289346, line 158:
```suggestion
  -- Else if (low priority) repair is needed, make sure there is a handyman task for it.
```
Combine that with what the code is doing.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
-  elseif threshold < 6 then
+  elseif remaining_use_count < 6 then
     -- If the job of repairing the machine isn't queued, queue it now (low priority)
```
Alberth289346, line 160: Ah right, you wrote that here. That's ok too.

Note that you can combine both if conditions with an `and` in one `if`.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+    end
+    if num_extinguishers == 0 or remaining_use_count < -3 then
+      -- If no enought extinguisher in room, or machine used 5 times over its strength always explode
+      must_explode = true
+    else
+      -- Explosion chance increases 20% with every use over strength, and reduced by 5% for every additional extinguisher (up to 3 extra) in the room bar the first one
+      explosion_chance = (2 / self.strength) + (remaining_use_count * -0.2) - (num_extinguishers * 0.05) + 0.05
+      -- Cap it until guaranteed explosion
+      explosion_chance = explosion_chance > 0.95 and 0.95 or explosion_chance
+      explosion_chance = explosion_chance < 0.05 and 0.05 or explosion_chance
```
Alberth289346, line 191:
```suggestion
      explosion_chance = min(0.95, max(0.05, explosion_chance))
```

I don't understand line 189.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+  if remaining_use_count < 1 then
+    -- If a fire extinguisher in the room, room has chance not to explode
+    -- Calculate count of extinguisher
+    for object, _ in pairs(room.objects) do
+      if object.object_type.id == "extinguisher" and num_extinguishers < 4 then
+        num_extinguishers = num_extinguishers + 1
+      end
+    end
+    if num_extinguishers == 0 or remaining_use_count < -3 then
+      -- If no enought extinguisher in room, or machine used 5 times over its strength always explode
```
Alberth289346, line 184:
```suggestion
      -- If not enough extinguishers in the room or machine is used 5 times over its strength, always explode.
```

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
 
-  -- Update whether smoke gets displayed for this machine (and if so, how much)
-  self:calculateSmoke(room)
+--! Call on machine use. Handles crashing the machine
+function Machine:calculateIsMachineMustExplode(room, remaining_use_count)
+  -- Find a queued task for a handyman coming to repair this machine
+  local num_extinguishers = 0
+  local explosion_chance
+  local must_explode = false
+  -- Room is set to explode
```
Alberth289346, line 174: I don't understand this line

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+  if remaining_use_count < 1 then
+    -- If a fire extinguisher in the room, room has chance not to explode
+    -- Calculate count of extinguisher
```
Alberth289346, line 177:
```suggestion
    -- Calculate the number of extinguishers in the room.
```

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+    for object, _ in pairs(room.objects) do
+      if object.object_type.id == "extinguisher" and num_extinguishers < 4 then
+        num_extinguishers = num_extinguishers + 1
+      end
+    end
```
Alberth289346, line 182: Instead of skipping incrementing the count after reaching 4, break out of the loop when you reach 4.

That avoids inspecting room objects for nothing.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -93,91 +96,130 @@ local function setSmoke(self, isSmoking)
...
+  self:clearDynamicInfo()
+  -- Prevent the machine from smoking, it's now just a pile of rubble
+  setSmoke(self, false)
+  -- If we have the window for this machine open, close it
+  local window = self.world.ui:getWindow(UIMachine)
+  if window and window.machine == self then
+    window:close()
+  end
+  -- Clear the icon showing a handyman is coming to repair the machine
+  self:setRepairing(nil)
```
Alberth289346, line 218: If you insert an empty line above each comment, it's much easier to read.

CorsixTH/Lua/entities/object.lua
```diff
@@ -621,6 +622,11 @@ function Object:isReservedFor(user)
   return false
 end
 
+--! Checks the object is a machine.
+function Object:isMachine()
+  return self.strength ~= nil
+end
```
Alberth289346, line 628: Not your fault, but this is weird.

Normally you would make a derived `Machine` class from the `Object` class.
This method then returns `false`, while the same method in `Machine` returns `true`.

CorsixTH/Lua/entities/object.lua
```diff
@@ -651,9 +657,10 @@ function Object:onClick(ui, button, data)
       if self.object_type.class == "Machine" then
         taskType = "repairing"
       end
-      local index = self.hospital:getIndexOfTask(self.tile_x, self.tile_y, taskType)
-      if index ~= -1 then
-        self.hospital:removeHandymanTask(index, taskType)
+      local task_index = self.hospital:getIndexOfTask(self.tile_x, self.tile_y, taskType)
+      local task_exist = task_index ~= -1
+      if task_exist then
```
Alberth289346, line 662: You can also do the test in the `if` condition, and write a comment behind it what it does.

CorsixTH/Lua/humanoid_actions/use_object.lua
```diff
@@ -386,12 +386,13 @@ action_use_object_tick = permanent"action_use_object_tick"( function(humanoid)
 
     -- Check if the room is about to be destroyed
     local room_destroyed = false
-    if object.strength then
+    if object:isMachine() then
       if humanoid.humanoid_class ~= "Handyman"  then
         room_destroyed = object:machineUsed(humanoid:getRoom())
       end
     elseif object:getDynamicInfo() and not object.master then
       -- Don't update if it is a slave object.
```
Alberth289346, line 394: `not object.master` is tricky here. It looks wrong at first sight, but it's actually correct. I think that could use a comment.

CorsixTH/Luatest/spec/entities/humanoids/vip_spec.lua
```diff
@@ -41,34 +41,34 @@ local function create_room(name, args)
 
   if args.alive_plants then
     for _=1,args.alive_plants do
```
Alberth289346, line 43:
```suggestion
    for _ = 1, args.alive_plants do
```

CorsixTH/Luatest/spec/entities/humanoids/vip_spec.lua
```diff
@@ -41,34 +41,34 @@ local function create_room(name, args)
...
     for _=1,args.dying_plants do
-      room.objects[{object_type = {id = "plant"}, isDying = function() return true end}] = true
+      room.objects[{object_type = {id = "plant"}, isMachine = function() return false end, isDying = function() return true end}] = true
     end
   end
 
   if args.working_machines then
     for _=1,args.working_machines do
-      room.objects[{object_type = {id = "machine"}, strength=_, isBreaking = function() return false end}] = true
+      room.objects[{object_type = {id = "machine"}, strength=_, isMachine = function() return true end, isBreaking = function() return false end}] = true
```
Alberth289346, line 56:
```suggestion
      room.objects[{object_type = {id = "machine"}, strength = _, isMachine = function() return true end, isBreaking = function() return false end}] = true
```
Also, the `_` variable is only used for unused place holders. In other words, don't use it. Instead, use a proper variable name in the `for` statement above.

Last but not least, if you split the table literal into several lines, it's simpler to read.

Alberth289346: Alternatively, perhaps create local variables for the `return true` and `return false` functions, and use those variables instead of duplicating the function definition all the time.

CorsixTH/Lua/entities/machine.lua
Alberth289346: Euhm, why are you counting up to 4 extinguishers

CorsixTH/Lua/earthquake.lua
```diff
@@ -136,8 +136,8 @@ function Earthquake:tick()
 
   for _, room in pairs(self.world.rooms) do
     for object, _ in pairs(room.objects) do
-      if object.strength then
-        object:machineUsed(room)
+      if object:isMachine() then
+        object:earthquakeImpact(room)
```
Alberth289346, line 140: Not your fault, but this looks weird.
Any reason why we can't tell all objects about an earth quake? (Rather than just machines).

Alberth289346:
> Refactored some bad smelling values declaration like
`self.times_used = -1 -- Incremented in the call on the next line`
and
`self.total_usage = -1 -- Incremented in the constructor of Object.`
to
`self.xxx = 0`

You also eliminated that increment in the next line resp constructor?
Otherwise you modified behavior.

CorsixTH/Lua/entities/machine.lua
ARGAMX: > The first line is wrong in two ways.
> 
> * It repeats the function name almost literally. There is never a need to write the same thing twice. Eg
> 
> ```
> local j = 1  -- Assign 1 to j
> ```
> 
> Your audience are programmers, they can read and understand `local j = 1`. Don't repeat what the code is already telling. Instead explain the higher level reason. Why is j assigned, and/or why is it 1?
> 
> As you also don't want to explain the obvious, eg if 1 is the only obvious good value here, don't explain anything. On the other hand, if it is an unexpected value but it is correct for a complicated reason, then do explain.
> 
> The code explains what it does. The comments should explain why that is the correct thing to do.
> 
> * The text should be readable without the reader knowing all the magic words. I read "Impact", but I don't understand what that means.

```
--! The function is called when an earthquake hits the machine.. 
--! Function defines the machine's reaction to an earthquake.
--! During an earthquake, this function is called one or more times.
```

Is such description would be better?

CorsixTH/Lua/entities/machine.lua
ARGAMX: > The comment doesn't add much relative to the function name. I think it can do without comment here.
> 
> Don't know if you created `calculateSmoke`, but given your comment, `updateSmoke` may also be a good name.

It's not mine, I just moved it. 
But yeah, it could probably be improved a little.
May be `updateSmokeDisplay`

CorsixTH/Lua/entities/machine.lua
ARGAMX: > If the code jumped to `else` from `if must_explode`, then yes, the room obviously didn't explode. No need to state in comment what you can quite easily read in the code.
> 
> I am not sure where "urgent repair" is coming from.
> 
> My first intuition in writing a comment here would be something like "Check if it needs maintenance." You want to foremost describe what the next code line is doing. How you got here is not that important, usually. The reader should read the code and comments above here for that.

I didn't write this, this not mine.
But I'll remove it.

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > Please indent comment lines just like normal statement lines.

Yes, that's right.

CorsixTH/Lua/entities/machine.lua
ARGAMX: > The function name already says use counts are incremented.
> 
> The goal of comment is to explain **why** the code is correct in doing what it does. Avoid repeating what the code is already saying.
> 
> Another use of comment is tell the high level steps of a longer process. For example if you need to 5 things in sequence, but each thing takes a page of code, then writing a comment line what step is being done above a page of code is helpful in guiding the reader to the code block that they are looking for.

Reasonable.

CorsixTH/Lua/entities/machine.lua
ARGAMX: I like this

CorsixTH/Lua/entities/machine.lua
ARGAMX: > Not your fault, but it's nicer if comment stays together.

Reasonable

CorsixTH/Lua/entities/machine.lua
ARGAMX: > The task index is used twice, which makes creating a variable for it better.
> 
> An alternative here is a comment after the original line `-- There is no repair task.`. That clarifies what the code is doing without creating a new variable.

Didn't get that.
I have already made variables where it was possible.
What place are we talking about here?

CorsixTH/Lua/entities/machine.lua
ARGAMX: > Not your fault, but the `1` and `2` (probably priorities) are magic here.
> 
> I don't know if there is a `Task` class, but if there is, you can create a nice name for these.

These magic numbers are all over the code. It would be good to change 1 and 2 not just in this place, but everywhere for consistency.
But I've already gone far beyond the initial scope of the task.
And here I'll need to make changes in 15 more places.
Maybe it's better not to do it in scope of this PR?..
Also, it is here that the comment above clearly describes what 1 and 2 are, so they're not that magical here...

Overall, the idea is interesting.

CorsixTH/Lua/entities/machine.lua
ARGAMX: > -- Else if (low priority) repair is needed, make sure there is a handyman task for it.

Reasonable.

CorsixTH/Lua/entities/machine.lua
ARGAMX: This is mostly not my code, I just refactored it a bit.

CorsixTH/Lua/entities/machine.lua
ARGAMX: > I don't understand line 189.

I am too

CorsixTH/Lua/earthquake.lua
ARGAMX: > Not your fault, but this looks weird. Any reason why we can't tell all objects about an earth quake? (Rather than just machines).

Didn't get here what you mean.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -56,11 +57,13 @@ function Machine:setCrashedAnimation()
 end
 
 --! Calculates the number of times the machine can be used before crashing (unless repaired first)
+--!return (int) remaining uses count
 function Machine:getRemainingUses()
   return self.strength - self.times_used
 end
 
 --! Returns true if a machine is smoking/needs repair
```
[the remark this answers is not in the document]
ARGAMX: > This states what value you can expect when the call to this function returns. The usual place for that is after your added `!return` line.
> 
> In general, consider the doc comments as a description how to call the function (what does it do as a general line of text, then parameters and their value, and finally values returned by the function and their value.
> 
> "their value" here should describe what range of values are allowed or returned, and a value means without reading the source code of the function.

It seems that now I understand that writing comments is even more difficult than doing the task...

CorsixTH/Lua/entities/machine.lua
ARGAMX: > I don't understand this line

Agree. Me too. I am removing this.

CorsixTH/Lua/entities/machine.lua
ARGAMX: Thanks

CorsixTH/Lua/entities/machine.lua
ARGAMX: > Instead of skipping incrementing the count after reaching 4, break out of the loop when you reach 4.
> 
> That avoids inspecting room objects for nothing.

Wise

ARGAMX:
> You also eliminated that increment in the next line resp constructor? Otherwise you modified behavior.

Ofcourse I have eliminated.

CorsixTH/Lua/entities/machine.lua
ARGAMX: > It is only the `last` use if the room is indeed exploding.

Ok

CorsixTH/Luatest/spec/entities/humanoids/vip_spec.lua
ARGAMX: Changed

CorsixTH/Lua/entities/object.lua
Alberth289346: Here you use additional code to explain WHAT it does, rather than just do things.

To explain what it does text often wins, but it's a trade-off.
Different people have different ideas about the best balance is.

That is why I stated it as an alternative, so you are aware of it. Whether you follow it, is up to you.

CorsixTH/Lua/entities/machine.lua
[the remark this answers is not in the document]
Alberth289346: Well, code can be horribly complicated too :)  There are always many small details to decide.

Explanation is generally not that difficult, it just takes experience.

The most difficult part is understanding the code such that you can write the text as clear as possible.

CorsixTH/Lua/entities/machine.lua
Alberth289346: Looks very good.

👍🏿

CorsixTH/Lua/entities/machine.lua
Alberth289346: I always write about everything that I see. Here it is difficult to understand what you copied and what you wrote yourself, since large code blocks were moved around.

Feel free to ignore comments about parts you didn't write :)

CorsixTH/Lua/entities/machine.lua
Alberth289346: There is a trade-off between creating additional statement(s) or saying what the code is doing in 3 words.

At some point, the amount of additional code becomes too large or too tricky for the additional explanation that it gives.

There is no universal agreement which is better, it depends on the exact problem as well as the person writing it.

CorsixTH/Lua/entities/machine.lua
Alberth289346: It's a common problem. Once you start improving code you find more and more things to improve :)

It's fine to draw a line and leave some things for a next patch.

CorsixTH/Luatest/spec/entities/humanoids/vip_spec.lua
Alberth289346: It's fine. I just point out things I notice.

Unless there is something truly broken, I usually don't make hard demands for changes.

Alberth289346:
Very good work!

LGTM

Alberth289346 (approved):

PR #2804, opened 2025-03-26
Fix campaign subfolders

tobylane:
*Fixes #*

- Set the origin folder for a custom campaign. This is the last part of permitting campaigns to be in their own folder.

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/fax.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/fax.lua
@@ -216,10 +216,10 @@ function UIFax:choice(choice_number)
       local campaign_info = self.ui.app.world.campaign_info
       for i, level in ipairs(campaign_info.levels) do
         if self.ui.app.world.map.level_filename == level then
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/dialogs/fullscreen/fax.lua
```diff
@@ -216,10 +216,10 @@ function UIFax:choice(choice_number)
       local campaign_info = self.ui.app.world.campaign_info
       for i, level in ipairs(campaign_info.levels) do
         if self.ui.app.world.map.level_filename == level then
-          local level_info, _ = self.ui.app:readLevelFile(campaign_info.levels[i + 1])
+          local level_info, _ = self.ui.app:readLevelFile(campaign_info.levels[i + 1], campaign_info.folder)
```
lewri, line 219: Existing campaigns will be missing this folder, no?

tobylane: Yes, but until the previous campaigns PR, they were all in the two folders that are checked after that parameter.

lewri (approved):

Issue #2488, opened 2024-02-11
Demo game has incorrect endConditions


lewri:
Looks like CorsixTH actually loads the base_config for the demo, which give no win/lose condition.
But a different issue is that the Demo has a defined endCondition in the readme:
```txt
You have to cure 20 patients in a year in this demo so look after your
patients well and provide for their every need.
```
And in the demo itself from DOSBOX
![image](https://github.com/CorsixTH/CorsixTH/assets/20030128/f80942db-e112-47c9-81c5-eca6866e9ac7)

I imagine it has an inferred lose condition of not treating 20 patients in 1 year is a lose game scenario. However, we don't have a method yet to implement that.

Alberth289346:
It could equally well be that initially they added the lose condition, and after evaluation they dropped it and forgot it's also mentioned in the text.

Then again, just mentioning it makes people take more effort to avoid it, so it may have worked after all :)

ARGAMX:
Do these win/lose conditions work in the original demo game?
If these conditions are met, does the demo game end in a win/loss?

tobylane:
This can be done now, but I'd prefer an open ended demo.

This video has a win at the end of the third quarter https://www.youtube.com/watch?v=k5WdsI6UDAM

Thia video loses mid June, don't know why https://youtu.be/xfyWdYhvq1o?t=779 But later loses at the end of the year https://youtu.be/xfyWdYhvq1o?t=2167

tobylane:
There could be a faithful demo level beside the open ended one, with these conditions, in a restricted Single Scenario list.

> #win_criteria[0].Criteria.MaxMin.Value.Group.Bound	4	1	20	1	0
#lose_criteria[0].Criteria.MaxMin.Value.Group.Bound	9	1	12	1	0

lewri:
Either faithful targets, or implement the demo game target as an optional check during launch? It only had a singular goal and I can't remember how difficult it was.

PR #2805, opened 2025-03-26
Add new criterion - year

tobylane:
- Add new endcondition criterion for the year. Is it clearer as complete years, without a decimal point for progress in the current year?  
The year needs 0.99 removed to work for the comparisons in checking win and lose. As with the internal values for year, it starts at 0 not 2000 and is only changed to this millennium for display.  
A requirement for over a certain year can be combined with others as a grace period, eg cure percentage must be over 60% after ten years
Examples for ending after 3 years.
#win_criteria[0].Criteria.MaxMin.Value.Group.Bound 9 1 3 1 0
#lose_criteria[0].Criteria.MaxMin.Value.Group.Bound 9 1 3 1 0

### DIFF
--- a/CorsixTH/Lua/date.lua
+++ b/CorsixTH/Lua/date.lua
@@ -263,6 +263,13 @@ function Date:isSameDay(other)
   return self._year == other._year and self._month == other._month and self._day == other._day
 end
 
+--[[ Calculates the number of months played until a time an hour in the future. This is used in EndConditions
+!return (number) Integer of months since Jan 2000
+]]
+function Date:getProgressInMonths()
+  return self:plusHours(1):monthOfGame()
+end
+
 -- METAMETHODS
 
 local Date_mt = Date._metatable
--- a/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
@@ -40,7 +40,8 @@ function UIProgressReport:UIProgressReport(ui)
     -- Load all sprite tables needed for all goal icons
     self.panel_sprites_table = {
       MPointer = gfx:loadSpriteTable("Data", "MPointer"),
-      Rep02V = gfx:loadSpriteTable("QData", "Rep02V", true, palette)
+      Rep02V = gfx:loadSpriteTable("QData", "Rep02V", true, palette),
+      Font04V = gfx:loadSpriteTable("QData", "Font04V"),
     }
     self.panel_sprites = self.panel_sprites_table.Rep02V -- The default goals icons
   end) then
@@ -55,6 +56,13 @@ function UIProgressReport:UIProgressReport(ui)
   self.selected = 1
   local hospital = world.hospitals[self.selected]
 
+  -- Get the localised name for the month
+  local function get_month_string(num)
+    num = math.fmod(num, 12)
+    if num == 0 then num = 12 end
+    return _S.months[num]
+  end
+
   -- Collect which criteria to show here, draw columns in draw
   -- Add the icons for the criteria
   local x = 263
@@ -68,11 +76,11 @@ function UIProgressReport:UIProgressReport(ui)
       crit_table.red = true
       res_value = crit_table.lose_value
     end
-    -- FIXME: res_value and cure_value are depersisted as floating points, using
+    -- FIXME: res_value and cur_value are depersisted as floating points, using
     -- string.format("%.0f", x) is not suitable due to %d (num) param in _S string
-    local tooltip
+    local tooltip, direction
     if crit_table.two_tooltips then
-      local direction = crit_table.win_value and "over" or "under"
+      direction = crit_table.win_value and "over" or "under"
       tooltip = _S.tooltip.status[direction][crit_name]
     else
       tooltip = _S.tooltip.status[crit_name]
@@ -82,6 +90,15 @@ function UIProgressReport:UIProgressReport(ui)
     elseif crit_table.formats == 3 then
       tooltip = tooltip:format(math.floor(res_value) / 1000,
           math.floor(cur_value) / 1000)
+    elseif crit_table.formats == 4 then -- Criterion months_played
+      -- Years are shown starting at 2000
+      if math.fmod(res_value, 12) == 0 then -- For whole years, use years_played string
+        tooltip = _S.tooltip.status[direction].years_played
+            :format(math.floor(res_value / 12) + 2000, math.floor((cur_value - 1) / 12) + 2000)
+      else
+        tooltip = tooltip:format(get_month_string(res_value), math.floor(res_value / 12) + 2000,
+          get_month_string(cur_value), math.floor((cur_value - 1) / 12) + 2000)
+      end
     else
       tooltip = tooltip:format(math.floor(res_value))
     end
--- a/CorsixTH/Lua/endconditions.lua
+++ b/CorsixTH/Lua/endconditions.lua
@@ -32,13 +32,15 @@ local local_criteria_variable = {
     -- New criteria
   {name = "staff_happiness",  icon = 16, icon_file = "MPointer", formats = 2, two_tooltips = true},
   {name = "patient_happiness",icon = 18, icon_file = "MPointer", formats = 2, two_tooltips = true},
+  {name = "months_played",    icon = 46, icon_file = "Font04V",  formats = 4, two_tooltips = true},
 }
 
 -- A table of functions for fetching criteria values that cannot be measured
---   directly from a hospital attribute of the same name.
+-- directly from a hospital attribute of the same name.
 local get_custom_criteria = {
   staff_happiness = function(hospital) return 100 * hospital:getAverageStaffAttribute("happiness", 0.5) end,
   patient_happiness = function(hospital) return 100 * hospital:getAveragePatientAttribute("happiness", 0.5) end,
+  months_played = function(hospital) return hospital.world.game_date:getProgressInMonths() end,
 }
 
 class "EndConditions"
--- a/CorsixTH/Lua/languages/english.lua
+++ b/CorsixTH/Lua/languages/english.lua
@@ -872,6 +872,7 @@ information = {
     cheat = "Hope you didn't click the Lose Level button by accident!",
     staff_happiness = "Your average staff happiness fell below %d%.",
     patient_happiness = "Your average patient happiness fell below %d%.",
+    months_played = "You didn't meet the level goals within %d months.",
   },
   cheat_not_possible = "You cannot use that cheat on this level.",
 }
@@ -1089,10 +1090,14 @@ tooltip.status = {
   over = {
     staff_happiness = "Your average staff happiness should be over %d%. Currently it's %d%",
     patient_happiness = "Your average patient happiness should be over %d%. Currently it's %d%",
+    months_played = "You need to reach the end of %s %d. Currently it's %s %d",
+    years_played = "You need to reach the year %d. Currently it's the year %d",
   },
   under = {
     staff_happiness = "Your average staff happiness should not be less than %d%. Currently it's %d%",
     patient_happiness = "Your average patient happiness should not be less than %d%. Currently it's %d%",
+    months_played = "You need to win by the end of %s %d. Currently it's %s %d",
+    years_played = "You need to reach the year %d. Currently it's the year %d",
   }
 }
 --------------------------------  UNUSED  -----------------------------------


TheCycoONE:
Are you using - 0.99 to get something like the floor? I don't quite understand though you described it.

Matroftt:
> What motivation for this feature in overall? Which benefits?

This is great, it will be possible to create, for example, such scenario which requires to cure 500 people in 3 years

tobylane:
> Are you using - 0.99 to get something like the floor? I don't quite understand though you described it.

The check for lose goal needs the measure to be over the target, where the win goal can just equal it. The start of year 4 is measured as 3.01, exceeding 3, causing a game to end. The wording of the tooltips is that the start of the year is needed for a win, but the end of the year for a loss. If this isn't acceptable, I'll look for an alternative solution to the measure function in get_custom_criteria.

TheCycoONE:
> > Are you using - 0.99 to get something like the floor? I don't quite understand though you described it.
> 
> The check for lose goal needs the measure to be over the target, where the win goal can just equal it. The start of year 4 is measured as 3.01, exceeding 3, causing a game to end. The wording of the tooltips is that the start of the year is needed for a win, but the end of the year for a loss. If this isn't acceptable, I'll look for an alternative solution to the measure function in get_custom_criteria.

It's acceptable but please comment as it's not obvious.

Alberth289346:
> The year needs 0.99 removed to work for the comparisons in checking win and lose.

Why not hide that in the implementation?

If a user wants 3 years, they write "3". That's the intuitive solution to a user. If the code cannot handle that, change code rather than forcing the user to enter non-intuitive values.

Not all users read everything or understand everything. Don't assume they will do the right thing when you force them in (for them) weird solutions.  
Trying that anyway just causes an endless amount of problems with levels not working as designed.

Alberth289346:
Why not use "month" or even "day"? It gives more options to the designer without much additional effort for us.

lewri:
> Why not use "month" or even "day"? It gives more options to the designer without much additional effort for us.

The original actually works in months for other settings. Month might be a more expected fit.

tobylane:
> Why not hide that in the implementation?

It is hidden, but as it needs a further explanation, it should be done another way. Eg the measure function is told it's measuring for a lose comparison, so it can adjust differently to measuring for win.

> Why not use "month" or even "day"? It gives more options to the designer without much additional effort for us.

The level writer's target needs to be a number (unless we assume they wrote something that can be treated as a date type), so it would be progress through the year. This can be converted to years and months and days in the tooltip.

lewri:
NB: assists with #2488

tobylane:
Added Date:getProgress to measure the days of the year accurately, and a tiny amount is added to this in the measure function to cause a lose end game in the correct year.  
The column height now changes each day, but the tooltips are set at dialog creation to just the year. As the dialog is closed at the end of the year for the hall of fame, it's never out of date. Rebased.

CorsixTH/Lua/date.lua
```diff
@@ -48,6 +48,22 @@ local month_length = {
...
+  90,
+  120,
+  151,
+  181,
+  212,
+  243,
+  273,
+  304,
+  334,
+}
```
Alberth289346, line 65: Don't add magic numbers that need to be kept in sync with other magic numbers.

Compute this table from the table above.

CorsixTH/Lua/date.lua
```diff
@@ -263,6 +279,13 @@ function Date:isSameDay(other)
   return self._year == other._year and self._month == other._month and self._day == other._day
 end
 
+--[[ Calculates the progress through the year as a decimal (including current day)
+!return (number)
+]]
+function Date:getProgress()
+  return self._year - 1 + (cumulative_days[self._month] + self._day) / 365
```
Alberth289346, line 286: Don't use a magic `365` that depends on previous magic numbers.
Compute it.

CorsixTH/Lua/date.lua
```diff
@@ -263,6 +279,13 @@ function Date:isSameDay(other)
   return self._year == other._year and self._month == other._month and self._day == other._day
 end
 
+--[[ Calculates the progress through the year as a decimal (including current day)
+!return (number)
```
Alberth289346, line 283: What does "decimal" mean?  To me it means you write a number in base 10.

Base 10 or eg base 16 is not about the value that you return, it's just about notation when you write the value on a piece of paper.

I think what you are looking for is "as a real" or "as floating point".

What is missing here is the unit. What does `1.0` mean? If I happen to pick 10 days as unit, it matches your description, but the code doesn't do that.

Also "through the year" can be interpreted as the fraction within the current year (ie it starts at 0 again each year) rather than adding all years from the start of the game (ie cumulative).

I'd suggest to keep the explanation short, and move the more technical details about meaning of a returned value after the "!return" line.
That also nicely fixes the problem of an empty return description.

CorsixTH/Lua/endconditions.lua
```diff
@@ -32,13 +32,15 @@ local local_criteria_variable = {
     -- New criteria
   {name = "staff_happiness",  icon = 16, icon_file = "MPointer", formats = 2, two_tooltips = true},
   {name = "patient_happiness",icon = 18, icon_file = "MPointer", formats = 2, two_tooltips = true},
+  {name = "year",             icon = 58, icon_file = "Font04V",  formats = 4, two_tooltips = true}, -- Starts from 0 not 2000
```
Alberth289346, line 35: If you rename to "years_played" (or something like that), you avoid the ambiguity between 0 and 2000.

CorsixTH/Lua/endconditions.lua
```diff
@@ -32,13 +32,15 @@ local local_criteria_variable = {
...
 }
 
 -- A table of functions for fetching criteria values that cannot be measured
 --   directly from a hospital attribute of the same name.
 local get_custom_criteria = {
   staff_happiness = function(hospital) return 100 * hospital:getAverageStaffAttribute("happiness", 0.5) end,
   patient_happiness = function(hospital) return 100 * hospital:getAveragePatientAttribute("happiness", 0.5) end,
+  year = function(hospital) return hospital.world.game_date:getProgress() + 0.001 end,
```
Alberth289346, line 43: The magic `0.001` needs a comment at least.
A new programmer is not going to understand why it's there.

I am not sure either :p

CorsixTH/Lua/endconditions.lua
```diff
@@ -32,13 +32,15 @@ local local_criteria_variable = {
     -- New criteria
   {name = "staff_happiness",  icon = 16, icon_file = "MPointer", formats = 2, two_tooltips = true},
   {name = "patient_happiness",icon = 18, icon_file = "MPointer", formats = 2, two_tooltips = true},
+  {name = "year",             icon = 58, icon_file = "Font04V",  formats = 4, two_tooltips = true}, -- Starts from 0 not 2000
 }
 
 -- A table of functions for fetching criteria values that cannot be measured
 --   directly from a hospital attribute of the same name.
```
Alberth289346, line 39: With your addition, this comment is now false.

While at it, remove the weird indentation at the 2nd line.

CorsixTH/Lua/languages/english.lua
```diff
@@ -808,6 +808,7 @@ information = {
     cheat = "Hope you didn't click the Lose Level button by accident!",
     staff_happiness = "Your average staff happiness fell below %d%.",
     patient_happiness = "Your average patient happiness fell below %d%.",
+    year = "You didn't meet the level goals within %d years.",
```
Alberth289346, line 811: Maybe `number_of_years` ?

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -82,6 +83,8 @@ function UIProgressReport:UIProgressReport(ui)
     elseif crit_table.formats == 3 then
       tooltip = tooltip:format(math.floor(res_value) / 1000,
           math.floor(cur_value) / 1000)
+    elseif crit_table.formats == 4 then
+      tooltip = tooltip:format(res_value + 2000, math.floor(cur_value) + 2000)
```
Alberth289346, line 87: Magic `2000` looks like a value that should be in `Date`.

CorsixTH/Lua/endconditions.lua
Alberth289346: Another option here is to add the base year in `Date`, and do everything in the absolute year. That may need to subtract the base year again in some computations, not sure.

Alberth289346:
I wrote several things about the dates that you use that don't entirely agree with each other. It is probably best if you first read all comments before starting to change things.

CorsixTH/Lua/endconditions.lua
tobylane: If the player has killed 10 patients on a level where a lose condition is killing 10 patients, they haven't _exceeded_ the condition yet. I think reaching day 1 of the year should count as exceeding it, but this is inconsistent with other criteria (the alternative reasoning is to say that once the player is past the first day of the year, any check of end game will result in losing, such as at the end of the year). A winning condition of kill 10 patients will be met when 10 have been killed, it only needs to be equalled, not exceeded. Adding a small number, less than a day, means that the whole number is not reached any sooner, the column height isn't changed. I'll add this as a general note to the wiki page for new criteria.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
tobylane: Wherever the year is shown, 1999 is added, as that's the real magic number. I'm using 2000 because I think it's easier on level editors for 5 to mean 2005 not 2004. I don't think it's right to have two magic numbers with the same meaning, so I did neither.

tobylane:
While I am continuing with two assumptions I made, I can see the case for reversing them. The number the level editor (person, not tool) uses to represent ingame years is 2000 off, not 1999 off as other ingame representations of the date are. The number the level editor uses should be consistent in win and lose, the same number in each means the game ends in the same year. Both of these make up for previous coding choices that didn't have a downside before.

CorsixTH/Lua/endconditions.lua
ARGAMX: > If the player has killed 10 patients on a level where a lose condition is killing 10 patients, they haven't _exceeded_ the condition yet.

Another option: 
Why we we don't want shift this logic to place where comparison is happening? 
I mean in that place we can use `>`, instead of `=>` to satisfy the exceeded logic.

Alberth289346: There is a smart trick in the comparison code to check for winning and losing conditions with the same operator. That may be a contributing factor to the complications here, but not sure.

TheCycoONE:
> While I am continuing with two assumptions I made, I can see the case for reversing them. The number the level editor (person, not tool) uses to represent ingame years is 2000 off, not 1999 off as other ingame representations of the date are. The number the level editor uses should be consistent in win and lose, the same number in each means the game ends in the same year. Both of these make up for previous coding choices that didn't have a downside before.

Probably want to be consistent with StartMonth/EndMonth in the level files here. It would be weird not to be.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
ARGAMX: I believe that using number without any explanation makes a number magical is the.
If there is an explanation or a variable name that gives it a name, the magic is gone.
So even if you do:
`local initial_year = 2000`
and then continue to operate on `initial_year`, there is significantly magic reducing.
And if you even write a comment to clarify why exactly 2000...
At a quick overview without context, it is not immediately obvious that 2000 is a year at all.

tobylane: I've added a comment on this line and in endconditions.lua to clarify this.

lewri: If you're going to do this, this is a time where you should be labelling the number probably so it provides some additional clarity. `start_year`, or `initial_year`; either would be fine.

CorsixTH/Lua/languages/english.lua
```diff
@@ -1021,10 +1022,12 @@ tooltip.status = {
   over = {
     staff_happiness = "Your average staff happiness should be over %d%. Currently it's %d%",
     patient_happiness = "Your average patient happiness should be over %d%. Currently it's %d%",
+    years_played = "You need to reach the year %d. Currently it's %d",
   },
   under = {
     staff_happiness = "Your average staff happiness should not be less than %d%. Currently it's %d%",
     patient_happiness = "Your average patient happiness should not be less than %d%. Currently it's %d%",
+    years_played = "You need to win by the year %d. Currently it's %d",
```
lewri: I don't think your PR summary explains clearly enough these are two different criteria being checked. E.g. Survive x years or Fulfil all other criteria within x years.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -40,7 +40,8 @@ function UIProgressReport:UIProgressReport(ui)
     -- Load all sprite tables needed for all goal icons
     self.panel_sprites_table = {
       MPointer = gfx:loadSpriteTable("Data", "MPointer"),
-      Rep02V = gfx:loadSpriteTable("QData", "Rep02V", true, palette)
+      Rep02V = gfx:loadSpriteTable("QData", "Rep02V", true, palette),
+      Font04V = gfx:loadSpriteTable("QData", "Font04V"),
```
lewri, line 44: Adding this requires closing the progress report for compatibility doesn't it?

CorsixTH/Lua/endconditions.lua
```diff
@@ -32,13 +32,16 @@ local local_criteria_variable = {
...
+  {name = "years_played",     icon = 58, icon_file = "Font04V",  formats = 4, two_tooltips = true},
 }
 
 -- A table of functions for fetching criteria values that cannot be measured
---   directly from a hospital attribute of the same name.
+-- directly from a hospital attribute of the same name.
 local get_custom_criteria = {
   staff_happiness = function(hospital) return 100 * hospital:getAverageStaffAttribute("happiness", 0.5) end,
   patient_happiness = function(hospital) return 100 * hospital:getAveragePatientAttribute("happiness", 0.5) end,
+  years_played = function(hospital) return hospital.world.game_date:getProgress() + 0.001 end,
```
lewri: I'm still a bit iffy in this method. Would it be possible to use `Date:plusDays(1)` instead and make an evaluation based on that?

You also have `Date:isLastDayOfYear()` in the class.

CorsixTH/Lua/date.lua
```diff
@@ -263,6 +264,13 @@ function Date:isSameDay(other)
   return self._year == other._year and self._month == other._month and self._day == other._day
 end
 
+--[[ Calculates the complete and partial years up to the current date
+!return (number) Floating point number of years since 2000
+]]
+function Date:getProgress()
```
lewri:
```suggestion
function Date:getProgressInYears()
```
?

lewri (commented):
I do apologise. But I'm still questioning why we're focussed on 'year' when the original game's files, and our custom ones, work in months.

Population: months
Visual diseases: months
Staff for hire: months
Earthquakes: months
Emergencies: months
Training (original only): months

A creator is going to expect definition values to work generally in months. Why are we bunking this now? Months are still in integer format.

Though I will also grant you the inverse of this where this feature makes sense: A player is going to work in years for a progress report. It's a troublesome division.

Generally happy to proceed on this path with caution.

TheCycoONE:
> apologise. But I'm still questioning why we're focussed on 'year' when the original game's files, and our custom ones, work in months.
...
> 
> reator is going to expect definition values to work generally in months. Why are we bunking this now? Months are still in integer format.
> 

I think this is the crux of it. Every 3 months is a valid option and every other setting in the file works this way.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
lewri: (Arguably, no game using this should exist before the change but I'm also cautious)

tobylane:
Screenshot 2025-06-23 at 18 40 06

This is months played. Is the years in the tooltip useful?

> I'm still a bit iffy in this method. Would it be possible to use Date:plusDays(1) instead and make an evaluation based on that?

`self:plusHours(1):monthOfGame()` would do the job. Other than with testing as in that screenshot, it's highly unlikely this will be used with such small numbers, where moving ahead an hour early at the end of the month would be visible.

TheCycoONE:
I suspect we can deal with the display side with a special formatting function and two or three strings %d years, %d months and another for just months, maybe one for just years. Pick based on how the math works.

Edit: saw your example. I like it when the months work out to even years I still suggest switching the tooltip.

tobylane:
> I suspect we can deal with the display side with a special formatting function and two or three strings %d years, %d months and another for just months, maybe one for just years. Pick based on how the math works.

Screenshot 2025-10-23 at 18 35 59
Screenshot 2025-10-23 at 18 35 32

```lua
    elseif crit_table.formats == 4 then -- Criterion months_played
      -- The months_played string is used if the goal is 3 years or less, or in the final year
      if res_value <= 36 or (res_value - cur_value) <= 12 then
        tooltip = tooltip:format(res_value, math.floor(res_value / 12), cur_value)
      else
        tooltip = _S.tooltip.status[direction].years_played
            :format(math.floor(res_value / 12), math.floor(cur_value / 12))
      end
```

This uses years, unless it's a short level or the end is near when it shows months. Is a year enough warning of a mid-year end date?

tobylane:
```lua
    elseif crit_table.formats == 4 then -- Criterion months_played
      -- The months_played string is used if the goal is 3 years or less, or in the final year
      -- Years are shown starting at 2000
      if res_value <= 36 or (res_value - cur_value) <= 12 then
        tooltip = tooltip:format(res_value, math.floor(res_value / 12) + 2000, cur_value)
      else
        tooltip = _S.tooltip.status[direction].years_played
            :format(math.floor(res_value / 12) + 2000, math.floor(cur_value / 12) + 2000)
      end
```
In a change from the previous images, this shows the year as the user expects, with better grammar. Rebased for the checks.

Alberth289346:
> This uses years, unless it's a short level or the end is near when it shows months. Is a year enough warning of a mid-year end date?

37 and 40 don't have much meaning to me. Any particular reason why you don't always break it down to year+month ?

Or even say the month name instead of "3 years, 5 months" (eg `may year 3`)?

EDIT: Possibly even "may 1st" (since "may 31st" is a lot further away, and a user may think the latter)

tobylane:
Screenshot 2026-01-25 at 10 46 32
Screenshot 2026-01-25 at 10 46 03

This uses years when the target is a year end (which I expect to be the norm), otherwise month names plus year.

CorsixTH/Lua/languages/english.lua
```diff
@@ -1089,10 +1090,14 @@ tooltip.status = {
   over = {
     staff_happiness = "Your average staff happiness should be over %d%. Currently it's %d%",
     patient_happiness = "Your average patient happiness should be over %d%. Currently it's %d%",
+    months_played = "You need to reach the end of %s %d. Currently it's %s %d",
```
lewri, line 1093: Are there any languages where it'd be reversed? (the `%s %d` component) None come to my mind but I'm no expert.

lewri: Actually, can you check if we fall into the `.0` problem here?

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
lewri, line 44: Is Font04V in the demo files? As this new criteria is needed for the true demo experience.

lewri (commented):
Nice, simpler.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
tobylane: The demo can't open this dialog.

lewri: Yep, realised now, they just get a information box with cures in
image

CorsixTH/Lua/languages/english.lua
tobylane: CJK languages may do. There's a math.floor to avoid a .0.

TheCycoONE (approved):

CorsixTH/Lua/endconditions.lua
```diff
@@ -32,13 +32,15 @@ local local_criteria_variable = {
     -- New criteria
   {name = "staff_happiness",  icon = 16, icon_file = "MPointer", formats = 2, two_tooltips = true},
   {name = "patient_happiness",icon = 18, icon_file = "MPointer", formats = 2, two_tooltips = true},
+  {name = "months_played",    icon = 46, icon_file = "Font04V",  formats = 4, two_tooltips = true},
```
lewri, line 35: I'm guessing the way this is called means Unicode doesn't overwrite this.

How does this work with the Russian edition?

lewri: Tag @ARGAMX for this one

ARGAMX: > I'm guessing the way this is called means Unicode doesn't overwrite this.
> 
> How does this work with the Russian edition?

Indeed, this could be a bit problematic. 
Especially if we add support for other versions built in other languages.
But I honestly have no idea how this could be implemented better.
Perhaps only creating our own icon.

In this particular case, we were lucky and everything went smoothly for "mik".
Untitled

TheCycoONE: Good enough until we make a bitmap.

PR #2808, opened 2025-03-29
Refactor drawable draw / hit_test / is_multiple_frame_animation methods

Alberth289346:
- Replaced the 'drawable' base class function pointers by abstract virtual methods.
- Implemented them in 'animation' and 'sprite_render_list'.
- Added an enumeration to 'animation' to track the kind of animation that is stored.

- Deleted all the functions of the removed function pointers.

- Simplify extending to secondary marker support.

### DIFF
--- a/CorsixTH/Lua/api_version.lua
+++ b/CorsixTH/Lua/api_version.lua
@@ -32,4 +32,4 @@ Note: This file compiles as both Lua and C++. */
 
 #endif /*]] --*/
 
-return 2688;
+return 2689;
--- a/CorsixTH/Lua/api_version.lua
+++ b/CorsixTH/Lua/api_version.lua
@@ -32,4 +32,4 @@ Note: This file compiles as both Lua and C++. */
 
 #endif /*]] --*/
 
-return 2689;
+return 2688;
--- a/CorsixTH/Src/th_gfx.cpp
+++ b/CorsixTH/Src/th_gfx.cpp
@@ -1344,7 +1344,8 @@ void animation::depersist(lua_persist_reader* pReader) {
   do {
     // Read the chain
     if (!pReader->read_stack_object()) break;
-    next = reinterpret_cast<link_list*>(lua_touserdata(L, -1));
+
+    next = get_animation_base_from_userdata(L, -1);
     if (next) next->prev = this;
     lua_pop(L, 1);
 
@@ -1827,7 +1828,7 @@ void sprite_render_list::depersist(lua_persist_reader* pReader) {
     return;
   }
 
-  next = reinterpret_cast<link_list*>(lua_touserdata(L, -1));
+  next = get_animation_base_from_userdata(L, -1);
   if (next) {
     next->prev = this;
   }
--- a/CorsixTH/Src/th_gfx.cpp
+++ b/CorsixTH/Src/th_gfx.cpp
@@ -1285,7 +1285,7 @@ void animation::persist(lua_persist_writer* pWriter) const {
   // Write the drawable fields
   pWriter->write_uint(flags);
 
-  if (anim_kind  == animation_kind::normal) {
+  if (anim_kind == animation_kind::normal) {
     pWriter->write_uint(1);
   } else if (anim_kind == animation_kind::child) {
     pWriter->write_uint(2);
@@ -1656,8 +1656,7 @@ void animation_base::set_layer(int iLayer, int iId) {
   }
 }
 
-sprite_render_list::sprite_render_list() : animation_base() {
-}
+sprite_render_list::sprite_render_list() : animation_base() {}
 
 sprite_render_list::~sprite_render_list() { delete[] sprites; }
 
--- a/CorsixTH/Src/th_gfx.cpp
+++ b/CorsixTH/Src/th_gfx.cpp
@@ -1249,55 +1249,6 @@ bool animation::hit_test_morph(int iDestX, int iDestY, int iTestX, int iTestY) {
          morph_target->hit_test(iDestX, iDestY, iTestX, iTestY);
 }
 
-namespace {
-bool THAnimation_hit_test_child(drawable* pSelf, int iDestX, int iDestY,
-                                int iTestX, int iTestY) {
-  return reinterpret_cast<animation*>(pSelf)->hit_test_child(iDestX, iDestY,
-                                                             iTestX, iTestY);
-}
-
-void THAnimation_draw_child(drawable* pSelf, render_target* pCanvas, int iDestX,
-                            int iDestY) {
-  reinterpret_cast<animation*>(pSelf)->draw_child(pCanvas, iDestX, iDestY);
-}
-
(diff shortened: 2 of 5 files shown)


Alberth289346:
I am not quite sure what is desired. I implemented the override methods in terms of calls to existing methods, but obviously, more refactoring is possible with inlining the existing methods (if that is possible, I didn't check).

CorsixTH/Lua/api_version.lua
```diff
@@ -32,4 +32,4 @@ Note: This file compiles as both Lua and C++. */
 
 #endif /*]] --*/
 
-return 2688;
+return 2689;
```
TheCycoONE, line 35: I don't think this does actually change the c++/lua api does it?

TheCycoONE:
Virtual functions cannot be inlined unless you mean to remove it from the base class somehow.

Good change, thanks!

CorsixTH/Lua/api_version.lua
Alberth289346: Indeed as far as I know, but when loading a current master savegame I got
```
Warning: THMap linked-lists are corrupted.
Warning: THMap linked-lists are corrupted.
Warning: THMap linked-lists are corrupted.
Warning: THMap linked-lists are corrupted.
Warning: THMap linked-lists are corrupted.
Warning: THMap linked-lists are corrupted.
Savegame version is 212 (Trunk), originally it was 212 (Trunk).
Segmentation fault (core dumped)
```

GDB (with abbreviated file paths):
```
Core was generated by `.../build/CorsixTH/corsix-th --i'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00005dd96a209190 in ?? ()
[Current thread is 1 (Thread 0x723a08346900 (LWP 9965))]
(gdb) bt
#0  0x00005dd96a209190 in ??? ()
#1  0x00005dd93ea10865 in level_map::draw
    (this=0x5dd96a266c38, pCanvas=0x5dd9676f2578, iScreenX=-1203, iScreenY=1621, iWidth=1575, iHeight=1174, iCanvasX=0, iCanvasY=0) at .../CorsixTH/Src/th_map.cpp:1094
#2  0x00005dd93e9fe790 in (anonymous namespace)::l_map_draw (L=0x5dd96a9298e8)
    at .../CorsixTH/Src/th_lua_map.cpp:741
#3  0x0000723a1474cf7e in ??? () at /lib/x86_64-linux-gnu/liblua5.3.so.0

(gdb) frame 1
(gdb) l
1090	      bool bRedrawAnimations = false;
1091	
1092	      drawable* pItem = static_cast<drawable*>(itrNode->entities.next);
1093	      while (pItem) {
1094	        pItem->draw_fn(pCanvas, itrNode.x(), itrNode.y());
1095	        if (pItem->is_multiple_frame_animation_fn()) {
1096	          bRedrawAnimations = true;
1097	        }
1098	        if (pItem->get_drawing_layer() == 1) {
```
It crashes while drawing the map.

I am guessing that the size of one or more classes changed due to the added virtual methods. No idea what assumptions the persistence code has.

Saving and loading from the branch seems to work fine.

Alberth289346:
> Virtual functions cannot be inlined unless you mean to remove it from the base class somehow.

I was more considering the various `draw` and `hit_test` methods that are now wrapped.
As an experiment I made those methods private:
``` diff
diff --git a/CorsixTH/Src/th_gfx.h b/CorsixTH/Src/th_gfx.h
index 9e9666c9..f2453818 100644
--- a/CorsixTH/Src/th_gfx.h
+++ b/CorsixTH/Src/th_gfx.h
@@ -565,6 +565,7 @@ class animation : public animation_base {
   void set_parent(animation* pParent);
 
   void tick();
+ private:
   void draw(render_target* pCanvas, int iDestX, int iDestY);
   bool hit_test(int iDestX, int iDestY, int iTestX, int iTestY);
   void draw_morph(render_target* pCanvas, int iDestX, int iDestY);
@@ -572,6 +573,7 @@ class animation : public animation_base {
   void draw_child(render_target* pCanvas, int iDestX, int iDestY);
   bool hit_test_child(int iDestX, int iDestY, int iTestX, int iTestY);
 
+ public:
   void draw_fn(render_target* pCanvas, int iDestX, int iDestY) override {
     switch (anim_kind) {
       case animation_kind::normal:
@@ -655,7 +657,9 @@ class sprite_render_list : public animation_base {
   ~sprite_render_list();
 
   void tick();
+ private:
   void draw(render_target* pCanvas, int iDestX, int iDestY);
+ public:
   bool hit_test(int iDestX, int iDestY, int iTestX, int iTestY);
 
   void set_sheet(sprite_sheet* pSheet) { sheet = pSheet; }
```
That showed one case of a draw call from a templated function that I then also modified:
``` diff
diff --git a/CorsixTH/Src/th_lua_anims.cpp b/CorsixTH/Src/th_lua_anims.cpp
index 24f14389..428ed736 100644
--- a/CorsixTH/Src/th_lua_anims.cpp
+++ b/CorsixTH/Src/th_lua_anims.cpp
@@ -591,7 +591,7 @@ template <typename T>
 int l_anim_draw(lua_State* L) {
   T* pAnimation = luaT_testuserdata<T>(L);
   render_target* pCanvas = luaT_testuserdata<render_target>(L, 2);
-  pAnimation->draw(pCanvas, static_cast<int>(luaL_checkinteger(L, 3)),
+  pAnimation->draw_fn(pCanvas, static_cast<int>(luaL_checkinteger(L, 3)),
                    static_cast<int>(luaL_checkinteger(L, 4)));
   lua_settop(L, 1);
   return 1;
```
That doesn't resolve the crash of old saves though.

CorsixTH/Lua/api_version.lua
TheCycoONE: hmm, that makes me want to load with asan on, because I'm not sure how the app version would affect save loading.

Alberth289346: Sorry, I didn't mention it, but I got this WITHOUT the API version change.

Still, it shouldn't happen, should it?

TheCycoONE: It should not. That's why I wonder if it's always performing bad memory access and only sometimes crashing. Asan or valgrind should confirm.

Alberth289346: I did some random digging, and this looks suspicious:

In template `lua_class_binding`:

```
/* Set the depersist size */
lua_pushinteger(pState->L, sizeof(T));
```
I added virtual methods to derived classes, so `sizeof(T)` in depersisting is likely not what gets pushed here.

TheCycoONE: This is the new size; as discussed in the channel.

TheCycoONE:
I pushed a fix for the segfaults and other errors that we were getting.

It is not allowed to reinterpret or static_cast a void* pointer to any class other than the original. Several places in the code were going animation* -> void* -> link_list* which particularly when you add a virtual function in the mix, does not have the same memory layout.

Alberth289346:
Added a commit that undoes the apt version increment.

Issue #2809, opened 2025-03-31
[META] Operating Theatre. Crashes and surgeons blocked overview.


ARGAMX:
I have spent some time investigating how a surgical operation is organized and how the Operating Theatre room works in CTH.

It is obvious that there are still several significant problems in Operating Theatre flow.

They are:

1. You can grab the Doctor and place him outside the Operating Theatre in a surgical gown or inside without it. This creates various unpredictable situations. Usually a game crashes. 
#1073, 
#1249, 
#1933, 
#2664 
![image](https://user-images.githubusercontent.com/20030128/173241532-11d4bfc2-1fb7-42b1-bdc4-ccf5c99bd369.png)

2. You can have more than 2 Surgeons in the same Operating Theatre. Usually after one or both grabbed or sacked. This creates various unpredictable situations. Usually just visual glitches but can crash on assert. 
#1162 ?, 
#1244, 
#2664 <img width="469" alt="Image" src="https://github.com/user-attachments/assets/12d396ab-6e55-4f74-b924-31888de10963" />

3. You can block one or both Surgeons in the Operating Theatre so that they could not be grabbed. The Operating Theatre no longer can function and could not be removed. The room is lost in fact. 
#866, 
#2029 <img width="737" alt="Image" src="https://github.com/user-attachments/assets/9b013e62-4e45-4241-87ee-b9df1cf6d641" />

..

..

..
Reasons why these problems are possible and occurs:

**A**. In CTH `MultiUseObjectAction` developed as an "uninterruptible state". However, the safe flow of this approach for the Operating Theatre has not been implemented. For `Surgeon1` it uses `MultiUseObjectAction`, For `Surgeon2` it uses `UseObjectAction`, which is interruptible. And:

**B**. Surgeons during Surgery process goes through an uninterruptible state, but this state is not correctly synchronized with Patient. (that's why surgeons can get blocked in Operating Theatre)

**C**. `Surgeon2` (who stands back to camera and does surgery manipulations) loses his uninterruptible state near the operating table and can be grabbed during the operation (while he is in  `UseObjectAction`). (why is this even possible? although maybe this is just a bug?)

**D**. When you remove a `Surgeon2` from the Operating Theatre during the operation, the necessary callbacks that could rollback the state to any consistent situation are not executed. (why was this allowed at all then?..)

[...]

Issue #1933, opened 2021-06-13
No animation for Surgeon using sofa facing south phase -5


Goldberg-Alex:
[gamelog.txt](https://github.com/CorsixTH/CorsixTH/files/6644435/gamelog.txt)]

```bash
Created game with savegame version 138.
Error in timer handler: 
Lua\humanoid_actions\use_object.lua:166: No animation for Surgeon using sofa facing south phase -5
stack traceback:
	[C]: in function 'error'
	Lua\humanoid_actions\use_object.lua:166: in upvalue 'action_use_phase'
	Lua\humanoid_actions\use_object.lua:435: in field '?'
	Lua\entities\humanoid.lua:536: in upvalue 'Humanoid_startAction'
	Lua\entities\humanoid.lua:636: in method 'finishAction'
	Lua\humanoid_actions\walk.lua:187: in local 'timer_function'
	Lua\entity.lua:216: in field 'tick'
	Lua\entities\humanoids\staff.lua:105: in field 'tick'
	Lua\entities\humanoids\staff\doctor.lua:93: in method 'tick'
	Lua\world.lua:1042: in method 'onTick'
	Lua\app.lua:1111: in function <Lua\app.lua:1108>
	(...tail calls...)
	Lua\app.lua:1012: in function <Lua\app.lua:1007>

```

lewri:
Thanks for the report. Are you able to provide us some info about what version of the game you were using and if possible the (auto)save before it happened?

Goldberg-Alex:
Hi!
I was using the GOG version. Sorry, I dont have the autosave.
I just noticed the log while trying to find the crash log of another issue, but there was only this, and this file was created a day before the crash.

TheCycoONE:
Spiffy, thank you for taking the time to file.

lewri:
CorsixTH version is 0.64 going by the savegame number.
(un)hopefully it might crop up again.

lewri:
I'm assuming somehow the Surgeon managed to leave gowned from the operating theatre and got to the staff room?

lewri:
I have managed to reproduce I believe....

1. Load save `...-3` from #2029 
2. Place a surgeon into the Operating Theatre from GP office south.
3. Wait for patient to enter, and attempt to pick up while he walks to the sink (it won't happen, but likes to queue "pickup" a lot in the action queue)
4. Allow surgeon to continue operation, eventually he will get picked up at the end of the surgery.
[...]

Issue #1249, opened 2017-09-19
Crash:  CorsixTH\Lua\humanoid_actions\walk.lua:304: Humanoid (Surgeon)


KeeperSt:
#### Describe the issue
Welcome to CorsixTH v0.60!

---------------------------------------------------------------

This window will display useful information if an error occurs.

---------------------------------------------------------------

Checking for CorsixTH updates...
You are running the latest version of CorsixTH.
An error has occurred!
Almost anything can be the cause, but the detailed information below can help the developers find the source of the error.
Running: The timer handler.
A stack trace is included below, and the handler has been disconnected.
...hivos de programa\CorsixTH\Lua\humanoid_actions\walk.lua:304: Humanoid (Surgeon) without door animations trying to walk through door (from operating_theatre to corridor)
stack traceback:
        [C]: in function 'error'
        ...hivos de programa\CorsixTH\Lua\humanoid_actions\walk.lua:304: in function <...hivos de programa\CorsixTH\Lua\humanoid_actions\walk.lua:210>
        (...tail calls...)
        ...hivos de programa\CorsixTH\Lua\humanoid_actions\walk.lua:188: in local 'timer_function'
        F:\Archivos de programa\CorsixTH\Lua\entity.lua:219: in field 'tick'
        F:\Archivos de programa\CorsixTH\Lua\entities\staff.lua:149: in method 'tick'
        F:\Archivos de programa\CorsixTH\Lua\world.lua:1075: in method 'onTick'
        F:\Archivos de programa\CorsixTH\Lua\app.lua:1000: in function <F:\Archivos de programa\CorsixTH\Lua\app.lua:997>
        (...tail calls...)
        F:\Archivos de programa\CorsixTH\Lua\app.lua:901: in function <F:\Archivos de programa\CorsixTH\Lua\app.lua:896>

Warning: No event handler for timer

#### Steps to Reproduce
  We have two doctors operating in the surgeon.  We put another one over the bed.  Few seconds later, it crashes.

#### Save Game
https://www.dropbox.com/s/74hokadasyz3w4z/keeper5.sav?dl=0

#### System Information
**CorsixTH Version:** 0.60 

**Operating System:** Windows 10 64 bits

**Theme Hospital Version:** CD

### Additional Info
![med general17](https://user-images.githubusercontent.com/31771827/30616655-3d4dfe8e-9d93-11e7-857b-5238b0f73a04.jpg)

TheCycoONE:
[...]

PR #2810, opened 2025-04-01
On pickup surgeon remove surgical gown

ARGAMX:
*Expected that will be fixed and are no longer reproducible: #1249, #1933*

Part of the Operating Theatre META #2809 (task **x101**)

-
- On a pickup surgeon from the operating room, remove his surgical gown.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -456,6 +456,15 @@ function Staff:setPickup(ui, window_to_close)
   end
 end
 
+-- Function resets "Surgeon" state for Doctor.
+-- Useful for case when on picking up Doctor from Operating Theatre
+-- we need to switching him back to regular clothes and type.
+function Staff:resetSurgeonState()
+  if self.humanoid_class == "Surgeon" then
+    self:setType("Doctor")
+  end
+end
+
 function Staff:onPlaceInCorridor()
   local world = self.world
   local notify_object = world:getObjectToNotifyOfOccupants(self.tile_x, self.tile_y)
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -456,13 +456,12 @@ function Staff:setPickup(ui, window_to_close)
   end
 end
 
--- Function resets "Surgeon" state for Doctor.
--- Useful for case when on picking up Doctor from Operating Theatre
--- we need to switching him back to regular clothes and type.
-function Staff:resetSurgeonState()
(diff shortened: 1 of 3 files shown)


lewri:
Do you want #1249 as being closed by this PR?

ARGAMX:
> Do you want #1249 as being closed by this PR?

Yes.

lewri (commented):
Where code specifically targets a class, better approach is to have it within that class itself i.e. `Doctor`. Though it might not be possible here...

TheCycoONE:
> Where code specifically targets a class, better approach is to have it within that class itself i.e. `Doctor`. Though it might not be possible here...

It's possible but it's not a clear win, they would need to check for doctor before calling or implement this as a stub in staff. A more generically named stub like onPickup with an override for Doctor is the best way if you wanted to go that route (similar to onPlaceInCorridor)

lewri:
> > Where code specifically targets a class, better approach is to have it within that class itself i.e. `Doctor`. Though it might not be possible here...
> 
> It's possible but it's not a clear win, they would need to check for doctor before calling or implement this as a stub in staff. A more generically named stub like onPickup with an override for Doctor is the best way if you wanted to go that route (similar to onPlaceInCorridor)
> 
> I would merge as is.

Yeah I think it's only the function name that irked me here more than anything else being wrong (specific instance inside a generic class). @ARGAMX would you take a suggestion to more generalise the function name so it more reflects a pickup event? Tbh if you did onPickup suggestion and left it completely in staff that'd be nice enough.

Alberth289346:
Code itself looks ok to me.

I believe this stuff should remain local to the operating theater room, but that is not even remotely feasible currently.

I agree with @TheCycoONE, `onPickup` would be somewhat nicer, but in `staff.lua` works too.

ARGAMX:
> Where code specifically targets a class, better approach is to have it within that class itself i.e. `Doctor`. Though it might not be possible here...

Agree, `staff.lua` is not best place for this since we have `doctor.lua`.
For some reason I was think that we don't have `doctor.lua`...

> A more generically named stub like onPickup with an override for Doctor is the best way if you wanted to go that route (similar to onPlaceInCorridor)

Thanks. Wise. I like it. 
`resetSurgeonState()` moved to `doctor.lua`.

Now `staff.lua` and `doctor.lua` have `onPickup()`.

Thanks guys. 👍

lewri (commented):
This is better. Pretty sure actions in flight should be fine (though who saves a game while having a staff member picked up?)

Alberth289346:
> This is better. Pretty sure actions in flight should be fine (though who saves a game while having a staff member picked up?)

Auto-save does.

PR #2812, opened 2025-04-03
Operation Theatre. Prevent pickup surgeon during surgery operation

### LINKED ISSUE #2664: [Bug] Surgeon can be grabbed during the operation
### Describe the issue

One of the surgeons can be grabbed during the operation. 
It gives possibility for various unexpected things.
One of them is Three surgeons perform an operation.

Screenshot 2024-10-11 at 21 03 51

### Steps to Reproduce

2. Try to grab secondary surgeon (surgeon1) who is stands front to the camera.
3. Observe that secondary surgeon is in uninterruptible state and can't be grabbed.
4. Dismiss that secondary surgeon.
5. Grab primary surgeon (surgeon2) who stand back to camera and do manipulations.
6. Put him in slack tongue clinic.
7. Observe that he was not in uninterruptible state. At the same time secondary surgeon is still near operation table.
8. Observe also that patient currently is trapped in hospital. 😈
9. Press 3 and observe that two doctors enter operating theatre and both of them take primary role.

### Expected Behaviour

Primary surgeon should not be grabbable during the operation.

### Save Game

[three_doctors_in_surgery_room.sav.zip](https://github.com/user-attachments/files/17346230/three_doctors_in_surgery_room.sav.zip)

### CorsixTH Version

0.68.0 Beta 1

### Operating System

MacOS Sonoma 14.5

ARGAMX:
*Fixes #2664*

Part of the Operating Theatre META #2809 (task **x102**)

-
- Prevent pickup surgeon during surgery operation.

### DIFF
--- a/CorsixTH/Lua/humanoid_action.lua
+++ b/CorsixTH/Lua/humanoid_action.lua
@@ -31,11 +31,12 @@ function HumanoidAction:HumanoidAction(name)
 
   self.name = name
   self.count = nil -- 'nil' means 'forever' (until finished), else the number to perform.
-  self.must_happen = false -- If true, action cannot be skipped.
+  self.must_happen = false -- If true, action cannot be skipped. still can be interrupted/cancelled by pick up.
   self.loop_callback = nil -- Periodic callback to check for termination conditions.
   self.after_use = nil -- Callback for performing updates afterwards.
   self.is_leaving = false -- Whether the humanoid is leaving.
   self.no_truncate = false -- If set, disable shortening the action.
+  self.uninterruptible = false -- If true, can not be interrupted/cancelled by pickup.
 end
 
 --! Set the number of times the action should happen.
@@ -55,6 +56,16 @@ function HumanoidAction:setMustHappen(must_happen)
   assert(type(must_happen) == "boolean", "Invalid value for parameters 'must_happen'")
 
   self.must_happen = must_happen
+  return self
+end
+
+--! Set the 'uninterruptable' flag (that is, action cannot be canceled or truncated by pickup).
+--!param uninterruptable (bool) Whether or not the action uninterruptable.
+--!return (action) Returning self, for daisy-chaining.
+function HumanoidAction:setUninterruptable(uninterruptable)
+  assert(type(uninterruptable) == "boolean", "Invalid value for parameters 'uninterruptable'")
+
+  self.uninterruptible = uninterruptable
   return self
 end
 
--- a/CorsixTH/Lua/humanoid_action.lua
+++ b/CorsixTH/Lua/humanoid_action.lua
@@ -31,7 +31,7 @@ function HumanoidAction:HumanoidAction(name)
 
   self.name = name
   self.count = nil -- 'nil' means 'forever' (until finished), else the number to perform.
-  self.must_happen = false -- If true, action cannot be skipped. still can be interrupted/cancelled by pick up.
+  self.must_happen = false -- If true, action cannot be skipped. Still can be interrupted/cancelled by pick up.
   self.loop_callback = nil -- Periodic callback to check for termination conditions.
   self.after_use = nil -- Callback for performing updates afterwards.
   self.is_leaving = false -- Whether the humanoid is leaving.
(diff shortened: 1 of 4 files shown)


lewri (dismissed):
Uninterruptible state needs some kind of feedback loop to the player; otherwise they will think it's a bug.

CorsixTH/Lua/humanoid_action.lua
```diff
@@ -31,11 +31,12 @@ function HumanoidAction:HumanoidAction(name)
 
   self.name = name
   self.count = nil -- 'nil' means 'forever' (until finished), else the number to perform.
-  self.must_happen = false -- If true, action cannot be skipped.
+  self.must_happen = false -- If true, action cannot be skipped. still can be interrupted/cancelled by pick up.
```
lewri, line 34: Extra sentence could be shortened to `Pickup overrides.` too; otherwise just needs a capital S.

CorsixTH/Lua/humanoid_action.lua
```diff
@@ -58,6 +59,16 @@ function HumanoidAction:setMustHappen(must_happen)
   return self
 end
 
+--! Set the 'uninterruptable' flag (that is, action cannot be canceled or truncated by pickup).
+--!param uninterruptable (bool) Whether or not the action uninterruptable.
+--!return (action) Returning self, for daisy-chaining.
+function HumanoidAction:setUninterruptable(uninterruptable)
+  assert(type(uninterruptable) == "boolean", "Invalid value for parameters 'uninterruptable'")
```
lewri, line 66: You only have one parameter :)

CorsixTH/Lua/rooms/operating_theatre.lua
```diff
@@ -128,8 +128,12 @@ function OperatingTheatreRoom._buildTableAction2(multi_use, operation_table_b)
...
   end
 
-  return UseObjectAction(operation_table_b):setLoopCallback(loop_callback_use_object)
-      :setAfterUse(after_use_use_object):setMustHappen(true):disableTruncate()
+  return UseObjectAction(operation_table_b)
+    :setLoopCallback(loop_callback_use_object)
+    :setAfterUse(after_use_use_object)
+    :setMustHappen(true)
+    :disableTruncate()
+    :setUninterruptable(true)
```
lewri, line 136: Replaced method was an acceptable way to write for us. However, this is all fine with me because of the amount 👍

lewri (commented):
Only nits from me, otherwise all fine.

CorsixTH/Lua/humanoid_action.lua
ARGAMX: > You only have one parameter :)

There are many such functions in this file.
This is a copy of other functions in the same file.
It is literally almost identical to them.
And they all have one parameter. 
Some of them is a Boolean one.
So I have no clue why it might be not perfect as other ones. 🙂

CorsixTH/Lua/humanoid_action.lua
ARGAMX: > Extra sentence could be shortened to `Pickup overrides.` too; otherwise just needs a capital S.

Capitalised 😎

CorsixTH/Lua/humanoid_action.lua
Alberth289346: The idea is to allow what you do below (lines 131-136).
You create an action with default settings and then adjust parameters by applying these simple methods. As they all return the action you can build a chain of such methods rather than having to repeat the action every time.

Eg
```
local a = Action():setLength(0):setHidden(false):setAuto(true)
```
rather than
```
local a = Action()
a:setLength(0)
a:setHidden(false)
a:setAuto(true)
```

This alternative works best if the class has many parameters, and you usually only need to change the default value of a few of the parameters

Here, actions have a lot of parameters. If you had to specify them in the constructor, you would need a line or more with `Action(true, 0, false, nil, ...)` values for all the arguments. Also, such a line does not tell you what is set with the `true` value, while above you have eg `setAuto(true)`, which is much more clear about what it does. (assuming 'auto' makes sense for the action).

PR #2813, opened 2025-04-05
Add support for AddressSanitizer, drop VLD

TheCycoONE:
VLD is no longer maintained so it's removed.

Added a convenience option for enabling AddressSanitizer: https://github.com/google/sanitizers/wiki/AddressSanitizer

The options provided here work in gcc and llvm. Address sanitizer is available in MSVC but may use a slightly different flag.

I have not enabled it in the windows dev preset because I cannot test it.


Alberth289346:
LGTM

PR #2814, opened 2025-04-05
Operation Theatre. Surgeons should not stuck 

### LINKED ISSUE #866: Stuck doctors in "unrecognized" operating theater
Greetings! I'm running CorsixTH Version 0.50 beta 2 for Windows and I encountered the following problem. I accepted an incoming emergency requiring surgery, but then none of the patients were moving in the queue, whether part of the emergency or not. I checked the operating theater, and both my doctors were "frozen": they wouldn't move, and while I could click on them and get the usual menu, I couldn't move them. I tried rearranging the room, but the cursor was stuck on Zzzz and they still didn't move. Plus, I was informed that I apparently didn't have an operating theater anymore...! The following screenshot shows both that notice and the room itself with its stuck doctors:

untitled

Here's a before save and an after save. I couldn't reproduce the bug, but if it helps, it is very possible I have moved a surgeon into the operating theater after accepting the emergency. I couldn't find another issue in github that presented all these, erm, symptoms...

Thank you!

ARGAMX:
*Fixes #866*

Part of the Operating Theatre META #2809 (task **x103**)

-
- Preventing sticking Surgeons if patient fed up

### DIFF
--- a/CorsixTH/Lua/rooms/operating_theatre.lua
+++ b/CorsixTH/Lua/rooms/operating_theatre.lua
@@ -237,10 +237,15 @@ function OperatingTheatreRoom:commandEnteringPatient(patient)
   local surgeon2 = next(self.staff_member_set, surgeon1)
   assert(surgeon1 and surgeon2, "Not enough staff in operating theatre")
 
-  -- Patient changes into surgical gown
   local screen, sx, sy = self.world:findObjectNear(patient, "surgeon_screen")
-  patient:walkTo(sx, sy)
-  patient:queueAction(UseScreenAction(screen))
+  -- Patient walk to surgeon screen
+  patient:queueAction(WalkAction(sx, sy)
+    :setMustHappen(true)
+    :disableTruncate())
+  -- Patient changes into surgical gown
+  patient:queueAction(UseScreenAction(screen)
+    :setMustHappen(true)
+    :disableTruncate())
 
   -- Meanwhile, surgeons wash their hands
   -- TODO: They sometimes overlap each other when doing that. Can we avoid that?


ARGAMX:
> I am not sure nailing every step to the floor is the answer. Sure, it works in some cases, but how to know it works in all cases?

Albert, you are 100% right.

But I have not found anything better.
We can rewrite everything regarding the logic of coordination. 
This will require more effort, time, and in the end we may still get new bugs. 
Or maybe we will not get new bugs.

For now in 0.68.0 it is terribly unreliable.
With this fix, although it is not ideal, I was no longer able to reproduce the bug.

As a QA, I am happy.
As a developer, I am not absolutely happy and I think exactly the same as you.

But for now I can not offer something better.
This not ideal fix will remove the problem for 0.69.0.

If someone can do better - I only welcome it.

Alberth289346:
Fair enough

lewri:
> As a developer, I am not happy enough and I think exactly the same as you.
> 
> But for now I can not offer something better. This not ideal fix will eliminate the problem in almost all cases for 0.69.0. (maybe even for everyone)

Feel free to add a TODO comment in the code, if you have a goal in mind on where to improve.

ARGAMX:
> Feel free to add a TODO comment in the code, if you have a goal in mind on where to improve.

Thank you, but I can't suggest anything specific and in a specific place yet.

One thing is obvious so far - that the actions of humanoids in the game are not properly synchronized and cannot be properly synchronized with each other yet. There are no appropriate mechanisms for this.

It's not just in the Operation Theatre. It's everywhere.

It's just that the Operation Theatre suffers from this the most.

lewri:
LGTM

PR #2818, opened 2025-04-07
Allow using the secondary marker

Alberth289346:
*Fixes animations with both a staff member and a patient, with different moods.

- Rename the first marker.
- Add a child animation that follows the second marker.
- Add a way to follow the second marker position of the parent animation.

Can be read by commit if so desired.


TheCycoONE:
Can clean up the old lua function names when they are migrated.

PR #2829, opened 2025-04-25
Boiler breakdown. prevent temp adjust during breakdown.

### LINKED ISSUE #2828: [Bug] During boiler breakdown user can adjust the temp
### Describe the issue

During boiler breakdown user can adjust the temperature.

But During boiler breakdown user should not be able to adjust the temperature as in vanilla TH.

This behaviour was implemented and working good in 0.60.0 and up to 0.64.0 before 0.65.0 where it was regressed.

Image

### Steps to Reproduce

'during' case:

1. *Load 'during' save game attached to this ticket*
2. This save made during boiler breakdown.
3. Open town map (F4).
4. Press "-" and observe how you can adjust temperature which should be not possible.

'before' case:

1. *Load 'before' save game attached to this ticket*
2. This save made before boiler breakdown.
3. Wait until boiler breakdown will happen on ~8 sept.
4. If it not happened back to 1.
5. Open town map (F4).
6. Press "-" or  "+" and observe how you can adjust temperature which should be not possible.

### Expected Behaviour

During boiler breakdown user should not be able to adjust the temperature as in vanilla TH.

### Save Game

[2828_boiler_brakedown.sav.zip](https://github.com/user-attachments/files/19849309/2828_boiler_brakedown.sav.zip)

### CorsixTH Version

0.68.0

### Operating System

MacOS Sonoma 14.5

ARGAMX:
*Fixes #2828*

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/town_map.lua
@@ -329,7 +329,7 @@ end
 function UITownMap:decreaseHeat()
   local h = self.ui.hospital
   local heat = math.floor(h.heating.radiator_heat * 10 + 0.5)
-  if not h.heating_broke then
+  if not h.heating.heating_broke then
     heat = math.max(heat - 1, 1)
     h.heating.radiator_heat = heat / 10
   end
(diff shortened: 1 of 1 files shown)


Alberth289346:
LGTM

PR #2831, opened 2025-04-25
Make TH's files check for corruption a option

Matroftt:
- Making it possible to not check files for corruption => no warning (attached), even if they are actually corrupted
- Option is accessible via customise

image


Alberth289346:
With respect to the failed build:  
If you look in the details of the failed 5.4 build (the `...` at the right, then select 'details'), you can see that the `WindowsInstaller/config_template.txt` needs to be adjusted as well.

I am not sure why you want this option. If the TH data files are not valid, the game will crash. Can you explain why you want this option please?

lewri (commented):
I agree this is useful. However, I'm not sure I agree it should be an option. 

Either it should be mandatory to run (a bit worse for ux if it runs every launch) or let it be a button "Check for TH corruption" (maybe game corruption, or Check game files integrity) in the options.

Does the corruption check still run when you select your TH install the first time? That's usually the ideal place to check and maybe the process is wrong there instead if errors aren't showing.

Alberth289346:
The error message states the problem is a `SM4` file. I am guessing that is a missing video file (given the reported size). Is that correct?

A missing file is detectable. Since video is optional, the program could avoid playing that video.

A corrupted file cannot be detected until it crashes the game, but that would be too late.

Matroftt:
> Does the corruption check still run when you select your TH install the first time? That's usually the ideal place to check and maybe the process is wrong there instead if errors aren't showing.

It does not. Upon selecting a folder with TH, game restarts immediately.

So we have two solutions for this? Rename and move it to the options page, or instead check it only when selecting TH folder?

TheCycoONE:
I can think of two legit use cases and one illegit:
Testing another version of the game e.g. some of the localized cd versions that we don't yet support.

Reverse engineering changes into the original files for fun and profit.

Running Warez copies of the game with corrupt files. These tend to crash and we do not want the bugs/do not support them. People tend to be less forthcoming and that's a reason we have this check.

What's your motivation @Matroftt

Matroftt:
> I can think of two legit use cases and one illegit: Testing another version of the game e.g. some of the localized cd versions that we don't yet support.
> 
> Reverse engineering changes into the original files for fun and profit.
> 
> Running Warez copies of the game with corrupt files. These tend to crash and we do not want the bugs/do not support them. People tend to be less forthcoming and that's a reason we have this check.
> 
> What's your motivation @Matroftt

I made this option for cut versions of the game, you can see that I am currently using it, since one of the intro files does not exist (this count as illegit I suppose?). But it will be also helpful for other two cases.

lewri:
I still think it should always run after setting the directory at least once imo. I agree having it run repeatedly is bad behaviour though.

As such, I still would edge towards preferring a check that is a 'check now' over a toggle.

Alberth289346:
I think the simpler option is to treat missing video files as missing rather than as being corrupted, and simply skip playing them.

lewri:
> I think the simpler option is to treat missing video files as missing rather than as being corrupted, and simply skip playing them.

This already happens; it's likely the wording might be better as `INTRO.SM4 is corrupt or missing: size Xkb`

Alberth289346:
> This already happens; it's likely the wording might be better as ...

And it shows the popup shown in the OP ?

I can see that being annoying.

Why not silently ignore the missing file if video is off, or add a log entry for it?

TheCycoONE:
Why an option, video files are optional and not present on a standard cd install. We can leave them out of the check if missing.

TheCycoONE:
I would like @Matroftt's use case to just work, no option.

TheCycoONE:
@Matroftt are you interested in continuing this? Otherwise I'll make the change I proposed to make the movies optional (accept standard install from CD)

TheCycoONE:
Testing the standard CD install - apparently it didn't install the level files either. Given that we probably don't care to support it?

Matroftt:
> @Matroftt are you interested in continuing this? Otherwise I'll make the change I proposed to make the movies optional (accept standard install from CD)

Not much. You can do it if you want

PR #2832, opened 2025-04-26
[WIP] Fix #2749: Standardize back button positions in UI windows

### LINKED ISSUE #2749: Have a consistent layout for buttons in the various windows

As an example, observe the position of the "back" button in some windows below:

image  
Campaign window, "back" is at the right.

image  
Custom Game, "back" is at the left.

image  
Load game, "back" is at the left.

image
Save game, "back" is at the left.

image  
Settings, "back" is at the center.

Having the same button at different places in different windows makes it more difficult to navigate, since the user has to carefully check where the desired button is.

I didn't check all buttons etc, other window widgets may also be inconsistent between different windows

Users will find it easier to navigate between different windows since the layout of them is more predictable.

Unsure

Something close to 101a4de3 (but window layout was not modified in it)

Perhaps

### Additional Information

_No response_

dennlin:
*Fixes #2749 *

-Standardizes the position of the back button in the Options (Settings) menu by moving it to the left side and widening it to match the width of other buttons
-Swaps the positions of Accept and Cancel buttons in the Hotkey Assignment dialog, moving Accept to the right and Cancel (renamed to "Back") to the left
-Adds new string keys for the renamed "Back" button

Note: These changes were tested with the demo version of the game, which limited access to some windows. Future work may be needed to apply similar changes to other windows.

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
+++ b/CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
@@ -487,21 +487,23 @@ function UIHotkeyAssign:UIHotkeyAssign(ui, mode)
 
   self:showKeyPane("global")
 
-  -- "Accept" button
+  -- "Cancel" button (rename to "Back" and move to left)
   self:addBevelPanel(self.panel_pos_table_x[1], 430, 200, 40, col_bg)
-      :setLabel(_S.hotkey_window.button_accept)
-      :makeButton(0, 0, 180, 40, nil, self.buttonAccept)
-      :setTooltip(_S.tooltip.hotkey_window.button_accept)
-  -- Reset to defaults button.
+      :setLabel(_S.hotkey_window.button_back)
+      :makeButton(0, 0, 180, 40, nil, self.buttonCancel)
+      :setTooltip(_S.tooltip.hotkey_window.button_back)
+
+  -- Reset to defaults button (keep in middle)
   self:addBevelPanel(self.panel_pos_table_x[2], 430, 200, 40, col_bg)
       :setLabel(_S.hotkey_window.button_defaults)
       :makeButton(0, 0, 180, 40, nil, self.buttonDefaults)
       :setTooltip(_S.tooltip.hotkey_window.button_defaults)
-  -- "Cancel" button
+
+  -- "Accept" button (move to right)
   self:addBevelPanel(self.panel_pos_table_x[3], 430, 200, 40, col_bg)
-      :setLabel(_S.hotkey_window.button_cancel)
-      :makeButton(0, 0, 180, 40, nil, self.buttonCancel)
-      :setTooltip(_S.tooltip.hotkey_window.button_cancel)
+      :setLabel(_S.hotkey_window.button_accept)
+      :makeButton(0, 0, 180, 40, nil, self.buttonAccept)
+      :setTooltip(_S.tooltip.hotkey_window.button_accept)
 
   self.built_in_font = built_in
 end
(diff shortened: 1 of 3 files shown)


Alberth289346:
Could you please change "Fixes #2749" in the first comment to "Addresses #2749" ?

The issue discusses UI layout in general, while this patch only addresses The 'Back' buttons. By doing the above, the issue stays open when this patch gets merged, for next patches addressing other UI elements.

Your choice to limit to back buttons only is good. Smaller patches are more easily reviewed.

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
```diff
@@ -487,21 +487,23 @@ function UIHotkeyAssign:UIHotkeyAssign(ui, mode)
...
 
-  -- "Accept" button
+  -- "Cancel" button (rename to "Back" and move to left)
   self:addBevelPanel(self.panel_pos_table_x[1], 430, 200, 40, col_bg)
-      :setLabel(_S.hotkey_window.button_accept)
-      :makeButton(0, 0, 180, 40, nil, self.buttonAccept)
-      :setTooltip(_S.tooltip.hotkey_window.button_accept)
-  -- Reset to defaults button.
+      :setLabel(_S.hotkey_window.button_back)
+      :makeButton(0, 0, 180, 40, nil, self.buttonCancel)
```
lewri, line 493: Suggest renaming the function too (otherwise you have `back` in some places and `cancel` in others).

CorsixTH/Lua/languages/english.lua
```diff
@@ -684,6 +684,7 @@ hotkey_window = {
   ingame_recallPosition_8 = "8",
   ingame_recallPosition_9 = "9",
   ingame_recallPosition_0 = "10",
+  back = "Back",
```
lewri, line 687: Line not needed here. I think you meant `options_window.back` but that already exists (line 473)

CorsixTH/Lua/languages/english.lua
```diff
@@ -700,6 +701,7 @@ tooltip.hotkey_window = {
   button_recallPosKeys = "Assign hotkeys to save and recall camera positions",
   panel_debugKeys = "Assign hotkeys for debugging",
   button_back_02 = "Go back to the main hotkey window. Hotkeys changed in this window can be accepted there",
+  button_back = "Return to the options screen without saving changes",
```
lewri, line 704: `hotkey_window.button_cancel` and `tooltip.hotkey_window.button_cancel` now become superfluous strings.
That's fine, but I would suggest checking the contents of the existing strings being removed, and see if they can be made to fit what already exists.

In my opinion here however, you can just delete them.

---

However, such a change does mean all other languages will lose this translation. I can rename the other translated languages' `hotkey_window.button_cancel` to `hotkey_window.button_back` so at least they have a translated string. We can then use our Discord to inform users of a string revision. **In short, you don't need to do this part of my comment yourself**

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
```diff
@@ -487,21 +487,23 @@ function UIHotkeyAssign:UIHotkeyAssign(ui, mode)
 
   self:showKeyPane("global")
 
-  -- "Accept" button
+  -- "Cancel" button (rename to "Back" and move to left)
```
lewri, line 490: You can just update the comment to `"Back" button` here. It'll all be understood 🙂

lewri (commented):
Thanks! Just some pointers for you

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
lewri: `UIHotkeyAssign:buttonCancel()` --> `UIHotkeyAssign:buttonBack()`

`:makeButton(0, 0, 180, 40, nil, self.buttonBack)

lewri:
@dennlin did you want to still continue with this one some time? Otherwise, we can work on finishing it off if needed for you from the current state.

PR #2834, opened 2025-04-27
Heatmaps should not applying to lakes

### LINKED ISSUE #1991: Heatmaps applying to lakes

ARGAMX:
*Fixes #1991*

PR with changed presented by Albert https://github.com/CorsixTH/CorsixTH/issues/1991#issuecomment-1221563728

[2834_Heatmaps_on_lakes_1.sav.zip](https://github.com/user-attachments/files/19929399/2834_Heatmaps_on_lakes_1.sav.zip)

[2834_Heatmaps_on_lakes_2.sav.zip](https://github.com/user-attachments/files/19929400/2834_Heatmaps_on_lakes_2.sav.zip)

v0.68.0 :

Screenshot 2025-04-27 at 14 55 11

this PR :

Screenshot 2025-04-27 at 14 56 11

### DIFF
--- a/CorsixTH/Src/th_lua_ui.cpp
+++ b/CorsixTH/Src/th_lua_ui.cpp
@@ -82,7 +82,7 @@ int l_town_map_draw(lua_State* L) {
           else
             goto dont_paint_tile;
         } else if (pNode->iParcelId == 0) {
-            goto dont_paint_tile;
+          goto dont_paint_tile;
         } else if (bShowHeat) {
           uint16_t iTemp = pMap->get_tile_temperature(pNode);
           if (iTemp < 5200)  // Less than 4 degrees
--- a/CorsixTH/Src/th_lua_ui.cpp
+++ b/CorsixTH/Src/th_lua_ui.cpp
@@ -81,6 +81,8 @@ int l_town_map_draw(lua_State* L) {
             iColour = iColourPurchasable;
           else
             goto dont_paint_tile;
+        } else if (pNode->iParcelId == 0) {
+            goto dont_paint_tile;
         } else if (bShowHeat) {
           uint16_t iTemp = pMap->get_tile_temperature(pNode);
           if (iTemp < 5200)  // Less than 4 degrees
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -485,7 +485,7 @@ bool level_map::load_from_th_file(const uint8_t* pData, size_t iDataLength,
 
       if (!(pData[5] & 1)) {
         pNode->flags.passable = true;
-        if (!(pData[7] & 16)) {
+        if (!(pData[7] & 16) && pNode->iParcelId != 0) {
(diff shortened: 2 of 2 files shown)


CorsixTH/Src/th_lua_ui.cpp
```diff
@@ -81,6 +81,8 @@ int l_town_map_draw(lua_State* L) {
             iColour = iColourPurchasable;
           else
             goto dont_paint_tile;
+        } else if (pNode->iParcelId == 0) {
+            goto dont_paint_tile;
```
lewri, line 85: A very counter-intuitive indent error imo from the logs; however, probably best to fix it.

@Alberth289346 this means the indents here have the same alignment as the nested if above. I don't like that personally. Could we put an empty line in instead to separate them above 84 to try and make it nicer? I'm not going to suggest Argus does a full refactor for such a small toe dip into our c++

lewri (commented):
Could you ping a screenshot into the PR please jut to save having to do a rebuild to verify?

CorsixTH/Src/th_lua_ui.cpp
ARGAMX: There was stuffed 2 extra spaces in here. 
That's why it issued a warning.
I removed them.
I think this will solve the problem.

ARGAMX:
> Could you ping a screenshot into the PR please jut to save having to do a rebuild to verify?

I have attached corresponding screenshots and savegames to the PR and issue 👌

TheCycoONE:
I suspect it would break if the lake was assigned a parcel number which is a perfectly valid thing for a map maker to do.

CorsixTH/Src/th_map.cpp
```diff
@@ -485,7 +485,7 @@ bool level_map::load_from_th_file(const uint8_t* pData, size_t iDataLength,
 
       if (!(pData[5] & 1)) {
         pNode->flags.passable = true;
-        if (!(pData[7] & 16)) {
+        if (!(pData[7] & 16) && pNode->iParcelId != 0) {
```
lewri, line 488: Albert also states in https://github.com/CorsixTH/CorsixTH/issues/1991#issuecomment-1221563728 it ideally needs an afterload to handle this change. What that would look like I'm not sure... There's probably a method to request a refresh of map data from lua.

lewri: Was thinking of `Map:setCellFlags` but generally this requires each tile to be given unless there's a way to efficiently pull all pond tiles from a map and then apply.

Alberth289346: Iterating over them and selection pond tiles is the best you can do here. Even in CPP that's what you need to do.

ARGAMX:
closed in preference to #2837

PR #2835, opened 2025-04-28
Fix animation refactor regression

### LINKED ISSUE #2826: [Bug] Some depersisted Objects/Animations are not fully rendered
### Describe the issue

After loading the savegame some objects are visually cutted.
Animation in cutted side is also invisible.
The save version does not matter.
The effect is reproducible even if you create a new game, make a save and load it.

Was brought in #2808

Image

### Steps to Reproduce

1. Start a new level/campaign.
2. Build a Staff room. Place the sofa facing southwest. 
3. Build General Diagnosis room.
4. Save game.
5. Load just created a savegame.
6. Observe that the sofa and crash trolley are visually cutted.

### Expected Behaviour

As was in 0.68.0 objects should not be cutted.

### Save Game

[sofa_cutted.sav.zip](https://github.com/user-attachments/files/19776844/sofa_cutted.sav.zip)

[issue_2826.sav.zip](https://github.com/user-attachments/files/19923054/issue_2826.sav.zip)

### CorsixTH Version

5ff228a trunk master

### Operating System

MacOS Sonoma 14.5

TheCycoONE:
luaT_testuserdata only works correctly if it is passed the correct
metatable reference, and the implementation we had assumed that was the
current environment table, but in the case of map loading it wasn't.

I wrote a replacement method for determining the class of the
animation_base userdata that depends on the __class_name stored in the
metatable.

There was an additional problem with persistence breaking any new saves:
it was using the void* of link_list rather than the class itself.
link_list was not polymorphic so I added the virtual destructor to fix
that and give it a vtable, then dynamic_cast to ensure the proper
pointer type was persisted.

*Fixes #2826*

### DIFF
--- a/CorsixTH/Src/persist_lua.cpp
+++ b/CorsixTH/Src/persist_lua.cpp
@@ -30,7 +30,6 @@ SOFTWARE.
 #include <cstdlib>
 #include <cstring>
 #include <string>
-#include <vector>
 
 #include "th_lua.h"
 #ifdef _MSC_VER
--- a/CorsixTH/Src/th.h
+++ b/CorsixTH/Src/th.h
@@ -30,7 +30,7 @@ SOFTWARE.
 class link_list {
  public:
   link_list();
-  ~link_list();
+  virtual ~link_list();
 
   link_list* prev;
   link_list* next;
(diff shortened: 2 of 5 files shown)


TheCycoONE:
Apparently in lua 5.1/luajit getfield doesn't return the type. That's slightly annoying but easy to workaround.
This fix points at a potential to use luaT_testuserdata if the right metatable index can be found.

The more alarming thing is that it does not fix new saves, so I'm still missing part of the problem.

ARGAMX (approved):
Works great for me 🔥

TheCycoONE, Great job 👍

PR #2837, opened 2025-04-29
[RFC] Set Original Campaign's Outdoor Flags

lewri:
*Fixes most of #1991 (see below)*

- For the original campaign, every map is checked for bad flags and removed where they shouldn't be (no hospital tiles outside buildings; and no passable tiles outside of regular paths/roads.
- This check runs every time a plot is bought, as flags can be changed depending what tile types were in a parcel
- Verified that valid paths still exist from edge of map to hospital in each level.
- Removes ponds from the town map*
- Fixes a bug in Level 4 where the pond could have objects placed on it 
image

*Except one instance in Level 4, where the pond is visible until bought, then it disappears. I'm not sure what has made this one an exception but likely because it's defined as purchasable.  On a custom map, the ponds inside a parcel don't show up.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local SDL = require("sdl")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 212 -- 'coverup_in_progress' renamed to 'coverup_selected'
+local SAVEGAME_VERSION = 213 -- Fix original campaign maps
 
 class "App"
 
--- a/CorsixTH/Lua/map.lua
+++ b/CorsixTH/Lua/map.lua
@@ -270,12 +270,7 @@ function Map:load(level, difficulty, level_name, map_file, level_intro, map_edit
     end
   end
 
-  -- fix original level 6 map
-  if level == 6 and not map_file and not map_editor then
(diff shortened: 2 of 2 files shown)


TheCycoONE:
Can you do something like the ch6 fix to target the explicit tiles in 4?

lewri:
> Can you do something like the ch6 fix to target the explicit tiles in 4?

Yes. I would like to, but my knowledge runs a bit dry here @TheCycoONE so will need help. My current workaround is to call `fixTiles` after every plot purchase.
There's two paths of a more robust solution here I imagine.
1. Work out where the original maps store the flags of the pond tile not yet purchased, and override them too in `fixTiles`. Before you purchase that parcel in Level 4, the hospital flag (and others) is false; but turns true on the pond once it's bought.
2. (outside my scope) Ignore all flags in the original maps and set them ourselves in C++ as if it was a custom map. That would be probably a more ideal solution, but you'd still need an afterload to fix existing saves (or have the C++ do something to re-read the map on load).

lewri:
Any input regarding the above message @TheCycoONE ?

TheCycoONE:
No, I haven't looked at the problem closely. Fixing on every purchase is probably fine.

lewri:
Okay, ready then.

lewri:
Rebased now, if to be merged

ARGAMX (approved):

PR #2839, opened 2025-04-30
Fix pick up surgeon skips hand washing

### LINKED ISSUE #2006: [Bug] Attempt to pick up surgeon skips hand washing
#### Describe the issue
If you attempt to pick up a surgeon who is heading to the wash basin in op_theatre, they will skip the basin and go straight to the table.

#### Steps to Reproduce
1. Have a patient enter the operating theatre

2. Wait for surgeons to start heading to wash basin

3. Attempt to pick one up.

#### Expected Behavior
Allow surgeon to be picked up, or don't skip any actions.

#### System Information
**CorsixTH Version:** 0.65

#### Gamelog.txt

```
Paste gamelog.txt output here
```

### Additional Info
illustrate the problem.*

### LINKED ISSUE #2809: [META] Operating Theatre. Crashes and surgeons blocked overview.

ARGAMX:
*Fixes #2006*
*Fixes #2809*

### DIFF
--- a/CorsixTH/Lua/rooms/operating_theatre.lua
+++ b/CorsixTH/Lua/rooms/operating_theatre.lua
@@ -183,17 +183,25 @@ end
 
 --! Sends the surgeon to the nearest operation sink ("op_sink1")
 --! and makes him wash his hands
---!param at_front (boolean): If true, add the actions at the front the action queue.
+--!param before_other_actions (boolean): If true, add the actions at the front the action queue.
 --! Add the actions at the end of the queue otherwise.
 --! Default value is true.
-function OperatingTheatreRoom:queueWashHands(surgeon, at_front)
+function OperatingTheatreRoom:queueWashHands(surgeon, before_other_actions)
   local sink, sink_x, sink_y = self.world:findObjectNear(surgeon, "op_sink1")
-  local walk = WalkAction(sink_x, sink_y):setMustHappen(true):disableTruncate()
+  local walk = WalkAction(sink_x, sink_y)
+    :setMustHappen(true)
+    :disableTruncate()
   local wait = wait_for_object(surgeon, sink, true)
-  local wash = UseObjectAction(sink):setMustHappen(true)
+    :setMustHappen(true)
+    :disableTruncate()
+    :setUninterruptable(true)
+  local wash = UseObjectAction(sink)
+    :setMustHappen(true)
+    :disableTruncate()
+    :setUninterruptable(true)
 
   for pos, action in pairs({walk, wait, wash}) do
-    if (at_front) then
+    if (before_other_actions) then
       surgeon:queueAction(action, pos)
     else
       surgeon:queueAction(action)


lewri (commented):
Asked for @mugmuggy if he's around to check wrt the truncate additions. 
If he's not, I can only guess it's probably fine.

ARGAMX:
> Asked for @mugmuggy if he's around to check wrt the truncate additions. If he's not, I can only guess it's probably fine.

This happen because of `Truncate`.
So the `disableTruncate` do the main job here.

I honestly don't understand why this param (`disableTruncate`) isn't added for most actions.
I can only assume that the operating theatre logic was created before the interrupt and truncation logic appeared.

Although honestly, I want to rewrite it all.
But this is such a big and problematic job that I don't even want to start.

So `disableTruncate` for today is enough.
`setUninterruptable` for addition insurance that grab will not interrupt or truncation this.

lewri (approved):
Can merge

PR #2841, opened 2025-05-01
Afterload minor fix in earthquake.lua

ARGAMX:
*Fixes error on load some old saved games. Example save game attached to issue 1249.*

Screenshot 2025-05-01 at 22 28 56

- Made proper check `if old_quake then` before accessing `old_quake.active` in `afterLoad` which was missed.


tobylane (approved):
Yes, but this only helps save games far older than we usually support.

lewri:
> Yes, but this only helps save games far older than we usually support.

@ARGAMX is pushing some fixes to get old bugged saves to load. As long as I tag them with `build/repo` I know not to put them in the patch notes 😅

PR #2844, opened 2025-05-03
General Diagnosis. Crash fixes on doctor change

### LINKED ISSUE #1328: [Bug] General Diagnosis. Moving doctor during room use led to broken room and patient
#### Describe the issue
I picked up a doctor in general diagnosis while a patient was about to change back into their normal clothes. The patient instead left the room (now dressed normally) and headed to their next room psychiatry. However, the general diagnosis room was not updated, so the patients clothes were also occupying the screen. Once the patient went to their next room, they attempted to change in that room, and froze up. This left the general diagnosis room unusable as it was partially used, and the psychiatry room unused, as there was a patient using it who would not progress further.

Video of the bug occuring:

https://www.twitch.tv/videos/217623614

#### Steps to Reproduce
1. Right click a doctor in a general diagnosis room to pick him up.
2. Move him out of the room as the patient changes back.

#### Save Game

[1328_2 Moving doctor during room use led to broken room and patient.sav.zip](https://github.com/user-attachments/files/19947317/1328_2.Moving.doctor.during.room.use.led.to.broken.room.and.patient.sav.zip)

#### Expected Behavior
Either:

The patient continues to change back and leaves the room with the diagnosis complete

or:

The patient changes back and leaves the room with the diagnosis progress lost.

After this, the patient resumes as normal, and the room is available to use again.

#### System Information
**CorsixTH Version:** *0.61*

#### Gamelog.txt
```
Created game with savegame version 122.
Error in timer handler: 
C:\Program Files\CorsixTH\Lua\rooms\general_diag.lua:59: attempt to index a nil value (local 'staff')
stack traceback:
	C:\Program Files\CorsixTH\Lua\rooms\general_diag.lua:59: in local 'after_use'
	...ogram Files\CorsixTH\Lua\humanoid_actions\use_screen.lua:46: in function <...ogram Files\CorsixTH\Lua\humanoid_actions\use_screen.lua:38>
	(...tail calls...)
	C:\Program Files\CorsixTH\Lua\entity.lua:216: in method 'tick'
	C:\Program Files\CorsixTH\Lua\world.lua:1051: in method 'onTick'
	C:\Program Files\CorsixTH\Lua\app.lua:1008: in function <C:\Program Files\CorsixTH\Lua\app.lua:1005>
	(...tail calls...)
	C:\Program Files\CorsixTH\Lua\app.lua:909: in function <C:\Program Files\CorsixTH\Lua\app.lua:904>

```

### Additional Info
illustrate the problem.*

ARGAMX:
*Fixes #1328*

- Fix. In `general_diag.lua`. Reworked `GeneralDiagRoom:commandEnteringPatient`. The sequence of events in the room has been made more sustainable to attempts to grab the doctor at inappropriate moments or to kick the patient out at inappropriate moments. Added comment to actions.
- Fix. In `humanoid.lua`. Improvements for `uninterruptible` actions logic. (was necessary to this General Diagnosis fix)
- Fix. In `room.lua`. Previously In `onHumanoidEnter` new staff was only able to replace old one only if `not old_staff_in_room.dealing_with_patient`. But `dealing_with_patient` was not really indicating dealing with patient, in fact only indicating that patient in the room. So I changed it to `not old_staff_in_room.dealing_with_patient and staff_in_room:isMeandering()`. Because while the patient is walking towards the screen `dealing_with_patient` is already true, but doctor is not occupied by the patient and can be grabbed. 
- Refactor. In `room.lua`. In `onHumanoidEnter` `humanoid` (as staff) renamed to `staff_entered`, `staff_member` renamed to `staff_in_room`, `humanoid` (as patient) renamed to `patient_entered`.

### DIFF
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -491,7 +491,8 @@ function Humanoid:startAction()
   -- Call the action start handler
   TheApp.humanoid_actions[action.name](action, self)
 
-  if action == self.action_queue[1] and action.todo_interrupt then
+  if action == self.action_queue[1] and action.todo_interrupt and not action.uninterruptible then
+    -- Interrupt an action because it has been marked as to be interrupted
     local high_priority = action.todo_interrupt == "high"
     action.todo_interrupt = nil
     local on_interrupt = action.on_interrupt
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -491,8 +491,9 @@ function Humanoid:startAction()
   -- Call the action start handler
   TheApp.humanoid_actions[action.name](action, self)
 
+  -- If action has been marked as to be interrupted
   if action == self.action_queue[1] and action.todo_interrupt and not action.uninterruptible then
-    -- Interrupt an action because it has been marked as to be interrupted
+    -- Interrupt this action
     local high_priority = action.todo_interrupt == "high"
     action.todo_interrupt = nil
     local on_interrupt = action.on_interrupt
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -510,9 +510,13 @@ function Staff:adviseWrongPersonForThisRoom()
 end
 
 --! Check whether staff are meandering
+--"meander" action always insert "move" or "idle" action before itself.
+--so when humanoid "meandering" his action queue usually looks like:
+--[1 idle, 2 meander] or [1 walk, 2 meander].
 --!return true if staff currently has a meander action
 function Staff:isMeandering()
-  if (self.action_queue[1].name == "meander") or
+  if ((self.action_queue[1].name == "idle") or
+    (self.action_queue[1].name == "walk")) and
     (self.action_queue[2].name == "meander") then
     return true
   end
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -509,6 +509,16 @@ function Staff:adviseWrongPersonForThisRoom()
   end
 end
 
+-- Function to decide if staff currently Meander
+function Staff:isMeandering()
+  for _, action in ipairs(self.action_queue) do
+    if action.name == "meander" then
+      return true
+    end
+  end
+  return false
+end
+
 -- Function to decide if staff currently has nothing to do and can be called to a room where they're needed
 function Staff:isIdle()
   -- Make sure we're not in an undesired state
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -509,12 +509,12 @@ function Staff:adviseWrongPersonForThisRoom()
   end
 end
 
--- Function to decide if staff currently Meander
+--! Check whether staff are meandering
+--!return true if staff currently has a meander action
 function Staff:isMeandering()
-  for _, action in ipairs(self.action_queue) do
-    if action.name == "meander" then
-      return true
-    end
+  if (self.action_queue[1].name == "meander") or 
+    (self.action_queue[2].name == "meander") then
+    return true
   end
   return false
 end
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -512,7 +512,7 @@ end
 --! Check whether staff are meandering
 --!return true if staff currently has a meander action
 function Staff:isMeandering()
-  if (self.action_queue[1].name == "meander") or 
+  if (self.action_queue[1].name == "meander") or
     (self.action_queue[2].name == "meander") then
     return true
   end
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -497,7 +497,7 @@ end
 
 --! When a valid member of staff enters the room this function is called.
 -- Can be extended in derived classes.
---!param staff (Staff) The staff in question
+--!param staff (object) The staff in question
 --!param already_initialized (bool) If true, this means that the staff has already got order
 -- what to do.
 function Room:commandEnteringStaff(staff, already_initialized)
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -497,7 +497,7 @@ end
 
 --! When a valid member of staff enters the room this function is called.
 -- Can be extended in derived classes.
---!param staff (object) The staff in question
+--!param staff (Staff) The staff in question
 --!param already_initialized (bool) If true, this means that the staff has already got order
 -- what to do.
 function Room:commandEnteringStaff(staff, already_initialized)
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -367,7 +367,7 @@ function Room:onHumanoidEnter(humanoid)
         if staff_in_room.humanoid_class == "Nurse" and self.room_info.id == "ward" then
           self.hospital:giveAdvice(nurse_desks)
         end
-        if (not staff_in_room.dealing_with_patient) or staff_in_room:isMeandering() then
+        if not staff_in_room.dealing_with_patient or staff_in_room:isMeandering() then
           -- Previous staff in the room not currently occupied by serving patient in room.
           -- Send out the previous staff and appoint new one.
           staff_in_room:setNextAction(self:createLeaveAction())
@@ -407,7 +407,7 @@ function Room:onHumanoidEnter(humanoid)
     -- been sent to an incorrect diagnosis room, they should leave and go
     -- back to the gp for redirection
     local patient_entered = humanoid
-    if (patient_entered.infected) and not patient_entered.diagnosed and
+    if patient_entered.infected and not patient_entered.diagnosed and
         not self:isDiagnosisRoomForPatient(patient_entered) then
       patient_entered:queueAction(self:createLeaveAction())
       patient_entered.needs_redirecting = true
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -356,42 +356,46 @@ function Room:onHumanoidEnter(humanoid)
   if class.is(humanoid, Staff) then
     -- If the room is already full of staff, or the staff member isn't relevant
     -- to the room, then make them leave. Otherwise, take control of them.
-    if not self:staffFitsInRoom(humanoid) then
-      if self:getStaffMember() and self:staffMeetsRoomRequirements(humanoid) then
-        local staff_member = self:getStaffMember()
-        self.humanoids[humanoid] = true
-          if staff_member.profile.is_researcher and self.room_info.id == "research" then
-            self.hospital:giveAdvice(researcher_desks)
-          end
-          if staff_member.humanoid_class == "Nurse" and self.room_info.id == "ward" then
-            self.hospital:giveAdvice(nurse_desks)
-          end
-        if not staff_member.dealing_with_patient then
-          staff_member:setNextAction(self:createLeaveAction())
-          staff_member:queueAction(MeanderAction())
-          self.staff_member = humanoid
-          humanoid:setCallCompleted()
-          self:commandEnteringStaff(humanoid)
+    local staff_entered = humanoid
+    if not self:staffFitsInRoom(staff_entered) then
+      if self:getStaffMember() and self:staffMeetsRoomRequirements(staff_entered) then
+        local staff_in_room = self:getStaffMember()
+        self.humanoids[staff_entered] = true
+        if staff_in_room.profile.is_researcher and self.room_info.id == "research" then
+          self.hospital:giveAdvice(researcher_desks)
+        end
+        if staff_in_room.humanoid_class == "Nurse" and self.room_info.id == "ward" then
+          self.hospital:giveAdvice(nurse_desks)
+        end
+        if (not staff_in_room.dealing_with_patient) or staff_in_room:isMeandering() then
+          -- Previous staff in the room not currently occupied by serving patient in room.
+          -- Send out the previous staff and appoint new one.
+          staff_in_room:setNextAction(self:createLeaveAction())
+          staff_in_room:queueAction(MeanderAction())
+          self.staff_member = staff_entered
+          staff_entered:setCallCompleted()
+          self:commandEnteringStaff(staff_entered)
         else
           if self.waiting_staff_member then
             self.waiting_staff_member.waiting_on_other_staff = nil
             self.waiting_staff_member:setNextAction(self:createLeaveAction())
             self.waiting_staff_member:queueAction(MeanderAction())
           end
-          self:createDealtWithPatientCallback(humanoid)
-          humanoid.waiting_on_other_staff = true
-          humanoid:setNextAction(MeanderAction())
+          self:createDealtWithPatientCallback(staff_entered)
+          staff_entered.waiting_on_other_staff = true
+          staff_entered:setNextAction(MeanderAction())
         end
       else
-        self.humanoids[humanoid] = true
-        humanoid:setNextAction(self:createLeaveAction())
-        humanoid:queueAction(MeanderAction())
-        humanoid:adviseWrongPersonForThisRoom()
+        -- Inappropriate staff for this room
+        self.humanoids[staff_entered] = true
+        staff_entered:setNextAction(self:createLeaveAction())
+        staff_entered:queueAction(MeanderAction())
+        staff_entered:adviseWrongPersonForThisRoom()
       end
     else
-      self.humanoids[humanoid] = true
-      humanoid:setCallCompleted()
-      self:commandEnteringStaff(humanoid)
+      self.humanoids[staff_entered] = true
+      staff_entered:setCallCompleted()
+      self:commandEnteringStaff(staff_entered)
     end
     self:tryAdvanceQueue()
     return
@@ -402,22 +406,23 @@ function Room:onHumanoidEnter(humanoid)
     -- An infect patient's disease may have changed so they might have
     -- been sent to an incorrect diagnosis room, they should leave and go
     -- back to the gp for redirection
-    if (humanoid.infected) and not humanoid.diagnosed and
-        not self:isDiagnosisRoomForPatient(humanoid) then
-      humanoid:queueAction(self:createLeaveAction())
-      humanoid.needs_redirecting = true
-      humanoid:queueAction(SeekRoomAction("gp"))
+    local patient_entered = humanoid
+    if (patient_entered.infected) and not patient_entered.diagnosed and
+        not self:isDiagnosisRoomForPatient(patient_entered) then
+      patient_entered:queueAction(self:createLeaveAction())
+      patient_entered.needs_redirecting = true
+      patient_entered:queueAction(SeekRoomAction("gp"))
       return
     end
     -- Check if the staff requirements are still fulfilled (the staff might have left / been picked up meanwhile)
     if self:testStaffCriteria(self:getRequiredStaffCriteria()) then
-      if self.staff_member  then
+      if self.staff_member then
         self:setStaffMembersAttribute("dealing_with_patient", true)
       end
-      self:commandEnteringPatient(humanoid)
+      self:commandEnteringPatient(patient_entered)
     else
-      humanoid:setNextAction(self:createLeaveAction())
-      humanoid:queueAction(self:createEnterAction(humanoid))
+      patient_entered:setNextAction(self:createLeaveAction())
+      patient_entered:queueAction(self:createEnterAction(patient_entered))
     end
   end
 end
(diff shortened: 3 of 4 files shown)


CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -509,6 +509,16 @@ function Staff:adviseWrongPersonForThisRoom()
   end
 end
 
+-- Function to decide if staff currently Meander
+function Staff:isMeandering()
+  for _, action in ipairs(self.action_queue) do
+    if action.name == "meander" then
+      return true
```
ARGAMX, line 516: Honestly seems iterating through whole `action_queue` don't looks like a good solution here. But I didn't found any other working way.

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
-end
+  -- Patient undress for procedure
+  local undress = UseScreenAction(screen):setAfterUse(after_use_screen1)
+  patient:queueAction(undress)
 
-function GeneralDiagRoom:onHumanoidLeave(humanoid)
-  if self.staff_member == humanoid then
-    self.staff_member = nil
-  end
-  Room.onHumanoidLeave(self, humanoid)
```
ARGAMX, line 95: This function just do the same as parent function do. So I don't see any meaning in this inheritor function.

CorsixTH/Lua/room.lua
```diff
@@ -356,42 +356,46 @@ function Room:onHumanoidEnter(humanoid)
...
+    if not self:staffFitsInRoom(staff_entered) then
+      if self:getStaffMember() and self:staffMeetsRoomRequirements(staff_entered) then
+        local staff_in_room = self:getStaffMember()
+        self.humanoids[staff_entered] = true
+        if staff_in_room.profile.is_researcher and self.room_info.id == "research" then
+          self.hospital:giveAdvice(researcher_desks)
+        end
+        if staff_in_room.humanoid_class == "Nurse" and self.room_info.id == "ward" then
+          self.hospital:giveAdvice(nurse_desks)
+        end
```
ARGAMX, line 369: No changes made here. Only indentation fixed.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -491,7 +491,8 @@ function Humanoid:startAction()
   -- Call the action start handler
   TheApp.humanoid_actions[action.name](action, self)
 
-  if action == self.action_queue[1] and action.todo_interrupt then
+  if action == self.action_queue[1] and action.todo_interrupt and not action.uninterruptible then
+    -- Interrupt an action because it has been marked as to be interrupted
```
Alberth289346, line 495:
```suggestion
    -- Interrupt the current action.
```
The first action in an action queue is currently being done, so it's "the" action rather than "an (any/some) action".

Inside the `if` the decision is already made, so why it's here isn't relevant anymore. Just state what is going to happen now.

The reason for not stating why it's here: All code is changing all the time, so eventually the "why" explanation becomes incomplete or even incorrect.
In other words, avoid duplicating information.

CorsixTH/Lua/room.lua
```diff
@@ -492,16 +497,16 @@ end
 
 --! When a valid member of staff enters the room this function is called.
 -- Can be extended in derived classes.
---!param humanoid The staff in question
---!param already_initialized If true, this means that the staff has already got order
+--!param staff (object) The staff in question
```
Alberth289346, line 500:
```suggestion
--!param staff (Staff) The staff in question
```
There is a `Staff` base class.

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
+    local staff = self:getStaffMember()
+    if staff then
+      local trolley, cx, cy = self.world:findObjectNear(patient, "crash_trolley")
+
+      -- Doctor walk to trolley
+      local ox, oy = trolley:getSecondaryUsageTile()
+      local staff_prepare = WalkAction(ox, oy)
+        :setMustHappen(true)
+        :disableTruncate()
+        :setUninterruptable(true)
```
Alberth289346, line 69: You may want to double-indent these kind of lines, They are continuations of line 66.

CorsixTH/Lua/room.lua
Alberth289346: You forgot about the added local rename at line 359 (and another one further down) and the added comments.

Have a look yourself. Open the "Changes" tab, and select "hide white space changes". You'll find almost all lines have changed.

Alberth289346: What you probably meant is "no semantical changes". That means "code behaves the same as before".

CorsixTH/Lua/rooms/general_diag.lua
Alberth289346: Euhm? The one in `room.lua` is about 80 lines long rather than 5 lines.
Line 576 to 656 in my copy of the room.lua file

Alberth289346: Oh, my bad. It does the same as you said. Sorry for the noise.

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
-    patient:walkTo(cx, cy, true)
-
-    local after_use_trolley = --[[persistable:general_diag_trolley_after_use]] function()
-      if #staff.action_queue == 1 then
+    local after_use_screen2 = function() end
+    local staff = self:getStaffMember()
+    if staff then
+      local trolley, cx, cy = self.world:findObjectNear(patient, "crash_trolley")
+
+      -- Doctor walk to trolley
```
lewri, line 64: `walks`

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
 
 function GeneralDiagRoom:commandEnteringPatient(patient)
   local screen, sx, sy = self.world:findObjectNear(patient, "screen")
+  -- Patient walk to screen
```
lewri, line 55: If you want to leave the singular word form in; then really a single line block comment would be nice here to denote they're action constructors.

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
+
+      -- Doctor walk to trolley
+      local ox, oy = trolley:getSecondaryUsageTile()
+      local staff_prepare = WalkAction(ox, oy)
+        :setMustHappen(true)
+        :disableTruncate()
+        :setUninterruptable(true)
+      staff:setNextAction(staff_prepare)
+
+      -- Doctor wait patient near trolley
```
lewri, line 72: `waits for`

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
+      local staff_idle = IdleAction()
+        :setMustHappen(true)
+        :disableTruncate()
+        :setUninterruptable(true)
+      staff:queueAction(staff_idle)
+
+      -- Patient walk to trolley
```
lewri, line 79: `walks`

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
-    end
-
-    patient:queueAction(MultiUseObjectAction(trolley, staff):setMustHappen(false)
-        :setProlongedUsage(false):setAfterUse(after_use_trolley))
-    patient:queueAction(WalkAction(sx, sy):setIsLeaving(true):setMustHappen(false):disableTruncate())
 
-    local after_use_screen2 = --[[persistable:general_diag_screen_after_use2]] function()
-      if #patient.action_queue == 1 then
-        self:dealtWithPatient(patient)
+      -- Patient use trolley
```
lewri, line 91: The doctor strictly speaking uses the trolley. It's just some silly quirk it gets applied to the patient.

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
+      local use_trolley = MultiUseObjectAction(trolley, staff)
+        :disableTruncate()
+        :setMustHappen(true)
+        :setProlongedUsage(false)
+        :setAfterUse(after_use_trolley)
+      patient:queueAction(use_trolley)
+
+      -- Patient walk back to screen
```
lewri, line 99: `walks`

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
       end
+    else
+      -- Patient didn't get served
+      self:makeHumanoidLeave(patient)
+      patient:queueAction(SeekRoomAction(self.room_info.id))
     end
 
-    patient:queueAction(UseScreenAction(screen):setIsLeaving(true):setMustHappen(true)
-        :setAfterUse(after_use_screen2))
+    -- Patient dress back to his clothes
```
lewri, line 117: `their` -- we have more than just male patients in the hospital :)

Also `dresses`

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
+      :setIsLeaving(true)
+      :setMustHappen(true)
+      :setAfterUse(after_use_screen2)
+    patient:queueAction(dress_back)
   end
 
-  patient:queueAction(UseScreenAction(screen):setAfterUse(after_use_screen1))
-  return Room.commandEnteringPatient(self, patient)
-end
+  -- Patient undress for procedure
```
lewri, line 125: `undresses`

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
+      staff:queueAction(staff_idle)
+
+      -- Patient walk to trolley
+      local go_to_trolley = WalkAction(cx, cy)
+        :setMustHappen(true)
+        :disableTruncate()
+      patient:queueAction(go_to_trolley)
+
+      local after_use_trolley = --[[persistable:general_diag_trolley_after_use]] function()
+        -- Doctor finish wait near trolley
```
lewri, line 86: ? I think for this one you could just say `-- Doctor and patient ready for begin procedure`

CorsixTH/Lua/room.lua
```diff
@@ -402,22 +406,23 @@ function Room:onHumanoidEnter(humanoid)
     -- An infect patient's disease may have changed so they might have
     -- been sent to an incorrect diagnosis room, they should leave and go
     -- back to the gp for redirection
-    if (humanoid.infected) and not humanoid.diagnosed and
-        not self:isDiagnosisRoomForPatient(humanoid) then
-      humanoid:queueAction(self:createLeaveAction())
-      humanoid.needs_redirecting = true
-      humanoid:queueAction(SeekRoomAction("gp"))
+    local patient_entered = humanoid
+    if (patient_entered.infected) and not patient_entered.diagnosed and
```
lewri, line 410: Not you, but those brackets can go.

lewri (commented):
Most of those grammar comments only apply if you don't do the constructor code comment instead. e.g. `--[[ Construct room usage cycle ]]`

CorsixTH/Lua/room.lua
ARGAMX: Actually, yes. 
I forgot about it at a certain point.

CorsixTH/Lua/rooms/general_diag.lua
lewri: Ahem... *ready to begin

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > The first action in an action queue is currently being done, so it's "the" action rather than "an (any/some) action".
> 
> Inside the `if` the decision is already made, so why it's here isn't relevant anymore. Just state what is going to happen now.
> 
> The reason for not stating why it's here: All code is changing all the time, so eventually the "why" explanation becomes incomplete or even incorrect. In other words, avoid duplicating information.

I didn't understand and didn't know that `startAction()` would delete actions out of the queue for some reason.
It's not obvious.
It was a bit of a surprise that `startAction()` does this.
This is exactly what I wanted to make more transparent in the code.

Perhaps I shouldn't have written this inside `if`, and it should be outside `if`.
And perhaps it should be written in the description of the function itself.

In general, the point is not only to explain what happens inside `if`.
But to explain why `startAction()` does this.

Perhaps this needs to be refactered. 
But it would be better to at least simply explain why this is so.

CorsixTH/Lua/rooms/general_diag.lua
ARGAMX: > ? I think for this one you could just say `-- Doctor and patient ready for begin procedure`

The thing is that this is fires after trolley used (this is why it is called `after_use_trolley`).

For some reason for General Diagnosis `MultiUseObjectAction` in patient action queue, not in staff queue. (why?...)
So to occupy Doctor he is getting `IdleAction()` at line 77.

And here is his `IdleAction()` must finish.
And this is what the commentary is trying to say...

CorsixTH/Lua/rooms/general_diag.lua
ARGAMX: > You may want to double-indent these kind of lines, They are continuations of line 66.

If it will not give me any error or warning, I would like.

CorsixTH/Lua/entities/humanoids/staff.lua
ARGAMX: > A quick look at the meander code indicates the meander might be second in the queue once it starts. You could run some observations; the first parameter in the for loop will be the index; so for testing you could do `for idx, action in ipairs(self.action_queue) do` then print the index of the meander.
> 
> If meander always has the same index (or two indexes at most) then you only ever need check those indexes.

Thank you, yes, it helped.
For some reason I checked it before PR (second in the queue) and it didn't work for me.
So I had to go through the whole queue.

But now I have rechecked again and everything worked.
Apparently I made a mistake when I checked the first time

CorsixTH/Lua/room.lua
```diff
@@ -356,42 +356,46 @@ function Room:onHumanoidEnter(humanoid)
...
+      if self:getStaffMember() and self:staffMeetsRoomRequirements(staff_entered) then
+        local staff_in_room = self:getStaffMember()
+        self.humanoids[staff_entered] = true
+        if staff_in_room.profile.is_researcher and self.room_info.id == "research" then
+          self.hospital:giveAdvice(researcher_desks)
+        end
+        if staff_in_room.humanoid_class == "Nurse" and self.room_info.id == "ward" then
+          self.hospital:giveAdvice(nurse_desks)
+        end
+        if (not staff_in_room.dealing_with_patient) or staff_in_room:isMeandering() then
```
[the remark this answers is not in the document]
ARGAMX: > I don't believe you really need the bracket; the not only tests against the first condition. But there's also no harm in it here.

I think you don't like brackets in `if`.

As for me, there is nothing really wrong with these brackets.

I had an experience with one language when the absence of brackets in the `if` condition was perceived by the compiler as an error.

That's why I am used to the fact that brackets in the `if` are never unnecessary.

CorsixTH/Lua/room.lua
```diff
@@ -356,42 +356,46 @@ function Room:onHumanoidEnter(humanoid)
...
-          self:commandEnteringStaff(humanoid)
+    local staff_entered = humanoid
+    if not self:staffFitsInRoom(staff_entered) then
+      if self:getStaffMember() and self:staffMeetsRoomRequirements(staff_entered) then
+        local staff_in_room = self:getStaffMember()
+        self.humanoids[staff_entered] = true
+        if staff_in_room.profile.is_researcher and self.room_info.id == "research" then
+          self.hospital:giveAdvice(researcher_desks)
+        end
+        if staff_in_room.humanoid_class == "Nurse" and self.room_info.id == "ward" then
```
ARGAMX, line 367: Does anybody feel that `staff_in_room` (alias for `room.staff_member`) is an error here?

I have a feeling that it must be `staff_entered`, not `staff_in_room`.

But I didn't dare to change it.

CorsixTH/Lua/room.lua
ARGAMX, line 369: Does anybody feel that staff_in_room (alias for room.staff_member) is an error here?

I have a feeling that it must be staff_entered, not staff_in_room.

But I didn't dare to change it.

CorsixTH/Lua/rooms/general_diag.lua
ARGAMX: > The doctor strictly speaking uses the trolley. It's just some silly quirk it gets applied to the patient.

For some strange reason for General Diagnosis `MultiUseObjectAction` in patient action queue, not in staff queue. (why?...)
So technically patient use trolley. It strange.

CorsixTH/Lua/rooms/general_diag.lua
ARGAMX: > The code is likely poorly named, but logically the comment I added is correct.
> 
> If not mine, the comment needs a rephrase a little to better convey the stage. The comment currently makes me wonder what exactly was he waiting for (I and you know from experience what, but to someone else it might not be crystal clear)? Inference of some kind about the patient in context would be nice.

`        -- Release doctor after procedure and finish his IdleAction() near trolley
`

CorsixTH/Lua/room.lua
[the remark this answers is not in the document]
lewri: It's not so much that I don't like them :p 
In the past I'd overbracket things then would get told to remove them.

I did say I was fine leaving this as though.

CorsixTH/Lua/room.lua
Alberth289346: It very likely makes no difference. The research room and the ward only have one type of staff members.

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
+
+      -- Doctor walks to trolley
+      local ox, oy = trolley:getSecondaryUsageTile()
+      local staff_prepare = WalkAction(ox, oy)
+          :setMustHappen(true)
+          :disableTruncate()
+          :setUninterruptable(true)
+      staff:setNextAction(staff_prepare)
+
+      -- Doctor waits patient near trolley
```
lewri, line 72: Sorry, I think I missed the `for` -- Doctor waits for patient near trolley

CorsixTH/Lua/rooms/general_diag.lua
```diff
@@ -52,47 +52,81 @@ end
...
+      staff:queueAction(staff_idle)
+
+      -- Patient walks to trolley
+      local go_to_trolley = WalkAction(cx, cy)
+          :setMustHappen(true)
+          :disableTruncate()
+      patient:queueAction(go_to_trolley)
+
+      local after_use_trolley = --[[persistable:general_diag_trolley_after_use]] function()
+        -- Release doctor after procedure and finish his IdleAction() near trolley
```
lewri, line 86: Ah.... I see now. Misleading order.
This is fine.

CorsixTH/Lua/rooms/general_diag.lua
lewri: Hmm..  I'd more argue the patient queues the action for the doctor to use the trolley but okay :)
If you're sticking with that then you'll need a `uses`

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -509,12 +509,12 @@ function Staff:adviseWrongPersonForThisRoom()
...
 
--- Function to decide if staff currently Meander
+--! Check whether staff are meandering
+--!return true if staff currently has a meander action
 function Staff:isMeandering()
-  for _, action in ipairs(self.action_queue) do
-    if action.name == "meander" then
-      return true
-    end
+  if (self.action_queue[1].name == "meander") or
```
TheCycoONE: I recall this was discussed in the room, but why does meander have to be in the first two actions to count?

Unintuitive enough to be commented on.

CorsixTH/Lua/entities/humanoid.lua
Alberth289346: Documenting is fine.

The point I am trying to make is that you your comments should always be about the code *below* it. If you want to say something about the code above, write a comment about that code at a line earlier in the file.

In that way, a reader doesn't have to look up and down the file, trying to find a comment that explains the line of code that gives trouble.

In your change, the "because it has been marked as to be interrupted" part is about the `if` statement above it. It's not about the code below (the code below just aborts blindly without caring why it must do that).  
It's fine that you want to state what the `if` is doing, but write that text above the `if` line.

Here, you could write above line 494 `-- Check whether the first action is interrupted.` (Instead if `first` you could also write `current`.)

Then at 495 you could write a short comment like `-- Abort the action.` or so.

CorsixTH/Lua/rooms/general_diag.lua
ARGAMX: > Hmm.. I'd more argue the patient queues the action for the doctor to use the trolley but okay :) If you're sticking with that then you'll need a `uses`

Ok
But nevertheless I don't know what to argue about here... it's not me saying this, it's the code saying this. This is strange for me.

Before I started refactoring General Diagnosis room, I thought the same as you.
In the Operating Theatre room it is the doctors who perform (hold in action queue) `MultiUseObjectAction` and `UseObjectAction`.

But after reading the code regarding the General Diagnosis, I can no longer stay this way.
Because this action not in Doctor queue.
If I remove `IdleAction()` from Doctor queue and replace it with `MeanderAction()` he will just start to walk across the room. I have recorded video for you:

[Screen Recording 2025-05-06 at 10.53.15.webm](https://github.com/user-attachments/assets/bb099877-ec73-46fe-ae1a-dda9dfbac05b)

From this video I see like that technically doctor don't really occupied by the trolley.
He connected with that action, yes.
But not occupied by it.
The patient is really occupied by this action and can't move.
At least from `GeneralDiagRoom:commandEnteringPatient` perspective.

So if you will write that the doctor uses a trolly, then your comment will contradict with this code:
`patient:queueAction(use_trolley)`

And in that case I will have the feeling what there is something wrong here.

I hope this comment will be more transparent:

`-- Patient uses trolley (trolly animation with patient and doctor)`

CorsixTH/Lua/entities/humanoids/staff.lua
ARGAMX: > I recall this was discussed in the room, but why does meander have to be in the first two actions to count?
> 
> Unintuitive enough to be commented on.

Reasonable, I agree.
I think meander can only be second action. Not first.
So I have change this condition again.

I have added a comment that describes how `action_queue` should looks to be a "meandering" case.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > Here, you could write above line 494 `-- Check whether the first action is interrupted.` (Instead if `first` you could also write `current`.)
> 
> Then at 495 you could write a short comment like `-- Abort the action.` or so.

OK, thank you.

PR #2847, opened 2025-05-05
Eliminate world.animation_manager

Alberth289346:
World supplies the length of an animation for some reason while the better way to do that is through `TheApp.animation_manager`.

- It goes there eventually anyway.
- World is about world and not about animations.
- No other uses of `world.animation_manager` exist.


Alberth289346:
Test saves did run

TheCycoONE:
Yeah this would be save safe, and fine.

PR #2849, opened 2025-05-05
Fix Linear Interpolation for markers

### LINKED ISSUE #2848: Linear-interpolation method for markers broken
When using linear interpolation methods to move markers with an entity they often jump and snap around.

This has likely been broken a long time and is due to executing a `math.floor` operation when calculation pixel position too early.

Example diff shows how to fix this in `Graphics:setMarkerRaw`
```diff
     if arg2 then
       -- Linear-interpolation positions
       local x1, y1 = positionToXy(arg1)
       local x2, y2 = positionToXy(arg2)
       for i = 0, anim_length - 1 do
-        local n = math.floor(i / (anim_length - 1))
-        self.anims[fn](self.anims, frame, (x2 - x1) * n + x1, (y2 - y1) * n + y1)
+        local n = (i / (anim_length - 1))
+        local pos_x = math.floor((x2 - x1) * n + x1)
+        local pos_y = math.floor((y2 - y1) * n + y1)
+        self.anims[fn](self.anims, frame, pos_x, pos_y)
         frame = self.anims:getNextFrame(frame)
       end
```

Patch already scheduled to come in soon.

lewri:
*Fixes #2848*

- Moves the `math.floor` operation until pixel positions are calculated
- Small refactor to accommodate repeated set of marker in frame
- Added a minimal assertion check with a hopefully helpful error message.

[screen-recording(10).webm](https://github.com/user-attachments/assets/e6c195a2-3351-4fd7-8266-435a77fba53c)

### DIFF
--- a/CorsixTH/Lua/graphics.lua
+++ b/CorsixTH/Lua/graphics.lua
@@ -695,6 +695,19 @@ function AnimationManager:setMarkerRaw(anim, fn, arg1, arg2, ...)
   local anim_length = self:getAnimLength(anim)
   local frame = self.anims:getFirstFrame(anim)
 
+  --! Set marker position for frame
+  --!param x1y1 (boolen) use values given by x1 and y1 only if true, otherwise use
+  -- linear interpolation between x1/y1 and x2/y2
+  --!param x1, y1, x2 (optional), y2 (optional) pixel positions
+  --!param n fraction between start and end frame of linear interpolation
+  local function setMarkerFramePosition(x1y1, x1, y1, x2, y2, n)
+    assert(type(x1) == "number" and type(y1) == "number",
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/graphics.lua
```diff
@@ -695,15 +695,24 @@ function AnimationManager:setMarkerRaw(anim, fn, arg1, arg2, ...)
   local anim_length = self:getAnimLength(anim)
   local frame = self.anims:getFirstFrame(anim)
 
+  -- Set marker for frame, x1y1 means position is defined in x1 and y1 already
```
Alberth289346, line 698: This doesn't make sense to me. Positions are defined in frames rather than combined x and y positions.

Imho, just make it a proper luadoc.

PR #2850, opened 2025-05-05
Extend animview

Alberth289346:
Adds the bottom row:

Screenshot from 2025-05-05 16-14-05

- You can enter a mood position, and press "This frame" to set the position.
Both pixel and tile positions work, except the latter may be less accurate.
- You can click the arrow buttons to move the mood 1 pixel in the indicated position.

Each time you set a mood, a text line is printed to `stdout` of the form `frame, {x, y, "px"},` and a comment what animation it is from.
```
2, {5, 20, "px"}, -- Anim 0
3, {15, 20, "px"}, -- Anim 0
```

### DIFF
--- a/AnimView/frmMain.cpp
+++ b/AnimView/frmMain.cpp
@@ -40,8 +40,6 @@ SOFTWARE.
 #include <wx/tokenzr.h>
 #include <wx/wfstream.h>
 
-#include <regex>
-
 #include "backdrop.h"
 
 BEGIN_EVENT_TABLE(frmMain, wxFrame)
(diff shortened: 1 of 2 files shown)


AnimView/frmMain.cpp
```diff
@@ -694,9 +731,68 @@ void frmMain::_onPanelClick(wxMouseEvent& evt) {
...
   if (m_bDrawMood) m_panFrame->Refresh(false);
 }
 
+void frmMain::_onMoodFrameApply(wxCommandEvent& evt) {
+  wxString wxText = m_txtMoodFramePos->GetLineText(0);
+  std::string text = wxText.ToStdString();
+
+  std::regex tilePosRegex(
+      "\\{ *([-0-9]+[.][0-9]*) *, *([-0-9]+[.][0-9]*) *\\}");
+  std::regex pixelPosRegex("\\{ *([-0-9]+) *, *([-0-9]+) *, *[\"]px[\"] *\\}");
```
TheCycoONE, line 746: `(-?\d+(?:\.\d+))?` maybe?

For a group containing a potentially negative number with an optional decimal followed by one or more digits?

TheCycoONE:
Looks ok, what happens with whole number tiles or is that an unexpected use case?

ARGAMX:
This window will be even bigger at height now? 
It already can't fit in height on 1280x800 resolution. 
Although this is not a big problem. 
But it seems to be slowly growing in height.

Screenshot 2025-05-05 at 20 23 21

TheCycoONE:
> This window will be even bigger at height now? It already can't fit in height on 1280x800 resolution. Although this is not a big problem. But it seems to be slowly growing in height.
> 

As a dev tool I'm not too concerned, as long as it works for the dev that needs it. In this case I assume @Alberth289346 added the extra functionality for himself.

Another ticket that splits this into tabs or collapsible sections or even adds a scrollbar would be welcome.

Alberth289346:
> Looks ok, what happens with whole number tiles or is that an unexpected use case?

You get entire tiles offset. It does happen with larger machines.

AnimView/frmMain.cpp
Alberth289346: What syntax is that?  I have been looking at the c++ standard, and they have https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04 

At least `(?:` isn't there.

Maybe it's simpler to write a character scanner. There is nothing complicated in there.

Alberth289346: https://en.cppreference.com/w/cpp/regex  is the regex cpp library

Alberth289346:
> This window will be even bigger at height now?

No, the left side where I added the row is still shorter than the right side.

AnimView/frmMain.cpp
TheCycoONE: The default c++ regex is ECMAScript - https://en.cppreference.com/w/cpp/regex/ecmascript check the section on sub-expressions. 

Specifically it's a subexpression that doesn't get captured as a match group.

Alberth289346: Ah ok. I know `(?:` from Python, but didn't find it in the Re documentation that I read. Should have picked a different one :)

Alberth289346: In the mean time, I switched to c character matching which does the job as well.

TheCycoONE:
I'm ok with merging.

PR #2852, opened 2025-05-07
Include what you use

TheCycoONE:
Applied IWYU recommendations (mostly).

I deviated on libraries, it always wanted me to include the inner headers for SDL, ffmpeg, etc. but the definitions were normally expected to be exported by a top level header. wxWidgets was particularly problematic since it wanted me to use the gtk specific includes instead of the generic ones.

I do not intend to add include-what-you-use to the CI/CD pipelines at this time but some pragmas were added to the code to make it report somewhat cleaner.

See: https://github.com/include-what-you-use/include-what-you-use

I am happy it sorted out some of the weird header dependencies around gfx that required including files in a specific order.

### DIFF
--- a/CorsixTH/Src/sdl_core.cpp
+++ b/CorsixTH/Src/sdl_core.cpp
@@ -20,12 +20,13 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
-#include "config.h"
+#include <SDL.h>
 
 #include <array>
 #include <cstdio>
 #include <cstring>
 
+#include "lua.hpp"
 #include "lua_sdl.h"
 #include "th_lua.h"
 
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -26,19 +26,23 @@ SOFTWARE.
 #ifdef CORSIX_TH_USE_FREETYPE2
 #include "th_gfx_font.h"
 #endif
+#include <SDL.h>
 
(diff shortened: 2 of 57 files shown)


Alberth289346 (approved):

Alberth289346:
Wow, so many includes !

Looks fine to me

TheCycoONE:
I'm not entirely happy with a few of the changes I made around SDL2 in particular. I'd like to look it over with fresh eyes before this gets merged.

CorsixTH/Src/sdl_core.cpp
```diff
@@ -20,12 +20,20 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
-#include "config.h"
+#include <SDL.h>
+#include <SDL_error.h>
```
TheCycoONE, line 24: SDL.h exports the rest of these (documented), drop the rest of them.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -26,19 +26,23 @@ SOFTWARE.
 #ifdef CORSIX_TH_USE_FREETYPE2
 #include "th_gfx_font.h"
 #endif
+#include <SDL.h>  // IWYU pragma: keep
```
TheCycoONE, line 29: The keep on SDL.h isn't particularly helpful here or the other spot I put it, drop them.

TheCycoONE:
I'm ok with merging this now. AnimView / wxwidgets could use some additional love (eliminate wx.h) but the gains are probably minimal.

PR #2854, opened 2025-05-08
Bump clang-format and clang-tidy for linux pipeline

TheCycoONE:
Switched to 20 of the clang-tools, the current latest.

It is fairly easy to acquire any version of these tools, either for debian based distros packaged by llvm, https://apt.llvm.org/ or static compiled 3rd party binaries at https://github.com/muttleyxd/clang-tools-static-binaries/releases - I will add these links to the wiki after this is merged.

For now I suppressed all rules that resulted in new warnings and only addressed the new errors. Future work should review the suppressed rules and fix any that we do want to enforce.


Alberth289346:
14 is the oldest in my distribution (Mint LTS, which is an Ubuntu derivative).

Using that does mean I'll be without `tidy` again in at most a few years.

TheCycoONE:
> 14 is the oldest in my distribution (Mint LTS, which is an Ubuntu derivative).
> 
> Using that does mean I'll be without `tidy` again in at most a few years.

I decided against 14; and went with 20. It is fairly easy to get and I will update the wiki to reflect that.

TheCycoONE:
Looks like the new clang-tidy found one more bug... yey

Alberth289346 (approved):

Issue #2794, opened 2025-03-11
Mood icons misalignment with humanoid


lewri:
In CorsixTH there are several states where a mood icon doesn't align properly. Reported issues below should be used for reference of the issue, Any new mood inconsistencies should be added here instead (old issues will be closed and merged in)

Patient mood icon misaligned in General Diagnosis ( #2018 )
![Image](https://github.com/user-attachments/assets/91dc21f1-e64a-4ef4-a0bc-99643476a493)

Mood icons do not transition smoothly through doors ( #1759 )
![Image](https://github.com/user-attachments/assets/1ffabf3a-05c1-4e12-8785-a90aa8049504)

Patient mood icon misaligned on Cast Remover ( #1758 )
![Image](https://github.com/user-attachments/assets/b9690818-6a6b-4ce1-918b-9cc9dd1df67e)

There had been initial chatter about finding a new method to target where the mood sprite should be drawn. Generally, once that exists then these mood issues should disappear.

Matroftt:
I believe some of these happen because mood displays over the object, in common cases they are patients/staff, but i suppose that on situation similar to #1758 next happens: patient walks into the machine, they become one whole object and now mood displays over the machine and not the patient. 

* #2660 this is also related to this issue

Alberth289346:
> I believe some of these happen because mood displays over the object

That's likely indeed.

Alberth289346:
I have dusted my java data-file loader code a bit, created sprite sheets of the VSPR file, and dumped the first frame of animations `34` and `184`, and linked the listed sprite numbers with the sprite sheets to get  a description.

That resulted in
```
Animation 34, Doctor walk east.
Frame 0
  if layers[0] == 0 then draw sprite 2503 at (-30, -30) // Shadow ?
  if layers[0] == 0 then draw sprite 162 at (-30, -79)  // Body + legs
  if layers[5] == 2 then draw sprite 285 at (-23, -79)  // ** Head **
  if layers[5] == 4 then draw sprite 281 at (-23, -79)  // ** Head **
  if layers[5] == 4 then draw sprite 977 at (-18, -48)  // Hand
  if layers[5] == 6 then draw sprite 283 at (-24, -80)  // ** Head **
  if layers[5] == 8 then draw sprite 4106 at (-24, -80) // ** Head **
  if layers[5] == 8 then draw sprite 977 at (-18, -48)  // Hand
```
and
```
Animation 184, walk through door.
Frame 0
[...]

PR #2855, opened 2025-05-08
Add moods for crash trolley animations.

Alberth289346:
Addresses #2794 

- Add mood markers to crash trolley animations.

It looks like they are not working though. Staff mood is turned off when the patient starts being examined, and turns on again afterwards.


Alberth289346:
Hacked the general diagnosis to give both patient and doctor an active mood.
As soon as the diagnosis starts though:

Screenshot from 2025-05-09 06-53-43

Both moods do survive the diagnosis process though:

Screenshot from 2025-05-09 06-54-36

PR #2859, opened 2025-05-14
Add test for setCellFlags

TheCycoONE:
Nearly all the library linkages were wrong. The rule is to link the library to the target where it is used. PRIVATE if it is in cpp code only, INTERFACE if it is in header code only, and PUBLIC if it is in both. Fixing this was required to get the test to link properly.

strings.h was invalid as a header. It could have been fixed by marking all the definitions inline, but I chose to put them in a new file instead.


TheCycoONE:
@lewri I'd like you to verify the windows artifacts still work; since I changed all the targets.

TheCycoONE:
Moved to WIP since we're getting a segfault in appveyor that I need to investigate.

lewri (commented):
I've tested `win-dev`;  it loads fine and works with `RelWithDebInfo`. Do you need `win-x64-release` testing too?

Edit: I suspect the segfaults might indicate a specific case where the the program might fall over, that I didn't find.

Alberth289346:
Saw nothing weird, but apparently something is broken in the new test.

I generally don't really trust all the magic numbers in Lua interface code, I keep thinking someone must have written a better solution for that.

TheCycoONE:
Yeah, the magic numbers in lua are all right, and it's not failing on clang/gcc on my local or github actions, or msvc 2022 on github actions, but consistently does with the appveyor which is also running msvc 2022.

It would be nice to have a trace for that segfault...

TheCycoONE:
Actually tests weren't running for windows in github action due to a path issue. Fixed that - let's see if it's a windows / msvc thing.

TheCycoONE:
Ready again.

Alberth289346:
Lots of nice changes there, looks good!

PR #2860, opened 2025-05-14
Fix swingdoor mood

Alberth289346:
*Fixes swingdoor mood position.

The old positions were not entirely correct (animations don't start and/or end center-tile). Also, it used hard-coded frame offsets but the various animations differ in the number of frames, so the end-position was often outside the frame range.

I ran a test in a somewhat flaky version, but the nurse worked:

Screenshot from 2025-05-14 22-32-20
Screenshot from 2025-05-14 22-32-43


Alberth289346:
Didn't try the test saves either. I can do that tomorrow, although I don't expect any problem.

lewri:
#2794

Alberth289346:
Ran another test. No picture, but the mood of a patient works too for the swing door.

Also ran the test saves, no problems.

PR #2863, opened 2025-05-16
Add mood position of the cast remover animations

Alberth289346:
Fix mood positions of cast remover animations.

Screenshot from 2025-05-16 21-16-44
Screenshot from 2025-05-16 21-17-09

Test saves also tried, and no problems.


Alberth289346:
Addresses #2794

Issue #2819, opened 2025-04-08
[Bug] Unreadable low contrast text for unicode languages in Annual Report


ARGAMX:
### Describe the issue

Unreadable low contrast text for unicode languages in Annual Report when select any Trophy.

0.68.0 and Before #2745 

<img width="912" alt="Image" src="https://github.com/user-attachments/assets/7e96bea9-4f8a-4f5e-9c0f-56f52d1fdea0" />

After #2745

<img width="912" alt="Image" src="https://github.com/user-attachments/assets/4b69a8a3-634d-4bf5-8fd6-3feade2a7acd" />

After #2782

<img width="912" alt="Image" src="https://github.com/user-attachments/assets/ee0ee367-17b2-44d4-83f5-4ce8cd3c2ce5" />

### Steps to Reproduce

1. Select any unicode language in Settings (Rus/Ukr/Pol/Cze/Hung) 
2. *Load save game attached to this ticket*
3. Press 5 and wait for Annual Report.
4. In Annual Report go to Awards and Trophies screen.
5. Select statuette at the bottom.
6. Observe unreadable low contrast text.

### Expected Behaviour

Text should be contrast and readable.

### Have you enabled any current CorsixTH Experimental Features?

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

Yes

### CorsixTH Version

5ff228a trunk master

### Operating System

MacOS Sonoma 14.5

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

Al-27:
why not give the player the option to toggle between the two modes...say `Linear RGB`(use a better name) can be `on` or `off` and perhaps a tooltip to further explain what it does.
```cpp
    int iRed = palette::get_red(iColour);
    int iGreen = palette::get_green(iColour);
    int iBlue = palette::get_blue(iColour);
    uint8_t cIntensity = static_cast<uint8_t>((iR + iG + iB) / 3);
    
    if (std::strcmp(sOption, "Linear RGB") == 0) 
    {
        // - Convert from nonlinear RGB to linear RGB.
        //   gamma = around 2.2 but depends on the screen.
        //   R, G and B are fractions.
        const double gamma = 2.2;
        double linR = std::pow(iRed / 255.0, gamma);
        double linG = std::pow(iGreen/ 255.0, gamma);
        double linB = std::pow(iBlue / 255.0, gamma);
        // - Compute luminance Y in XYZ space.
[...]

PR #2871, opened 2025-05-24
Allow colour override for unicode fonts.

Alberth289346:
- Can be set with the second parameter of the 'setSheet' Lua call.
- The colour is ignored by the bitmap fonts.

Addresses #2819


Alberth289346:
This patch only enables colour override in CPP.

CorsixTH/Src/th_gfx_font.cpp
```diff
@@ -262,20 +262,22 @@ FT_Error freetype_font::set_face(const uint8_t* pData, size_t iLength) {
...
     int iWidth;
     int iHeight;
+    argb_colour sheet_colour = 0;
     unsigned int iSprite = *sCharToTry - 31;
-    if (pBitmapFontSpriteSheet->get_sprite_size(iSprite, &iWidth, &iHeight) &&
-        pBitmapFontSpriteSheet->get_sprite_average_colour(iSprite, &colour) &&
+    if (font_spritesheet->get_sprite_size(iSprite, &iWidth, &iHeight) &&
+        font_spritesheet->get_sprite_average_colour(iSprite, &sheet_colour) &&
         iWidth > 1 && iHeight > 1) {
+      font_colour = ((colour & 0xFF000000) == 0) ? sheet_colour : colour;
```
TheCycoONE, line 280: Any fully transparent colour feels unnecessary when the default is just 0

Alberth289346: Only recognize `0` now.

CorsixTH/Src/th_lua_gfx.cpp
```diff
@@ -277,11 +280,18 @@ int l_freetype_font_new(lua_State* L) {
...
 
 int l_freetype_font_set_spritesheet(lua_State* L) {
+  // Any fully transparent colour falls back to getting the font colour from
+  // the sprite sheet.
+  argb_colour colour = 0;
+
   freetype_font* pFont = luaT_testuserdata<freetype_font>(L);
   sprite_sheet* pSheet = luaT_testuserdata<sprite_sheet>(L, 2);
+  if (!lua_isnoneornil(L, 3)) {
+    colour = static_cast<uint32_t>(luaL_checkinteger(L, 3));
```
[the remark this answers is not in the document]
Alberth289346: it has r, g, and b parameters now

CorsixTH/Src/th_lua_gfx.cpp
```diff
@@ -277,11 +280,23 @@ int l_freetype_font_new(lua_State* L) {
...
+  // Colour 0 falls back to using the average colour of the sprite sheet.
+  argb_colour colour = 0;
+
   freetype_font* pFont = luaT_testuserdata<freetype_font>(L);
   sprite_sheet* pSheet = luaT_testuserdata<sprite_sheet>(L, 2);
+  if (!lua_isnoneornil(L, 3)) {
+    int red = static_cast<int>(luaL_checkinteger(L, 3));
+    int green = static_cast<int>(luaL_checkinteger(L, 4));
+    int blue = static_cast<int>(luaL_checkinteger(L, 5));
+    red = std::max(0, std::min(255, red));
```
TheCycoONE, line 292: good excuse for `int alpha = static_cast<int>(lua_optinteger(L, 6, 255));` ?
not sure anyone would ever want that... probably not.

PR #2874, opened 2025-05-25
Unicode font colour support (lua side)

TheCycoONE:
Adds the lua infrastructure to support specifying a colour when loading a font.

Also includes an undefined behaviour fix for unicode font drawing that I saw while developing the change.

I got rid of the varargs and overloaded parameters in Graphics:drawFont, for me at least it is much easier to read this way.


ARGAMX:
Thank you TheCycoONE. 👍
This is a very useful PR. 
Yesterday after Albert's 2871 `Allow colour override for unicode fonts` I have started the same work but I have lost fight against damned varargs.

TheCycoONE:
I fixed the tests and the indentation. I've found it's not being applied when switching language from bitmap to ttf; but does after restart - looking into that.

lewri:
> I fixed the tests and the indentation. I've found it's not being applied when switching language from bitmap to ttf; but does after restart - looking into that.

Could you call an `App:restart()`? like we do when selecting TH install?

CorsixTH/Lua/graphics.lua
```diff
@@ -432,52 +436,34 @@ function Graphics:loadLanguageFont(name, sprite_table, ...)
...
-      end
-    end
-    sprite_table = self:loadSpriteTable(unpack(arg, 1, n_pass_on_args))
-    if n_pass_on_args < #arg then
-      x_sep, y_sep = unpack(arg, n_pass_on_args + 1, #arg)
-    else
-      x_sep, y_sep = nil, nil
-    end
-  end
+function Graphics:loadFontAndSpriteSheet(dir, name, complex, palette, x_sep, y_sep, ttf_col)
```
lewri, line 443: Is the font not already a spritesheet?

In my experience, and isn't you, the previous `loadFont` function is hard to grab without any added doc. If you're making a new function to replace it; please can I suggest adding doc this time around.

CorsixTH/Lua/graphics.lua
```diff
@@ -432,52 +436,34 @@ function Graphics:loadLanguageFont(name, sprite_table, ...)
...
-    -- The font doesn't contain an uppercase M, so (in all likelihood) is used
-    -- for drawing special symbols rather than text, so the original bitmap
-    -- font should be used.
-  elseif self.language_font then
-    use_bitmap_font = false
-  end
+  return self:loadFont(sprite_table, x_sep, y_sep, ttf_col, force_bitmap)
+end
+
+function Graphics:loadFont(sprite_table, x_sep, y_sep, ttf_col, force_bitmap)
```
lewri, line 450: I am assuming this is now a private function?

CorsixTH/Lua/dialogs/resizables/main_menu.lua
```diff
@@ -30,6 +30,12 @@ local col_bg = {
   blue = 198,
 }
 
+local label_ttf_col = { -- off white
+  red = 250,
+  green = 249,
+  blue = 245,
+}
```
lewri, line 37: Wasn't this pure white before? Or am I remembering it wrong.

lewri (commented):
I'm not sure I'll get to reviewing save behaviour as quick as you'd like; but previously breaks happened in these kind of scenarios:

1. Loading an old savegame that cause a very_old_save UIInformation box to appear.
2. Loading savegames with specific font sheets open (previously these were Award (I think), Bank Manager ledger, Casebook(?)).
3. Loading a savegame which fails to load, then trying to load something else (new or a different save).

In general, if a font glitch happens it persists across the program until said program is started again. That was a slightly off-topic issue from this previously but somewhat related. We only concluded that in general that isn't the intended behaviour of what the program should be doing with loaded fonts.

Also I think isn't there somewhere in the C++ or TH.lua that also loads a font? Might be related to the minimalist error handler.

CorsixTH/Lua/graphics.lua
TheCycoONE: It is not.

CorsixTH/Lua/graphics.lua
TheCycoONE: No, the font contains a sprite sheet but it is not a sprite sheet. I can add docs.

TheCycoONE:
> > I fixed the tests and the indentation. I've found it's not being applied when switching language from bitmap to ttf; but does after restart - looking into that.
> 
> Could you call an `App:restart()`? like we do when selecting TH install?

Not necessary, I'm working on the fix. It impacts more things so you might want to delay testing.

TheCycoONE:
Ok, uploaded a version with a fixed cache key. This fixes both switching languages and how widely the colours apply. Before this patch it was setting it too broadly (everywhere).

ARGAMX:
after cache fix If I set any Unicode language and then load any savegame I get error: 

Image

ARGAMX (approved):

PR #2876, opened 2025-05-26
Check file handle of output file as well.

Alberth289346:
Reported by @tobylane in the chat.

Opening output file of rmc-decode CLI was not checked.


lewri (approved):
To me it looks fine, matches the other check formats.

PR #2877, opened 2025-05-26
Dont play drop sound on cancel hire

ARGAMX:

In CTH If you click Esc while placing a new employee, the hiring will cancel but you still will hear the placement sound.

In original TH there is no any sound in this case.

This PR removes placement sound in CTH for this case.

### DIFF
--- a/CorsixTH/Lua/dialogs/place_staff.lua
+++ b/CorsixTH/Lua/dialogs/place_staff.lua
@@ -58,16 +58,16 @@ function UIPlaceStaff:close()
   if newcomer_hired then
     self.ui:playSound("plac_st2.wav")
   elseif old_staff_member then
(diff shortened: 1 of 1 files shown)


ARGAMX:
> Is it only escape, and not right click then?

Both.

lewri:
> * Your comment `-- Drop sound`, I assume you meant `Play sound` instead. Drop would imply you don't want to play it.

Actually I see now what you meant you wanted to comment what the sound was. The fact it took me a few minutes probably means the comment itself isn't very clear

ARGAMX:
> Hope those help.

Thank you.
Okay.
The suggestions are good.
I applied them all.

The option with alternative 1 seemed the simplest and easiest to read to me.

PR #2879, opened 2025-05-29
[Pending] Puke sound on vomit

ARGAMX:

In original TH when patient pukes you firstly hear the puke sound. And then vomit hitting the floor sound.

In CTH If patient pukes you don't hear any puke sound. Only sound vomit hitting the floor sound.

This PR add patient puke sound in CTH for this case.

[vommit_kingdom2.sav.zip](https://github.com/user-attachments/files/20506492/vommit_kingdom2.sav.zip)

btw
puke sound is `puke.wav` and `puke2.wav`.
vomit hitting the floor sound is `vomsplat.wav`.
you can hear vomit hitting the floor sound.
but there is no any `vomsplat.wav` mention on the source code.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/vomit.lua
+++ b/CorsixTH/Lua/humanoid_actions/vomit.lua
@@ -36,6 +36,9 @@ local action_vomit_end = permanent"action_vomit_end"( function(humanoid)
 end)
 
 local function action_vomit_start(action, humanoid)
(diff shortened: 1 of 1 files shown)


lewri:
NB: Pending a check of adding this to the cpp sound  override roster instead to match the correct frames

lewri:
As some vomit sounds now stray 10 frames from the start of the animation; #2892 is more favourable

ARGAMX:
closed in preference to #2892

PR #2880, opened 2025-05-29
Refactor th map cpp

Alberth289346:
- Improve name of two 'block' map class variables.
- Eliminate magic constant block array indices.
- Further smaller improvements.

### DIFF
--- a/CorsixTH/Src/th_map.h
+++ b/CorsixTH/Src/th_map.h
@@ -188,7 +188,7 @@ enum tile_layer : uint8_t {
   west_wall = 2,
   ui = 3,
 
-  _num_tile_layers = 4  //!< Number of parts in a tile.
+  _num_tile_layers = 4  //!< Number of layers in a tile.
 };
 
 struct map_tile {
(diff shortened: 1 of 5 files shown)


CorsixTH/Src/th_map.h
```diff
@@ -182,6 +182,15 @@ enum class temperature_theme {
   yellow_red     //!< Gradients of yellow, orange, and red
 };
 
+enum tile_part : uint8_t {
+  ground = 0,
+  north_wall = 1,
+  west_wall = 2,
+  ui = 3,
+
+  _num_tile_parts = 4  //!< Number of parts in a tile.
```
TheCycoONE, line 191: I'm not familiar with this leading underscore use in c++; did you source it from anywhere?

TheCycoONE:
I see you avoided the 'layer' naming, but historically they were drawn bottom up, so layer made some sense. Do you plan on changing that?

Alberth289346:
> I see you avoided the 'layer' naming, but historically they were drawn bottom up, so layer made some sense. Do you plan on changing that? If not I'd rather go with tile_layer than tile_part.

A renderer may want to think in layers, but a map is not a renderer. For this reason I opted for the more neutral 'part'.

I don't mind much though, so I changed it.

CorsixTH/Src/th_map.h
Alberth289346: I checked for a way to get the value from the enum declaration itself, but SO only listed highly convoluted solutions.

Settling for adding it as literal in the enumeration, I think it's not at equal footing with the other values, it's something more meta. So I wanted to make that clear in the name (which seems to be working :p ). As both Python and Lua have `_` prefixes for more private things, I thought it a good idea to do that here too.

Not that I care much about the precise name, dropped the leafing `_`.

PR #2886, opened 2025-06-01
Fix font cache and lua bind casting

### LINKED ISSUE #2885: [Bug] Member call on address which is not freetype_font
### Describe the issue

```
/home/stephen/projects/CorsixTH/CorsixTH/Src/th_lua_gfx.cpp:322:21: runtime error: member call on address 0x6e7b212ad138 which does not point to an object of type 'freetype_font'
0x6e7b212ad138: note: object is of type 'font'
 be be be be  c0 c0 d4 92 37 5d 00 00  80 bb 22 21 db 6d 00 00  ff ff ff ff 01 be be be  00 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'font'
/home/stephen/projects/CorsixTH/CorsixTH/Src/th_gfx_font.cpp:245:17: runtime error: member call on address 0x6e7b212ad138 which does not point to an object of type 'freetype_font'
0x6e7b212ad138: note: object is of type 'font'
 be be be be  c0 c0 d4 92 37 5d 00 00  80 bb 22 21 db 6d 00 00  ff ff ff ff 01 be be be  00 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'font'
```

### Steps to Reproduce

1. Compile CorsixTH with address sanitizer enabled 
2. Start the game
3. Go to settings
4. Change the screen resolution

### Expected Behaviour

Screen resolution should change, no errors.

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

### CorsixTH Version

6123fa8ace141ef96f7a350ab9c57e7c40d4ed74

### Operating System

Arch Linux

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

### Additional Information

Possibly related to #2884

TheCycoONE:
Fixes #2885

Fixing the cache eliminated the problem. The cache being broken was a regression in my code for allowing font colors.

For correctness and because I wrote the code already this patch also fixes the chain of pointer casting when the base font class is desired, the same way animation_base was solved.

### DIFF
--- a/CorsixTH/Src/th_lua.h
+++ b/CorsixTH/Src/th_lua.h
@@ -25,8 +25,12 @@ SOFTWARE.
 #include "config.h"
 
 #include <cassert>
+#include <cstdio>
+#include <cstring>
+#include <initializer_list>
 #include <memory>
 #include <new>
+#include <type_traits>
 
 #include "lua.hpp"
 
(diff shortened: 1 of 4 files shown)


CorsixTH/Src/th_lua.h
```diff
@@ -463,4 +467,79 @@ void luaT_pushtablebool(lua_State* L, const char* k, bool v);
...
+}  // namespace
+
+// Find the appropriate derived class of base for a given
+// userdata object.
+//
+// th_lua_internal.h defines our c++ binding class metatables:
+// metatable: { __index = method_table, ... }
+// method_table: { ... }
+// method_table metatable: { __class_name = class_name }
+template <typename B, typename T1, typename T2>
```
TheCycoONE, line 516: I'm not really a fan that this is limited to exactly two types, and extending it requires copying most of the code.

This was the reason I moved the class name lookup logic out of this function.

I think if we run into a situation where we want more than 2 types I will move to a lambda based solution instead.

i.e.

```cpp
luaT_touserdata_base(lua_State* L, int idx, std::unordered_map<const char*, std::function<B*(void*)>> cast_map) {...}
```

Alberth289346:
LGTM, although complicated.

PR #2793, opened 2025-03-08
 Add machine menu

Matroftt:
image
* Performs #2104 

I will use RS for Remaining Strength / Current Strength, and TS for Total Strength.

**Describe what the proposed change does**
- Implements 480x400 machine menu which indicates following information:
  - Machine's name and position;
  - Machine's RS & TS & Their ratio;
  - How many times machine was used;
  - Is handyman assigned to fix the machine;
  - Warning if machine is going to blow up soon.
 
- Buttons:
  - Clicking on a machine in list will bring you to that machine and open its window;
  - Clicking on a smoke indicator will open replace machine dialog;
  - Clicking on a assign indicator will open handyman window which is going to fix it.
  - Clicking on RS, TS, % and Used headers will sort machines.

- Bind: It is F10 by default, but you can change it in settings -> hotkeys -> panel.
- Currently it can be opened using charts menu, but i am not sure if this is an optimal way.

PR #2887, opened 2025-06-01
Fix machine menu crash on get marker

ARGAMX:

Fixes machine menu crash.

After merging machine menu #2793 it crash on any machine select:
1. Take/build master 91658705
2. Load any save. (for example [room_crashed_doc_in_grab_surg.sav.zip](https://github.com/user-attachments/files/17475797/room_crashed_doc_in_grab_surg.sav.zip))
3. Open machine menu (F10).
4. Press any machine name to locate camera on it.
5. Get Lua crash `attempt to call a nil value (method 'getMarker')`

This happened because there was another parallel PR which changed name of `getMarker` function to `getPrimaryMarker` and `getSecondaryMarker`, so it not available anymore by old name.


Alberth289346:
Markers for machines are somewhat tricky, especially when they are used.
In that situation, there is perhaps a patient and a machine in the animation, and likely both are at a different marker position.

As such, markers for machines is probably something that needs some more thought.

As for the problem here, is it really needed to show the position that precise?
The code is pointing the entire display of the user to that spot, is at the center of the tile versus a 30 pixels above the tile making any significant difference?

If not, then perhaps the better solution is to drop the marker?

ARGAMX:
> Markers for machines are somewhat tricky, especially when they are used. In that situation, there is perhaps a patient and a machine in the animation, and likely both are at a different marker position.
> 
> As such, markers for machines is probably something that needs some more thought.
> 
> As for the problem here, is it really needed to show the position that precise? The code is pointing the entire display of the user to that spot, is at the center of the tile versus a 30 pixels above the tile making any significant difference?
> 
> If not, then perhaps the better solution is to drop the marker?

I agree to have a better and more accurate method. 

Especially if the game will not crash.

I didn't quite understand what is meant by dropping the marker.

Alberth289346:
> I didn't quite understand what is meant by dropping the marker.

I propose to remove the marker completely.

```
 local x, y = self.ui.app.map:WorldToScreen(entity.tile_x, entity.tile_y)
 local px, py = entity.th:getSecondaryMarker()
 self.ui:scrollMapTo(x + px, y + py)
```
`x` and `y` is the tile position in the map with the machine. As far as I know, if you scroll to that, you can see for example 1/4th of the hospital centered at the machine tile. `px` and `py` are offsets from `x` and `y`, but they are very small, my guess is less than the size of a tile.

Given the somewhat dubious nature of markers for machines, especially when it is being used, is it a good idea to NOT use `px` and `py`?

The user is seeing 1/4th of the hospital, will they even notice that small `(px, py)` difference?

ARGAMX:
> I propose to remove the marker completely.
> 
> The user is seeing 1/4th of the hospital, will they even notice that small `(px, py)` difference?

Ok, I got the point. 

I have removed the getMarker and get even better result on 800x600.
At high resolution, yes, the result is not even noticeable.

Thanks.

PR #2892, opened 2025-06-03
Allow one or two sounds to be played with an animation frame

Alberth289346:
- Allow a random choice of one of two sounds to be played with a frame of an animation.
- Add sounds to frames of various vomit animations.

### DIFF
--- a/CorsixTH/Src/th_gfx.cpp
+++ b/CorsixTH/Src/th_gfx.cpp
@@ -1466,10 +1466,11 @@ void animation::set_animation_kind(animation_kind anim_kind) {
 }
 
 namespace {
-//! Storage of one or two sounds. If two sounds are stored, both are selected in 50% of the cases.
+//! Storage of one or two sounds. If two sounds are stored, both are selected in
+//! 50% of the cases.
 struct sound_pair {
-  sound_pair(int16_t sound): soundA(sound), soundB(-1) { }
-  sound_pair(int16_t soundA, int16_t soundB): soundA(soundA), soundB(soundB) { }
+  sound_pair(int16_t sound) : soundA(sound), soundB(-1) {}
+  sound_pair(int16_t soundA, int16_t soundB) : soundA(soundA), soundB(soundB) {}
 
   int16_t get_sound() const {
     if (soundB < 0) {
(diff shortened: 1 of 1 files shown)


CorsixTH/Src/th_gfx.cpp
```diff
@@ -1466,23 +1466,64 @@ void animation::set_animation_kind(animation_kind anim_kind) {
...
     // Female flying to heaven (anim 3220)
-    {6987, 123},
+    {6987, sound_pair(123)},
+
     // Using Computer (anim 2098)
-    {4213, 35},
-    {4215, 35},
-    {4224, 35},
-    {4230, 35},
+    {4213, sound_pair(35)},
```
lewri, line 1502: I'm not convinced `sound_pair` is a necessary call where only one sound is provided? But I also don't know C++ well so maybe it makes more sense to always run this function

CorsixTH/Src/th_gfx.cpp
```diff
@@ -1466,23 +1466,64 @@ void animation::set_animation_kind(animation_kind anim_kind) {
...
+    {1902, sound_pair(58, 114)}, // Animation 1034
+    {4149, sound_pair(58, 114)}, // Animation 2056
+    {6901, sound_pair(58, 114)}, // Animation 3184
+    {8819, sound_pair(58, 114)}, // Animation 4138
+    {9105, sound_pair(58, 114)}, // Animation 4204
+    {9565, sound_pair(58, 114)}, // Animation 4324
+    {9654, sound_pair(58, 114)}, // Animation 4384
+    {9944, sound_pair(58, 114)}, // Animation 4452
+    {10007, sound_pair(58, 114)}, // Animation 4476
+//    {??, sound_pair(58, 114)}, // Animation 4792
```
lewri:
```suggestion
    {10989, sound_pair(58, 114)}, // Animation 4792
```
But is also a duplicate of 4452,  maybe add but leave commented out?

lewri:
Works fine to me

Edit: Can we add a description to the PR once ready?

CorsixTH/Src/th_gfx.cpp
Alberth289346: The `{a, b}` pair are a key/value pair of the `std::map`, which is similar to key/value pair in a lua table except you write an `=` between the values of the pair then.
The key/value pair used to be `size_t, int` (2 integer numbers). I changed it to `size_t, sound_pair` (an integer + class instance). For this reason the second parameter must construct an object.
Lines 1471 and 1472 give 2 options for that. One with 1 sound value, and one with two sound values for the pair.

Alberth289346: In our cpp, class names are also lower snake case. They like to hide in sea of other names in the code :)

ARGAMX (approved):
This PR resolves the `Puke sound` problem described in #2879. Verified. 👍

PR #2824, opened 2025-04-13
Marker split add and improve

Alberth289346:
Use primary marker for patients and secondary marker for staff.

- Allow setting marker positions both for patients and staff.
- Adjust existing marker positions
- Fix sweeping and normal door animations
- Add first shot marker positions for 'simple' ones (standing animations).

Probably simplest to read by individual commit

*Addresses #2794

Issue #2177, opened 2022-05-29
[Bug] Give a summary of the problem


Alberth289346:
### Describe the issue

Handyman window does not show the handyman properly while busy.
![Screenshot from 2022-05-29 16-41-30](https://user-images.githubusercontent.com/3254232/170882392-1476eb19-8554-4f1f-a2c8-2033aad40bc2.png)
![Screenshot from 2022-05-29 16-40-25](https://user-images.githubusercontent.com/3254232/170882416-e2fc5d4a-5f22-4229-9f96-888c030b800c.png)

As you can see, the handyman is not centered in the circular area in his window.

Bug doesn't seem new, both shots are from a development version, but I have confirmed the sweeping screenshot also in fd7a9420 (2 fixes further than `v0.66-beta1`) and in `v0.65`.

### Steps to Reproduce

Likely any not-too-old save will work, but from scratch, build a minimal hospital, and wait a while.

### Expected Behaviour

Handyman should be standing properly in front of the camera of course!

### Save Game

Save game attached to report: Yes/No

### CorsixTH Version

0.65 to 0.66-beta1

### Operating System

Mint LTS

### Theme Hospital Version

CD

### Gamelog.txt

```bash
nothing of interest.
```

lewri:
It looks like the mini window doesn't know where to focus as the handyman repairs... Unless when repairing it makes the game think he is standing on a specific tile in respect of the machine?

Alberth289346:
I am pretty sure the handyman is standing on the tile being sweeped (liked that better than swept :) ). I checked the sweep animation (1878), and the handy man is standing at tile (0,0):
![image](https://user-images.githubusercontent.com/3254232/171044890-107cd0a2-cc2a-4a9d-98a1-f90e88395a43.png)

It also has a `setMarker` call, and for fun I changed that:
``` diff
diff --git a/CorsixTH/Lua/humanoid_actions/sweep_floor.lua b/CorsixTH/Lua/humanoid_actions/sweep_floor.lua
index bc4a9d51..bcab4655 100644
--- a/CorsixTH/Lua/humanoid_actions/sweep_floor.lua
+++ b/CorsixTH/Lua/humanoid_actions/sweep_floor.lua
@@ -35,7 +35,7 @@ local animation_numbers = {
   1874,
   1878,
 }
-TheApp.animation_manager:setMarker(animation_numbers, {-1.05, -0.05})
[...]

PR #2898, opened 2025-06-07
Handle in-flight actions without mood marker

lewri:
Fixes a regression from #2824 where some entities had no marker for a current action.

Argument is that the save system could be investigated not to try and cache this data, so it pulls the new mood marker information.

Untested, but I think this change may have some additional knock on effects to #2177


Alberth289346:
Thanks

PR #2907, opened 2025-06-11
Remove dynamic info for GrimReaper

lewri:
Accidentally added in a previous PR a request to update the GrimReaper's dynamic info, but this method doesn't exist for that entity.

Instead, remove the line but leave a comment so people know they're special :)

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/grim_reaper.lua
+++ b/CorsixTH/Lua/entities/humanoids/grim_reaper.lua
@@ -44,6 +44,6 @@ function GrimReaper:afterLoad(old, new)
     self.mood_marker = 2
   end
 
-  self:updateDynamicInfo()
+  -- The grim repear has no dynamic info currently, so don't refresh it
   Humanoid.afterLoad(self, old, new)
 end


ARGAMX (approved):

CorsixTH/Lua/entities/humanoids/grim_reaper.lua
```diff
@@ -44,6 +44,6 @@ function GrimReaper:afterLoad(old, new)
     self.mood_marker = 2
   end
 
-  self:updateDynamicInfo()
+  -- The grim repear has no dynamic info currently, so don't refresh it
```
Alberth289346: But there is no such function at all here, so the remark makes no sense by itself.

The only way that the comment makes sense is **if you know** that other humanoids do have a call to update the dynamic info at this spot.

It could be made clear with a line like `-- self:updateDynamicInfo()` below it, but that seems a bit silly too. I am sure you know, but that may not be universal.

As an alternative, it's also possibly to make the grim reaper less surprising, eg by adding
``` lua
function GrimReaper:updateDynamicInfo()
  -- The grim reaper has no dynamic info to update.
end
```
Then you can treat the grim reaper more as a normal humanoid with very little content.

lewri: That causes a luacheck error if you do it that way, which we also want to avoid.

However, removing it completely with no note may well lead us to adding it back again

lewri: I'm going to potentially suggest the worst of both but:

```lua
-- self:updateDynamicInfo() [The grim reaper has no dynamic info to update.]
```

lewri: Okay, I swear these kind of stub functions used to throw errors. Guess not anymore. Updating it

Alberth289346:
Nice that it works !

LGTM

PR #2917, opened 2025-06-17
Fix raw map overlays

### LINKED ISSUE #2916: [Bug] Byte x map overlays broken loading custom games
### Describe the issue

If you attempt to view byte information as a map overlay in a custom game, the map code will use the wrong path for the map file and crash.

Forcing the game to give us the error also highlights that it tries to also target the path of where the map originally came from, even if a save is transferred to another device
```yaml
Warning: Could not find file
Level.L/Users/argus/Documents/Games/Theme-Hospital/CorsixTH/Development/Corsix.Lab.app/Contents/MacOS/../Resources/Campaigns/hoppalong.level,
the Theme Hospital data is incomplete.
```

The issue stems partly from how Map attempts to get byte information:

https://github.com/CorsixTH/CorsixTH/blob/5431213426af4233654962ec38fed917dfba2d7b/CorsixTH/Lua/map.lua#L620-L630
https://github.com/CorsixTH/CorsixTH/blob/5431213426af4233654962ec38fed917dfba2d7b/CorsixTH/Lua/map.lua#L502-L517

### Steps to Reproduce

1.  Load a save from #2913 (unfortunately for @ARGAMX this will work as it was his save)

2.  Go to Debug > Map Overlay > Byte 0 & 1

3. Observe crash

### Expected Behaviour

When retrieving the byte flags, it needs to do two things:
1. Attempt to retrieve the map data from the game itself rather than a external file. Or if not possible, look for the map file in common locations and error cleanly if it can't be found
2. When calling these overlays, work out whether this is a TH game or a custom game and what path to choose to get `thData`.

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

### CorsixTH Version

0.69.0 Beta 1

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml
Lua\map.lua:628: attempt to index a nil value (local 'thData')
stack traceback:
        Lua\map.lua:628: in method 'loadDebugText'
        Lua\dialogs\menu.lua:672: in field 'handler'
        Lua\dialogs\menu.lua:399: in method 'onMouseUp'
        Lua\window.lua:1642: in field 'onMouseUp'
        Lua\ui.lua:870: in function <Lua\ui.lua:857>
        (...tail calls...)
        Lua\app.lua:1189: in function <Lua\app.lua:1184>
```

### Additional Information

NB: This is a developer tool, so priority will always be low.

TheCycoONE:
Store raw map information in the tile_nodes so that it can be used for debug overlays instead of reloading the map file.

Fixes #2916

Note: Saves from older versions will not show overlay flags since they need to be saved with the map in this version to avoid the need for the original map file.


CorsixTH/Lua/map.lua
```diff
@@ -578,11 +578,11 @@ function Map:updateDebugOverlayHeliport()
   end
 end
 
-function Map:loadDebugText(base_offset, xy_offset, first, last, bits_)
+function Map:loadDebugText(first, last, bits_)
```
Alberth289346, line 582:
```suggestion
-- If both 'first' and 'last' are numbers, they are indices starting at 0.
function Map:loadDebugText(first, last, bits_)
```

Alberth289346:
LGTM

PR #2922, opened 2025-06-20
0.69.0 Beta 2 changelog

lewri:
*Fixes #*

-
-
-

### DIFF
--- a/CorsixTH/com.corsixth.corsixth.metainfo.xml
+++ b/CorsixTH/com.corsixth.corsixth.metainfo.xml
@@ -22,30 +22,33 @@
   <releases>
     <release version="0.69.0~beta2" date="2025-06-21" type="development">
       <description>
-        <p>Features</p>
+        <p>Features:</p>
         <ul>
-          <li>Custom level/campaign creators can now use an existing level file
+          <li>Custom level&sol;campaign creators can now use an existing level file
               as an overlay to copy and extend their level file with the
               &num;overlay variable. See the &apos;Custom Level&apos; Wiki page
               for more information.</li>
         </ul>
-        <p>Changes</p>
+        <p>Changes:</p>
         <ul>
           <li>You can now hear patients vomiting, instead of just the vomit
               splatters</li>
           <li>Lakes no longer show in the Town Map as per the original game</li>
         </ul>
-        <p>Translations</p>
+        <p>Translations:</p>
         <ul>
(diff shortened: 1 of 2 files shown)


CorsixTH/com.corsixth.corsixth.metainfo.xml
```diff
@@ -20,6 +20,38 @@
   </provides>
   <launchable type="desktop-id">com.corsixth.corsixth.desktop</launchable>
   <releases>
+    <release version="0.69.0~beta2" date="2025-06-21" type="development">
+      <description>
+        <p>Features</p>
+        <ul>
+          <li>Custom level/campaign creators can now use an existing level file
```
lewri:
```suggestion
          <li>Custom level&sol;campaign creators can now use an existing level file
```

changelog.txt
```diff
@@ -1,3 +1,29 @@
...
+# Translations
+* Russian translation has been updated. Thanks @ARGAMX
+
+# Bug Fixes
+* Fixed a crash in General Diagnosis
+* Fixed a compatibility issue with mood markers on old saves
+* Fixed a crash with the grim reaper
+* Fixed a bug with the hotkey assignment menu
+* Fixed some anomalies in the original campaign's map files
+* Fixed an issue with handymen ignoring cleaning duties. Please note existing
```
TheCycoONE, line 25: This was a regression in beta 1 right?

I'd like to track regressions so we can write them out of the final release notes.

changelog.txt
```diff
@@ -1,3 +1,29 @@
...
+# Changes
+* You can now hear patients vomiting, instead of just the vomit splatters
+* Lakes no longer show in the Town Map as per the original game
+
+# Translations
+* Russian translation has been updated. Thanks @ARGAMX
+
+# Bug Fixes
+* Fixed a crash in General Diagnosis
+* Fixed a compatibility issue with mood markers on old saves
```
TheCycoONE: This was also a regression.

changelog.txt
lewri: I think most bug fixes here are regressive behaviours? Do you have a preference on how to mark them as such or are you fine as "Fixed regression"?

TheCycoONE: Yes, fixed 0.69-beta1 regression:

lewri: All but one of those bug fixes was a regression. It looked silly adding shared text to all of them so it has its own header instead.

Issue #1862, opened 2021-04-27
Update Screenshot tool to save as PNG/JPG


lewri:
**What feature or enhancement would you like to add?**
Change the SDL2 libraries to support screenshots in PNG (or lossless JPG) format rather than the ancient and bloated BMP format

**How would this feature benefit the gameplay of CorsixTH?**
BMPs are not a common image format any more by modern standards. As such, we should be moving to PNG or JPG.

**Was the feature in the original game?**
No

**What version of CorsixTH are you using?**
 0.64

**Does this feature/enhancement relate to a current problem?**
No

**Additional information**
Some quick investigation does point to SDL2 image libraries existing that should be able to capture in PNG. We may use them already.

TheCycoONE:
So it's captured here, my preference is to use libpng directly rather than pulling in SDL_Image.

tobylane:
Does this qualify? https://github.com/mohammadul/SDL_SavePNG

Alberth289346:
Not sure if the license is good, but the approach seems ok at first sight.

lewri:
The license looks compatible to me, I think it just needs the author to be mentioned in our license file
https://en.wikipedia.org/wiki/Zlib_License?wprov=sfla1

mugmuggy:
The original fork https://github.com/driedfruit/SDL_SavePNG/ has a null ptr deference fix.

PR #2925, opened 2025-06-25
Add Fatal Error window

lewri:
- CorsixTH lacks any visual indicator when a game crashes outside the timer handler / empty action queue. Only that a message is printed to the console, which may not be visible to the player depending on setup.
- Adds a fatal error window when the game crashes into a state where it can no longer continue. Depending on the context it can either be:
   -  Show an error message to a player to reset the program, if key handlers are still functioning
   -  Show an error message to close the message for debugging, if debug mode is on
   -  Show an error message to all use cases if the needed key handlers to control the mouse no longer function
- Makes the timer error recovery message translatable.

As I'm not entirely sure if the error may become bothersome for devs I'm open to adding a config item to disable them entirely if necessary.

image
image


CorsixTH/Lua/languages/english.lua
```diff
@@ -783,6 +783,10 @@ errors = {
     missing_setting = "No difficulty and level number given in overlay field of custom level.",
   },
   cannot_restart_missing_files = "Sorry, but this level cannot be restarted because of missing files %s or %s.",
+  recoverable = "Sorry, but an error has occurred. There can be many reasons - see the console for details. Would you like to attempt a recovery?",
```
lewri, line 786: This could be expanded too if needed to note it's possible to report.

CorsixTH/Lua/app.lua
```diff
@@ -1246,14 +1246,16 @@ function App:run()
...
-        "log window for details. Would you like to attempt a recovery?",
+        _S.errors.recoverable,
         --[[persistable:app_attempt_recovery]] function()
         self.world:gameLog("Recovering from error in timer handler...")
         entity.ticks = false
         self.eventHandlers.timer = handler
       end
       ))
+    elseif self.world then
+      local can_reset = handler ~= "buttonup" and handler ~= "buttondown" and handler ~= "motion"
```
lewri, line 1257: This needs changing to instead just reference last dispatch type

TheCycoONE:
I'd love to have that link to the file or directory. That would need to have some c++ support and os-specific code, so we would probably also need a way to indicate support.

TheCycoONE:
I suppose that could be built later.

Alberth289346:
Instead of a magic `x`, have a "restart" button?

ARGAMX:
I like this type of window. We clearly needed it. 👍

My thoughts: 

Often when a Lua crash occurs game UI becomes unresponsive. Will this button on the window work in this case?

If this button works successfully, then how should I act if I still want to send this file to the developers? Based on the context of the message, I should press this button. Why should I restart the game in this case?

Matroftt:
Russian strings
```
recoverable = "Сожалеем, возникла ошибка. Причин может быть несколько - подробности смотрите в консоли. Хотите попытаться восстановить?",
fatal_can_reset = "Фатальная ошибка возникла в CorsixTH, которая не может быть восстановлена. Нажмите "X" чтобы перезапустить программу. Ваше последнее автосохранение, скорее всего, было Autosave%d.// Если вы хотите сообщить об этом разработчикам, журнал для этой сессии находится в:://",
fatal_can_debug = "Фатальная ошибка возникла в CorsixTH, которая не может быть восстановлена. Ваше последнее автосохранение, скорее всего, было Autosave%d.// Разработчики могут нажать "X", чтобы закрыть это сообщение для отладки. Журнал для этой сессии находится в:://",
fatal_cant_reset = "Фатальная ошибка возникла в CorsixTH, которая не может быть восстановлена. Пожалуйста, закройте программу и запустите ее снова. Ваше последнее автосохранение, скорее всего, было Autosave%d.// Если вы хотите сообщить об этом разработчикам, журнал для этой сессии находится в:://"
```

lewri:
> Instead of a magic `x`, have a "restart" button?

I'm open to making a reset button using a graphic.



> Often when a Lua crash occurs game UI becomes unresponsive. Will this button on the window work in this case?

I've made it so that if the motion and button handlers are responsive, the reset is available. If they aren't the button and message change.



> If this button works successfully, then how should I act if I still want to send this file to the developers? Based on the context of the message, I should press this button. Why should I restart the game in this case?

This adds to the concern in the OP `NB: Currently a downside is that this prevents the user from saving the crashed game. This could be changed too.` -- I _could_ force it to drop a save on crash.

ARGAMX:
> This adds to the concern in the OP `NB: Currently a downside is that this prevents the user from saving the crashed game. This could be changed too.` -- I _could_ force it to drop a save on crash.

You mean save game after crashes automatically? Just like we save logs? I think this is good idea.
Then the user will have a crash save anyway, no matter whether he needs it or not.

And we can add a "close game" button to this crash window. The user will be able to close the game and then think about whether he wants to send this event to immediately us, or just restart and continue playing.

I think if we just suggest restarting the game to continuing to play, then the possibility that user will file a crash report decreases. I know from my own experience. 🙂

Also in this window we can give the user a link to github issues where he can upload crash log and save.

May be in the text in this crash window we can suggest user to take a photo of the current window on the phone, so that he can then go through the steps and do everything.

TheCycoONE:
getting so close to a full blown crash reporter. The screenshot can be automated, the save can as long as the state isn't too corrupt. Creating a github issue is a challenge, I'm not sure how to do it without distributing personal access tokens with the app, and that's out of the question; or else setting up another service to translate crash reports into github issues.

lewri:
> getting so close to a full blown crash reporter. The screenshot can be automated, the save can as long as the state isn't too corrupt. Creating a github issue is a challenge, I'm not sure how to do it without distributing personal access tokens with the app, and that's out of the question; or else setting up another service to translate crash reports into github issues.

I've seen apps launch mail clients to send out the information, but mainly for Android.
I previously pondered such level of interaction, but deemed it to get quite messy fast.

For screenshots I'd rather have #1862 first as github issues does not support BMP embeds.

Alberth289346:
The light blue text on the orange background is (despite my appreciation for orange), somewhat hard to read for me.

lewri:
> The light blue text on the orange background is (despite my appreciation for orange), somewhat hard to read for me.

I can understand that. Red is hard to work with sometimes, in this case, not making the blue work. I'll have a play about with the colours.

ARGAMX:
Red color is pretty tough for backgrounds.
Only white and yellow looks pretty good and contrast on it.
May be you can consider black as background. 
Black as background looks pretty gloomy. Good choice for crash 🙂

yvw2e9td70561

PR #2929, opened 2025-07-06
Add png screenshot support

### LINKED ISSUE #1862: Update Screenshot tool to save as PNG/JPG

Alberth289346:
*Fixes #1862 

Allows writing PNG screenshot files by giving a `.png` (case-insensitive) extension to the name of the file to write.

The feature in itself works. Further / future work is likely:
- Currently the program always needs `libPNG` and `ZLib` to build. Whether that is a good idea may need a policy decision.
- Whether to keep the BMP format is another policy decision.
- Tell Lua about the available file type(s).
- Tell the users.
- ??
- Profit!

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -10,7 +10,6 @@
     "luafilesystem",
     "lpeg",
     "libpng",
-    "zlib",
     {
        "name": "fluidsynth",
        "platform": "!osx",


CorsixTH/Src/th_lua_gfx.cpp
```diff
@@ -715,10 +715,32 @@ int l_surface_rect(lua_State* L) {
...
+
 int l_surface_screenshot(lua_State* L) {
   render_target* pCanvas = luaT_testuserdata<render_target>(L);
-  const char* sFile = luaL_checkstring(L, 2);
-  if (pCanvas->take_screenshot(sFile)) {
+  const char* file_path = luaL_checkstring(L, 2);
+
+  // Decide the image format.
+  size_t file_path_length = strlen(file_path);
+  bool write_bmp = str_endswith(file_path, file_path_length, ".bmp");
```
lewri: Are we at the point of immediate implementation? In which case I've favour just dropping bmp support entirely.

Alberth289346: Is also a good option.

CorsixTH/CMakeLists.txt
```diff
@@ -249,6 +249,24 @@ else()
   message("Lua modules are not linked and will be loaded at runtime")
 endif()
 
+# Find libpng
+if(VCPKG_TARGET_TRIPLET)
+  find_package(PNG REQUIRED)
```
TheCycoONE: If these conditions are the same maybe we don't need the VCPKG branch? Confirmed the usage notes for libpng, and zlib.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -722,7 +724,69 @@ void render_target::set_cursor_position(int iX, int iY) {
...
+  }
+  png_infop info_write_data = png_create_info_struct(png_write_data);
+  if (info_write_data == nullptr) {
+    png_destroy_write_struct(&png_write_data, nullptr);
+    std::fclose(fp);
+    return false;
+  }
+
+  // Setup error handling.
+  if (setjmp(png_jmpbuf(png_write_data))) {
```
TheCycoONE: To review further, there can be complications with jmp in c++.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -722,7 +724,69 @@ void render_target::set_cursor_position(int iX, int iY) {
   cursor_y = iY;
 }
 
-bool render_target::take_screenshot(const char* sFile) {
+namespace {
+
+bool write_rgb_png(int width, int height, png_bytep pixels, int pitch,
+                   const char* file_path) {
+  // Open the file for writing
+  FILE* fp = std::fopen(file_path, "wb");
```
TheCycoONE: You don't use this file until much later in png_init_io but need to handle it in the error case many times before that, it likely makes sense to delay opening the file until later.

CorsixTH/Src/th_lua_gfx.cpp
TheCycoONE: Also in favour of dropping the runtime switching. On the fence about using compile time switching to avoid a hard dependency on libpng, but leaning towards just always enabling - it is pretty small and ubiquitous.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -27,6 +27,8 @@ SOFTWARE.
 #include "th_gfx_font.h"
 #endif
 #include <SDL.h>
+#include <png.h>
+#include <zlib.h>
```
TheCycoONE: What are we doing with zlib?

CorsixTH/Src/th_gfx_sdl.cpp
TheCycoONE: So the caveat here is that it's really easy to trigger undefined behaviour, if any non-trivial c++ objects would have been destroyed by the unwinding of the longjmp.

I don't think that is the case here, but I find it hard to reason about longjmp. I think you may be leaking the array below? You also couldn't turn that c array into a c++ one because that would definitely trigger the ub.

I think libpng offers alternative error handling.

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: libpng uses it to write pixels

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: The libpng document suggested to open the file first, in order to avoid committing "considerable resources" to PNG writing if the file cannot be opened, but may be old advice.

CorsixTH/CMakeLists.txt
Alberth289346: I tried understanding CMake a few times already, but its tutorial is a list of "monkey do this, copy that, click here" without explaining the system behind it at all. I am very bad at remembering random sequences of things without a higher level context why the stuff is how it is.
It's reference contains too many details which are no doubt fine if you understand the system behind it, except I don't.

So I asked for a little help, and Toby then kindly added the "if" parts with vcpkg. vcpkg has no explanation how to use it either, and our wiki only points at its home page. The home page seems to be about "integrating vcpkg in builds" instead of explaining how to use it as a user. After more thinking I concluded I likely don't even need vcpkg at all, since I have the libraries compiled and installed at my system already with `sudo apt install libpng zlib`.

The addition by Toby didn't work by itself. As a blunt experiment I then copied the "if" parts to the "else" parts and that worked.
Obviously they can be simplified, but as they needed some policy decisions first anyway, I decided to just keep it as-is.

CorsixTH/Src/th_lua_gfx.cpp
Alberth289346: I'd drop the screenshot feature instead of falling back to `bmp`.

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: Hmm, yes good spot.

Row pointers allocation should be moved then indeed.

Alternative error handling seems to need a custom compile:
```
If you would rather avoid the complexity of setjmp/longjmp issues,
you can compile libpng with PNG_NO_SETJMP, in which case
errors will result in a call to PNG_ABORT() which defaults to abort().

You can #define PNG_ABORT() to a function that does something
more useful than abort(), as long as your function does not
return.
```

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: Actually it's the entire file: `Libpng uses zlib for its compression and decompression of PNG files.`

But close enough.

CorsixTH/Src/th_lua_gfx.cpp
Alberth289346: With compile time switching, Lua should know the file type to give to C++ and/or the user. If the feature is dropped instead, Lua should also know, so it can tell the player the screenshot didn't work.

Thus simplest is to have the feature always available.

Alberth289346: Lua should also know if the feature is dropped.

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: After moving some of the stuff here I found new leaks. To fix once and for all, I centralized memory management at the top and bottom of the function.

CorsixTH/Src/th_gfx_sdl.cpp
TheCycoONE: Right but why include the header?

CorsixTH/Src/th_gfx_sdl.cpp
TheCycoONE: I understand you can specify an error function and use that to throw a c++ exception which avoids the jump. There's an example for read here: https://gist.github.com/xmartin-yao/42d9d4b034f6a41b7c9409a45781c65f which could probably be adapted.

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: Deleted the include.

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: Moved.

CorsixTH/Src/th_gfx_sdl.cpp
Alberth289346: I don't see the problems here. the PNG library is doing the jump, and it ends here. I centralized managing allocated data, so it always gets released.

But as apparently that is not enough, I introduced a `throw`, which in my view creates much convolution in managing the data, but whatever.

Alberth289346:
Can't close some of the boxes for some reason.

Anyway, to move forward, this patch needs a policy decision on whether or not to
- drop BMP support.
- make PNG support optional. If so, it needs a path to tell Lua about having PNG support (possibly by returning a list of supported formats or so).

- CMake configuration needs work in any case.

CorsixTH/CMakeLists.txt
```diff
@@ -249,6 +249,24 @@ else()
...
+  target_link_libraries(CorsixTH_lib PUBLIC PNG::PNG)
+else()
+  find_package(PNG REQUIRED)
+  target_link_libraries(CorsixTH_lib PUBLIC PNG::PNG)
+endif()
```
tobylane:
```suggestion
find_package(PNG REQUIRED)
target_link_libraries(CorsixTH_lib PUBLIC PNG::PNG)
```

CorsixTH/CMakeLists.txt
```diff
@@ -249,6 +249,24 @@ else()
...
+
+# Find zlib
+if(VCPKG_TARGET_TRIPLET)
+  find_package(ZLIB REQUIRED)
+  target_link_libraries(CorsixTH_lib PUBLIC ZLIB::ZLIB)
+else()
+  find_package(ZLIB REQUIRED)
+  target_link_libraries(CorsixTH_lib PUBLIC ZLIB::ZLIB)
+endif()
```
tobylane:
```suggestion
find_package(ZLIB REQUIRED)
target_link_libraries(CorsixTH_lib PUBLIC ZLIB::ZLIB)
```

TheCycoONE:
> Can't close some of the boxes for some reason.
> 
> Anyway, to move forward, this patch needs a policy decision on whether or not to
> 
>     * drop BMP support.
> 
>     * make PNG support optional. If so, it needs a path to tell Lua about having PNG support (possibly by returning a list of supported formats or so).
> 
>     * CMake configuration needs work in any case.

We're ok requiring libpng and dropping bmp support.

Alberth289346:
> As we are considering making Freetype required, which uses libpng, it's then no benefit to make this use of png optional.

This sounded weird, so I looked it up ( https://gitlab.freedesktop.org/freetype/freetype/-/blame/master/include/freetype/config/ftoption.h#L263 ):
```
   * PNG bitmap support.
   *
   *   FreeType now handles loading color bitmap glyphs in the PNG format.
   *   This requires help from the external libpng library.  Uncompressed
   *   color bitmaps do not need any external libraries and will be supported
   *   regardless of this configuration.
   *
   *   Define this macro if you want to enable this 'feature'.
   *
   *   If you use a build system like cmake or the `configure` script,
   *   options set by those programs have precedence, overwriting the value
   *   here with the configured one.
```
Looks not needed for now.

Alberth289346:
Dropped BMP support, applied the CMake patches by Toby, and changed the `.bmp` file extension in Lua.

I think this is all that needs to be done.

CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -740,12 +827,21 @@ bool render_target::take_screenshot(const char* sFile) {
...
     SDL_UnlockSurface(pRgbSurface);
 
-    if (readStatus != -1) SDL_SaveBMP(pRgbSurface, sFile);
+    if (readStatus != -1) {
+      try {
+        write_rgb_png(width, height,
+                      static_cast<png_bytep>(pRgbSurface->pixels),
+                      pRgbSurface->pitch, file_path);
+        ok = true;
+      } catch (int zero) {
```
TheCycoONE: This is c++, use an unnamed parameter instead of a bogus assignment.

Alberth289346:
```suggestion
      } catch (int) {
```

Alberth289346: Hmm, why can't I apply my suggestion?

Alberth289346: Done.

Don't think I ever used that before. Even with base classes I like to have a variable name for documentation purposes.

Alberth289346:
Windows build seems to have died on:
```
error: https://www.inf.puc-rio.br/~roberto/lpeg/lpeg-1.1.0.tar.gz: WinHttpSendRequest failed with exit code 12175. A security error occurred
```
I just now downloaded it successfully.

Alberth289346:
` vcpkg` died on the same failed download, it seems.

Alberth289346:
Restarted the windows build

TheCycoONE:
Conflicts due to merging Toby's pr first.

Alberth289346:
Rebased with resolved conflicts in
``` diff
diff --git a/CorsixTH/Lua/api_version.lua b/CorsixTH/Lua/api_version.lua
index 75660ce3..dfc61d77 100644
--- a/CorsixTH/Lua/api_version.lua
+++ b/CorsixTH/Lua/api_version.lua
@@ -32,4 +32,4 @@ Note: This file compiles as both Lua and C++. */
 
 #endif /*]] --*/
 
-return 2695;
+return 2696;
diff --git a/vcpkg.json b/vcpkg.json
index 842e6101..a9d9127f 100644
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -9,6 +9,8 @@
     "sdl2",
     "luafilesystem",
     "lpeg",
+    "libpng",
+    "zlib",
     {
        "name": "fluidsynth",
        "platform": "!osx",
```

Alberth289346:
No errors in appveyor log. The message however states:
```Build execution time has reached the maximum allowed time for your plan (90 minutes).```

lewri:
As long as the github version succeeds we can dismiss appveyor's failure

PR #2935, opened 2025-07-12
Refactor rendering

Alberth289346:
Refactor the current rendering code.

The complete patch seems non-readable, but each commit should be doable.


TheCycoONE:
Looks good to me, nice commit organization.

PR #2941, opened 2025-07-19
Remove unknown HTML entities from metainfo

rmnvgr:
The Flatpak build [failed](https://github.com/flathub-infra/vorarbeiter/actions/runs/16390734671/job/46316385526#step:14:2471) because the metainfo file contains unknown entities: `&sol;` and `&num;`. Since this is an XML file, these HTML entities are unknown and the validation fails.

- Replace the unknown entities with the proper character

### DIFF
--- a/CorsixTH/com.corsixth.corsixth.metainfo.xml
+++ b/CorsixTH/com.corsixth.corsixth.metainfo.xml
@@ -126,9 +126,9 @@
       <description>
         <p>Features:</p>
         <ul>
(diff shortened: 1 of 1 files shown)


lewri:
@TheCycoONE does this commit need pushing into the 0.69.0 tag too?

lewri (approved):
This would've been my fault in being overly cautious with html codes in the past

TheCycoONE:
The tag has been moved

PR #2942, opened 2025-07-20
Savegame bump post-release

lewri:
**Merge only after release is published**

- Allocate the 10 savegame version reserve for patches.


lewri:
This can merge now

PR #2945, opened 2025-07-26
Make freetype and sdl mixer always required

tobylane:
- Change the CMake, vcpkg and lua parts to always build with freetype and mixer. The C changes are naive, I just dealt with the ifdefs and cmakedefines. All reordering is from clang-format. Does this change anything for the IWYU comments?
- Remove the check for compile_opts.os before using it
- ~~Freetype won't include the default features (libpng, bzip, brotli and zlib) because they aren't needed to show the Noto font (or Arial). Should we have these in case other fonts need them?~~ removed, a question for later


vcpkg.json
```diff
@@ -8,34 +8,33 @@
...
+          "platform": "!osx"
+        },
+        "libflac",
+        "mpg123",
+        "opusfile"
       ]
     },
+    {
+      "name": "freetype",
+      "default-features": false
```
TheCycoONE: Can you leave this out of this PR so it's just dropping the option; we can explore minimizing features later.

PR #2957, opened 2025-08-03
Include machine and cost parameters for replacement dialog

### LINKED ISSUE #2811: `_S.confirmation.replace_machine` requires amendment in French
As noted at the top of Regional Differences on this wiki article: https://tcrf.net/Theme_Hospital#Regional_Differences

For CorsixTH we pass the machine upgrade cost to format this string. However, French is missing this cost and the machine name.
There is an entrance in `French.lua` below that uses the non-cost format:
https://github.com/CorsixTH/CorsixTH/blob/9ec1281f0d74855e6793cb7bc8edb0e1c92d0054/CorsixTH/Lua/languages/french.lua#L503

Simple fix, translate the English string to French and include the machine name `%s` cost format `%d`:
`Are you sure you want to replace the %s for $%d?` --- french genders things so it might need some amending to e.g. `the machine %s`?

lewri:
*Fixes #2811*

- Add the `%s` and `%d` parameters we pass when replacing machines for other languages.

### DIFF
--- a/CorsixTH/Lua/languages/french.lua
+++ b/CorsixTH/Lua/languages/french.lua
@@ -31,7 +31,7 @@ Inherit("original_strings", 1)
 ----------------------------------------------------------- Override -----------------------------------------------------------
 rooms_short.ward = "Salle des malades"
 rooms_long.ward = "Salle des malades"
-confirmation.replace_machine = "Voulez-vous vraiment remplacer cette machine %s pour $%d ?" -- French originally had no parameters
+confirmation.replace_machine = "Voulez-vous vraiment remplacer cette machine %s pour $%d ?" -- French originally did not show machine name or cost
 
 misc.save_failed = "ERREUR : partie non sauvegardée." -- Much more french
 misc.cant_treat_emergency = "Votre hôpital ne peut pas traiter cette urgence car la maladie n'a pas été découverte. N'hésitez pas à réessayer."
--- a/CorsixTH/Lua/languages/french.lua
+++ b/CorsixTH/Lua/languages/french.lua
@@ -31,6 +31,7 @@ Inherit("original_strings", 1)
 ----------------------------------------------------------- Override -----------------------------------------------------------
 rooms_short.ward = "Salle des malades"
 rooms_long.ward = "Salle des malades"
+confirmation.replace_machine = "Voulez-vous vraiment remplacer cette machine %s pour $%d ?" -- French originally had no parameters
 
 misc.save_failed = "ERREUR : partie non sauvegardée." -- Much more french
 misc.cant_treat_emergency = "Votre hôpital ne peut pas traiter cette urgence car la maladie n'a pas été découverte. N'hésitez pas à réessayer."
@@ -503,7 +504,6 @@ confirmation = {
   overwrite_save = "Il y a déjà une partie sauvegardée ici. Êtes-vous sûr de vouloir l'écraser ?",
   delete_room = "Voulez-vous vraiment détruire cette salle ?",
   sack_staff = "Êtes-vous sûr de vouloir licencier ?",
-  replace_machine = "Voulez-vous vraiment remplacer cette machine ?"
 }
 
 -- The originals of these strings contain one space too much


CorsixTH/Lua/languages/french.lua
```diff
@@ -503,7 +504,6 @@ confirmation = {
   overwrite_save = "Il y a déjà une partie sauvegardée ici. Êtes-vous sûr de vouloir l'écraser ?",
   delete_room = "Voulez-vous vraiment détruire cette salle ?",
   sack_staff = "Êtes-vous sûr de vouloir licencier ?",
-  replace_machine = "Voulez-vous vraiment remplacer cette machine ?"
```
ARGAMX, line 506: Why did we remove this parameter and move it to another place in the document? 🤔

Here is the `confirmation` block. And we extract it from here, but the other confirmations remain in place?

The change is small, but it is not clear why...

CorsixTH/Lua/languages/french.lua
```diff
@@ -31,6 +31,7 @@ Inherit("original_strings", 1)
 ----------------------------------------------------------- Override -----------------------------------------------------------
 rooms_short.ward = "Salle des malades"
 rooms_long.ward = "Salle des malades"
+confirmation.replace_machine = "Voulez-vous vraiment remplacer cette machine %s pour $%d ?" -- French originally had no parameters
```
ARGAMX: `-- French originally had no parameters` 
What meaning will this comment have for those who will read it here? 🤔

Considering that they will have no context, that this parameter value was different in the past.

What is it for?..

CorsixTH/Lua/languages/french.lua
lewri: So this string exists in the original language file. Arguably somebody didn't need to add it again. 
However, if we need to change a string from the original it goes to the top of the file in the overrides section (or should).

CorsixTH/Lua/languages/french.lua
lewri: It was only added to explain to other translators why this override exists. I'm not aware our translators always read the source PRs. It is not necessary perse, it was more an audience consideration. I can remove it, however.

Though if to keep, the comment could be expanded on better.

ARGAMX: ok, I get it. 👍
Yes, that makes sense in light of the overriding.

CorsixTH/Lua/languages/french.lua
ARGAMX: ah, interesting. 🙂

I hadn't touched the languages built into the game before, such as
French
Spanish
German
Swedish
Italian
But now I see that they are somehow strangely arranged with these text moving in case of overriding... it's was not obvious at all...

I went to the Wiki to find answers.
Opened the `Localization` page.
https://github.com/CorsixTH/CorsixTH/wiki/Localization

But I did not find an explanation or instructions that some lines should be moved if they modified. And where it should be transferred. 🙂

It seems like this is some kind of generally audience accepted rule that is not described in wiki.

Yes, perhaps there is some sense in this.

CorsixTH/Lua/languages/french.lua
TheCycoONE: Is it right to say cette machine, and then then name of the machine? It sounds wrong to me but I'm not native.

lewri: I asked @Sanndow and they said that this was fine. I imagine french also normally present the machine name with << >> but we don't support that character (they have a specific character where << and >> are a singleton).

Machines are always feminine and as such you could always drop machine and use 'la'.

CorsixTH/Lua/languages/french.lua
lewri: The translators wiki page is likely still a bit unloved :(

CorsixTH/Lua/languages/french.lua
lewri:
```suggestion
confirmation.replace_machine = "Voulez-vous vraiment remplacer cette machine %s pour $%d ?" -- French originally did not show machine name or cost
```

ARGAMX (approved):

Issue #1638, opened 2020-06-09
[META] Graphical Glitches in game


lewri:
#### Describe the issue
As found yesterday, we imported a load of graphical glitches from the googlecode that were originally linked to a meta issue.
As a result, we don't have a meta issue in github. This issue is to serve that purpose.

The following issues are documented as graphical glitches:
#403 - Radiation Shield on X-ray visible through wall
#460 - Larger machines stick out of the view window
#463 - Doctor appears under couch 
#472 - Cardio machine glitches with wall
#475 - Training chair glitches with wall
#487 - Additional doors appearing for rooms
#495 - Patient who dies from hair restoration gets bloaty head graphic
#509 - Toilet size glitch on back wall
#510 - Crash trolley in general diagnosis obscures doctor
#512 - Placing two video games together in staff room causes glitch
#514 - Doctor appears they can walk through slack tongue machine
#565 - Destroyed operating theatre can appear through wall
#1255 - Ward bed can glitch through wall

If any of these have since been resolved, they can be crossed off (some are quite old).
Any new graphical glitches should be linked to this meta.

Alberth289346:
Reporting more of these glitches isn't really needed. We know they exist in half a zillion variations, and more if you start looking for them,
Cyco looked at the problem already and established that the current way of painting rooms to the screen is simply broken beyond repair. The current solution is never going to fix all render bugs.

As such, it's not going to be a small local fix with a testcase like you made to verify it's indeed fixed. Instead, the entire paint code must be taken out and replaced by something better. Trouble is, we don't yet know what that "something" should be. I think we also established that likely the current paint code doesn't get enough information to solve the problem, so that needs some thinking too.

This is basically the point where we are currently wrt this problem. No easy solution available, I am afraid.

Alberth289346:
A small overview of experiments I have been doing.

1. Currently, rendering happens by traversing the map diagonally and painting the tiles as they are visited by the traversal.
   Some tiles contain multi-tile animations and thus touch more than their own tile. The algorithm is complicated and speaks about traversal in 2 directions. Also, the code has an early-list, a normal-list, and a bottom-of-list notion.
   Given the existence of this issue, it's clearly not working as intended.

2. My first experiment is a baseline experiment. Drop the complications, and render all tiles diagonally, one at a time. What becomes clear is that large (ie multi-tile) animations glitch sometimes with neighbouring tiles due to incorrect drawing order.

[...]

PR #2958, opened 2025-08-03
Add drawing layers constants

Alberth289346:
*Fixes Rat hole layer

- Introduce constants for the various drawing layers.
- Apply constants instead of magic numbers.
- Drop layers of objects, humanoids, and staff, as they are derived entities.
- Draw a rathole before a side object (instead of after it).

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -23,7 +23,8 @@ self = false
 globals = { -- Globals
             "_A", "_S",
             "corsixth",
-            "action_queue_leave_bench", "class", "compare_tables",
+            "action_queue_leave_bench", "class", "compare_tables", "DrawFlags",
+            "DrawingLayers",
             "destrict", "flag_clear", "flag_isset", "flag_set", "flag_toggle",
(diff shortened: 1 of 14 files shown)


Alberth289346:
Screenshot from 2025-08-03 17-09-05
Screenshot from 2025-08-03 17-09-33

lewri:
I assume related to the meta #1638 ?

Alberth289346:
> I assume related to the meta https://github.com/CorsixTH/CorsixTH/issues/1638 ?

Drawing layers are used there indeed, but they are about order of animations at a single tile (side objects at the back before the main animation, for example).

Issue #1638 is about multiple (multi-tile) animations at different tiles that have conflicts in order at shared tiles between them, usually at tiles (partly) in front of each other.

Here I foremost wanted to cleanup all the magic numbers, and add the rathole fix. The latter is much simpler if you do the former (less searching what values are used where).

(I Couldn't find an issue about the rathole, I guess it didn't get further than the chat.)

PR #2961, opened 2025-08-04
RFC: Cpp 17 and RAII refactor

TheCycoONE:
Update target from C++14 to C++17

Use C++17 string_view and filesystems APIs where appropriate.
Class initializer lists were updated throughout, not due to 17, but just cleanup.

An earlier version of this PR also moved to Designated Initializers, but that is a C++20 feature so it was reverted.

std::filesystems::path is meant for storing owned strings representing filesystem paths. Unfortunately we depend on a lot of libraries that use the narrow convention (fopen) for windows, and that API is built around the wide convention (_wfopen) https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170&redirectedfrom=MSDN so they aren't as convenient for our use case as they might otherwise be.

This bumps the minimum macOS version that can be targeted to 10.15.

C++17 also has std::variant as a type safe replacement for union, but the two places we use union didn't look like good candidates because with variant the individual members aren't named and all of our unions are anonymous with named members that have little in common except that they're mutually exclusive.


TheCycoONE:
~@lewri can you check this on Windows. std::filesystem uses wchar_t there instead of char, and I'm not convinced everything works (assuming it compiles.)~

~If not I will probably revert that code.~

It doesn't compile - I suspected as much.

CorsixTH/Src/iso_fs.cpp
```diff
@@ -499,7 +498,9 @@ void iso_filesystem::build_file_lookup_table(uint32_t iSector,
     if (prefix.empty()) {
       path = ent.filename;
     } else {
-      path = prefix + path_seperator + ent.filename;
+      path += prefix;
+      path += path_seperator;
+      path += ent.filename;
```
Alberth289346, line 468: Original: `path = prefix + path_seperator + ent.filename;`

Ie it doesn't append the path.

CorsixTH/Src/persist_lua.cpp
```diff
@@ -613,10 +609,10 @@ class lua_persist_basic_writer : public lua_persist_writer {
 
  private:
   lua_State* L;
-  uint64_t next_index;
+  uint64_t next_index{1};
   std::string data;
-  size_t data_size;
-  bool had_error;
+  size_t data_size{0};
+  bool had_error{false};
```
Alberth289346, line 611: Not sure I like this initialization. It has magic values seemingly copied from `init()`. Are these supposed to stay in sync with each other?

`had_error` is probably fine, as you cannot go back to "didn't have an error". But then, the initialization should only be here, right?

Alberth289346: `line_sequence` does something similar.

lewri:
Let me know when you need Windows testing again

CorsixTH/Src/iso_fs.cpp
TheCycoONE: Path is an empty string to start, it makes no difference.

CorsixTH/Src/persist_lua.cpp
TheCycoONE: I don't like that init exists; c++ is designed around RAII and we ignore it all over the place which causes friction in modernizing.

Anyway that results in double initialization; because the class is initialized with the parameter defaults or constructor init list, and then the body of the constructor runs.

I guess I can go further and try to eliminate the init / reconstruct the object to reset it.

CorsixTH/Src/iso_fs.cpp
TheCycoONE: This was necessary because prefix is read only now. There is no operator for string_view + string_view.

TheCycoONE:
@Alberth289346 I have updated iso_fs / iso_filesystem to fully flush out where I intend to go with this, which is in line with the direction the language has been heading for the last ~15 years. Could you take a look at that.

CorsixTH/Src/persist_lua.cpp
TheCycoONE: It is not easy, perhaps not possible, to eliminate the init() method in persist_lua because of it's interaction with the lua state which involves referencing the userdata for the class.

However I have eliminated the redundant initialization of member variables pointed out above.

Alberth289346: I don't remember, but `init()` could also be used to draw a fresh graph while the game is running

TheCycoONE: Maybe it should be `reset` then.

Alberth289346:
Phew, lots of changes.

Looks good.

CorsixTH/Src/th_lua_iso.cpp
```diff
@@ -22,38 +22,69 @@ SOFTWARE.
...
-  } else {
-    lua_pushnil(L);
-    lua_pushstring(L, pSelf->get_error());
-    return 2;
+int l_isofs_is_valid_root(lua_State* L) {
+  // Static method so we don't need an instance
+  try {
+    iso_filesystem fs(luaL_checkstring(L, 1));
+    lua_pushboolean(L, true);
+  } catch (std::runtime_error& e) {
```
lewri: Build generated warning here, do you need to fix? Will show up in all future PRs unless suppressed.

PR #2962, opened 2025-08-05
Hotfix room crash on placing staff

### LINKED ISSUE #2954: [Bug] Lua crash on placing staff in room
### Describe the issue

This is a 0.69.0 regression. 
Was brought by #2844

We can get a Lua crash in several rooms by placing staff in that room while a patient is being served in this room.

https://github.com/user-attachments/assets/a4cb0ccc-df39-424b-8902-5b1258c8b062

Affected rooms:

💥 Fracture Clinic - crash
💥 Slack Tongue Clinic - crash
💥 Ultrascan - crash
💥 Cardiogram - crash 
💥 Inflation - crash
💥 General Diagnosis - crash

Not affected rooms:

✅ GP's Office - ok, not affected
✅ Pharmacy - ok, not affected
✅ Ward - ok, not affected
✅ Operating Theater - ok, not affected
✅ Psychiatric - ok, not affected
✅ Blood Machine - ok, not affected
✅ X-Ray - ok, not affected
✅ Scanner - ok, not affected
✅ Jelly Vat - ok, not affected
✅ Hair Restoration - ok, not affected
✅ Decontamination - ok, not affected
✅ Electrolysis - ok, not affected

<details>
<summary>Fixing General Diagnosis PR broke General Diagnosis lol</summary>
The funny thing is that the PR which was intended to fix the flow in General Diagnosis ended up by breaking the General Diagnosis and not only. 
This happened because I tested the General Diagnosis before creating the PR. But during the process of working with the PR, I made some changes, which were not covered by the test targeted on the bug. These new changes in second commit in PR introduced a new problem. Although the targeted bug was fixed.
</details>

----

### Steps to Reproduce

2. Wait for the patient to enter the Fracture Clinic.
3. Go to the recruitment interface and select a new nurse. 
4. Press the recruitment button, but do not put a new Nurse in the hospital yet.
5. Wait for the patient to start sitting in the machine 
6. Now put the nurse in this Fracture Clinic.
7. Observer game crash.

----

### Expected Behaviour

Should not crash

----

### Save Game

[Battenburg_level12_2021_fracture.sav.zip](https://github.com/user-attachments/files/21548281/Battenburg_level12_2021_fracture.sav.zip)

----

### CorsixTH Version

0.69.0

### Gamelog.txt

```yaml

Error in buttonup handler: 
...ents/MacOS/../Resources/Lua/entities/humanoids/staff.lua:520: attempt to index a nil value (field 'integer index')
stack traceback:
	...ents/MacOS/../Resources/Lua/entities/humanoids/staff.lua:520: in method 'isMeandering'
	...Corsix.69.0.app/Contents/MacOS/../Resources/Lua/room.lua:370: in method 'onHumanoidEnter'
	.../Contents/MacOS/../Resources/Lua/dialogs/place_staff.lua:153: in method 'onMouseUp'
	...rsix.69.0.app/Contents/MacOS/../Resources/Lua/window.lua:1642: in field 'onMouseUp'
	...s/Corsix.69.0.app/Contents/MacOS/../Resources/Lua/ui.lua:870: in function <...s/Corsix.69.0.app/Contents/MacOS/../Resources/Lua/ui.lua:857>
	(...tail calls...)
	.../Corsix.69.0.app/Contents/MacOS/../Resources/Lua/app.lua:1189: in function <.../Corsix.69.0.app/Contents/MacOS/../Resources/Lua/app.lua:1184>

```

ARGAMX:
*Fixes #2954*

- Before extracting elements from an array by index, a check has been added to ensure that there are enough elements.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -515,11 +515,18 @@ end
 --[1 idle, 2 meander] or [1 walk, 2 meander].
 --!return true if staff currently has a meander action
 function Staff:isMeandering()
-  if ((self.action_queue[1].name == "idle") or
-    (self.action_queue[1].name == "walk")) and
-    (self.action_queue[2].name == "meander") then
+  local actions_count = #self.action_queue
+  -- if the action queue is empty, we assume that staff is meandering
+  if actions_count == 0 then return true end
+
+  if actions_count >= 2 and
+    (
+      ((self.action_queue[1].name == "idle") or (self.action_queue[1].name == "walk"))
+      and (self.action_queue[2].name == "meander")
+    ) then
     return true
   end
+
   return false
 end
 
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -512,17 +512,16 @@ end
 --! Check whether staff are meandering
 --!return true if staff currently has a meander action
 function Staff:isMeandering()
+  if #self.action_queue < 2 then return false end
+
   -- "meander" action always insert "move" or "idle" action before itself.
   -- so when humanoid "meandering" his action queue usually looks like:
   -- [1 idle, 2 meander] or [1 walk, 2 meander].
-  if #self.action_queue >= 2 and
-    (((self.action_queue[1].name == "idle") or
-    (self.action_queue[1].name == "walk")) and
-    (self.action_queue[2].name == "meander")) then
-    return true
-  end
+  local idle_is_first = self.action_queue[1].name == "idle"
+  local walk_is_first = self.action_queue[1].name == "walk"
+  local meander_is_second = self.action_queue[2].name == "meander"
 
-  return false
+  return (idle_is_first or walk_is_first) and meander_is_second
 end
 
 -- Function to decide if staff currently has nothing to do and can be called to a room where they're needed
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -510,16 +510,12 @@ function Staff:adviseWrongPersonForThisRoom()
 end
 
 --! Check whether staff are meandering
---"meander" action always insert "move" or "idle" action before itself.
---so when humanoid "meandering" his action queue usually looks like:
---[1 idle, 2 meander] or [1 walk, 2 meander].
 --!return true if staff currently has a meander action
 function Staff:isMeandering()
-  local actions_count = #self.action_queue
-  -- if the action queue is empty, we assume that staff is meandering
-  if actions_count == 0 then return true end
-
-  if actions_count >= 2 and
+  -- "meander" action always insert "move" or "idle" action before itself.
+  -- so when humanoid "meandering" his action queue usually looks like:
+  -- [1 idle, 2 meander] or [1 walk, 2 meander].
+  if #self.action_queue >= 2 and
     (((self.action_queue[1].name == "idle") or
     (self.action_queue[1].name == "walk")) and
     (self.action_queue[2].name == "meander")) then
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -520,10 +520,9 @@ function Staff:isMeandering()
   if actions_count == 0 then return true end
 
   if actions_count >= 2 and
-    (
-      ((self.action_queue[1].name == "idle") or (self.action_queue[1].name == "walk"))
-      and (self.action_queue[2].name == "meander")
-    ) then
+    (((self.action_queue[1].name == "idle") or
+    (self.action_queue[1].name == "walk")) and
+    (self.action_queue[2].name == "meander")) then
     return true
   end
 


CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -515,11 +515,17 @@ end
 --[1 idle, 2 meander] or [1 walk, 2 meander].
 --!return true if staff currently has a meander action
 function Staff:isMeandering()
-  if ((self.action_queue[1].name == "idle") or
+  local actions_count = #self.action_queue
+  -- if the action queue is empty, we assume that staff is meandering
+  if actions_count == 0 then return true end
```
lewri: Could you justify the need to return true on this one instead of false? I can't currently infer at all why we want true. I'm sure there are reasons though. (Arguably, we're already in a critically broken state if the action queue is empty)

lewri:
Tested on the general diagnosis and that's fine. Be good for anyone else reviewing just to double check rooms function fine now.

Alberth289346:
```
--"meander" action always insert "move" or "idle" action before itself.
--so when humanoid "meandering" his action queue usually looks like:
--[1 idle, 2 meander] or [1 walk, 2 meander].
```
Lines 513 to 515 do not belong in the function description. They are implementation details of the current solution. Move them into the function.

The idea of the comment above the function is to state what the objective of the function is (ie line 512), what parameter values it expects (none here), and what it returns in all relevant cases. In other words, it describes how to call the function, and what answers can be expect when the function returns. It should **not** explain how the function arrives at its result.

The reason is that you don't need to know what 'meander' means in detail, in order to just call the function. By hiding those details inside the function you avoid reading details that are not relevant for **calling the function**.

The function also becomes more resistant against code changes. If we eg change the meander action internally (for example, if we move the "walk" into the meander action in some way), then currently the function description becomes broken while meandering as a whole is still the same.
To fix you would currently also need to edit the function header comment, while it makes no differences at all in how to call the function.

In general, this should even go as far as not caring about the implementation at all for calling a function.
If we would throw away the action queue and replace it with something else, would the question "is the staff meandering?" be asked differently? My guess is no, a simple boolean answer would still suffice.
That is however not possible in all cases.

The comment is however **very relevant for understanding the internals** of the function. By stating the shape of meander action queue, you explain a lot of things that are otherwise just plain magic. For reading the internals, it's crucial.
If you are reading at function implementation level however, you do read the code and the comments in the code. A developer going there will find them and appreciate them.

ARGAMX:
> Lines 513 to 515 do not belong in the function description. They are implementation details of the current solution. Move them into the function.

Yes, I completely agree.
It should have been written inside the function.
As it does not apply to the function interface.
Albert, Thank you for such a detailed explanation. 😉

CorsixTH/Lua/entities/humanoids/staff.lua
ARGAMX: > Could you justify the need to return true on this one instead of false? I can't currently infer at all why we want true. I'm sure there are reasons though. (Arguably, we're already in a critically broken state if the action queue is empty)

To be honest, I can't justify either false or true. 

I wanted to make the function more safe. For this I made a entry check. But since the function still has to return something, I had to make a choice.

But I don’t understand how the queue works, whether it’s empty or not. And I think function don't need to know these to be safe.

But in fact, yes, in essence this check is not necessary.
The second check does enough to ensure there is no crash. And the first one is simply redundant.

Based on the logic of the task for which I wrote this function, it should return true. But if we read the name and description of the function, then everything is clearly written there and it should return false.

So I'll just remove it and then all the questions will self-disappear. 👍

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -510,16 +510,18 @@ function Staff:adviseWrongPersonForThisRoom()
...
---"meander" action always insert "move" or "idle" action before itself.
---so when humanoid "meandering" his action queue usually looks like:
---[1 idle, 2 meander] or [1 walk, 2 meander].
 --!return true if staff currently has a meander action
 function Staff:isMeandering()
-  if ((self.action_queue[1].name == "idle") or
+  -- "meander" action always insert "move" or "idle" action before itself.
+  -- so when humanoid "meandering" his action queue usually looks like:
+  -- [1 idle, 2 meander] or [1 walk, 2 meander].
+  if #self.action_queue >= 2 and
```
lewri: I'd like to suggest a refactor

```lua
function Staff:isMeandering()
  if #self.action_queue < 2 then return false end

  return (self.action_queue[1].name == "idle" or self.action_queue[1].name == "walk") and
      self.action_queue[2].name == "meander"
end
```
This makes an early return for any invalid action queue states; and means you can run the meander test as the end of the function instead. I also broke a bit of our line width limits, because I think putting your first test on one line (it's x or y); then your second test (and this is z) it might be cleaner.

No need to take it, it might be too concise for preference; but that triple bracket at the start of line 519 isn't ideal.
Edit: width limits could be avoided by having `local queue = self.action_queue`

ARGAMX: Long records can be shortened through local variables. 
This adds clarity and ease of reading.
This resolves triple bracket.

```lua
function Staff:isMeandering()
  if #self.action_queue >= 2 then
    local idle_is_first = self.action_queue[1].name == "idle"
    local walk_is_first = self.action_queue[1].name == "walk"
    local meander_is_second = self.action_queue[2].name == "meander"
    if (idle_is_first or walk_is_first) and meander_is_second then
      return true
    end
  end
  return false
end
```

Alberth289346: Maybe you are not aware, but a condition at `if ... then` produces a boolean value `true` or `false`.
Ie it is fine to return the `...` value itself, instead of first testing if it is `true` in order to `return true`.

Alberth289346: Nothing major though.

ARGAMX: > Maybe you are not aware, but a condition at `if ... then` produces a boolean value `true` or `false`. Ie it is fine to return the `...` value itself, instead of first testing if it is `true` in order to `return true`.

Yes, of course, sure.

```lua
function Staff:isMeandering()
  if #self.action_queue < 2 then return false end

  local idle_is_first = self.action_queue[1].name == "idle"
  local walk_is_first = self.action_queue[1].name == "walk"
  local meander_is_second = self.action_queue[2].name == "meander"

  return (idle_is_first or walk_is_first) and meander_is_second
end
```

How about that?

lewri: I'm happy with this

lewri:
Please add now into master when you're free to :)

PR #2968, opened 2025-08-10
Let staff return to training/research

### LINKED ISSUE #2964: [Bug] Student drops out study and responds to room call
### Describe the issue

When a student goes back to the Training Room from the Staff Room he can answer room calls.
I am unable to check if that 0.69.0 regression or not.

----

https://github.com/user-attachments/assets/a1ae1989-8523-4fa8-8a64-f6aa07d91500

----

### Steps to Reproduce

1. Load attached save game '2964_Student_27_FENBERRY_abandon_study_and_answer_call.sav'.
2. You will see `27 FENBERRY`. He is studying in the Training Room. But currently returning from the Staff Room to the Training Room.
3. Press 3 and soon you will hear the room call to the Scanner.
4. Observe how `27 FENBERRY` abandons his study and heads to the Scanner room.

----

### Expected Behaviour

Students should not pick up room calls while studying, even on returning to the Training Room.

----

### Save Game

[2964_Student_27_FENBERRY_abandon_study_and_answer_call.sav.zip](https://github.com/user-attachments/files/21689363/2964_Student_27_FENBERRY_abandon_study_and_answer_call.sav.zip)

----

### CorsixTH Version

0.69.0

----

lewri:
*Fixes #2964*

- Reworked staff idle check to properly handle returning doctors to research/training.

@ARGAMX please extensively test I've not broken other call states e.g. dropping answers to rooms.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -565,8 +565,8 @@ function Staff:isIdle()
     -- In the corridor and not on_call (e.g. watering or going to room).
     -- The staff is free, unless going back to the training/research.
     room = self.last_room
-    if room then
-      return room.room_info.id ~= "training" and room.room_info.id ~= "research"
+    if room and (room.room_info.id == "training" or room.room_info.id == "research") then
+      return false
     end
 
     return true
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -562,15 +562,13 @@ function Staff:isIdle()
     -- or if the only one in sight is actually leaving.
     return not room:isRoomInDemand()
   else
-    -- In the corridor and not on_call (watering or going to room), the staff is free
-    -- unless going back to the training room or research department.
-    local x, y = self:getCurrentAction().x, self:getCurrentAction().y
-    if x then
-      room = self.world:getRoom(x, y)
-      if room and (room.room_info.id == "training" or room.room_info.id == "research") then
-        return false
-      end
+    -- In the corridor and not on_call (e.g. watering or going to room).
+    -- The staff is free, unless going back to the training/research.
+    room = self.last_room
+    if room then
+      return room.room_info.id ~= "training" and room.room_info.id ~= "research"
     end
+
     return true
   end
 end
--- a/CorsixTH/Lua/entities/humanoids/staff/doctor.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff/doctor.lua
@@ -173,7 +173,7 @@ function Doctor:updateSkill(consultant, trait, amount)
       if self:getRoom().room_info.id == "training" then
         self:setNextAction(self:getRoom():createLeaveAction())
         self:queueAction(MeanderAction())
-        self.last_room = nil
+        self.last_room = nil -- Consultant no longer needs to return to this room
       end
       self:updateStaffTitle()
     end
--- a/CorsixTH/Lua/rooms/training.lua
+++ b/CorsixTH/Lua/rooms/training.lua
@@ -209,7 +209,7 @@ function TrainingRoom:commandEnteringStaff(humanoid)
         self.hospital:giveAdvice({_A.staff_place_advice.not_enough_lecture_chairs})
         humanoid:setNextAction(self:createLeaveAction())
         humanoid:queueAction(MeanderAction())
-        humanoid.last_room = nil
+        humanoid.last_room = nil -- Prevent Doctor returning to this room automatically
       end
     end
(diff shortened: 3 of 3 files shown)


Alberth289346:
Hmm, `last_room` to send them back eh?

So what happens if the doctor graduates (ie finishes their studies), and is waiting for the next job at hand?

lewri:
> Hmm, `last_room` to send them back eh?
> 
> So what happens if the doctor graduates (ie finishes their studies), and is waiting for the next job at hand?

https://github.com/CorsixTH/CorsixTH/blob/3d2b3b199658155d44355c10cee67d7e85e85672/CorsixTH/Lua/rooms/training.lua#L208-L213
and
https://github.com/CorsixTH/CorsixTH/blob/3d2b3b199658155d44355c10cee67d7e85e85672/CorsixTH/Lua/entities/humanoids/staff/doctor.lua#L171-L177
Will fix those (arguably the first is not related to the q but relates to training)

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -562,15 +562,13 @@ function Staff:isIdle()
...
-    if x then
-      room = self.world:getRoom(x, y)
-      if room and (room.room_info.id == "training" or room.room_info.id == "research") then
-        return false
-      end
+    -- In the corridor and not on_call (e.g. watering or going to room).
+    -- The staff is free, unless going back to the training/research.
+    room = self.last_room
+    if room then
+      return room.room_info.id ~= "training" and room.room_info.id ~= "research"
```
ARGAMX: Maybe we should leave it as it was?
I didn't understand why this particular check was changed.
It was correct in itself...

If we look at the whole function, its philosophy is to `return true` at the end, but
on the way to this `return true` there are many barriers with `return false`.

This pattern is easy to read and the function's operation becomes more obvious.

And here we 'remove' one of these `return false` barriers.
And now the function as a whole is no longer reads so fluently.

I would leave it as is.
Moreover, it does not fix anything as I understand. Correct?
I changed it back and the fix still works.

```suggestion
    if room and (room.room_info.id == "training" or room.room_info.id == "research") then
      return false
```

But In general, fix works good to me, I tested it. Good job. 👍
I like the rest. 👌

ARGAMX:
> Hmm, `last_room` to send them back eh?
> 
> So what happens if the doctor graduates (ie finishes their studies), and is waiting for the next job at hand?

Very good question.
I thought about it too.
But when I tested it, this freshly graduated student still responded to the call.
So this case works fine.

Alberth289346:
@lewri : Perhaps add a comment to the `nil` assignments why they're there?

lewri:
> @lewri : Perhaps add a comment to the `nil` assignments why they're there?

Sorry, where in the code?

CorsixTH/Lua/entities/humanoids/staff.lua
lewri: Would it make it worse if I said I actually wanted to do an actual one line statement 🤣 -- this was actually meant to be my compromise.
I can extend it back out if absolutely necessary, though from a coder's perspective I don't think the `false` added anything we couldn't ascertain already.

Alberth289346:
> Sorry, where in the code?

Sorry for not being clear, I should have referred more precisely.

I meant at https://github.com/CorsixTH/CorsixTH/pull/2968#issuecomment-3172853623 , lines 212 respectively 176

CorsixTH/Lua/entities/humanoids/staff.lua
ARGAMX: > Would it make it worse if I said I actually wanted to do an actual one line statement 🤣 -- this was actually meant to be my compromise. I can extend it back out if absolutely necessary, though from a coder's perspective I don't think the `false` added anything we couldn't ascertain already.

Yes, seems it would be better to return it back to how it was.

Yes, functionally nothing will change. but in previous form it was easier to read.
And in previous form it also was possible to add more conditions after this one.
Because in this new variant it turns out that now it can be only the last one.
Because now it always makes an `return`.
And if we want to add something else after it, then we need to turn it back to the form it was.

So I don't see any benefits in change it from the form it was.

TheCycoONE:
Since you're fiddling with last_room, I don't think we clear it when staff are placed in the hall. Maybe we want to?

ARGAMX:
> Since you're fiddling with last_room, I don't think we clear it when staff are placed in the hall. Maybe we want to?

I think that it actually exists and is cleaned quite well.
I always put docs in the hallway first, and then in the staff room if I want to keep them from going back to their room.
And this trick works good.
So looks like nothing to tweak.

ARGAMX (approved):
Good work 👍

Alberth289346 (approved):
LGTM

PR #2969, opened 2025-08-11
Eliminate pragma pack casts and raw memory in sound

TheCycoONE:
Use std containers for memory.
Avoid non-portable packed structure casting.
Fix rule of 3 violations.

I'd expect a further refactor to load the sound data on construction, but that's further than I have time to take this before I leave.


TheCycoONE:
Bah format. This will need to sit.

Alberth289346:
Nice improvements !

LGTM

PR #2970, opened 2025-08-11
Remote repeat in World:nextEmergency()

lewri:
- Convert the deprecated `repeat .. until` loop to a while loop in `World:nextEmergency()`


Alberth289346:
LGTM

PR #2971, opened 2025-08-13
Fix disrupt service in rooms by staff replacing

### LINKED ISSUE #2950: [Bug] Error in room logic that allows to disrupt service almost in any room
Image

### Describe the issue

This is not a 0.69.0 regression.

During the investigating and fixing #1328 I discovered that this bug is affects not only the General Diagnosis room.
Similar bug can be reproduced in many other rooms.

The main point of the room bug is that after the patient has entered the room, but before the manipulations in the room have started, put new staff in this room, and grab the old ones and put him outside. After that the room stops functioning correctly or we get visual glitches or the game crashes.

While fixing #1328 I didn't find the solution to fix all rooms at once.

Affected rooms:

Not fatal:
👎 GP's Office - not fatal service suspension 
👎 Inflation - not fatal service suspension
👎 Hair Restoration - not fatal service suspension
👎 Blood Machine - not fatal service suspension
👎 Psychiatric - not fatal service suspension
👎 X-Ray - not fatal service suspension
👎 Scanner - not fatal service suspension
👎 Pharmacy - not fatal service suspension
👎 Jelly Vat - not fatal service suspension
👎 Decontamination - not fatal service suspension
👎 Electrolysis - not fatal service suspension

Fatal or VIsual Glitches:
💥Slack Tongue Clinic - continue to service but glitches (scenario 1) or crash (scenario 2) are happening
💥Ultrascan - continue to service but glitches (scenario 1) or crash (scenario 2) are happening
💥Fracture Clinic - continue to service but glitches (scenario 1) or crash (scenario 2) are happening
💥Cardiogram - continue to service but fatal crash after first serve (scenario 1)

Successful rooms:

✅ Ward - ok, not affected.
✅ Operating Theater - ok, resolved by #2812.
✅ General Diagnosis - ok, resolved by #2844.

Not fatal service suspension - the staff does nothing in the room. Staff can be grabbed from the room, further patient will leave the room. After this, the room is ready to operate as usual. 
But if you don't grab the doctor, and just kick the patient out, the game may crash when the next patient enters the room.

Not fatal service suspension video example:

https://github.com/user-attachments/assets/a2a6a778-9f06-4b7e-bb09-ccce569b76da

Fatal service suspension video example:

https://github.com/user-attachments/assets/c6fd0102-c639-4981-870a-daf23e546901

----

### Steps to Reproduce

Steps to Reproduce similar to described here
https://github.com/CorsixTH/CorsixTH/issues/1328#issuecomment-2836264269

’Softcore’ scenario 1 with one room:

1. Choose a specific room in the hospital or build it. For example, the GP Office.
2. Prepare/Find/hire staff for this room.
3. Put this staff in this room.
4. Wait until any patient will enter this room or change dress near the dress screen, if the room has it. Set speed to 1.
5. Hire another staff and put it into this room before the patient begins to be served in that room.
6. Pick up the first staff who was in this room initially.
7. Put it outside of the room in the hospital free space.
8. Set speed to 5. Observe as the workflow in the room is disrupted.

Hardcore scenario 2 with two rooms:

1. Choose a specific room in the hospital or build it. For example Fracture Clinic. This is the First Room.
2. Choose any other room in the hospital or build it where the same staff can work. For example Pharmacy. This is the Second Room.
3. Prepare/Find/hire staff for the First Room. This is Primary Staff.
4. Prepare/Find/hire additional reserved staff for this room. This is the Secondary Staff.
5. Wait until any patient will enter this room or change dress near the dress screen, if the room has it.
6. Put Primary Staff in the First Room.
7. Wait until any patient from the queue will enter the First Room. Set speed to 1.
8. Grab the Secondary Staff and put it in the First Room.
9. Grab the Primary Staff from the First Room and put it in the Secondary Room. Patient from the queue will enter the Second Room too.
10. Set speed to 5. Observe as the workflow in the First Room/Second Room is disrupted. Observe visual glitches.

----

### Expected Behaviour

Room should serve patients as usual.
To avoid this bug patient can leave the room and re enter it again.

----

### Save Game

[Battenburg_level12_2021.sav.zip](https://github.com/user-attachments/files/21527486/Battenburg_level12_2021.sav.zip)

[ChrizCampaign_level7_2024.sav.zip](https://github.com/user-attachments/files/21527489/ChrizCampaign_level7_2024.sav.zip) 

----

### CorsixTH Version

0.68.0, 0.69.0

ARGAMX:
*Fixes #2950*

- Removed functionality of staff replacing after a patient has entered which was implemented here: https://github.com/CorsixTH/CorsixTH/commit/39b324b179c15aa3cc2531d338ce5756fe60032f (removed `waiting_on_other_staff` flag, and `waiting_staff_member` flag, and `dealt_patient_callback`)

### DIFF
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -1171,7 +1171,7 @@ end
 
 ----- BEGIN Save game compatibility -----
 -- These function are merely for save game compatibility.
--- For 0.69 gamesaves and below.
+-- For 0.69.x gamesaves and below.
 -- And they does not participate in the current game logic.
 local --[[persistable:room_dealt_with_patient_callback]] function _(staff_humanoid) end
 ----- END Save game compatibility -----


CorsixTH/Lua/room.lua
```diff
@@ -623,12 +632,8 @@ function Room:onHumanoidLeave(humanoid)
     end
   end
   if class.is(humanoid, Staff) then
-    if humanoid.waiting_on_other_staff then
-      humanoid.waiting_on_other_staff = nil
-      self.dealt_patient_callback = nil
-    end
     -- Make patients leave the room (except wards) if there are no longer enough staff
-    if not self:testStaffCriteria(self:getRequiredStaffCriteria()) then
+    if not self:testStaffCriteria(self:getRequiredStaffCriteria()) or self:getStaffMember() == nil then
```
ARGAMX, line 606: The patient will now leave the room also in case if no staff are assigned.

If the necessary staff is in the room, but they are not assigned to the room, this is no longer enough for patient to stay.

CorsixTH/Lua/room.lua
```diff
@@ -427,27 +434,27 @@ function Room:onHumanoidEnter(humanoid)
   end
 end
 
+----- BEGIN Save game compatibility -----
+-- These function are merely for save game compatibility.
+-- For 0.69 gamesaves and below.
+-- And they does not participate in the current game logic.
+-- Do not move or edit
```
lewri: The question here is why. It's likely an approach that relies on leaving a permanently orphaned function is not a good solution.

Is the issue that the `persistable` needs to exist? If so, you can make it a stub to the bottom of the file and delete what you no longer need. (ironically, doing the above but to a lesser extent, at least.)
Example: https://github.com/CorsixTH/CorsixTH/blob/e48ea317cbbb2381ee1e68c5cfec67819f2e495e/CorsixTH/Lua/dialogs/fullscreen/research_policy.lua#L51-L54
Wrt making stubs -- it should really be documented on this Wiki page but it's not: https://github.com/CorsixTH/CorsixTH/wiki/Persistence

If it's not, might need some more context what you're trying to achieve.

CorsixTH/Lua/rooms/training.lua
```diff
@@ -172,40 +172,35 @@ function TrainingRoom:commandEnteringStaff(humanoid)
   local profile = humanoid.profile
 
   if profile.humanoid_class == "Doctor" then
-    -- Consultants try to use the projector and/or skeleton
     if profile.is_consultant then
+      -- Consultant entered. Try to use the projector and/or skeleton
```
lewri: The line below only looks for the projector? Likely for training room, `entered/placed in room` might be more relevant.

CorsixTH/Lua/rooms/training.lua
```diff
@@ -172,40 +172,35 @@ function TrainingRoom:commandEnteringStaff(humanoid)
...
-          humanoid:walkTo(ox, oy)
-          self:doStaffUseCycle(humanoid)
-          self:setStaffMember(humanoid)
-        else
-          humanoid:setNextAction(self:createLeaveAction())
-          humanoid:queueAction(MeanderAction())
-        end
+        -- Leave room
+        humanoid:setNextAction(self:createLeaveAction())
+        humanoid:queueAction(MeanderAction())
```
lewri: This appears to be a change in behaviour, but I haven't loaded up the artefact to check.,
For the training room, if you place a staff member in a room it will call `onHumanoidEnter` (above)
https://github.com/CorsixTH/CorsixTH/blob/e48ea317cbbb2381ee1e68c5cfec67819f2e495e/CorsixTH/Lua/dialogs/place_staff.lua#L151-L156
In turn this then calls `commandEnteringStaff`. With this patch it appears you can't change the consultant by dropping them in (they will in turn immediately leave)

CorsixTH/Lua/rooms/training.lua
```diff
@@ -234,15 +229,6 @@ function TrainingRoom:onHumanoidLeave(humanoid)
...
 
-    if humanoid.profile.is_consultant and humanoid == self.staff_member then
-      local staff = self.waiting_staff_member
-      self:setStaffMember(nil)
-      if staff then
-        staff.waiting_on_other_staff = nil
-        self.waiting_staff_member = nil
-        self:commandEnteringStaff(staff)
-      end
-    end
```
lewri, line 245: I'm curious what purpose this tried to previously serve. Did it? Was it meant for rooms that weren't training?

CorsixTH/Lua/room.lua
```diff
@@ -427,27 +434,27 @@ function Room:onHumanoidEnter(humanoid)
...
-    self:commandEnteringStaff(staff_humanoid)
-    self:setStaffMember(staff_humanoid)
+    staff_humanoid:setNextAction(self:createLeaveAction())
+    staff_humanoid:queueAction(MeanderAction())
+    staff_humanoid:setMood("staff_wait", "deactivate")
+    staff_humanoid:setDynamicInfoText("")
     self.waiting_staff_member = nil
     self.dealt_patient_callback = nil
   end
-  self.waiting_staff_member = humanoid
```
lewri, line 450: This is the only point `waiting_staff_member` is set inside the room code. However, other parts of the file depend on checking this variable. One part of `gp.lua` also depends on this variable.

lewri:
To note: I understand generally the goal and good work.

CorsixTH/Lua/rooms/training.lua
ARGAMX: > The line below only looks for the projector? Likely for training room, `entered/placed in room` might be more relevant.

Yes, I also thought that there was some incorrect information about the skeleton here. But I didn't go into too much detail. I was just sure that the place where the comment was posted was incorrect. So I moved it.
But I like your suggestions.

CorsixTH/Lua/rooms/training.lua
ARGAMX: > I'm curious what purpose this tried to previously serve. Did it? Was it meant for rooms that weren't training?

I asked myself the same question when I saw this...

I think the purpose of it the similar as `function Room:createDealtWithPatientCallback(humanoid)` in `room.lua`.

But as you understand `createDealtWithPatientCallback` is never called in the training room.

To be more specific, after the main lecturer has left, the reserve lecturer must take on the role of the main one and the status of being a reserve must be reset.

CorsixTH/Lua/room.lua
ARGAMX: > This is the only point `waiting_staff_member` is set inside the room code. However, other parts of the file depend on checking this variable. One part of `gp.lua` also depends on this variable.

Oh, yes, that's right, thank you.
I didn't write in `gp.lua` that calling `self.dealt_patient_callback(self.waiting_staff_member)` is now also part of backward compatibility.
Corrected.

CorsixTH/Lua/room.lua
ARGAMX: > The question here is why.

Thanks, I was thinking about that too.

This wording was taken from here.
There was no detailed explanation why.
But it's better than nothing.
And I myself couldn’t come up with anything better.

https://github.com/CorsixTH/CorsixTH/blob/8a1de5ad0cad19e42dfb6e4f42ace915250b65c1/CorsixTH/Lua/rooms/operating_theatre.lua#L256-L261

I would gladly remove this function, as I myself do not like to keep such unused code.

But without it, saves of previous versions do not load.

I had to leave it for backward compatibility reasons.

I also thought about moving it to another part of the file.
But then in the change history it will be like my function written by me.
This whole fix was possible only because I was able to find the original commit in the GIT history via `blame/author`.

I wrote about it here.
https://github.com/CorsixTH/CorsixTH/issues/2950#issuecomment-3175034483
and moving this code means destroying the connection that helped me.
That's why I gave up on it.

This code is already strange. And its transfer will destroy part of its connections to the history that can explain more why it is here. 

Maybe we just need to leave a link to the original commit? But I haven't seen anyone do that with our code. That's why I didn't paste link and move it too.

So I don't think it's worth moving.

CorsixTH/Lua/rooms/training.lua
ARGAMX: > With this patch it appears you can't change the consultant by dropping them in (they will in turn immediately leave)

Well, yeah, true.
Seems that's a bug.
I'll think about it.
Thanks.

CorsixTH/Lua/room.lua
lewri: Ah no, it's likely your solution is just:

```lua
-- Stub for backwards compatibility
--[[persistable:room_dealt_with_patient_callback]] function (staff_humanoid) end
```
Move to end of file. The functional code inside the callback will already exist in the permanent objects table (at least in theory!)

Delete `Room:createDealtWithPatientCallback(humanoid)` function entirely.


This code that has been commented, I'd suggest saying less is more for your code documentation.

```lua
-- Saves <= 0.69.1 used patient callbacks.
-- Retain until 2028 release for backward compatibility.
 if self.dealt_patient_callback then
    self.dealt_patient_callback(self.waiting_staff_member)
  end
```

The retention aspect is optional, but I've started doing it with superfluous strings that have been since removed from the code to make sure we eventually clean up what is no longer required.

If those changes don't work, can you show the error it throws?

<sup>Persistables are a PITA.</sup>

lewri: The stub persistable might also need at end of line `-- luacheck: ignore 212` (must be on same line)

CorsixTH/Lua/room.lua
```diff
@@ -30,6 +30,11 @@ function Room:Room(x, y, w, h, id, room_info, world, hospital, door, door2)
   self.world = world
   self.hospital = hospital
 
+  -- Serving staff in single occupancy rooms (like GD, Pharmacy and etc)
```
lewri, line 33: GD or GP? (liekly either but GP might fit the line length)

ARGAMX: > GD or GP? (liekly either but GP might fit the line length)

Both. GP and GD.
I thought that GD is a more special case for better understanding.

CorsixTH/Lua/rooms/training.lua
ARGAMX: > In turn this then calls `commandEnteringStaff`. With this patch it appears you can't change the consultant by dropping them in (they will in turn immediately leave)

This resolved now. 👍

CorsixTH/Lua/room.lua
ARGAMX: > Ah no, it's likely your solution is just:
> 
> ```lua
> -- Stub for backwards compatibility
> --[[persistable:room_dealt_with_patient_callback]] function (staff_humanoid) end
> ```
> 
> Move to end of file. The functional code inside the callback will already exist in the permanent objects table (at least in theory!)
> 
> Delete `Room:createDealtWithPatientCallback(humanoid)` function entirely.
> 
> This code that has been commented, I'd suggest saying less is more for your code documentation.
> 
> If those changes don't work, can you show the error it throws?
>

The game won't start.

Screenshot 2025-09-20 at 20 23 47

lewri: Yes, I'm rather confused why the normal behaviour of stubbing isn't working. @TheCycoONE am I missing something, that prevents the normal approach of retiring a persistable?

CorsixTH/Lua/rooms/gp.lua
```diff
@@ -155,9 +155,15 @@ function GPRoom:dealtWithPatient(patient)
     patient:queueAction(IdleAction())
   end
 
+  ----- BEGIN Save game compatibility -----
+  -- These function call are merely for save game compatibility.
+  -- For 0.69 gamesaves and below.
+  -- And they does not participate in the current game logic.
+  -- Do not move or edit
   if self.dealt_patient_callback then
```
TheCycoONE: hmm - why? usually keeping persisted function definitions is required, but not calls?

CorsixTH/Lua/room.lua
TheCycoONE: The syntax is wrong.

Try:
```lua
local --[[persistable:room_dealt_with_patient_callback]] function _(staff_humanoid) end
```

TheCycoONE:
A lot of behavior was cut, reading the issue I see why. I expect someone will try to implement this again at some point. Fix up the persistence and it's probably good after some play testing.

TheCycoONE (changes requested):

CorsixTH/Lua/rooms/gp.lua
ARGAMX: > hmm - why? usually keeping persisted function definitions is required, but not calls?

Basically, we can remove this, thereby losing some backward compatibility.

If we remove this, then in those saved games where this staff replacement process is already in progress, the old doctor won't leave room after dealt with patient, but will remain in the room.

And the new doctor won't be able to take his place, but will simply meandering around in the room endlessly.

It looks like a bug, but it's not critical and doesn't crash the game. It's just one doctor wandering around aimlessly in the room. and until the player intervenes, this will never end.

I figured that if I didn't keep this here, I'd be asked to put it back to keep the backward compatibility.

In principle, I wouldn't mind removing it. But it's probably worth doing this somewhere no earlier than 2 releases in the future.

That's why I noted the version number.

CorsixTH/Lua/room.lua
ARGAMX: > The syntax is wrong.
> 
> Try:
> 
> ```lua
> local --[[persistable:room_dealt_with_patient_callback]] function _(staff_humanoid) end
> ```

Ok, thanks, that works.
But I get not the result I had expected:
I have tried this and get the same result as there:

https://github.com/CorsixTH/CorsixTH/pull/2971#discussion_r2384012021

Game will not crash with this `persistable` record. 
But old save will not be able to resolve ongoing swap perfectly. 

If that's exactly what we want, I can do it.

We are consciously not wanting described backward compatibility here? @lewri

CorsixTH/Lua/rooms/gp.lua
TheCycoONE: I would rather use afterLoad to fix state.

ARGAMX: > I would rather use afterLoad to fix state.

Make sense. I will make change with `afterLoad`.

ARGAMX: > I would rather use afterLoad to fix state.

`afterLoad` implemented 👍

CorsixTH/Lua/room.lua
```diff
@@ -1184,3 +1168,10 @@ function Room:calculateHappinessFactor()
 
   self.happiness_factor = window_factor + space_factor
 end
+
+----- BEGIN Save game compatibility -----
+-- These function are merely for save game compatibility.
+-- For 0.69 gamesaves and below.
```
lewri: 0.69.1
```suggestion
-- For 0.69.1 gamesaves and below.
```

Edit 🤔 Actually calling out is potentially a moving target. Maybe 0.69.x
```suggestion
-- For 0.69.x gamesaves and below.
```

lewri: This generally is a nit, as in our actual codebase we shouldn't be casually dropping the revision reference if we've now started using it in practice.

PR #2972, opened 2025-08-13
Update screenshot function

lewri:
*Fixes #*

- Updates Screenshots to be more date/time relationship rather than random numbers. This is more in line with *nix/Android behaviour. It also helps with our goal for better approachability to bug reports. Format `Screenshot_YYYYMMDD-HHMMSS.png`
- Removes the deprecated repeat loop as no longer needed
- Adds a blocker for > 1 screenshot per second.

### DIFF
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1106,16 +1106,20 @@ end
 function UI:tutorialStep(...)
 end
 
+--! Perform the Screenshot action (usually by key bind 'global_screenshot')
 function UI:makeScreenshot()
-   -- Find an index for screenshot which is not already used
-  local i = 0
-  local filename
-  repeat
-    filename = TheApp.screenshot_dir .. ("screenshot%i.png"):format(i)
-    i = i + 1
-  until lfs.attributes(filename, "size") == nil
-  print("Taking screenshot: " .. filename)
-  local res, err = self.app.video:takeScreenshot(filename) -- Take screenshot
+  -- Generate filename
+  local timestamp = os.date("%Y%m%d-%H%M%S")
+  local filename = TheApp.screenshot_dir .. ("Screenshot_%s.png"):format(timestamp)
+
+  -- It's very unlikely you intentionally want multiple screenshots a second
+  if lfs.attributes(filename, "size") ~= nil then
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/ui.lua
```diff
@@ -1106,16 +1106,20 @@ end
...
-  local res, err = self.app.video:takeScreenshot(filename) -- Take screenshot
+  -- Generate filename
+  local timestamp = os.date("%Y%m%d-%H%M%S")
+  local filename = TheApp.screenshot_dir .. ("Screenshot_%s.png"):format(timestamp)
+
+  -- It's very unlikely you intentionally want multiple screenshots a second
+  if lfs.attributes(filename, "size") ~= nil then
+    print("Screenshot failed: File already exists")
+    return
+  end
```
lewri, line 1119: The check isn't needed here if  Line 1122 already handles this error really.

Alberth289346 (approved):

Alberth289346:
There is probably still the problem of switching between summer and winter time, where you can have the same time twice in the same day.
I am not sure if it is relevant enough though.

If you want to fix it, including a notion of "summer time" may do the trick (not sure if you can get that). Adding a unique sequence number should work too, but that is somewhat complicated too.
If you have a unique number each day, that may be simpler. That however needs getting all screenshots filenames of today, and do some pattern matching to find the largest unique number.

lewri:
> There is probably still the problem of switching between summer and winter time, where you can have the same time twice in the same day. I am not sure if it is relevant enough though.
> 
> If you want to fix it, including a notion of "summer time" may do the trick (not sure if you can get that). Adding a unique sequence number should work too, but that is somewhat complicated too. If you have a unique number each day, that may be simpler. That however needs getting all screenshots filenames of today, and do some pattern matching to find the largest unique number.

I think a more elegant fix is adding an incrementing number e.g. `_1` at the end if it exists, but as you say, I don't think these have a relevant need to accommodate.

CorsixTH/Lua/ui.lua
Alberth289346: Deciding what to do here is the only thing left then. I'd be fine with removing it.

PR #2976, opened 2025-08-17
[Draft] Cleanup map-tile storage

Alberth289346:
Use a vector to store the map instead of an array.

Still has knowledge about tile order in storage at least, and probably a few other things can be improved too.

### DIFF
--- a/CorsixTH/Lua/api_version.lua
+++ b/CorsixTH/Lua/api_version.lua
@@ -32,4 +32,4 @@ Note: This file compiles as both Lua and C++. */
 
 #endif /*]] --*/
 
-return 2697;
+return 2698;
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -285,7 +285,7 @@ bool level_map::set_size(int iWidth, int iHeight) {
   delete[] original_cells;
   delete[] parcel_adjacency_matrix;
   delete[] purchasable_matrix;
-  _width = iWidth;
+  width = iWidth;
   height = iHeight;
   cells = nullptr;
   cells = new (std::nothrow) map_tile[iWidth * iHeight];
@@ -299,7 +299,7 @@ bool level_map::set_size(int iWidth, int iHeight) {
     delete[] original_cells;
     original_cells = nullptr;
     cells = nullptr;
-    _width = 0;
+    width = 0;
     height = 0;
     return false;
   }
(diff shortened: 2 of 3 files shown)


Alberth289346:
I want to have multiple single time animations from 1 `animation`, but ended up refactoring the map code instead.
While it's not my goal, I think at least some of the changes here are useful to keep.

I tried to move knowledge about about ordering of tiles in the vector into the `encode` and `decode` coordinate functions, but didn't succeed completely. Basically it would mean you always must iterate over `x` and `y`, unless you really don't care about the order of processing of tiles.

Please share your thoughts how to proceed.

TheCycoONE:
The loops where you emplace_back with no arguments n times can be replaced with resize(n)

Alberth289346:
> The loops where you emplace_back with no arguments n times can be replaced with resize(n)

I had plans to add a tile coordinate to the tiles, but currently not sure it's actually needed.

TheCycoONE:
The depersist class of level_map calls a placement new constructor over itself. I think that may leak the vectors because it instantiates a new object in place over the existing one without calling the destructors, even the default - it's pretty awful in my opinion.

Can you verify whether we're leaking memory on load now? If so it's the depersist I'd rather change.

Alberth289346:
Didn't check yet, but given that c++ allows
```
vector<X> *p = new vector<X>();
p->resize(10000);
```

means in my view that what-ever `p` points to cannot move (I might have setup `q` pointing to it too, etc), but the vector storage likely needs to re-allocated and thus might move.

In other words, I don't see how a vector can avoid having 2 storage locations at least.

TheCycoONE:
I checked, we're not constructing the same object twice, we are making a new level_map every time we load a game. We appear to be allocating uninitialized memory, casting it to a level_map, calling depersist on it, and doing placement construction on our this pointer from there.

My understanding is this is undefined behavior :frowning: 


```
level_map::level_map() 0x7d0ff59e1bf8 # called from level_map::depersist
Savegame changed from 218 (v0.69.0-beta2) to 230 (v0.69.1-dev230). The save was created using 194 (v0.68.0).
level_map::level_map() 0x7d0ff5e0bd78
Savegame changed from 218 (v0.69.0-beta2) to 230 (v0.69.1-dev230). The save was created using 194 (v0.68.0).
level_map::level_map() 0x7d0ff5d25378
Savegame changed from 218 (v0.69.0-beta2) to 230 (v0.69.1-dev230). The save was created using 194 (v0.68.0).
level_map::~level_map() 0x7d0ff5d25378
level_map::~level_map() 0x7d0ff5e0bd78
level_map::~level_map() 0x7d0ff59e1bf8
```

PR #2977, opened 2025-08-20
Late diagnosis termination bug fix.

### LINKED ISSUE #2974: [Bug] Late diagnosis termination bug
Image
<br/><br/>

### A little non-technical introduction

Image
<br/><br/>
There is a ‘Diagnosis termination’ slider in ‘Hospital Policy’ (F9).

By default it is set to 100% and allows you to complete diagnostics and send patients for treatment as early as possible.

But if we want to shake more money out of patients perhaps it is not in our interests to rush with diagnosis…
This is especially relevant at the beginning of the level, when there is little money and not many patients. 
There are no queues yet and you can torment them for a little longer.

So we can move this slider further to 150% or even 200% and no one will punish us for this... 
This way, patients will continue to undergo diagnostics even if they are no longer needed.
Which will bring more money to the hospital.

I like this setting, it makes you think about whether when we are sent in real life for expensive diagnostics to a single specific clinic, is this diagnostic really necessary at that moment or…
Thanks Bullfrog, they are geniuses.

### Describe the issue

So what went wrong in CTH?
‘Diagnosis termination’ 100% value works well in CTH.
But if we move this slider further, we will increase the probability that the patient will go home without being sent for treatment, even though the required minimum diagnostic threshold has been reached.

For example, you don't have extra diagnostic equipment for this type of disease. You only have one.
Then in CTH there is a high risk that the patient with this type of disease will go home after the first diagnosis room, since there is no more diagnose room for him.

But logically this shouldn't happen.
Deception is deception, but if a patient goes home not cured we do not receive part of the money at all and then there is no point in this function.

Let's see what the tooltip says about this function:

Image
<br/><br/>
It says:

<A patient's diagnosis will continue until Doctors are as certain as the STOP PROCEDURE percentage or until all the diagnosis machines have been tried>

As I understand here, if the doctor has not reached the required percentage, but there are no more machines for diagnostics, the diagnostics should be STOPPED. And after this STOP, the patient should go for treatment, not to home.

In the same situation with the same patient, if I put 200%, he will most likely go home. 
If I put 100%, he always goes for treatment.
If at 100% he always goes for treatment, then at 200% he should always go for treatment too.
But due to a mistake in CTH logic, this is not the correct case.

----

### Steps to Reproduce

https://github.com/user-attachments/assets/42c8bd1d-79d5-4781-9b96-6a2667936d57

1. Load savegame attached to this issue. The ‘Diagnosis termination’ slider is set to 200% in ‘Hospital Policy’ in this save.<br/>
2. There is a patient with transparency near the GP door at the bottom right.<br/>
3. You can press him and check his history to see that this is his second visit to GP. And now he is back to the GP after a Scanner diagnostic. There is a consultant doctor in the GP's office and this is more than enough to successfully diagnose transparency.<br/>
4. Press 5 and observe that the patient will pay for the GP and after that he is going home. Not for treatment.<br/>
5. If this experiment is repeated, the patient goes home with a 100% probability and is sent for treatment with a 0% probability.<br/>
6. Load savegame attached to this issue again.<br/>
7. Open ‘Hospital Policy’ (F9) and move the ‘Diagnosis termination’ slider left to 100%. Close ‘Hospital Policy’.<br/>
8. Press 5 and observe that the patient will pay for the GP and after that he is going to the Pharmacy.<br/>
9. If this experiment is repeated, the patient goes for treatment with a 100% probability and goes home with a 0% probability.<br/>

----

### Expected Behaviour

If the minimum diagnostic threshold has been reached for the patient, but there is no additional diagnostic equipment, the patient must be sent to treatment regardless of the ‘Diagnosis termination’ value.

----

### Save Game

[2974_GP_cant_diagnose_transparecny.sav.zip](https://github.com/user-attachments/files/21812943/2974_GP_cant_diagnose_transparecny.sav.zip)

----

### CorsixTH Version

0.68, 0.69

ARGAMX:
*Fixes #2974*

- Added earlier check for a GP that there are not enough diagnostic rooms for over-diagnosis. So if minimum diagnostic threshold has been reached the diagnosis should be set in such a case.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -334,7 +334,7 @@ function Patient:isTreatmentEffective()
   return (cure_chance >= math.random(1,100))
 end
 
-function Patient:isHaveMorePossibleDiagnosticRooms()
+function Patient:hasMoreDiagnosisRoomsAvailable()
   return #self.available_diagnosis_rooms ~= 0
 end
 
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -108,9 +108,9 @@ function Patient:setDisease(disease)
     self.available_diagnosis_rooms[i] = room
   end
   -- 25% of the patients pay via insurance
-  local insurance_chance = math.random(1,12)
-  if insurance_chance < 4 then
-    self.insurance_company = insurance_chance
+  if math.random(1,4) == 1 then
+    -- randomly select one of three insurance companies
+    self.insurance_company = math.random(1,3)
   end
   -- Randomise thirst and the need to visit the loo soon.
   -- Alien patients do not have the needed animations for these things, so exclude them
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -51,6 +51,9 @@ function Patient:Patient(...)
   self.vaccination_candidate = false
   -- Has the patient passed reception?
   self.has_passed_reception = false
+  -- Diagnostics progress to diagnose the patient.
+  -- Usually this value is in the range from 0 to ~2.5
+  self.diagnosis_progress = 0
 
   -- Is the patient trying to get to the toilet? ("yes", "no", or "no-toilets")
   self.going_to_toilet = "no"
@@ -104,9 +107,10 @@ function Patient:setDisease(disease)
   for i, room in ipairs(self.disease.diagnosis_rooms) do
     self.available_diagnosis_rooms[i] = room
   end
-  local company = math.random(1,12)
-  if company < 4 then
-    self.insurance_company = company
+  -- 25% of the patients pay via insurance
+  local insurance_chance = math.random(1,12)
+  if insurance_chance < 4 then
+    self.insurance_company = insurance_chance
   end
   -- Randomise thirst and the need to visit the loo soon.
   -- Alien patients do not have the needed animations for these things, so exclude them
@@ -328,6 +332,10 @@ function Patient:isTreatmentEffective()
   cure_chance = cure_chance + (service_base * service_factor)
 
   return (cure_chance >= math.random(1,100))
+end
+
+function Patient:isHaveMorePossibleDiagnosticRooms()
+  return not (#self.available_diagnosis_rooms == 0)
 end
 
 --! Change patient internal state to "cured".
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -335,7 +335,7 @@ function Patient:isTreatmentEffective()
 end
 
 function Patient:isHaveMorePossibleDiagnosticRooms()
-  return not (#self.available_diagnosis_rooms == 0)
+  return #self.available_diagnosis_rooms ~= 0
 end
 
 --! Change patient internal state to "cured".
--- a/CorsixTH/Lua/rooms/gp.lua
+++ b/CorsixTH/Lua/rooms/gp.lua
@@ -134,7 +134,11 @@ function GPRoom:dealtWithPatient(patient)
   elseif patient.disease and not patient.diagnosed then
     self.hospital:receiveMoneyForTreatment(patient)
     patient:completeDiagnosticStep(self)
-    if patient.diagnosis_progress >= self.hospital.policies["stop_procedure"] then
+    local no_need_diagnose_further =
+      patient.diagnosis_progress >= self.hospital.policies["stop_procedure"]
+    local cant_diagnose_further_but_can_set_diagnosis =
+      (patient.diagnosis_progress >= 1.0) and (not patient:isHaveMorePossibleDiagnosticRooms())
+    if no_need_diagnose_further or cant_diagnose_further_but_can_set_diagnosis then
       patient:setDiagnosed()
       if patient:agreesToPay(patient.disease.id) then
         patient:queueAction(SeekRoomAction(patient.disease.treatment_rooms[1]):enableTreatmentRoom())
@@ -164,7 +168,7 @@ function GPRoom:dealtWithPatient(patient)
 end
 
 function GPRoom:sendPatientToNextDiagnosisRoom(patient)
-  if #patient.available_diagnosis_rooms == 0 then
+  if not patient:isHaveMorePossibleDiagnosticRooms() then
     -- The very rare case where the patient has visited all his/her possible diagnosis rooms
     -- There's not much to do then... Send home
     patient:goHome("kicked")
--- a/CorsixTH/Lua/rooms/gp.lua
+++ b/CorsixTH/Lua/rooms/gp.lua
@@ -137,7 +137,7 @@ function GPRoom:dealtWithPatient(patient)
     local no_need_diagnose_further =
       patient.diagnosis_progress >= self.hospital.policies["stop_procedure"]
     local cant_diagnose_further_but_can_set_diagnosis =
-      (patient.diagnosis_progress >= 1.0) and (not patient:isHaveMorePossibleDiagnosticRooms())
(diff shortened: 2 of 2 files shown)


CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -51,6 +51,9 @@ function Patient:Patient(...)
   self.vaccination_candidate = false
   -- Has the patient passed reception?
   self.has_passed_reception = false
+  -- Diagnostics progress to diagnose the patient.
+  -- Usually this value is in the range from 0 to ~2.5
+  self.diagnosis_progress = 0
```
ARGAMX, line 56: Just refactoring which does not affect functionality.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -104,9 +107,10 @@ function Patient:setDisease(disease)
   for i, room in ipairs(self.disease.diagnosis_rooms) do
     self.available_diagnosis_rooms[i] = room
   end
-  local company = math.random(1,12)
-  if company < 4 then
-    self.insurance_company = company
+  -- 25% of the patients pay via insurance
+  local insurance_chance = math.random(1,12)
+  if insurance_chance < 4 then
+    self.insurance_company = insurance_chance
```
ARGAMX: Just refactoring which does not affect functionality

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -330,6 +334,10 @@ function Patient:isTreatmentEffective()
   return (cure_chance >= math.random(1,100))
 end
 
+function Patient:isHaveMorePossibleDiagnosticRooms()
```
lewri: Not sure on the best phrasing, though 'is' and 'have' are related words in this context. 
What about `hasMoreDiagnosisRoomsAvailable` or just replace the `isHave` with `has`

It's a bit hard as we don't want the function name too long. You could get fancy and have `areDiagnosticOptionsExhausted`. The trouble is that one doesn't convey that you might discover a new room.

lewri:
Other parts seem fine

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: > Not sure on the best phrasing, though 'is' and 'have' are related words in this context. What about `hasMoreDiagnosisRoomsAvailable` or just replace the `isHave` with `has`

Yes, thank you, I agree. 
I like `hasMoreDiagnosisRoomsAvailable`. 👍
I constantly have difficulties with English. 😀

Alberth289346: `moreDiagnosisPossible` ?

CorsixTH/Lua/entities/humanoids/patient.lua
TheCycoONE: Usually?

CorsixTH/Lua/entities/humanoids/patient.lua
TheCycoONE: 1/4rd chance of being insured by any of three companies; the same random number controlling both. In that sense both company and chance are appropriate names.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: > Usually

Usually 🙂

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: > 1/4rd chance of being insured by any of three companies; the same random number controlling both. In that sense both company and chance are appropriate names.

to me name `company` not looks like an integer variable.
`if company < 4` seems weird.

ARGAMX: probably shouldn't have used one value for two cases here. very confusing.

Alberth289346: `company_nr`

CorsixTH/Lua/entities/humanoids/patient.lua
lewri: I think the question is more, in what scenarios does this number go outside those bounds? 
The graphic related to stop procedure that should set the maximum diagnosis process means we *should* be capping at 200%
image

However, knowing CorsixTH I wouldn't be surprised if we don't hard clamp diagnosis progress at `<= 200%` for this value. But that's a separate issue. That is likely where this 'usually' is coming from.

CorsixTH/Lua/entities/humanoids/patient.lua
lewri: If we're struggling for a consensus maybe revert the name back, as the baseline soluton.
Maybe add a second comment line below the first `A result of 1-3 denotes which insurance company to use.`

Edit (a different nit is I prefer `<= 3` to `< 4` for clarity.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: > However, knowing CorsixTH I wouldn't be surprised if we don't hard clamp diagnosis progress at `<= 200%` for this value. But that's a separate issue. That is likely where this 'usually' is coming from.

Yes, logic don't cut `diagnosis_progress` value at `Hospital policy diagnosis termination` value.

`diagnosis_progress` value is patient diagnosis progress.
`Hospital policy diagnosis termination` value does not cut this value at certain threshold.
`Patient diagnosis` value can be grater that `Hospital policy diagnosis termination` value.

To understand, a simple example:

We have `Hospital policy diagnosis termination` = 1.0
Patient spawn. his `diagnosis_progress` = 0.0 
Patient progresses in GP. +0.48. his `diagnosis_progress` 0.0+0.48 = 0.48
Patient progresses in Scanner. +0.18. his `diagnosis_progress` 0.48+0.18 = 0.66
Patient progresses in GP. +0.42. his `diagnosis_progress` 0.66 + 0.42 = 1.08

now `diagnosis_progress` >= `Hospital policy diagnosis termination`.
at this point we can call `patient:setDiagnosed()`.

as you can see, the result is not 1.00, but 1.08

if you set `Hospital policy diagnosis termination` = 2.0
the same thing will happen after 2.0
we will have 2.08, and 2.15, and 2.45, and 2.52, and any values ​​up to some specific maximum value that I don't know. 🤷‍♂️

I think max value is in the range
`Hospital policy diagnosis termination` ... `Hospital policy diagnosis termination`+1.5
Maybe it's 3.2 or something like that. (Why 3.2 ????? lol)
I don't know.
In any case, this research was not included in this scope as it too tedious.

but without this comment it was not obvious at all that, for example, `diagnosis_progress` is not in the range from 0 to 100. nothing was said about this.
and I had to figure and ensure it out myself.

I agree that this comment can generate new questions.
but without it I had a question - what is stored in this value?
If such a comment had been made before me, I wouldn't have had to spend time on this.

Maybe it is possible to formulate this somehow so that it says what is needed and does not raise new questions?

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: > If we're struggling for a consensus maybe revert the name back, as the baseline soluton. Maybe add a second comment line below the first `A result of 1-3 denotes which insurance company to use.`
> 
> Edit (a different nit is I prefer `<= 3` to `< 4` for clarity.

I think the number `12` should not appear here at all.
it is a natural magic number here.
12 is 3 * 4.

It is become obvious after TheCycoONE comment above.
So I'll probably rewrite this.

ARGAMX:
```lua
  -- 25% of the patients pay via insurance
  if math.random(1,4) == 1 then
    -- randomly select one of three insurance companies
    self.insurance_company = math.random(1,3)
  end
  ```
  
I think this variant is much more transparent.
And without using intermediate variables, for which we had a hard time finding a suitable name. 🙂

TheCycoONE:
Any reason this is still RFC?

ARGAMX:
> Any reason this is still RFC?

Probably no

PR #2980, opened 2025-08-25
Remove target memory layout assumptions from CorsixTH resource loading

### LINKED ISSUE #2983: [Bug] CorsixTH cannot run on s390x (big endian)
### Describe the issue

Immediately on launch:
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==5205==ERROR: AddressSanitizer: SEGV on unknown address 0x52209a070000 (pc 0x02aa337a63be bp 0x03ffd24f8348 sp 0x03ffd24f8348 T0)
==5205==The signal is caused by a UNKNOWN memory access.
    #0 0x2aa337a63be in chunk_renderer::decode_chunks(unsigned char const*, int, bool) (/home/stephen/CorsixTH/build/dev/CorsixTH/corsix-th+0x5263be) (BuildId: 88443102836b8a60e7c4dcd09229bae7fe6a389f)
    #1 0x2aa3383bb8b in sprite_sheet::load_from_th_file(unsigned char const*, unsigned long, unsigned char const*, unsigned long, bool, render_target*) (/home/stephen/CorsixTH/build/dev/CorsixTH/corsix-th+0x5bbb8b) (BuildId: 88443102836b8a60e7c4dcd09229bae7fe6a389f)
    #2 0x2aa3365da53 in (anonymous namespace)::l_spritesheet_load(lua_State*) (/home/stephen/CorsixTH/build/dev/CorsixTH/corsix-th+0x3dda53) (BuildId: 88443102836b8a60e7c4dcd09229bae7fe6a389f)
    #3 0x3ffaa392eb1  (/lib/s390x-linux-gnu/liblua5.4.so.0+0x12eb1) (BuildId: dbdde232f9bd9cc03aeb19f800c940eea2b2a6ce)
    #4 0x3ffaa3a4933  (/lib/s390x-linux-gnu/liblua5.4.so.0+0x24933) (BuildId: dbdde232f9bd9cc03aeb19f800c940eea2b2a6ce)
    #5 0x3ffaa3931b5  (/lib/s390x-linux-gnu/liblua5.4.so.0+0x131b5) (BuildId: dbdde232f9bd9cc03aeb19f800c940eea2b2a6ce)
    #6 0x3ffaa38bf77 in lua_callk (/lib/s390x-linux-gnu/liblua5.4.so.0+0xbf77) (BuildId: dbdde232f9bd9cc03aeb19f800c940eea2b2a6ce)
    #7 0x2aa335f99b3 in lua_main(lua_State*) (/home/stephen/CorsixTH/build/dev/CorsixTH/corsix-th+0x3799b3) (BuildId: 88443102836b8a60e7c4dcd09229bae7fe6a389f)
    #8 0x3ffaa392eb1  (/lib/s390x-linux-gnu/liblua5.4.so.0+0x12eb1) (BuildId: dbdde232f9bd9cc03aeb19f800c940eea2b2a6ce)
    #9 0x3ffaa39319d  (/lib/s390x-linux-gnu/liblua5.4.so.0+0x1319d) (BuildId: dbdde232f9bd9cc03aeb19f800c940eea2b2a6ce)
    #10 0x3ffaa391aed  (/lib/s390x-linux-gnu/liblua5.4.so.0+0x11aed) (BuildId: dbdde232f9bd9cc03aeb19f800c940eea2b2a6ce)
    #11 0x3ffaa3935b1  (/lib/s390x-linux-gnu/liblua5.4.so.0+0x135b1) (BuildId: dbdde232f9bd9cc03aeb19f800c940eea2b2a6ce)
    #12 0x3ffaa38c075 in lua_pcallk (/lib/s390x-linux-gnu/liblua5.4.so.0+0xc075) (BuildId: dbdde232f9bd9cc03aeb19f800c940eea2b2a6ce)
    #13 0x2aa335fdb41 in main (/home/stephen/CorsixTH/build/dev/CorsixTH/corsix-th+0x37db41) (BuildId: 88443102836b8a60e7c4dcd09229bae7fe6a389f)
    #14 0x3ffa602c9c9  (/lib/s390x-linux-gnu/libc.so.6+0x2c9c9) (BuildId: dab82547e89e6c245d83627bfce415cff7bb4f61)
    #15 0x3ffa602cac7 in __libc_start_main (/lib/s390x-linux-gnu/libc.so.6+0x2cac7) (BuildId: dab82547e89e6c245d83627bfce415cff7bb4f61)
    #16 0x2aa335f890f  (/home/stephen/CorsixTH/build/dev/CorsixTH/corsix-th+0x37890f) (BuildId: 88443102836b8a60e7c4dcd09229bae7fe6a389f)
```

### Steps to Reproduce

1. Create an s390x vm in qemu
2. Install debian
3. Install CorsixTH dependencies
4. Build CorsixTH with linux-dev preset
5. Install tigervnc
6. Launch tigervnc with xterm
7. Connect to vnc from host
8. Run corsix-th

### Expected Behaviour

No crash

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

### CorsixTH Version

76cc7bbb

### Operating System

Debian Trixie

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

### Additional Information

_No response_

TheCycoONE:
Should improve compatibility with additional architectures and compilers.

The pragma pack structures and reinterpret_cast usage assumed that unaligned access was allowed and that the target architecture was little endian.

Fixes #2983 

Screenshot_20250827_211309
Screenshot_20250827_211435

AnimView still heavily uses packed structures for loading and is out of scope of this PR.


Alberth289346:
I agree that deleting the pragmas is a good idea.

Having constants for offsets, I am not sure. The offsets are not going to ever change, and now there is no single point in the code where both field sizes and offsets are near each other.
You cannot easily see what bytes in a structure are skipped or read (and not eg used twice).

In my Java code I took a simpler approach, based on the idea that these numbers are never going to change.
``` java
    public static RawFrame newFrame(byte[] data, int offset, int listCount, int frameCount) {
        int dataOffset = 0;
        if (!GraphicsUtils.canRead32(data, offset + dataOffset)) return null;
        int element_list_index = GraphicsUtils.read32(data, offset + dataOffset);
        if (element_list_index >= listCount) element_list_index = 0;
        dataOffset += 4;

        dataOffset += 2; // Skip over width and height.

        if (!GraphicsUtils.canRead8(data, offset + dataOffset)) return null;
        int sound = GraphicsUtils.read8(data, offset + dataOffset);
        dataOffset++;

        if (!GraphicsUtils.canRead8(data, offset + dataOffset)) return null;
        byte flags = (byte)GraphicsUtils.read8(data, offset + dataOffset);
        dataOffset++;

        if (!GraphicsUtils.canRead16(data, offset + dataOffset)) return null;
        int nextFrame = GraphicsUtils.read16(data, offset + dataOffset);
        if (nextFrame >= frameCount) nextFrame = 0;
        dataOffset += 2;

        Common.check(dataOffset == SIZE);
        return new RawFrame(element_list_index, sound, flags, nextFrame);
    }
```
Easy to see that the amount of read data and the `dataOffset` increments match. `SIZE` is not shown here, but it's `public static final int SIZE = 10;`. Not sure why it's a constant anymore, I'd have to check. 

----
Pointing to the spec is good, but that is a moving target. Assuming you didn't take my new additions into account here, the docs and the code currently disagree for animations iirc.

TheCycoONE:
I'm not attached to the constexpr offsets, feels like a nit.

The specs are probably pretty rigid at this point; we don't actually know animation index size. We know that there are two bytes that are definitely the little endian index, followed by two bytes of 0s, so it could be uint16/uint16 (always 0), or it could be uint32, but it makes no practical difference.

Alberth289346:
> The specs are probably pretty rigid at this point; we don't actually know animation index size. We know that there are two bytes that are definitely the little endian index, followed by two bytes of 0s, so it could be uint16/uint16 (always 0), or it could be uint32, but it makes no practical difference.

There is much more structure in the animation start frames than CTH uses.

I used 32 bit in my Java code so you get consecutive animation numbering instead of all multiples of 2.

Reading the spec again, https://github.com/CorsixTH/theme-hospital-spec/blob/master/format-specification.md#start  I wrote that an animation is 8 byte long.
The second 4 bytes of the 8 were weird I remember.

At the time I didn't know, but I checked what most animations were showing, and I found it has groups of 2 or 4 "the same". Some were in a sense mirrored, while others were really the same. Eventually, I figured it out. It's the same animation in 2 or 4 directions. 

I now wonder if the "weird" 4 bytes contain the hflag somewhere. It would make sense to encode that in the file I think.

I am not sure it's a good idea to make these changes in CTH, as a lot is going to fall over.

TheCycoONE:
I'll revert the 4 byte thing, and maybe make some of the other suggested changes. Notably this PR does fix the last remaining endian problems / opens up CorsixTH to more systems. (AnimView not included)

TheCycoONE:
16 -> 32 bit change has been reverted.

Switched structs to namespaces where they were just a collection of consts.

I don't agree that it's clearer not to have the consts so I'm not changing that. I will point out they are defined directly above the function they are used.

TheCycoONE:
I found a compromise I like better; see if you prefer that too.

changelog.txt
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------------------
+CorsixTH 0.70.0 - ???
+-------------------------------------------------------------------------------
+# Packager Notes
```
lewri, line 4: As a nit, not sure your need this right now, as this will be documented in the project board and made at release time.

It'd be nice to find a more visible way though for people who want to try it on a big endian system though from master.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -153,15 +153,55 @@ void animation_manager::set_sprite_sheet(sprite_sheet* pSpriteSheet) {
...
+constexpr size_t bytes_per_animation_property = 4;
+
+// Frame information structure reinterpreted from Theme Hospital data.
+// https://github.com/CorsixTH/theme-hospital-spec/blob/master/format-specification.md#frame
+struct th_frame_properties {
+  constexpr static size_t size = 10;
+
+  explicit th_frame_properties(const uint8_t* pData)
+      : list_index{bytes_to_uint32_le(pData)},
+        sound{pData[6]},
```
Alberth289346, line 166:
```suggestion
        // Skipping width and height (2 bytes).
        sound{pData[6]},
```

CorsixTH/Src/th.cpp
```diff
@@ -158,9 +159,10 @@ th_string_list::th_string_list(const uint8_t* data, size_t length) {
   uint8_t* sDataOut = string_buffer.data();
   sections.resize(iSectionCount);
   for (size_t i = 0; i < iSectionCount; ++i) {
-    size_t section_size = reinterpret_cast<const uint16_t*>(data)[i + 1];
-    sections[i].reserve(section_size);
-    for (size_t j = 0; j < section_size; ++j) {
+    size_t sectionSizeOffset = (i + 1) * 2;
+    size_t sectionSize = bytes_to_uint16_le(data + sectionSizeOffset);
```
Alberth289346, line 165:
```diff
-    size_t sectionSizeOffset = (i + 1) * 2;
-    size_t sectionSize = bytes_to_uint16_le(data + sectionSizeOffset);
+    size_t sectionSizeOffset = i * 2;
+    // The 2 bytes have been read already as iSection count above.
+    size_t sectionSize = bytes_to_uint16_le(data + 2 + sectionSizeOffset);
```
Trying to get rid of the need for much brain power to grasp where the off-by-one is coming from.

Suggestion button was making a mess, so I changed it to a diff.

CorsixTH/Src/th.cpp
```diff
@@ -158,9 +159,10 @@ th_string_list::th_string_list(const uint8_t* data, size_t length) {
...
+    sections[i].reserve(sectionSize);
```
Alberth289346, line 172:
```suggestion
    // Read the data of the section.
    sections[i].reserve(sectionSize);
```

Alberth289346: Comment may seem silly and trivial, but I found it makes a lot of difference if you are new to the code or didn't look at it for 10 years.

CorsixTH/Src/th_gfx.cpp
```diff
@@ -242,22 +281,21 @@ bool animation_manager::load_from_th_file(
...
+                                          i * th_element_properties::size);
 
     element oElement;
-    oElement.sprite = pTHElement->table_position / 6;
-    oElement.flags = pTHElement->flags & 0xF;
-    oElement.x = static_cast<int>(pTHElement->offx) - 141;
-    oElement.y = static_cast<int>(pTHElement->offy) - 186;
-    oElement.layer = static_cast<uint8_t>(
-        pTHElement->flags >> 4);  // High nibble, layer of the element.
+    oElement.sprite = thElement.table_position / 6;
```
Alberth289346: `6` is actually `th_sprite_properties::size` ?

`// Size of a sprite.` works too of course.

Alberth289346:
> I found a compromise I like better; see if you prefer that too.

Converting in the constructor is a good idea.

CorsixTH/Src/th_gfx.cpp
TheCycoONE: That predated my change and I didn't know what it was. That makes enough sense to change

TheCycoONE:
I think I made all of the suggested changes.

CorsixTH/Src/th_gfx.cpp
Alberth289346: I verified in my Java code. It looks like
``` java
    public static RawElement newElement(byte[] data, int offset, SpriteSheet sheet) {
        int dataOffset = 0;
        if (!GraphicsUtils.canRead16(data, offset + dataOffset)) return null;
        int spriteNumber = GraphicsUtils.read16(data, offset + dataOffset);
        spriteNumber /= 6; // Likely position / size, XXX Sprite.SIZE ??
        dataOffset += 2;
```

Apparently, I concluded that at some earlier point in time as well.

Link4Electronics:
Thanks for bringing big endian support, I understand it's already merged but just to confirm that it works on a PowerMac G5 PPC64 with a Radeon R5 230
Screenshot_2025-09-10_10-15-13
Only thing missing is music that fluidsynth spams on console complaining about SMPTE.
EDIT: ended up finding a bug, when trying to put a door on the blueprint, it crashes, will open an issue about it later.
Best regards,
Link.

TheCycoONE:
Spiffy. Thanks for testing, I'll look into your crash.

PR #2985, opened 2025-08-30
Minor cleanup load_multi_buffer

TheCycoONE:
This is a small one; just fixes the encapsulation, improves the comments, and I hope makes it slightly more readable.

### DIFF
--- a/CorsixTH/Src/persist_lua.cpp
+++ b/CorsixTH/Src/persist_lua.cpp
@@ -73,42 +73,77 @@ int l_crude_gc(lua_State* L) {
 
 }  // namespace
 
+constexpr size_t load_multi_buffer_capacity = 3;
+
 //! Structure for loading multiple strings as a Lua chunk, avoiding
 //! concatenation
 /*!
-    luaL_loadbuffer() is a good way to load a string as a Lua chunk. If there
+   luaL_loadbuffer() is a good way to load a string as a Lua chunk. If there
    are several strings which need to be concatenated before being loaded, then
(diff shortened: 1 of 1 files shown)


Alberth289346:
I am missing the loop, but I am guessing the load function is called 3 times.

The function is plain magic, as it doesn't state what it is supposed to be doing.

In the old version, my first guess would be `// Load all strings in the multi buffer.` (since it would have a reason to loop over the array, right?

In the new version seems to `// Append a loaded string into the multi buffer.`. Two completely different things. Now if the function would state what it's supposed to do, I could judge correctness in this fragment.

Currently, I would need to dig around in all cpp code, find calls to this function, and check all of them for how they are used. Hopefully that gives enough information to judge the purpose, but no guarantees.

Obviously, the latter is going to take ages compared with the former. This is how documentation effectively saves time.

----
Why not rename `s` and `i` to something more descriptive? Currently the names in themselves are pretty much meaningless.

TheCycoONE:
The comment for the class says exactly what it does though you have to understand lua_load first and if you don'tit won't make sense.

That loop was misleading, it never looped, it's a callback, the loop is internal to lua_load.

TheCycoONE:
Yeah, I think you misunderstood the old logic, which is understandable.

I added the loop for skipping empty chunks back in though in my tests it's never used. Hopefully the new code is much clearer to the reader though.

Also incorporated other suggestions.

Alberth289346 (approved):
Nice!

Alberth289346:
Can't see what was changed, but it LGTM.

PR #2987, opened 2025-08-30
Store the tile position of an animation in the animation.

Alberth289346:
Instead of afterwards re-constructing the animation position from the map, store it in the animation itself.

Commits:
- First commit does renaming of the existing pixel position offset.
- Second commit introduces storage of the animation tile.
- Update api.
- Reformat cpp.

With respect to the pixel position offset, the docs claim that it is updated in the `tick`, but that could have been written afterwards (possibly even by me).
- The `animation` actually does that.
- The `render_sprite_list` however does it both in `tick` and in `draw`.
  That looks wrong, since if a sprite is ever active (didn't check but I am guessing we currently don't have that at all), then movement is twice as fast as it should be.
- Morph target updating also looks fishy at first sight, but didn't check either.

### DIFF
--- a/CorsixTH/Src/th_gfx.h
+++ b/CorsixTH/Src/th_gfx.h
@@ -524,21 +524,20 @@ class animation_base : public drawable {
   void attach_to_tile(map_tile* pMapNode, int layer);
 
   uint32_t get_flags() const { return flags; }
-  int get_x() const { return x_relative_to_tile; }
-  int get_y() const { return y_relative_to_tile; }
+  int get_x() const { return pixel_offset.x; }
+  int get_y() const { return pixel_offset.y; }
 
   void set_flags(uint32_t iFlags) { flags = iFlags; }
-  void set_position(int iX, int iY) {
-    x_relative_to_tile = iX, y_relative_to_tile = iY;
+  void set_position(int x, int y) {
+    pixel_offset.x = x;
+    pixel_offset.y = y;
   }
   void set_layer(int iLayer, int iId);
   void set_layers_from(const animation_base* pSrc) { layers = pSrc->layers; }
 
  protected:
-  //! X position on tile (not tile x-index)
-  int x_relative_to_tile{};
-  //! Y position on tile (not tile y-index)
-  int y_relative_to_tile{};
+  //! Offset in pixels relative to the center of the animation tile.
+  xy_pair pixel_offset{0, 0};
 
   ::layers layers{};
 };
--- a/CorsixTH/Src/th_gfx.h
+++ b/CorsixTH/Src/th_gfx.h
@@ -532,7 +532,7 @@ class animation_base : public drawable {
     tile.x = x;
     tile.y = y;
   }
-  void set_position(int x, int y) {
+  void set_pixel_offset(int x, int y) {
     pixel_offset.x = x;
     pixel_offset.y = y;
   }
--- a/CorsixTH/Src/th_gfx.h
+++ b/CorsixTH/Src/th_gfx.h
@@ -521,13 +521,17 @@ class animation_base : public drawable {
   animation_base();
 
   void remove_from_tile();
-  void attach_to_tile(map_tile* pMapNode, int layer);
+  void attach_to_tile(int x, int y, map_tile* node, int layer);
 
   uint32_t get_flags() const { return flags; }
-  int get_x() const { return pixel_offset.x; }
-  int get_y() const { return pixel_offset.y; }
+  const xy_pair &get_pixel_offset() const { return pixel_offset; }
+  const xy_pair &get_tile() const { return tile; }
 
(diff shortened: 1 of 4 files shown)


CorsixTH/Src/th_gfx.h
```diff
@@ -518,21 +518,32 @@ class animation_base : public drawable {
...
-  //! X position on tile (not tile x-index)
-  int x_relative_to_tile{};
-  //! Y position on tile (not tile y-index)
-  int y_relative_to_tile{};
+  //! X tile position. Negative value indicates the animation is not active.
+  int x_tile{-1};
+  //! Y tile position. Negative value indicates the animation is not active.
+  int y_tile{-1};
+  //! X pixel position relative to animation tile center.
+  int x_tile_pixels{};
```
TheCycoONE: I find these names less clear than the old ones, but they seem to serve the same purpose (x_tile_pixels seems like it could be the width, or the position of the tile).

Maybe
```
struct pixel_offset {int x{}, int y{}};
...
pixel_offset tile_center_offset;
```

TheCycoONE: Or pixel_diff, like the ptrdiff_t type

CorsixTH/Src/th_lua_anims.cpp
```diff
@@ -422,25 +424,17 @@ int l_anim_get_tile(lua_State* L) {
   if (lua_isnil(L, 2)) {
     return 0;
   }
-  level_map* pMap = (level_map*)lua_touserdata(L, 2);
-  const link_list* pListNode = pAnimation->get_previous();
-  while (pListNode->prev) {
-    pListNode = pListNode->prev;
+  lua_touserdata(L, 2);  // Not used, but returned from the call.
```
TheCycoONE: We don't need to call lua_touserdata if we don't use it, it has no side effects - consider it a [[nodiscard]] result (but C doesn't have that concept)

CorsixTH/Src/th_gfx.h
Alberth289346: We have a few x/y diffs iirc, I'll have a look tomorrow.

TheCycoONE:
I didn't notice any compatibility changes to the lua API? Did I miss it?

Alberth289346:
> I didn't notice any compatibility changes to the lua API? Did I miss it?

Aside from not crashing on non-active animations (and returning a nil-pair), there is no change in the api indeed.

Last time I had that however, you wanted to have a bump anyway. So here I thought it to be useful to do that as well.


Btw, should we have a change description here as well, to avoid missing an update if cpp modifications get merged at around the same time?

TheCycoONE:
> Btw, should we have a change description here as well, to avoid missing an update if cpp modifications get merged at around the same time?

Probably not a bad idea, though it hasn't come up yet. The lua/c++ compatible comment syntax is annoying but doable.

CorsixTH/Src/th_lua_anims.cpp
```diff
@@ -512,21 +503,23 @@ int l_anim_get_flag(lua_State* L) {
...
 
 template <typename T>
-int l_anim_set_position(lua_State* L) {
+int l_anim_set_pixel_offset(lua_State* L) {
   T* pAnimation = luaT_testuserdata<T>(L);
 
-  pAnimation->set_position(static_cast<int>(luaL_checkinteger(L, 2)),
-                           static_cast<int>(luaL_checkinteger(L, 3)));
+  int x = static_cast<int>(luaL_optinteger(L, 2, 0));
+  int y = static_cast<int>(luaL_optinteger(L, 3, 0));
```
Alberth289346:
```suggestion
  int x = static_cast<int>(luaL_optinteger(L, 2));
  int y = static_cast<int>(luaL_optinteger(L, 3));
```

CorsixTH/Src/th_gfx.h
Alberth289346: It became `pixel_offset` (with comment about tile center), and
`tile` (for the tile position in the map).

CorsixTH/Src/th_lua_anims.cpp
Alberth289346: dropped

Alberth289346:
> Last time I had that however, you wanted to have a bump anyway. So here I thought it to be useful to do that as well.

Dropped the api increment.

TheCycoONE:
LGTM

CorsixTH/Src/th_gfx.h
```diff
@@ -510,38 +510,43 @@ class animation_manager {
   void fix_next_frame(uint32_t iFirst, size_t iLength);
 };
 
+// Integer 2D pair.
+struct xy_pair {
```
TheCycoONE, line 514: I have a future vision for different structs for different coordinate systems so they can't be mixed up. This is fine now, only used for pixel offsets current.

PR #2988, opened 2025-08-30
Don't allow patient entry if staff want rest

### LINKED ISSUE #2982: GP. Doctor abandons patient
### Describe the issue

In a GP office a doctor can leave a patient alone.
In this particular case the Doctor has reached a given level of fatigue and decided to go to the staff room.

https://github.com/user-attachments/assets/8224953d-cd79-4cf5-a2e0-f31014f2e0ef

----

### Steps to Reproduce

1. Load savegame attached to the issue.
2. Press 4.
3. Observe how a Doctor `#1 PETLET` in a GP abandons a patient.

----

### Expected Behaviour

The doctor must finish serving the current patient and only after that leave the GP.

----

### Save Game

[2982_GP_doctor_leaves_gp_in_middle_of_appointment.v69.sav.zip](https://github.com/user-attachments/files/21993248/2982_GP_doctor_leaves_gp_in_middle_of_appointment.v69.sav.zip)

----

### CorsixTH Version

0.69.0

----

lewri:
*Fixes #2982*

- Fixes the issue where staff wanting to go to the staff room shouldn't be considered as available to the room.
- This is due to some use_object actions (e.g. GP desk) being a prolonged cycle, which can let a patient enter and then gets abandoned when the staff member leaves.

### DIFF
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -251,7 +251,8 @@ function Room:getMissingStaff(criteria)
       -- check if answering a call to another room
       if class.is(humanoid, Staff) and humanoid:fulfillsCriterion(attribute) and
           not humanoid:isLeaving() and not humanoid.fired and
-          not (humanoid.on_call and humanoid.on_call.object ~= self) then
+          not (humanoid.on_call and humanoid.on_call.object ~= self) and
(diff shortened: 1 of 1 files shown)


ARGAMX (approved):
Nice 👍

PR #2989, opened 2025-08-31
Remove debug falling flag

lewri:
- Remove all debug_falling flags, with anticipated permanent go-live for 0.70.0 release
- Sets a probability roll (50%) on eligible patients falling in an earthquake (previously, it was 100%)
- Updates issue template

Changes supporting falling actions: #1848

### DIFF
--- a/CorsixTH/Lua/earthquake.lua
+++ b/CorsixTH/Lua/earthquake.lua
@@ -149,7 +149,7 @@ function Earthquake:tick()
   local hospital = self.world:getLocalPlayerHospital()
   -- loop through the patients and allow the possibility for them to fall over
   for _, patient in ipairs(hospital.patients) do
-    if not patient.in_room and patient.falling_anim then
+    if not patient.in_room and patient.falling_anim and math.random(1,2) == 2 then
       patient:falling(false)
     end
   end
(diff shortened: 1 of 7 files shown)


lewri:
Forgot to push 3rd commit, oops

CorsixTH/Lua/earthquake.lua
```diff
@@ -145,15 +145,11 @@ function Earthquake:tick()
...
-  -- It is currently disabled except for debugging purposes in the config file.
-  -- Current behaviour can cause empty action queues or other undesired behaviours.
-  -- Once working, the debugging flag can be removed.
-  if not TheApp.config.debug_falling then return end
+  -- A patient might fall over during an earthquake.
   local hospital = self.world:getLocalPlayerHospital()
   -- loop through the patients and allow the possibility for them to fall over
   for _, patient in ipairs(hospital.patients) do
-    if not patient.in_room and patient.falling_anim then
+    if not patient.in_room and patient.falling_anim and math.random(1,2) == 2 then
```
TheCycoONE, line 152: I suspect this is still too high, though I guess you can test it.

As I recall this code runs multiple times during the earthquake more for larger ones.

lewri: Triggers once per game day, The random value here is untested yet (I will), but iirc in the original sometimes the same patient did fall more than once.

lewri: Wait is it? The code in earthquake makes it look like faster speeds will trigger a damage phase in less ticks.

lewri: > Wait is it? The code in earthquake makes it look like faster speeds will trigger a damage phase in less ticks.

No they don't. But it feels a bit peculiar how we do it. Every 16 'hours' out of 50.

PR #2990, opened 2025-08-31
Avoid calling members on uninitialized objects

TheCycoONE:
When __depersist is called, in most circumstances the object has not yet be initialized (i.e. by calling new), it is simply a block of malloc'd memory the right size.

This patch fixes everywhere that was calling methods or accessing members without calling placement new first.

I'm using void* and _ud as conventions to help future developers not to accidentally access the object too early; for lack of better enforcement in the language.

layers is zero initialized by the constructor which removes the need for the memset.

### DIFF
--- a/CorsixTH/Src/persist_lua.cpp
+++ b/CorsixTH/Src/persist_lua.cpp
@@ -918,6 +918,12 @@ class lua_persist_basic_reader : public lua_persist_reader {
           if (lua_isnil(L, -1))
             lua_pop(L, 1);
           else {
+            // Call the __depersist function with the userdata.
+            // Note: Unless the __pre_depersist method was called above the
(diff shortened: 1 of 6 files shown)


CorsixTH/Src/persist_lua.cpp
```diff
@@ -918,6 +918,12 @@ class lua_persist_basic_reader : public lua_persist_reader {
           if (lua_isnil(L, -1))
             lua_pop(L, 1);
           else {
+            // Call the __depersist function with the userdata.
+            // Note: Unless the __pre_depersist method was called above the
```
Alberth289346:
```suggestion
            // Note: Unless the __pre_depersist method was called above, the
```

Alberth289346 (approved):

Issue #2778, opened 2025-02-12
[Feature Request] Cheat Code to Make Machines Invulnerable to Damage


CommonSenseGuy:
### What feature or enhancement would you like to add?

Pretty self explanatory, if some sort of flag (the cheat) is set to true then all machines that can sustain damage during earthquakes won't receive any damage at all.

I guess a cheat to disable earthquakes would be an alternate version that does the same thing (unless someone likes the earthquakes, but doesn't want any damage taking place?)

### How would this feature benefit the gameplay of CorsixTH?

Just for fun

### Was the feature in the original game?

No

### What version of CorsixTH are you using?

0.68

### Does this feature/enhancement relate to a current problem?

No

lewri:
With the amount of possible cheat options; the window may need redesigning again into a scrollbar/dropdown/pages; especially given our minimum window size.

MandrakeHorse:
I've been thinking about how both of these cheats could be implemented.

### Repair All Machines
1. Fetch all the machines that exist in the hospital (we should already be doing something like this for the Machines menu - check `/Lua/dialogs/resizables/machine_menu.lua`)
2. Create a for loop that goes through the list of machines.
3. For each machine, set the Times Used to 0 and clear all "needs repairing" flags (i.e. room marked as needs repairing, smoke animation, broken gear icon, any queued repair jobs).

We can't simply call [`machineRepaired()`](https://github.com/CorsixTH/CorsixTH/blob/cc647c3de58ad15bb7667af6529d00fbd841ea2b/CorsixTH/Lua/entities/machine.lua#L346C1-L356C1) in `/Lua/entities/machine.lua`  on every machine because this method also calls `reduceStrengthOnRepair()`, and we don't want to potentially reduce each machine's strength by 1. So we will need to either create a separate method that's nearly identical except for this line, or update this existing method so that `reduceStrengthOnRepair()` does not get called if the user has just clicked the Repair All Machines cheat button.

### **Machines Never Break**
This could be a toggle which sets every machine's Remaining Uses to a sufficiently high value while preventing it from ever decreasing (i.e. [`getRemainingUses()`](https://github.com/CorsixTH/CorsixTH/blob/cc647c3de58ad15bb7667af6529d00fbd841ea2b/CorsixTH/Lua/entities/machine.lua#L62) always returns 9999) when enabled. I personally prefer this approach, since players can still track how many patients have been treated by each machine. However, this introduces a few concerns: 
[...]

PR #2998, opened 2025-09-10
Add cheat to repair all machines

ShiroAka:
Linked to issue #2778 

Added a new cheat to repair all the machines in a hospital, and made some refactors to make it easy to get the hospital machines (based on some code found in the machine menu module).

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/machine_menu.lua
+++ b/CorsixTH/Lua/dialogs/resizables/machine_menu.lua
@@ -281,9 +281,9 @@ function UIMachineMenu:update()
   self.machine_list = {}
   local world = self.ui.app.world
   local dispatcher = world.dispatcher
-  local playerMachines = world:getPlayerMachines()
+  local machines = world:getPlayerMachines()
 
-  for _, machine in ipairs(playerMachines) do
+  for _, machine in ipairs(machines) do
     if not machine:getRoom().crashed then
       local assigned_handyman
       local repair_call = dispatcher.call_queue[machine]


CorsixTH/Lua/entities/machine.lua
```diff
@@ -345,13 +345,16 @@ end
...
   room.needs_repair = nil
-  self:reduceStrengthOnRepair(room)
   self.times_used = 0
   self:setRepairing(nil)
   setSmoke(self, false)
   self:removeHandymanRepairTask()
+
+  if dontReduceStrength == nil or not dontReduceStrength then
+    self:reduceStrengthOnRepair(room)
+  end
```
```suggestion
  if not dontReduceStrength then
    self:reduceStrengthOnRepair(room)
  end
```

I believe in Lua you can just `if not dontReduceStrength then`. The result will be the same.

ShiroAka: Interesting. Never used Lua before, but some things seem to work just like JS.

CorsixTH/Lua/dialogs/resizables/machine_menu.lua
```diff
@@ -281,24 +281,17 @@ function UIMachineMenu:update()
...
-        if not machine:getRoom().crashed then
-          local assigned_handyman
-          local repair_call = dispatcher.call_queue[machine]
-          if repair_call then
-            assigned_handyman = repair_call["repair"].assigned
-          end
-          local machine_for_list = machineForList(machine, assigned_handyman)
-          table.insert(self.machine_list, machine_for_list)
-        end
+  local playerMachines = world:getPlayerMachines()
```
lewri: The localisation isn't strictly necessary, the function name from world is also clear enough.

CorsixTH/Lua/entities/machine.lua
```diff
@@ -345,13 +345,16 @@ end
...
+function Machine:machineRepaired(room, dontReduceStrength)
   room.needs_repair = nil
-  self:reduceStrengthOnRepair(room)
   self.times_used = 0
   self:setRepairing(nil)
   setSmoke(self, false)
   self:removeHandymanRepairTask()
+
+  if not dontReduceStrength then
+    self:reduceStrengthOnRepair(room)
```
lewri: A nit that you don't have to fix, but the `room` argument in the `self:reduceStrengthOnRepair` function is no longer required at all and could be dropped out (line 372).

TheCycoONE:
Thanks @ShiroAka

PR #3001, opened 2025-09-10
Change Queue Position Change Behaviour

### LINKED ISSUE #1646: Patients in queue change seating when next patient is to enter room.
#### Describe the issue
I actually believe this was a bug even in the original to be honest (but maybe not as pronounced).
When you have a group of patients queueing for a room; if the next patient is called forward it can cause all other patients in the queue to get up and sit back down in a new seat.

I really don't understand what sense it makes to do that.

#### Steps to Reproduce
1. Load any game which has a room with a decent sized queue.

2. Watch when a patient leaves the room, everyone in the queue gets up.

3. Patients go back and sit somewhere else.

#### Expected Behavior
Patients to stay in their seats until they are called to enter the room.

#### System Information
**CorsixTH Version:** Trunk 139 (and 0.64 rc1)

**Operating System:** W10

### Additional Info
This could just also be a mask though... the patients could all be being shifted to the GP Office next along in the area.... But even then that's odd.
https://www.dropbox.com/s/2tl01f999ub5oys/rrbcy-0q3ed.webm?dl=0

lewri:
*Fixes #1646*

- Changes the tolerances for finding a bench in relation to the door/desk (i.e. goal) the patient wants to use (mostly refactoring labels and steps).
- Prevents a patient unnecessarily moving bench if they're already close to the goal (I believe this is what caused loads of people to move at once).

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/queue.lua
+++ b/CorsixTH/Lua/humanoid_actions/queue.lua
@@ -234,6 +234,8 @@ local action_queue_on_change_position = permanent"action_queue_on_change_positio
       error("Patient in a queue has no calculated distance from themselves to their target door/desk!")
     end
 
+    -- Look for a free bench that is closer to the goal. If we are standing
+    -- expand the search radius to increase the chance of getting a seat.
     local bench_look_distance = not action:isStanding() and
         math.max(2 * distance_from_goal / 3, DISTANCE_THRESHOLD) or MAX_DISTANCE
 


CorsixTH/Lua/humanoid_actions/queue.lua
```diff
@@ -215,23 +215,38 @@ local action_queue_on_change_position = permanent"action_queue_on_change_positio
...
+        action.current_bench_distance <= DISTANCE_THRESHOLD then
+      return
+    end
+    local hum_x, hum_y = humanoid.tile_x, humanoid.tile_y
+    local goal_x, goal_y = action.x, action.y
+
+    local distance_from_goal = humanoid.world:getPathDistance(hum_x, hum_y,
+                                                              goal_x, goal_y)
+    if not distance_from_goal then
+      error("Patient in a queue has no caclulated distance from themselves to their target door/desk!")
```
lewri: On some rigorous testing I don't believe the code should ever fall into this state at least.

CorsixTH/Lua/humanoid_actions/queue.lua
```diff
@@ -215,23 +215,38 @@ local action_queue_on_change_position = permanent"action_queue_on_change_positio
...
+    local goal_x, goal_y = action.x, action.y
+
+    local distance_from_goal = humanoid.world:getPathDistance(hum_x, hum_y,
+                                                              goal_x, goal_y)
+    if not distance_from_goal then
+      error("Patient in a queue has no calculated distance from themselves to their target door/desk!")
     end
-    local bench, bx, by, dist = humanoid.world:getFreeBench(action.x, action.y, bench_max_distance)
+
+    local find_closer_bench = 2 * distance_from_goal / 3
```
ARGAMX: `find_closer_bench` - what is this variable responsible for?
i'm just a little confused. 😇
"find_closer_bench" - sounds like a function name.
(verbs are more suitable for naming functions)

lewri: In truth, there is no useful name I can think of without turning becoming way too long. I was merely trying to avoid a bit of math magic being unlabelled ( `x / 2` became ` 2 * x / 3`).

I think I've made this localised variable without benefit and it's not necessary at all, so may delete it.

lewri: Yep, deleting the local label.

CorsixTH/Lua/humanoid_actions/queue.lua
```diff
@@ -215,23 +215,37 @@ local action_queue_on_change_position = permanent"action_queue_on_change_positio
...
+
+    local distance_from_goal = humanoid.world:getPathDistance(hum_x, hum_y,
+                                                              goal_x, goal_y)
+    if not distance_from_goal then
+      error("Patient in a queue has no calculated distance from themselves to their target door/desk!")
     end
-    local bench, bx, by, dist = humanoid.world:getFreeBench(action.x, action.y, bench_max_distance)
+
+    local bench_look_distance = not action:isStanding() and
+        math.max(2 * distance_from_goal / 3, DISTANCE_THRESHOLD) or MAX_DISTANCE
```
TheCycoONE, line 240: why 2/3rds?

lewri: With the changes to having a threshold, the patient now sometimes will barely get to at all to change closer if it's only 1/2

TheCycoONE: Maybe a comment so people are less afraid to tweak it.

lewri: I can make a `local modifier = 2 / 3` but might be redundant. Comment also fine though.

Alberth289346: It may help to express the idea:

```
-- Looking for a free bench. If the humanoid is further away from the goal of the queue, expand the radius of the search.
-- In that way, the queue is spread out to many nearby benches.
```

CorsixTH/Lua/humanoid_actions/queue.lua
```diff
@@ -215,23 +215,37 @@ local action_queue_on_change_position = permanent"action_queue_on_change_positio
...
+    local goal_x, goal_y = action.x, action.y
+
+    local distance_from_goal = humanoid.world:getPathDistance(hum_x, hum_y,
+                                                              goal_x, goal_y)
+    if not distance_from_goal then
+      error("Patient in a queue has no calculated distance from themselves to their target door/desk!")
     end
-    local bench, bx, by, dist = humanoid.world:getFreeBench(action.x, action.y, bench_max_distance)
+
+    local bench_look_distance = not action:isStanding() and
```
lewri:
```suggestion

    -- Look for a free bench that is closer to the goal. If we are standing
    -- expand the search radius to increase the chance of getting a seat.
    local bench_look_distance = not action:isStanding() and
```

CorsixTH/Lua/humanoid_actions/queue.lua
lewri: Something akin to Albert's suggestion has been added.

PR #3005, opened 2025-09-13
Restore tile of animations while loading the map.

Alberth289346:
Set the tile of the animations while loading the map.

### DIFF
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -1545,6 +1545,18 @@ void level_map::persist(lua_persist_writer* pWriter) const {
   oEncoder.pump_output(pWriter);
 }
 
+namespace {
+//! Assign the tile position in the map to the just loaded animations.
+void restore_map_position(animation_base* anim, int tile_x_pos,
+                          int tile_y_pos) {
+  while (anim != nullptr) {
+    anim->set_tile(tile_x_pos, tile_y_pos);
+    anim = dynamic_cast<animation_base*>(anim->next);
+  }
+}
+
+}  // namespace
+
 void level_map::depersist(lua_persist_reader* pReader) {
   lua_State* L = pReader->get_stack();
   int iWidth, iHeight;
@@ -1600,8 +1612,17 @@ void level_map::depersist(lua_persist_reader* pReader) {
     }
   }
 
+  int tile_x_pos = -1;
+  int tile_y_pos = 0;
   for (map_tile *pNode = cells, *pLimitNode = cells + width * height;
        pNode != pLimitNode; ++pNode) {
+    // Update tile 2D position.
+    tile_x_pos++;
+    if (tile_x_pos == width) {
+      tile_x_pos = 0;
+      tile_y_pos++;
+    }
+
     uint32_t f;
     if (!pReader->read_uint(f)) return;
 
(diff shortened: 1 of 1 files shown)


Alberth289346:
While the code in map restoring seems to handle only one animation, it's actually a chain. I printed the animations that it restores (address and position), and got output like
```
0x57b7d89751e8 at tile (68, 68)
0x57b7d8974bb8 at tile (61, 69), 0x57b7d8974588 at tile (61, 69)
0x57b7d8975e48 at tile (62, 69)
0x57b7d897abd8 at tile (66, 69)
0x57b7d8972cc8 at tile (74, 69)
0x57b7d8975818 at tile (75, 69)
0x57b7d89732f8 at tile (62, 70), 0x57b7d8973f58 at tile (62, 70), 0x57b7d8973928 at tile (62, 70)
0x57b7d897b208 at tile (64, 70)
0x57b7d897a5a8 at tile (68, 71)
```

Alberth289346:
To understand what it does exactly, I printed what it sets. A fragment:
```
0x57b7d762e818 at tile (68, 47)                 -- Not in picture
0x57b7d733fe38 at tile (61, 52)  -- Radiator
0x57b7d69a7d98 at tile (63, 52)  -- Machine
0x57b7d69a8198 at tile (65, 52)  -- Radiator
0x57b7d78717b8 at tile (58, 53)                 -- Not in picture
0x57b7d698d8b8 at tile (62, 54)  -- Radiator
0x57b7d840aff8 at tile (58, 55), 0x57b7d69811b8 at tile (58, 55)  -- Not in picture
0x57b7d71c07b8 at tile (62, 55)                 -- Not in picture
0x57b7d7873e38 at tile (64, 55)  -- Door
0x57b7d62f1dd8 at tile (58, 56)                 -- Not in picture
0x57b7d62ed6f8 at tile (60, 56)                 -- Not in picture
0x57b7d6b1f0b8 at tile (63, 56)                 -- Not in picture
```
In-game:

Image

Comparing positions, I added a black dot to the tiles where the code set the map position.

Note that walls are not in this list.
That makes sense, since they are in the 4 layers (eg `tile_layer::north_wall`).

I think the old code never properly worked, because the layer animations are not in in the `link_list` base class. It smells like UB to me.

TheCycoONE:
Why would they need to be in the base class? I believe I fixed all the UB when I made everything load it's original class from lua and use dynamic_cast in c++ when using the base class?

CorsixTH/Src/th_map.cpp
```diff
@@ -1545,6 +1545,18 @@ void level_map::persist(lua_persist_writer* pWriter) const {
   oEncoder.pump_output(pWriter);
 }
 
+namespace {
+//! Assign the tile position in the map to the juts loaded animations.
+void restore_map_position(animation_base* anim, int tile_x_pos,
+                          int tile_y_pos) {
+  while (anim != nullptr) {
+    anim->set_tile(tile_x_pos, tile_y_pos);
+    anim = static_cast<animation_base*>(anim->next);
```
TheCycoONE: Use a dynamic_cast here.

Alberth289346:
> Why would they need to be in the base class? I believe I fixed all the UB when I made everything load it's original class from lua and use dynamic_cast in c++ when using the base class?

``` c++
struct map_tile {
  map_tile();
  ~map_tile() = default;

  link_list entities;
  link_list oEarlyEntities;
  uint16_t tile_layers[tile_layer::num_tile_layers];
  // Other fields omitted.
};

map_tile[...] the_map;
```
I don't know exactly what the wall Lua object points at when `getTile` gets executed, but at best it's `&map[x].tile_layers[north_wall]`. Then you basically computed `int tile_index = static_cast<tile>(&map[x].tile_layers[north_wall]) - &map[0];` in the previous implementation. The first address is not at the start of the `map[x]` tile.

A worse case is if the wall in Lua is something outside the map array. I have no doubt it will compute something, but it's just garbage.

CorsixTH/Src/th_map.cpp
Alberth289346: Done

PR #3006, opened 2025-09-13
Fix bug where the repair machine icon would not show

### LINKED ISSUE #2949: [Bug] Paging Handyman to repair machine does not visibly mark it for repair
### Describe the issue

When the player calls a handyman to fix a machine (by clicking the Page Handyman button in the machine's popup window), the machine is not being visibly marked for repair. The animated Machine Due For Repair icon does not appear above the machine, and the machine's name is not surrounded by brackets in the popup. There is nothing to indicate that the machine is currently in the repair queue, even after a Handyman starts making his way towards the room to fix it.

Image

### Steps to Reproduce

1. Click on any machine to open the machine's popup window
2. Click the Page Handyman button

### Expected Behaviour

In Theme Hospital, when the user clicks the Page Handyman button for a machine:

1. The Machine Due For Repair icon appears above the machine (an animated gear that repeatedly splits into halves and separates)
2. The machine's name becomes surrounded by brackets in the popup window.

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

### CorsixTH Version

0.68.0, 0.69.0

### Operating System

Windows 10

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

### Additional Information

_No response_

ShiroAka:
*Fixes #2949 *

- Fixes an issue where the repair machine icon will not show despite clicking the "Page Handyman" button

### DIFF
--- a/CorsixTH/Lua/calls_dispatcher.lua
+++ b/CorsixTH/Lua/calls_dispatcher.lua
@@ -112,7 +112,7 @@ function CallsDispatcher:callForRepair(object, urgent, manual, lock_room)
     dropped = nil
   }
 
-  object:setRepairingMode(lock_room and true or false)
+  object:setRepairingMode(lock_room and true or false, true)
 
   if not manual and urgent then
     object.hospital:giveAdvice({_A.warnings.machines_falling_apart})
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -346,7 +346,7 @@ end
 --! Call on machine repaired.
 --!param room (object) machine room
 function Machine:machineRepaired(room)
-  room.needs_repair = nil
+  room.needs_repair = false
   self:reduceStrengthOnRepair(room)
   self.times_used = 0
   self:setRepairing(nil)
@@ -383,13 +383,14 @@ end
 --! Tells the machine to start showing the icon that it needs repair.
 --!   also lock the room from patient entering
 --!param handyman The handyman heading to this machine. nil if repairing is finished
-function Machine:setRepairing(handyman)
-  -- If mode is set to true manually through the dialog, boost the urgency despite of the strength
+--!param is_manual_repair (bool) true if the repairing mode was set manually through the dialog; nil otherwise
+function Machine:setRepairing(handyman, is_manual_repair)
   local anim = {icon = 4564} -- The only icon for machinery
   local room = self:getRoom()
-  self:setMoodInfo(handyman and anim or nil)
-  room.needs_repair = handyman
-  if handyman then
+  local should_repair = handyman or is_manual_repair
+  self:setMoodInfo(should_repair and anim or nil)
+  room.needs_repair = should_repair
+  if should_repair then
     self.ticks = true
   else
     self.ticks = self.object_type.ticks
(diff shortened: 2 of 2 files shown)


ShiroAka:
One thing I noted when trying to fix this is that when the machine is at full strength, clicking the button does nothing. So I will keep testing to see if I can fix that issue as well, but the main bug should be fixed

lewri:
> One thing I noted when trying to fix this is that when the machine is at full strength, clicking the button does nothing. So I will keep testing to see if I can fix that issue as well, but the main bug should be fixed

Arguably, if the machine is already at full strength (new or repaired) there is no need for a handyman call. You could even go as far as making the button disabled in this state (not sure if there's actually a graphic for it though).

ShiroAka:
> > One thing I noted when trying to fix this is that when the machine is at full strength, clicking the button does nothing. So I will keep testing to see if I can fix that issue as well, but the main bug should be fixed
> 
> Arguably, if the machine is already at full strength (new or repaired) there is no need for a handyman call. You could even go as far as making the button disabled in this state (not sure if there's actually a graphic for it though).

Yeah, you're right, but I have some memories of being able to do that in the original game (I may be mistaken) and that's why I wanted to check that.

Don't think there's a graphic for the disabled button (I will check later), but maybe the image could be turned into a grey scale with some coding (even better if those buttons are in bmp or some easy to handle format).

ARGAMX:
> > > One thing I noted when trying to fix this is that when the machine is at full strength, clicking the button does nothing. So I will keep testing to see if I can fix that issue as well, but the main bug should be fixed
> > 
> > 
> > Arguably, if the machine is already at full strength (new or repaired) there is no need for a handyman call. You could even go as far as making the button disabled in this state (not sure if there's actually a graphic for it though).
> 
> Yeah, you're right, but I have some memories of being able to do that in the original game (I may be mistaken) and that's why I wanted to check that.
> 
> Don't think there's a graphic for the disabled button (I will check later), but maybe the image could be turned into a grey scale with some coding (even better if those buttons are in bmp or some easy to handle format).

I have just checked original game and yes, you can call handyman to repair full strength machine.

But from my point of view, this is absolutely pointless and I don't see any sense in such functionality.

Therefore, I think that if the handyman is not called to repair an already fully functional machine (and gear will not appear), this will be quite valid behavior.

I'm not even sure that we need to change the graphics for the button. If nothing happens in this case - in general, it's acceptable in my opinion.

But if it will be very easy to do, to disable the button - then why not. But it seems to me that it is not so easy. 🙂

ARGAMX (approved):
I have tested this change. Works smoothly for me. 👍

MandrakeHorse:
Thanks for fixing this bug. Would it be possible to surround the machine's name with brackets in the popup window and machine list as well?

ShiroAka:
> Thanks for fixing this bug. Would it be possible to surround the machine's name with brackets in the popup window and machine list as well?

The machine name should be surrounded by parentheses in the dialog when set to repair.
I think doing the same in the machine list could be possible, but I would do it in another PR.

Alberth289346:
> I have just checked original game and yes, you can call handyman to repair full strength machine.
> 
> But from my point of view, this is absolutely pointless and I don't see any sense in such functionality.

While the machine doesn't improve its strength, repairing has the side-effect that max strength is decreased.
By not repairing, you eliminate this side-effect in this case.

In other words, in the original game you needed to be careful when to repair the machine, as otherwise its strength would decrease. In CTH this is now impossible to do.

ARGAMX:
> > I have just checked original game and yes, you can call handyman to repair full strength machine.
> > But from my point of view, this is absolutely pointless and I don't see any sense in such functionality.
> 
> While the machine doesn't improve its strength, repairing has the side-effect that max strength is decreased. By not repairing, you eliminate this side-effect in this case.
> 
> In other words, in the original game you needed to be careful when to repair the machine, as otherwise its strength would decrease. In CTH this is now impossible to do.

If the machine was in perfect at maximum strength, then no, the repair will not reduce its maximum strength.
neither in TH nor in CTH.

Reduce maximum strength in that case is equal to 0%.

my very first PR here was dedicated to researching this thing and bringing it into line with TH. #2648

Therefore, I cannot say that any side-effect exists at all in this scenario. Just a waste of the handyman's time.

PR #3013, opened 2025-09-15
Add screen shake configuration option

### LINKED ISSUE #3007: Option to disable screenshake during Earthquakes

Whenever an Earthquake occurs, the entire screen will shake. This can cause motion sickness or headaches for some players. Although Earthquakes can be completely disabled using cheats, this will technically make the campaign easier.

We should provide an option that disables screenshake during Earthquakes. While this would remove the main visual indicator that an Earthquake is taking place, the sound cues will remain (the receptionist's announcement, loud rumbling in the background).

Players have the option to disable screenshake during Earthquakes.

No

0.69

No

### Additional Information

The ability to disable screenshake is a common accessibility guideline (for example: https://learn.microsoft.com/en-us/gaming/accessibility/xbox-accessibility-guidelines/117)

MandrakeHorse:
*Fixes #3007*

- Adds a configuration option that allows the player to disable screen shake during earthquakes.
- Appears in Custom Settings menu
- Enabled by default

### DIFF
--- a/CorsixTH/Lua/config_finder.lua
+++ b/CorsixTH/Lua/config_finder.lua
@@ -125,6 +125,7 @@ local config_defaults = {
   enable_avg_contents = false,
   remove_destroyed_rooms = false,
   machine_menu_button = true,
+  enable_screen_shake = true,
   audio_frequency = 22050,
   audio_channels = 2,
   audio_buffer_size = 2048,
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -965,11 +965,15 @@ function GameUI:scrollMap(dx, dy)
   self.screen_offset_y = floor(dy + 0.5)
 end
 
---! Start shaking the screen, e.g. an earthquake effect
+--! Start shaking the screen, e.g. an earthquake effect (unless disabled in config)
 --!param intensity (number) The magnitude of the effect, between 0 for no
 -- movement to 1 for significant shaking.
 function GameUI:beginShakeScreen(intensity)
-  self.shake_screen_intensity = intensity
+  if self.app.config.enable_screen_shake then
+    self.shake_screen_intensity = intensity
+  else
+    self.shake_screen_intensity = 0
+  end
 end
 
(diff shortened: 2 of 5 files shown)


CorsixTH/Lua/config_finder.lua
```diff
@@ -125,6 +125,7 @@ local config_defaults = {
   enable_avg_contents = false,
   remove_destroyed_rooms = false,
   machine_menu_button = true,
+  enable_screen_shake = true,
```
lewri, line 128: You'll need to add the text you made in the config template also here in the config finder. You can mostly copy paste an existing setting and adapt it as needed.

MandrakeHorse: I've added `enable_screen_shake` to `string_01`.

CorsixTH/Lua/game_ui.lua
```diff
@@ -965,11 +965,15 @@ function GameUI:scrollMap(dx, dy)
...
+--! Start shaking the screen, e.g. an earthquake effect (unless disabled in config)
 --!param intensity (number) The magnitude of the effect, between 0 for no
 -- movement to 1 for significant shaking.
 function GameUI:beginShakeScreen(intensity)
-  self.shake_screen_intensity = intensity
+  if self.app.config.enable_screen_shake then
+    self.shake_screen_intensity = intensity
+  else
+    self.shake_screen_intensity = 0
+  end
```
MandrakeHorse, line 976: I've coded this section so that the screen will only shake if `enable_screen_shake` is set to `true`. The alternative would have been to keep the screen still if `enable_screen_shake` is set to `false` and to let it shake otherwise. I picked the first approach for simplicity, although it does mean that the screen won't shake by default (for example, if the player has manually edited the configuration file and deleted or accidentally set this variable to something like "truu")

lewri: In lua, any value except `false` or `nil` is true. Your layout is the standard approach. Though `self.screen_shake_intensity = self.app.config.enable_screen_shake and intensity or 0` is also valid.

Issue #2467, opened 2023-12-28
[Bug] UIInformation has first line paragraph indent


lewri:
### Describe the issue

Indentation is occurring on UIInformation text in at least English at the first line. Tested in Toby's tutorial PR (2457) and 0.66

Edit: Not present in dutch. It's likely something to do with how `//` gets processed in original strings if it's at the start of a string.

### Steps to Reproduce

1. Start a new campaign with tutorial
2. Notice UIInformation text get indented.

### Expected Behaviour

The text should all align together.

### Save Game

Save game attached to report: No

### CorsixTH Version

0.67 (PR2457) and 0.66

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Additional Information

![image](https://github.com/CorsixTH/CorsixTH/assets/20030128/5c438a97-a1c9-4d4b-a68f-c916dbd3d80a)

lewri:
Original game does not indent.
![image](https://github.com/CorsixTH/CorsixTH/assets/20030128/4d71bd62-8693-4c15-b077-1d2aac5ce3a1)

PR #2068, opened 2021-10-10
Add duplication checker for Adviser

lewri:
*Fixes #2067*

**Describe what the proposed change does**
- Messages sent to the adviser are now checked for duplicates already queued. Unfortunately if the same message is already up we skim over that one.
- Obeys the override request as it clears the queue anyway (out of scope here but that sounds like bad behaviour)

PR #3016, opened 2025-09-17
Advisor message history window

### LINKED ISSUE #530: history of adviser messages screen
_From [maartenp...@hotmail.nl](https://code.google.com/u/103280305005036074093/) on March 14, 2013 16:54:26_

What steps will reproduce the problem? 1.Open a screen like Town Map in a very low or a very high resolution.
2.The advisor message dissapears behind the screen when the screen is placed where the message pops up.

The advisor message should be shown ''over'' the screen, otherwise users will miss important information.

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1532_

ShiroAka:
*Fixes #530 *

- Adds a new dialog which lets the user see a history of the adviser's messages, which displays latest messages first
- The new dialog can be opened via the "Charts" menu, and also by using the hotkeys (by default CTRL + H)

image

As of now, the dialog lets users delete single messages (by clicking the red button next to each message), or clear the whole history (by clicking the header above the red buttons).


Matroftt:
Add a screenshot of this menu and tell how it could be opened in "describe what the proposed change does"


Also, are you sure I made this menu? :)

ShiroAka:
> Also, are you sure I made this menu? :)

Oops, sorry about that. Honestly I copied the machine menu code and didn't pay attention to the copyright part

Alberth289346:
Align the top of the messages list with the top of the 'x' at the right?

With respect to messages:
- What happens if the same message appears again?
- Do messages auto-expire after some time?
- Would it be useful to know when a message appeared?

ARGAMX:
Sorry, I couldn't resist.
I couldn't resist spamming the window with my favorite (well, no) message in the game.

Screenshot 2025-09-17 at 17 49 02

ARGAMX:
I also noticed that after save and loading the game all messages disappear.
(Why should they disappear after load?)

Screenshot 2025-09-17 at 17 59 36

ShiroAka:
> Align the top of the messages list with the top of the 'x' at the right?
> 
> With respect to messages:
> 
> * What happens if the same message appears again?
> * Do messages auto-expire after some time?
> * Would it be useful to know when a message appeared?

As seen in ARGAMX's screenshot, messages repeat without limits. While reading the comments here I just came up with the idea to bring a message to the top of the list if was already in it.
But for that I should implement a way to identify the messages in the code (ideally without having to compare strings, which is already done in some other parts of the code).

Alternatively I could still let messages repeat, but avoid adding it to the history if a similar message was added X time ago (this would only fix the issue in ARGAMX's screenshot)

There is no message expiration; I don't see the benefit of adding that.
There's only a max. amount of messages that can be added to the history (I set it to 20, but it's hardcoded) and, if the limit is reached, the oldest messages are replaced.

About knowing when a message appeared, if implemented correctly, it may be useful. Perhaps I could add another row showing how many seconds (or minutes) ago was the message sent.
The thing is that in order to add that info I should make the table a bit wider (some didn't like this idea) or increase the row height (which would add a lot of free space for most messages and wouldn't look nice, IMO)

ARGAMX:
> About knowing when a message appeared, if implemented correctly, it may be useful. Perhaps I could add another row showing how many seconds (or minutes) ago was the message sent.
> The thing is that in order to add that info I should make the table a bit wider (some didn't like this idea) or increase the row height (which would add a lot of free space for most messages and wouldn't look nice, IMO)

The message history window is already quite functional even without the message posting time.
But if we add the time, then given that time in the game flows at different speeds depending on the selected speed, how relevant will that time be?
In my opinion, in this case, it makes sense to display only the in-game date.

Screenshot 2025-09-18 at 09 20 35

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,181 @@
...
+local col_bg = { red = 60, green = 174, blue = 201, }
+local col_highlight = { red = 47, green = 154, blue = 190, }
+local col_shadow = { red = 36, green = 138, blue = 158 }
+local col_delete_button = { red = 219, green = 36, blue = 36 }
+
+-- Timer, used to update the dialog only once per X ticks to reduce the burden.
+local ticks_to_skip = 5
+local tick_timer = ticks_to_skip
+
+local row_height = 40
```
lewri: It might be possible to make your row heights dynamic with some help from the draw functions on text (I think UIInformation and UIConfirmation do this). But I need to look into this patch deeper as well.

TheCycoONE:
I wouldn't deduplicate, this is the real notification history.

ShiroAka:
> I wouldn't deduplicate, this is the real notification history.

But maybe I should filter certain messages that aren't very useful, like the errors when placing staff.

ShiroAka:
> The message history window is already quite functional even without the message posting time. But if we add the time, then given that time in the game flows at different speeds depending on the selected speed, how relevant will that time be? In my opinion, in this case, it makes sense to display only the in-game date.

In a ledger it makes sense to display the in-game date because it refers to what's actually happening in the simulation, but for the notifications I'm not so sure. 

The way I see it, the notifications are more related to general advice or results of your direct actions, and are something that could impact your actions directly (instead of a being a direct result of the simulation). So the focus is not on the simulation per-se, but on the user. That's why I suggested displaying how many real seconds/minutes ago did the notification appear, instead of tracking that based on the simulation's/in-game time.

But you have a good point on the game speed. I thought about that too.

ShiroAka:
Added a message filter that compares strings (for now). Doesn't seem to degrade performance as I thought it would, but I don't have a slow/old PC to test on.

Now the staff placement messages shouldn't appear in the history.

ARGAMX:
> In a ledger it makes sense to display the in-game date because it refers to what's actually happening in the simulation, but for the notifications I'm not so sure.
>
> The way I see it, the notifications are more related to general advice or results of your direct actions, and are something that could impact your actions directly (instead of a being a direct result of the simulation). So the focus is not on the simulation per-se, but on the user. That's why I suggested displaying how many real seconds/minutes ago did the notification appear, instead of tracking that based on the simulation's/in-game time.

Yes, sounds logical.

The thing is that as I think most user-initiated notifications make less sense to store in history.
Like those messages about trying to put an employee in an inappropriate tile.
Or when we can't buy something.
Or when we put a nurse to the GP office and the Adviser says we can't put her here.
The nurse will leave the office on her own.

But why should we reread that type of messages in history when the moment has already passed?
For such messages, I agree, game time really doesn't matter.
But I don't think there's much point in showing these kinds of messages in history at all. 
They lose relevance instantly.

When we talk about Advisor notifications, I actually think primarily about events in the simulation.

Why them? Because the player isn't the cause of these notifications. They pop up when the player is busy with buildng the room and can't switch to another task or something else and misses them. That's where a notification history would be most useful. I constantly miss messages like these as a player.

And in light of this problem, your implementation of the history for such notifications is incredibly useful!

I thought you were trying to solve this exact problem.
This is exactly the problem I'm experiencing as a player.

What I'm constantly missing:

Student gained X skill or degree (result of the simulation)
New room researched (result of the simulation)
Some research direction is fully researched (result of the simulation)
Boiler breakdown (result of the simulation)
You've reached the balance plus and it's time to pay off the bank loan (result of the simulation)

That's why the question:
What specific general advices or results of your direct actions would you like to keep in history?

ShiroAka:
> The thing is that as I think most user-initiated notifications make less sense to store in history. Like those messages about trying to put an employee in an inappropriate tile. Or when we can't buy something. Or when we put a nurse to the GP office and the Adviser says we can't put her here. The nurse will leave the office on her own.
> 
> But why should we reread that type of messages in history when the moment has already passed? For such messages, I agree, game time really doesn't matter. But I don't think there's much point in showing these kinds of messages in history at all. They lose relevance instantly.

That's why I added a commit to filter those messages. I tested it and they don't show anymore in the history

I only added a few, so the filter list is incomplete.

For example, some of the strings in the "adviser.information" and "adviser.information.initial_general_advice" may be helpful, but if the user misses them they aren't missing much, so they may be filtered from the history.


> What I'm constantly missing:
> 
> Student gained X skill or degree (result of the simulation) New room researched (result of the simulation) Some research direction is fully researched (result of the simulation) Boiler breakdown (result of the simulation) You've reached the balance plus and it's time to pay off the bank loan (result of the simulation)
> 
> That's why the question: What specific general advices or results of your direct actions would you like to keep in history?

Some of the warnings, for example, are in some way a direct result of your actions (you missed adding something to the hospital, or you are doing something wrong), and they are useful to the player.

The "room_requirements" and "staff_advice" strings are other examples of the above, but some of them may be removed from the history, because they don't add much info (and I think they appear again after a while if you don't do what they say)

lewri:
> I wouldn't deduplicate, this is the real notification history.

It's worth noting, that #2068 does do some de-duplication before a message is served; if the previous message has been waiting in the queue so long it's come up again.

Alberth289346:
> ... But why should we reread that type of messages in history when the moment has already passed?

I typically notice the advisor when he dives down behind the bottom window again, and completely miss the message.

New users may not immediately catch what is wrong, and currently you cannot get the message back for more careful reading.

Different users have different needs. By default log everything. Then add settings to skip logging some messages. In that way everybody can pick what they prefer.

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,181 @@
...
+  self:UIResizable(ui, width, height, col_bg)
+  self.ui = ui
+  self.modal_class = "adviser_history"
+
+  self.esc_closes = true
+  self.resizable = false
+  self:setDefaultPosition(0.05, 0.05)
+
+  self.adviser_messages = {}
+  self.list_table = {}
```
Alberth289346: For the data fields (`rows_shown` below too), can you add a line of comment describing what it contains?

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,181 @@
...
+local header_height = 20
+local window_margin = 15
+
+function UIAdviserHistory:UIAdviserHistory(ui)
+  local app = ui.app
+
+  -- Calculate menu's height
+  local height_divisor = 4
+  local approx_rows_space = math.floor(app.config.height / height_divisor)
+  self.rows = math.ceil(approx_rows_space / row_height)
```
Alberth289346: This is actually `self.num_visible_rows` ?

`rows` is a bit too generic. The name doesn't make it clear what is stored in it. (Eg the name does not exclude rows of messages).

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,181 @@
...
+  local rows_space = self.rows * row_height
+
+  local width = 500
```
Alberth289346: You may want to move this line to the other constants at lines 36-38.
In that way, all magic sizes are near each other. No need to go look through the entire file to find them then.

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
Alberth289346: `rows_available` perhaps?

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,181 @@
...
+    local message_width = 400
+    local x = window_margin
+    local y = window_margin
+
+
+    -- Draw headers
+    local delete_header = self:addBevelPanel(x, window_margin, delete_width, header_height, col_highlight)
+    local delete_all_button = delete_header:makeButton(0, 0, delete_width, header_height, nil, deleteAllButtonClicked)
+          :setTooltip(_S.tooltip.adviser_history.header.delete_message):setLabel("X", self.white_font, "center")
+    x = x + delete_width
```
Alberth289346: Instead of resetting + incrementing `x`, you could have 3 different variables that denote the position of the left edge of the window area.

CorsixTH/Lua/dialogs/adviser.lua
```diff
@@ -58,6 +61,65 @@ function UIAdviser:UIAdviser(ui)
...
+    _A.staff_place_advice.receptionists_only_at_desk,
+    _A.staff_place_advice.only_psychiatrists,
+    _A.staff_place_advice.only_surgeons,
+    _A.staff_place_advice.only_nurses_in_room,
+    _A.staff_place_advice.only_doctors_in_room,
+    _A.staff_place_advice.only_researchers,
+    _A.staff_place_advice.nurses_cannot_work_in_room,
+    _A.staff_place_advice.doctors_cannot_work_in_room,
+    _A.placement_info.staff_cannot_place,
+  }
```
Alberth289346: What happens if we decide to delete one of those strings?

(Or, if we decide to add more strings?)

ShiroAka: Adding more strings should cause no issues performance wise (I tested with 50 elements in that array and saw no problems), so changing that array would control whether the messages are shown in the history.

I added those strings because they produce the most spam (and add no useful info), but I have been thinking about adding more.

ShiroAka:
I just had an idea regarding the message filtering (similar to what Alberth proposed above)

So far the staff placement messages are being filtered from the history, and I have been thinking about adding more messages to the filter...

But what if I add a column that lets the user block a certain message from the history?
Blocking a message should delete all other instances from the history and prevent it from appearing again.

The only issue is that I would need a way to identify the messages once they are in the history (not necessarily with an ID).

And I would also need to add a section somewhere in the config to unblock the messages (and possibly a way to block them from there)

CorsixTH/Lua/dialogs/adviser.lua
ShiroAka: An alternative to this could be to check if a message already exists in the list and, if so, bring it to the top instead of adding a new instance.

ShiroAka:
Found a little bug with the message history. I'll leave it here so I don't forget

If I change the language after saving the game with the window open, it saves all the strings in the original language.

But what makes it more interesting is that the type of specialist in the 3rd message is translated to the language in use after loading the save.
image

It may be related to this line in the "doctor.lua" file. 
image

Although that line comes before the message is added to the history (so the string should already be formatted by then), it makes me think the program is somehow formatting the string after loading the game...

ShiroAka:
Also, in the original commit for this PR I added some lines in the GameUI:afterLoad function just in case, but maybe they are not needed.
image

Having those lines there prevents the history from getting saved with the game (and I think it would be useful to be able to save that info), but let me know if I should add them back.

CorsixTH/Lua/dialogs/adviser.lua
```diff
@@ -41,6 +41,9 @@ function UIAdviser:UIAdviser(ui)
   self.queued_messages = {}        -- There might be many messages in a row
   self.timer = nil                 -- Timer which hides adviser at the end
 
+  self.message_history = {}        -- Store past messages to be able to view them later
+  self.message_history_limit = 20  -- Limit the number of messages stored
```
lewri, line 45: Shiro, for the message you sent this morning the issues you encountered are because the advisor is missing these values in old saves.

A basic solution is just to set these values to defaults in an afterload. A smarter afterload would be trying to retrospectively fill history but I don't believe that's really necessary.

ARGAMX:
> Found a little bug with the message history. I'll leave it here so I don't forget
> 
> If I change the language after saving the game with the window open, it saves all the strings in the original language.
> 
> But what makes it more interesting is that the type of specialist in the 3rd message is translated to the language in use after loading the save. image
> 
> It may be related to this line in the "doctor.lua" file. image
> 
> Although that line comes before the message is added to the history (so the string should already be formatted by then), it makes me think the program is somehow formatting the string after loading the game...

It's okay.
This is fine considering how the game currently handles language switching.
If you switch between a non-Unicode languages, you'll get the bug shown in your screenshot.
But the bugs are especially significant if you're switching between pair of Unicode and non-Unicode languages.
Then, in many texts, you'll just get the question mark "?????????"
Check the video in issue 2974.
So, in light of all the above, I'd say this isn't a bug in your functionality.
I think you can leave it and you don't have to resolve it as part of your PR.

ARGAMX:
> Also, in the original commit for this PR I added some lines in the GameUI:afterLoad function just in case, but maybe they are not needed. image
> 
> Having those lines there prevents the history from getting saved with the game (and I think it would be useful to be able to save that info), but let me know if I should add them back.

I don't think it should delete history when saving or loading.
This is especially true when loading an autosave.
And in general, it would be cool to load a save and get a contextual understanding of what happened before.
So, I don't think you should restore these clearing code.
I tried this version with clearing, and frankly, I was upset when the history was empty after loading.

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ShiroAka: A lot of this code was copied from the "machine_menu.lua" and I didn't improve it, so it may need some improvements there too.
I'll review it later and reorder the code a bit

TheCycoONE:
> It's okay. This is fine considering how the game currently handles language switching. If you switch between a non-Unicode languages, you'll get the bug shown in your screenshot. But the bugs are especially significant if you're switching between pair of Unicode and non-Unicode languages. Then, in many texts, you'll just get the question mark "?????????" Check the video in issue 2974. So, in light of all the above, I'd say this isn't a bug in your functionality. I think you can leave it and you don't have to resolve it as part of your PR.

 Can you report if it isn't already. This is supposed to work and does work some places. It's not a good reason to break it further if we can help it.

ARGAMX:
> Can you report if it isn't already. This is supposed to work and does work some places. It's not a good reason to break it further if we can help it.

Screenshot 2025-10-10 at 11 57 25

Screenshot 2025-10-10 at 11 57 59

Screenshot 2025-10-10 at 12 01 47

ARGAMX:
> Can you report if it isn't already. This is supposed to work and does work some places. It's not a good reason to break it further if we can help it.

Ok, I will report an issue.

ARGAMX (commented):
I merged this change for myself and played with it.
This is a very useful enchahment! 👍
I enjoyed playing with it!

What I didn't like was that after loading, I can't see the Adviser history.
Sometimes I want to look at something in another save and then go back and continue playing current level.
So I save, load another save, then load my first save — now the Adviser message history is empty — and then I regret that I loaded the game.
So, saving the game doesn't save this state, which isn't very good or convenient.

On the other hand, if enable displaying the history after loading, then when switching languages, you might get artifacts like the text "?????????".

But I think the damage from this defect is less than the damage from loosing the history.
Overall, I recommend keeping the history after loading.
Otherwise, this cool feature loses some of its usefulness.
I have disabled history wipe on load for me localy.

I hope this PR haven't hung on?

ARGAMX (approved):
_chuck-norris

ShiroAka:
Just in case, when merging try to do a squash; otherwise, the main branch will get filled with several commits with bad names from this PR :)

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,182 @@
...
+  tick_timer = tick_timer - 1
+  if tick_timer <= 0 then
+    tick_timer = ticks_to_skip
+    self:update()
+  end
+end
+
+function UIAdviserHistory:close()
+  Window.close(self)
+end
```
TheCycoONE: newline nit

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,182 @@
...
+    end
+  end
+end
+
+function UIAdviserHistory:onTick()
+  -- Dialog does not update on every tick
+  tick_timer = tick_timer - 1
+  if tick_timer <= 0 then
+    tick_timer = ticks_to_skip
+    self:update()
```
TheCycoONE, line 183: Why is the scrollbar updated on a timer instead of when the message_history is updated?

TheCycoONE (approved):

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ShiroAka: I copied that from the machine menu code, but you are right, I should have changed that (didn't do it because I didn't want to break anything).
Good catch!

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,191 @@
...
+  local new_history_top = self.ui.adviser.message_history[1]
+
+  if current_history_top ~= new_history_top then
+    self:update()
+  end
+end
+
+function UIAdviserHistory:close()
+  Window.close(self)
+end
```
ARGAMX:
```suggestion
end

```

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ARGAMX: > newline nit

newline added 👌

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ShiroAka: I added the newline but it seems my github client chose to remove it (or maybe it was VSCode) :S

ARGAMX: > I added the newline but it seems my github client chose to remove it (or maybe it was VSCode) :S

Interesting. 🙂
If this is the case then I'm sure there must be an option in the settings to turn this on/off.

But specifically for this PR, if a new line has already been added, everything is ok here now.

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,192 @@
...
+          col_shadow.red, col_shadow.green, col_shadow.blue)
+        :makeScrollbar(col_bg, scrollbarMovedCallback, 1, 1, 10, 1)
+    -- Add close button
+    self:addPanel(337, x,  8):makeButton(0, 0, 24, 24, 338, self.close)
+        :setTooltip(_S.tooltip.adviser_history.close)
+  end
+end
+
+function UIAdviserHistory:update()
+  -- self.adviser_messages = self.ui.adviser.message_history --> WRONG, this way both variables point to the same table
```
lewri: ?

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,192 @@
...
+function UIAdviserHistory:UIAdviserHistory(ui)
+  local app = ui.app
+
+  self.ui = ui
+  self.modal_class = "adviser_history"
+
+  self.esc_closes = true
+  self.resizable = false
+
+  self.adviser_messages = {} -- Contains all the messages in the adviser.adviser_history prop
```
lewri: What's the prop for/mean?

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,192 @@
...
+        delete_panel = delete_panel,
+        delete_button = delete_button,
+        message_panel = message_panel,
+      })
+    end
+
+    x = x + 10 -- Add 10 pixels from end of the message rows
+
+    -- Add scrollbar
+    self.scrollbar = self:addColourPanel(x, self.window_margin + 20, 24, self.row_height * rows, --y = self.window_margin + close button height
```
lewri: I'm confused what the comment is referencing. It'd be saner to distinctly have a line of `self.window_margin = self.window_margin + close_button_height`

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,192 @@
...
+  self:scrollbarMoved()
+end
+
+function UIAdviserHistory:scrollbarMoved()
+  local scroll_pos = self.scrollbar.value
+  for i = 1, self.rows_shown, 1 do
+    local message = self.adviser_messages[i + scroll_pos - 1]
+    local row = self.list_table[i]
+    if message then
+      row.message_panel:setLabel(" " .. message, nil, "left"):setColour(col_bg)
```
lewri: Does it to weird things if it's not just `""`

lewri: What is the space in aid of actually? If you need that it sounds like some offset is wrong.

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -881,6 +883,16 @@ function UIBottomPanel:dialogMachineMenu(enable)
   self.ui:playSound("selectx.wav")
 end
 
+function UIBottomPanel:dialogAdviserHistory(enable)
```
lewri, line 913: Nice the functionality is available if someone wants to add a button.

CorsixTH/Lua/languages/english.lua
```diff
@@ -1034,6 +1036,21 @@ tooltip.machine_menu = {
...
+
+tooltip.adviser_history = {
+  delete_message = "Click to dismiss this message",
+  message = "List of adviser messages - Newest messages appear first",
+  header = {
+    message = "Adviser messages",
+    delete_message = "Click to delete all messages"
+  },
+  close = "Close the adviser history dialog",
+}
```
lewri, line 1081: Are these in the Discord yet at all on the translations page?

lewri (commented):
Please luadoc the functions of your new file. I believe you're mostly there though.

CorsixTH/Lua/config_finder.lua
```diff
@@ -828,7 +829,8 @@ param(hotkeys_values, 'ingame_panel_research') ..
 param(hotkeys_values, 'ingame_panel_status') ..
 param(hotkeys_values, 'ingame_panel_charts') ..
 param(hotkeys_values, 'ingame_panel_policy') ..
-param(hotkeys_values, 'ingame_panel_machineMenu') ..
+param(hotkeys_values, 'ingame_panel_adviserHistory') ..
+param(hotkeys_values, 'ingame_panel_machineMenu') ..
```
ARGAMX:
```suggestion
param(hotkeys_values, 'ingame_panel_machineMenu') ..
param(hotkeys_values, 'ingame_panel_adviserHistory') ..
```

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ShiroAka: old comment that I made while refactoring and forgot to delete

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ShiroAka: Property/attribute
But I'll be removing that list, so this is not relevant anymore

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ShiroAka: I added that space to avoid the text being so close to the edge of the column. I feel it's easier to read with that extra space
Here you can clearly see the difference
image

CorsixTH/Lua/languages/english.lua
ShiroAka: Now they are :)

CorsixTH/Lua/dialogs/bottom_panel.lua
ShiroAka: I can add a button later, but not in this PR

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,187 @@
...
+
+  self.rows_shown = 0
+
+  self.row_height = 40
+  self.header_height = 20
+  self.window_margin = 15
+  self.delete_col_width = 30
+  self.message_col_width = 400
+
+  local total_width = 500
```
tobylane:
```suggestion
  local total_width = 489
```
The margin + columns + line 132 spacing + scrollbar width + margin is 489, 500 looks too wide.

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,187 @@
...
+
+    x = x + 10 -- Add 10 pixels from end of the message rows
+    y = self.window_margin + self.header_height
+
+    -- Add scrollbar
+    self.scrollbar = self:addColourPanel(x, y, 24, self.row_height * rows, col_shadow.red, col_shadow.green, col_shadow.blue)
+        :makeScrollbar(col_bg, scrollbarMovedCallback, 1, 1, 10, 1)
+    -- Add close button
+    self:addPanel(337, x,  8):makeButton(0, 0, 24, 24, 338, self.close)
+        :setTooltip(_S.tooltip.adviser_history.close)
```
tobylane:
```suggestion
    y = self.window_margin + self.header_height + 5

    -- Add scrollbar
    self.scrollbar = self:addColourPanel(x, y, 24, self.row_height * rows - 5,
        col_shadow.red, col_shadow.green, col_shadow.blue)
        :makeScrollbar(col_bg, scrollbarMovedCallback, 1, 1, 10, 1)
    -- Add close button
    self:addPanel(337, x, 14):makeButton(0, 0, 24, 24, 338, self.close)
        :setTooltip(_S.tooltip.adviser_history.close)
```
Not sure on this suggestion. Is the close button better off with the middle or bottom matching the column header?

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,187 @@
...
+  -- Checking the lenght of both lists would fail if a message was moved from the middle to the top due to duplication.
+  -- So we just check if the top string between lists has changed to determine if we need to update
+  local new_history_top = self.ui.adviser.message_history[1]
+
+  if self.last_seen_history_top ~= new_history_top then
+    self:update()
+  end
+end
+
+function UIAdviserHistory:close()
```
tobylane: Not necessary as it's inherited.

CorsixTH/Lua/game_ui.lua
```diff
@@ -1318,6 +1318,8 @@ function GameUI:afterLoad(old, new)
   if old < 240 then
     self.subtitles = Subtitles(self)
     self:addWindow(self.subtitles)
+    self.adviser.message_history = {}
+    self.adviser.message_history_limit = 20
```
tobylane: This can go in `UIAdviser:afterLoad`, with a savegame version bump in App.

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
lewri: I think you've managed to somehow recreate #2467 here. Really, your spacing should be equal for each line.

CorsixTH/Lua/dialogs/adviser.lua
```diff
@@ -58,6 +61,52 @@ function UIAdviser:UIAdviser(ui)
...
+      return true, index
+    end
+  end
+  return false, -1
+end
+
+--! Adds a message to the adviser's message history.
+--! If the history exceeds a certain limit, the oldest messages are removed.
+function UIAdviser:addMessageToHistory(speech)
+  local isMessageInHistory, messageIndexInHistory = self:isMessageInHistory(speech)
```
lewri: Camel case doesn't really belong here (line 76). But also why run a boolean when you're already trying to find a value. You can use the existence of a value here for your checks. Also while index can't be 0 in this case, it's weird not to account for it.

```lua
function UIAdviser:getIndexInMessageHistory(message)
  for index, msg in ipairs(self.message_history) do
    if msg == message then
      return index
    end
  end
  return -1
end
```

```lua
local message_history_index = self:getIndexInMessageHistory(speech)

if message_history_index > 0 then
...
```

CorsixTH/Lua/dialogs/adviser.lua
```diff
@@ -233,6 +283,8 @@ function UIAdviser:onMouseDown(button, x, y)
     end
   elseif button == "right" then
     self:hide()
+    --Dump dismissed messages to history
```
lewri: Missing a space

CorsixTH/Lua/window.lua
```diff
@@ -311,7 +312,7 @@ function Panel:drawLabel(canvas, x, y, limit)
 
   if not multi_line then
     text = {text}
-    wrapped = false
+    wrapped = false or self.force_text_wrap
```
lewri: Just write `wrapped = self.force_text_wrap`. A default false value is already false.

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ARGAMX: > Not sure on this suggestion. Is the close button better off with the middle or bottom matching the column header?

I think so too. Applied

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ARGAMX: > The margin + columns + line 132 spacing + scrollbar width + margin is 489, 500 looks too wide.

I think so too. Applied

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,187 @@
...
+  Window.close(self)
+end
```
ARGAMX:
```suggestion
```

CorsixTH/Lua/dialogs/adviser.lua
ARGAMX:
```suggestion
    -- Dump dismissed messages to history
```

CorsixTH/Lua/dialogs/adviser.lua
```diff
@@ -58,6 +61,52 @@ function UIAdviser:UIAdviser(ui)
...
+
+--! Adds a message to the adviser's message history.
+--! If the history exceeds a certain limit, the oldest messages are removed.
+function UIAdviser:addMessageToHistory(speech)
+  local isMessageInHistory, messageIndexInHistory = self:isMessageInHistory(speech)
+
+  if isMessageInHistory then
+    -- If the message is already in history, remove it first
+    table.remove(self.message_history, messageIndexInHistory)
+  end
```
ARGAMX:
```suggestion
function UIAdviser:messageIndexInHistory(message)
  for index, msg in ipairs(self.message_history) do
    if msg == message then
      return index
    end
  end
  return nil
end

--! Adds a message to the adviser's message history.
--! If the history exceeds a certain limit, the oldest messages are removed.
function UIAdviser:addMessageToHistory(speech)
  local message_index_in_history = self:messageIndexInHistory(speech)

  if message_index_in_history then
    -- If the message is already in history, remove it first
    table.remove(self.message_history, message_index_in_history)
  end
```

CorsixTH/Lua/window.lua
ARGAMX:
```suggestion
    wrapped = self.force_text_wrap
```

ARGAMX: Applied

CorsixTH/Lua/dialogs/adviser.lua
ARGAMX: Applied

CorsixTH/Lua/dialogs/adviser.lua
ARGAMX: Applied

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ARGAMX: Applied

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,184 @@
...
+
+    x = x + 10 -- Add 10 pixels from end of the message rows
+    y = self.window_margin + self.header_height
+
+    -- Add scrollbar
+    self.scrollbar = self:addColourPanel(x, y, 24, self.row_height * rows, col_shadow.red, col_shadow.green, col_shadow.blue)
+        :makeScrollbar(col_bg, scrollbarMovedCallback, 1, 1, 10, 1)
+    -- Add close button
+    self:addPanel(337, x,  8):makeButton(0, 0, 24, 24, 338, self.close)
+        :setTooltip(_S.tooltip.adviser_history.close)
```
ARGAMX:
```suggestion
    y = self.window_margin + self.header_height + 10

    -- Add scrollbar
    self.scrollbar = self:addColourPanel(x, y, 24, self.row_height * rows - 10,
        col_shadow.red, col_shadow.green, col_shadow.blue)
        :makeScrollbar(col_bg, scrollbarMovedCallback, 1, 1, 10, 1)

    -- Add close button
    self:addPanel(337, x, 15):makeButton(0, 0, 24, 24, 338, self.close)
        :setTooltip(_S.tooltip.adviser_history.close)
```

CorsixTH/Lua/game_ui.lua
ARGAMX: > This can go in `UIAdviser:afterLoad`, with a savegame version bump in App.

Applied

CorsixTH/Lua/dialogs/adviser.lua
ARGAMX: Applied

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ARGAMX: I like the space ShiroAka added here.
I agree with him that it's easier to read with that extra space.
So I wouldn't want to remove it either.

ARGAMX:
PR ready for review

ARGAMX:
Bump for review

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
```diff
@@ -0,0 +1,185 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE. --]]
+
+--! Adviser message history
+class "UIAdviserHistory" (UIResizable)
+
+---@type UIAdviserHistory
+local UIAdviserHistory = _G["UIAdviserHistory"]
+
+local col_bg = { red = 60, green = 174, blue = 201, }
```
TheCycoONE: Is it possible to use the new colour constants here?

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
TheCycoONE: It's a hack and not found anywhere else in CorsixTH. Can keep it but add an issue to replace it with padding support for panel labels.

CorsixTH/Lua/window.lua
```diff
@@ -311,7 +312,7 @@ function Panel:drawLabel(canvas, x, y, limit)
 
   if not multi_line then
     text = {text}
-    wrapped = false
+    wrapped = self.force_text_wrap
```
TheCycoONE: What's the git blame on this one anyway. Seems like there's been a messy history with the wrapping (auto_clip?)

It's strange that passing in the text as a table is the only way to get wrapping.

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
lewri: I'd rather we kept a uniform method to update the paragraph formatting and not pick and choose as a hack. I'd recommend the alternative of removing this hack and adding a universal controller after (someone then has more reason to actually do the work).

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
lewri: I don't think they have overlap with the standards, unless you want this window purple. Though adding them as their own definitions in utility is a fine option too.

CorsixTH/Lua/window.lua
ARGAMX: > What's the git blame on this one anyway. Seems like there's been a messy history with the wrapping (auto_clip?)
> 
> It's strange that passing in the text as a table is the only way to get wrapping.

Indeed, it is strange.
Git blame points to 92305d15e345f291dfbf07988d01d0559a20c46b.
This logic has not changed since 2013.
What should we do with this?

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ARGAMX: Ok, that make sense. 
That space removed. 👌

CorsixTH/Lua/dialogs/resizables/adviser_history.lua
ARGAMX: Colors moved to utility. 👌

CorsixTH/Lua/window.lua
TheCycoONE: I don't love the force wrap name. I'd like a setWrapped true/false and fall back to the table check if it is nil.

Separately in another PR I'd like everywhere that passes a table to add setWrapped() and I would like auto_clip removed.

ARGAMX: > I don't love the force wrap name. I'd like a setWrapped true/false and fall back to the table check if it is nil.
> 
> Separately in another PR I'd like everywhere that passes a table to add setWrapped() and I would like auto_clip removed.

Okay.
I made it so that if `setWrapped` isn't set, it simply falls back to the behavior that was before this PR.
Also removed `force` from the name.

CorsixTH/Lua/window.lua
```diff
@@ -304,17 +305,17 @@ end
...
   local center_y = false
-  local s = self.apply_ui_scale and TheApp.config.ui_scale or 1
+  local wrapped = not self.auto_clip
 
+  local multi_line = type(text) == "table"
   if not multi_line then
     text = {text}
-    wrapped = false
     center_y = true
+    wrapped = (self.wrap_text ~= nil) and self.wrap_text or false
```
lewri, line 315: Does this break wrapping on any other windows?  I'm guessing nothing currently uses this

ARGAMX: > Does this break wrapping on any other windows? I'm guessing nothing currently uses this

On any other windows it still stay `false` as it was `false`.

ARGAMX (approved):

PR #3026, opened 2025-09-19
render_target RAII

TheCycoONE:
Some refactoring I did while working on the movie text. The rest of my PR isn't useful anymore but I figured it was worth saving these changes.


Alberth289346 (approved):
Looks good.

The only thing bothering me are the old `pParam...` and `oParam...` names, maybe rename them now that we do access by reference?

TheCycoONE:
> Looks good.
> 
> The only thing bothering me are the old `pParam...` and `oParam...` names, maybe rename them now that we do access by reference?

Resolved

Issue #3025, opened 2025-09-18
[Bug] Click sound effect is missing in various places


MandrakeHorse:
### Describe the issue

There is a number of places where we are not playing the "click" sound effect (SELECTX.wav)

When furnishing rooms in CorsixTH:
- Clicking the Buy/sell objects button
- Clicking the Purchase item(s) button in the Choose Items panel
- Clicking individual items in the list of bought items (to select them for placement)
<img alt="Building GP's office" src="https://github.com/user-attachments/assets/72eb0e7c-cfee-40e7-9155-374677146922" />

When using the Staff Management panel:
- Clicking on a staff member's row to select them 

<img alt="Staff management panel" src="https://github.com/user-attachments/assets/e5e7a73f-ddc7-4b77-a1d0-d91694935585" />

### Steps to Reproduce

1. Start/load a game
2. Start building any room
3. Draw the walls and confirm
4. Add the door and confirm
5. Start furnishing the room
6. Hire some staff (if necessary)
7. Open the Staff Management panel
8. Click a staff member's row to select it

### Expected Behaviour

- [ ] A click sound is heard when clicking the Buy/sell objects button
- [ ] A click sound is heard when clicking the Purchase item(s) button in the Choose Items panel
- [ ] A click sound is heard when clicking individual items in the list of bought items
- [ ] A click sound is heard when clicking individual staff members in the Staff Management panel
- [ ] A click sound is heard when clicking any of the sweeping/watering/repair buttons in the Handyman panel
- [ ] A click sound is heard when clicking the Sack button on any of the Staff Information Panels
- [ ] A click sound is heard when clicking the Health Graph in the Patient Information Panel
- [ ] A click sound is heard when clicking the "Staff leave rooms" check and cross buttons on the Policy screen

### Have you enabled any current CorsixTH Experimental Features?

- [ ] Right mouse panning
- [ ] Falling actions (applies to CorsixTH versions < 0.70.0)

### Save Game

Save game attached to report: No

### CorsixTH Version

0.69.1

### Operating System

Windows 10

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

Matroftt:
Also handymen dialog

MandrakeHorse:
When using the Handyman's panel:
- [ ] Clicking on a Handyman's Sweeping/Watering/Repair button to increase its priority

When using any staff member's panel:
- [ ] Clicking the Sack button to bring up the confirmation window

[...]

Issue #3018, opened 2025-09-17
[Bug] Missing hover sound in Furnish Corridor panel


MandrakeHorse:
### Describe the issue

In Theme Hospital, a soft "ding" sound effect can be heard when hovering over items in the Build Room and Furnish Corridor panels.

In CorsixTH, although this sound can be heard when hovering over items in the Build Room panel, the highlight sound effect (HLIGHTP2) is not heard when hovering over items in the Furnish Corridor panel.

<img alt="furnish corridor panel" src="https://github.com/user-attachments/assets/ec5c4cd6-4a9f-4d4b-b66f-4543a99ddfc2" />

`/CorsixTH/Lua/dialogs/build_room.lua` includes this line of code which does not appear in `/CorsixTH/Lua/dialogs/furnish_corridor.lua`
https://github.com/CorsixTH/CorsixTH/blob/cad1d4c5a5667f4c5c4d9b670650565bfac1682d/CorsixTH/Lua/dialogs/build_room.lua#L209

### Steps to Reproduce

1. Start a new game
2. Click the Furnish Corridor button
3. Hover over the items in the Furnish Corridor panel

### Expected Behaviour

A soft "ding" sound is heard when hovering over items in the Furnish Corridor panel.

### Have you enabled any current CorsixTH Experimental Features?

- [ ] Right mouse panning
- [ ] Falling actions (applies to CorsixTH versions < 0.70.0)

### Save Game

Save game attached to report: No

### CorsixTH Version

0.69.1

### Operating System

Linux Mint 22.2 Cinnamon

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

Matroftt:
I almost finished this along with other sounds. Going to make a PR next day

ARGAMX:
In the Build Room phase when Furnish Room this sound also missing while hovering over room items list.

Matroftt:
> In the Build Room phase when Furnish Room this sound also missing while hovering over room items list.

Furnish room is a part of furnish corridor

Matroftt:
Was this ding sound used somewhere else? (excluding hovering over humanoids and TH main menu)

MandrakeHorse:
> Was this ding sound used somewhere else? (excluding hovering over humanoids and TH main menu)

[...]

PR #3030, opened 2025-09-19
Add missing hover and click sound effects in various places

Matroftt:
Fixes 
* #3025 
* #3018 

- Lots "ding!" sounds (HLight5.wav and HLightP2.wav) are implemented, in these situations:
  - Build room (switched from HLightP2.wav to HLight5.wav which TH uses in this dialog)
  - Furnish corridor / Furnish room
  - Top menu
  - Timer buttons
  - UIInformation close button
  - Hovering over humanoid with cold/hot mood
  - Hovering over room/door with queue
  - Hovering over object list in place objects dialog
 
- Lots select sounds (Selectx.wav) are implemented, in these situations:
  - Switch between treatment history / graph in patient dialog
  - Fire staff button
  - Handymen priorities change
  - Handymen plot change
  - Selecting staff in staff management panel
  - Toggle buttons in policy dialog
  - Furnish room and move items buttons
  - Buy items button

- Makes behavior of timer button and UIInformation button as it was in TH; Now it turns red on hover instead of on click.
- Improves behavior of place_dialog.lua making it closer to TH:

place_objects.lua updated dialog

- Improves staff management panel (select sfx plays when you select a doctor; hovering cursor over name makes it blue and ding sfx plays;

{7C72566A-2884-4AC9-AD33-F1843C5290A7}

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
@@ -34,6 +34,7 @@ function UIStaffManagement:UIStaffManagement(ui)
     local palette = gfx:loadPalette("QData", "Staff01V.pal", true)
     self.panel_sprites = gfx:loadSpriteTable("QData", "Staff02V", true, palette)
     self.title_font = gfx:loadFontAndSpriteTable("QData", "Font01V", false, palette)
+    self.blue_font = gfx:loadFontAndSpriteTable("QData", "Font02V", false, palette)
     self.face_parts = ui.app.gfx:loadRaw("Face01V", 65, 1350, nil, "Data", "MPalette.dat")
   end) then
     ui:addWindow(UIInformation(ui, {_S.errors.dialog_missing_graphics}))
@@ -49,6 +50,8 @@ function UIStaffManagement:UIStaffManagement(ui)
   self:updateStaffList()
 
   self.default_button_sound = "selectx.wav"
+  self.hover_id = 0
+  self.visual_hover_id = 0
 
   -- Close button
   self:addPanel(0, 603, 443):makeButton(0, 0, 26, 26, 10, self.close):setTooltip(_S.tooltip.staff_list.close)
@@ -277,9 +280,13 @@ function UIStaffManagement:draw(canvas, x, y)
       local row_no = i - (self.page-1)*10
       self.row_blankers[row_no].visible = false
       titles:draw(canvas, row_no + 10*(self.page-1), x + 58, y + 63 + row_no*27)
-      titles:draw(canvas, staff.profile:getFullName(),
+      local font = self.title_font
+      if i == self.visual_hover_id then
+        font = self.blue_font
+      end
+      font:draw(canvas, staff.profile:getFullName(),
           x + 88, y + 63 + row_no*27)
-      titles:draw(canvas, "$" .. staff.profile.wage, x + 230, y + 63 + row_no*27, 80, 0)
+      font:draw(canvas, "$" .. staff.profile.wage, x + 230, y + 63 + row_no*27, 80, 0)
 
       -- Draw the morale, tiredness and skill for this staff member
       if happiness_bar_width ~= 0 then
@@ -299,6 +306,7 @@ function UIStaffManagement:draw(canvas, x, y)
       end
     end
   end
+
   -- Make sure the other ones are not visible
   for i = #staff_list + 1 - (self.page-1)*10, 10 do
     self.row_blankers[i].visible = true
@@ -424,6 +432,7 @@ function UIStaffManagement:onMouseDown(code, x, y)
       -- Hit staff row
       if #self.staff_members[self.category] - (self.page - 1)*10 > math_floor((y - 81)/27) then
         self.selected_staff = math_floor((y - 81)/27) + 1 + (self.page - 1)*10
+        TheApp.audio:playSound("selectx.wav")
       end
     else
       local inside_view_of_the_staff_area = (x > 497 and x < 580) and (y > 373 and y < 455)
@@ -433,6 +442,33 @@ function UIStaffManagement:onMouseDown(code, x, y)
     end
   end
   return UIFullscreen.onMouseDown(self, code, x, y)
+end
+
+function UIStaffManagement:onMouseMove(x, y, dx, dy)
+  local current_hover_id
+  local inside_staff_list_area
+
+  if self.page*10 >= #self.staff_members[self.category] then
+    inside_staff_list_area = (x > 50 and x < 624) and
+    (y > 82 and y < 82 + 27 * (#self.staff_members[self.category] % 10))
+  else
+    inside_staff_list_area = (x > 50 and x < 624) and (y > 82 and y < 351)
+  end
+
+  if inside_staff_list_area then
+    if #self.staff_members[self.category] - (self.page - 1)*10 > math_floor((y - 81)/27) then
+      current_hover_id = math_floor((y - 81)/27) + 1 + (self.page - 1)*10
+      if self.hover_id ~= current_hover_id then
+        self.ui:playSound("Hlight5.wav")
+        self.hover_id = current_hover_id
+        self.visual_hover_id = current_hover_id
+      end
+    end
+  else
+    self.hover_id = nil
+    self.visual_hover_id = nil
+  end
+  return Window:onMouseMove(x, y, dx, dy)
 end
 
 function UIStaffManagement:onMouseUp(code, x, y)
--- a/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
@@ -447,17 +447,20 @@ end
 function UIStaffManagement:onMouseMove(x, y, dx, dy)
   local current_hover_id
   local inside_staff_list_area
-
-  if self.page*10 >= #self.staff_members[self.category] then
+  local header_height = 81
+  local row_height = 27
+  local active_hover_id = math_floor((y - header_height)/row_height)
+  if self.page*10 > #self.staff_members[self.category] then
+    -- Make sure area contains no hollow space
     inside_staff_list_area = (x > 50 and x < 624) and
-    (y > 82 and y < 82 + 27 * (#self.staff_members[self.category] % 10))
+    (y > header_height and y < header_height + row_height * (#self.staff_members[self.category] % 10))
   else
-    inside_staff_list_area = (x > 50 and x < 624) and (y > 82 and y < 351)
+    inside_staff_list_area = (x > 50 and x < 624) and (y > header_height and y < 351)
   end
 
   if inside_staff_list_area then
-    if #self.staff_members[self.category] - (self.page - 1)*10 > math_floor((y - 81)/27) then
-      current_hover_id = math_floor((y - 81)/27) + 1 + (self.page - 1)*10
+    if #self.staff_members[self.category] - (self.page - 1)*10 > active_hover_id then
+      current_hover_id = active_hover_id + 1 + (self.page - 1)*10
       if self.hover_id ~= current_hover_id then
         self.ui:playSound("Hlight5.wav")
         self.hover_id = current_hover_id
--- a/CorsixTH/Lua/dialogs/information.lua
+++ b/CorsixTH/Lua/dialogs/information.lua
@@ -127,16 +127,6 @@ end
 
 function UIInformation:onMouseMove(x, y, dx, dy)
   self.active_hover = self:hoverTest(self.active_hover, x, y, self.width - 29, self.width - 10, self.height - 29, self.height - 10)
-  --[[
-  local current_hover = x > self.width - 29 and x < self.width - 10 and y > self.height - 29 and y < self.height - 10
-  if current_hover and not self.active_hover then
-    self.active_hover = true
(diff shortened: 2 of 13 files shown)


TheCycoONE:
Great attention to detail. Thanks!

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
```diff
@@ -435,6 +444,33 @@ function UIStaffManagement:onMouseDown(code, x, y)
   return UIFullscreen.onMouseDown(self, code, x, y)
 end
 
+function UIStaffManagement:onMouseMove(x, y, dx, dy)
+  local current_hover_id
+  local inside_staff_list_area
+
+  if self.page*10 >= #self.staff_members[self.category] then
```
ARGAMX: It's hard to understand what's being checked here.
Maybe we can make this condition more human-readable?

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
```diff
@@ -435,6 +444,33 @@ function UIStaffManagement:onMouseDown(code, x, y)
...
+    inside_staff_list_area = (x > 50 and x < 624) and
+    (y > 82 and y < 82 + 27 * (#self.staff_members[self.category] % 10))
+  else
+    inside_staff_list_area = (x > 50 and x < 624) and (y > 82 and y < 351)
+  end
+
+  if inside_staff_list_area then
+    if #self.staff_members[self.category] - (self.page - 1)*10 > math_floor((y - 81)/27) then
```
ARGAMX: It's hard to understand what's being checked here.
Maybe we can make this condition more human-readable?

This function could probably have some more comments.
Right now, it's hard to say what this function is for, what it does, or how it does it.
At least I can't answer it.

CorsixTH/Lua/dialogs/watch.lua
```diff
@@ -103,6 +124,19 @@ function UIWatch:onCountdownEnd()
   end
 end
 
+function UIWatch:onMouseMove(x, y, dx, dy)
+  local current_hover = x > 4 and x < 31 and y > 0 and y < 29
+  if current_hover and not self.hover then
+    self.hover = true
```
ARGAMX: Does the variable `hover` mean we've already played that sound and won't play it again?
It's not obvious from this variable name.
I assumed it after reading the entire function.

Perhaps we can give this variable a name that would make it obvious what it does?
For example, `hover_sound_already_played`.

just compare:
`hover_sound_already_played? - yes`
or
`hover? - yes` (what "yes"?)

Or we could write a comment that would make it obvious.
The same applies to other similar cases and `hover` variables in this PR.

CorsixTH/Lua/dialogs/watch.lua
```diff
@@ -84,6 +97,14 @@ function UIWatch:UIWatch(ui, count_type)
...
   else
     self:addPanel(timer_sprite, 0, 28):setTooltip(tooltips[count_type])
+    .custom_draw = --[[persistable:open_hospital_timer_button]] function(panel, canvas, x, y)
+      x = x + panel.x
+      y = y + panel.y
+      panel.window.panel_sprites:draw(canvas, panel.sprite_index, x, y)
+      if self.hover then
+        self.panel_sprites:draw(canvas, 17, x+4, y-28)
+      end
+    end
```
ARGAMX, line 107: I am seeing this code block three times in this PR. Apparently, it's copy-pasted. 
Copy-pasted code usually indicates that actually this code can be defined in one place and reused by calling it everywhere where it needed. Different numbers can be passed as parameters. May be in parent `Window` class.
I'm not sure what to do with this. But copy-pasted code always smells suspiciously.

CorsixTH/Lua/game_ui.lua
```diff
@@ -518,16 +522,24 @@ function GameUI:onCursorWorldPositionChange()
     self.cursor_room = room
   end
 
-  -- Any hoverable mood should be displayed on the new entity
```
ARGAMX, line 521: Why did we deleted this comment?

CorsixTH/Lua/dialogs/information.lua
```diff
@@ -116,6 +125,20 @@ function UIInformation:draw(canvas, x, y)
...
+  if current_hover and not self.hover then
+    self.hover = true
+    self.ui:playSound("Hlight5.wav")
+  else
+    if not current_hover then
+      self.hover = false
+    end
+  end
+  return Window:onMouseMove(x, y, dx, dy)
+end
```
ARGAMX, line 131: I think all this code is united by a single idea.
and can be moved to a separate function.

`onMouseMove` can simply call this function.

This will make a lot of sense when, in the future, `onMouseMove` will contains more code related to other tasks, and it will no longer be so obvious which part of `onMouseMove` is intended for what.

So I believe, `onMouseMove` should look like this:
```
func onMouseMove()
  doX()
  doY()
  doZ()
end
```

The same applies to other similar cases in this PR.

Matroftt: something like this i guess

```lua
function hover(current_hover, was_hovered)
  if current_hover and not was_hovered then
    return true
    self.ui:playSound("Hlight5.wav")
  else
    if not current_hover then
      return false
    end
  end
end

function onMouseMove(x, y, dx, dy)
  local current_hover = .......
  self.was_hovered = self:hover(current_hover, self.was_hovered)
  return Window:onMouseMove(x, y, dx, dy)
end

Comment (ARGAMX) [reply CorsixTH/Lua/dialogs/information.lua:131]: > something like this i guess

you can even pass `x, y` to that function, to completely incapsulate that functionality to that function.
`self.was_hovered` also can be managed/set inside that function.

`onMouseMove` no need to be involved in these "hover things".

File: CorsixTH/Lua/dialogs/watch.lua (hunk 3 above)
Comment (TheCycoONE) [reply CorsixTH/Lua/dialogs/watch.lua]: I disagree here. The action is hovering over the control and the sound is played on transition into that state but the variable should reflect the state not the side effect.

hover_active maybe.

File: CorsixTH/Lua/dialogs/watch.lua (hunk 4 above)
Comment (TheCycoONE) [reply CorsixTH/Lua/dialogs/watch.lua:107]: Yes this function has no upvalues it would be better at the top level defined once.

Comment (TheCycoONE) [reply CorsixTH/Lua/dialogs/watch.lua:107]: Sorry, it has one upvalue 'self' but that could be an argument or it could be defined on the class.

File: CorsixTH/Lua/game_ui.lua (hunk 5 above)
Comment (Matroftt) [reply CorsixTH/Lua/game_ui.lua:521]: my bad, deleted it by accident

**Comment by ARGAMX** [comment]
I tried this change but discovered one unpleasant effect.

In the original game, if you hover the cursor over the staff list or the top menu, the sound plays only once at a time. If the cursor moves to the next element while the previous element's sound hasn't finished playing, it's interrupted and the next element's sound starts over.

With this PR in CTH if you quickly hover the cursor over the staff list or the top menu, each new sound doesn't interrupt the previous one, and the same sound instantly plays multiple times in parallel. The sounds overlap, resulting in a louder noise than if only one sound were playing at a time. This increases noise and you want to reduce the game's volume, but then other sounds are hard to hear.

Basically, we need to be able to interrupt sounds that have already started. Without it, this feature has unpleasant effects.

When hovering over patients, it worked well. but it's not comfortable on lists. 

Try this PR and listen.

File: CorsixTH/Lua/dialogs/fullscreen/staff_management.lua (hunk 2 above)
Comment (TheCycoONE) [reply CorsixTH/Lua/dialogs/fullscreen/staff_management.lua]: I can, it echo's the code above in the hit test for onMouseDown; There is a list of staff members that are displayed, each row is 27 pixels tall, and each page of results has up to 10 entries, but obviously the final page has less.

So the check you commented on is to see if they are hovering in the list or past the end. Assuming they are in the list it grabs the current index in the list based on the mouse position and page.

This whole file is full of these repeating magic numbers - it could use a heavy refactor but it seems beyond the scope of this PR.

That said; now that you're finding the id under the mouse on hover you could get rid of finding it on mouse down (you already have it) - and maybe that's a good excuse to improve the readability of at least this block.

Comment (ARGAMX) [reply CorsixTH/Lua/dialogs/fullscreen/staff_management.lua]: > I can, it echo's the code above in the hit test for onMouseDown; There is a list of staff members that are displayed, each row is 27 pixels tall, and each page of results has up to 10 entries, but obviously the final page has less.
> 
> So the check you commented on is to see if they are hovering in the list or past the end. Assuming they are in the list it grabs the current index in the list based on the mouse position and page.

Ok, thanks.
We could probably write it that way.
Or at least put the result in a named Bool.
Because, unfortunately, I couldn't figure it out after reading this.
I suspect I'm not the only one who's that stupid on the earth. 😁 😁 😁

**Comment by TheCycoONE** [comment]
Ok, played through all of the changes.

I'm not bothered as much as @ARGAMX about the ringing when hovering over e.g. the staff list, though it's likely to happen in normal use and of the sounds it's the most controversial. The blue font is definitely nice.

The red box looks just like the original, so even if I don't love the style it gets +1 for faithfulness:
image
image

There is some commented out code that should be removed (is this WIP?)

The original played a different sound here when the change was invalid, e.g. - when already at 0, and CorsixTH with your patch does not.
image

**Comment by ARGAMX** [comment]
> The original played a different sound here when the change was invalid, e.g. - when already at 0, and CorsixTH with your patch does not.

That's true. But if you're referring to the "-" and "+" buttons from your screenshot, it seems like this PR and the Issue wasn't intended to fix these problems. There was an incorrect sound there before PR. Perhaps we should fix this in another PR. This PR seems to be growing rapidly and already going beyond of initial scope. 🙂

**Comment by TheCycoONE** [comment]
Fair

File: CorsixTH/Lua/dialogs/watch.lua [hunk 7]
```diff
@@ -84,6 +97,14 @@ function UIWatch:UIWatch(ui, count_type)
         self.scrollToTimerEventPatient, nil, self.cycleTimerEventPatient)
   else
     self:addPanel(timer_sprite, 0, 28):setTooltip(tooltips[count_type])
+    .custom_draw = --[[persistable:open_hospital_timer_button]] function(panel, canvas, x, y)
+      x = x + panel.x
+      y = y + panel.y
+      panel.window.panel_sprites:draw(canvas, panel.sprite_index, x, y)
+      if self.active_hover then
+        self.panel_sprites:draw(canvas, 17, x+4, y-28)
```
lewri: Space out your arithmetic e.g. `x + 4`
This is also magic.

CorsixTH/Lua/window.lua
```diff
@@ -1531,6 +1531,24 @@ function Window:hitTestPanel(x, y, panel)
...
 
+--[[ Used to test if cursor is currently hovering over certain area in dialog.
+!param active_hover (bool) Is cursor is already hovering over it?
+!params mouse_x, mouse_y (integers) Mouse X and Y coordinate
+!params min_x, max_x, min_y, max_y (integers) Coordinates of the area.
+]]
+function Window:hoverTest(active_hover, mouse_x, mouse_y, min_x, max_x, min_y, max_y)
+  local is_currently_hovering = mouse_x > min_x and mouse_x < max_x and mouse_y > min_y and mouse_y < max_y
+  if is_currently_hovering and not active_hover then
+    self.ui:playSound("HLight5.wav")
```
lewri: Arguably, I don't think Window should be calling the sound. The daughter class should be asking for the hover result, then actioning accordingly.

lewri:
All good things, hopefully we don't go ping mad after :P

CorsixTH/Lua/window.lua
TheCycoONE: I'm not sold on this design either but I think it can be refactored. It'd be best IMO if those rows were their own controls and they got the onMouseMove but used their knowledge of their own location data to implement onEnter / onExit events which were passed in like the buttons.

CorsixTH/Lua/dialogs/watch.lua
lewri: `offset_x` and `offset_y`?

I'm still a little iffed we have to almost identical persistables.

CorsixTH/Lua/window.lua
```diff
@@ -1531,6 +1531,24 @@ function Window:hitTestPanel(x, y, panel)
...
+    return true
+  else
+    if not is_currently_hovering then
+      return false
+    end
+    return true
```
lewri: Return the `is_currently_hovering` value (i.e. `return not is_currently_hovering`), isn't it already a boolean? The triple return is a little unsightly.

CorsixTH/Lua/window.lua
Matroftt: > Arguably, I don't think Window should be calling the sound. The daughter class should be asking for the hover result, then actioning accordingly.

daughter class is going to rapidly repeat the sound this way

CorsixTH/Lua/window.lua
```diff
@@ -1531,6 +1531,22 @@ function Window:hitTestPanel(x, y, panel)
   return false
 end
 
+--[[ Used to test if cursor is currently hovering over certain area in dialog.
+!param active_hover (bool) Is cursor is already hovering over it?
+!params mouse_x, mouse_y (integers) Mouse X and Y coordinate
```
TheCycoONE: The build failure is because of the use of `params` which is not supported by our documentation tool.

List each parameter independently with !param

integer is the type, not integers.

PR #3038, opened 2025-09-21
[RFC] Add IDs to adviser message strings

### LINKED ISSUE #3037: Allow adviser message strings to have an ID

Allow the adviser message strings to have an ID, so they could be filtered easier and the existing functions don't have to compare whole strings to, for example, avoid duplicate adviser messages.

This would make it easier to filter adviser messages in the adviser history dialog, for example, so unnecessary messages could be filtered and removed from that view (e.g.: messages related to staff placement).

No

0.69

No

ShiroAka:
*Fixes #3037 *

- Refactors the adviser messages system so the lang strings can have an ID and are easier to filter. 
- As a result, the existing "Adviser:checkForDuplicates" no longer compares whole strings. Instead it only compares integer IDs, which should be faster

### DIFF
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -18,6 +18,8 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
+require("languages.string_ids")
+
 --! The multi-purpose panel for launching dialogs / screens and dynamic information.
 class "UIBottomPanel" (Window)
 
@@ -413,7 +415,7 @@ end
 function UIBottomPanel:queueMessage(type, message, owner, timeout, default_choice, callback)
   -- Show a helpful message if there has been no messages before - only in campaign though
   if not self.ui.hospital.message_popup and tonumber(self.world.map.level_number) then
-    self.world.ui.adviser:say(_A.information.fax_received)
+    self.world.ui.adviser:say(_A.information.fax_received, AdviserStringIds.information.fax_received)
     self.ui.hospital.message_popup = true
   end
   local fax = {
@@ -811,8 +813,11 @@ function UIBottomPanel:giveResearchAdvice()
     end
   end
   local msg = can_build_research and _A.warnings.research_screen_open_1 or _A.warnings.research_screen_open_2
+
+  local msg_id = can_build_research and AdviserStringIds.warnings.research_screen_open_1 or AdviserStringIds.warnings.research_screen_open_2
+
   if not TheApp.using_demo_files then
-    self.ui.adviser:say(msg)
+    self.ui.adviser:say(msg, msg_id)
   end
 end
 
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -547,7 +547,7 @@ end
 
 --[[ Spawns the inspector who will walk to the reception desk. ]]
 function Epidemic:spawnInspector()
-  self.world.ui.adviser:say(_A.information.epidemic_health_inspector)
+  self.world.ui.adviser:say(_A.information.epidemic_health_inspector, AdviserStringIds.information.epidemic_health_inspector)
   local inspector = self.world:newEntity("Inspector", 2, 2)
   self.inspector = inspector
   inspector:setType("Inspector")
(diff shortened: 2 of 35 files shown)


TheCycoONE:
Did you encounter any issues or profile and see advisor message comparison as a hot spot?

This looks incorrect (ignores parameters passed to formatting like room name), more effort to maintain, and of questionable value as short string compares are admittedly not as fast as an integer compare, still very fast.

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -413,7 +415,7 @@ end
 function UIBottomPanel:queueMessage(type, message, owner, timeout, default_choice, callback)
   -- Show a helpful message if there has been no messages before - only in campaign though
   if not self.ui.hospital.message_popup and tonumber(self.world.map.level_number) then
-    self.world.ui.adviser:say(_A.information.fax_received)
+    self.world.ui.adviser:say(_A.information.fax_received, AdviserStringIds.information.fax_received)
```
Alberth289346, line 418: One of the first rules in software is not to duplicate information, since it can get out of sync due to changes in the program.

If you want to go here, hide your functionality such that a programmer here does not have to write the ID themselves. That is, instead, derive it in some way from the `_A.information.fax_received` text that they write.

CorsixTH/Lua/languages/string_ids.lua
```diff
@@ -0,0 +1,414 @@
...
+        placing_object_blocks_door    = 7,
+    },
+
+    cheats = {
+        th_cheat                      = 8,
+        roujin_on_cheat               = 9,
+        roujin_off_cheat              = 10,
+        norest_on_cheat               = 11,
+        norest_off_cheat              = 12,
+    },
```
Alberth289346, line 21: You have along list of consecutive numbers. The collection strings however is not static. New strings are added, old strings are retired.

Imagine we add an `advisor_cheat`. What number does it get?

(And no, renumbering all lines below this is not an option. It leads to large file changes that are not readable.)

----
As a suggestion, why not use the string name itself? (That is, why must it be translated to a number at all?)

Alberth289346: As a second reason why these tables with numbers don't work: When writing a savegame, all data in the running program is stored.

When you load a savegame (eg one year later), all data is restored to its original spot. If you store these numbers anywhere (eg inside your window class), you thus get numbers of strings as they were one year ago.
Obviously, those numbers are wrong since in the mean time, strings have been added and have been removed. A number that you store now for (by example) `advisor_cheat` may be used for the completely unrelated string `mutiplayer_disconnected` one year later.

CorsixTH/Lua/game_ui.lua
```diff
@@ -1033,102 +1033,202 @@ local function make_tutorial_phases()
...
-    _A.tutorial.object_in_invalid_position,            -- 14
-    { text = _A.tutorial.confirm_room,                 -- 15
+    {
+      text = _A.tutorial.place_objects,                         -- 13
+      speech_id = AdviserStringIds.tutorial.place_objects,
+    },
+    {
+      text = _A.tutorial.object_in_invalid_position,            -- 14
+      speech_id = AdviserStringIds.tutorial.object_in_invalid_position,
+    },
```
Alberth289346, line 1184: If you look at the original file, the `-- XY` numbers are nicely aligned vertically, to simplify reading.

In your edit, the numbers are not vertically aligned any more.

In general, try to avoid touching things around your modifications. Try to reduce the impact of your code changes for non-related parts as much as possible.

CorsixTH/Lua/epidemic.lua
```diff
@@ -547,7 +547,7 @@ end
 
 --[[ Spawns the inspector who will walk to the reception desk. ]]
 function Epidemic:spawnInspector()
-  self.world.ui.adviser:say(_A.information.epidemic_health_inspector)
+  self.world.ui.adviser:say(_A.information.epidemic_health_inspector, AdviserStringIds.information.epidemic_health_inspector)
```
Alberth289346, line 550: We generally limit line length to around 80 characters.

If it is longer, continue at the next line while inserting a double indent (ie 4 spaces). That is:
```
  self.world.ui.adviser:say(_A.information.epidemic_health_inspector,
      AdviserStringIds.information.epidemic_health_inspector)
```

(double indent ensure that the "then" code is findable, see, eg line702 below, where the condition needs multiple lines)

CorsixTH/Lua/dialogs/bottom_panel.lua
ShiroAka: tried to think of a better way, but I don't know all the ins and outs of Lua, and it is so simple it gets a bit hard to implement things that are present or easier in other languages

ShiroAka:
@Alberth289346 Thanks for taking the time to review the code and explain things. You are right, this changes too much and may cause issues in the future.
To be honest, I got carried away and made changes without thinking a lot, which is not ok. Lesson learned

I tested a bit more and it seems Lua can compare many "long" strings without affecting performance, so I should be able to implement the adviser's messages filtering without much trouble

If I can find a better way to do the same, I'll reopen this PR. But since the main issue got solved, this may not be needed.

PR #3039, opened 2025-09-21
Add pause movie support

TheCycoONE:
Another step towards headlines that I thought would be useful as it's own PR.

This adds the ability to pause and resume movies, by default with the space bar.


CorsixTH/Src/th_movie.cpp
```diff
@@ -522,6 +522,17 @@ void movie_player::play_audio(int requested_audio_channel) {
 
 void movie_player::stop() { aborting = true; }
 
+void movie_player::togglePause() {
+  bool wasPaused = paused.exchange(!paused);
```
TheCycoONE, line 532: This is so close to thread safe, if c++ had an atomic not, or if fetch_xor worked on bools I would have done it. As is I'd need a mutex or a compare_exchange loop and it didn't seem worth it since we only call from the event thread anyway.

MandrakeHorse:
While the majority of Theme Hospital's movies are only a couple of seconds long, with the intro being the longest video at 88 seconds, the ability to pause and resume movies is a commonly-requested accessibility feature. (And at the very least, it could be useful if someone wanted to take a screenshot of a particular frame.)

Bear in mind that if we ship this feature without pausing the audio (DICEYFIN.wav) with the level transition movies, we might subsequently receive a bug report about how pausing the board game movies desyncs the audio.

Speaking of the board game movie, it's worth mentioning that unlike with the intro/win/lose cutscenes, it is not possible to skip these cutscenes by clicking the left mouse button or pressing the escape key, presumably because the audio is being played separately. This might be worth pursuing as a separate issue (especially because as a tester, I personally find being unable to skip these cutscenes annoying enough to disable global movies.)

TheCycoONE:
I think pausing and skipping the board game are related. I prefer to defer that, but maybe plan to address both in 0.70 (it's just not my active focus at the moment)

lewri (commented):
I'm not sure I see a benefit in exposing a pause movie behaviour to an end user; especially if it causes quirks. For debugging, maybe.

Though I'm not against adding the function for wider goals.

CorsixTH/Src/th_movie.h
```diff
@@ -331,6 +331,12 @@ class movie_player {
   //! Stop the currently playing movie
   void stop();
 
+  //! Pause the currently playing movie
+  //!
+  //! This method is not thread safe, it should only be called from the event
+  //! thread.
```
lewri, line 337: Could you maybe align this comment to use wording within the function itself. You use 'close to thread safe' but then say 'not' here (arguably both are true, but the wording infer differences). You could point back to the function for deeper explanation, I'd already be going "why?" reading this anyway.

TheCycoONE: No. This describes the api (not thread safe) the other is an implementing comment explaining my thought process. A nearly thread safe implementation results in a not thread safe api. Nearly is only relevant if you're changing the function not if you are calling it.

TheCycoONE:
> I'm not sure I see a benefit in exposing a pause movie behaviour to an end user; especially if it causes quirks. For debugging, maybe. Though I also wouldn't stop it happening. Could it be `P` to mimic the in game pause? Or is the game still running in the background?
> 
> Though I'm not against adding the function for wider goals.

Space felt more natural to me but nothing stops using p. Maybe play with it and see what you think. "Quirks" already exist as @MandrakeHorse mentioned. Actually this same quirk.

TheCycoONE:
For the sake of this PR I will disable pause when can_skip is false (board game) - and for the issue we'll want to get rid of that flag for both #3040 

(do not merge until I add that)

TheCycoONE:
Done, pause is now p, and board game cannot be paused.

MandrakeHorse:
Happy for pause to be bound to P by default, since Space (and Enter) is often used to skip cutscenes in other games

tobylane:
Pause the intro movie then skip it, then start a campaign. The board game movie is paused, and the game is stuck.

TheCycoONE:
Thanks @tobylane

TheCycoONE:
Playing a movie now unpauses, fixing the issue @tobylane pointed out.

TheCycoONE:
Rebased on the latest. The refresh() method had to change a little due to the lose movie changes.

PR #3045, opened 2025-09-23
Lose movies newspaper headlines

### LINKED ISSUE #3020: [Bug] Newspaper headlines are missing from Lose cutscenes
### Describe the issue

The Lose movies do not include the randomly-selected tabloid headline ("SMASHED SURGEON", "BANK BYPASS OPERATION" etc) that Theme Hospital adds to these videos, resulting in the newspaper having a blank headline.

Lose movie with missing text

### Steps to Reproduce

1. Enable debug mode
2. Set Global Movie Control to On in Settings
3. Load any game
4. Open the Cheats menu
5. Click Lose Level

### Expected Behaviour

An appropriate tabloid headline is randomly-selected and displayed over the final frame of the various Game Over movies.

- [ ] Right mouse panning

### Save Game

### CorsixTH Version

0.69.0

### Operating System

Linux Mint 22.2 Cinnamon

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

### Additional Information

A list of tabloid headlines used by these cutscenes can be found here: https://gaming.stackexchange.com/questions/203627/how-do-i-know-why-i-failed-the-level-in-theme-hospital

TheCycoONE:
At a high level this PR brings support for the headlines that appear in the lose movies of the original game. It also takes a few steps towards supporting the win level letter where it made sense to do so.

It is intended that each commit be reviewed independently. Each has a message describing the changes.

Fixes #3020 

I am not satisfied with the scaling artifacts of the font when drawing on modern display resolutions. If anyone has suggestions there, I welcome them; otherwise I'm not certain it's possible to fix without much bigger rendering changes. True type fonts look pretty good here, and if pressed I'd probably bundle one or two suitable ttf fonts rather than try to fix arbitrary scaling the bitmap.

Screenshot 2025-09-29 at 20 35 36


CorsixTH/Lua/graphics.lua
```diff
@@ -208,11 +208,12 @@ end
 
 local function makeGreyscaleGhost(pal)
   local remap = {}
+  local stride = #pal / 256
   -- Convert pal from a string to an array of palette entries
   local entries = {}
-  for i = 1, #pal, 3 do
+  for i = 1, #pal, stride do
     local entry = {pal:byte(i, i + 2)} -- R, G, B at [1], [2], [3]
```
Alberth289346, line 216: Since it now computes stride, shouldn't it here either take `stride` bytes or check that `stride == 3` ?

Alberth289346 (approved):
Nice work!

ARGAMX (commented):
I have builded this branch, but on app run I get
missing `Bitmap/lose.pl8`

Image

I have inspected build and in `Bitmap` there is no `lose.pl8`. 
Although I see that this file was in sources `../CorsixTH/Bitmap/lose.pl8`.

I don't understand how these resources get into the build.
Perhaps they're in some build configs.
I created these configs a long time ago once and haven't updated them since.
Possible in `install_manifest` which I never update.
I only run `make CorsixTH` and `make install` each time I want to rebuild.

So, I manually added this `lose.pl8` to the build, and after that, it launched successfully.

Anyone can test this on this save:
[lose_level.sav.zip](https://github.com/user-attachments/files/22603341/lose_level.sav.zip)

Image

Letters itself looks good. Especially letters on Unicode fonts, looks just gorgeous. 
Even better than native letters on English and other non-Unicode.

English:
Screenshot 2025-09-29 at 20 35 36

French:
Screenshot 2025-09-29 at 21 14 36

French:
Screenshot 2025-09-29 at 21 14 59

Russian:
Screenshot 2025-09-29 at 20 44 44

Czech:
Screenshot 2025-09-29 at 20 45 43

Greek:
Screenshot 2025-09-29 at 20 39 37

Czech:
Screenshot 2025-09-29 at 20 40 05

Czech:
Screenshot 2025-09-29 at 20 40 29

Ukrainian:
Screenshot 2025-09-29 at 20 41 26

Russian:
Screenshot 2025-09-29 at 20 38 36

Russian:
Screenshot 2025-09-29 at 20 43 31

As you can see, due to frequent use diacritical marks in languages like Czech, text can slide down. Also several letters in Russian and Ukrainian also have such diacritical marks. One such letter is enough to make the entire title shift.

Also, some text may extend beyond the width of the borders. But since this still depends heavily on the chosen font, it's difficult to predict.

Nevertheless, the results of the work are quite impressive. 👍

TheCycoONE:
Thanks for the report. @ARGAMX - I think this build should resolve the missing files.

CorsixTH/Lua/graphics.lua
TheCycoONE: This is the grayscale mapping, I really don't know what to do with alpha here. However, I neglected to include it in palette and I've fixed that thanks to this comment.

TheCycoONE: For clarity here, it checks stride to find if the file it's reading is RGBARGBARGBA, then continues to take [RGB]A[RGB]A; but this spot is not what loads the palette for sprites, it's just looks for similar entries that are gray. I mostly needed it not to crash.

Alberth289346: Not something that has to be done here, just showing what I'd do in general:

----
I was mostly worried about `stride < 3` which would create a mess.

As a precaution I would add something like
``` lua
assert(stride >= 3, "Stride too small.") -- Ensured during loading of the file.
```
It makes sure you won't get messy results, and it points to where to look if the condition ever fails. As a side-effect, it makes clear that `/ 256` is supposed to work correctly without having to check.

----
In my work, correctness is way more important than not crashing, so I add things like this everywhere. The goal is to crash as fast as possible if something isn't right. 
While it sounds counter-intuitive, in my experience, that goal leads to solid programs that hardly ever crash. [We keep the checks in deployed code as well.]

While changing a program however, it often crashes at unexpected spots. Typically it then points to a case that I missed in the change.

(I think you get the solid behavior because there isn't much room left to compute with invalid data without crashing very quickly. This means you catch bugs quickly in the development cycle, and often near the code that is wrong.)

TheCycoONE: It should be unreachable with a file that isn't 768 or 1024 bytes exactly (palette load should fail).

However I see there is missing error handling for that condition on the lua side and as of the follow commit nothing to prevent the exception from unwinding through the binding. Will address before merging.

ARGAMX:
> Thanks for the report. @ARGAMX - I think this build should resolve the missing files.

Thanks, missing `Bitmap/lose.pl8` resolved. 👍

CorsixTH/Lua/graphics.lua
TheCycoONE: Fixed.

TheCycoONE:
@ARGAMX can you paste some of the text and fonts you used here. I'm not going to do anything about the size in this PR but I'd like to take one more attempt at placement.

ARGAMX (approved):

ARGAMX:
> @ARGAMX can you paste some of the text and fonts you used here. I'm not going to do anything about the size in this PR but I'd like to take one more attempt at placement.

Thin
[SF-Compact-Display-Regular.otf.zip](https://github.com/user-attachments/files/22641051/SF-Compact-Display-Regular.otf.zip)

Bold
[Roboto-Bold.ttf.zip](https://github.com/user-attachments/files/22641058/Roboto-Bold.ttf.zip)

ARGAMX:
> @ARGAMX can you paste some of the text and fonts you used here. I'm not going to do anything about the size in this PR but I'd like to take one more attempt at placement.

Czech
"DIVNÝ PRAKTICKÝ LÉKAŘ SI HRAJE NA BOHA"
"ZNIČUJÍCÍ VYŠETŘENÍ"
"ŠOK ŠPRÝMOŠTAINA"

Brazilian Portuguese
`"O MÉDICO AÇOUGUEIRO"`

Rus
"ВРАЧ УПОРОЛСЯ В ХЛАМ"
"ВРАЧ ГРЕБ ДЕНЬГИ ЛОПАТКОЙ"
"ОБЛАВА НА СУМАСШЕДШИХ ВРАЧЕЙ"

Ukr
"ЩО БУЛО НА ЛАБОРАТОРНОМУ СТОЛІ?"
`"ОПЕРАЦІЇ З ПЕРЕКАЗАМИ"`
"ДИВАКУВАТИЙ ТЕРАПЕВТ ДУМАЄ ЩО ВІН БОГ"

TheCycoONE:
Thanks!

TheCycoONE:
Thanks, I think this is an improvement. I now vertically center the TTF fonts in the headline area (as well as I can).

lewri (commented):
Merge once you're satisfied @TheCycoONE  =]

Issue #2991, opened 2025-09-01
[Bug] Half couch renders in test save


TheCycoONE:
### Describe the issue

As reported on discord/matrix. The sofa is only half rendered in some cases (regression)

This worked properly in 0.69

### Steps to Reproduce

1. Load the 0.69.1 test save: https://github.com/lewri/CorsixTH-Tests/blob/main/Test%20Saves/0.69.1%20test%20save.sav

2. Observe the sofa in the staff room

### Expected Behaviour

Couch should be fully rendered, but it is only half rendered.

### Have you enabled any current CorsixTH Experimental Features?

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

Save game attached to report: Yes

### CorsixTH Version

22c2546ccd7403fa963f7f95e50306ea9b71fc97

### Operating System

Arch Linux

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

### Additional Information

git bisect results:
```
git bisect start
# status: waiting for both good and bad commits
# good: [5c4e06397a1594e4c94c52533541608e6802208b] Bump version and add changelog (#2963)
git bisect good 5c4e06397a1594e4c94c52533541608e6802208b
# status: waiting for bad commit, 1 good commit known
# bad: [22c2546ccd7403fa963f7f95e50306ea9b71fc97] Merge pull request #2987 from Alberth289346/store-tile-in-anim
git bisect bad 22c2546ccd7403fa963f7f95e50306ea9b71fc97
# good: [194e4c5d58d52593b0e9983b21bc55af8e52c9d4] Format
git bisect good 194e4c5d58d52593b0e9983b21bc55af8e52c9d4
# bad: [b64a91b2017cb3567b2b844deccf3d6574c24a9b] Move th_sprite_properties to cpp
git bisect bad b64a91b2017cb3567b2b844deccf3d6574c24a9b
# good: [3d2b3b199658155d44355c10cee67d7e85e85672] Merge pull request #2958 from Alberth289346/add-drawing-layers-constants
git bisect good 3d2b3b199658155d44355c10cee67d7e85e85672
# bad: [d304aba1cc0e2d8ced12daeb22ce92f1d9a986a3] Merge pull request #2953 from Alberth289346/fix-tile-cropping
git bisect bad d304aba1cc0e2d8ced12daeb22ce92f1d9a986a3
[...]

PR #3048, opened 2025-09-23
Fix #3046 tooltips

### LINKED ISSUE #3046: [Bug] Tooltips don't work
Image

### Describe the issue

Tooltips used to appear when hovering over a UI element. 
They are no longer displayed in trunk-master.
They disappeared everywhere in the game.
Win criteria in Progress Report now unable to display as they were working on tooltips.

----

### Steps to Reproduce

1. Run the game.
2. Point cursor on any button in the main menu.
3. Observe that Tooltips don't appear.

----

### Expected Behaviour

Tooltips should appear as before.

----

### CorsixTH Version

[29bba7d1190cff77bf3ad5b052da5f83e50a2d2f](https://github.com/CorsixTH/CorsixTH/commit/29bba7d1190cff77bf3ad5b052da5f83e50a2d2f)

Regressed in #3026

TheCycoONE:
I was under the impression that the renderer logical size defaulted to the window size.

It seems that is mostly true, but when calling
SDL_RendererGetLogicalSize it returns a 0 rather than the window size.

This breaks a couple places where we ask for the dimensions of the window.

This will revert the behaviour change to #2991 as well.

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -424,9 +424,9 @@ render_target::scoped_target_texture::~scoped_target_texture() {
 }
 
 render_target::render_target(const render_target_creation_params& params)
-    : width(params.width),
-      height(params.height),
-      direct_zoom(params.direct_zoom) {
+    : width{params.width},
+      height{params.height},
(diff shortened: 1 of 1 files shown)


Alberth289346:
Seems fine to me.

PR #3060, opened 2025-10-04
Fix smoke machine position

### LINKED ISSUE #2872: [Bug] Machines smoke animations position misalign
Image

### Describe the issue

Machine smoke animations is mismatch the actual Machine position.

----

### Expected Behaviour

It should be like in the original Theme Hospital.

----

### Save Game

Inflation smoke animations is mismatch the actual machine position.
v0.69.0
[smoke_misalign_inflation.sav.zip](https://github.com/user-attachments/files/21667736/smoke_misalign_inflation.sav.zip)

Image

Cardio smoke animations is mismatch the actual machine position.
v0.69.0
[smoke_misalign_cardio.sav.zip](https://github.com/user-attachments/files/21667618/smoke_misalign_cardio.sav.zip)

Image

Scanner smoke animations is mismatch the actual machine position.
v0.69.0
[smoke_misalign_hair_restoration.sav.zip](https://github.com/user-attachments/files/21667619/smoke_misalign_hair_restoration.sav.zip)

Image

Hair Restoration smoke animations is mismatch the actual machine position.
v0.69.0
[smoke_misalign_scanner.sav.zip](https://github.com/user-attachments/files/21667629/smoke_misalign_scanner.sav.zip)

Image

Slicer
[room_crashed_doc_in_grab_surg.sav.zip](https://github.com/user-attachments/files/17475797/room_crashed_doc_in_grab_surg.sav.zip)

Image

----

### CorsixTH Version

v.0.68.0, v.0.69.1

marcassin92:
xray mp4

*Fixes #2872*

- Select the best animation for each machines
- Store the smoke animation position into `object.orientations` of each machine definition
- Apply the object translation to the smoke position, and mirror if necessary the animation

Note : 
- The smoke position isn't part of the footprint definition because some machine ( operating_table ) doesn't have a `(0, 0)` tile.
- An improvment could be to store the `x` and `y` offset/translation instead of applying it to `smoke_position` in _entities/object.lua_, and remove the `smoke_position` for every machines.
- Seeing that some animations are not nice from the Original game, maybe we can choose another better ( if possible ) ?
  - 3 machines (operating, blood, hair) use the same `3472` animation, when `3444` and `3452` arn't used.

Question for reviewers : 
- What's the rule about backward compatibility for saves ? should we try as possible to be non breaking ?
=> In the `getSmokeTile` function, `self.object_type` is used to get the smoke position but it can be found in `self.smoke_offset` (stored in the constructor), i did that (temporarily) to make it work for machines without this info ( ex: from an save without this fix ).

### DIFF
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -43,8 +43,6 @@ function Machine:Machine(hospital, object_type, x, y, direction, etc)
 
   self.total_usage = 0
   self:setHandymanRepairPosition(direction)
-
-  self.smoke_offset = object_type.orientations[direction].smoke_position
 end
 
 --!param room (object) machine room
@@ -81,10 +79,10 @@ end
 local function getSmokeTile(self)
   local map, _, _ = self.th:getTile()
   local x, y = self.tile_x, self.tile_y
-
-  if self.object_type.orientations[self.direction].smoke_position then
-    x = x + self.object_type.orientations[self.direction].smoke_position[1] -- todo : replace by self.smoke_offset
-    y = y + self.object_type.orientations[self.direction].smoke_position[2]
+  local smoke_position = self.object_type.orientations[self.direction].smoke_position
+  if smoke_position then
+    x = x + smoke_position[1]
+    y = y + smoke_position[2]
   end
 
   return map, x, y
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -43,6 +43,13 @@ function Machine:Machine(hospital, object_type, x, y, direction, etc)
 
   self.total_usage = 0
   self:setHandymanRepairPosition(direction)
+
+  self.smoke_offset = object_type.orientations[direction].smoke_position
+  if object_type.smoke_animation then
+    self.smoke_animation = object_type.smoke_animation
+  else
+    self.smoke_animation = 3428 -- Default to the (0,0) based animation
+  end
 end
 
 --!param room (object) machine room
@@ -77,19 +84,13 @@ end
 --! Return Tile position for smoke animation
 --!return (map, int, int)
 local function getSmokeTile(self)
-  local map, x, y = self.th:getTile()
-  if self.footprint then
-    for _, footprintCell in ipairs(self.footprint) do
-      if footprintCell.smoke_position then
-        x = x + footprintCell[1]
-        y = y + footprintCell[2]
-        break
-      end
-    end
-  end
+  local map, _, _ = self.th:getTile()
+  local x, y = self.tile_x, self.tile_y
 
-  -- adjust the offset of the original animation.
-  x = x - 1
+  if self.smoke_offset or self.object_type.orientations[self.direction].smoke_position then
+    x = x + self.object_type.orientations[self.direction].smoke_position[1] -- todo : replace by self.smoke_offset
+    y = y + self.object_type.orientations[self.direction].smoke_position[2]
+  end
 
   return map, x, y
 end
@@ -111,7 +112,8 @@ local function setSmoke(self, isSmoking)
       self.ticks = true
     end
     -- TODO: select the smoke icon based on the type of machine
-    self.smokeInfo:setAnimation(self.world.anims, 3424)
+    local mirror = self.direction == "east" and 1 or 0
+    self.smokeInfo:setAnimation(self.world.anims, self.smoke_animation or self.object_type.smoke_animation, mirror)
   else -- Otherwise, turning smoke off
     -- If there is currently a smoke animation, remove it
     if self.smokeInfo then
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -45,11 +45,6 @@ function Machine:Machine(hospital, object_type, x, y, direction, etc)
   self:setHandymanRepairPosition(direction)
 
   self.smoke_offset = object_type.orientations[direction].smoke_position
-  if object_type.smoke_animation then
-    self.smoke_animation = object_type.smoke_animation
-  else
-    self.smoke_animation = 3428 -- Default to the (0,0) based animation
-  end
 end
 
 --!param room (object) machine room
@@ -87,7 +82,7 @@ local function getSmokeTile(self)
   local map, _, _ = self.th:getTile()
   local x, y = self.tile_x, self.tile_y
 
-  if self.smoke_offset or self.object_type.orientations[self.direction].smoke_position then
+  if self.object_type.orientations[self.direction].smoke_position then
     x = x + self.object_type.orientations[self.direction].smoke_position[1] -- todo : replace by self.smoke_offset
     y = y + self.object_type.orientations[self.direction].smoke_position[2]
   end
@@ -98,7 +93,7 @@ end
 --! Set whether the smoke animation should be showing
 local function setSmoke(self, isSmoking)
   -- If turning smoke on for this machine
-  if isSmoking then
+  if isSmoking and self.object_type.smoke_animation then
     -- If there is no smoke animation for this machine, make one
     if not self.smokeInfo then
       self.smokeInfo = TH.animation()
@@ -111,9 +106,8 @@ local function setSmoke(self, isSmoking)
       -- tick to animate over all frames
       self.ticks = true
     end
-    -- TODO: select the smoke icon based on the type of machine
     local mirror = self.direction == "east" and 1 or 0
-    self.smokeInfo:setAnimation(self.world.anims, self.smoke_animation or self.object_type.smoke_animation, mirror)
+    self.smokeInfo:setAnimation(self.world.anims, self.object_type.smoke_animation, mirror)
   else -- Otherwise, turning smoke off
     -- If there is currently a smoke animation, remove it
     if self.smokeInfo then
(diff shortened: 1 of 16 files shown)


marcassin92:
Here are some 3 layers smoke animation from the Theme Hospital game for ref :
Note :
- Blood machine & Hair remover have strange animations
- small cliping/hiding issue with the operating table
- DNA fixer doesnt have strength, neither repair capabilities in the origin game. I didn't see any smoke after several earthquakes

hair_restorer mp4
blood mp4
cardio mp4
operating mp4
remov mp4
xray mp4
scanner mp4
slicer mp4
inflator mp4
shower mp4
ultrascan mp4
electrolyzer mp4
Jelly Moudler : See [comment](https://github.com/CorsixTH/CorsixTH/pull/3060#issuecomment-3369081490)

Alberth289346:
Looks good !   Thanks for the animations.

marcassin92:
Preview of current fix so far : 
Cliping/hiding issue with : 
- inflator
- hait restorer
- ultrascan
- blood machine

corsixth mp4

CorsixTH/Lua/entities/machine.lua
```diff
@@ -74,24 +76,38 @@ function Machine:isMachine()
...
+--! Return Tile position for smoke animation
+--!return (map, int, int)
+local function getSmokeTile(self)
+  local map, _, _ = self.th:getTile()
+  local x, y = self.tile_x, self.tile_y
+
+  if self.object_type.orientations[self.direction].smoke_position then
+    x = x + self.object_type.orientations[self.direction].smoke_position[1] -- todo : replace by self.smoke_offset
+    y = y + self.object_type.orientations[self.direction].smoke_position[2]
+  end
```
```suggestion
  local smoke_position = self.object_type.orientations[self.direction].smoke_position

  if smoke_position then
    x = x + smoke_position[1] -- todo : replace by self.smoke_offset
    y = y + smoke_position[2]
  end
```

Perhaps it will be nicer this way?

CorsixTH/Lua/objects/machines/dna_fixer.lua
```diff
@@ -30,7 +30,6 @@ object.build_preview_animation = 5070
 object.default_strength = 7
 object.crashed_animation = 3376 -- TODO correct?
 object.show_in_town_map = true
-object.smoke_animation = 3428 -- TODO
```
ARGAMX: > remove smoke animation for dna fixer. Add possible anim for jelly

Why we can't use smoke animation for DNA fixer?

marcassin92: I'm not able to select the DNA fixer for repair in TH original game, neither it has a smoke animation after several earth quakes. ( That's what i observed, i may be wrong)
Is that a bug from the original Game ? Or is it a choice that this machine never breaks ?

Note :  There are 14 machines for 13 animations

ARGAMX:
Jelly Moudler

Original TH:

[Screen Recording 2025-10-05 at 16.23.44.webm](https://github.com/user-attachments/assets/2cc5d24f-3ed6-4031-94e4-f8359cdd09de)

This PR:

Screenshot 2025-10-05 at 17 01 17

Still slightly offset.

ARGAMX:
Blood Machine

Original TH:

[Screen Recording 2025-10-05 at 16.27.24.webm](https://github.com/user-attachments/assets/8ca2fd88-1101-4ffc-8cd2-27d646ce26ac)

This PR:

Screenshot 2025-10-05 at 16 48 53

Although it looks strange, it is consistent with the original. ✅

Maybe we should try to position it better, not as in original? Is this possible?

ARGAMX:
Hair restorer

same strange as Blood Machine

Original TH:

[Screen Recording 2025-10-05 at 16.33.57.webm](https://github.com/user-attachments/assets/def1f918-d1a9-49a8-bf3e-061c1cf6edc9)

This PR:

Screenshot 2025-10-05 at 17 01 23

Although it looks strange, it is consistent with the original. ✅

I support the idea to make it better ( if possible )

ARGAMX:
> Question for reviewers :
> 
> * What's the rule about backward compatibility for saves ? should we try as possible to be non breaking ?
>   => In the `getSmokeTile` function, `self.object_type` is used to get the smoke position but it can be found in `self.smoke_offset` (stored in the constructor), i did that (temporarily) to make it work for machines without this info ( ex: from an save without this fix ).

Related to backward compatibility: 
I believe that saves from previous version should be loadable and playble.
If machine was smoking in old save, than it should smoke with this PR.
It is OK if smoke will be misaligned in old save with this PR.

marcassin92:
Thanks for the Jelly reference. I was able to set the good animation
image

CorsixTH/Lua/objects/machines/dna_fixer.lua
ARGAMX: > I'm not able to select the DNA fixer for repair in TH original game, neither it has a smoke animation after several earth quakes. ( That's what i observed, i may be wrong) Is that a bug from the original Game ? Or is it a choice that this machine never breaks ?
> 
> Note : There are 14 machines for 13 animations

Yes, the situation here is confusing.

In original TH DNA fixer don't have any functionality of wear and tear.
Thus In original TH DNA fixer is indestructible.

Is this a bug or why did they do this?..
I think they rushed this disease for some reason and didn't fully implement it. 
And in the end, they didn't have an animation for repairing this machine.
That's probably why they couldn't come up with anything more than making it indestructible. That way, there's no need to repair it, and no animation is needed.

But for some reason, CorsixTH didn't do the same.
I don't know why.
Someone decided to make the machine destructible.
Although there's no animation for it being repaired.
So you might notice in CTH that the handyman simply walks up to DNA fixer, it's instantly repaired, and he leaves.
Looks like a bug. But so far there has been no decision on how to do this better.

So, ever since CorsixTH decided to stand out and not follow original TH in this, we still need smoke for DNA Fixer. It's unclear what type of smoke, but we need it.

CorsixTH/Lua/entities/machine.lua
marcassin92: Totally agree. I was waiting an answer for the saves related question to clear that part.
I applied the change you suggest + removed the TODO comment.

lewri:
Fwiw, I'm not overly fussed if the smoke is a little off from the original. It's only a graphical effect to warn the player and as long as it's appropriately there I think we're good enough.

marcassin92:
Animation 3460 is a little better for both machines (Not sure if the cliping/hiding issue is related to the choice of animation)
 
blood3460 mp4
Screenshot from 2025-10-05 16-33-23
It's just a small change. Tell me if you have preferences.

ARGAMX:
> Animation 3460 is a little better for both machines (Not sure if the cliping/hiding issue is related to the choice of animation)
> 
> [ ](https://private-user-images.githubusercontent.com/1659596/497559310-ea0354bb-e3b5-4a1b-844d-fd1b7046c8e4.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTk2NzYwMzYsIm5iZiI6MTc1OTY3NTczNiwicGF0aCI6Ii8xNjU5NTk2LzQ5NzU1OTMxMC1lYTAzNTRiYi1lM2I1LTRhMWItODQ0ZC1mZDFiNzA0NmM4ZTQuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MTAwNSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTEwMDVUMTQ0ODU2WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZDFkZTkzNWEwYjliMDA0YTViZmEwZDNkZTc5NjdlYzllNTk4NjViYWVhNjZkNThiMWJkZGU1Yjg2ZGU4YjBlOSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.v8PQTfib1daZU-s9P7JWrrVix4pBT4RzTlk-AMh6Gt0) Screenshot from 2025-10-05 16-33-23 It's just a small change. Tell me if you have preferences.

I like this with Hair restorer 👍

CorsixTH/Lua/objects/machines/dna_fixer.lua
marcassin92: 3444 seems quite ok but has the cliping issue on one direction ( like a lot of others anim on this machine )
image
Screenshot from 2025-10-05 17-05-05

ARGAMX: > 3444 seems quite ok but has the cliping issue on one direction ( like a lot of others anim on this machine ) image Screenshot from 2025-10-05 17-05-05

Nice. I think it fits. 👍

ARGAMX:
> Animation 3460 is a little better for both machines (Not sure if the cliping/hiding issue is related to the choice of animation)
> 
> blood3460 mp4  It's just a small change. Tell me if you have preferences.

This Blood machine smoke is also probably better than how it now. 👍

ARGAMX (approved):

TheCycoONE:
Thanks @marcassin92!

PR #3062, opened 2025-10-05
Set emergency patient health to 0 after emergency timer expires

### LINKED ISSUE #3055: [Bug] Emergency patients can't die
Image

----

### Describe the issue

Emergency patients cannot die if they are in the room when the timer ends.

----

### Steps to Reproduce

1. Load savegame `emergency_patients_goes_toilet.sav` attached to this issue.
2. Press 5 and wait until the emergency timer ends.
3. Observe that some patients who were in the toilet did not die and continue to live.

----

1. Load savegame `emergency_cant_die_in_room.sav` attached to this issue.
2. Press on the Nurse in Pharmacy to open the Staff window.
3. Press 2 and wait until the next patient in yellow enters the Pharmacy.
4. Sack Nurse via Staff window.
5. Observe that the emergency timer just ended.
3. Observe that the patient in yellow leaves the Pharmacy and continues to live.

----

### Expected Behaviour

After timer end Emergency patients should die after they exit the room.

----

### Save Game

[emergency_patients_goes_toilet.sav.zip](https://github.com/user-attachments/files/22476507/emergency_patients_goes_toilet.sav.zip)

[emergency_cant_die_in_room.sav.zip](https://github.com/user-attachments/files/22476531/emergency_cant_die_in_room.sav.zip)

----

### CorsixTH Version

0.68.0, 0.69.1

----

marcassin92:
*Fixes #3055*

At the end of Emergency time : `patient:die()` method is not call because the patient is in a room, and it's not possible to know if the patient is surely going to be cured.
Setting the `health` attributes to 0.0 will make the patient died next time he's out of a room, and the day [end](https://github.com/CorsixTH/CorsixTH/blob/afdec16eb78837ea931f08f30d35d6d9286831a1/CorsixTH/Lua/entities/humanoids/patient.lua#L650-L654). Unless the patient is [cured](https://github.com/CorsixTH/CorsixTH/blob/afdec16eb78837ea931f08f30d35d6d9286831a1/CorsixTH/Lua/entities/humanoids/patient.lua#L349-L352) before leaving the room and then the `health` is set to 1, so no death.

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1136,8 +1136,13 @@ function Hospital:resolveEmergency()
   local rescued_patients = emer.cured_emergency_patients
   for _, patient in ipairs(self.emergency_patients) do
     if patient and not patient.cured and not patient.dead
-        and not patient.going_home and not patient:getRoom() then
-      patient:die()
+        and not patient.going_home then
+      if not patient:getRoom() then
+        patient:die()
+      else
+        -- Patient is considered dead, unless cured by staff, will die after leaving the room.
+        patient.attributes["health"] = 0.0
+      end
     end
   end
   local total = emer.victims


ARGAMX:
Oops, I assigned that task to myself and was working on it too. 🙂
I managed to fix it, but I haven't had time to make a PR for it yet. 🙂
I'll test this PR and think about what we should do in this situation. 🙂

marcassin92:
> Oops, I assigned that task to myself and was working on it too. 🙂

Damn my bad.
If your solution is working, use yours and close my PR :-)

CorsixTH/Lua/hospital.lua
```diff
@@ -1136,8 +1136,13 @@ function Hospital:resolveEmergency()
   local rescued_patients = emer.cured_emergency_patients
   for _, patient in ipairs(self.emergency_patients) do
     if patient and not patient.cured and not patient.dead
-        and not patient.going_home and not patient:getRoom() then
-      patient:die()
+        and not patient.going_home then
+      if not patient:getRoom() then
```
TheCycoONE, line 1140: patient.lua has an extra condition here:

```lua
 if not self:getRoom() and not self:getCurrentAction().is_leaving then
      self:setMood("dying5", "deactivate")
      self:die()
    end
```

I also wonder why this isn't considered at all for Patient:treatDisease which calls die() inside a room without a check.

Patient:die itself appears to handle being inside a room.

```lua
  if self:getRoom() then
    self:queueAction(MeanderAction():setCount(1))
  else
    self:setNextAction(MeanderAction():setCount(1))
  end
```

What I'm saying is this condition seems suspect to me; I'm not sure the motivation for having room checked in some places and not others. My best guess is to give the treatment a chance of being successful. If that's the case maybe a comment, or some logic consolidation in patient.

patient:dieIfNotTreated or similar.

TheCycoONE: @ARGAMX if you're also working on this consider the same question. It might be all right as is.

marcassin92: I agree that the idea for this condition `getRoom()` may be to let the patient being possibly cured.
I've tested on the original TH Game in case of emergency: If the patient is in process of being cured when the countdown finish, the patient will survive (still not counted in the reward). If you remove the staff ( right click ), patient will stop the cure animation, leave the room and die.
> patient.lua has an extra condition here:

I guess the logic is the same for a patient being cured, but that doesn't make sense for a patient in a diagnosis room.

Note : I didn't succeed to have the countdown finish as soon as the patient has opened the door, as it seems that patients start dying few seconds before the countdown if not being cured.

TheCycoONE: From what I can tell, places where the patients are skipped and set to health 0 are killed at the end of the day after leaving in CorsixTH.

ARGAMX: The thing is that dying only at the end of the day is also a suspicious solution to the problem as a whole.
Emergency patients should die instantly as soon as they can. Not at the end of the day.
If they will start dying process at the end of the day then player have a time gap to sack patient and avoid dead.
For example, if the Emergency timer runs out while the patient is in the room, we can remove the doctor from the room. After that the patient leaves the room, and there's a time window until the end of the day.
Original TH never allow this.
Moreover At this moment we can set doctor back to the room and patient will enter again. And this strange situation can continue indefinitely.
I am preparing PR which will resolve this correctly.

ARGAMX: > Note : I didn't succeed to have the countdown finish as soon as the patient has opened the door, as it seems that patients start dying few seconds before the countdown if not being cured.

As far as I remember from the original, the timer simply can't run out until at least one patient is not cured and not at the dying procedure in hallway.
Correct me if I'm wrong.

CorsixTH/Lua/hospital.lua
```diff
@@ -1136,8 +1136,13 @@ function Hospital:resolveEmergency()
...
+        patient:die()
```
lewri, line 1141: Might I suggest, instead of letting the emergency kill the patient, just set health to 0 and let the tickDay function handle it? It already deals with health at 0 at the end of day.

TheCycoONE: We don't love that behavior, though. That particular check may move to tick rather than tickDay at which point setting health to zero is likely good.

CorsixTH/Lua/hospital.lua
ARGAMX: > @ARGAMX if you're also working on this consider the same question. It might be all right as is.

So, I looked into it and here's what I think:

```lua
 if not self:getRoom() and not self:getCurrentAction().is_leaving then
      self:setMood("dying5", "deactivate")
      self:die()
    end
```
I refactored and removed this part from here in #3113.

Related to:
```lua
  if self:getRoom() then
    self:queueAction(MeanderAction():setCount(1))
  else
    self:setNextAction(MeanderAction():setCount(1))
  end
```
Agree, this condition seems suspect.
`Patient:die` for room cases only can handle the state when then manipulations with patient are ended.
If we call `Patient:die` when patient entering the room die process will fail. (This is actually the problem with these bugs)
However, calling `Patient:die` in the corridor is a mosly working scheme.

I also think that something is somehow confused here and the names are not very clear.
I don't know what to do with it.
Also problem here is that it manipulating the patient action queue.
It's not great, but without getting involved in refactoring this, it seems like there's nothing much that can be done.
Unless to write a comment somewhere.

ARGAMX:
PR #3113 merged, which resolves not only 3055, but also 1584, which has a similar problem.
Also it makes Emergency patients to die as soon as they can, not only the next game day. 
I am closing this PR in prefer to PR #3113. 
@marcassin92 thank you for spending effort on it. 👍 🙏

PR #3063, opened 2025-10-05
Refactor to use font_options table

TheCycoONE:
With a growing number of options, passing as method arguments was becoming unwieldy.

The table is a bit more forgiving with ttf_color, it can now be specified as a single rgb or argb number i.e. 0xRRGGBB or 0xAARRGGBB if desired.

There are a couple new features:
* Alpha components on the ttf_color are now supported.
* Overriding the font size is now possible with ttf_width and ttf_height.


Alberth289346:
Seems fine to me

TheCycoONE:
@Alberth289346 lewri added a rule that PRs have to be approved, are you able to check the box (which I think is a button, github's UI here is confusing)

PR #3066, opened 2025-10-09
Add ttf shadow option

TheCycoONE:
The shadow supports arbitrary pixel offsets and color, and extends the text box accordingly. The default settings match the common shadow used in CorsixTH.

The default UI font now uses the shadow which helps it pop.

### DIFF
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -1693,25 +1693,49 @@ void freetype_font::free_texture(cached_text* pCacheEntry) const {
   }
 }
 
-void freetype_font::make_texture(render_target* pEventualCanvas,
+void freetype_font::make_texture(const render_target* pEventualCanvas,
                                  cached_text* pCacheEntry) const {
-  uint32_t* pPixels = new uint32_t[pCacheEntry->width * pCacheEntry->height];
-  std::memset(pPixels, 0,
-              pCacheEntry->width * pCacheEntry->height * sizeof(uint32_t));
-  uint8_t* pInRow = pCacheEntry->data;
-  uint32_t* pOutRow = pPixels;
-  uint32_t colorBase = font_color & 0xFFFFFF;
-  uint32_t alphaBase = (font_color >> 24) & 0xFF;
-  for (int iY = 0; iY < pCacheEntry->height;
-       ++iY, pOutRow += pCacheEntry->width, pInRow += pCacheEntry->width) {
-    for (int iX = 0; iX < pCacheEntry->width; ++iX) {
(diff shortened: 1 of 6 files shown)


CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -1693,25 +1693,49 @@ void freetype_font::free_texture(cached_text* pCacheEntry) const {
...
+  const uint32_t alphaBase = (color >> 24) & 0xFF;
+
+  const uint8_t* inData = cacheEntry->data;
+
+  for (int iY = 0; iY < cacheEntry->height; ++iY) {
+    for (int iX = 0; iX < cacheEntry->width; ++iX) {
+      if (const int inIndex =
+              (iY - offsetY) * cacheEntry->width + (iX - offsetX);
+          inIndex >= 0 && inIndex < cacheEntry->width * cacheEntry->height &&
+          inData[inIndex]) {
```
Alberth289346, line 1733: An assignment inside the if statement?

I am not even sure why the compiler accepts that. Moving the assignment to before the `if` seems a better solution to me.

(Maybe it's me, but I don't like side-effects in tests.)

CorsixTH/Src/th_lua_gfx.cpp
```diff
@@ -356,7 +374,29 @@ int l_freetype_font_set_font_options(lua_State* L) {
...
+  if (lua_istable(L, -1)) {
+    shadowOptions.enabled = true;
+
+    lua_getfield(L, -1, "offset_x");
+    shadowOptions.offset_x = static_cast<int>(luaL_optinteger(L, -1, 1));
+    lua_pop(L, 1);
+
+    lua_getfield(L, -1, "offset_y");
+    shadowOptions.offset_y = static_cast<int>(luaL_optinteger(L, -1, 1));
+    lua_pop(L, 1);
```
Alberth289346, line 388: Thanks, this way of getting the data makes much more sense than the above `read_color_from_lua`.

CorsixTH/Src/th_gfx_sdl.cpp
TheCycoONE: It was added in C++17 https://skebanga.github.io/if-with-initializer/

if became:
`if (init; condition)`, similar to `for (init; cond; stmt)`

Alberth289346: That example doesn't convince me at all. For the general solution you need a loop, and then the scoping problem doesn't exist at all.

Also, often init is much longer, and then the `if` line isn't a good spot for it anyway.

So, at first sight this idea looks like an edge-case, which adds more confusion than improve things.

But enough bashing. If you want it, go ahead.

PR #3071, opened 2025-10-11
Refactor Staff:tickDay()

lewri:
*Fixes #*

- General refactor of repeated code
- I've also made a change where as long as you are resting (and not greatly tired), your happiness will go up rather than be negatively impacted. This may have been part of the impact on staff could get unhappy to easily. However, it could be dropped out.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -52,17 +52,18 @@ function Staff:tickDay()
 
   -- if you overwork your Dr's then there is a chance that they can go crazy
   -- when this happens, find him and get him to rest straight away
-  if not self:isVeryTired() and self:isResting() then
+  if self:isVeryTired() or not self:isResting() then
+    -- Working when you should be taking a break will make you unhappy
+    if self:getAttribute("fatigue") >= self.hospital.policies["goto_staffroom"] then
+      self:changeAttribute("happiness", -0.02)
+    end
+    -- You will also start to become unhappy as you become tired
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -72,32 +73,31 @@ function Staff:tickDay()
...
+
+  -- Seeing various nearby objects boost your happiness, some more than others
+  local good_objects = {
+    ["extinguisher"] = 0.002, -- Makes you feel safe
+    ["bin"]          = 0.001,
+    ["bookcase"]     = 0.003,
+    ["skeleton"]     = 0.002,
+    ["tv"]           = 0.0005,
+  }
+  for obj_name, happiness_score in pairs(good_objects) do
```
Alberth289346, line 85: This is about happiness, so `good_object` likely follows from that idea.

However, 'good' is also covered in the positive numbers. As such, the 'good_' prefix is not really needed. Discarding the prefix would open the path to adding bad objects to the same table, if so desired.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -72,32 +73,31 @@ function Staff:tickDay()
...
+    self.world:findObjectNear(self, obj_name, 2, function()
+      self:changeAttribute("happiness", happiness_score)
```
Alberth289346, line 87: It's a bit of a waste to perform a full search on the same area for every object type. I looked for a better implementation but failed to find one.
Not sure it exists at all.

PR #3072, opened 2025-10-15
Fix сrash when edit room while patient undressing

### LINKED ISSUE #3051: [Bug] Crash when try to edit cardio room while patient undressing
Image

### Describe the issue

Game crashes when go to edit cardio room while patient undressing.

As posted to discord/matrix by Mickmane:

> I took doc out of cardio when I wanted to move cardio room, not seeing the patient inside. Removed doc, clicked to unbuild it, saw zzz..., announcer said doctor immediately wanted in cardio. Dumped doc back in. Now you can see the poor lady in her underwear heading for the door. :))

### Steps to Reproduce

1. Load savegame attached to this issue.
2. Press 1 to unpause the game.
3. While the patient started to use the Change Dress Screen in the Cardio Room, before she put her white item on screen, press “Edit Room” at the bottom and select Cardio Room.
4. Cursor becomes “Zzzzz” as there doctor and patient in the room.
5. Observe how the patient dresses back and leaves the room. And after she is outside of the room she suddenly teleports back to the room undressed, goes to the door and game crashes.

[Image](https://github.com/user-attachments/assets/561fca4d-81d4-4bab-9795-28d208d5289f)

### Expected Behaviour

Game shouldn't crash.

### Save Game

[3051 Cardio Autosave-6-22.sav.zip](https://github.com/user-attachments/files/22528228/3051.Cardio.Autosave-6-22.sav.zip)

### CorsixTH Version

v0.69.1

### Operating System

Windows 10 Home

### Gamelog.txt

```yaml
An error has occurred!

Error in timer handler: 
...ontents/MacOS/../Resources/Lua/humanoid_actions/walk.lua:342: Humanoid (Stripped Female Patient 3) without door animations trying to walk through door (from cardiogram to corridor)
stack traceback:
	[C]: in function 'error'
	...ontents/MacOS/../Resources/Lua/humanoid_actions/walk.lua:342: in function <...ontents/MacOS/../Resources/Lua/humanoid_actions/walk.lua:263>
	(...tail calls...)
	...ontents/MacOS/../Resources/Lua/humanoid_actions/walk.lua:241: in local 'timer_function'
	...orsix.Lab.app/Contents/MacOS/../Resources/Lua/entity.lua:223: in method 'tick'
	...Corsix.Lab.app/Contents/MacOS/../Resources/Lua/world.lua:911: in method 'onTick'
	...t/Corsix.Lab.app/Contents/MacOS/../Resources/Lua/app.lua:1311: in function <...t/Corsix.Lab.app/Contents/MacOS/../Resources/Lua/app.lua:1308>
	(...tail calls...)
	...t/Corsix.Lab.app/Contents/MacOS/../Resources/Lua/app.lua:1211: in function <...t/Corsix.Lab.app/Contents/MacOS/../Resources/Lua/app.lua:1206>
```

ARGAMX:
*Fixes #3051*

- Before setting `setNextAction(use_screen)` check if it's not already in the actions queue.

### DIFF
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -1001,7 +1001,7 @@ function Room:makeHumanoidDressIfNecessaryAndThenLeave(humanoid)
 
     if humanoid:getCurrentAction().name == "use_screen" then
       --The humanoid must be using the screen to undress because this isn't a leaving action:
-      if not humanoid:isDressingAndLeaving() then
+      if not humanoid:hasDressingAndLeavingAction() then
         humanoid:getCurrentAction().after_use = nil
         humanoid:setNextAction(use_screen)
       end


lewri:
Your pr branch has used a Cyrillic c instead of a normal c apparently? I'm not aware of that causing issues though?

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -662,6 +662,16 @@ function Humanoid:hasLeavingAction()
   return false
 end
 
+-- Check if there is "dress is_leaving" action in the action queue
+function Humanoid:isDressAndLeaving()
```
lewri: I'm tempted to say this function should live in room, because you're asking about the humanoid state in regard to the room's objects. 

If room is the right place, function name becomes `isHumanoidDressingToLeave`

The only reason I say this is it feels like it should live where we also have this such function 
`makeHumanoidDressIfNecessaryAndThenLeave`

ARGAMX:
> Your pr branch has used a Cyrillic c instead of a normal c apparently? I'm not aware of that causing issues though?

Yeap.

I like that GitHub didn't warn about this beforehand, only after the branch was already created, and it didn't give any options to do anything about it. At least I can't do anything.

I renamed the original branch, but it didn't affect the PR at all.

I don’t understand what it wants from us with this inscription?

If this is a problem, why wasn't there a warning at the creation stage?

Strange late warning.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > I'm tempted to say this function should live in room, because you're asking about the humanoid state in regard to the room's objects.
> 
> If room is the right place, function name becomes `isHumanoidDressingToLeave`
> 
> The only reason I say this is it feels like it should live where we also have this such function `makeHumanoidDressIfNecessaryAndThenLeave`

The thing is that, technically, the `action.name == "use_screen"` has no relation to the room, since the term `use_screen` isn't part of the room.
We think so because we know there's a screen in the room.
But the code does not know this.
It's just the name of the action, and from a code perspective, there's no connection to the room at all.

While `makeHumanoidDressIfNecessaryAndThenLeave` function manages the patient and it know its for sure.
It digs into patient action queue, make search in it and adds something to it. I don't much like this and I think it shouldn't have been done this way. It violates OOP encapsulation principles.
Although this is already the popular case throughout our code, it's already difficult to resolve it everywhere.
But I am not very happy to the spread of this problem even more.
In fact, this very bug is a consequence of that approach: 
### the responsibility for managing the patient queue is spread throughout the code.
There is no single center of responsibility for patient queue management.
That's why we have so many bugs with the patient queue, and they're incredibly difficult to debug and fix.
And each fix like this both fixes something and adds new bugs.

In this perspective `Humanoid:isDressAndLeaving()` is a more healthier function, since it works in its own area of ​​responsibility, in self object, and does not dig around in the room or some other object.

Alberth289346:
> I like that GitHub didn't warn about this beforehand, only after the branch was already created, and it didn't give any options to do anything about it. At least I can't do anything.

Version control things are all done at git level. GH isn't much involved there. GH is essentially a website layer on top of (many) git repos for extra features.

CorsixTH/Lua/entities/humanoid.lua
TheCycoONE: The grammar is off for english; isDressedAndLeaving or isDressingAndLeaving; based on the body probably the second.

ARGAMX: > The grammar is off for english; isDressedAndLeaving or isDressingAndLeaving; based on the body probably the second.

ok, fixed 👌

CorsixTH/Lua/room.lua
```diff
@@ -1000,9 +1001,12 @@ function Room:makeHumanoidDressIfNecessaryAndThenLeave(humanoid)
 
     if humanoid:getCurrentAction().name == "use_screen" then
       --The humanoid must be using the screen to undress because this isn't a leaving action:
-      humanoid:getCurrentAction().after_use = nil
-      humanoid:setNextAction(use_screen)
+      if not humanoid:isDressAndLeaving() then
```
ARGAMX:
```suggestion
      if not humanoid:isDressingAndLeaving() then
```

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -662,6 +662,16 @@ function Humanoid:hasLeavingAction()
...
+    for _, action in ipairs(self.action_queue) do
```
ARGAMX:
```suggestion
function Humanoid:isDressingAndLeaving()
  for _, action in ipairs(self.action_queue) do
```

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -662,6 +662,16 @@ function Humanoid:hasLeavingAction()
   return false
 end
 
+-- Check if there is "dress is_leaving" action in the action queue
```
lewri: More specifically from a code perspective, you mean are they using a screen during their leave cycle? We interpret it as undressing, however.

```suggestion
-- Check if the humanoid's leaving process also involves using a screen to get dressed.
```

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -662,6 +662,16 @@ function Humanoid:hasLeavingAction()
...
+function Humanoid:isDressingAndLeaving()
+  for _, action in ipairs(self.action_queue) do
+    if action.name == "use_screen" and action.is_leaving then
+      return true
```
lewri, line 669: `Humanoid:isLeaving()` only looks at the first action in the queue. Is there a reason why we're checking on any action? In other parts of our code, we only ever check the leaving flag on the current action.

Depending how the actions are constructed, couldn't that mean an action inserted has the leaving flag when they aren't actually in the state of leaving yet?

CorsixTH/Lua/entities/humanoid.lua
lewri: My other comment would make this suggestion technically wrong, if we're not meant to be grabbing any action.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > `Humanoid:isLeaving()` only looks at the first action in the queue. Is there a reason why we're checking on any action? In other parts of our code, we only ever check the leaving flag on the current action.
> 
> Depending how the actions are constructed, couldn't that mean an action inserted has the leaving flag when they aren't actually in the state of leaving yet?

At the same time it have `Humanoid:hasLeavingAction()` which is not looks at the first action in the queue.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > More specifically from a code perspective, you mean are they using a screen during their leave cycle? We interpret it as undressing, however.

Yes, perhaps "use_screen" would be more appropriate here instead of "dress".

ARGAMX:
```suggestion
-- Check if there is "use_screen is_leaving" action in the action queue
```

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > `Humanoid:isLeaving()` only looks at the first action in the queue. Is there a reason why we're checking on any action? In other parts of our code, we only ever check the leaving flag on the current action.
> 
> Depending how the actions are constructed, couldn't that mean an action inserted has the leaving flag when they aren't actually in the state of leaving yet?

Perhaps `hasDressingAndLeavingAction` is more correct name for this function.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -662,6 +662,16 @@ function Humanoid:hasLeavingAction()
   return false
 end
 
+-- Check if there is "use_screen is_leaving" action in the action queue
+function Humanoid:isDressingAndLeaving()
```
ARGAMX:
```suggestion
function Humanoid:hasDressingAndLeavingAction()
```

CorsixTH/Lua/room.lua
```diff
@@ -1000,9 +1001,12 @@ function Room:makeHumanoidDressIfNecessaryAndThenLeave(humanoid)
 
     if humanoid:getCurrentAction().name == "use_screen" then
       --The humanoid must be using the screen to undress because this isn't a leaving action:
-      humanoid:getCurrentAction().after_use = nil
-      humanoid:setNextAction(use_screen)
+      if not humanoid:isDressingAndLeaving() then
```
ARGAMX:
```suggestion
      if not humanoid:hasDressingAndLeavingAction() then
```

CorsixTH/Lua/entities/humanoid.lua
lewri: Yes I agree with that name.

lewri:
LGTM @TheCycoONE can you check the Cyrllic c won't have an impact for us before merging?

TheCycoONE (approved):

PR #3074, opened 2025-10-17
Only visit the same object tile only once in a search.

### LINKED ISSUE #3070: World:findObjectNear / pathfinder.findObject -- callback can trigger twice
I've generally noticed this while refactoring for something else. If you pass a callback to the `World:findObjectNear` function, it has the potential to trigger twice.

Currently I've only noticed this when looking for objects as part of `Staff:tickDay()`. I don't have any code for a definitive reproduction but some hacky prints show it triggered twice when looking for a nearby skeleton (the callback is to change attribute):

```yaml
Attribute change for happiness at amount 0.002 because of skeleton
table: 00000210619AF200 [Humanoid]
1-09-04T49 [Date]
Attribute change for happiness at amount 0.002 because of skeleton
table: 00000210619AF200 [Humanoid -- same one!]
1-09-04T49 [Date -- also same]
```

This is from a dev version of 0.69.1

Another hacky print from earlier, where the 3rd line is at the end of a loop, shows the attribute change went through twice before the loop end:
```yaml
Attribute change for    happiness       at amount       0.003
Attribute change for    happiness       at amount       0.003
bookcase        0.003
```

Alberth289346:
*Fixes #3070 

- Skip visiting the same neighbour tile in each object search.
- Doing this in Lua was simpler than in CPP.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -83,8 +83,16 @@ function Staff:tickDay()
     ["tv"]           = 0.0005,
   }
   for obj_name, happiness_score in pairs(good_objects) do
-    self.world:findObjectNear(self, obj_name, 2, function()
-      self:changeAttribute("happiness", happiness_score)
+    -- findObjectNear() returns the object from every reachable direction.
+    -- To eliminate double counting, visited_positions only allows one visit
+    -- to every tile in each search.
+    local visited_positions = {}
+    self.world:findObjectNear(self, obj_name, 2, function(x, y)
+      local xy = x + y *256
+      if not visited_positions[xy] then
+        self:changeAttribute("happiness", happiness_score)
+        visited_positions[xy] = true
+      end
     end)
   end
 


lewri:
To note, `Patient:tickDay()` is also a casualty of 3070.
https://github.com/CorsixTH/CorsixTH/blob/69dd11350d5c9ccdaf4c6789e373a14dc5b47a5a/CorsixTH/Lua/entities/humanoids/patient.lua#L693-L740

Less so for vomitings, more so for the other parts further down.

Edit: I also think the plant lookup in Staff:tickDay also falls foul of this too.
(cross posting in the actual issue ticket)

TheCycoONE:
I would like to look at the c++ side before accepting a workaround in the callback. Give me the weekend?

Alberth289346:
> I would like to look at the c++ side before accepting a workaround in the callback. Give me the weekend?

Sure.

For actually registering visits in C++:
* The probably simplest path in cpp is to use the `visited` property of path nodes (line 84 `th_pathfind.h`). You'd set that property when the callback is performed.  
That field however has the semantics `True if the cell has already been visited (popped from the open heap)`, which not something that is known at that point.
* Another option is to add a bit to the `path_node` class for tracking visits, or
* Add an XY-map to the `object_visitor` class, similar to what Lua does now.

A minor secondary hurdle is the call interface from Lua. Somewhere it's necessary to express how to visit the objects. The interface already has a lot of parameters.
* This morning however, I had the thought to add a new Lua visit function for this case. That would avoid adding more parameters.
* Splitting the call into several separate calls is another path.

Alberth289346:
A function to look at all tiles in a given rectangle might work.

TheCycoONE:
Maybe I shouldn't have looked at this; now I'm confused why we're using euclidean distance instead of manhattan, and why it's not using just a basic breadth-first/dijkstra search; and since most of the use cases only want a single result, why not let it exit early....

TheCycoONE:
Yes the particular case of happiness affected by what is near by could use the entity map already in lua; path finding is irrelevant. Line of sight would add a little realism but completely unnecessary.

Alberth289346:
> Maybe I shouldn't have looked at this; now I'm confused why we're using euclidean distance instead of manhattan, and why it's not using just a basic breadth-first/dijkstra search; and since most of the use cases only want a single result, why not let it exit early....

It is standard Dijkstra:
`int object_visitor::guess_distance(path_node* pNode) { return 0; }`
The heuristic doesn't steer at all. 

We do that in most classes, so path finding often isn't very fast at all. Lots to gain if we want.

Possibly the worst offender is at the start of a level, where a path to the hospital is attempted from each edge tile of the map. There is also an existence check for such a path somewhere (but could be the same function), for sanity checking of a map after editing.

TheCycoONE:
Could move the callback to the main loop (where visited is set to true) and calculate the direction from prev. That seems reasonably clean and keeps the semantics of visited.

Edit: I suppose the concern there is that the object isn't passable so it wouldn't be added to the list of nodes; but since we overload try_node, it could be.

Edit 2: However if you did you could travel through an object to reach another object, unless you added a special provision for that, and now it becomes less clean. So we are left reaching for another 'seen' property. Seen isn't awful -

Alberth289346:
I switched to implementing a Lua solution using the `entity_map`.

PR #3076, opened 2025-10-18
Eliminate last layer

Alberth289346:
Code had a magic connection between `setDrawingLayer` and `setTile` through the `last_layer` local variable:
``` c++
/* this variable is used to determine the layer of the animation, it should be
  rewritten at some
  point so that the it is passed as an argument in the function l_anim_set_tile
*/
int last_layer = 2;
```
Eliminating it reduces magic, which is never a bad thing.

### DIFF
--- a/CorsixTH/Lua/utility.lua
+++ b/CorsixTH/Lua/utility.lua
@@ -228,7 +228,8 @@ DrawingLayers.AtomAnalyser = 3
 DrawingLayers.ReceptionistFacingUser = 3 -- Facing east or south.
 DrawingLayers.Entity = 4 -- All 'normal' animations.
 DrawingLayers.ReceptionistFacingAway = 5 -- Facing west or north.
--- Values 6 and 7 not used.
+-- Value 6 is not used.
+DrawingLayers.FloatingDollars = 7
 DrawingLayers.EastSideObject = 8
 DrawingLayers.SouthSideObject = 9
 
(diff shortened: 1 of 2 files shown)


TheCycoONE:
Could you merge the middle 3 commits into one, so that full refactor is together?

CorsixTH/Src/th_lua_anims.cpp
```diff
@@ -396,17 +396,16 @@ int l_anim_set_tile(lua_State* L) {
...
-      pAnimation->attach_to_tile(x, y, node, last_layer);
+      pAnimation->attach_to_tile(x - 1, y - 1, node, last_layer);
     } else {
-      luaL_argerror(L, 3,
-                    lua_pushfstring(L,
-                                    "Map index out of bounds (" LUA_NUMBER_FMT
-                                    "," LUA_NUMBER_FMT ")",
-                                    lua_tonumber(L, 3), lua_tonumber(L, 4)));
+      // Off-map, report an error.
+      std::string msg = "Map index out of bounds (" + std::to_string(x) + ", " +
```
TheCycoONE, line 399: In C++20 this will be std::format - I look forward to it; for now I suppose we might as well do it this way; sstream is kinda annoying.

Alberth289346: I tried `std::format` and even installed the newest `clang-20` of the distribution, but it didn't work.
Apparently, the version number is not related to the cpp version.

Alberth289346:
> Could you merge the middle 3 commits into one, so that full refactor is together?

Done.

CorsixTH/Src/th_lua_anims.cpp
TheCycoONE: no, it's the CXX_STANDARD in CMake that drives that

TheCycoONE (approved):

PR #3085, opened 2025-10-19
Show every level's criteria in progress report by сompacting

### LINKED ISSUE #2750: [Bug] Progress report don't show unfulfilled criteria.
### Describe the issue

If the player kills more than a certain percentage of patients, then in Progress report instead of showing required hospital cost, the death rate criterion is displayed. So death criteria overlaps hospital cost criteria.
Thus, the player does not see hospital cost criterion at all.
If player fulfills all the other 4 criteria, he will be in a situation where all the displayed criteria are fulfilled, but the level does not end.
And the player doesn't understand why.

Screenshot 2024-12-26 at 21 31 13

Example of such case: #2675

### Steps to Reproduce

1. *Load save game attached to #2675 ticket*
2. Observe that level can't be cleared despite fulfilling all visible criteria.

### Expected Behaviour

Expected Behaviour should resolve that issue.

As a variants:
1. Instead of showing fulfilled criteria, Progress report can show unfulfilled first. With next priorities: I. Critical criteria (in red zone), II. Unfulfilled criteria, III. Fulfilled criteria.
2. Scrolling like left and right arrows around the table. (@tobylane proposed)

Also take into account that there are currently 5 criteria for victory, but Since CorsixTH 0.69 and up are going to have 2 more win conditions.

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

1

### CorsixTH Version

0.68.0

### Operating System

1

### Theme Hospital Version

CD

### Gamelog.txt

_No response_

### Additional Information

_No response_

Matroftt:
* Fixes #2750 

- Progress report shows every criteria that is used in level
- Example with six criterias:
{9EEB944B-0D66-452E-AF4A-E77F1FB4C9AF}

This is not a perfect solution due to overlapping, it seems that later balance and hospital value icons graphics should be replaced with icons that take less space.

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
@@ -61,7 +61,7 @@ function UIProgressReport:UIProgressReport(ui)
   local x = 263
   local crit_amount = #crit_data
   -- 30 gap width value is used when there is 5 criterias or less.
-  local gap_width = math.floor(math.min(150/crit_amount, 30))
+  local gap_width = math.floor(math.min(150 / crit_amount, 30))
   for _, crit_table in ipairs(crit_data) do
     crit_table.visible = true
     local crit_name = crit_table.name
--- a/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
@@ -57,8 +57,11 @@ function UIProgressReport:UIProgressReport(ui)
 
   -- Collect which criteria to show here, draw columns in draw
   -- Add the icons for the criteria
-  local x = 263
   local crit_data = world.endconditions:generateReportTable(hospital)
+  local x = 263
+  local crit_amount = #crit_data
+  -- 30 gap width value is used when there is 5 criterias or less.
+  local gap_width = math.floor(math.min(150/crit_amount, 30))
   for _, crit_table in ipairs(crit_data) do
     crit_table.visible = true
     local crit_name = crit_table.name
@@ -88,9 +91,10 @@ function UIProgressReport:UIProgressReport(ui)
     if not crit_table.icon_file then -- Icons from QData/Rep02V
       self:addPanel(crit_table.icon, x, 240)
     end
-    self:makeTooltip(tooltip, x, 180, x + 30, 180 + 90)
-    x = x + 30
+    self:makeTooltip(tooltip, x, 180, x + gap_width, 180 + 90)
+    x = x + gap_width
   end
+  self.gap_width = gap_width
   self.crit_data = crit_data
 
   self:addPanel(0, 606, 447):makeButton(0, 0, 26, 26, 8, self.close):setTooltip(_S.tooltip.status.close)
@@ -214,7 +218,7 @@ function UIProgressReport:draw(canvas, x, y)
         local icon_sprites = self.panel_sprites_table[crit_table.icon_file]
         icon_sprites:draw(canvas, crit_table.icon, x + lx, y + 240)
       end
-      lx = lx + 30
+      lx = lx + self.gap_width
     end
   end
 
--- a/CorsixTH/Lua/endconditions.lua
+++ b/CorsixTH/Lua/endconditions.lua
@@ -151,16 +151,9 @@ function EndConditions:generateReportTable(hospital)
     table.insert(report_table, crit_table)
     count = count + 1
   end
-  -- Limit the table to five goals, ordered by progress towards meeting the lose goal
-  if count > 5 then
-    table.sort(report_table, function(a,b) return a.progress > b.progress end)
-    for n = 6, #report_table do report_table[n] = nil end
-  end
-
   -- Fill up the report table with win criteria not already present as lose criteria
   for _, crit in pairs(report_table) do unique_criteria_set[crit.name] = true end
   for i = 1, #local_criteria_variable do
-    if count == 5 then break end
     local name = local_criteria_variable[i].name
     if win_group[name] and not unique_criteria_set[name] then
       count = count + 1


lewri (commented):
I would reiterate my feeling towards putting this behind a flag, but I would not stop this if it didn't happen.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -57,8 +57,11 @@ function UIProgressReport:UIProgressReport(ui)
 
   -- Collect which criteria to show here, draw columns in draw
   -- Add the icons for the criteria
-  local x = 263
   local crit_data = world.endconditions:generateReportTable(hospital)
+  local x = 263
+  local crit_amount = #crit_data
+  -- 30 gap width value is used when there is 5 criterias or less.
+  local gap_width = math.floor(math.min(150/crit_amount, 30))
```
lewri:
```suggestion
  local gap_width = math.floor(math.min(150 / crit_amount, 30))
```

CorsixTH/Lua/endconditions.lua
```diff
@@ -151,16 +151,9 @@ function EndConditions:generateReportTable(hospital)
     table.insert(report_table, crit_table)
     count = count + 1
   end
-  -- Limit the table to five goals, ordered by progress towards meeting the lose goal
-  if count > 5 then
-    table.sort(report_table, function(a,b) return a.progress > b.progress end)
-    for n = 6, #report_table do report_table[n] = nil end
-  end
```
lewri, line 158: No max buffer then?

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
```diff
@@ -57,8 +57,11 @@ function UIProgressReport:UIProgressReport(ui)
 
   -- Collect which criteria to show here, draw columns in draw
   -- Add the icons for the criteria
-  local x = 263
   local crit_data = world.endconditions:generateReportTable(hospital)
+  local x = 263
+  local crit_amount = #crit_data
+  -- 30 gap width value is used when there is 5 criterias or less.
```
lewri, line 63: Where we're > 5, I'd recommend at least considering adjusting the dollar sign icon. I think the floating dollar icon is the same size as this one isn't it?

You could make a square version of the icon.

ARGAMX:
> I would reiterate my feeling towards putting this behind a flag, but I would not stop this if it didn't happen.

What are the benefits of a flag for this feature?

lewri:
> > I would reiterate my feeling towards putting this behind a flag, but I would not stop this if it didn't happen.
> 
> What are the benefits of a flag for this feature?

Arguably this is a feature that enables something more than a user should have access to. Also given there is UI degradation as the criteria gets bigger it could be less accessible to some people, more accessible to others. 

The max buffer query did briefly come up in discord. If you don't have a flag of boolean, you could have one to set the maximum criteria to show instead (defaulted to 5) as an example. This also means we still retain some of Toby's code to criteria management.

MandrakeHorse:
It wouldn't be too difficult to add a configuration option called "Detailed Report", especially since the overlapping icons could result in players reporting it as a bug.

ARGAMX:
> > What are the benefits of a flag for this feature?
> 
> Arguably this is a feature that enables something more than a user should have access to. Also given there is UI degradation as the criteria gets bigger it could be less accessible to some people, more accessible to others.
> 
> The max buffer query did briefly come up in discord. If you don't have a flag of boolean, you could have one to set the maximum criteria to show instead (defaulted to 5) as an example. This also means we still retain some of Toby's code to criteria management.

I agree that it would be great to cover all modifications with flags.

But I'm concerned that if we set a flag here and leave it false by default, then this PR doesn't actually close issue 2750.

And since it doesn't close the issue 2750, it's as if this PR is pointless for regular players. Because apparently the idea of this change was to solve the issue 2750.

if we set a flag here and leave it true, then essentially the opposite will happen: your expectations won't be met.

From all this, I get the feeling that the flag here is essentially meaningless.

The flag is well-suited for enhancements.

If PR is a bug fix — and this PR seems a bug fix, not an enchantment — then it's unclear why anyone would want to continue playing with a bug.

lewri:
> If PR is a bug fix — and this PR seems a bug fix, not an enchantment — then it's unclear why anyone would want to continue playing with a bug.

Again, you could argue that the 5 is by design of the game itself, this isn't a bug and enhancement to the original (in tandem to the extra criteria that got added). 

@Matroftt (or @tobylane ) do you still suppress win/lose criteria in this pr that hasn't reached any activation threshold yet? (I.e. when the win criteria overrides the lose (and vice versa) or the lose criteria has not hit the start trigger)

Until the ui elements do manage to work nicely I'm hesitant to fully back something without a flag. I'm pretty sure we should also have tools in the code itself to shrink a sprite down to get around https://github.com/CorsixTH/CorsixTH/issues/2750#issuecomment-2842716518

tobylane:
> do you still suppress win/lose criteria in this pr that hasn't reached any activation threshold yet? (I.e. when the win criteria overrides the lose (and vice versa) or the lose criteria has not hit the start trigger)

Yes, that's in EndConditions:_checkWin/Losegroup functions.

I think we should strongly recommend level writers limit themselves to five. I did not follow this in [PrestigeBronzeEleven](https://github.com/CorsixTH/CommunityContent/blob/main/Campaigns/PrestigeBronze/PrestigeBronzeEleven.level), the only published example where the five limit in UI hinders the player. With that in mind, yes to this PR and yes to later narrowing those icons.

CorsixTH/Lua/dialogs/fullscreen/progress_report.lua
Matroftt: Should this be done in this PR or in the future? I was going to make icons for patient and staff happiness too

lewri: You could bundle them together elsewhere.

TheCycoONE:
Some discussion in the related issue: #2750

Matroftt:
Closed since it was decided to approach this issue in different way

PR #3087, opened 2025-10-22
Right click in build mode remove window

### LINKED ISSUE #2682: [Bug] Can't remove doors or windows in build mode
### Describe the issue

In the original TH, you could remove windows and doors when in build mode by right clicking on them. On corsix, you must "cancel" the building and start over.

### Steps to Reproduce

1.  make a room, place a door and a window
2. while still in build mode try to erase them with right click

### Expected Behaviour

Right click should remove windows or doors while in build mode.

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

### CorsixTH Version

0.68

### Operating System

Windows 10 x64

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml
No gamelog generated for me on 0.68
```

### Additional Information

No gamelog generated for me on 0.68

marcassin92:
*Fixes #2682*

**Changes :**
- In "Window" phase, the right click remove the window from the wall and reset to the default transparent wall.
- Move right and left click logic in dedicated method.

Note : 
As the cursor is above a wall in "window" phase, after right click, the window is still there because the `blueprint_window` is still valid and we're in build mode. Move the cursor outside this cell to see the window disappear. 
Should we hide the blueprint/window in this case ?

Wall selection to delete is the same as wall construction (Test: Try delete a window with the cursor in the room corner)

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -1122,7 +1122,9 @@ function UIEditRoom:onLeftButtonDown(x, y)
   end
 end
 
--- Remove window
+--! Attempt to remove a window placed on the room blueprints
+--!param x co-ordinate
+--!param y co-ordinate
 function UIEditRoom:tryRemoveWindowFromWall(x, y)
   local cell_x, cell_y, wall_dir = self:screenToWall(self.x + x, self.y + y)
   if not cell_x then


CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1072,32 +1072,93 @@ function UIEditRoom:draw(canvas, ...)
...
+      elseif wall_dir == "east" then
+        wall_x = cellx + 1
+      end
+
+      local anim = self.blueprint_wall_anims[wall_x][wall_y]
+      if anim then
+        -- reset tile to basic floor/wall
+        map:setCell(cellx, celly, 4, 24)
+        anim:setAnimation(self.anims, 120, direction_flag)
+        anim:setTag(nil)
```
Matroftt:
```suggestion
        anim:setTag(nil)
        self.ui:playSound("de_build.wav")
```
Dont forget to play sfx

Matroftt (changes requested):

lewri:
> Note :
> As the cursor is above a wall in "window" phase, after right click, the window is still there because the `blueprint_window` is still valid and we're in build mode. Move the cursor outside this cell to see the window disappear.
> Should we hide the blueprint/window in this case ?

I imagine there will be some way to update (or at least give the illusion of) the blueprint wall on click. Though I haven't touched blueprint code really.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1072,32 +1072,94 @@ function UIEditRoom:draw(canvas, ...)
...
+    end
+  elseif self.phase == "windows" then
+    self:placeWindowBlueprint()
+  end
+end
+
+-- Remove window
+function UIEditRoom:onRightButtonDown(button, x, y)
+  if self.phase == "windows" then
+    local cellx, celly, wall_dir = self:screenToWall(self.x + x, self.y + y)
```
lewri: `cell_x` and `cell_y`

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1072,32 +1072,94 @@ function UIEditRoom:draw(canvas, ...)
...
+    else
+      self.ui:tutorialStep(3, 9, 10)
+    end
+  elseif self.phase == "windows" then
+    self:placeWindowBlueprint()
+  end
+end
+
+-- Remove window
+function UIEditRoom:onRightButtonDown(button, x, y)
```
lewri: Why not call the function what it does? `function UIEditRoom:removeWindow(x, y)` (the button parameter is unused).

lewri: In tandem to this, if you moved the phase check inside the onMouseDown function you have a nicely defined function above.

e.g.

```lua
elseif button == "right" then
  if self.phase == "windows" then
    self:removeWindow()
  end
end
```

Ideally a good refactor is also thinking if the left button would benefit from the same treatment. But it is out of scope of this PR it was just simply a code move.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1072,32 +1072,94 @@ function UIEditRoom:draw(canvas, ...)
...
+
+local window_floor_blueprint_markers = {
+  north = 33,
+  east = 34,
+  south = 35,
+  west = 36,
+}
+
+
+function UIEditRoom:onLeftButtonDown(button, x, y)
```
lewri: Button parameter is unused

lewri (commented):
Good stuff

CorsixTH/Lua/dialogs/edit_room.lua
marcassin92: > Why not call the function what it does

In my first implementation removing the door was possible too so that made sense. Now as it's only for windows, it make sense to put a more describing name.
I choosed something more verbose, as `removeWindow` is already defined in a grand parent `Window` class.

ARGAMX:
> As the cursor is above a wall in "window" phase, after right click, the window is still there because the `blueprint_window` is still valid and we're in build mode. Move the cursor outside this cell to see the window disappear.
> Should we hide the blueprint/window in this case ?

In the original Theme Hospital the window is visually still there after right click.
Based on this, I would recommend to stick the same behavior. 👍
So I think we should not hide the blueprint/window in this case.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -1072,31 +1072,92 @@ function UIEditRoom:draw(canvas, ...)
...
+      self:confirm(true)
+    else
+      self.ui:tutorialStep(3, 9, 10)
+    end
+  elseif self.phase == "windows" then
+    self:placeWindowBlueprint()
+  end
+end
+
+-- Remove window
```
lewri: Sorry, I should've helped with doc on this one.

```suggestion
--! Attempt to remove a window placed on the room blueprints
--!param x co-ordinate of target tile
--!param y co-ordinate of target tile
```

TheCycoONE:
Thanks @marcassin92

PR #3089, opened 2025-10-23
Find nearby objects from a square area around the humanoid using the entity map.

### LINKED ISSUE #3070: World:findObjectNear / pathfinder.findObject -- callback can trigger twice

Alberth289346:
*Fixes #3070 

- Instead of searching for objects, it iterates over tiles around the humanoid to find objects, using the entity map.
- It only examines tiles of the same room.

### DIFF
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -527,7 +527,7 @@ end
 
 --! Collect the objects in a square around the humanoid within the same room.
 --!param size (int) Size of the square to be searched, as number of rows/columns
---  around the humanoid. Size 0 only searched the tile of the humanoid, size 1
+--  around the humanoid. Size 0 only searches the tile of the humanoid, size 1
 --  adds the 8 tiles around the humanoid (one tile in all directions), and so on.
 --!param object_spec (str or array str) Name(s) of objects to find.
 --!return (array objs, or table str -> array objects) Found requested objects.


CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -82,10 +80,10 @@ function Staff:tickDay()
     ["skeleton"]     = 0.002,
     ["tv"]           = 0.0005,
   }
+  local happy_objects = {"extinguisher", "bin", "bookcase", "skeleton", "tv"}
```
lewri: Seems sad we're repeating the good_objects list with a happy_objects list?

CorsixTH/Lua/utility.lua
```diff
@@ -75,6 +75,15 @@ function table_length(table)
   return count
 end
 
+--! Get a random item from an array.
+--!param array Array with 0 or more items.
+--!return (nil or an item)
+function getRandomEntryFromArray(array)
+  if #array == 0 then return nil end
+  if #array == 1 then return array[1] end
+  return array[math.random(1, #array)]
```
lewri, line 84: Wouldn't math.random(1,1) still just return 1?

```lua
for i=1,10 do
  print(math.random(1,1))
end
```

Output:
```
1
1
1
1
1
1
1
1
1
1
```

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -525,6 +525,50 @@ function Humanoid:getCurrentMood()
   end
 end
 
+--! Collect the objects in a rectangle around the humanoid within the same room.
+--!param radius (int) Number of tiles distance from the humanoid to search.
+--!param object_spec (str or array str) Name(s) of objects to find.
+--!return (array objs, or table str -> array objects) Found requested objects.
+--  One array is returned if an object name was supplied, else a table with
+--  object names to array of objects.
+function Humanoid:findObjectsInSquare(radius, object_spec)
```
lewri: `findObjectsWithinRadius`?

lewri (commented):
Not 100% this belongs in humanoid and not world but I could be otherwise convinced :)

Nice work

CorsixTH/Lua/entities/humanoid.lua
TheCycoONE: Within Radius implies a circle; Manhattan Distance implies a diamond, this is a square. Maybe the radius parameter name is wrong but I don't know the right term for this.

lewri: Distance/range for the parameter?

TheCycoONE: It's not a distance under any normal definition. Inradius seems to be the accepted math term for the radius of the circle that fits inside a polygon, which this is. It's not a common term.

TheCycoONE: Which is to say I'm fine with the current naming.

CorsixTH/Lua/utility.lua
Alberth289346: It does, but why ask `math.random` to compute a known answer?

CorsixTH/Lua/entities/humanoids/staff.lua
Alberth289346: We are. I agree it's non-optimal.

I pondered about requiring the form used by the function, ie `{extinguisher={}, bin={}, bookcase={}, skeleton={}, tv={}}`, but that means more work here, and it's somewhat ridiculous for single object types to return `{litter={ <<objects>> }}` from the collection function.

Shall insert a loop to build the array.

CorsixTH/Lua/utility.lua
lewri: It was only as I thought the second if was a wasted execution for arrays > 1, when arrays == 1 can also use the random call. In reality, the cost is very minute so we can leave it as is :)

Alberth289346:
> Not 100% this belongs in humanoid and not world but I could be otherwise convinced :)

Why would it fit in `world`?

I checked all `findObjectsNear` occurrences for the kind of search that was done here, and more cases don't exist as far as I can see.

How humanoids are affected by their surroundings seems a property of humanoids to me. (ie it's 'their' problem.)
In my view that's why happiness changes are in `staff` and `patient` and not elsewhere.
That code needs information to work with, and that computation can be shared between both kinds of humanoids (rather than be implemented in both `staff` and `patient`).

Arguably, the proposed solution is not optimal either. Code between `staff` and `patient` looks very similar. There seems room for more sharing. however, that was not the goal here.

CorsixTH/Lua/entities/humanoid.lua
Alberth289346: I'll switch to `size`. That includes the tile of the humanoid itself, so it needs a little more explanation, and some additional computations to handle even numbers correctly.

Alberth289346:
Done

New changes:
``` diff
diff --git a/CorsixTH/Lua/entities/humanoid.lua b/CorsixTH/Lua/entities/humanoid.lua
index 7e4399f3..33f20d97 100644
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -525,13 +525,15 @@ function Humanoid:getCurrentMood()
   end
 end
 
---! Collect the objects in a rectangle around the humanoid within the same room.
---!param radius (int) Number of tiles distance from the humanoid to search.
+--! Collect the objects in a square around the humanoid within the same room.
+--!param size (int) Size of the square including the tile of the humanoid.
+--  Size 1 only covers the tile of the humanoid, size 3 adds the 8 tiles around
+--  it, size 5 gives two tiles in all directions, and so on.
 --!param object_spec (str or array str) Name(s) of objects to find.
 --!return (array objs, or table str -> array objects) Found requested objects.
 --  One array is returned if an object name was supplied, else a table with
 --  object names to array of objects.
-function Humanoid:findObjectsInSquare(radius, object_spec)
+function Humanoid:findObjectsInSquare(size, object_spec)
   -- Prepare for the search.
   local objs_table = {}
   if type(object_spec) == "string" then
@@ -548,9 +550,12 @@ function Humanoid:findObjectsInSquare(radius, object_spec)
   -- Search the rectangle for as far as it is within the world.
   local width, height = world_map.th:size()
   local entity_map = self.world.entity_map
-  for x = self.tile_x - radius, self.tile_x + radius do
+
+  size = (size >= 1) and size or 1 -- Force to include the center tile.
+  local half_size = math.floor((size - 1) / 2)
+  for x = self.tile_x - half_size, self.tile_x + size - half_size do
     if x >= 1 and x <= width then
-      for y = self.tile_y - radius, self.tile_y + radius do
+      for y = self.tile_y - half_size, self.tile_y + size - half_size do
         if y >= 1 and y <= height and world_map:getRoomId(x, y) == self_room_id then
           for _, obj in ipairs(entity_map:getObjectsAtCoordinate(x, y)) do
             local entry = objs_table[obj.id]
diff --git a/CorsixTH/Lua/entities/humanoids/patient.lua b/CorsixTH/Lua/entities/humanoids/patient.lua
index 65d2cc48..afd93da2 100644
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -689,7 +689,7 @@ function Patient:tickDay()
     local nausea = (1.0 - self:getAttribute("health")) * initialVomitMult
     local numVomit = 0
 
-    for _, litter in ipairs(self:findObjectsInSquare(2, "litter")) do
+    for _, litter in ipairs(self:findObjectsInSquare(5, "litter")) do
       -- Count vomit inducing litter.
       if litter:vomitInducing() then numVomit = numVomit + 1 end
 
@@ -713,13 +713,13 @@ function Patient:tickDay()
   end
 
   -- It is nice to see plants, but dead plants make you unhappy.
-  local plant = getRandomEntryFromArray(self:findObjectsInSquare(2, "plant"))
+  local plant = getRandomEntryFromArray(self:findObjectsInSquare(5, "plant"))
   if plant then
     self:changeAttribute("happiness", -0.0003 + (plant:isPleasingFactor() * 0.0001))
   end
 
   -- It always makes you happy to see you are in safe place.
-  local extinguisher = getRandomEntryFromArray(self:findObjectsInSquare(2, "extinguisher"))
+  local extinguisher = getRandomEntryFromArray(self:findObjectsInSquare(5, "extinguisher"))
   if extinguisher then self:changeAttribute("happiness", 0.0002) end
 
   -- sitting makes you happy whilst standing and walking does not
diff --git a/CorsixTH/Lua/entities/humanoids/staff.lua b/CorsixTH/Lua/entities/humanoids/staff.lua
index 90135342..cc49acaa 100644
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -67,7 +67,7 @@ function Staff:tickDay()
   end
 
   -- It is nice to see plants, but dead plants make you unhappy
-  local plant = getRandomEntryFromArray(self:findObjectsInSquare(2, "plant"))
+  local plant = getRandomEntryFromArray(self:findObjectsInSquare(5, "plant"))
   if plant then
     self:changeAttribute("happiness", -0.003 + (plant:isPleasingFactor() * 0.001))
   end
@@ -80,8 +80,13 @@ function Staff:tickDay()
     ["skeleton"]     = 0.002,
     ["tv"]           = 0.0005,
   }
-  local happy_objects = {"extinguisher", "bin", "bookcase", "skeleton", "tv"}
-  happy_objects = self:findObjectsInSquare(2, happy_objects)
+
+  -- Construct an array with the object names.
+  local happy_objects = {}
+  for name, _ in pairs(good_objects) do happy_objects[#happy_objects + 1] = name end
+
+  -- Look what's around the humanoid, and adapt the happiness.
+  happy_objects = self:findObjectsInSquare(5, happy_objects)
   for obj_name, happiness_score in pairs(good_objects) do
     self:changeAttribute("happiness", #happy_objects[obj_name] * happiness_score)
   end
@@ -139,7 +144,7 @@ function Staff:tick()
   end
 
   -- seeing litter will make you unhappy. If it is pee or puke it is worse
-  for _, litter in ipairs(self:findObjectsInSquare(2, "litter")) do
+  for _, litter in ipairs(self:findObjectsInSquare(5, "litter")) do
     if litter:anyLitter() then
       self:changeAttribute("happiness", -0.0002)
     else

```

Alberth289346:
Back to the `radius` definition, except it's now named `size`.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -525,6 +525,54 @@ function Humanoid:getCurrentMood()
   end
 end
 
+--! Collect the objects in a square around the humanoid within the same room.
+--!param size (int) Size of the square to be searched, as number of rows/columns
+--  around the humanoid. Size 0 only searched the tile of the humanoid, size 1
```
Alberth289346:
```suggestion
--  around the humanoid. Size 0 only searches the tile of the humanoid, size 1
```

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -82,10 +80,15 @@ function Staff:tickDay()
...
+  local happy_objects = {}
+  for name, _ in pairs(good_objects) do happy_objects[#happy_objects + 1] = name end
+
+  -- Look what's around the humanoid, and adapt the happiness.
+  happy_objects = self:findObjectsInSquare(2, happy_objects)
   for obj_name, happiness_score in pairs(good_objects) do
-    self.world:findObjectNear(self, obj_name, 2, function()
-      self:changeAttribute("happiness", happiness_score)
-    end)
+    self:changeAttribute("happiness", #happy_objects[obj_name] * happiness_score)
```
lewri, line 91: We only ever counted the first object found; as the bug was that we were adjusting more than once. Otherwise we become very unbalanced without changing other factors.

lewri: Happy to consider it as a wider scope enhancement though to review all factors in future.

TheCycoONE: That's not true, the callback happens for every object found in the before and then some due to the revisit issue. Whether intentional or not is unknown.

lewri: Yep sorry

PR #3090, opened 2025-10-23
Fix Task reassignment error with dead handyman

### LINKED ISSUE #2918: [Bug] Handyman path finding error "bad argument #3 to 'findDistance' (number expected, got nil)"
### Describe the issue

[HandymanToSlicer.sav.zip](https://github.com/user-attachments/files/20779931/HandymanToSlicer.sav.zip)

The handyman in the Slicer can't repair the machine. This is the first handyman visit to the machine. If the machine is rotated, the handyman can repair it. Is the repair task at an inaccessible tile?

### Steps to Reproduce

2. See the handyman from the Inflator go to the Slicer

3. See the error that (usually) happens around the 17-21st of the month

Alternatively, hire a new handyman and drop him in the Slicer.

### Expected Behaviour

Repair and leave

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

### CorsixTH Version

7ce9f65f (recent master)

### Operating System

macOS 14.2

### Theme Hospital Version

CD

### Gamelog.txt

```yaml
Error in timer handler: 
CorsixTH/CorsixTH/Lua/world.lua:1389: bad argument #3 to 'findDistance' (number expected, got nil)
stack traceback:
	[C]: in method 'findDistance'
	CorsixTH/Lua/world.lua:1389: in method 'getPathDistance'
	CorsixTH/CorsixTH/Lua/hospital.lua:2223: in method 'searchForHandymanTask'
	...sixTH/CorsixTH/Lua/entities/humanoids/staff/handyman.lua:128: in method 'searchForHandymanTask'
	CorsixTH/CorsixTH/Lua/calls_dispatcher.lua:334: in method 'answerCall'
	CorsixTH/CorsixTH/Lua/humanoid_actions/answer_call.lua:33: in field '?'
	CorsixTH/CorsixTH/Lua/entities/humanoid.lua:541: in method 'startAction'
	CorsixTH/CorsixTH/Lua/entities/humanoid.lua:647: in method 'finishAction'
	CorsixTH/CorsixTH/Lua/humanoid_actions/use_object.lua:410: in local 'timer_function'
	CorsixTH/CorsixTH/Lua/entity.lua:223: in field 'tick'
	CorsixTH/CorsixTH/Lua/entities/humanoids/staff.lua:112: in method 'tick'
	CorsixTH/CorsixTH/Lua/world.lua:911: in method 'onTick'
	CorsixTH/CorsixTH/Lua/app.lua:1289: in function <CorsixTH/CorsixTH/Lua/app.lua:1286>
	(...tail calls...)
	CorsixTH/CorsixTH/Lua/app.lua:1189: in function <CorsixTH/CorsixTH/Lua/app.lua:1184>
```

### Additional Information

_No response_

marcassin92:
### Description
*Fixes #2918*

The hospital is notified when a handyman die in order to clean the current Tasks,
Fixes path finding error when trying to reassign task to dead handyman referenced in tables `Hospital:handymanTasks`.
New method to search for tasks assigned to a Handyman.
 
### Note:
After discussion and more investigation :
This PR is a "hotfix" and does not fix the root issue of multiple tasks assigned to one Handyman.
This fix avoid crash for game saves with Handymans having multiple tasks assigned.
This PR can be closed if #3136 is accepted.

### Test :

This fix cannot be tested on issue's game save ( Handyman already dead ).

To reproduce / test the PR :
- Download the save [ScannerExplodeHandymanDie.zip](https://github.com/user-attachments/files/23109835/ScannerExplodeHandymanDie.zip)
- Keep the only handyman into the scanner room
- Wait for the scanner to explode
- Recruit another handyman ( can be done before explosion )
- Wait few days for the error to happen ( or not if patched )

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff/handyman.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff/handyman.lua
@@ -37,7 +37,7 @@ end
 
 function Handyman:die()
   Staff.die(self)
-  self.hospital:onHandymanDeath(self)
+  self.hospital:unassignHandymanTasks(self)
 end
 
 function Handyman:dump()
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1616,7 +1616,7 @@ end
 
 --! Handyman has died. Remove all tasks assigned to this staff
 --!param handyman The handyman deceased
-function Hospital:onHandymanDeath(handyman)
+function Hospital:unassignHandymanTasks(handyman)
   local tasks = self:findAssignedTasksToHandyman(handyman)
   for _, task in ipairs(tasks) do
     task.assignedHandyman = nil


CorsixTH/Lua/entities/humanoids/staff/handyman.lua
```diff
@@ -38,6 +38,7 @@ end
 function Handyman:die()
   self:unassignTask()
   Staff.die(self)
+  self.hospital:handymanDeath(self)
```
lewri: onHandymanDeath might be nicer. But I note in hospital the humanoid one doesn't follow that trend either... Maybe it can stay.

CorsixTH/Lua/hospital.lua
```diff
@@ -1614,6 +1614,18 @@ function Hospital:humanoidDeath(patient)
   end
 end
 
+--! Handyman has died. Remove all tasks assigned to this staff
+--!param handyman The handyman deceased
+function Hospital:handymanDeath(handyman)
+  for i = 1, #self.handymanTasks do
```
lewri: There is a function buried in the code called removeHandymanTask. Is there a chance it looks similar to this?

CorsixTH/Lua/entities/humanoids/staff/handyman.lua
```diff
@@ -38,6 +38,7 @@ end
 function Handyman:die()
   self:unassignTask()
```
ARGAMX: Interestingly, here is where the function is called to unassign the task from the handyman.
But this isn't happening, right?

So, two lines down, we now again call a function to remove the task from the handyman.
Just reading the code and not knowing about the unassign issue this looks like double work.
It turns out this line needs to be deleted? Or not?

It's somehow unclear here..

CorsixTH/Lua/hospital.lua
```diff
@@ -1614,6 +1614,18 @@ function Hospital:humanoidDeath(patient)
...
+    for j = 1, #self.handymanTasks[i].subTable do
+      if self.handymanTasks[i].subTable[j].assignedHandyman == handyman then
+        self.handymanTasks[i].subTable[j].assignedHandyman = nil
+      end
+    end
+  end
+end
+
```
ARGAMX, line 1625: A good feature that sorts through all the tasks of handyman's and searches for tasks assigned to a specific handyman.

But possible the proper name for all of this should be `findTaskAssignedToHandyman`?

I think we could reuse this search function. It's strange that it hasn't been coded before.

Thus, `Hospital:handymanDeath` can call `Hospital:findTaskAssignedToHandyman`, get the desired task, and clear the `assignedHandyman` field of that task.

Sounds good?

Just come to idea that due to errors in the code there could be multiple such tasks assigned to that handyman?

So it turns out we need to return not just one task, but an array of tasks.

Possible this might be the answer to why the first task cleanup call fails. First call cleans only one task, trusting that everything is working correctly and there are no other tasks assigned to that handyman.

CorsixTH/Lua/hospital.lua
lewri: This is a comment in the pipeline I had before there were other reviews

CorsixTH/Lua/hospital.lua
marcassin92: > But possible the proper name for all of this should be findTaskAssignedToHandyman?

:+1: 

> Just come to idea that due to errors in the code there could be multiple such tasks assigned to that handyman?

That's exactly the case, while debugging/printing tasks in the ScannerExplodeHandymanDie.zip save file, we can observe that the handyman have 4 assigned tasks in the `Hospital:handymanTasks` table, then 3 after the `Handiman:unassignTask()` call.

```lua
function Handyman:die()
  local tasks_before = self.hospital:findTasksAssignedToHandyman(self)
  for i, t in ipairs(tasks_before) do print(i, " => ", t) end
  self:unassignTask()
  Staff.die(self)
  local tasks_after = self.hospital:findTasksAssignedToHandyman(self)
  for i, t in ipairs(tasks_after) do print(i, " => ", t) end
end
```

If multiple assigned tasks to a handyman is an error, then fixing the multiple assignments would be enough to fix this crash.
If we want to fix this later, we can let the `Hospital:onHandymanDeath(handyman)` remove all assigned task and then remove the `unassignTask()` in the `die()` method as it's doing the same thing for only one task.

I guess the fix of the multiple assignments is way more cleaner.

( At really first sight i thought multiple assignments are a normal thing (tasks queue system or idk), but i guess you're right, a handyman should have only 1 task assigned. )

CorsixTH/Lua/hospital.lua
marcassin92: `removeHandymanTask` seems to remove assignment for a particular Task ( Called when a machine is repaired, floor cleaned up, plant watered ... ) without checking the handyman assigned. It may differs from my first implementation where i wanted to unassigned all (the) task(s) of a particular Handyman (May not be the best solution cf [comment](https://github.com/CorsixTH/CorsixTH/pull/3090#discussion_r2483701554) above).

CorsixTH/Lua/entities/humanoids/staff/handyman.lua
Alberth289346: Whether or not the handyman is dead doesn't matter to the function. If you call it with a living handyman, it will also remove its tasks.

Please try to make the function description as general as possible by not mentioning conditions that are not relevant to the function.

PR #3091, opened 2025-10-24
Build ffmpeg with openssl on macos

tobylane:
- Vcpkg's ffmpeg uses securetransport on macos and ios, which Apple deprecated and curl dropped earlier this year. As Apple hasn't made a replacement, this adds openssl to ffmpeg. Outside of vcpkg this isn't needed, just build without securetransport or openssl. I include ios because it has the same fault, but there's another issue before ios can be a build platform.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -61,7 +61,11 @@
             "avcodec",
             "avformat",
             "swresample",
-            "swscale"
+            "swscale",
+            {
+              "name": "openssl",
+              "platform": "osx, ios"
+            }
           ]
         }
       ]


TheCycoONE:
Well we don't need transport, I didn't think our ffmpeg port included it but if it does you can fix it there?

TheCycoONE (approved):
Approved but skeptical.

tobylane:
Yes, closed in favour of https://github.com/microsoft/vcpkg/issues/47955 or I'll try to write a fix for our port.

TheCycoONE:
@tobylane we have a fork of the ffmpeg port in CorsixTH/vcpkg-registry that you can patch, or I could apply your suggested patch to try.

Issue #1883, opened 2021-05-18
Emscripten support for CorsixTH


alexandre-lavoie:
**What feature or enhancement would you like to add?**
Provide a web port for CorsixTH using Emscripten.

**How would this feature benefit the gameplay of CorsixTH?**
This would make CorsixTH extremely portable - with high reliability. Additionally, the effort of porting with Emscripten is relatively low compared to a total rewrite.

**Was the feature in the original game?**
No

**What version of CorsixTH are you using?**
0.64

**Does this feature/enhancement relate to a current problem?**
I've been having a go at this issue here https://github.com/alexandre-lavoie/CorsixTH/tree/build-ext . This issue aims mostly for discussion/progress.

**Additional information**
As of first post, I was able to at least fully link the components of the very minimal build. I'm using Nixpkgs also to be forced to have the build steps solid and reproducible. The current was able to render literally one frame (as seen below). I think it gives an idea of what could be possible, but threads seem to have put me mostly in a stalemate. Emscripten seems to not support PThreads and dynamic linking through `dlopen` (which I think is the only way to link Lua packages) at the same time. I am also very limited in the resources regarding Emscripten - so I ended my development with these possible directions in mind:

- If we could find a way to make Lua "inner-build" packages in a way to not have to `dlopen` them, we could get rid of that feature and use pthread.

- We could offer a `single threaded` option that rewrite all threaded code. I had a try in a latest commit and I am not very familiar with C to/from Lua. It could be useful for anyone that also wants to port on lighter hardware.

- We could use C++ alternatives for the Lua pkgs to have a clean Lua-package-free version. 

I have a few more ideas in "this scale", but I am worried that it would change CorsixTH too harshly. I am also worried that I would lose interest/lack time for a major "patch". I feel there must be a way to do this fully through builds, but hours of `nix-build web.nix` didn't find anything fully working yet. Therefore, I leave this discussion open!

![CorsixTH](https://user-images.githubusercontent.com/36684879/118584898-df3f9c80-b765-11eb-87ab-0d3c4250920f.png)

TheCycoONE:
Item 1 I know how to do. I was already considering adding a cmake option for it, so that's one more argument in favor.

Item 2 is a little harder. One might start by disabling movies and music.

Item 3 I don't see the need for once 1 was resolved.

alexandre-lavoie:
I'll make a PR to sub-branch `emscripten` so we can thinker on the configs. 

[...]

PR #3093, opened 2025-10-24
[WIP] Adds browser platform

ololoken:
Demo available *Unlinked until copyrighted files are removed*
th


lewri:
Ah, @alexandre-lavoie would be proud someone else picked up his original work on getting the game working with emscripten :)

ololoken:
hey @lewri , would you share link to the Alexandre's work?

tobylane:
Their issue is #1883, their PR is #1891, also see branch https://github.com/alexandre-lavoie/CorsixTH/tree/build-ext

I'm happy to see this, well done. For legality, you may want to switch to the demo files.

Some of these changes should be behind a check for emscripten, see [this commit](https://github.com/CorsixTH/CorsixTH/commit/60bb9accd599c39cb565711319918aee10c7bbcd) for how to add emscripten to app.os. What alternatives are there to including the lpeg and lfs sources?

ololoken:
@tobylane thanks, I already see some ideas I can re-use.

TheCycoONE:
Changes unrelated to emscripten should be submitted independently. Some of these Theme Hospital to CorsixTH naming changes make sense, others do not.

TheCycoONE:
Embedding libraries like lpeg and lfs will not be accepted, better to pull them in the build step like normal.

CorsixTH/Levels/demo.level
```diff
@@ -23,7 +23,7 @@ SOFTWARE.
 %MapFile = "LEVEL.L1"
 
 %LevelBriefing = "This is the only map available as long as you use
- the Theme Hospital Demo files."
+ the CorsixTH Demo files."
```
TheCycoONE: Theme Hospital was right here.

CorsixTH/Src/CMakeLists.txt
```diff
@@ -4,12 +4,30 @@ configure_file(${CMAKE_SOURCE_DIR}/CorsixTH/Src/config.h.in ${CMAKE_BINARY_DIR}/
...
 if (CMAKE_SYSTEM_NAME MATCHES "Emscripten")
+    include(FetchContent)
+    FetchContent_Declare(lpeg
+            URL  https://github.com/roberto-ieru/LPeg/archive/refs/heads/master.zip
+    )
+    FetchContent_Declare(lfs
+            URL  https://github.com/lunarmodules/luafilesystem/archive/refs/tags/v1_8_0.zip
+    )
+    FetchContent_Populate(lpeg)
+    FetchContent_Populate(lfs)
```
tobylane:
```suggestion
    FetchContent_Declare(lpeg
            URL  https://github.com/roberto-ieru/LPeg/archive/refs/heads/master.zip
            DOWNLOAD_EXTRACT_TIMESTAMP false
    )
    FetchContent_Declare(lfs
            URL  https://github.com/lunarmodules/luafilesystem/archive/refs/tags/v1_8_0.zip
            DOWNLOAD_EXTRACT_TIMESTAMP false
    )
    FetchContent_MakeAvailable(lpeg lfs)
```
This avoids CMake warnings and appears to do the same job.

tobylane (commented):
I have built corsixth.js, without html (because of modularize?). Is your html and other js generated or hand written?

tobylane:
https://github.com/tobylane/CorsixTH/commit/2d4681e2ad65b0d1155ee531fb0b16eef94b02a9

This wraps the Lua changes you've made in a check for emscripten. You can turn movies off in the same way as the update check is done in this.

ololoken:
@tobylane I think it's better to keep crossplatform code here https://github.com/CorsixTH/CorsixTH/pull/3093/files#diff-6c6af139aff1f6a63e202d5bd0b6a468d97e9e758e8cf1d7b9907782c6b26269R222

CorsixTH/Src/CMakeLists.txt
```diff
@@ -1,6 +1,37 @@
...
 
+
+set(LPEG_SRC_FILES "")
+set(LPEG_HRC_FILES "")
+set(LFS_SRC_FILES "")
+set(LFS_HRC_FILES "")
+if (CMAKE_SYSTEM_NAME MATCHES "Emscripten")
+    include(FetchContent)
+    FetchContent_Declare(lpeg
+            URL  https://github.com/roberto-ieru/LPeg/archive/refs/heads/master.zip
```
TheCycoONE: You almost certainly want a specific tag or commit.

TheCycoONE: It would also be better to fetch into their own target rather than in CorsixTH/Src which would require fewer other changes to cmake.

TheCycoONE: Nevermind, I see why this was done... ok.

TheCycoONE:
Can you make your demo page require the user to select the demo files from their computer. EA did a take down of Ultima 4 when they did something very similar, despite the game being free.

ololoken:
@TheCycoONE yep, on weekend.

CorsixTH/Lua/dialogs/resizables/main_menu.lua
```diff
@@ -50,9 +50,11 @@ function UIMainMenu:UIMainMenu(ui)
     {_S.main_menu.continue,        self.buttonContinueGame,   _S.tooltip.main_menu.continue},
     {_S.main_menu.load_game,       self.buttonLoadGame,       _S.tooltip.main_menu.load_game},
     {_S.main_menu.options,         self.buttonOptions,        _S.tooltip.main_menu.options},
-    {_S.main_menu.map_edit,        self.buttonMapEdit,        _S.tooltip.main_menu.map_edit},
-    {_S.main_menu.exit,            self.buttonExit,           _S.tooltip.main_menu.exit}
+    {_S.main_menu.map_edit,        self.buttonMapEdit,        _S.tooltip.main_menu.map_edit}
   }
+  if TheApp and TheApp.os != 'emscripten' then
```
tobylane:
```suggestion
  if TheApp and TheApp.os ~= 'emscripten' then
```

CorsixTH/Src/main.cpp
```diff
@@ -49,7 +49,7 @@ SOFTWARE.
 
 extern "C" {
 int luaopen_random(lua_State* L);
-#ifdef CORSIX_TH_LINK_LUA_MODULES
+#if CORSIX_TH_LINK_LUA_MODULES || __EMSCRIPTEN__
```
tobylane:
```suggestion
#if defined(CORSIX_TH_LINK_LUA_MODULES) || defined(__EMSCRIPTEN__)
```

CorsixTH/Src/main.cpp
```diff
@@ -159,7 +159,7 @@ int lua_main_no_eval(lua_State* L) {
   preload_lua_package(L, "persist", luaopen_persist);
   preload_lua_package(L, "sdl", luaopen_sdl);
 
-#ifdef CORSIX_TH_LINK_LUA_MODULES
+#if CORSIX_TH_LINK_LUA_MODULES || __EMSCRIPTEN__
```
tobylane:
```suggestion
#if defined(CORSIX_TH_LINK_LUA_MODULES) || defined(__EMSCRIPTEN__)
```

TheCycoONE:
From what I've read the emscripten toolchain for cmake supports static linking.

Not urgent, but I think it would be better to keep lpeg / lfs as targets like the current FindPackage, even if we use Fetch (like we do with catch2) - possibly some other things we keep a little more consistent in the cmake too.

Edit: Looks like there is an emscripten triplet for vcpkg `wasm32-emscripten` - no idea how well it works.

ololoken:
need advice, what I'm doing wrong? 
https://github.com/ololoken/ems-runner-corsix-th/actions/runs/18954535011/job/54127801071
err: "At least one entry of URL is a path" on FetchContent_MakeAvailable

lewri:
> need advice, what I'm doing wrong? https://github.com/ololoken/ems-runner-corsix-th/actions/runs/18954535011/job/54127801071 err: "At least one entry of URL is a path" on FetchContent_MakeAvailable

Looking it up briefly means a local directory path (e.g. `C:/Foo/Bar.zip`) got passed where it expected a URL to download from

I'm not good with cmake things but cyco or Toby probably can pick at it.

CorsixTH/Src/CMakeLists.txt
```diff
@@ -1,6 +1,40 @@
...
+            URL  "https://github.com/roberto-ieru/LPeg/archive/refs/heads/master.zip"
+            SOURCE_DIR "${CMAKE_BINARY_DIR}/ep_lpeg"
+            DOWNLOAD_EXTRACT_TIMESTAMP false
+    )
+    FetchContent_Declare(lfs
+            URL  "https://github.com/lunarmodules/luafilesystem/archive/refs/tags/v1_8_0.zip"
+            SOURCE_DIR "${CMAKE_BINARY_DIR}/ep_lfs"
+            DOWNLOAD_EXTRACT_TIMESTAMP false
+    )
+    FetchContent_MakeAvailable(lpeg)
```
lewri, line 23: For reference yesterday's build failure on the emscripten build points here

CorsixTH/Src/CMakeLists.txt
```diff
@@ -1,6 +1,40 @@
...
+
+set(LPEG_SRC_FILES "")
+set(LPEG_HRC_FILES "")
+set(LFS_SRC_FILES "")
+set(LFS_HRC_FILES "")
+if (EMSCRIPTEN)
+    include(FetchContent)
+    FetchContent_Declare(lpeg
+            URL  "https://github.com/roberto-ieru/LPeg/archive/refs/heads/master.zip"
+            SOURCE_DIR "${CMAKE_BINARY_DIR}/ep_lpeg"
```
lewri, line 14: From what I can read, this looks like should be in FetchContent_Populate
https://cmake.org/cmake/help/v3.14/module/FetchContent.html

TheCycoONE:
This is the way I recommend using FetchContent: https://github.com/CorsixTH/CorsixTH/blob/540826f8e8a2d4a32c6aa2650070d736c9a4f713/CorsixTH/CppTest/CMakeLists.txt#L23-L35

Actually I started playing with this myself and it doesn't work well because these are not cmake projects. vcpkg is much easier.

TheCycoONE:
The above would go in CorsixTH/ not CorsixTH/Src/

TheCycoONE:
Curious if you have tried emcmake - I think it should be possible to make fewer changes in cmake, though I don't know without trying it myself. SDL needs to be skipped, and maybe some others that are builtin, but mostly static linking normal find packages is the way to go.

lewri:
@ololoken just checking in if you wanted to continue still with this one

ARGAMX:
@ololoken how are you? Is there any chance that you will be interested in contributing to this PR some more?

ololoken:
@lewri @ARGAMX all good, just was focused on serious sam port. Yeah I will try to update the PR this weekend.

TheCycoONE:
@ololoken it might be worth waiting for or rebasing on top of the SDL3 PR. The impact on this is probably large.

TheCycoONE:
SDL3 is merged, good time to continue. https://wiki.libsdl.org/SDL3/README-emscripten

tobylane:
Thank you to ololoken and osuritz for your efforts here and in the [wasm branch](https://github.com/osuritz/corsixth-web/tree/wasm). For anyone looking to do the remaining work, with knowledge of emscripten and without AI, please jump in. The questions in #3464 will be of some guide, and we can help in Discord/Matrix.

PR #3095, opened 2025-10-25
FIx crash on hovering exploded rooms

Matroftt:
- Fixes crash that occurs when player hovers over a destroyed room. It is happening because when room destroyed queue is nil, and lua cant get array length from nil


ARGAMX:
Regression from #3030

PR #3097, opened 2025-10-26
Don't assume staff_member_set for all rooms

### LINKED ISSUE #3094: [Bug] Female cycling between door and screen in general diagnosis
### Describe the issue

Female was sent to general diagnosis.
Savegame is just before the visit (doctor and female still need to enter).

After the doctor, the female enters, changes cloths (I think twice), walks out again, turns 180 degrees, knocks at the door, enters, changes cloths ....

### Steps to Reproduce

[p1.zip](https://github.com/user-attachments/files/23141160/p1.zip)

2. Unpause

3. Watch female

### Expected Behaviour

After changing cloths, visit the crash trolley for an examination.

- [ ] Right mouse panning

### Save Game

### CorsixTH Version

f0cb8fd6 (current master)

### Operating System

Linux

### Theme Hospital Version

CD

### Gamelog.txt

```yaml
Nothing relevant exists in the log.
```

### Additional Information

_No response_

lewri:
*Fixes #3094*

- Staff member set was applying to all rooms at once following #2971 
- Reverts this, adds a note to override in derived class instead (we already do)
- Option commit for afterLoad, if we want it

### DIFF
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -1076,6 +1076,14 @@ function Room:afterLoad(old, new)
     self.waiting_staff_member = nil
     self.dealt_patient_callback = nil
   end
+  -- Patch the saves post 233 that labelled all rooms as having a staff_member_set
+  if old >= 233 and old < 235 then
+    local room_name = self.room_info.id
+    if room_name ~= "ward" and room_name ~= "operating_theatre" and
+        room_name ~= "research" then
+      self.staff_member_set = nil
+    end
+  end
 end
 
 --[[ Is the room one of the diagnosis rooms for the patient?
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local SDL = require("sdl")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 234 -- Overhaul training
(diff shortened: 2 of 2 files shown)


CorsixTH/Lua/room.lua
```diff
@@ -1075,6 +1075,13 @@ function Room:afterLoad(old, new)
     self.waiting_staff_member = nil
     self.dealt_patient_callback = nil
   end
+  -- Patch the saves post 233 that labelled all rooms as having a staff_member_set
+  if old >= 233 and old < 235 then
```
lewri, line 1080: I know we generally don't like doing this format of old. If we want the afterload and don't like this it can just be old < 235

TheCycoONE: It's fine

CorsixTH/Lua/room.lua
```diff
@@ -33,7 +33,7 @@ function Room:Room(x, y, w, h, id, room_info, world, hospital, door, door2)
   -- Serving staff in single occupancy rooms (like GD, Pharmacy and etc)
   self.staff_member = nil
   -- Serving staff list in multi-occupancy rooms (like Operating Theatre)
-  self.staff_member_set = {}
+  self.staff_member_set = nil -- Override with {} in derived class
```
TheCycoONE, line 37: Please explain why this has to be nil here / difference in behaviour so that the next dev doesn't come along and change it back.

Alberth289346: Alternatively, eliminate the single staff member case, so we don't have this confusion any more.

Issue #3098, opened 2025-10-26
Review Good First Issues


lewri:
Currently, there have been discussions that the current process of `Good First Issue` is generally unhelpful and full of traps or lack of information.

As part of a project to make this process better there needs to be changes to how these issues are handled. The following tasks will set us upon that road. The aim is to create a standardised format for how Good First Issues are presented in a clean and uncluttered manner:

- [x] Create a form for Good First Issues to be submitted by and merge this
- [x] Remove all existing `Good First Issue` labels from existing issues
- [x] Collate all affected issues, and consider their realistic feasibility to be a candidate as a `Good First Issue` again
- [ ] Convert any passing issues into the new form, closing their old issue in the process
- [ ] Create documentation for this process in the Wiki

The onus will rely on developers in future to properly undertake the task to ensure new Good First Issues are properly represented for someone new to the project.

lewri:
Existing GFIs:

## Wiki Related
~#1657 - Update Translators Wiki Documentation~ ✔️ 
~#1716 - Update FluidSynth compile instructions for Linux to Wiki~
#2064 - Wiki Updates for patients through Insurance

## Code related
#1730 - Emergency cheat breaks current epidemic
#1835 - Incorrect adviser message when dropping 3rd surgeon in Operating Theatre
#1914 - Font option is slightly confusing
~#2254 - mustPause windows can unpause a user-initiated pause~ ❌
~#2602 - disease.cure_price should use config values~
#2749 - Have a consistent layout for buttons in the various windows
~#2778 - Cheat Code to Make Machines Invulnerable to Damage~

In general Wiki changes are usually good ones to pick up for new people.

lewri:
Although 2254 was labelled a good first issue, I feel like it has the potential for hidden traps, which is what led to that issue coming about in the first place.

PR #3099, opened 2025-10-26
Gfi template

lewri:
*Contributes to #3098*

- Adds Good First Issue form template

Template visible at https://github.com/lewri/CorsixTH-Tests/issues

There is no way to actually lock down a template to certain users unfortunately, so the best we have is a 'Developer Use Only' message.


lewri:
Dirty branch

PR #3100, opened 2025-10-26
Good First Issue Template

lewri:
*Contributes to #3098*

- Adds Good First Issue form template

Template visible at https://github.com/lewri/CorsixTH-Tests/issues

There is no way to actually lock down a template to certain users unfortunately, so the best we have is a 'Developer Use Only' message.

### DIFF
--- a/.github/ISSUE_TEMPLATE/good_first_issue.yml
+++ b/.github/ISSUE_TEMPLATE/good_first_issue.yml
@@ -2,7 +2,7 @@
 name: Good First Issue [Developer Use Only]
 title: "Good First Issue: Enter Summary Here"
 description: For developer use only. Create a guided ticket for a first-time
-  contributor to the code to fix or implement something new.
+  contributor so that they fix the code or implement something new.
 projects: []
 labels:
   - Good First Issue
@@ -29,7 +29,7 @@ body:
   - type: input
     id: version
     attributes:
-      label: What CorsixTH version was the issue first discovered in?
+      label: Which CorsixTH version was the issue first discovered in?
       description: Ideally a release version
       placeholder: "0.69.1"
     validations:
@@ -37,9 +37,9 @@ body:
   - type: textarea
     id: reproduction
     attributes:
-      label: How can a player reproduce/see this in CorsixTH?
+      label: How can this be reproduced in CorsixTH?
       description: If there is a problem method to replicate, this helps the
-        contributor to check any fix works as intended
+        contributor to check that any fix works as intended
       value: |-
         1. *Load save game attached to this ticket*
 
@@ -63,13 +63,13 @@ body:
     validations:
       required: true
   - type: dropdown
-    id: dropdown-0
+    id: complexity
     attributes:
       label: Complexity Score
       description: |
        How difficult will this be for someone to undertake?
 
-       **Note: If this requires more than basic knowledge of Lua or C++ (arugably 
+       **Note: If this requires more than basic knowledge of Lua or C++ (arguably 
        any C++ is bad) it is not a Good First Issue!**
       options:
         - "1: Very Easy (no code knowledge required)"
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/good_first_issue.yml
```diff
@@ -0,0 +1,111 @@
...
+    attributes:
+      label: If this has been created from an existing issue, please add it below
+      description: Separate multiple issues with a comma
+      placeholder: "#1234"
+  - type: textarea
+    id: savegame
+    attributes:
+      label: Is a savegame to be provided?
+      description: Drag or upload it in the box below
+      value: "Savegame provided: Yes/No/Already in existing ticket"
```
MandrakeHorse: I think "save game" is preferred over "savegame"

.github/ISSUE_TEMPLATE/good_first_issue.yml
```diff
@@ -0,0 +1,111 @@
+# Note: This form can't be locked down to organisation memebers in GitHub currently.
+name: Good First Issue [Developer Use Only]
+title: "Good First Issue: Enter Summary Here"
+description: For developer use only. Create a guided ticket for a first-time
+  contributor to the code to fix or implement something new.
```
MandrakeHorse: Maybe adjust the wording from "to the code to fix" to something like "so that they can fix the code"

.github/ISSUE_TEMPLATE/good_first_issue.yml
```diff
@@ -0,0 +1,111 @@
...
+    id: what-is-issue
+    attributes:
+      label: Describe the issue
+      description: Give as much detail as possible
+    validations:
+      required: true
+  - type: input
+    id: version
+    attributes:
+      label: What CorsixTH version was the issue first discovered in?
```
MandrakeHorse: Change "What" to "Which"

.github/ISSUE_TEMPLATE/good_first_issue.yml
```diff
@@ -0,0 +1,111 @@
...
+      description: Ideally a release version
+      placeholder: "0.69.1"
+    validations:
+      required: true
+  - type: textarea
+    id: reproduction
+    attributes:
+      label: How can a player reproduce/see this in CorsixTH?
```
MandrakeHorse: "How can this issue be reproduced in CorsixTH?"

.github/ISSUE_TEMPLATE/good_first_issue.yml
```diff
@@ -0,0 +1,111 @@
...
+      description: If there is a problem method to replicate, this helps the
+        contributor to check any fix works as intended
```
MandrakeHorse: Change to "to check that any fix works as intended"

.github/ISSUE_TEMPLATE/good_first_issue.yml
```diff
@@ -0,0 +1,111 @@
...
+    validations:
+      required: true
+  - type: dropdown
+    id: dropdown-0
+    attributes:
+      label: Complexity Score
+      description: |
+       How difficult will this be for someone to undertake?
+
+       **Note: If this requires more than basic knowledge of Lua or C++ (arugably
```
MandrakeHorse: Typo: change to "arguably"

MandrakeHorse (changes requested):
Read through the template and have posted some suggested adjustments to the wording

lewri:
Updated. Changes not reflected in preview as just wording fixes.

TheCycoONE:
I'm concerned about an approach that requires identifying issues before they're created; and I'd rather not close and reopen issues; though the description could be updated to conform to the template instead.

lewri:
> I'm concerned about an approach that requires identifying issues before they're created; and I'd rather not close and reopen issues; though the description could be updated to conform to the template instead.

So a problem with some issues is that by the time they get identified as such to be a Good First Issue is they can be swathed in a bunch of noise that could deter from working on a solution. They might have loose or even dead ends in conversation, or conflicting ideas that are no longer relevant.  I also refer to notes of a comment in https://github.com/nodejs/admin/issues/50#issuecomment-359158182

The idea of this template is to 'spoon-feed' people for lack of a less crude word. It should provide everything needed, and a clear path to achieve it. It shouldn't have any distracting fluff for the ticket itself that the original may have in it to start with. It takes time to invest in a new person who wants to contribute, too, who may not do well being thrown in at the deep end from the start.

Of course, there's nothing stopping us from just submitting this template from the get-go, if it has been agreed elsewhere what the problem and solution is already known.

Issue #3040, opened 2025-09-21
Allow player to skip board game movies


MandrakeHorse:
### What feature or enhancement would you like to add?

The board game movies (AREA01V to AREA14V) that play before the start of every level (assuming that global movies is not disabled) cannot be skipped by pressing either the left mouse button or escape key. This is in contrast to the intro movie, win game movie, and lose movies, which can all be skipped. Presumably, this inconsistency relates to the fact that the audio for all the boardgame movies (DICEYFIN.wav and DICE122M.wav) are played separately.

### How would this feature benefit the gameplay of CorsixTH?

The player will be able to skip the level transition movies by pressing the escape key or the left mouse button.
Pressing either button will immediately terminate both the movie and the audio.

### Was the feature in the original game?

Unsure

### What version of CorsixTH are you using?

0.69.1

### Does this feature/enhancement relate to a current problem?

Yes - #3039

TheCycoONE:
This is indeed because the audio is played separately. There are a couple situations in the game where it would be preferable to interrupt the currently playing audio. If that is addressed they could fix this issue as well as the over aggressive chime that happens when you move the mouse over a list of staff in the current master.

PR #3104, opened 2025-10-27
Support for stop/pausing sound effects

TheCycoONE:
Fixes: #3040 

Adds support for pausing and stopping sound effects played through audio.lua.

To illustrate the change, the board game movie can now be paused and skipped like all other movies.

### DIFF
--- a/CorsixTH/Lua/movie_player.lua
+++ b/CorsixTH/Lua/movie_player.lua
@@ -90,6 +90,10 @@ local destroyMovie = function(me)
   else
     me.audio:playRandomBackgroundTrack()
   end
+  if me.sound then
+    me.audio:stopSound(me.sound)
+    me.sound = nil
+  end
   me.refresh_overlay = nil
   me.playing = false
   me.movie_over = false
@@ -156,13 +160,13 @@ function MoviePlayer:MoviePlayer(app, audio, video)
   self.playing = false
   self.holding_bg_music = false
   self.channel = -1
+  self.sound = nil
   self.lose_movies = {}
   self.advance_movies = {}
   self.intro_movie = nil
   self.demo_movie = nil
   self.win_game_movie = nil
   self.win_level_movie = nil
-  self.can_skip = true
   self.wait_for_stop = false
   self.wait_for_over = false
   self.movie_over = false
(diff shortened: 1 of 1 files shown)


CorsixTH/Src/th_sound.cpp
```diff
@@ -248,8 +248,7 @@ void sound_player::populate_from(sound_archive* pArchive) {
   sounds = new Mix_Chunk*[pArchive->get_number_of_sounds()];
   for (; sound_count < pArchive->get_number_of_sounds(); ++sound_count) {
     sounds[sound_count] = nullptr;
-    SDL_RWops* pRwop = pArchive->load_sound(sound_count);
-    if (pRwop) {
+    if (SDL_RWops* pRwop = pArchive->load_sound(sound_count)) {
```
Alberth289346, line 251: Please don't have side effects in conditions.

The compiler will merge the assignment and the test.

CorsixTH/Src/th_sound.h
```diff
@@ -72,20 +72,73 @@ class sound_player {
...
+  //! \param iX X coordinate of the sound effect.
+  //! \param iY Y coordinate of the sound effect.
+  //! \return The channel the sound is played on, or -1 on error.
+  int play_at(size_t iIndex, double dVolume, int iX, int iY);
+
+  //! Pause playback on a given channel if playing, or resume if paused.
+  //!
+  //! \param channel The channel to toggle pause on.
+  //! \return 1 if paused, 0 if resumed, -1 on invalid input.
+  int toggle_pause(int channel);
```
Alberth289346: Not too sure about the `1` meaning 'stopped'. You get inverted logic like `bool running = toggle_pause(x) == 0`.

Possibly the name is the root cause. Everything else seems to be about  playing, perhaps `toggle_play` as function name ?

CorsixTH/Src/th_sound.cpp
TheCycoONE: This is perfectly valid and encoraged c++ style. Instantiation in if is unambiguous and reduces scope.

CorsixTH/Src/th_sound.h
TheCycoONE: Pause is distinct from stop, not playing is ambiguous.

CorsixTH/Src/th_sound.cpp
Alberth289346: I didn't recognize it as such, since there is no semi-colon in the `if` condition.

Also, apparently the scope is the entire 'if' body.

CorsixTH/Src/th_sound.h
TheCycoONE: I'll switch this to an enum; I considered it at the time.

CorsixTH/Src/th_sound.cpp
TheCycoONE: Can recognize because in an equality check there can't be a type name at the start.

CorsixTH/Lua/audio.lua
```diff
@@ -300,12 +299,36 @@ function Audio:playSound(name, where, is_announcement, played_callback, played_c
...
+  if sound.channel and sound.channel >= 0 then
+    sound_fx:togglePause(sound.channel, sound.played_callback_id)
+  end
+end
+
+function Audio:stopSound(sound)
+  local sound_fx = self.sound_fx
+  if not sound_fx then
+    return
+  end
```
lewri, line 349: Not that I hope this is a silly question, but does stop mean the sound is effectively destroyed?
Luadoc in general on both functions needed.

TheCycoONE: Yes it does.

TheCycoONE:
I've added the luadoc and addressed the paused result.

CorsixTH/Src/th_sound.h
TheCycoONE: Converted to enum.

CorsixTH/Lua/audio.lua
TheCycoONE: Addressed in luadoc

Alberth289346 (approved):

PR #3109, opened 2025-10-29
Remove the correct diagnosis room

### LINKED ISSUE #100: Patient is heading to one room but sent to another.
During diagnosis patient goes to the GP and then is sent to a new room as shown in the debug action queue:

Put the seek_room action finds a different room even though the cardiogram is available (albeit the other side of the hospital).  

And the patient ends up in the ward.

I can understand why a new room might be chosen if the other was not available but there seems to be some form of randomisation going on when seeking a room. https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/humanoid_actions/seek_room.lua#L21 The room being sought isn't always the one found! 

This probably has little effect for gameplay but for game logic can produce unexpected behaviour with available treatment rooms. I noticed this because it gets in the way when I was trying to implement something. Any idea why this randomisation happens?

### LINKED ISSUE #3105: [Bug] Patient visits the same room several times
----

https://github.com/user-attachments/assets/a2740298-35d8-4c3d-921b-796bb068331b

### Describe the issue

The patient visits the same diagnostic room several times.
It should not generate new diagnostic points and iirc it was impossible in the original game.

----

### Steps to Reproduce

1. Load savegame attached to the issue.
2. There's only one patient in the entire hospital who goes from the GP's office to the psychiatry department. He has already visited the GP once. Keep an eye on him.
3. The patient will take the following route: GP -> Psycho -> GP -> Psycho -> GP -> Pharmacy.

----

### Expected Behaviour

The same patient should not visit the same room twice.
Instead of this the patient must:
1. Either via fax request for more diagnostic rooms.
2. Either ask via fax should he be treated with a certain probability of success.
3. Either the patient must decide to go home.

----

### Save Game

[3105 Patient Visit Psych Twice.sav.zip](https://github.com/user-attachments/files/23167167/3105.Patient.Visit.Psych.Twice.sav.zip)

[3105 Patient Visit Psych Twice 2 GP state.sav.zip](https://github.com/user-attachments/files/23195453/3105.Patient.Visit.Psych.Twice.2.GP.state.sav.zip)

----

### CorsixTH Version

0.69.1

----

lewri:
*Fixes #100*
*Fixed #3105*

- Fixes the seek room action removing the wrong diagnosis room from the `available_diagnosis_rooms` list
- Stops patients being sent to the same diagnosis room twice, hurrah!
- Some renaming of local variables

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/seek_room.lua
+++ b/CorsixTH/Lua/humanoid_actions/seek_room.lua
@@ -41,6 +41,9 @@ function SeekRoomAction:enableTreatmentRoom()
   return self
 end
 
+--! Sets the index of the room to use from the disease's available diagnosis
+-- rooms
+--!param room (integer) key in the available diagnosis room list
 function SeekRoomAction:setDiagnosisRoom(room)
   assert(type(room) == "number", "Invalid value for parameter 'room'")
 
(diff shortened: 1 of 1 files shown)


sadger:
Who disturbs my slumber? :imp:

lewri:
That's a separate pipeline. I am aware of it. This PR is not to fix that.

I'd actually need a save for that cycle too.

ARGAMX:
> That's a separate pipeline. I am aware of it. This PR is not to fix that.
> 
> I'd actually need a save for that cycle too.

Okay.
But then we fix one defect and expose another.
Because it turns out the fixable defect was hiding this one.
Because I've never seen anything like this before.
So we'll have to fix it anyway?
What do you propose to do with this?
How do we usually deal with PRs that knowingly create/reveal new defects?

This is save 2 from the issue.

lewri:
I'll have a look, but if I can't fix that it'll need a new ticket

Edit: I cant reproduce from the save

lewri:
Did we reach consensus if we believe on balance this pr hasn't caused this behaviour in the end?

ARGAMX (approved):
Okay 👌

PR #3110, opened 2025-10-30
Cut down on hover chime

TheCycoONE:
Only play one hover sound at a time in various dialogs. This keeps the hover noise from escalating when moving the mouse quickly over a list of items.

### DIFF
--- a/CorsixTH/Lua/dialogs/place_objects.lua
+++ b/CorsixTH/Lua/dialogs/place_objects.lua
@@ -234,6 +234,7 @@ function UIPlaceObjects:addObjects(object_list, pay_for)
 
   self.active_index = 0 -- avoid case of index changing from 1 to 1
   self.active_hover_index = 0
+  self.hover_sound = nil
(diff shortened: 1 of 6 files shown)


CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -596,7 +597,11 @@ function UIPlaceObjects:onMouseMove(x, y, dx, dy)
     if (header_height+y) % bar_height < 21 then
       current_hover_id = math_floor((y - header_height)/bar_height) + 1
       if self.active_hover_index ~= current_hover_id then
-        self.ui:playSound("Hlight5.wav")
+        if self.hover_sound then
+          require('debugger')()
```
ARGAMX: Is this some debug code that slipped in here?

TheCycoONE: Yep, I caught that, thought I already pushed the fix but it's just sitting on my local so I'll get that.

PR #3112, opened 2025-10-31
World: apply level_config StartPrice to cure_price (init + afterLoad)

### LINKED ISSUE #2602: disease.cure_price should use config values
Currently all our disease files use a hard coded value for cure cost.
However, the level config files can define these using `#expertise[x].StartPrice`

Changing these shouldn't be too difficult; it will just require applying a for loop to world.available_diseases during game creation to set `disease.cure_price`. `World:initLevel(...)` is likely a good entry point, as the available diseases table is constructed there.
E.g.
```lua
disease.cure_price = level_config.expertise[disease.expertise_id].StartPrice
```

For applying this change to an existing game, it will need an afterload in `World` to cycle through available diseases and update the value.

jur4jmiklo5:
### Summary
- On new game (`World:initLevel`), set each `disease.cure_price` from `level_config.expertise[expertise_id].StartPrice` .
- On save load (`World:afterLoad`), re-apply the same so updated level data also affect existing saves.
- Uses `~= nil` so `StartPrice = 0` is respected.

### Testing
- **Init:** Verified in logs that `bloaty_head` follows `expertise[2].StartPrice`.
- **Load:** Temporarily modified `StartPrice` in `Map:afterLoad` for a save; logs show `AFTERLOAD` updating `cure_price`.
- **UI:** Drug Casebook reflects the StartPrice (with Hospital Policy multiplier at default).

Fixes #2602

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -249,20 +249,26 @@ function World:initLevel(app, avail_rooms)
     end
   end
   -- Apply level-defined pricing to diseases on new game:
-  if level_config and level_config.expertise and #self.available_diseases > 0 then
-    for _, dis in ipairs(self.available_diseases) do
-      local ex = dis.expertise_id and level_config.expertise[dis.expertise_id]
-      if ex and ex.StartPrice then
-        dis.cure_price = ex.StartPrice
-      end
-    end
-  end
+  self:applyLevelStartPrices()
   if #self.available_diseases == 0 and self.map.level_number ~= "MAP EDITOR" then
     -- No diseases are needed if we're actually in the map editor!
     print("Warning: This level does not contain any diseases")
   end
 
   self:determineWinningConditions()
+end
+
+-- Apply level-defined pricing to all available diseases.
+function World:applyLevelStartPrices()
+  local lc = self.map and self.map.level_config
+  if not (lc and lc.expertise and self.available_diseases) then return end
+
+  for _, dis in ipairs(self.available_diseases) do
+    local ex = dis.expertise_id and lc.expertise[dis.expertise_id]
+    if ex and ex.StartPrice then
+      dis.cure_price = ex.StartPrice
+    end
+  end
 end
 
 function World:toggleInformation()
@@ -2212,17 +2218,7 @@ end
 function World:afterLoad(old, new)
   self:setUI(self.ui)
   -- Re-apply level-defined pricing when loading saves:
-  do
-    local lc = self.map and self.map.level_config
-    if lc and lc.expertise and self.available_diseases then
-      for _, dis in ipairs(self.available_diseases) do
-        local ex = dis.expertise_id and lc.expertise[dis.expertise_id]
-        if ex and ex.StartPrice then
-          dis.cure_price = ex.StartPrice
-        end
-      end
-    end
-  end
+  self:applyLevelStartPrices()
   -- insert global compatibility code here
   if old < 4 then
     self.room_built = {}


CorsixTH/Lua/world.lua
```diff
@@ -248,6 +248,15 @@ function World:initLevel(app, avail_rooms)
       end
     end
   end
+  -- Apply level-defined pricing to diseases on new game:
+  if level_config and level_config.expertise and #self.available_diseases > 0 then
+    for _, dis in ipairs(self.available_diseases) do
+      local ex = dis.expertise_id and level_config.expertise[dis.expertise_id]
+      if ex and ex.StartPrice ~= nil then
```
TheCycoONE: Based on your commit message you may not be aware, 0 is truthy in lua (unlike javascript) so the ~= nil isn't needed.

jur4jmiklo5: Good catch — updated both blocks to use `if ex and ex.StartPrice`. Thanks!

CorsixTH/Lua/world.lua
```diff
@@ -2202,7 +2211,18 @@ end
...
+  do
+    local lc = self.map and self.map.level_config
+    if lc and lc.expertise and self.available_diseases then
+      for _, dis in ipairs(self.available_diseases) do
+        local ex = dis.expertise_id and lc.expertise[dis.expertise_id]
+        if ex and ex.StartPrice then
+          dis.cure_price = ex.StartPrice
+        end
+      end
+    end
```
ARGAMX: It seems like this added block of code and the above block of code essentially do the same thing?
Does that mean they can both be moved into a separate function and just called here?

jur4jmiklo5: Refactored into `World:applyLevelStartPrices()` and replaced duplicate code in `initLevel`/`afterLoad`. Thanks!

CorsixTH/Lua/world.lua
```diff
@@ -2202,7 +2217,8 @@ end
 --!param new (integer) The current version of the save game format.
 function World:afterLoad(old, new)
   self:setUI(self.ui)
-
+  -- Re-apply level-defined pricing when loading saves:
+  self:applyLevelStartPrices()
```
lewri, line 2221: I'm curious why this isn't after the end of all the old blocks. But also can't see a problem why it can't exist at the top too..
It is non-conventional to reload start prices on every save from a glance. Editing the `.LEVEL` file would actually have no effect on saved games as far as I can tell; map class already has its data and refuses to change without an afterload (like your testing shows).

CorsixTH/Lua/world.lua
```diff
@@ -256,6 +258,19 @@ function World:initLevel(app, avail_rooms)
...
 
+-- Apply level-defined pricing to all available diseases.
+function World:applyLevelStartPrices()
+  local lc = self.map and self.map.level_config
+  if not (lc and lc.expertise and self.available_diseases) then return end
+
+  for _, dis in ipairs(self.available_diseases) do
+    local ex = dis.expertise_id and lc.expertise[dis.expertise_id]
+    if ex and ex.StartPrice then
+      dis.cure_price = ex.StartPrice
```
lewri, line 269: We can afford proper named local variables here, there's no need to be stingy.

lc = config
dis = disease
ex = expertise (arguably expertise is a bad name from TH)

CorsixTH/Lua/world.lua
TheCycoONE: Unconventional but perhaps the right thing to do - I've talked to level designers who wanted to be able to see other changes in saves. The main counter-argument I could see is that it might not work for everything and that makes it harder to predict what will / won't take when loading a save.

CorsixTH/Lua/world.lua
TheCycoONE: length of a variable name proportional to the scope - none of these are long lived enough to have any difficulty tracing their meaning. I was fine with it.

PR #3113, opened 2025-11-02
Improving patients dying

### LINKED ISSUE #1584: [Bug] Crash on trying to cure a dying patient in Pharmacy
#### Describe the issue
In the attached saved game, the game will freeze after a few seconds - people stop moving, and the zoom function no longer works. On return to the menu, the mouse cursor is hidden and will only appears when clicking.

On loading a different game, the action is still frozen, unless the application is completely closed and restarted

I'm not sure if this is related, but the saved game recently experienced an earthquake, damaging some of machinery. 

Image

#### Steps to Reproduce
There are two saved games in the zip file. The first `before ` will run for a few in-game days until 8 December when it will freeze, and the second is in frozen already on loading.

#### Save Game
[Saves.zip](https://github.com/CorsixTH/CorsixTH/files/4240558/Saves.zip)

#### Expected Behavior
The game should continue to run past December 8 and the mouse cursor should remain visible when returning to the menu.

#### System Information
.63, pre-compiled binary

Windows 10

CD, circa 2001

#### Gamelog.txt
```
Paste gamelog.txt output here
```
```
Created game with savegame version 134.

Savegame version is 134 (v0.63), originally it was 134 (v0.63)
Savegame version is 134 (v0.63), originally it was 134 (v0.63)

Error in timer handler: 
Lua\entities\humanoid.lua:906: Action queue was empty. This should never happen.
humanoid - class: Standard Female Patient
Warmth: 0.288   Happiness: 1.000   Fatigue: 0.000  Thirst: 0.219  Toilet_Need: 0.017   Health: 0.000
Actions: []
stack traceback:
	[C]: in function 'error'
	Lua\entities\humanoid.lua:906: in method 'getCurrentAction'
	Lua\hospital.lua:698: in method 'countSittingStanding'
	Lua\hospital.lua:728: in method 'checkFacilities'
	Lua\hospital.lua:982: in method 'onEndDay'
	Lua\world.lua:1004: in method 'onTick'
	Lua\app.lua:1111: in function <Lua\app.lua:1108>
	(...tail calls...)
```

### LINKED ISSUE #3055: [Bug] Emergency patients can't die

ARGAMX:
*Fixes #3055*
*Fixes #1584*

- Now if it's time for a patient to die by the timer, instead of trying to kill him instantly, he is transferred to a state `set_to_die`.
- `patient.set_to_die = true` a more transparent and understandable state name instead of `patient.attributes["health"] = 0.0`.
- Each tick `going_to_die` is being checked. If it is true and all other conditions and situation are suitable, only then the dying scenario is launched.
- Added checks for the patient knocking on the door or opening the door as inappropriate moments for running the dying scenario.
- The scenario of dying after unsuccessful treatment does not fall under this category, since there is no delay and the patient's situation is known in advance at treatment end phase.
- Previously when the patient's dying scenario was triggered, the `patient.going_home` flag was set to true (If patient goes to heaven, this is this how patient goes home?...), which is logically incorrect and created traps in other places. Instead there is new a `patient.going_to_die`.

```lua
  -- Indicates that the patient is being put on death. Patient may be busy,
  -- But as soon as he is deoccupy he will start the dying scenario based on this flag
  -- Except cases are when he is cured or sent home while being busy
  self.set_to_die = false
  -- Indicates that the patient on the dying scenario
  -- Cure should not happen at this state. Patient can't be sent home
  self.going_to_die = false
    -- To distinguish between actually being dead and having a nil hospital
  self.dead = false
  ```
  
  Perhaps more understandable names could be thought up for these flags.

### DIFF
--- a/CorsixTH/Lua/dialogs/patient.lua
+++ b/CorsixTH/Lua/dialogs/patient.lua
@@ -285,7 +285,7 @@ function UIPatient:updateInformation()
   self.disease_blanker.visible = not show_dis_btn
 
   -- Show go home?
-  local show_home_btn = not patient.going_home
+  local show_home_btn = not patient.going_home and not patient.going_die
   self.home_button.enabled = show_home_btn
   self.home_button.visible = show_home_btn
   self.home_blanker.visible = not show_home_btn
@@ -294,6 +294,7 @@ function UIPatient:updateInformation()
   local show_guess_btn = not (patient.is_debug or
       patient.diagnosis_progress == 0 or
       patient.diagnosed or
+      patient.going_die or
       patient.going_home)
   self.guess_button.enabled = show_guess_btn
   self.guess_button.visible = show_guess_btn
@@ -330,7 +331,7 @@ function UIPatient:guessDisease()
   local patient = self.patient
   -- NB: the first line of conditions should already be ruled out by button being disabled, but just in case
   if patient.is_debug or patient.diagnosis_progress == 0 or patient.diagnosed or
-      patient.going_home or patient:getRoom() or
+      patient.going_home or patient.going_die or patient:getRoom() or
       not patient.hospital.disease_casebook[patient.disease.id].discovered then
     self.ui:playSound("wrong2.wav")
     return
--- a/CorsixTH/Lua/dialogs/patient.lua
+++ b/CorsixTH/Lua/dialogs/patient.lua
@@ -285,7 +285,7 @@ function UIPatient:updateInformation()
   self.disease_blanker.visible = not show_dis_btn
 
   -- Show go home?
-  local show_home_btn = not patient.going_home and not patient.going_die
+  local show_home_btn = not patient.going_home and not patient.going_to_die
   self.home_button.enabled = show_home_btn
   self.home_button.visible = show_home_btn
   self.home_blanker.visible = not show_home_btn
@@ -294,7 +294,7 @@ function UIPatient:updateInformation()
   local show_guess_btn = not (patient.is_debug or
       patient.diagnosis_progress == 0 or
       patient.diagnosed or
-      patient.going_die or
+      patient.going_to_die or
       patient.going_home)
   self.guess_button.enabled = show_guess_btn
   self.guess_button.visible = show_guess_btn
@@ -331,7 +331,7 @@ function UIPatient:guessDisease()
   local patient = self.patient
   -- NB: the first line of conditions should already be ruled out by button being disabled, but just in case
   if patient.is_debug or patient.diagnosis_progress == 0 or patient.diagnosed or
-      patient.going_home or patient.going_die or patient:getRoom() or
+      patient.going_home or patient.going_to_die or patient:getRoom() or
       not patient.hospital.disease_casebook[patient.disease.id].discovered then
     self.ui:playSound("wrong2.wav")
     return
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -700,9 +700,9 @@ function Humanoid:isLeaving()
   return self.action_queue[1].is_leaving and true or false
 end
 
--- Check if the humanoid is knocking door
+-- Check if the humanoid is knocking on a door
 function Humanoid:isKnockingDoor()
-  return class.is(self.action_queue[1], KnockDoorAction)
+  return class.is(self:getCurrentAction(), KnockDoorAction)
 end
 
 -- Check if the humanoid is entering room
@@ -906,7 +906,7 @@ end
 -- feeling of warmth accordingly. Returns whether the calling function should proceed.
 function Humanoid:tickDay()
   -- No use doing anything if we're going home/fired (or dead)
-  if self.going_home or self.going_die or self.dead then
+  if self.going_home or self.going_to_die or self.dead then
     return false
   end
 
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -700,6 +700,19 @@ function Humanoid:isLeaving()
   return self.action_queue[1].is_leaving and true or false
 end
 
+-- Check if the humanoid is knocking door
+function Humanoid:isKnockingDoor()
+  return class.is(self.action_queue[1], KnockDoorAction)
+end
+
+-- Check if the humanoid is entering room
+function Humanoid:isEnteringRoom()
+  if self.user_of and self.user_of.object_type.id == "door" then
+    return true
+  end
+  return false
+end
+
 -- Check if there is "is_leaving" action in the action queue
 function Humanoid:hasLeavingAction()
   for _, action in ipairs(self.action_queue) do
@@ -893,7 +906,7 @@ end
 -- feeling of warmth accordingly. Returns whether the calling function should proceed.
 function Humanoid:tickDay()
   -- No use doing anything if we're going home/fired (or dead)
-  if self.going_home or self.dead then
+  if self.going_home or self.going_die or self.dead then
     return false
   end
 
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -39,13 +39,13 @@ function Patient:Patient(...)
   self.pay_amount = 0
   -- To distinguish between actually being dead and having a nil hospital
   self.dead = false
-  -- Indicates that the patient is being put on death. Patient may be busy,
-  -- But as soon as he is deoccupy he will start the dying scenario based on this flag
-  -- Except cases are when he is cured or sent home while being busy
+  -- Indicates that the patient is being marked for death. The patient may be busy,
+  -- But as soon as the patient is unoccupied he will start the dying scenario based
+  -- on this flag, except the cases when he is cured or sent home while being busy.
   self.set_to_die = false
   -- Indicates that the patient on the dying scenario
   -- Cure should not happen at this state. Patient can't be sent home
-  self.going_die = false
+  self.going_to_die = false
   -- Is the patient reserved for a particular nurse when being vaccinated
   self.reserved_for = false
   self.vaccinated = false
@@ -100,7 +100,7 @@ function Patient:onClick(ui, button)
     self.user_of:onClick(ui, button)
   elseif button == "right" then
     -- Attempt to push patient over
-    if not self.world:isPaused() and not (self.cured or self.going_die or self.dead or self.going_home)
+    if not self.world:isPaused() and not (self.cured or self.going_to_die or self.dead or self.going_home)
          and math.random(1, 2) == 2 then
       self:falling(true)
     end
@@ -373,7 +373,7 @@ function Patient:die()
   -- Remove any messages and/or callbacks related to the patient.
   self:unregisterCallbacks()
 
-  self.going_die = true
+  self.going_to_die = true
   if self:getRoom() then
     self:queueAction(MeanderAction():setCount(1))
   else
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: CorsixTH/Lua/dialogs/patient.lua [hunk 1]
```diff
@@ -294,6 +294,7 @@ function UIPatient:updateInformation()
   local show_guess_btn = not (patient.is_debug or
       patient.diagnosis_progress == 0 or
       patient.diagnosed or
+      patient.going_die or
```
lewri: going_to_die?
I'm surprised we didn't have a self.dying variable before.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -700,6 +700,19 @@ function Humanoid:isLeaving()
   return self.action_queue[1].is_leaving and true or false
 end
 
+-- Check if the humanoid is knocking door
```
lewri: Knocking on a door?

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -39,6 +39,13 @@ function Patient:Patient(...)
   self.pay_amount = 0
   -- To distinguish between actually being dead and having a nil hospital
   self.dead = false
+  -- Indicates that the patient is being put on death. Patient may be busy,
```
lewri: "marked for death" might fit better, plus gives a nice yarr harr vibe ☠️

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -39,6 +39,13 @@ function Patient:Patient(...)
...
+  -- But as soon as he is deoccupy he will start the dying scenario based on this flag
```
lewri: Patients are not a singular gender, call them patients, he/she, or they.

It isn't your fault the code has got the singular references already written, but we shouldn't be adding more. Maybe one day I'll go fix them.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -962,7 +990,7 @@ end
 function Patient:updateDynamicInfo()
   local action_string = self.action_string or ""
   local info = ""
-  if self.going_home then
+  if self.going_home or self.going_die then
```
lewri: While fine, just note that Patient:die does update dynamic info directly. Not sure why it doesn't use this function instead but that's an aside.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -39,6 +39,13 @@ function Patient:Patient(...)
   self.pay_amount = 0
   -- To distinguish between actually being dead and having a nil hospital
   self.dead = false
+  -- Indicates that the patient is being put on death. Patient may be busy,
+  -- But as soon as he is deoccupy he will start the dying scenario based on this flag
+  -- Except cases are when he is cured or sent home while being busy
+  self.set_to_die = false
+  -- Indicates that the patient on the dying scenario
+  -- Cure should not happen at this state. Patient can't be sent home
+  self.going_die = false
```
lewri: going_to_die?
I'm surprised we didn't have a self.dying variable before.

CorsixTH/Lua/hospital.lua
```diff
@@ -1136,8 +1136,8 @@ function Hospital:resolveEmergency()
   local rescued_patients = emer.cured_emergency_patients
   for _, patient in ipairs(self.emergency_patients) do
     if patient and not patient.cured and not patient.dead
-        and not patient.going_home and not patient:getRoom() then
-      patient:die()
+        and not patient.going_home then
+      patient:setToDying()
```
lewri, line 1140: We still end up with a floating health of not 0. This affects the health graph in the patient's UI as not tying up with the format expected of other patients. Generally speaking, you only die if you have no health. The factor should also be adjusted to 0 as a mop-up.

In fact, without it, Patient:tickDay will keep firing its whole shebang on emergency patients whereas it is stopped for others.

CorsixTH/Lua/entities/humanoids/patient.lua
lewri: I think you mean 'unoccupied' for the other part of that line.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -39,6 +39,13 @@ function Patient:Patient(...)
   self.pay_amount = 0
   -- To distinguish between actually being dead and having a nil hospital
   self.dead = false
+  -- Indicates that the patient is being put on death. Patient may be busy,
+  -- But as soon as he is deoccupy he will start the dying scenario based on this flag
+  -- Except cases are when he is cured or sent home while being busy
+  self.set_to_die = false
```
lewri, line 45: Again, my brain yearns for `marked_for_death` but I can't think of any reason not to use `set_to_die`

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -700,6 +700,19 @@ function Humanoid:isLeaving()
...
+function Humanoid:isKnockingDoor()
+  return class.is(self.action_queue[1], KnockDoorAction)
```
lewri: The proper call is `self:getCurrentAction()`

lewri (commented):
You really want to bait the next version's changelog notes don't you 😄 

Hesitation on adding new variable checks to functions, not having a helper function instead to ask for the info instead. Otherwise, will be good to see how things go on this change.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: I didn't even pay attention to it.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: 👍

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: 👍

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: 👍

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: 👍

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: 👍

CorsixTH/Lua/hospital.lua
ARGAMX: > We still end up with a floating health of not 0. This affects the health graph in the patient's UI as not tying up with the format expected of other patients. Generally speaking, you only die if you have no health. The factor should also be adjusted to 0 as a mop-up.
> 
> In fact, without it, Patient:tickDay will keep firing its whole shebang on emergency patients whereas it is stopped for others.

Mostly, I agree.
But this isn't related to the bugs that is being fixed.
And it wasn't in the scope of the task.
I haven't really investigated this problem.
I have a proposal: you could create a PR to fix this. 👍

TheCycoONE:
going_die (ick) and going_to_die are a little close for comfort.

ARGAMX:
> going_die (ick) and going_to_die are a little close for comfort.

Sorry, I don't understand what this is meant

CorsixTH/Lua/entities/humanoids/patient.lua
TheCycoONE: self.going_die how is this different than going_to_die, why such similar names. Easily confused.

ARGAMX: > self.going_die how is this different than going_to_die, why such similar names. Easily confused.

Ahh, now I get it.
Yes, this is a mistake.
It's the same variable.
I was working on Lewri's suggestion to rename `going_die` to `going_to_die` and I somehow fail to rename it everywhere. I missed two places.
I've fixed that now.
It should be `going_to_die` everywhere.

CorsixTH/Lua/hospital.lua
```diff
@@ -1136,8 +1136,8 @@ function Hospital:resolveEmergency()
   local rescued_patients = emer.cured_emergency_patients
   for _, patient in ipairs(self.emergency_patients) do
     if patient and not patient.cured and not patient.dead
-        and not patient.going_home and not patient:getRoom() then
-      patient:die()
+        and not patient.going_home then
```
lewri, line 1139:
```suggestion
        and not patient.going_home then
          patient:changeAttribute("health", patient:getAttribute("health") * -1)
```
Will that work instead of a new PR?

CorsixTH/Lua/hospital.lua
lewri: See https://github.com/CorsixTH/CorsixTH/pull/3113/files#r2491630152 , if you're happy to just have a commit with it on. Otherwise you'll end up in rebase hell.

CorsixTH/Lua/hospital.lua
ARGAMX: > Will that work instead of a new PR?

May be just 
`patient.attributes["health"] = 0.0` 
?

`patient:changeAttribute("health", patient:getAttribute("health") * -1)`
looks a bit complicated and it’s not clear why it was done this way.
The problem here that you don't want to work with either float or int?
This code raises questions.

I think instead of such direct changes, we should call functions.
If we want to reset the patient's health, then we should call the function that's called that.
And it should be the function's responsibility to take into account whether it's a float or an int.

this is why I call `patient:setToDying()` here instead of just `patient.set_to_die = true`.

lewri: But this is calling the function `Patient:changeAttribute` to make the change, and is also calling to function `Patient:getAttribute` to get the current health value, no?
I don't agree direct editing the attribute when functions are available is correct, especially if you invoke it from the hospital.

It could use a comment or local, if so be it.

lewri:
```suggestion
        and not patient.going_home then
          -- Set the patient's health to 0 before we die
          patient:changeAttribute("health", patient:getAttribute("health") * -1)
```

lewri (approved):

Alberth289346:
Late to the party so I understand if you don't want to do this here or now.

``` lua
  -- Except cases are when he is cured or sent home while being busy
  self.set_to_die = false
  -- Indicates that the patient on the dying scenario
  -- Cure should not happen at this state. Patient can't be sent home
  self.going_to_die = false
  -- To distinguish between actually being dead and having a nil hospital
  self.dead = false
```
States are good, but in this way you can have both `self.going_to_die = true` and `self.dead = true` at the same time, and then you're lost what the real state is.  
Also, finding the current state is now a lot of searching (you need to check all the flags), and changing the state is less than trivial.

The simple solution is to have `self.state` (or even `self._state` making it explicit others are not to use it)), and assigning constants to it.

String literals or unique numbers as constants is one solution, but typos and mixed-up numbers may be troublesome.
To avoid these, create constants as part of the class definition, eg
`Patient._is_dead = "_is_dead"` and use `Patient._is_dead` everywhere, eg `if self._state == Patient._is_dead then ...`.

PR #3114, opened 2025-11-02
[WIP] Add proxies for partial rendering of the same animation from multiple tiles.

Alberth289346:
*Fixes #*

- Adds animation proxy objects at tiles with possible cropping.
- The proxy objects are like the `render_attach_points` in Lua, except they are in cpp.
- Also, unlike in Lua, the animation is not duplicated. Instead the same animation is painted once for each proxy object. By setting cropping limits at the proxies, duplicate painting of the same parts can be avoided.

Much of the cpp code seems to be there. Things to do:
- [ ] Actually attach proxies to their tile.
- [x] Test if it does anything useful.
  - Render code is a simple double loop to draw in positive X direction now, without any early entities or multiple `rap`s.
  - Graphics are at the expected place for the simple case that I tried (since loading a save is not feasible is my guess).
- [ ] Extend the cpp persistance save and load code in some way.
- [ ] Take out the render attach code.
  - Disabled it by commenting it away for now.
- [ ] Add proxies to animations that need it.

### DIFF
--- a/CorsixTH/Src/th_gfx.h
+++ b/CorsixTH/Src/th_gfx.h
@@ -56,9 +56,11 @@ enum draw_flags : uint32_t {
   thdf_flip_horizontal = 1 << 0,
   //! Draw with the top becoming the bottom and vice versa
   thdf_flip_vertical = 1 << 1,
-  //! Draw with 50% transparency. If set together with thdf_alpha_75, don't draw at all.
+  //! Draw with 50% transparency. If set together with thdf_alpha_75, don't draw
+  //! at all.
   thdf_alpha_50 = 1 << 2,
-  //! Draw with 75% transparency. If set together with thdf_alpha_50, don't draw at all.
+  //! Draw with 75% transparency. If set together with thdf_alpha_50, don't draw
+  //! at all.
   thdf_alpha_75 = 1 << 3,
   //! Draw using a remapped palette
   thdf_alt_palette = 1 << 4,
@@ -497,7 +499,7 @@ class animation_base : public drawable {
   ::layers layers{};
 };
 
-class animation_proxy: public animation_base {
+class animation_proxy : public animation_base {
  public:
   animation_proxy(animation* const parent_anim, int8_t dx, int8_t dy,
                   int8_t crop_base, int8_t crop_width);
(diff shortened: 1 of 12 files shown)


CorsixTH/Src/th_lua_anims.cpp
```diff
@@ -55,7 +55,7 @@ int l_anims_set_spritesheet(lua_State* L) {
 
 //! Set the video target for the sprites.
 /*!
-    setCanvas(<video-surface>)
+ *  setCanvas(<video-surface>)
```
TheCycoONE, line 58: Not sure how consistent we've been on this. In th_gfx.h when we use this style we don't have a `*` on each line.

Maybe see how it renders.

Alberth289346: My editor added the `*` at each line. As I am used to that in Java, I added a few more.

I don't have strong preferences here.

Alberth289346:
I am somewhat considering to make `animation` purely an object that renders images without being connected to a tile.

Not now however.

CorsixTH/Src/th_gfx.h
```diff
@@ -495,12 +497,41 @@ class animation_base : public drawable {
   ::layers layers{};
 };
 
+class animation_proxy: public animation_base {
```
TheCycoONE: Preferably with some documentation about what a proxy is.

CorsixTH/Src/th_lua_anims.cpp
TheCycoONE: I'm fine with it as long as doxygen doesn't throw *s in the output; and doxygen mentions this style in their documentation so it shouldn't: https://www.doxygen.nl/manual/docblocks.html

CorsixTH/Src/th_map.h
```diff
@@ -285,6 +285,10 @@ class level_map {
   void update_temperatures(uint16_t iAirTemperature,
                            uint16_t iRadiatorTemperature);
 
+  bool is_on_map(int x, int y) {
```
TheCycoONE, line 288: const function

Alberth289346:
I pondered about rendering of tiles, and figured that you cannot render anything until the north and east side have been drawn. That holds at least for the north and east tiles that may get partly hidden.
So I concluded that the code should render when it reaches a far west or south tile of the object. For the far west case, it should perhaps only draw the part that could be hidden by its neighbour at the right (ie half a tile). For the far south tiles, it should draw the full tile, except for the part already covered by neighbour tiles at the west (mostly due to  performance).

If that is the case, then an algorithm could compute the tiles and parts to draw from a footprint.

PR #3115, opened 2025-11-04
Expand variables on 3112

lewri:
Give abbreviated locals in new function full names

https://github.com/CorsixTH/CorsixTH/pull/3112


ARGAMX (approved):
👍

TheCycoONE (approved):
As I said on the other PR, I don't think this is necessary where the lifetime of the variable is only a couple of lines; it's trivial to see what it means. 🤷

PR #3117, opened 2025-11-06
Midi device support

### LINKED ISSUE #1236: Feature Request: Midi output
Hi, Please add an option to allow MIDI output to be redirected to an external device/Midi port. I have a Roland Sound Canvas 55 Hardware Midi Synthesizer and would love to play the game as it was originally intended to be heard. 
pmidi -l on Linux shows
 24:0     CH345                             CH345 MIDI 1
I can set Dosbox to output Midi to this port, and pmidi -p 24:0 midi.mid will play a midi file on the synthesizer.

TheCycoONE:
Fixes #1236

Add support for playback over a selected MIDI device. Additionally adds a consolidated option screen for sound related settings including the soundfont file selection.

image

To use a MIDI device you must go into the new Sound options section and select an API other than the default, e.g. Windows MM for Windows; as well as the device you want to target.

### DIFF
--- a/CorsixTH/Lua/window.lua
+++ b/CorsixTH/Lua/window.lua
@@ -2023,7 +2023,9 @@ function Window:getTooltipAt(x, y)
     end
   end
   for _, btn in ipairs(self.buttons) do
-    if btn.visible ~= false and btn.tooltip and btn.x <= x and x < btn.r and btn.y <= y and y < btn.b then
+    if btn.panel_for_sprite.visible ~= false and
+        btn.tooltip and
+        btn.x <= x and x < btn.r and btn.y <= y and y < btn.b then
       return self:getTooltipForElement(btn.tooltip, x, y)
     end
   end


CorsixTH/Src/xmi2mid.cpp
```diff
@@ -236,48 +253,56 @@ uint8_t* transcode_xmi_to_midi(const unsigned char* xmi_data, size_t xmi_length,
...
+                    time_multiplier,
+            (iTokenType & 0x0F) | 0x80);
         pToken->data = iExtendedType;
         pToken->buffer = "\0";
         break;
       case 0xF0:
         iExtendedType = 0;
-        if (iTokenType == 0xFF) {
+        if (iTokenType ==
+            0xFF) {  // Meta event, should not be sent to synthesizer
```
Alberth289346:
```suggestion
        if (iTokenType == 0xFF) {
          // Meta event, should not be sent to synthesizer
```
Seems a bit nicer to me.

TheCycoONE: I've changed all this since then, now uses named constants.

TheCycoONE: Still - much too early to review.

CorsixTH/Src/midi_player.h
```diff
@@ -0,0 +1,309 @@
...
+   * XMI files
+   * @param api The MIDI API to use, one of:
+   *            - <empty string> or "Native" - Any available API
+   *            - "ALSA" - Linux ALSA
+   *            - "JACK" - JACK on Linux and Unix where available
+   *            - "CoreMIDI" - MacOS CoreMIDI
+   *            - "Windows MM" - Windows MIDI Api
+   *            - "Web MIDI API" - Web MIDI API
+   *            - "Windows UWP" - Universal Windows Platform MIDI API (RtMidi
+   * 6+)
```
Alberth289346: Line is just too long. Perhaps break the line after "API"?
(For some reason I cannot suggest that.

CorsixTH/Src/midi_player.h
```diff
@@ -0,0 +1,309 @@
...
+   */
+  void pause();
+
+  /**
+   * Resume the current MIDI track if paused
+   */
+  void resume();
+
+ private:
+  /**!
```
Alberth289346: Does it need the exclamation mark?  Other comments don't have it.

CorsixTH/Lua/config_finder.lua
```diff
@@ -488,6 +491,26 @@ audio_music = nil -- [[X:\ThemeHospital\Music]]
...
+-------------------------------------------------------------------------------
+-- Midi API and Device settings.
+-- By default, CorsixTH uses FluidSynth or build defined MIDI synthesizer.
+-- You can change the API to target other available MIDI backends using these
+-- settings on supported platforms.
+-- Possible values for midi_api are:
+--   <nil>      - Uses SDL_Mixer's MIDI backend, typically FluidSynth
+--   Native     - Use any available platform MIDI API
+--   ALSA       - Use the ALSA MIDI API (Linux only)
+--   JACK       - Use the JACK MIDI API (Unix-like systems with JACK)
```
lewri, line 503: @colinjmatt our Appimage seems to try to latch to JACK or ALSA in some circumstances (and notably throws warnings that have no impact). Any impact here?

CorsixTH/Lua/config_finder.lua
```diff
@@ -488,6 +491,26 @@ audio_music = nil -- [[X:\ThemeHospital\Music]]
...
+--   CoreMIDI   - Use the CoreMIDI API (MacOS only)
+--   WindowsMM  - Use the Windows MultiMedia API (Windows only)
+--
+-- Possible values for midi_port depend on the selected midi_api, and can
+-- be left nil to use the system default port. A list of available ports
+-- can be obtained from the midi settings screen in game.
+--
+midi_api = nil -- [[Native]]
```
lewri, line 511: Most of our config values do use the Windows value for their commented default. But I'm actually much more happy with native.

CorsixTH/Lua/dialogs/resizables/sound_setting.lua
```diff
@@ -0,0 +1,457 @@
...
+function UISoundSettings:reinitAudio()
+  local app = self.ui.app
+  app.audio:stopBackgroundTrack()
+  app.audio.has_bg_music = false
+  app.audio.not_loaded = not app.config.audio
+  app.audio.speech_file_name = nil
+  app.audio:init()
+  app:initLanguage()
+  app.audio:playRandomBackgroundTrack()
+end
```
lewri, line 242: There's no sanity check here for if `app.config.play_music == false`. Well there is, it's just buried deep into `Audio:playBackgroundTrack()`

Actually, you may not need line 241 as `Audio:init()` calls to play a background track already.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -225,9 +216,9 @@ function UIOptions:UIOptions(ui, mode)
   self:addBevelPanel(320, lower_row_y_pos, BTN_WIDTH, 30, col_bg):setLabel(_S.options_window.hotkey)
     :makeButton(0, 0, BTN_WIDTH, 30, nil, self.buttonHotkey):setTooltip(_S.tooltip.options_window.hotkey)
 
-  -- "Jukebox" button
-  self:addBevelPanel(465, lower_row_y_pos, BTN_WIDTH, 30, col_bg):setLabel(_S.options_window.jukebox)
-    :makeButton(0, 0, BTN_WIDTH, 30, nil, self.buttonJukebox):setTooltip(_S.tooltip.options_window.jukebox)
+  -- "Sound Options" button
+  self:addBevelPanel(465, lower_row_y_pos, BTN_WIDTH, 30, col_bg):setLabel(_S.options_window.sound)
+    :makeButton(0, 0, BTN_WIDTH, 30, nil, self.buttonSound):setTooltip(_S.tooltip.options_window.sound)
```
lewri, line 221: You may wish to hide this menu if the `WITH_AUDIO` compile option is off? Or disable it with a tooltip?

CorsixTH/Lua/config_finder.lua
```diff
@@ -488,6 +491,26 @@ audio_music = nil -- [[X:\ThemeHospital\Music]]
 --
 soundfont = nil -- [[X:\ThemeHospital\FluidR3.sf3]]
 
+-------------------------------------------------------------------------------
+-- Midi API and Device settings.
```
lewri, line 495: I would be tempted to say this now crosses territory into an advanced setting. It is probably best most players don't even try to touch it.

CorsixTH/Lua/config_finder.lua
```diff
@@ -488,6 +491,26 @@ audio_music = nil -- [[X:\ThemeHospital\Music]]
...
+-- By default, CorsixTH uses FluidSynth or build defined MIDI synthesizer.
```
lewri, line 496: Does this allow any better expansion into timidity support? Even though we're trying to move people away from that.

CorsixTH/Src/midi_player.h
TheCycoONE: clang-format doesn't give us freedom here, unless the wording is changed.

Edit: going to reformat this so it's not broken. clang-format did a bad thing here.

CorsixTH/Lua/dialogs/resizables/sound_setting.lua
TheCycoONE: Not anymore, that was broken.

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE: We removed WITH_AUDIO awhile ago.

CorsixTH/Lua/config_finder.lua
TheCycoONE: No change.

CorsixTH/Lua/dialogs/resizables/sound_setting.lua
TheCycoONE: I'd rather not guard on play_music here, we don't anywhere else we call play sounds, that is always in the audio code itself.

CorsixTH/Lua/config_finder.lua
TheCycoONE: Well... if someone runs timidity themselves then it will have a port and they can target it I suppose; so in that sense yes.

CorsixTH/Lua/config_finder.lua
TheCycoONE: There isn't an advanced settings section, except my new advanced midi option section.

There is a special settings, but that's for things not exposed in game so this doesn't qualify.

TheCycoONE: I suspect most players would be better off adjusting these in game; port is nearly impossible to get right otherwise, API is fairly simple though.

TheCycoONE:
I've noticed the tooltip for the invisible control pops in (e.g. soundfont when showing midi) - another window system fix to add.

Alberth289346 (approved):

Issue #3111, opened 2025-10-30
ui:playSound / audio:playSound -- Callback appears not persisted


lewri:
While trying to add a callback into a playSound function call, I noticed when you load the save, the callback is effectively destroyed and no longer called.

I tracked back to where the C++ seems to set the callback.
https://github.com/CorsixTH/CorsixTH/blob/b85d62011924528e34d542e500fe55e8720caf54/CorsixTH/Src/th_lua_sound.cpp#L245-L266

You can likely reproduce by adding a simple `--[[persistable:test]] function() print("callback") end` into the playsound and see if it prints out after loading a save with it present.

For the actual code I was trying to implement is below for #2113

```lua
function UIResearch:playBgSound()
  self.bg_sound = self.ui:playSound("Research.wav",
    --[[persistable:research_policy_window_reset_bg_sound]] function()
      self.playing = false
      self.bg_sound = nil
  end)
  self.playing = true
end
```

Inside the `UIResearch:onTick()` function
```lua
  if not self.playing then
    self:playBgSound()
  end
```

lewri:
This is most likely a `wontfix`/`cantfix` unless you can think of anything reasonable @TheCycoONE ?

We know the sound with the callback is only alive for the same game session and once it finishes playing it never comes back to a save; whether same or new session.

Alberth289346:
In the development chat (at 30 October), it appears that CPP does not track what sound(s) is (are) playing at any time. Control of sounds is achieved by exchanging the sound channel number between Lua and CPP. This works as long as the game is not saved and aborted, and continued at another point in time.

This could be fixed by adding more information to the CPP state data to improve the situation, but to get enough functionality to reload or restart sounds at the time of loading the saved game in CPP is non-trivial. Paths of sound files may have changed, sound hardware may be different etc. CPP is currently a minimal layer between Lua and the OS. It has no advanced functionality.

As an alternative, maybe it's better to store that additional information in Lua. By creating a central point in Lua that manages all sound, there is one spot in the Lua code that can be extended to restore sounds as much as possible they were at the time of the save. For example, maybe sounds can only be restarted instead of continuing at the point of playing at the time of the save.

Adding more state information in order to be able to restore a sound again in Lua is quite doable. Restoring that information on loading the game is currently trivially out of the box. The central point does need a special call that it must consider all in-flight information (such as allocated sound channels) as invalid, and attempt to rebuild the situation as it was at the time of the save as much as possible.

[...]

Issue #2113, opened 2022-03-07
[Bug] Sounds missing from Research Window


lewri:
#### Describe the issue
In Theme Hospital, the game would play the noise of the rat going round in its wheel. In CorsixTH this is missing.

#### Steps to Reproduce
1. Load a game with a research dept

2. Open research window.

#### Expected Behavior
The sounds should be playing

#### System Information
**CorsixTH Version:** 0.65.1

**Operating System:** Windows 10
**Theme Hospital Version:** GOG

#### Gamelog.txt

```
Paste gamelog.txt output here
```

### Additional Info
*Paste any screen shots or other additional information that might help
illustrate the problem.*

I did some tinkering first to try getting this working, but it isn't elegant enough because the sound will play its full duration even if the window is closed (e.g. the loop just starts again, you close the window, the sound now has to finish even though the window is gone)
*In `UIResearch:UIResearch(ui)`*
```lua
self.ui = ui
self.playing = false -- for background sound
```
*New function*
```lua
function UIResearch:playBgSound()
  self.ui:playSound("Research.wav", function() self.playing = false end)
  self.playing = true
end
```
*Added to `UIResearch:onTick()`*
```lua
if not self.playing then self:playBgSound() end
```

Potentially, to make this work better, we're missing the ability to stop sounds on the fly (like we can with music).

lewri:
Note: #3118 is now no longer a permanent fix for this. See #3111

PR #3118, opened 2025-11-07
Enable sound in research policy

lewri:
*Provides stopgap for #2113*

- Note this works around the documented bug of #3111 
- Adds the `Research.wav` sound to the research policy window which was in the original game.

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/research_policy.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/research_policy.lua
@@ -47,8 +47,11 @@ function UIResearch:UIResearch(ui)
   self.number_font  = gfx:loadFontAndSpriteTable("QData", "Font43V", false, palette)
   self.hospital = ui.hospital
   self.research = ui.hospital.research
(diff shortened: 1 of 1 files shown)


Alberth289346:
I have no doubt this will work currently. What I am worried about is the future. The check and start as a special case in the tick function is a bit of a sledge hammer approach. The shortest path to the desired functionality as a special case.

The trouble will come when sound requirements change in the future. Then this special case is going to cause a headaches.

I'd like to plea for a more general solution. Avoid special cases and very local decision making for something global.

In my view, the `ui` is likely the best owner for sound management. Make it responsible for managing sound of windows. In that way, you keep one spot where all "sound" is happening. That makes it simpler to adjust it in the future if the need arises.

Likely that central system needs information what it should do, so there must be some data about that somewhere, such that eg any window can have sound if so desired.

lewri:
I have now reclassified this as a stopgap solution, and does not close the original issue. Function also made private. #3111 should lead the way on a better long-term solution; and the code here should not be used as universally good practice.

PR #3119, opened 2025-11-07
Get up before we die

lewri:
- If a patient has fallen while about to die, they need to get up first (to fall down again!).
- Fixes the very specific case where the patient has to: fall over, die, and have the day change for this bug to happen.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -603,6 +603,15 @@ function Patient:tickDay()
   else
     self:setMood("sad2", "deactivate")
   end
+
+  -- Avoid an empty action queue error if the player clicks a patient to fall
+  -- at the same time as the day changes. There is now a between "neutral" stage.
+  if self.has_fallen == 3 then
+    self.has_fallen = 1
+  elseif self.has_fallen == 2 then
+    self.has_fallen = 3
+  end
+
   -- Now call the parent function - it checks
   -- if we're outside the hospital or on our way home.
   if not Humanoid.tickDay(self) then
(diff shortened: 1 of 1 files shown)


Alberth289346 (approved):

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -609,6 +609,14 @@ function Patient:tickDay()
...
   end
 
+  -- Avoid an empty action queue error if the player clicks a patient to fall
+  -- at the same time as the day changes. There is now a between "neutral" stage.
+  if self.has_fallen == 3 then
+    self.has_fallen = 1
+  elseif self.has_fallen == 2 then
+    self.has_fallen = 3
+  end
+
```
ARGAMX, line 614: This isn't about your PR:
But I noticed that these magic numbers 1, 2, and 3 don't make any sense to me.
Typical place where enum could improve readability.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -609,6 +609,14 @@ function Patient:tickDay()
     return
```
ARGAMX: This `return` is triggered if `self.going_home or self.going_to_die or self.dead`.

I wonder if the patient falls and is kicked out of the hospital while he's down.
And then have the day change.
Will he ever get up after that?
The same applies to when he gets `going_to_die = true`.
Will he get up?...

If not, then maybe we need to move this `has_fallen` block a little higher.
But it's best to double-check all of this.

lewri: Yes it can move up again

PR #3120, opened 2025-11-08
Set health to 0 on failed emergency patients

lewri:
*Implements https://github.com/CorsixTH/CorsixTH/pull/3113/files#r2491630152*

-  Make sure emergency patients who die get zero health, or they risk triggering additional activities in tickDay()

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1137,6 +1137,8 @@ function Hospital:resolveEmergency()
   for _, patient in ipairs(self.emergency_patients) do
     if patient and not patient.cured and not patient.dead
         and not patient.going_home then
+      -- Set the patient's health to 0 before we die
+      patient:changeAttribute("health", patient:getAttribute("health") * -1)
       patient:setToDying()
     end
   end


ARGAMX (dismissed):
👍

Alberth289346:
Euhm, why is that not done in `setToDying()` ?

Much further away:
I think that `tickDay()` is too powerful. A humanoid should perform mode switches by itself. Doing it from  "the outside" has much risk in getting out of sync with other considerations, because while looking at humanoid code, this now happens by black magic (Currently, nowhere inside the humanoid code you can find the conditions and the decision of this mode switch.)
New programmers will simply miss these external cases and mode switches, because they don't know it exists.

lewri:
> Euhm, why is that not done in `setToDying()` ?

By the time we reach this function, health is already 0 in normal cases. At least that function didn't have an assert for health not being 0.

There is no easy win without a larger refactor 🤷

Originally I suggested the resolve emergency just sets the health to 0 on patients and let the tick functions work out it after. But we're not there yet either.

Alberth289346:
> Originally I suggested the resolve emergency just sets the health to 0 on patients and let the tick functions work out it after. But we're not there yet either.

I wrote some more text in the Humbe project last Friday, and had the thought that perhaps humanoids should look around for their eg happiness etc on their own instead of at a central point.
If you then also add a poisson distribution, then all humanoids will look around them at different points in time, and you get a more consistent cycle time of the Lua code.  
The central point idea is good idea if you can make it more performant by combining computations for different humanoids, but I don't think that is happening much in `atEndOfDay` for humanoids.

lewri:
@Alberth289346 I can probably appease this a bit better by moving the code inside the for loop to Patient. But maybe as a new PR after this.

TheCycoONE:
I'm unconvinced this couldn't go in setToDying. What's the harm?

lewri:
> I'm unconvinced this couldn't go in setToDying. What's the harm?

Actually there's another hole because failing to cure a patient directly calls die instead of setToDying, which makes sense really because you don't want to delay it's visual effect compared to cure. The health is also not put to 0 there.

You could instead have every instance of changeattribute check if health hit 0 and then call setToDying (or was it goingToDie) if it reached 0.

TheCycoONE:
I'm still not understanding the opposition to making setToDying change the attribute to zero if it's not zero. Where is the harm?

PR #3125, opened 2025-11-10
Take SideObject out of the bin

lewri:
*Fixes #*

- Puts wrongly placed SideObject code from bin.lua to object.lua (NB: conversation happened about if it should be its own file, was concluded not necessary)
- Slight change of comments

### DIFF
--- a/CorsixTH/Lua/entities/object.lua
+++ b/CorsixTH/Lua/entities/object.lua
@@ -923,10 +923,10 @@ function SideObject:getDrawingLayer()
   else
     if self.direction == "east" then
       if self.object_type.thob == 50 then
-      --[[ Some side objects, such as the bin, have two orientations north and east
-        but they are displayed in the north and west part of the tile respectively.
-        This could lead to a graphical glitch where a side object in the west part
-        of the tile is displayed over a doctor in the middle of the tile. ]]
+        --[[ The bin has orientations north and east but they are displayed in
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/entities/object.lua
```diff
@@ -904,3 +904,35 @@ function Object:setState(state)
...
+  elseif self.direction == "west" then
+    return DrawingLayers.WestSideObject
+  else
+    if self.direction == "east" then
+      if self.object_type.thob == 50 then
+      --[[ Some side objects, such as the bin, have two orientations north and east
+        but they are displayed in the north and west part of the tile respectively.
+        This could lead to a graphical glitch where a side object in the west part
+        of the tile is displayed over a doctor in the middle of the tile. ]]
+        return DrawingLayers.WestSideObject
```
lewri, line 930: This case is strictly only the bin; not sure why they used the thob. So this comment was right before.

Alberth289346:
```suggestion
      if self.object_type.thob == 50 then
        --[[ The bin has orientations north and east but they are displayed in
        the north and west part of the tile respectively. This could lead to a
        graphical glitch where a side object in the west part of the tile is
        displayed over a doctor in the middle of the tile. ]]
        return DrawingLayers.WestSideObject
```
Since it's only the bin, make that explicit?

PR #3127, opened 2025-11-12
Don't show dialog "humanoid had empty queue"

### LINKED ISSUE #2653: [Bug] Moving room causes patient routing problems
### Describe the issue

If room had moved to another place when patient already moved from queue to state when he heading to enter the room, game will give error.

Screenshot 2024-10-02 at 21 18 49

### Steps to Reproduce

2. Move pharmacy to another place. There is a free space in the northeast. Or just delete pharmacy.
3. Unpause, set game speed 3.
4. Observe that patient in hat heading to old pharmacy place and then error on screen appears.

### Expected Behaviour

Considering that the patient then successfully returns to the reception and after that able to find Pharmacy on his own there is no need to show this dialog to user.

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

[humanoid_empty_action_queue.sav.zip](https://github.com/user-attachments/files/17234434/humanoid_empty_action_queue.sav.zip)

### CorsixTH Version

0.67.0

### Operating System

MacOS Sonoma 14.5

### Theme Hospital Version

CD

### Gamelog.txt

_No response_

### Additional Information

_No response_

ARGAMX:
*Fixes #2653*

- Wrap show dialog into `if TheApp.config.debug then`

### DIFF
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -742,9 +742,16 @@ function Humanoid:_handleEmptyActionQueue()
   end
 
   -- Log what just happened.
+  local message = "Warning: Empty action queue. For humanoid at coord x:"
+    .. self.tile_x .. " y:" .. self.tile_y .. ". Last action: " .. self.previous_action.name
+  if not TheApp.config.debug then
+    -- Notify the user about this in console even if debug mode is disabled.
+    -- Because gameLog() will be silent in console if debug mode is disabled.
+    print(message)
+    print("Automatic recovery has been attempted for empty queue.")
+  end
   self.world:gameLog("")
-  self.world:gameLog("Warning! Empty action queue! For humanoid at coord x:"
-    .. self.tile_x .. " y:" .. self.tile_y .. ". Last action: " .. self.previous_action.name)
+  self.world:gameLog(message)
   self.world:gameLog(self:tostring())
 
   -- First find out if this humanoid is in a room.


CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -752,38 +756,37 @@ function Humanoid:_handleEmptyActionQueue()
     self:queueAction(MeanderAction())
   elseif class.is(self, GrimReaper) then
```
lewri, line 766: How does the grim reaper even manage to get an empty action queue? Not that's for you to worry about lol.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -741,6 +741,10 @@ function Humanoid:_handleEmptyActionQueue()
     return
   end
 
+  -- Log what just happened.
+  self:dump()
```
lewri: Good addition. However, this just prints the clicked humanoid to console. Instead call `self.world:gameLog(self:tostring())` (dump could be adapted to go to gamelog with a switch but not necessary).

If we have debug on, this is also printed in the console anyway.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -752,38 +756,37 @@ function Humanoid:_handleEmptyActionQueue()
...
+        else
+          -- Set these variables to increase the likelihood of the humanoid managing to get out of the hospital.
+          self.going_home = false
+          self.hospital = self.world:getLocalPlayerHospital()
+          self:goHome("kicked")
+        end
+      end,
+      nil -- Do nothing on cancel
+    ))
+  end
```
lewri, line 800: else:
print("Warning: Humanoid encountered an empty action queue. Recovery has been attempted.")

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -752,38 +756,37 @@ function Humanoid:_handleEmptyActionQueue()
     self:queueAction(MeanderAction())
   elseif class.is(self, GrimReaper) then
     self:queueAction(IdleAction())
-  else
+  elseif class.is(self, Patient) then
```
lewri, line 768: Arguably better yes.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -752,38 +756,37 @@ function Humanoid:_handleEmptyActionQueue()
...
     self:queueAction(SeekReceptionAction())
+  else
+    self:queueAction(IdleAction())
```
lewri: This is likely the only case where something might not resolve itself. Do we want to note anything about it?

CorsixTH/Lua/entities/humanoid.lua
lewri: Actually, this check could be deleted; it's the same as the else block.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -752,38 +756,37 @@ function Humanoid:_handleEmptyActionQueue()
...
+    ui:addWindow(UIConfirmDialog(ui, true, _S.errors.dialog_empty_queue,
+      --[[persistable:humanoid_leave_hospital]] function()
+        self.world:gameLog("The humanoid was told to leave the hospital...")
+        if class.is(self, Staff) then
+          self:fire()
+        else
+          -- Set these variables to increase the likelihood of the humanoid managing to get out of the hospital.
+          self.going_home = false
+          self.hospital = self.world:getLocalPlayerHospital()
+          self:goHome("kicked")
```
lewri, line 793: Not you, but this looks like it might make some funky behaviour with non-patients. Not to resolve here though.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -741,6 +741,10 @@ function Humanoid:_handleEmptyActionQueue()
     return
   end
 
+  -- Log what just happened.
+  self:dump()
+  self.world:gameLog("Empty action queue! Last action: " .. self.previous_action.name)
```
lewri: There are downsides to removing any information/interruption to the player:
1. We don't get told of any issue that may have a pattern,
2. We don't get any broken saves at the time of the event.

Please preserve the existing log data that was removed:

```lua
  self.world:gameLog("")
  self.world:gameLog("Empty action queue!")
  self.world:gameLog("Last action: " .. self.previous_action.name)
  self.world:gameLog(debug.traceback())
```

Ideally, knowing their tile location would be useful if they did somehow get stuck. This helps if the game crashes again later we know where to look for that event. That can be done with `self.tile_x` and `self.tile_y`.

lewri (changes requested):
I'm cautious to suppress this to debug only.

However, please note some adjustments necessary:

1. We still should provide some real time feedback we hit an empty action queue at this time. I'm not comfortable completely isolating it from the player. Hence the print warning in the console (though anyone outside of Windows doesn't usually see it).
2. More information needs pushing to the gamelog with this change; if we end up crashing later as a result.

The change to make the patient seek reception is good. However if a patient died, got an empty action queue, and was sent to reception; they would inevitably break again.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: 👌

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: 👌

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > This is likely the only case where something might not resolve itself. Do we want to note anything about it?

This originally was `SeekReceptionAction()`. 
I changed it to `IdleAction()`.
But now I'm starting to doubt it.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > Ideally, knowing their tile location would be useful if they did somehow get stuck. This helps if the game crashes again later we know where to look for that event. That can be done with `self.tile_x` and `self.tile_y`.

I have added `self.tile_x` and `self.tile_y`. 
I agree that this would be useful.

ARGAMX: > Please preserve the existing log data that was removed:

The rest of the logging remains essentially the same.
Except the `debug.traceback()`.

The problem with `debug.traceback()` is that it's useless for an empty queue case, and every time it shows the same trace — that the tick attempted to execute a command from the queue, but the queue is empty.
It's completely useless and that trace obvious.
You won't know if the patient was dying or was treated in the room from trace. No.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > Not you, but this looks like it might make some funky behaviour with non-patients. Not to resolve here though.

I agree, it shouldn't have been done that way.
I have wrapped it into `if class.is(self, Patient) then` check.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: I changed it back and now it's not like that. So there's no need to delete it now.

CorsixTH/Lua/entities/humanoid.lua
lewri: Ok fine, I agree with that.

CorsixTH/Lua/entities/humanoid.lua
lewri: You have seek reception called for both patients and any other entity now? You arguably no longer would need to filter for patients. But you could add an in-line comment like as `else -- Patients etc`?

CorsixTH/Lua/entities/humanoid.lua
lewri: Gamelog gives no player feedback in the console unless debug is on. Did you want to stick with that or use a print (which doesn't go in gamelog). (Or do both if you like)

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -752,38 +758,41 @@ function Humanoid:_handleEmptyActionQueue()
...
-      else
-        -- Set these variables to increase the likelihood of the humanoid managing to get out of the hospital.
-        self.going_home = false
-        self.hospital = self.world:getLocalPlayerHospital()
-        self:goHome("kicked")
-      end
-    end,
-    nil -- Do nothing on cancel
-  ))
+  self.world:dumpGameLog()
```
lewri, line 802: So if we dump, we'll create a gamelog file that may not relate to the time of a full on game crash. You likely don't want to dump if we're removing any player interaction with this patch.

It makes sense to hold fire until we crash for real. If we recovered successfully nobody needs to know 😉
image

CorsixTH/Lua/entities/humanoid.lua
lewri: Can you check if VIPs and Inspectors still get a going home process, just before we finalise locking this off if you are changing it.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > Gamelog gives no player feedback in the console unless debug is on. Did you want to stick with that or use a print (which doesn't go in gamelog). (Or do both if you like)

I don't think it true.
`if TheApp.config.debug` don't cover the `self.world:gameLog` calls.
Thats why I moved `self.world:gameLog` calls outside of the block covered by `if TheApp.config.debug`.

If you didn't see any output in your console before this PR, then I don't know how to help, as I've never had and still don't have this console.
What's displayed there is a mystery to me.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > So if we dump, we'll create a gamelog file that may not relate to the time of a full on game crash. You likely don't want to dump if we're removing any player interaction with this patch.
> 
> It makes sense to hold fire until we crash for real. If we recovered successfully nobody needs to know 😉

Firstly. 😎
You can run game v0.69.1 and instantly close it. 
Then open Logs folder and you will see new log file.
This new gamelog file is not relate to the time of a full on game crash. 
How do you live with that? 😁
Looks like `dumpGameLog()` was already called.
Next  `dumpGameLog()` call will not create any new file. It already exist.
So looks like logfiles time was never corresponded to the time of the crash. 😛

Secondly. 😎
In the #2653 start you asked to show logs of this event, apparently expecting to see something in log.
The logs was empty due to missing that  `dumpGameLog()` call in that scenario.
So I had nothing to show. My logs was empty.
This `dumpGameLog()` call should prevent this situation from happening.
And that dialogue In the #2653 simply would not have existed if that had been implemented earlier.
Don't you fell this proposal looks like a contradiction? 😛

So, I don't understand why you are afraid to dump here.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > Can you check if VIPs and Inspectors still get a going home process, just before we finalise locking this off if you are changing it.

If VIPs and Inspectors can get here, we would see this window, but I never saw it for them.
Usually only for patients.

CorsixTH/Lua/entities/humanoid.lua
lewri: Print() will output to the console, if one is present
Gamelog will log into the table of things to write on dump.

However, if you call gamelog with debug on, every thing ever added to the table is also printed to the console.

Code for gamelog addition:
```lua
--! Append a message to the game log.
--!param message (string) The message to add.
function World:gameLog(message)
  self.game_log[#self.game_log + 1] = message
  -- If in debug mode also show it in the command prompt
  if TheApp.config.debug then
    print(message)
  end
end
```

CorsixTH/Lua/entities/humanoid.lua
lewri: Every time you call dumpGamelog it appends everything in the gamelog's table to its file. My bad about the init behaviour.

Worst case, you get duplication:

```
System information
[First dump call]
Error 1 : foo
Warning 1: Recovered
Warning 2: blahblah
... [Second dump call]
Error 1 : foo
Warning 1: Recovered
Warning 2: blahblah
Error 2: Bar
```

lewri: Wait no... we use `w` mode; so it overwrites the gamelog. I thought we used `a` append more.

Please ignore me 🙃

CorsixTH/Lua/entities/humanoid.lua
lewri: The print is only really useful for Windows and some Linux users.... Mac probably has no care for the terminal unless you go out your way to launch the game with one.

ARGAMX: > Print() will output to the console, if one is present Gamelog will log into the table of things to write on dump.
> 
> However, if you call gamelog with debug on, every thing ever added to the table is also printed to the console.
> 
> Code for gamelog addition:
> 
> ```lua
> --! Append a message to the game log.
> --!param message (string) The message to add.
> function World:gameLog(message)
>   self.game_log[#self.game_log + 1] = message
>   -- If in debug mode also show it in the command prompt
>   if TheApp.config.debug then
>     print(message)
>   end
> end
> ```

Regarding the `print()`, now I know more, thank you.

But I still don't understand why a user with debugging disabled would need this information in the console?
What should they do with it?
How can it help them?

This is real debugging information.
Meaningless for the average user.

In 0.68.0 - 0.69.1, I never had a situation where:
1. The situation would not have resolved itself without kick the patient or the doctor. And I would have found it useful to kick a patient through this window.
2. The empty queue information would have given me some meaningful information that would have helped me understand something important.

So far, ALL empty queue cases I've had, and those Mickman empty queue cases had, have been resolved automatically and didn't require any intervention.

Mickman had debug mode enabled, so he saw these messages in the console.
And what did this give to him?
He asked about these messages in chat, because it was an incomprehensible message to him.
In the end, he didn't have to do anything about it, and it resolved itself.
And I understand him here.

These messages are not needed by those who have debugging disabled.
All my cases and Mickman cases were not useful.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > You have seek reception called for both patients and any other entity now? You arguably no longer would need to filter for patients. But you could add an in-line comment like as `else -- Patients etc`?

Yes, that code no longer would need for Interpreter, as `Patient` can be the part of `else`.
But I honestly think that the fewer real sets are listed in a switch, the more difficult it is to maintain that switch.
Especially in weakly typed and weakly constrained languages ​​like Lua.
Explicity is better here I think.

So the patients are added here exclusively for others reading this code, not for the interpreter.
A good example of the opposite is the function `function _getHumanoidQueuePriority(queue, humanoid)` in `queue.lua`, where until I refactored a bit and added a comment, it was not at all obvious who was getting the priority 5.
I don't understand this style.
I believe the code should be written for people, not for machines.

Regarding this section of code I still don't have a clear understanding of who will be in the 'else' here in `humanoid.lua`. VIPs? It's not good that readers like me don't have a clear understanding here.
But at least now it is clear that patients do not belong that `else`.
So, I wouldn't want to remove this, even though it's functional and unnecessary.

CorsixTH/Lua/entities/humanoid.lua
lewri: So as I said earlier, I'm not comfortable cutting off feedback to the user outright. Hence, use a print to the console.
Whether you want it as just a print or a print followed by a gamelog is up to you.

ARGAMX: > Whether you want it as just a print or a print followed by a gamelog is up to you.

Okay, I've added `print()` output for all users.

CorsixTH/Lua/entities/humanoid.lua
lewri: Sure. It still looks weird but I'm not going to block it.

PR #3134, opened 2025-11-20
Get rid of `object.humanoid_class == className`

ARGAMX:

We have a lot of calls in our code like `object.humanoid_class == "Handyman"`

I found 69 matches in 25 files (!).
(Do you also smell a lot of git conflicts?) 😁

Some objects directly read the fields of another object and also externally compare this field value.

It was a bit bothering me as I think it's a typical violation of encapsulation for me.

Logically, this should be a call to a method of this object, to which the object for comparison is passed.
And the object itself decides the results of this comparison.

Filed `humanoid_class` should not be exposed outside and should not participate in external operations.

So I propose to rework this to call the appropriate method.

`object.humanoid_class == "Handyman"`
->
`object:isType("Handyman")`

----

Overall, it a minor issue.

I was more concerned that this strange approach was being copied and pasted everywhere and was growing. So reading and externally comparing the `humanoid_class` field seemed like a common practice.

I was working on another task, and it required me to copy this comparison to other places one more.
But I realised I didn't want to copy that and decided to tidy things up a bit.

Oh yeah, and on top of that, it'll possible break something somewhere. How could it be otherwise? 😂

But I went and lit a candle in the church to prevent this from happening. 🤞🏻 🕯️

### DIFF
--- a/CorsixTH/Lua/rooms/operating_theatre.lua
+++ b/CorsixTH/Lua/rooms/operating_theatre.lua
@@ -368,7 +368,7 @@ function OperatingTheatreRoom:canHumanoidEnter(humanoid)
   if can and class.is(humanoid, Patient) then
     -- Patients can only enter once all doctors are in surgeon clothes
     for staff, is_ready in pairs(self.staff_member_set) do
-      if class.is(staff, Doctor) or is_ready ~= "ready" then
+      if not staff:isType("Surgeon") or is_ready ~= "ready" then
         return false
       end
     end


CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -815,6 +815,10 @@ function Humanoid:setType(humanoid_class)
   self.th:setPartialFlag(self.permanent_flags or 0)
 end
 
+function Humanoid:isType(humanoid_class)
+  return self.humanoid_class == humanoid_class
+end
+
```
lewri, line 821: We don't need this; as we already have code made to support a check in a better way: `class.is(humanoid, ClassName)`

https://github.com/CorsixTH/CorsixTH/blob/39d721148dc1ea71e06e12131ae50a0876734fc3/CorsixTH/Lua/class.lua#L127-L141

lewri: Or you could make this function call `class.is` instead too.

Alberth289346:
> I was more concerned that this strange approach was being copied and pasted everywhere and was growing. So reading and externally comparing the humanoid_class field seemed like a common practice.

Lots of newbie programmers hacked the code. They tend to copy/paste already existing code blocks.
Not that you can blame them. The Lua programming model of global read and write access is essentially broken for our scale of code.

Alberth289346:
Of course, much of the code after class checks should probably be moved the proper derived class.
The Humanoid base class is generic both as base class and as collection of code pieces that affect one or more specific derived classes.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > We don't need this; as we already have code made to support a check in a better way: `class.is(humanoid, ClassName)`

Yes, perhaps that would be even better.
I changed `isType()` to `class.is()` almost everywhere.
But there are several `humanoid_classes` that don't have their own class.
For example, "Chewbacca Patient" or "Slack Female Patient."
So for those cases, I had to use `isType()`.

CorsixTH/Lua/dialogs/staff_dialog.lua
```diff
@@ -394,7 +394,7 @@ function UIStaff:openStaffManagement()
 end
 
 function UIStaff:hitTest(x, y)
-  return Window.hitTest(self, x, y) or is_in_view_circle(x, y, self.staff.profile.humanoid_class == "Handyman")
+  return Window.hitTest(self, x, y) or is_in_view_circle(x, y, self.staff.profile:isType("Handyman"))
```
Alberth289346: `self.staff.profile:isType("Handyman")`

This implies `class.is(self.staff, Handyman)` right?

EDIT: Fixed call.

CorsixTH/Lua/dialogs/staff_rise.lua
```diff
@@ -81,7 +81,7 @@ function UIStaffRise:UIStaffRise(ui, staff, rise_amount)
   self:makeTooltip(_S.tooltip.staff_window.salary, 14, 171, 168, 193)
   self:makeTooltip(_S.tooltip.staff_window.ability, 12, 213, 89, 243)
 
-  if profile.humanoid_class == "Doctor" then
+  if profile:isType("Doctor") then
```
Alberth289346: `profile` is derived from the `staff` parameter, so `class.is(staff, Doctor)` (or `self.staff` would work too).

ARGAMX: > `profile` is derived from the `staff` parameter, so `class.is(staff, Doctor)` (or `self.staff` would work too).

Yeah, right, thanks. 👍
I refactored that too.

CorsixTH/Lua/dialogs/staff_dialog.lua
ARGAMX: > `self.staff.profile:isType("Handyman")`
> 
> This implies `class.is(self.staff, Handyman)` right?
> 
> EDIT: Fixed call.

Seems so.
I changed it.
Seems to be working fine. ✌️

CorsixTH/Lua/rooms/training.lua
```diff
@@ -200,7 +200,7 @@ function TrainingRoom:commandEnteringStaff(humanoid)
   local obj, ox, oy
   local profile = humanoid.profile
 
-  if profile.humanoid_class == "Doctor" then
+  if class.is(humanoid, Doctor) then
```
ARGAMX, line 203: this is not a mistake (profile -> humanoid)

CorsixTH/Lua/dialogs/staff_dialog.lua
```diff
@@ -67,7 +67,7 @@ function UIStaff:UIStaff(ui, staff)
   self.ui = ui
   self.modal_class = "humanoid_info"
   self.width = 220
-  if profile.humanoid_class == "Handyman" then
+  if class.is(staff, Handyman) then
```
ARGAMX, line 70: this is not a mistake (profile -> staff)

CorsixTH/Lua/dialogs/staff_dialog.lua
```diff
@@ -86,7 +86,7 @@ function UIStaff:UIStaff(ui, staff)
   self:addPanel(301,   15, 114) -- Skills/Abilities
   self:addColourPanel(35, 51, 71, 81, 208, 252, 252):makeButton(0, 0, 71, 81, nil, self.openStaffManagement):setTooltip(_S.tooltip.staff_window.face) -- Portrait background
 
-  if profile.humanoid_class == "Handyman" then
+  if class.is(staff, Handyman) then
```
ARGAMX, line 89: this is not a mistake (profile -> staff)

CorsixTH/Lua/dialogs/staff_dialog.lua
```diff
@@ -105,7 +105,7 @@ function UIStaff:UIStaff(ui, staff)
     self:addPanel(302,   5, 178) -- View circle top/Wage
     self:addPanel(303,   0, 226) -- View circle midpiece
     self:addPanel(304,   6, 274) -- View circle bottom
-    if profile.humanoid_class ~= "Doctor" then
+    if not class.is(staff, Doctor) then
```
ARGAMX, line 108: this is not a mistake (profile -> staff)

CorsixTH/Lua/entities/humanoid.lua
lewri: Worrying... sounds like those should be a 'subclass'.

<del>I wonder if you could just isolate this kind of function to Patient method.</del> ignore that

CorsixTH/Lua/staff_profile.lua
```diff
@@ -273,3 +273,7 @@ function StaffProfile.translateStaffClass(staff_class)
   }
   return staffclass_to_string[staff_class]
 end
+
+function StaffProfile:isType(humanoid_class)
+  return self.humanoid_class == humanoid_class
+end
```
lewri, line 279: Is this used in the file? 
Oh... it's used in place_staff. Ew. This could do with a note about why you have this here separate from Humanoid:isType.

mugmuggy:
Probably worth noting that if you aren't using LuaJit, there is significant overhead, in converting some of these accesses from the lookup of what is essentially locally scoped variables to the global lookup/access through the class.is method. Most of the updates are around UI methods so probably haven't been touched since the move to separate all the classes out and being UI related, probably not a big deal, but anything that forms part of the main loop, should be wary of for older or lower clcok speed processors.

ARGAMX:
> but anything that forms part of the main loop, should be wary of for older or lower clcok speed processors.

For the main loop, calling `isType()` might be preferable to `class.is()` for this reason?

lewri:
Likely you're safe with calls that in functions that only fire occasionally. Ones that perform part of tick/tickDay/endMonth loops are less so.

We use to do some odd `local Foo = Foo` stuff in really old code. Likely less desired now (but think only worked for standard lua tools)

CorsixTH/Lua/staff_profile.lua
ARGAMX: > Is this used in the file? Oh... it's used in place_staff. Ew. This could do with a note about why you have this here separate from Humanoid:isType.

I think this comment is useful within the context of this PR and review.
Because in the context of this PR, it's a valid question to ask ("why there are two copies of this function in different files?").

But if the reader isn't aware of this PR and that the exact same function exists in Humanoid, I think this comment will be a strange one.

Some strange and useless comment.

CorsixTH/Lua/dialogs/staff_dialog.lua
lewri: As an example to comments earlier. No harm in this window staying as class.is; since it only fires on window creation.

Alberth289346:
> We use to do some odd local Foo = Foo stuff in really old code. Likely less desired now (but think only worked for standard lua tools)

It avoids a useless search in the local scope (before trying the global scope), so it works for everything global. The difference however is quite small.

The local class variables does the same thing (`local Foo = _G["Foo"]`) I am far from sure it makes a difference since you hardly ever use the class name.

CorsixTH/Lua/rooms/operating_theatre.lua
```diff
@@ -368,7 +368,7 @@ function OperatingTheatreRoom:canHumanoidEnter(humanoid)
   if can and class.is(humanoid, Patient) then
     -- Patients can only enter once all doctors are in surgeon clothes
     for staff, is_ready in pairs(self.staff_member_set) do
-      if class.is(staff, Doctor) or is_ready ~= "ready" then
+      if not staff:isType("Surgeon") or is_ready ~= "ready" then
         return false
       end
```
Alberth289346, line 373: You probably considered it already, but does `staff_member_set` includes handy men that visit the plants?

ARGAMX: > You probably considered it already, but does `staff_member_set` includes handy men that visit the plants?

Nope, handymen never gets into `room.staff_member_set`.
`room.staff_member_set` is only serving patients staff.
Handymen goes to `room.humanoids` array.
So, in this regard, everything is fine here.

CorsixTH/Lua/entities/humanoids/staff.lua
```diff
@@ -43,7 +43,7 @@ function Staff:tickDay()
   if not Humanoid.tickDay(self) then
     return false
   end
-  if self.humanoid_class == "Receptionist" then return true end
+  if class.is(self, Receptionist) then return true end
```
lewri, line 46: tickDay might be too frequent for class.is going by mugmuggy's comments but also most use cases of CorsixTH shouldn't be running on a potato.

If someone wants to port it to potato hardware and we fine these are bottlenecks that contributor can always give a patch to change them (or suggest they use LuaJIT).

lewri:
Additional reviewer tags are just in case cyco has any view on it; and Toby has done some work with LuaJIT before. Not a blocker to merge if nothing comes back in the next day imo.

CorsixTH/Lua/entities/humanoids/staff.lua
ARGAMX: > tickDay might be too frequent for class.is going by mugmuggy's comments but also most use cases of CorsixTH shouldn't be running on a potato.
> 
> If someone wants to port it to potato hardware and we fine these are bottlenecks that contributor can always give a patch to change them (or suggest they use LuaJIT).

I conducted an experiment and didn't find any noticeable drop in performance.

PR #3143, opened 2025-11-23
Fix the aspect ratio for some FMVs

TheCycoONE:
On old DOS games sometimes 320x200 was used but it was intended to be stretched to fit a 4:3 monitor. The intro movie in Theme Hospital is an example of this.


CorsixTH/Lua/movie_player.lua
```diff
@@ -45,6 +46,10 @@ local calculateSize = function(me)
   local screen_w, screen_h = me.app.config.width, me.app.config.height
   local native_w = me.moviePlayer:getNativeWidth()
   local native_h = me.moviePlayer:getNativeHeight()
+  if native_w == 320 and native_h == 200 then
```
lewri, line 49: I would've preferred detection of the movie itself to be played but this works too.

lewri (approved):

PR #3144, opened 2025-11-23
small Windows installer related improvements

goodusername123:
- Optimizes the file size (in a fully lossless way) of the Windows icon used by both the Windows installer and the main Windows game executable.
- Changes the compression used in the Windows installer from non-solid zlib/Deflate to non-solid LZMA netting a file size decrease of around 8MB in my rough testing. NOTE: this increases the time it takes to build/compile the installer however I assume this is fine.
- Sets the Windows installer to be DPI scaled for crisp text and bitmap. NOTE: after this change the window(s) for installer no longer changes size when moved to a different monitor, If this is unacceptable then I do know of an alternative method for DPI scaling but it comes at the cost of splash bitmap no longer being crisp (like before), The text only being half as crisp/sharp, and it only works on Win10+ (falls back to something else on older versions of Windows).

### before DPI awareness:
Capture before

### after DPI awareness:
Capture after

### alternative DPI scaling:
Capture after alt

There are some extra random details in the commit messages. If desired I can attach said extra words/details to this PR's description.

### DIFF
--- a/WindowsInstaller/Win32Script.nsi
+++ b/WindowsInstaller/Win32Script.nsi
@@ -53,6 +53,10 @@ InstallDir "$PROGRAMFILES\${PRODUCT_NAME}"
 ; This is needed to be able to install into the program files directory
 RequestExecutionLevel admin
 
+; Set installer compression
+SetCompressor lzma
+SetCompressorDictSize 16
+
 ; -------------------- Definitions and macros that create the graphical interface -----------------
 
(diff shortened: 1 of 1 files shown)


lewri:
I wouldn't imagine we care about changes between monitors. It's very unlikely someone is going to do such a thing anyway during install.

TheCycoONE:
Thank you!

PR #3146, opened 2025-11-25
Sound: add isPlaying, handles, and loops

### LINKED ISSUE #2113: [Bug] Sounds missing from Research Window

TheCycoONE:
Rather than directly using channels to reference sound effects use handles. This way sounds are less likely to coincidentally have the same ID which was not suitable for an isPlaying method.

Taking advantage of the new handles, add an isPlaying method.

Convert the research dialog to use this method for checking isPlaying.

In the process I discovered a race condition with the reserved channels which this PR also addresses.

Also adds a loop parameter that can be used for repeated or continuous sounds.

Fixes #2113

### DIFF
--- a/CorsixTH/Lua/audio.lua
+++ b/CorsixTH/Lua/audio.lua
@@ -314,7 +314,7 @@ local wilcard_cache = permanent "audio_wildcard_cache" {}
 --  finished playing. Can be nil.
 --!param played_callback_delay (integer) An optional delay in milliseconds
 --  before the played_callback is called.
---!param loops (integer) number of times to play the audio
+--!param loops (integer) number of times to play the audio. -1 for infinite.
 --!return (table) A `sound` table for passing into functions that act on the
 --  playing sound. The fields are an implementation detail that should not be
 --  used outside of the Audio class.
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -225,7 +225,17 @@ function UI:setupGlobalKeyHandlers()
   self:addOrRemoveDebugModeKeyHandlers()
 end
 
--- Used for everything except music and announcements
+--! Play a sound effect
+--!param name (string) The name of the sound to be played. Can include
+--  wildcards (*).
+--!param played_callback (function) A function to be called when the sound has
+--  finished playing. Can be nil.
+--!param played_callback_delay (integer) An optional delay in milliseconds
+--  before the played_callback is called.
+--!param loops (integer) number of times to play the audio. -1 for infinite.
+--!return (table) A `sound` table for passing into functions that act on the
+--  playing sound. The fields are an implementation detail that should not be
+--  used outside of the Audio class.
 function UI:playSound(name, played_callback, played_callback_delay, loops)
   if self.app.config.play_sounds then
     return self.app.audio:playSound(name, nil, false, played_callback, played_callback_delay, loops)


lewri:
Is this enough to close the issue or would we need more? If it's the former, it may want to live in Window instead. I have not yet reviewed though.

TheCycoONE:
I don't understand the question?

Alberth289346 (approved):

lewri:
> I don't understand the question?

The question was if this closes #2113 or not is all. Or if you'd still like a better permanent solution in the future.

CorsixTH/Src/th_sound.cpp
```diff
@@ -323,25 +331,37 @@ void sound_player::set_sound_effects_enabled(bool bOn) {
...
-  for (; (available_channels_bitmap & (1 << iChannel)) == 0; ++iChannel) {
+  std::unique_lock lock(channel_mutex);
+  for (size_t i = 0; i < channels.size(); ++i) {
+    if (channels[i] == null_handle) {
+      // Don't assign a null handle to a playing track
+      if (next_playing_track_handle == null_handle) {
+        next_playing_track_handle++;
+      }
+      channels[i] = next_playing_track_handle++;
+      return i;
```
lewri: This will permanently warn on every future pr for reference. Can you suppress it?

TheCycoONE: I'll add a cast before merging yes

TheCycoONE:
Yes this can close that I think. I had the idea after that an even better solution is to add a loop option so we don't need to do anything on tick in lua. But there is always room to refactor.

CorsixTH/Lua/dialogs/fullscreen/research_policy.lua
```diff
@@ -51,7 +51,6 @@ function UIResearch:UIResearch(ui)
   -- lewri: This is a stopgap solution to work around sound states not being
   -- implemented. It should be replaced as soon as it is possible to do so.
```
lewri: These could now be removed

CorsixTH/Lua/dialogs/fullscreen/research_policy.lua
```diff
@@ -167,17 +166,12 @@ end
 --! Construct and play this window's background sound.
 function UIResearch:_playBgSound()
   -- Note: See UIResearch:UIResearch function for improving this.
-  self.bg_sound = self.ui:playSound("Research.wav",
-    --[[persistable:research_policy_window_reset_bg_sound]] function()
```
lewri, line 171: Risking a failed persistable I think? Though arguably the gap between my pr and your pr is so short it should be acceptable.

CorsixTH/Lua/audio.lua
```diff
@@ -374,11 +374,25 @@ function Audio:stopSound(sound)
...
-  if sound.channel and sound.channel >= 0 then
-    sound_fx:stop(sound.channel, sound.played_callback_id)
+  if sound and sound.handle then
+    sound_fx:stop(sound.handle, sound.played_callback_id)
   end
 end
 
+--! Determine if a given sound is still playing
+--!param sound (table) The `sound` table returned by `Audio:playSound`.
+function Audio:isPlaying(sound)
```
lewri, line 385: Could this also have a pass through function in UI like the others?

CorsixTH/Lua/dialogs/fullscreen/research_policy.lua
```diff
@@ -167,17 +166,12 @@ end
 --! Construct and play this window's background sound.
 function UIResearch:_playBgSound()
   -- Note: See UIResearch:UIResearch function for improving this.
```
lewri: Also not needed anymore

CorsixTH/Lua/dialogs/fullscreen/research_policy.lua
```diff
@@ -167,17 +166,12 @@ end
 --! Construct and play this window's background sound.
 function UIResearch:_playBgSound()
```
lewri, line 168: From the previous discussed PR; a more robust solution might be better these functions living in the Window class instead. But I'd understand if that would be too consuming to move now.

lewri (commented):
This review submitted under the guise that this will close the issue 2113. If it won't, I can make a different review.

Works fine for me too.

TheCycoONE:
Thanks Lewri, I plan to fix everything except moving to window. It doesn't make sense to me if this is the only window that does it. If there are other windows I'd reconsider.

A little surprised how coherent it turned out given it was one rather late evening of coding.

CorsixTH/Lua/dialogs/fullscreen/research_policy.lua
TheCycoONE: I don't want to move this unless there are more use cases for it. Otherwise it's bloating window just for research room.

CorsixTH/Lua/audio.lua
TheCycoONE: No longer used, so not for now.

CorsixTH/Lua/dialogs/fullscreen/research_policy.lua
TheCycoONE: Fixed.

CorsixTH/Lua/dialogs/fullscreen/research_policy.lua
TheCycoONE: Removed.

TheCycoONE:
Uploaded a new version that I'm happy with. It uses loops in SDL_Mixer instead of doing anything with callbacks or playback checks on tick which is cleaner and more efficient.

I expect the two commits to be merged after review.

CorsixTH/Lua/audio.lua
```diff
@@ -314,10 +314,11 @@ local wilcard_cache = permanent "audio_wildcard_cache" {}
 --  finished playing. Can be nil.
 --!param played_callback_delay (integer) An optional delay in milliseconds
 --  before the played_callback is called.
+--!param loops (integer) number of times to play the audio
```
lewri: Suggest: -1 is continuous/infinite (or some comment thereabouts)

PR #3149, opened 2025-11-28
Modernize use scoped lock

TheCycoONE:
scoped_lock replaces lock_guard in c++ 17, as equally light for a single mutex but also capable of locking multiple mutexes at once with a deadlock avoidance stategy.

### DIFF
--- a/CorsixTH/Src/th_movie.cpp
+++ b/CorsixTH/Src/th_movie.cpp
@@ -106,7 +106,7 @@ movie_picture_buffer::~movie_picture_buffer() {
 
 void movie_picture_buffer::abort() {
   aborting = true;
-  std::lock_guard<std::mutex> lock(mutex);
+  std::scoped_lock lock(mutex);
   cond.notify_all();
 }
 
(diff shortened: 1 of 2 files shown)


CorsixTH/Src/th_movie.cpp
```diff
@@ -132,20 +132,20 @@ void movie_picture_buffer::allocate(SDL_Renderer* pRenderer, int iWidth,
...
-    std::lock_guard<std::mutex> lock(mutex);
+    std::scoped_lock lock(mutex);
     allocated = false;
   }
 
   for (movie_picture& picture : picture_queue) {
-    std::lock_guard<std::mutex> pictureLock(picture.mutex);
+    std::scoped_lock lock(mutex);
     picture.deallocate();
   }
```
Alberth289346, line 150: I am guessing this is intended?

It uses a different mutex, but as it seems to deallocate the object, I can see that can be tricky.

Alberth289346: Line 180 still uses the picture lock. Should not be the same lock?

Alberth289346 (approved):
The introduced discrepancy between lines 149 and 180 is unclear to me. If this is intended, the change looks fine.

CorsixTH/Src/th_movie.cpp
TheCycoONE: Nope, this was a mistake, I deleted the whole line and didn't remember what I put there.

PR #3159, opened 2025-12-03
UI scale

### LINKED ISSUE #1210: Integer-based UI scaling

TheCycoONE:
*Fixes #1210*

* Adds the option to scale to 200% or 300% for UI elements.
* Fixes the position of the English tooltip for the language dropdown entries
* Fixes the internal size of the confirm dialogs to match the visual size.
* Draws true type fonts larger rather than scaling for improved appearance.
* Nearest neighbour scaling for bitmap fonts to preserve readability.
* Dynamically scalable (all dialogs)
* Improves the appearance of bitmap text on the lose movie headline

What it looks like:

<details>
  <summary>Before. Default 100% scale. Press here 👈</summary>
  
Screenshot 2026-03-19 at 11 39 56
  
</details>

<details>
  <summary>After. 200% scale. Press here 👈</summary>
  
  Screenshot 2026-03-19 at 11 29 13
  
</details>


CorsixTH/Src/th_lua_gfx.cpp
```diff
@@ -43,6 +43,35 @@ SOFTWARE.
 #include FT_TYPES_H
 
 namespace {
+uint32_t flags_from_table(lua_State* L, int index) {
```
Alberth289346: Are you aware that the draw flags also exist as named constants in Lua?

```
Lua/utility.lua
217:DrawFlags = {}
218:DrawFlags.FlipHorizontal  = 2^0
219:DrawFlags.FlipVertical    = 2^1
220:DrawFlags.Alpha50         = 2^2
221:DrawFlags.Alpha75         = 2^3
222:DrawFlags.AltPalette      = 2^4
223:DrawFlags.EarlyList       = 2^10
224:DrawFlags.ListBottom      = 2^11
225:DrawFlags.BoundBoxHitTest = 2^12
226:DrawFlags.Crop            = 2^13
```

`utility` also has a number of bit operation functions (`flag_*` functions around lines 190-210, and two more around lines 295-310)

TheCycoONE: I didn't, I'll switch to using those and add the missing flags.

TheCycoONE:
Alright, I think that's everything. Though I need to go back through all the dialogs and test all the buttons to make sure nothing I did later broke anything earlier.

TheCycoONE:
I'm just going to squash merge this one; too much effort to tidy.

TheCycoONE:
There is some naming convention inconsistency that should be fixed before merging. Feel free to point it out but if you do you also have to tell me the right style 😛 

There should also be some documentation on ui vs screen pixels and when which is used. Mainly draw calls and mouse events use screen pixels; widget placement uses ui pixels.

lewri:
Could we push this to the beta tester channel at all as well for pre merge testing too? They'll need an accessible win64 portable build probably linked direct, Linux users should be fine themselves.

Alberth289346:
It's too big indeed. I looked at the changes and didn't see anything weird.
I am fine with the patch.

A test is likely useful as Lewri proposed.

Alberth289346 (approved):

tobylane (approved):
Minor points for the future

- Handle unexpected scale numbers
- Provide a way out of menu dialogs when the cancel button is out of view, eg esc.
- A cancel button/timeout on changing the scale factor? Maybe not needed, as the settings button and its dialog are in the centre of the screen.

It feels so authentic to have a larger ui!

TheCycoONE:
@tobylane 

I spent some time thinking on that too, there should be a different maximum scale factor depending on the resolution; or when shrinking the window by dragging the border, a different minimum size depending on the scale factor.

Not as sure about what to do when people edit the config file directly - maybe that's on them.

Will look at this in a follow up PR like you suggested.

Alberth289346:
> Not as sure about what to do when people edit the config file directly - maybe that's on them.

Clamp to the max that completely fits their screen, imho.
I doubt anyone would want to go beyond that. If someone does, let them file a ticket to explain why.

----
I pondered about the large jump from 1 to 2.

I think to fix, you seem to need to define scale 1.0 with a larger number, eg 32. Then the initial step is 33/32 or about 3% as far as I can see.
The downside of this approach is that you don't have the same number of pixels for each original pixel. Rounding must be applied somewhere (not exactly sure where). It's definitely less simple to implement.

TheCycoONE:
No fractional scaling right now; all the draw calls will need to be changed.

I need to do it for SDL3, as it uses float coordinates for everything, so maybe then; but some things like the face portraits are multiple assets that need to be pixel perfect on the seams so no guarantee.

That said the original assets are for 640x480; so most people will probably like 2x at 1080p.

lewri:
A query from me:

Does the PR have a sanity check to make sure the scale doesn't end up with the UI going out of bounds?

E.g. 3x at 1080p would make the UI elements too big to fit the resolution.

640 x 480 *2 = 1280 x 960 min
640 x 480 *3 = 1920 x 1440 min

I would also wonder if those resolutions are used with those scales, full screen windows should render as full screen again rather than in a window.

TheCycoONE:
@lewri It does not. That's what @tobylane and I were talking about. Follow item.

TheCycoONE:
@lewri also I suppose you didn't test it. Full screen does render full screen again with those scales and those resolutions.

TheCycoONE:
Pushed one more change, related to the machine menu button on the bottom panel. It now hides / shows based on the ui_scale. Also fixed dynamically updating it when the window size changes.

TheCycoONE:
image

Fullscreen research window at 1920x1440

TheCycoONE:
Also, while not ideal - even without any changes. At 300% and 640x480 resolution you can change the scale back to something reasonable so no one should get stuck:

image

lewri:
> @lewri also I suppose you didn't test it. Full screen does render full screen again with those scales and those resolutions.

Nope, missing that luxury right now :(

ARGAMX (approved):
Great work and a great improvement! 💪
I'm impressed 👍
We really missed that functionality.

Related to situations when the cancel button is out of view I found that I can drag any window up and down and this make cancel button visible, when I drag it DOWN (if I need to get button at bottom).

Screenshot 2025-12-23 at 20 29 41

I also noticed that some of the overlays now have borders that they didn't have at 100% before.

For example, you can notice how the advisor on 200% now has a thin, fashionable mustache.

Screenshot 2025-12-23 at 20 11 40

I also noticed that at the 200% the elements on the salary increase window are misaligned.

Screenshot 2025-12-23 at 20 11 40

Also game freezes/crash when I try to drag that windows in edge positions.

Steps to reproduce:
1. Select 200% and 1280x960 and close the game.
2. Start the game.
3. Load attached save game. [0.69.1-Battenberg.sav.zip](https://github.com/user-attachments/files/24316274/0.69.1-Battenberg.sav.zip)
4. Open notification.
5. Drag the window.

Fractional scaling also came to mind and possible will be useful feature for this functionality in future.

Everything else is very cool! 🆒

Thank you, Cyco

TheCycoONE:
Thanks @ARGAMX

I've corrected the position of the marker and the skill indicator on the staff_rise dialog. I couldn't reproduce the crash but I fixed the positions first, perhaps they were drawing out of bound?

The advisor mouth is annoying me now that you pointed it out. I'm not quite sure what to do about it yet; that's an animation currently using the direct_zoom scaling - the issue goes away if you set direct_zoom = false in the config file.

TheCycoONE:
@flackr any ideas here? I seem to remember there were a few of these gap issues when direct_zoom was first added. I think I'd prefer if the animation wasn't split into different textures.

ARGAMX:
> Thanks @ARGAMX
> 
> I've corrected the position of the marker and the skill indicator on the staff_rise dialog. I couldn't reproduce the crash but I fixed the positions first, perhaps they were drawing out of bound?
> 
> The advisor mouth is annoying me now that you pointed it out. I'm not quite sure what to do about it yet; that's an animation currently using the direct_zoom scaling - the issue goes away if you set direct_zoom = false in the config file.

Yes, `direct_zoom = false` resolved it, thanks. 👍

Also this resolved some issues with overlays which I was observed for a long time. For example, such border lines were visible on the surgeon's back during the operation on all previous releases that I played. Before that scale functionality.

Another thing. I noticed that the objects in the preview window on the interface for selecting a room to build are not get scale.

-----------

100%

Screenshot 2025-12-23 at 23 55 39

-----------

200%

Screenshot 2025-12-23 at 23 55 14

ARGAMX:
Mouse right click on doctor name on UIStaff (staff dialog) iterates over all staff of that kind.

Screenshot 2025-12-24 at 00 07 47

But on 200% this click area is misaligned now above and left related to the actual text

TheCycoONE:
Fixed those two items. Thanks again @ARGAMX 

I had tested the circle, but missed that the name was another option.

PR #3163, opened 2025-12-14
Avoid emergency patients vomit, drink and use toilet

### LINKED ISSUE #3067: [Bug] Emergency patients shouldn't drink and use toilet
Image

----

### Describe the issue

In the Original TH emergency patients, never use the drink machine and never go to the toilet. 
Also they never pee on the floor.

In the Corsix TH emergency patients use the drink machine, go to the toilet. 
Also they can pee on the floor.

What's especially interesting is that they're running to the toilet right before the timer runs out. Therefore, periodically to save the last patient is simply impossible, as he is stuck in the toilet instead of saving his life.

----

### Steps to Reproduce

1. Load savegame attached to the ticket. 
2. Observe that some emergency patients are heading to the toilet.

----

### Expected Behaviour

As in the Original TH.

----

### Save Game

[emergency_patients_goes_toilet.sav.zip](https://github.com/user-attachments/files/22601103/emergency_patients_goes_toilet.sav.zip)

----

### CorsixTH Version

0.68.0, 0.69.1

----

ARGAMX:
*Fixes #3067*

I also ran an experiment in the original TH and did not register any emergency patients vomiting or urinating in the hallway.

- Avoid vomiting, urinating in the hallway, going to the toilet, or using vending machines on emergency patients.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -709,6 +709,12 @@ function Patient:tickDay()
     self.health_history["last"] = last
   end
 
+  -- Avoid vomiting, urinating in the hallway, going to the toilet,
+  -- or using vending machines on emergency patients.
+  if self.is_emergency then
+    return
+  end
+
   -- Vomitings.
   if self.vomit_anim and not self:getRoom() and not self:getCurrentAction().is_leaving and not self:getCurrentAction().is_entering then
     --Nausea level is based on health then proximity to vomit is used as a multiplier.


TheCycoONE (approved):

lewri:
Side note: terminating this early will also stop emergency sleeping illness patients yawning (L800-802).
Also cuts out a lot of happiness events for emergencies.
Maybe also of note is the section L876-879 that seems to handle a bad queue scenario.

(I had a period of downtime before a flight, though only looking on my phone).

ARGAMX:
> Side note: terminating this early will also stop emergency sleeping illness patients yawning (L800-802). Also cuts out a lot of happiness events for emergencies. Maybe also of note is the section L876-879 that seems to handle a bad queue scenario.
> 
> (I had a period of downtime before a flight, though only looking on my phone).

Yes, thank you.
I noticed this too and thought about it.
But I couldn't come up with anything better than this solution.
In fact, this entire function really needs to be refactored.
A function that's 300 lines long is not normal.
I'm not too fond of early terminating either.
But when a function is so long, there's nothing else to do.

lewri:
Yes the happiness is most likely insignificant for emergencies. 

The yawning not happening would be more noticeable to someone. The queue fix might also be noticeable (though while not seen it happen before, I also was never actively looking for it).

I agree with the refactor. Multiple private functions called by tickDay would be a good way to go. I'll leave it with you regarding the above, a refactor you could probably fix those new niggles that are created by this patch proposal.

PR #3166, opened 2025-12-21
Explain the animation system of the current implementation

Alberth289346:
Explain how the current animations work.


Alberth289346:
It has been sitting here for a month, let's merge it.

PR #3167, opened 2025-12-22
Refactor Patient:tickDay

### LINKED ISSUE #3164: Refactor Patient:tickDay due to new bugs from 3163
See PR for more context.

Side note: terminating this early will also stop emergency sleeping illness patients yawning (L800-802).
Also cuts out a lot of happiness events for emergencies.
Maybe also of note is the section L876-879 that seems to handle a bad queue scenario.

_Originally posted by @lewri in https://github.com/CorsixTH/CorsixTH/issues/3163#issuecomment-3652725787_

lewri:
Merry Christmas!

*Fixes #3164*

- Refactors Patient:tickDay() into multiple smaller functions.
- Better flexibility

There's currently a few code improvements too. But this isn't ready yet.

Edit: Sorry, there's no super easy way to read this patch, as I'm working off my less used laptop I didn't have full luxury to break it down as easily.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -664,7 +664,7 @@ function Patient:_dailyHealthChecks()
     return health
   end
 
-  local old_dying_mood, new_dying_mood = rangeMapLookup(health_treshold_rangemap, health, true)
+  local old_dying_mood, new_dying_mood = rangeMapLookup(health, health_treshold_rangemap, true)
   if not old_dying_mood then
     if new_dying_mood then
       -- Patient has entered the road of deterioration.


CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -599,102 +599,98 @@ function Patient:tick()
...
-      self:tapFoot()
-    elseif self.waiting == 20 then
-      self:yawn()
-    elseif self.waiting == 30 then
-      self:checkWatch()
-    end
-
-    if self.has_vomitted and self.has_vomitted > 0 then
-      self.has_vomitted = 0
+-- Different events happen while a patient is waiting, based on the total number of days.
```
Alberth289346:
```suggestion
-- Different events happen while a patient is waiting, based on the total number of remaining days.
```
I guess.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -599,102 +599,98 @@ function Patient:tick()
...
-    self:changeAttribute("happiness", -0.0002) -- waiting too long will make you sad
-    -- There is a 1/3 chance that the patient will get fed up and leave.
-    -- This is potentially run 10 ((0.22-0.18)/0.004) times, hence the 1/30 chance.
-    -- If patient is already in the cure room, let the treatment happen.
-    if not self:_checkIfCureRoom(self:getRoom()) and math.random(1,30) == 1 then
-      self:setDynamicInfoText(_S.dynamic_info.patient.actions.fed_up)
-      self:setMood("dying1", "deactivate")
-      self:goHome("kicked")
+  -- Define different thresholds for deteriorating health and their mood changes.
+  -- First value is the mood to deactivate, the second to activate (as applicable).
```
Alberth289346, line 644: As an enhancement, you could first disable all 'dying' moods, and then enable the one you want to have.

That catches cases when a patient 'skips' a stage for some reason. (One reason may be setting health hard-coded to 0, but not sure.)

Likely this doesn't have much impact, due to the priorities in showing moods.

CorsixTH/Lua/entities/humanoids/patient.lua
lewri: Yeah I forgot that one :D

CorsixTH/Lua/entities/humanoids/patient.lua
lewri: I'm not even sure why we fire the mood update daily tbh and not just once at the threshold. If I think of a nice way of doing it I'll add it.

lewri:
Okay I should be finished with this for now. However, I haven't had an ability to run the code patch yet if someone else is able to check that.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -764,119 +781,107 @@ function Patient:tickDay()
...
-      -- If we are queueing, let the queue handle the situation.
-      for _, current_action in ipairs(self.action_queue) do
-        if current_action.name == "queue" then
-          local callbacks = current_action.queue.callbacks[self]
-          if callbacks then
-            callbacks:onGetSoda(self, machine, lx, ly, after_use)
-            return
-          end
+  -- Take the necessary steps to let the patient to get a soda.
+  local function constructGetSodaAction(machine, lx, ly)
```
TheCycoONE: Why is this one not _constructGetSodaAction?

TheCycoONE (approved):
Quite a bit clearer, thanks

CorsixTH/Lua/entities/humanoids/patient.lua
lewri: It's already inside a private function. But I can rename it if you'd prefer.

TheCycoONE: no, I missed that it was a local function. Sorry.

TheCycoONE: It is only used once, as a tail call; I suspect it shouldn't be a function.

lewri: Yeah it is. It was only this way for readability but it can be dropped

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -708,45 +704,66 @@ function Patient:tickDay()
...
-    -- as a temp fix for this I have added 0.5 to the < nausea equation,
-    -- this may want adjusting or removing when the other factors are in the game MarkL
-    if self:getAttribute("health") <= 0.8 or numVomit > 0 or self:getAttribute("happiness") < 0.6 then
-      nausea = nausea * ((numVomit + 1) * proximityVomitMult)
-      if math.random() < nausea + 0.5 then
-        self:vomit()
+--! Process a patient's sudden need for the toilet.
+--! We will either do it on the floor (10% chance, daily), or seek out a toilet.
+function Patient:handleToiletNeed()
+  if math.random(1, 10) < 5 then
```
Alberth289346: That's 40% rather than the claimed 10%.

CorsixTH/Lua/entities/humanoids/patient.lua
TheCycoONE: Could it be converted to a top level local (no upvalues)

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -599,102 +599,98 @@ function Patient:tick()
...
+    {upper = 0.22, value = {false, "dying1"}},    -- Getting rather unwell now
+    {value = false}
+  }
+  -- Special handling for a patient almost passed.
+  local almost_dead_value, dying_value = 0.01, 0.0
+  if health < almost_dead_value then
+    if health == dying_value then
+      -- Patient died, will die when they leave the room, will be cured, or is leaving
+      -- the hospital. Regardless we do not need to adjust any other attribute
+      self:setToDying()
```
ARGAMX: In fact, after getting health 0 and after this `setToDying` function call, the patient may even be cured and not die.
That is, the patient has not died and will not necessarily die when their health drops to 0.

This will happen, for example, when his health has dropped to 0 after he has already entered the cure room.
And if he is will be cured there, he will survive despite the fact that this code was executed.

That's why this function is called `setToDying`, and not, say, `setToDead`.
I don't know, maybe it's not obvious from the name.
But that's exactly what I was trying to say naming it that way. 🙂

And in reality, dying patients don't always die when they go to the hospital. 🙂
Sometimes they are saved.
So, I think the comment should be adjusted to reflect these thing. 🙂

Although perhaps the comment takes this into account, and I may not have understood what is written here, but for some reason the part "Patient died, will die when they leave the room" bothers me.

By the way, if the patient is already leaving the hospital, his health will not decrease. So this code will not execute.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -764,119 +781,107 @@ function Patient:tickDay()
...
+  end
+
+  -- Go get a drink.
+  constructGetSodaAction(machine, lx, ly)
+end
+
+--! If the patient is sitting on a bench or standing and queued, it may be a situation
+--! where he/she is not in the queue anymore, but should be. If this is the case for
+--! more than 2 ticks, go to reception.
+function Patient:checkPatientIsStillQueueingForARoom()
```
ARGAMX: Not you,
I wonder in what cases this might happen?..
What a strange situation.
You probably don't know as I don't.
So it's a rhetorical question.
Maybe if the room deleted?..
But it's strange that we have such code.
Looks like a workaround for some kind of bug.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: Basically, this comment (`Patient died, will die when they leave the room`) was previously written after `self:setToDying()` and explained the `return` call, not the `self:setToDying()` call.
Now it's moved up, and now I think it looks bit weird.

Screenshot 2025-12-27 at 15 30 11

ARGAMX: `if self:_dailyHealthChecks() == 0.0 then return end`

So it turns out this comment should be moved here. ☝️
Or just deleted.

lewri: I'm confused what's being said here. Can you write the block again for what you mean please? 

RE going home patients, that'll be done from `Humanoid:tickDay()` just before this function is called in `Patient:tickDay()` (or at least it used to, that might have changed)

Alberth289346: > "Patient died, will die when they leave the room" bothers me

Perhaps there is lack of consistent terminology here. But if comment bothers you, try to improve it.

Eg "Patient is doomed, but cannot die in the room. They will die as soon as they are outside the room."

"is doomed" can be rewritten to "cannot be cured".

CorsixTH/Lua/entities/humanoids/patient.lua
Alberth289346: Looks like a workaround indeed. Unfortunately adding more code looks simpler than analyzing why the existing code fails.

If you want to find out more, read the commit that added the code. With some luck there is also an issue number for more information.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: I not sure how to improve this comment, since the "return" line it described no longer follows it.
It no longer performs the function it done before.
I provided a screenshot of what this comment previously described.
Now the code has changed and it doesn't do that.
It just not needed here.
I think we can just delete it.

```suggestion
      self:setToDying()
```

ARGAMX: Another option.
`self:setToDying()` will set `set_to_die` to `true`.
So we can copy the comment that describes this `set_to_die` flag?

```suggestion
      -- Mark patient for death. 
      -- The patient may be currently busy, but as soon as the patient is unoccupied
      -- the patient will start the dying scenario, except the cases when
      -- the patient is cured or sent home while being busy.
      self:setToDying()
```

lewri: Ah just fix the comment. Got it, sorry

CorsixTH/Lua/entities/humanoids/patient.lua
lewri: I converted it to a private function by mistake instead of a local. The amount of uses to self already in it, might make more sense to leave it private.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -764,119 +781,110 @@ function Patient:tickDay()
...
-  -- anymore, but should be. If this is the case for more than
-  -- 2 ticks, go to reception
+  -- Go get a drink.
+  self:_constructGetSodaAction(machine, lx, ly)
+end
+
+--! If the patient is sitting on a bench or standing and queued, it may be a situation
+--! where he/she is not in the queue anymore, but should be. If this is the case for
+--! more than 2 ticks, go to reception.
+function Patient:checkPatientIsStillQueueingForARoom()
```
TheCycoONE: Should be underscore?

lewri: Maybe, I wasn't sure if the function could be useful elsewhere but I see no usage right now

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -903,6 +911,82 @@ function Patient:tickDay()
...
+  end
+
+  -- See if a patient should yawn today.
+  -- Don't attempt if the patient is already waiting with a yawn event
+  local will_yawn = self.waiting and waiting_events[self.waiting] == "yawn"
+  if self.disease.yawn and not will_yawn and math.random(1, 10) == 5 then
+    self:yawn()
+  end
+
+  self:checkPatientIsStillQueueingForARoom()
```
lewri:
```suggestion
  self:_checkPatientIsStillQueueingForARoom()
```

CorsixTH/Lua/entities/humanoids/patient.lua
lewri:
```suggestion
function Patient:_checkPatientIsStillQueueingForARoom()
```

ARGAMX (approved):
Thank you. 
Refactor result is even better than I imagined 👍

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -599,102 +599,98 @@ function Patient:tick()
...
-  -- is there time to say a prayer
-  elseif health == 0.0 then
-    self:setToDying()
-    -- Patient died, will die when they leave the room, will be cured, or is leaving
-    -- the hospital. Regardless we do not need to adjust any other attribute
-    return
+    return health
+  end
+
+  local old_dying_mood, new_dying_mood = rangeMapLookup(health_treshold_rangemap, health, true)
```
ARGAMX:
```suggestion
  local old_dying_mood, new_dying_mood = rangeMapLookup(health, health_treshold_rangemap, true)
```

I am sure the parameters `health` and `health_treshold_rangemap` here should be swapped.
because the order of parameters of the `rangeMapLookup` function is different:

```Lua
--! Find the smallest bucket with its upper value less or equal to a given number,
--! and return the value of the bucket, or its index.
--!param number (number) Value to accept by the bucket.
--!param buckets (list) Available buckets, pairs of {upper=x, value=y} tables,
-- in increasing x value, where nil is taken as infinite. The y value is
-- returned for the first bucket in the list where number <= x (in normal mode).
-- If y is nil, the index of the bucket in the list is returned.

--!param alt_mode (boolean) If true, the comparison becomes number < x instead.
--!return (number) Value or index of the matching bucket.

function rangeMapLookup(number, buckets, alt_mode)
...
```

lewri: health = number
health_threshold_rangemap = buckets
true = alt_mode

Generally I haven't executed this code, but this matches up with the format of the rangemaps used in VIP class.
e.g.
https://github.com/CorsixTH/CorsixTH/blob/d7bb75756d77c417e192a443b310ad4130f0481f/CorsixTH/Lua/entities/humanoids/vip.lua#L307-L318

lewri: Oops, only saw the other half of this comment (the important half 🤣) now. Will update it.

PR #3169, opened 2025-12-24
Do not show the mood of the VIP and Inspector (#3145)

### LINKED ISSUE #3145: Do not show the mood of the VIP and Inspector
Image

----

### Describe the issue

In the original TH if you hover over Inspector or VIP you will never see the mood above the head.
In the CTH if you hover over Inspector or VIP you can see the mood above the head.

----

### Steps to Reproduce

1. Load attached save game.
2. Observe that the incoming notification fax contains a proposal to invite a VIP.
3. Go to Town Map (F4) and set Heat Level to max or min value. 
4. Close the Town Map and open the VIP fax.
5. Send an invitation to a VIP.
6. Press 5 and wait until VIP reaches Reception.
7. Follow him and hover to see his mood.
8. Observe that he has a hot or cold mood.

----

### Expected Behaviour

The Inspector and VIP mood should not show as in the original game.

----

### Save Game

[event-vip-fax.0.65.0.sav.zip](https://github.com/user-attachments/files/23711252/event-vip-fax.0.65.0.sav.zip)

----

### CorsixTH Version

0.68.0, 0.69.1

----

Fraggenard:
*Fixes #3145*

- VIPs and Inspectors will no longer show hot or cold moods on hover as in the original game


tobylane:
Thanks for your first contribution!

Please can you add an afterload to add this attribute to all relevant existing humanoids? This [recent commit](https://github.com/CorsixTH/CorsixTH/commit/a8ec73e5859df48f393ab3296e16609dccda5740) is a useful example.

Fraggenard:
> Thanks for your first contribution!
> 
> Please can you add an afterload to add this attribute to all relevant existing humanoids? This [recent commit](https://github.com/CorsixTH/CorsixTH/commit/a8ec73e5859df48f393ab3296e16609dccda5740) is a useful example.

Thanks for the helpful example! I wasn't sure if this was a drastic enough change to warrant incrementing the save file version, but if it is it will be a quick change.

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -353,6 +353,8 @@ function Humanoid:Humanoid(...)
   self.build_callbacks  = {--[[set]]}
   self.remove_callbacks = {--[[set]]}
   self.staff_change_callbacks = {--[[set]]}
+
+  self.hover_moods = true
```
lewri: If you would like a cleaner improvement, you could make a function in humanoid called enableHoverMoods e.g.

```lua
--! Set if hover moods apply to this humanoid (true by default).
--! Override in derived classes when not to apply.
function Humanoid:setIfHoverMoodsVisible()
  self.hover_moods = true
end
```

Then in VIP and Inspector add the overrides e.g. (adjust for inspector)
```lua
function Vip:setIfHoverMoodsVisible()
  self.hover_moods = false
end
```

Then at your line 357 in this file you call the function instead:
```lua
self:setIfHoverMoodsVisible()
```
This will look at the subclass first for the function if it exists, and execute that instead (i.e. the VIP and Inspector)

This eliminates the line in epidemic.lua and 493 of world.lua.

CorsixTH/Lua/world.lua
```diff
@@ -2604,6 +2605,15 @@ function World:afterLoad(old, new)
...
 
+  -- Fix VIPs to not show mood on hover
+  for _, entity in ipairs(self.entities) do
+    if class.is(entity, Humanoid) and entity:isType() then
+      if entity:isType("VIP") or entity:isType("Inspector") then
+        entity.hover_moods = false
+      end
+    end
+  end
+
```
lewri: This may be cleaner with my comment above (note, for the afterload you need to wrap in an `old < xyz` where xyz is the new savegame version number you set in `app.lua` (the variable called `local SAVEGAME_VERSION = xxx -- Some comment you update`. For new version, just bump the current number in master by 1.

```lua
if old < xyz then
  for _, entity in ipairs(self.entities) do
    if class.is(entity, Humanoid) then
      entity:setIfHoverMoodsVisible()
    end
  end
end
```

^ For the above, `ipairs` loop is well used on `self.entities`. `pairs` is safer but shouldn't be necessary.

lewri (commented):
Alternative recommended approach added that may be cleaner.

ARGAMX (commented):
Working solution!
Nice job!

I wonder if we could achieve the same thing without adding an additional variable for this?

Then we won't need `afterLoad` for that.

How about to go there:

https://github.com/CorsixTH/CorsixTH/blob/1d1b7ed8f0c9f621ecb5daf8267ce82b1076eee0/CorsixTH/Lua/entity.lua#L263-L264

And add a one line here, which will make `return` for VIPs and Inspectors?
This would looks like this:

```lua
function Entity:setMoodInfo(new_mood)
  if class.is(self, Vip) or class.is(self, Inspector) then return end
  if new_mood then
```

What do you think about this? 🙂

CorsixTH/Lua/world.lua
Fraggenard: I like this. Will clean everything up when I'm at my PC next. As a general rule, is the savegame version updated every time an afterload function is added?

lewri: Yes :)

lewri (approved):

PR #3170, opened 2025-12-27
UI Scale Guards

TheCycoONE:
Adds a minimum resolution / maximum scale condition to the options screen and when drag resizing the window.

I also added some resolution options that were conspicuously missing from a 2025 game.

This is a continuation of some of the discussion in #3159

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1014,12 +1014,20 @@ function App:fixConfig()
     if key == "language" and type(value) == "string" then
       self.config[key] = value:lower()
 
-    -- For resolution, check that resolution is at least 640x480
-    elseif key == "width" and type(value) == "number" and value < 640 then
-      self.config[key] = 640
+    -- For resolution, clamp resolution to at least 640x480
+    elseif key == "width" and (type(value) ~= "number" or value < App.MIN_WINDOW_WIDTH) then
+      self.config[key] = App.MIN_WINDOW_WIDTH
 
-    elseif key == "height" and type(value) == "number" and value < 480 then
-      self.config[key] = 480
+    elseif key == "height" and (type(value) ~= "number" or value < App.MIN_WINDOW_HEIGHT) then
+      self.config[key] = App.MIN_WINDOW_HEIGHT
+
+    -- For scale, clamp to integer scale >= 1
+    elseif key == "ui_scale" then
+      if type(value) == "number" then
+        self.config[key] = math.max(math.floor(value), 1)
+      else
+        self.config[key] = 1
+      end
 
     elseif (key == "scroll_speed" or key == "shift_scroll_speed") and
         type(value) == "number" then
@@ -1041,6 +1049,14 @@ function App:fixConfig()
       end
       self.config[key] = value
     end
+  end
+
+  -- clamp scale to suitable values for current resolution
+  if self.config.ui_scale * App.MIN_WINDOW_WIDTH > self.config.width or
+      self.config.ui_scale * App.MIN_WINDOW_HEIGHT > self.config.height then
+    self.config.ui_scale = math.floor(math.min(
+        self.config.width / App.MIN_WINDOW_WIDTH,
+        self.config.height / App.MIN_WINDOW_HEIGHT))
   end
 end
 


TheCycoONE:
Some thoughts:

~Maybe passing in the min size explicitly to surface would be better than scale.~

~If a custom resolution is entered the minimum should be enforced.~

~Should we clamp or override values in the config file if they aren't legal? How?~

~Why is the scale drop-down always tall enough for 3 entries when there are 1 or two?~

CorsixTH/Lua/app.lua
```diff
@@ -1013,6 +1013,10 @@ function App:fixConfig()
     elseif key == "height" and type(value) == "number" and value < 480 then
       self.config[key] = 480
 
+    -- For scale, check that it is an integer scale >= 1
```
Alberth289346:
```suggestion
    -- For scale, force it to an integer scale >= 1
```
It doesn't check.

Alberth289346 (approved):

CorsixTH/Lua/app.lua
```diff
@@ -1013,6 +1013,10 @@ function App:fixConfig()
...
+    elseif key == "ui_scale" and type(value) == "number" then
+      self.config[key] = math.min(math.floor(value), 1)
```
tobylane:
```suggestion
    elseif key == "ui_scale" then
      self.config[key] = type(value) == "number" and math.min(math.floor(value), 1) or 1
```
Reset if it isn't a number

tobylane (commented):
Looks good. Should it clean up non numbers?

ARGAMX:
For me all values ​​other than 100% is dissapeared.
I had to increase the screen resolution to get it back to at least 200%. 🙂
But I guess that's okay.

Scale value not saves for me between session.
It always reset to 100% after close-open game.
So I have to set it manually to 200% after each run.

On 200% I also noticed the same borders that were around the advisor's mouth; 
they can now be seen on the timer clock.
I set direct zoom to false, so the advisor doesn't have a mustache. 
But for some reason, this not help for the clock.

Screenshot 2025-12-28 at 16 34 57
(This also can be obtain in "UI scale" #3159)

It can seen on the same save game I attached to "UI scale" PR.

I also noticed that the notification window is always half open.

Screenshot 2025-12-28 at 16 38 02
(This also can be obtain in "UI scale" #3159)

Use the same save to receive some notifications.

ARGAMX (approved):

CorsixTH/Lua/app.lua
TheCycoONE: Probably, though none of the other fixes seem to either, e.g. resolution just above it.

TheCycoONE:
> For me all values ​​other than 100% is dissapeared. I had to increase the screen resolution to get it back to at least 200%. 🙂 But I guess that's okay.

I suspect this was intentional and you had a scale that would have made the full screen dialogs too big? What was your resolution?
> 
> Scale value not saves for me between session. It always reset to 100% after close-open game. So I have to set it manually to 200% after each run.

Oops, fixed.
> 
> On 200% I also noticed the same borders that were around the advisor's mouth; they can now be seen on the timer clock. I set direct zoom to false, so the advisor doesn't have a mustache. But for some reason, this not help for the clock.
> Screenshot 2025-12-28 at 16 34 57 (This also can be obtain in "UI scale" #3159)

Hmm, that wouldn't have regressed due to this patch. I'll open another PR after if I figure out these artifacts.
> 
> It can seen on the same save game I attached to "UI scale" PR.
> 
> I also noticed that the notification window is always half open.
> Screenshot 2025-12-28 at 16 38 02 (This also can be obtain in "UI scale" #3159)
> 
> Use the same save to receive some notifications.

Missed that earlier - I'll open another PR for this.

CorsixTH/Lua/app.lua
TheCycoONE: Fixed for both

TheCycoONE:
Thanks for all the feedback. I hope I have addressed everything (including my own concerns about 640x480 being scattered around too much).

ARGAMX:
Screenshot 2025-12-29 at 18 23 52

I thought maybe this was a strange situation:
1. I launched the game.
2. It launched at 800x600 (I launched it at that resolution last time).
3. I went into the settings and only saw 100% in the UI Scale list. So, it looks like scaling isn't available to me.
4. Even though my screen supports 1680x1050, you can see it in the screenshot.
5. The 200% option only becomes available if I select 1280x960 or higher. But how am I supposed to know this?

In general, I have a feeling that without knowing about this function and how it works, user might not even understand what it exist and what he shoud do to make something other than 100% available.

TheCycoONE:
I think if you don't play at a high enough resolution to see the scale you don't need it? Or people look for it because of the release notes, or a tooltip on the scale option I suppose.

PR #3171, opened 2025-12-28
UI Scale. Misc scale fixes

TheCycoONE:
The purpose of this PR is to fix some of the visual artifacts that happen when UI scaling.

- [x] artifact on watch when red pie slice is visible
- [x] artifact on adviser mouth when speaking
- [x] bottom panel message slot half open with no messages.


ARGAMX (approved):
_chuck-norris

PR #3172, opened 2025-12-29
Fix rendering ground tile

### LINKED ISSUE #3142: Floor should be drawn just before drawing the tile objects and animations
### Describe the issue

Image

### Steps to Reproduce

Build shown chair items as in screenshot.

### Expected Behaviour

Sprites should be shown correctly

- [ ] Right mouse panning
- [x] Falling actions (applies to CorsixTH versions < 0.70.0)

### Save Game

_No response_

### CorsixTH Version

0.69

### Operating System

Win11

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

### Additional Information

_No response_

Alberth289346:
*Fixes #3142 

Do not draw all floor tiles (and other ground tiles) in a separate function; draw them as part of rendering each tile instead.

This causes the ground to be painted *after* painting the tiles behind each tile. This causes hiding any pixels that "stick out" from the former tiles.

Screenshot from 2025-12-29 13-08-31

### DIFF
--- a/CorsixTH/Src/th_map.cpp
+++ b/CorsixTH/Src/th_map.cpp
@@ -959,30 +959,18 @@ bool level_map::layer_exists(uint16_t layer, int& height) const {
   return false;
 }
 
-void level_map::draw_floor(render_target* pCanvas, int iScreenX, int iScreenY,
-                           int iWidth, int iHeight, int iCanvasX,
-                           int iCanvasY) const {
-  for (map_tile_iterator itrNode1(this, iScreenX, iScreenY, iWidth, iHeight);
-       itrNode1; ++itrNode1) {
-    int tile_x = itrNode1.tile_x_position_on_screen() + iCanvasX - 32;
-    int tile_y = itrNode1.tile_y_position_on_screen() + iCanvasY;
+void level_map::draw_ground(const map_tile* tile, int tile_x, int tile_y,
+                            render_target* pCanvas) const {
+  draw_layer(tile, tile_x, tile_y, tile_layer::ground, pCanvas);
 
-    // First, draw the floor tile as it should be below everything else.
-    int height = 32;
-    uint16_t layer = itrNode1->tile_layers[tile_layer::ground];
-    wall_blocks->get_sprite_size(layer & 0xFF, nullptr, &height);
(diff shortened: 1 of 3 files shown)


CorsixTH/Src/th_map.cpp
```diff
@@ -249,12 +249,12 @@ map_tile_flags::operator uint32_t() const {
 }
 
 map_tile::map_tile()
-    : iParcelId(0),
+    : tile_layers{},
```
TheCycoONE, line 252: As none of these depend on the constructor or the parameters they could be set on the members themselves. That would make ordering them more local.

CorsixTH/Src/th_map.h
```diff
@@ -402,8 +402,8 @@ class level_map {
   void set_overlay(map_overlay* pOverlay, bool bTakeOwnership);
 
  private:
-  void draw_floor(render_target* pCanvas, int iScreenX, int iScreenY,
-                  int iWidth, int iHeight, int iCanvasX, int iCanvasY) const;
+  void draw_ground(const map_tile* tile, int tile_x, int tile_y,
```
TheCycoONE, line 405: We seem to be confused about camel or snake case for parameters. This is at least consistent with other tile parameters but I'm never sure which I should be using.

TheCycoONE:
image

We seem to lose wheels with this PR.

TheCycoONE:
image

TheCycoONE:
Currently the regressions seem worse than the improvements unfortunately.

TheCycoONE:
image

TheCycoONE:
image

TheCycoONE:
image

Alberth289346:
Hmm, what a mess :(

CorsixTH/Src/th_map.h
Alberth289346: As far as I know, all names with type prefixes are old. I tend to change them to snake case, but not always.

The old `draw_floor` was likely copied from another render function.

Alberth289346:
This approach seems to need at least needs different tiles being painted, which suggests #1638 should be done first, or maybe solve this at the same time.

PR #3173, opened 2025-12-31
Fix skipped rooms when seeking a new diagnosis room

### LINKED ISSUE #3116: [Bug] Skipping rooms in seek_room cycle
Edit: Actual issue here was that #3109 caused a regression where only half a patient's rooms were checked as possible diagnosis rooms. The vomit just exposed this in investigation.

### Describe the issue

Video from https://github.com/CorsixTH/CorsixTH/pull/3109#pullrequestreview-3405055103:

[Screen Recording 2025-10-31 at 19.09.34.webm](https://github.com/user-attachments/assets/65bbd78f-ff7a-47be-8f29-edffd872d8a8)

When a patient wait gets interrupted by a vomit action (potentially others); they will start seek_room from the beginning again and potentially try to look for a different room.

### Steps to Reproduce

1. I don't have a reliable save for this

2. Have a hospital with a GP office and reception only and a non-consultant
3. Get a patient to get seen in GP office and try get a ? as no rooms available for diagnosis. Don't open the fax.
4. Build a room used by their disease before a patient vomits
5. Watch patient vomit and see if they eventually pick up the new room after a vomit.

### Expected Behaviour

In essence trying to find a different room you could go to can be a good thing; I don't think TH locked a patient down to just one room when they were waiting for more diagnosis rooms.

However, the vomit shouldn't trigger the search, it likely should be part of a build callback or something similar.

### CorsixTH Version

0.69.1

ARGAMX:
*Fixes #3116*

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/seek_room.lua
+++ b/CorsixTH/Lua/humanoid_actions/seek_room.lua
@@ -63,7 +63,6 @@ local action_seek_room_find_room = permanent"action_seek_room_find_room"( functi
       if room then return room end
     end
 
-    local tried_rooms = 0
     local diagnosis_rooms = humanoid.available_diagnosis_rooms
     -- Make numbers for each available diagnosis room. A random index from this list will be chosen,
     -- and then the corresponding room index is taken as next room. (The list decreases for each room
@@ -73,7 +72,7 @@ local action_seek_room_find_room = permanent"action_seek_room_find_room"( functi
       rooms_to_check[i] = i
     end
 
-    while tried_rooms < #diagnosis_rooms do
+    while #rooms_to_check > 0 do
       -- Choose a diagnosis room from the list at random. Note: This ignores the initial diagnosis room!
       local room_at_index = math.random(1,#rooms_to_check)
       room_type = diagnosis_rooms[rooms_to_check[room_at_index]]
@@ -90,7 +89,6 @@ local action_seek_room_find_room = permanent"action_seek_room_find_room"( functi
         action.room_type = room_type
         return room
       else
-        tried_rooms = tried_rooms + 1
         -- Remove the index of this room from the list of indices available in available_diagnosis_rooms
         table.remove(rooms_to_check, room_at_index)
         -- If the room can be built, set the flag for it.


CorsixTH/Lua/humanoid_actions/seek_room.lua
```diff
@@ -73,7 +73,7 @@ local action_seek_room_find_room = permanent"action_seek_room_find_room"( functi
       rooms_to_check[i] = i
     end
 
-    while tried_rooms < #rooms_to_check do
+    while tried_rooms < #diagnosis_rooms do
```
TheCycoONE: I don't see how this changes anything?

lewri: At the start of the while loop `#diagnosis_rooms` == `#rooms_to_check`.

`rooms_to_check` is essentially a shallow clone of `diagnosis_rooms`? Because we don't actually want to perform any action on the latter.

That's not an answer to what changes but I believe is the observation.

Edit: referenced wrong variable in comment.

lewri: Sorry, they're actually equivalent to. I think diagnosis rooms listed them in both in iterative pairs and a list, but tried rooms is just the iterative I think.

lewri: I'm still not entirely sure how this works, you may need to explain more. But after coming back after a few weeks it looks like it's possible that when `tried_rooms` increases by 1, `rooms_to_check` thus also decreases by 1. This creates a scenario where we actually only check half the diagnosis rooms available, I believe.

If that's true, it'd be possibly a separate bug I introduced in #3109; and that while loop should just be `while #rooms_to_check > 0`

Not sure how that relates to the issue to be fixed though.

lewri: Yeah, looks like I made a shrinking table by accident that is meant to be a static reference point in the current loop. If the change above was made to the while loop,  the `tried_rooms` stuff could be entirely dropped too.

I can push that fix in a separate pr if you like, but it overlaps with this patch as a change on the same line (yay)

TheCycoONE: Interested in seeing your PR lewri, assume @ARGAMX doesn't mind if you take this issue?

ARGAMX: I am pushed Lewri version in this PR.

Although both versions worked exactly the same. 😛

lewri: Yep that's the patch in a nutshell. Though how does this prevent vomit causing a restart still? This just makes sure every room is tried when looking for next diagnosis room.

ARGAMX: > Though how does this prevent vomit causing a restart still? This just makes sure every room is tried when looking for next diagnosis room.

Yeap, this just makes sure every room is tried when looking for next diagnosis room and it not preventing vomit restarts seek_room.

These two are essentially different problems. 

I was specifically concerned about the regression described here: https://github.com/CorsixTH/CorsixTH/pull/3109#pullrequestreview-3405055103 and fixing it.

If that issue 3116 wasn't about that regression, then seems we created the issue not about what I wrote and we never created an issue about the regression. We just merged regression bug and that is.

If that issue was about that, then we could create a separate issue specifically about the problem of the search restarting over and over again.

In general, it seems that in any case another issue needs to be created.

Also I think search restarts not only because of the vomit.

lewri: I agree. In essence then this PR just fixes skipped rooms, which should have this PR's title updated as such.

A new issue regarding the logic then probably helps after this.

@TheCycoONE hopefully we're at the bottom of it now.

lewri (approved):
As previous comments, description of #3116 is fixed by this, but the actual title of the issue is about something else that still exists.

PR #3180, opened 2026-01-04
Inherit from Japanese audio files if present

Fraggenard:
- In the Japanese PC release of Theme Hospital, Japanese announcer audio is found in SOUND-6.DAT, which is formatted like and based on the English SOUND-0.DAT. This small change to the Japanese translation file would allow for those audio files to be used if the user has pointed CorsixTH to the Japanese version of the game, otherwise English audio will be used as a fallback.


TheCycoONE:
I actually don't know if there's any downside to this approach. I trust you have tested it.

PR #3183, opened 2026-01-06
Fix range function used for displaying heat

### LINKED ISSUE #3181: Malfunctions in the heating display
----

### Describe the issue

Seem there is some Malfunctions in the heating display.

Heating display mode can be changed:
1. Start or load any game.
2. On the top menu select Options -> Warmth Colours 

This is what the BLUE GREEN RED heating display mode looks like:

BLUE - cold
GREEN - ok
RED - overheat

Image

This is what the YELLOW ORANGE RED heating display mode looks like:

Image

You can only see YELLOW and WHITE.
This mode doesn't display heat properly.

----

### Save Game

[3056 Operating Theatre blocks resolved 4.sav.zip](https://github.com/user-attachments/files/22984109/3056.Operating.Theatre.blocks.resolved.4.sav.zip)

----

### Steps to Reproduce

Load save game and open “Town map”.

----

### Expected Behaviour

YELLOW ORANGE RED mode should work somehow. It's meaningless now.

Moreover, the BLUE GREEN RED mode was apparently supposed to display not only three colors, but also their gradations. 
But that doesn't work either.

Check `temperature_theme` in `th_lua_ui.cpp`

----

### CorsixTH Version

0.60, 0.62, 0.63, 0.64, 0.65, 0.66, 0.67, 0.68, 0.69

TheCycoONE:
I broke this in bef0689e2c40356aa8102b2f6fee687fbca422c3 back in 2015 when I mixed up max vs min for clamping the value.

Fixes #3181

Good find @ARGAMX 

image

image

image

### DIFF
--- a/CorsixTH/CppTest/test_th_lua_ui.cpp
+++ b/CorsixTH/CppTest/test_th_lua_ui.cpp
@@ -1,16 +1,17 @@
 #include <catch2/catch_test_macros.hpp>
 
-#include "../Src/th_lua_ui.cpp"  // NOLINT(bugprone-suspicious-include)
+#include "../Src/th_lua_ui.h"
 
-TEST_CASE("range_scale scales correctly", "[range_scale]") {
-  REQUIRE(range_scale(0, 100, 0, 10, 20) == 10);
-  REQUIRE(range_scale(0, 100, 50, 10, 20) == 15);
-  REQUIRE(range_scale(0, 100, 100, 10, 20) == 20);
-  REQUIRE(range_scale(0, 200, 100, 0, 100) == 50);
-  REQUIRE(range_scale(50, 150, 100, 0, 100) == 50);
-  REQUIRE(range_scale(0, 100, 25, 200, 100) == 175);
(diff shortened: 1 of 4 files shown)


CorsixTH/CppTest/test_th_lua_ui.cpp
```diff
@@ -0,0 +1,12 @@
+#include <catch2/catch_test_macros.hpp>
+
+#include "../Src/th_lua_ui.cpp"  // NOLINT(bugprone-suspicious-include)
```
TheCycoONE: This isn't safe and I should be extracting range_scale into a header. Annoying.

TheCycoONE:
It's pretty small, but the two commits may help the review; the first shows the bug fix; the second moves it to a new header file to make it testable.

ARGAMX (commented):
Thanks, works perfect for me. 👍

Screenshot 2026-01-06 at 09 02 50

Screenshot 2026-01-06 at 09 08 46

PR #3184, opened 2026-01-08
Lua 5.5 support

TheCycoONE:
Updates CorsixTH for compatibility with Lua 5.5

Could use a bit more testing before merging.

`global` is also now a reserve word that according to the lua reference manual cannot be used in table variable names: See https://www.lua.org/manual/5.5/manual.html#3.1

I didn't notice a problem, but out of caution I have removed all references.


lewri:
NB: Wiki compile deps table to be updated post merge.
(likely in notes of lua, 5.5 support in master and  CorsixTH release 0.70.0+)

tobylane (approved):

Alberth289346 (approved):

PR #3190, opened 2026-01-11
Fix duration calculation for playback callback

### LINKED ISSUE #3187: Sound announcements overlap each other
----

### CorsixTH Version

Any trunk master build with c088006

----

### Describe the issue

The announcement queue isn't functioning properly.
The receptionist's audio announcements don't play sequentially, but overlap each other.

Previously, In version 0.69.1, announcements were voiced one by one, only one at a time.

git bisect points to #3146

----

### Save Game

[Middle Op Theater.zip](https://github.com/CorsixTH/CorsixTH/files/10112581/Middle.Op.Theater.zip)

----

### Steps to Reproduce

1. Prepare 2 builds. build 0.69.1 and any trunk master build with c088006.
2. Run 0.69.1.
3. Load attached savegame.
4. You will hear after loading several announcements in the queue are voiced one by one in turn without overlapping.
5. Close 0.69.1. and run trunk master build.
6. Load attached savegame.
7.  You will hear after loading several announcements are voiced simultaneously and overlap each other.

----

### Expected Behaviour

Announcements queue should work as in 0.69.1.

TheCycoONE:
Off by one error on loop multiplier

Fixes #3187

### DIFF
--- a/CorsixTH/Src/th_lua_sound.cpp
+++ b/CorsixTH/Src/th_lua_sound.cpp
@@ -263,8 +263,9 @@ int l_soundfx_play(lua_State* L) {
         static_cast<int>(luaL_checkinteger(L, 6));
     int& callback_id = played_sound_callback_ids[played_sound_callback_index];
 
-    Uint32 interval = static_cast<Uint32>(
-        pArchive->get_sound_duration(iIndex) * loops + iPlayedCallbackDelay);
+    Uint32 interval =
+        static_cast<Uint32>(pArchive->get_sound_duration(iIndex) * (loops + 1) +
+                            iPlayedCallbackDelay);
     SDL_TimerID timersID =
         SDL_AddTimer(interval, played_sound_callback, &callback_id);
     map_sound_timers.emplace(std::pair<int, map_timer_info>(


CorsixTH/Src/th_lua_sound.cpp
```diff
@@ -263,8 +263,9 @@ int l_soundfx_play(lua_State* L) {
         static_cast<int>(luaL_checkinteger(L, 6));
     int& callback_id = played_sound_callback_ids[played_sound_callback_index];
 
-    Uint32 interval = static_cast<Uint32>(
-        pArchive->get_sound_duration(iIndex) * loops + iPlayedCallbackDelay);
+    Uint32 interval =
+        static_cast<Uint32>(pArchive->get_sound_duration(iIndex) * (loops + 1) +
```
lewri, line 267: It is line 253 that handles infinite loop playback right?

ARGAMX (approved):
Resolves the issue for me. Thanks. 👍

CorsixTH/Src/th_lua_sound.cpp
lewri: Argamx has tested, we good.

TheCycoONE: Yes, the if condition handles infinite.

PR #3191, opened 2026-01-11
UI Scale. Fix font scaling

### LINKED ISSUE #3192: UI Scaling: Builtin Font not scaling
The folder selection path font does not scale consistently. It did not originally scale at 200%, but did at 300% then back to 200%. May have also needed to load a game to apply the font scaling change?

TheCycoONE:
When scale changes, the 'builtin' font was not being scaled.

*Fixes #3192*

### DIFF
--- a/CorsixTH/Lua/graphics.lua
+++ b/CorsixTH/Lua/graphics.lua
@@ -372,7 +372,7 @@ local font_proxy_mt = {
     end,
     isBitmap = function(self)
       return self._proxy:isBitmap()
-    end,
+    end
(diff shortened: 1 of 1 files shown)


TheCycoONE:
May want a new review; this is a larger fix to cover more cases and work around what appears to be a garbage collector bugs - the c++ destructor was being called while the font was still in the weak table (clearCache), and once I saw it try to draw to a freed font.

Since I now update the scale of the existing font instead of deleting references I haven't seen the issue.

PR #3198, opened 2026-01-14
Fix for weak key issue

### LINKED ISSUE #3196: [Bug] Lua use after finalize
### Describe the issue

The finalizer (__gc) metamethod may be called on an object before lua actually frees the memory for that object. More importantly weak tables with keys referencing these objects retain their keys until after the object memory is freed.

This combination means that the destructor in C++ can be called but the object is still alive and may have methods called on it.

We use tables with weak keys in a handful of places:

```
CorsixTH/Luatest/non_strict.lua
27:local allowed_globals = setmetatable({}, {__mode = "k"})

CorsixTH/Lua/strict.lua
29:local allowed_globals = setmetatable({}, {__mode = "k"})

CorsixTH/Lua/dialogs/edit_room.lua
289:  local humanoids_to_watch = setmetatable({}, {__mode = "k"})

CorsixTH/Lua/strings.lua
131:local shadows = setmetatable({}, {__mode = "k"})

CorsixTH/Lua/graphics.lua
75:    cursors = setmetatable({}, {__mode = "k"}),
81:  self.load_info = setmetatable({}, {__mode = "k"})
88:  self.reload_functions = setmetatable({}, {__mode = "k"})
91:  self.reload_functions_last = setmetatable({}, {__mode = "k"})
```

I observed this issue specifically with fonts and reload_functions_last which changing the scale factor #3192 - but it is potentially an issue in all of these places.

Consider avoiding weakly keyed tables or modifying the metadata in the finalizer to mark it as invalid and respecting that anywhere using weak keys. It would be better if the destructor wasn't called multiple times too.

Read carefully about the new garbage collection: https://www.lua.org/manual/5.5/manual.html#2.5

### Steps to Reproduce

1. Compile with ASAN to detect invalid access
2. Start the game using lua 5.5 on 2e841d4c
3. Open the options screen and change the scale / language repeatedly, drag the window size
4. Observe occasional errors

### Expected Behaviour

Never access an object that has been finalized

### CorsixTH Version

2e841d4c

### Operating System

ArchLinux

### Additional Information

The behavior is not actually new in Lua 5.5; lua has been allowed to call the finalizer multiple times since lua 5.3, and weak keys have been left after finalization since 5.2.

It's possible I only observed it now because of the amount of use with testing scaling, along with the santizers.

TheCycoONE:
Fixes #3196

Prevent weak keys with finalizers from being removed form tables while iterating. This was primarily a problem for the tables in graphics.lua

The fix causes any unrefererenced key to be removed before the iteration is done and disables garbage collection so no new keys can be finalized during iteration.

This PR also removes an unnecessary weak table on entities that is short lived.

As an advantage the game doesn't use a ton of memory and slow to a crawl when flipping back and forth between languages with this patch.

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -286,7 +286,7 @@ function UIEditRoom:clearArea()
   local rect = self.blueprint_rect
   local world = self.ui.app.world
   world:clearCaches() -- To invalidate idle tiles in case we need to move people
-  local humanoids_to_watch = setmetatable({}, {__mode = "k"})
+  local humanoids_to_watch = {}
   do
     local x1 = rect.x - 1
     local x2 = rect.x + rect.w


CorsixTH/Lua/utility.lua
```diff
@@ -458,3 +458,11 @@ function shallow_clone(tbl)
   setmetatable(target, meta)
   return target
 end
+
+function pause_gc_and_use_week_keys(fn, ...)
```
TheCycoONE: weak

PR #3199, opened 2026-01-17
Overcharge payment rejection improvement

ARGAMX:
screenshot from 2014-09-27 18 24 12 

The change makes the probability of patients refusing to pay be pretty similar to the original game.
By formula suggested here https://github.com/CorsixTH/CorsixTH/discussions/2428#discussioncomment-15526798

----

### Describe the issue

Currently payment refuse probability does not align with original game.

When the price of procedures increased by the player, the probability of patients refusing to pay does not align with the probability in the original game.

Read this topic for more details:
https://github.com/CorsixTH/CorsixTH/discussions/2428 

----

### Save Game

https://www.dropbox.com/sh/zfb57sosk92nbc9/AABcO6Q2Q70WDp1BORASEB9Sa/prices_impact?dl=0 

----

### CorsixTH Version

0.69.2

----

### Expected Behaviour

The probability should be similar to the original game.

----

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -321,13 +321,22 @@ end
 function Patient:agreesToPay(disease_id)
   local hosp = self.hospital
   local casebook = hosp.disease_casebook[disease_id]
-  local price_distortion = self:getPriceDistortion(casebook)
-  local is_over_priced = price_distortion > hosp.over_priced_threshold
-
-  if is_over_priced and math.random(1, 5) == 1 then return false end
-  self.pay_amount = hosp:getTreatmentPrice(disease_id)
-
-  return true
+  local agrees_to_pay = true
+  local price = casebook.price
+  local is_over_priced = price > 1.0
+  if is_over_priced then
+    local overprice_size = price - 1.0
+    -- payment chance modificator was chosen experimentally to match the chance in the original game.
+    -- check github issue 2428 for details.
+    local payment_chance_modificator = 4
+    local payment_chance = math.exp(-1 * payment_chance_modificator * overprice_size)
+    agrees_to_pay = math.random() <= payment_chance
+  end
+  if agrees_to_pay then
+    -- save the price the patient agrees to
+    self.pay_amount = hosp:getTreatmentPrice(disease_id)
+  end
+  return agrees_to_pay
 end
 
 --! Either the patient is cured, or he/she dies.
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -322,12 +322,11 @@ function Patient:agreesToPay(disease_id)
   local hosp = self.hospital
   local casebook = hosp.disease_casebook[disease_id]
   local agrees_to_pay = true
-  local price = casebook.price
-  local is_over_priced = price > 1.0
+  local price_multiplier = casebook.price
+  local is_over_priced = price_multiplier > 1.0
   if is_over_priced then
-    local overprice_size = price - 1.0
+    local overprice_size = price_multiplier - 1.0
     -- payment chance modificator was chosen experimentally to match the chance in the original game.
-    -- check github issue 2428 for details.
     local payment_chance_modificator = 4
     local payment_chance = math.exp(-1 * payment_chance_modificator * overprice_size)
     agrees_to_pay = math.random() <= payment_chance


CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -321,13 +321,22 @@ end
 function Patient:agreesToPay(disease_id)
   local hosp = self.hospital
   local casebook = hosp.disease_casebook[disease_id]
-  local price_distortion = self:getPriceDistortion(casebook)
-  local is_over_priced = price_distortion > hosp.over_priced_threshold
```
lewri, line 325: This is the only use of this variable `over_priced_threshold` in the code. You can clean up the reference in  hospital if you remove it here.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -321,13 +321,22 @@ end
...
-
-  if is_over_priced and math.random(1, 5) == 1 then return false end
-  self.pay_amount = hosp:getTreatmentPrice(disease_id)
-
-  return true
+  local agrees_to_pay = true
+  local price = casebook.price
+  local is_over_priced = price > 1.0
+  if is_over_priced then
+    local overprice_size = price - 1.0
```
lewri: Comment what `1.0` is, though is the `.0` necessary?

I'm worried that someone thought the value of `casebook.price` should be a multiplier was a good idea (not you, likely was TH)

Edit: It's meaning is rather obvious, but the confusion is we've labelled `price` against something that isn't actually a price if that makes sense.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -321,13 +321,22 @@ end
...
+    -- payment chance modificator was chosen experimentally to match the chance in the original game.
+    -- check github issue 2428 for details.
```
lewri: discussion 2428

However, it's likely the research knowledge should've been transferred to the wiki instead by now.

lewri (commented):
Please could we note what we're actually doing in this PR rather than relying solely on the related ticket? Summarising the outcome of the discussion likely is good enough, with referencing back to it for more depth.

CorsixTH/Lua/entities/humanoids/patient.lua
TheCycoONE: Referencing issue numbers in code has a way of going stale (e.g. github is the not the first source control CorsixTH has been hosted at).

Better to detail here.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: > I'm worried that someone thought the value of `casebook.price` should be a multiplier was a good idea (not you, likely was TH)
> 
> Edit: It's meaning is rather obvious, but the confusion is we've labelled `price` against something that isn't actually a price if that makes sense.

Yes, perhaps the name `price_multiplier` would have been more accurate.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: Nope, that's not the only use.

CorsixTH/Lua/entities/humanoids/patient.lua
ARGAMX: this problem resolved 👍

CorsixTH/Lua/entities/humanoids/patient.lua
lewri: Yeah I agree with that. Other parts can be tackled later.

lewri (approved):
Approved but please could you expand the `Describe what the proposed change does` part of the original post more.

ARGAMX:
> Approved but please could you expand the `Describe what the proposed change does` part of the original post more.

Thank you. Okay. I'll do it.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -321,13 +321,21 @@ end
...
-
-  return true
+  local agrees_to_pay = true
+  local price_multiplier = casebook.price
+  local is_over_priced = price_multiplier > 1.0
+  if is_over_priced then
+    local overprice_size = price_multiplier - 1.0
+    -- payment chance modificator was chosen experimentally to match the chance in the original game.
+    local payment_chance_modificator = 4
+    local payment_chance = math.exp(-1 * payment_chance_modificator * overprice_size)
```
TheCycoONE, line 331: e^(-4*[0.1-1])

Image

Is this correct; it looks harsher than I remember the original being, but I haven't tested. What were the experimental results?

TheCycoONE: I see in the ticket; very good.

Issue #3188, opened 2026-01-10
UI Scaling Feedback


lewri:
### This ticket is a meta for issues and possible enhancements discovered during the UI Scaling implementation. You may summarise the issue in a comment, but immediately after should be spun out to its own issue.
NB: Repo members, please attach spun out tickets as sub issues.
 

- The folder selection path font does not scale consistently. It did not originally scale at 200%, but did at 300% then back to 200%. May have also needed to load a game to apply the font scaling change? - #3192
<img width="523" height="401" alt="Image" src="https://github.com/user-attachments/assets/5114aecb-6fea-46fe-b383-0888f5c69eda" />

- More scaling stages would be beneficial (likely a longer term project goal).
- Cursor icon doesn't expand with scaling (see image above) - #3193
- Resolution options can be confusing with scaling options. Dependent on the scaling chosen, some resolutions disappear until scaling is changed back without reason (i.e. the dropdown may need an entry `Scaling applied, some resolutions unavailable`.) #3280 
- It may be useful to prompt for scaling at first runtime, as the initial navigation of scaling could be difficult for some users.

TheCycoONE:
Cursor was intentional. It can have it's own scale if desired if it still always points as expected (I'm not sure it will)

An earlier version did scale the cursor.

I think this ticket needs to be split. The only bug is the font size, and the others need discussion - except more scaling options, the challenge there is technical (and it may not look good)

ARGAMX:
<img width="122" height="197" alt="Image" src="https://github.com/user-attachments/assets/2877264c-5021-4324-bbd2-ef93ad37d439" />

- I also noticed that a gray border has also appeared around the buttons on the 200% scale on the fax window. - #3194 

- On English language the folder selection path is unaffected only until game restart or until scale changed again. On Unicode languages it always small on 200%. - #3192

- About disappearing resolution and scale options I also bit concerned. https://github.com/CorsixTH/CorsixTH/pull/3170#issuecomment-3696838843

TheCycoONE:
This is better suited as a discussion unless you have an actionable recommendation

TheCycoONE:
Please spin off individual bugs for the font and any visual artifacts.

TheCycoONE:
I tried to address the resolution surprise with a tooltip. Short of that I need suggestions not complaints.

ARGAMX:
I detected some deviatiatin in patients drag and drop to another rooms. 
[...]

PR #3206, opened 2026-01-26
Make UI Scale & Resolution option button aware

lewri:
*Fixes #*

- Disables the UI Scale button if no scale other than 100% can be selected
- Also updates the tooltip with a hint.

image \
(string has the word 'first' at the end, and replaces UI with 'user interface' at the start: this screenshot was before that).

Link into #3188

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/options.lua
+++ b/CorsixTH/Lua/dialogs/resizables/options.lua
@@ -339,7 +339,6 @@ function UIOptions:selectResolution(number)
       local err = {_S.errors.unavailable_screen_size}
       self.ui:addWindow(UIInformation(self.ui, err))
     end
-    self:processWindowResizeEvent()
   end
 
   if res.custom then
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -343,6 +345,7 @@ function UIOptions:selectResolution(number)
   else
     callback(res.width, res.height)
   end
+  self:updateUIScaleAvailabilityState()
```
TheCycoONE: This needs to be in the callback above, here it won't get called when the custom resolution dialog is used.

lewri:
Holes fixed. Also fixed it so that the resolution button updates on a resize event too (unrelated to UI Scaling)

CorsixTH/Lua/ui.lua
```diff
@@ -975,8 +975,12 @@ end
 --!param width (integer) New window width
 --!param height (integer) New window height
 function UI:onWindowResize(width, height)
-  if not self.app.config.fullscreen then
-    self:changeResolution(width, height)
+  if not self.app.config.fullscreen and self:changeResolution(width, height) then
+    -- Update the Options window, if open
+    local window = self:getWindow(UIOptions)
```
TheCycoONE: Just implement UIOptions:onChangeResolution and it will get called.

Issue #3211, opened 2026-01-31
Enable the game to automatically detect music folders


ShiroAka:
### What feature or enhancement would you like to add?

Enable the game to automatically detect a music folder named “Music” or “music” within either the CorsixTH installation directory or the configuration directory. This would remove the need for users to manually configure the folder path.

### How would this feature benefit the gameplay of CorsixTH?

This enhancement would improve the user experience by allowing players to simply place a music folder in the installation or configuration directories to enable custom music.
It eliminates the current requirement to configure the folder manually through the UI, which can feel slow and inconvenient.

### Was the feature in the original game?

No

### What version of CorsixTH are you using?

0.69.2

### Does this feature/enhancement relate to a current problem?

No

ShiroAka:
Already working on this. I've added this issue just for tracking purposes

Issue #2842, opened 2025-05-02
[Bug] Error text don't respect window width


ARGAMX:
### Describe the issue

Error text don't respect window width. 
This happens only on non-Unicode languages.

<img width="577" alt="Image" src="https://github.com/user-attachments/assets/c35c1edb-769c-4c34-9b72-df91cade948a" />

For any Unicode language this problem disappears.

<img width="330" alt="Screenshot 2025-05-01 at 22 28 56" src="https://github.com/user-attachments/assets/1eb811b6-28de-487d-a914-2ae0097f7777" />

### Steps to Reproduce

1. Download savegame attached to issue 873 or 1162. 
2. Place it your CorsixTH savegames folder.
3. Run CorsixTH v0.68.0 or build latest master.
4. Select in setting any non-Unicode language (English for example).
5. Load downloaded save game.
6. Observe error message on the screen where text going outside of the error window.

### Expected Behaviour

The text should wrap to a new line as for Unicode languages.

### Have you enabled any current CorsixTH Experimental Features?

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

Save game attached to report: Yes/No

### CorsixTH Version

0.68.0

### Operating System

macOS Sonoma 14.5

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

Alberth289346:
Please break the line at whitespace, that makes it easier to read.

lewri:
I can only assume `l_font_draw_wrapped` is not finding a suitable place to cut the string because there's no valid whitespace character. However, I'm not sure why unicode is fine with this but the spritesheet font it is not.

TheCycoONE:
The filename can't be split and is wider than the container. I guess at that point it becomes bugged.

ARGAMX:
This bug Regressed in 0.66.0. 
0.65.0 was handling this more successfully.

============================================================
v0.66.0
<img width="473" alt="Image" src="https://github.com/user-attachments/assets/432cd94f-3c8a-4034-af54-a5d399f95ace" />

============================================================
v0.65.0 
<img width="386" alt="Image" src="https://github.com/user-attachments/assets/ebeca07a-be6b-40a7-bec9-f16520d9f229" />

lewri:
#2045 might be related, even though it was originally targeted only to deal with unicode @TheCycoONE ?

We have found a few irregularities in font rendering over the last few years after that change.

PR #3212, opened 2026-01-31
Change saves & screenshots default location labels in settings

### LINKED ISSUE #3207: Finding the save game folder is too difficult
----

### Describe the issue

Finding the saved game folder isn't easy, especially on macOS.
The first time I encountered this in `0.65.0` and I had to google where I could find the save game folder.
The same problem with the `config.txt` file which is located in the same place as this folder.

----

### Expected Behaviour

Finding the saved game folder should be somehow simpler.
For example, when hovering over the `Saves` item in the `Folder Locations` section the tooltip could show the path to the saved game folder.
Currently, both the item and the tooltip say `Default Location`, which is hardly useful.

Image

----

### CorsixTH Version

0.69.2

----

### Operating System

Any

----

ShiroAka:
*Fixes #3207

- Changed the saves & screenshots "Default Location" labels so it shows the full path.

- Changed the reset location buttons to show the path to that default location without having to reset the config

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/folder_settings.lua
+++ b/CorsixTH/Lua/dialogs/resizables/folder_settings.lua
@@ -84,7 +84,7 @@ function UIFolder:UIFolder(ui, mode)
   self.fonts_panel = self:addBevelPanel(160, 75, 180, 20, col_bg)
     :setLabel(font_location_label, built_in)
     :setAutoClip(true)
-    :makeButton(0, 0, 160, 20, nil, self.buttonBrowseForFont)
+    :makeButton(0, 0, 180, 20, nil, self.buttonBrowseForFont)
     :setTooltip(tooltip_font)
 
   -- Location saves alternative


ShiroAka:
I'm currently seeing an issue with how the tooltips are rendered.

It seems the text doesn't get split in lines if there are no spaces.
image

I doubt it has to do with what I changed, but if it does, please let me know!

ARGAMX:
> I'm currently seeing an issue with how the tooltips are rendered.
> 
> It seems the text doesn't get split in lines if there are no spaces. image
> 
> I doubt it has to do with what I changed, but if it does, please let me know!

Don't worry, It is an old issue.
If there is no spaces, for some reason it unable to split text to new line.
I think it is related to #2842.

ShiroAka:
> > I'm currently seeing an issue with how the tooltips are rendered.
> > It seems the text doesn't get split in lines if there are no spaces. image
> > I doubt it has to do with what I changed, but if it does, please let me know!
> 
> Don't worry, It is an old issue. If there is no spaces, for some reason it unable to split text to new line. I think it is related to #2842.

I remember recently seeing something in the code that calculated the number of new lines after trying to split the spaces (it used a variable called "number_newlines" or something similar)

If I'm not mistaken, what I saw was related to the advisor bubble, but it may be also related to this (or both things share some code)

ARGAMX (commented):
Nice 👌

I have tested on macOS.

For saves and screenshots part works perfect 👍

Image

Then I have created `Music` folder in the same path and put some .mp3 here.
Music folder detects and generally works.

I tried naming the folder both way: `Music` and `music` and set `audio_music` to `nil` before attempts - the result is the same. 
The game displays the folder for both cases as lowercased `music`. For me It doesn't really matter, the main thing is that the folder is detected successfully in both uppercase and lowercase. But for some reason, I thought it would look different in the game, not just in `music` always, but depending on the real case. Overall, that is ok for me, just highlighting. But I would double check on Linux.

For some reason, in the tooltip, a line break appears at the end of the path before the folder name.
This looks like a defect?

Image
.

The `X` button says it will reset my path 
```
/Users/argus/.config/CorsixTH/music
```
to
```
/Users/argus/.config/CorsixTH

music
```
Which essentially the same?
But after resetting, the music stops and the player says that's it, there won't be music any more...
This looks like a defect?

Then I restart the game and it finds the folder again. And music play fine.
So, `X` is acting suspiciously.

At the same time in main field path showed as `/Users/argus/.config/CorsixTH//music`, not `/Users/argus/.config/CorsixTH/music` (double "//")

As you can see in the screenshot, the text in the main field has shifted upward to the right. This happens at both 100% and 200%. Could this also be due to the line break?
This looks like a defect?

I also noticed that the Music folder must not be empty. Otherwise, it can't find it. 
But that is ok for me, just highlighting.

By the way, I was thinking about to move this music autodetection part to another PR, this makes sense?
I like it, but seems tasks #3207 and #3211 is unrelated and so that change could be split into two different PRs?
It would be easier to track everything in the future if they weren't mixed up?

👍

tobylane:
// means new paragraph, you can see it used intentionally in the level introductions.

The check for music by extensions is too short. If it needs doing, reuse from Audio:init, https://github.com/CorsixTH/CorsixTH/blob/1a5f0d2f8d5fa3d3d512cd3d7de4c5503a368d4b/CorsixTH/Lua/audio.lua#L100

"Reset to default [dir name]" doesn't sound right for the tooltip, if it's found a dir, it only offers to reset to the same value. Clicking the X clears the value, as expected.

When it checks that the dir is already configured, check it exists as a readable dir ready for later use. If it's not, do you want to ignore or warn the user? If it finds a dir with the right name, but it's not readable, don't warn or save it.

CorsixTH/Lua/app.lua
```diff
@@ -514,6 +515,52 @@ function App:initScreenshotsDir()
...
+    return true
+  end
+
+  -- Checks if the specified directory exists and has music files
+  local function dirHasMusic(path)
+    if lfs.attributes(path, "mode") ~= "directory" then
+      return false
+    end
+    for file in lfs.dir(path) do
+      if file:match("%.ogg$") or file:match("%.wav$") or file:match("%.mp3$") then
```
lewri: This excludes a person who wants a music directory with MIDI files. In fact, we support many music formats, see audio.lua.

CorsixTH/Lua/app.lua
```diff
@@ -514,6 +515,52 @@ function App:initScreenshotsDir()
...
+      end
+    end
+    return nil
+  end
+
+  local corsixth_path = lfs.currentdir()
+  local conf_path = self.command_line["config-file"] or "config.txt"
+  conf_path = conf_path:match("^(.-)[^" .. pathsep .. "]*$")
+
+  -- Check corsixth's install directory first, then the config file directory
```
lewri: A final fallback should be the TH SOUND directory, if we're going to start setting this. The 'restore to default' behaviour, should also, imo, fall back to the TH SOUND directory (maybe it needs some more logic to decide between TH SOUND and the predefined music path for choice)

NB: This would require some shakeup in audio.lua; or some way to show the path but not actually set it as the config setting (the second option itself risks running the music folder init code though and disobeying a player choice to use TH).

There isn't a good answer here; it might just be enough to always assume the music folder presence means the player never wants TH midi; in which case, a lack of path should just show the TH Sound folder even though its not set (I sometimes dump my music to play inside the TH folder).

CorsixTH/Lua/app.lua
```diff
@@ -514,6 +515,52 @@ function App:initScreenshotsDir()
...
+  end
+
+  -- Checks if the specified path contains a music directory and if it has music files
+  -- If found, the directory's path is returned
+  local function findMusicDir(path)
+    local posible_dirs = {"music", "Music"}
+    for _, name in ipairs(posible_dirs) do
+      local candidate = path .. pathsep .. name
+      if dirHasMusic(candidate) then
+        return candidate
```
lewri: Very unlikely someone is going to have both a `music` and `Music` folder; this is good enough.

CorsixTH/Lua/dialogs/resizables/folder_settings.lua
```diff
@@ -83,31 +83,34 @@ function UIFolder:UIFolder(ui, mode)
   self:addBevelPanel(20, 100, 130, 20, col_shadow, col_bg, col_bg)
     :setLabel(_S.folders_window.savegames_label):setTooltip(_S.tooltip.folders_window.savegames_location)
     .lowered = true
-  local tooltip_saves = app.config.savegames and _S.tooltip.folders_window.browse_saves:format(app.config.savegames) or _S.tooltip.folders_window.default
+  local saves_location = app.config.savegames or app.savegame_dir
+  local tooltip_saves = saves_location and _S.tooltip.folders_window.browse_saves:format(saves_location)
   self.saves_panel = self:addBevelPanel(160, 100, 160, 20, col_bg)
-  self.saves_panel:setLabel(app.config.savegames and app.config.savegames or tooltip_saves , built_in):setAutoClip(true)
+  self.saves_panel:setLabel(saves_location , built_in):setAutoClip(true)
```
lewri: Because we're now showing default location, it might be a good idea to consider if we should expand the window's width.

lewri (changes requested):
I would've encouraged the music change to be its own PR, as both fixes are serving different goals, especially as we'll be scrutinising two different ideas. 

On the balance of a 'gotcha' with defaulting off the standard TH music if a folder exists, this is fine as long as it requires the user to explicitly go out their way to make the folder (because that implies they want it). However, to play devil's advocate, wouldn't a user also still just stick their music wherever they like? E.g. `C:\Users\Foo\Music\` (the default OS folder in Window).

CorsixTH/Lua/app.lua
lewri: Toby commented this an hour ago, but it hadn't loaded up yet in my review at the time :(

CorsixTH/Lua/app.lua
ShiroAka: I added both names because, according to the AI, the lfs.attributes is case sensitive in linux and mac

CorsixTH/Lua/app.lua
ShiroAka: Earlier I split the PRs.
I'll try to make the extensions list public (instead of local vars) in the audio.lua file and use those to match extensions

ShiroAka:
I still need to find a fix for the text going beyond the tooltip.
Any ideas would be welcome

I could try to force the paths to be split after a \ or /. 
But is the tooltip's width dynamic or fixed? If dynamic, how is it set?

CorsixTH/Lua/app.lua
TheCycoONE: Discussion point:

Do we even want to support non-game MIDI files? What's the use case?

None of the MIDI changes I made to use real devices will support them because I focused on our XMI files and the capabilities they needed - which is somewhat less than a fully functional general purpose midi player; particularly around tempo shifts.

I would strongly encourage anyone bringing their own music to use one of the modern audio formats.

CorsixTH/Lua/app.lua
TheCycoONE: The TH stuff is entirely special cased, it won't work as a detected folder. It uses it's own filesystem object so that it can access ISO images e.g.

CorsixTH/Lua/app.lua
lewri: Dropping support doesn't make sense if the functionality is already there and we don't get complaints, which would mean one of two scenarios:
1. People use it and it's 'good enough'
2. People don't use it.

I wouldn''t wish to alienate people of scenario 1 by depriving them of a method that works for them. However, we don't have any statistics to back up what scenario is true. (I also think this is meant to be in #3217 -- but I haven't cleaned up my own comments yet). A middle ground is updating the information about it, denoting for best experience you should use modern sound formats.

lewri: Moved over to 3217

CorsixTH/Lua/dialogs/resizables/folder_settings.lua
lewri: This earlier comment is up to your own interest if you wanted to implement it.

CorsixTH/Lua/languages/english.lua
```diff
@@ -628,7 +628,7 @@ tooltip.folders_window = {
   no_font_specified = "No font location specified yet!",
   not_specified = "No folder location specified yet!",
   default = "Default location",
-  reset_to_default = "Reset the directory to its default location",
+  reset_to_default = "Reset the directory to its default location ( %1% )",
```
lewri, line 631: This isn't a big stretch to edit the string of other languages to add the `( %1% )` if you want. A translator can move it to the start of the string later if necessary.

ShiroAka: It wouldn't bother me to manually add that to the existing lang files

As for the tooltip issue. I could pick that up next, no problem

TheCycoONE: Some folders don't have a default location (e.g. music and font), probably need two tooltips if you add that.

ARGAMX: Yeap, good point from Cyco.

Screenshot 2026-02-10 at 14 20 15

lewri:
It did just occur to me, that changing the save location will stop CorsixTH seeing any saves previously in the default location. There are probably swathes of potential 'fixes' to that, but the minimum might be just making the user aware. This is most probably out of scope.

ARGAMX:
Thanks for adding `( %1% )`in all languages 👍

Now for some reason when I press `X` opposite the music path, text appears on the screen outside the border.

Screenshot 2026-02-14 at 19 21 14

ARGAMX:
I also noticed that if I open this window with paths, then select a path, and then hover the cursor over the `X` button, then for some reason the tooltip says, what the currently set path is the default.

Screenshot 2026-02-14 at 19 26 46

ARGAMX:
Also if set any path for Music, then you can not reset it by pressing the `X` buttons.

ARGAMX:
The new text `clear_directory` been machine translated into all languages?

ShiroAka:
> The new text `clear_directory` been machine translated into all languages?

@ARGAMX  Yeah, I used AI

The issue with the reset button was because I didn't read the code well and thought the "app.savegame_dir" var only had the default dir. Sorry about that

MandrakeHorse:
> > The new text `clear_directory` been machine translated into all languages?
> 
> @ARGAMX Yeah, I used AI

I'm concerned about the use of machine translation for `clear_directory`. Until someone fluent in each language confirms, we won't be able to tell if any of the translations you've generated are accurate or makes sense. It might just be a simple tooltip, but we should strive to release CorsixTH with the best translation possible.

We do have a localisation project on Discord, where I can notify our volunteers that this pull request contains a new string that needs translating.

CorsixTH/Lua/dialogs/resizables/folder_settings.lua
```diff
@@ -64,79 +64,121 @@ function UIFolder:UIFolder(ui, mode)
...
   self:addBevelPanel(20, 75, 130, 20, col_shadow, col_bg, col_bg)
-    :setLabel(_S.folders_window.font_label):setTooltip(_S.tooltip.folders_window.font_location)
+    :setLabel(_S.folders_window.font_label)
+    :setTooltip(_S.tooltip.folders_window.font_location)
     .lowered = true
-  local tooltip_font = app.config.unicode_font and _S.tooltip.folders_window.browse_font:format(app.config.unicode_font) or _S.tooltip.folders_window.no_font_specified
-  self:addBevelPanel(160, 75, 180, 20, col_bg)
-    :setLabel(app.config.unicode_font and app.config.unicode_font or tooltip_font, built_in):setAutoClip(true)
-    :makeButton(0, 0, 180, 20, nil, self.buttonBrowseForFont):setTooltip(tooltip_font)
+  self.fonts_panel = self:addBevelPanel(160, 75, 160, 20, col_bg)
```
ARGAMX:
```suggestion
  self.fonts_panel = self:addBevelPanel(160, 75, 180, 20, col_bg)
```
to remove this new gap
Image

CorsixTH/Lua/dialogs/resizables/folder_settings.lua
```diff
@@ -64,79 +64,121 @@ function UIFolder:UIFolder(ui, mode)
...
+    :setLabel(font_location_label, built_in)
+    :setAutoClip(true)
+    :makeButton(0, 0, 160, 20, nil, self.buttonBrowseForFont)
```
ARGAMX:
```suggestion
    :makeButton(0, 0, 180, 20, nil, self.buttonBrowseForFont)
```
to remove this new gap
Image

PR #3213, opened 2026-01-31
Only advise about low drug prices if prices have been lowered

### LINKED ISSUE #3201: Prevent Low Prices adviser notification when treatment costs 100%

If the player sets their treatment prices to 100%, the advisor may occasionally tell them that "You're charging too little for %s. This will bring people to your hospital, but you won't make a lot of profit from each one."

https://github.com/CorsixTH/CorsixTH/blob/6005581722b895d835b60a26f1f16cedd07c9aae/CorsixTH/Lua/hospitals/player_hospital.lua#L558-L573

https://github.com/CorsixTH/CorsixTH/blob/6005581722b895d835b60a26f1f16cedd07c9aae/CorsixTH/Lua/hospital.lua#L2338-L2363

However, 100% is considered to be the ideal price to charge your patients. Indeed, once the revamped overcharging formula is released in version 0.70 (see https://github.com/CorsixTH/CorsixTH/pull/3199), patients will never refuse to pay if treatment is set to 100% or below, but become more likely to refuse to pay if it is set above 100%. By encouraging the player to raise prices beyond this safe level, they are likely to start seeing patients walking out, meaning that the adviser has pushed to player into sabotaging their own hospital.

It would make more sense for this notification to only appear if the player has set the treatment to 99% or below.

The adviser should only encourage the player to raise prices if they are charging 99% or below for a treatment.
The adviser should not encourage the player to raise prices above 100% for a treatment.

Yes

0.69.2

#3200

### Additional Information

_No response_

tobylane:
*Fixes #3201*

- Check that the drug price has been lowered before advising the player the price is too low.

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -2164,10 +2164,11 @@ end
 function Hospital:computePriceLevelImpact(patient, casebook)
   local price_distortion = patient:getPriceDistortion(casebook)
   patient:changeAttribute("happiness", -(price_distortion / 2))
-
   if price_distortion < self.under_priced_threshold then
     if math.random(1, 100) == 1 then
-      self:advisePriceLevelImpact("under", casebook.disease.name)
+      if casebook.price < 1 then
+        self:advisePriceLevelImpact("under", casebook.disease.name)
+      end
       self:changeReputation("under_priced")
     end
   elseif price_distortion > self.over_priced_threshold then


CorsixTH/Lua/hospital.lua
```diff
@@ -2164,10 +2164,11 @@ end
 function Hospital:computePriceLevelImpact(patient, casebook)
   local price_distortion = patient:getPriceDistortion(casebook)
   patient:changeAttribute("happiness", -(price_distortion / 2))
-
   if price_distortion < self.under_priced_threshold then
     if math.random(1, 100) == 1 then
```
lewri, line 2168: Wow, 1% chance 🙃

PR #3215, opened 2026-01-31
Close daughter UIOptions Windows when clicking outside boundary

### LINKED ISSUE #3205: [Bug] Custom Resolution window does not stay on top of options menu (and others)
### Describe the issue

When choosing a custom resolution, if you click outside of that window it gets sent backwards.

### Steps to Reproduce

1. Open game and navigate to options.

2.  Set a custom resolution

3. Click off of the custom resolution window.

### Expected Behaviour

Either keep the window in front; or close it if clicked outside of.

- [ ] Right mouse panning

### Save Game

### CorsixTH Version

605a8dcfd7ec95de670a6636ea94c1e126d05c9b

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml

```

### Additional Information

https://github.com/user-attachments/assets/0d0ea4e8-2d0e-4bc0-8ea3-f411af5cb574

lewri:
*Fixes #3205*

- Will close the daughter windows from UIOptions if they are clicked off of, treating it as a 'cancel' action.

There is an alternative solution, where `UIResizable` has a variable added called `clickoff_closes`; which instead rewrites UIResizable instead of this patch. (essentially `if self.clickoff_closes and not self:hitTest(x, y) then self:close() end` -- though some windows may do different things on cancel which wouldn't be accounted for here as easy).

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/options.lua
+++ b/CorsixTH/Lua/dialogs/resizables/options.lua
@@ -541,6 +541,13 @@ function UIResolution:ok()
   end
 end
 
+function UIResolution:onMouseDown(button, x, y)
+  if not self:hitTest(x, y) then
+    self:close(false)
+  end
+  UIResizable.onMouseDown(self, button, x, y)
+end
+
 --! Closes the resolution dialog
 --!param ok (boolean or nil) whether the resolution entry was confirmed (true) or aborted (false)
 function UIResolution:close(ok)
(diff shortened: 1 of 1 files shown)


TheCycoONE:
An alternate alternate is another dialog type that inherits UIResizable - e.g. Modal, that these inherit from.

PR #3217, opened 2026-01-31
Enable game to automatically detect music folders

### LINKED ISSUE #3211: Enable the game to automatically detect music folders

ShiroAka:
*Fixes #3211 *

Enable the game to autodetect music folders in either the config dir or CorsixTH's install dir

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -109,31 +109,6 @@ function App:init()
   print("")
 
   -- Prereq 1: C++ and lua sides are compatible
-  -- Note: These errors cannot be translated, as the config file specifies the language
-  local conf_path = self.command_line["config-file"] or "config.txt"
-  local conf_chunk, conf_err = loadfile_envcall(conf_path)
-  if not conf_chunk then
-    error("Unable to load the config file. Please ensure that CorsixTH " ..
-      "has permission to read/write " .. conf_path .. ", or use the " ..
-      "--config-file=filename command line option to specify a writable file. " ..
-      "For reference, the error loading the config file was: " .. conf_err)
-  else
-    conf_chunk(self.config)
-  end
-  self:fixConfig()
-  corsixth.require("filesystem")
-  local good_install_folder, error_message = self:checkInstallFolder()
-  self.good_install_folder = good_install_folder
-  self.level_dir = self:getFullPath("Levels", true)
-  self.campaign_dir = self:getFullPath("Campaigns", true)
-  self:initUserDirectories()
-  self:initSavegameDir()
-  self:initScreenshotsDir()
-
-  -- Create the window
-  if not SDL.init("video", "timer", "audio") then
-    return false, "Cannot initialise SDL"
-  end
   local compile_opts = TH.GetCompileOptions()
   local api_version = corsixth.require("api_version")
   if api_version ~= compile_opts.api_version then


ShiroAka:
Here is some test code I made to test the feature without having to run the game (I executed it with VSCode + Lua + Lua Debugger extensions, and by dropping the correct "lfs.dll" file next to the ".lua" file with the code)

To obtain the paths commented in the test code below, paste the following lines somewhere in the "app.lua" file and run the game:

<details>
  <summary> Code to obtain paths for test code </summary>
  
  local corsixth_path = lfs.currentdir()
  local th_install_path = self.config.theme_hospital_install
  local conf_path = self.command_line["config-file"] or lfs.currentdir()  -- Gets the path of the config file
  conf_path = conf_path:match("^(.-)[^" .. pathsep .. "]*$") -- Removes the config file name from the above path
  
  print(corsixth_path)
  print(th_install_folder)
  print(conf_path)

</details>

==========================

Here is the test code you can use to easily test this feature:

<details>
  <summary>Test Code</summary>
local allowed_waveform_formats = {
  "OGG", "OPUS", "FLAC", "WV", "WAV", "WAVE",
  "MPG", "MPEG", "MP3", "MAD", "AIFF", "AIFC", "AIF"
}

local allowed_instructional_formats = {
  "MID", "MIDI", "KAR", "669", "AMF", "AMS", "DBM",
  "DSM", "FAR", "GDM", "IT", "MED", "MDL", "MOD", "MOL", "MTM", "NST", "OKT", "PTM",
  "S3M", "STM", "ULT", "UMX", "WOW", "XM", "XMI"
}

local lfs = require("lfs")

function initMusicDir()

  local function stripTrailingSlash(path)
    -- Remove one or more trailing / or \ characters
    return path:gsub("[/\\]+$", "")
  end

  -- Checks if the given directory name is a valid music directory name
  local function isValidMusicDirName(name)
    local posible_dirs = {"music", "sound"}

    for _, valid_name in ipairs(posible_dirs) do
      if name == valid_name then
        return true
      end
    end
    return false
  end

  -- Checks if the specified directory has music files
  local function dirHasMusic(path)
    -- Build a lookup table of allowed extensions (waveform + instructional)
    local allowed = {}
    for _, extension in ipairs(allowed_waveform_formats) do
      allowed[extension:lower()] = true
    end
    for _, extension in ipairs(allowed_instructional_formats) do
      allowed[extension:lower()] = true
    end

    -- Scan directory for files with matching extensions
    for file in lfs.dir(path) do
      local ext = file:match("%.([^.]+)$") -- capture extension without dot
      if ext and allowed[ext:lower()] then
        return true
      end
    end

    return false
  end

  -- Checks if the specified path contains a music directory and if it has music files
  -- If found, the directory's path is returned
  local function findMusicDir(path)
    -- Used to guard against the case where the TH install path is not valid/set
    if path == nil then
      return nil
    end

    local normalized_path = stripTrailingSlash(path)

    -- Check every file and folder inside the given path
    for entry in lfs.dir(normalized_path) do
      local entry_path = normalized_path .. "/" .. entry
      local isDirectory = lfs.attributes(entry_path, "mode") == "directory"

      if isDirectory and isValidMusicDirName(entry:lower()) then
        if dirHasMusic(entry_path) then
          return entry_path
        end
      end
    end
    return nil
  end

  -- Print the following paths in the game code and paste them here for faster testing
  --local corsixth_path =
  --local conf_path =
  --local th_install_path =



  -- Check corsixth's install directory first, then the config file directory
  local music_dir = findMusicDir(corsixth_path) or findMusicDir(conf_path) or findMusicDir(th_install_path)
  if music_dir then
    print(music_dir)
  end

  return true
end

initMusicDir()
</details>

ShiroAka:
If what I did with the "conf_path" variable is right (replacing the "config.txt" section found in other parts of the "app.lua" file with "lfs.currentdir()"), later I may change that so the algorithm doesn't have to check the same path twice.

Currently the algorithm doesn't check nested folders, only the files inside the "music" or "sound" folders inside CorsixTH's dir, the config dir or TH's install dir (in that order).

If enough people tell me to include that, we'll have to think which folder to choose in case it finds more than one.
The way the algorithm is done, it'll chose the first it finds, but I'm not certain if the OS orders them alphabetically or it's just random

CorsixTH/Lua/audio.lua
```diff
@@ -41,6 +41,15 @@ function Audio:Audio(app)
   self.played_sound_callbacks = {}
   self.entities_waiting_for_sound_to_be_enabled = {}
   self.midi_player = nil
+  self.allowed_waveform_formats = {
+    "OGG", "OPUS", "FLAC", "WV", "WAV", "WAVE",
+    "MPG", "MPEG", "MP3", "MAD", "AIFF", "AIFC", "AIF"
+  }
+  self.allowed_instructional_formats = {
```
lewri, line 48: Pulling from other PR:


@TheCycoONE 
Discussion point:

Do we even want to support non-game MIDI files? What's the use case?

None of the MIDI changes I made to use real devices will support them because I focused on our XMI files and the capabilities they needed - which is somewhat less than a fully functional general purpose midi player; particularly around tempo shifts.

I would strongly encourage anyone bringing their own music to use one of the modern audio formats.


@lewri 
Dropping support doesn't make sense if the functionality is already there and we don't get complaints, which would mean one of two scenarios:
1. People use it and it's 'good enough'
 2. People don't use it.

I wouldn''t wish to alienate people of scenario 1 by depriving them of a method that works for them. However, we don't have any statistics to back up what scenario is true. (I also think this is meant to be in https://github.com/CorsixTH/CorsixTH/pull/3217 -- but I haven't cleaned up my own comments yet). A middle ground is updating the information about it, denoting for best experience you should use modern sound formats.

CorsixTH/Lua/app.lua
```diff
@@ -514,6 +515,94 @@ function App:initScreenshotsDir()
...
+function App:initMusicDir()
+  -- No need to set the music dir if it's already configured
+  if self.config.audio_music then
+    return true
+  end
+
+  local function stripTrailingSlash(path)
+    -- Remove one or more trailing / or \ characters
+    return path:gsub("[/\\]+$", "")
+  end
```
lewri: Such a function doesn't belong here, or should just be in the main function raw. Though I feel like we have other methods to remove the trailing slash.

CorsixTH/Lua/app.lua
```diff
@@ -514,6 +515,94 @@ function App:initScreenshotsDir()
...
+        if dirHasMusic(entry_path) then
+          return entry_path
+        end
+      end
+    end
+    return nil
+  end
+
+  local corsixth_path = lfs.currentdir()
+  local th_install_path = self.config.theme_hospital_install
```
lewri: From other PR:

@TheCycoONE 
The TH stuff is entirely special cased, it won't work as a detected folder. It uses it's own filesystem object so that it can access ISO images e.g.

lewri: This is correct, see audio.l.ua (now) Line 116.
My recommendation was just to 'pretend' that is the set music path.

This still doesn't quite solve the issue we run music directory lookup every launch though.

CorsixTH/Lua/app.lua
```diff
@@ -514,6 +515,94 @@ function App:initScreenshotsDir()
...
+  -- Checks if the specified directory has music files
+  local function dirHasMusic(path)
+    -- Build a lookup table of allowed extensions (waveform + instructional)
+    local allowed = {}
+    for _, extension in ipairs(self.audio.allowed_waveform_formats) do
+      allowed[extension:lower()] = true
+    end
+    for _, extension in ipairs(self.audio.allowed_instructional_formats) do
+      allowed[extension:lower()] = true
+    end
```
lewri: If you use a list_to_set (see utility.lua) then you can just lookup the extension in the set, if it exists it returns true. Your conversion of the extension would have to up `:upper` instead though.

CorsixTH/Lua/app.lua
ShiroAka: I thought about putting that in a utils file. I think there's one, but not sure if that would be the right place

CorsixTH/Lua/app.lua
ShiroAka: It wouldn't search on every launch if the "audio_music" setting is set

Maybe we could just tell set TH's sound folder as the default path and let the user change it if they desire.

Instead of looking through the TH's directory, I could change it to search CorsixTH's dir, the config one and if everything fails, set TH's sound/midi dir as the default

CorsixTH/Lua/audio.lua
ShiroAka: I like your idea more. Give the user the power to choose.
But specifically XMI files are rare IMO. I've only seen them in this game (maybe it was common when the game was released). I think nowadays people just use midi or modern formats (mp3, ogg, flac, etc)

CorsixTH/Lua/app.lua
lewri: A value of "TH" might work to stop checks? You'll have to update the detection in audio.lua though.

CorsixTH/Lua/audio.lua
TheCycoONE: XMI is extremely exotic, it was only ever shipped with about 20 games in the 90s and never in any other context. As far as I know Theme Hospital was the last game to ship with it. That's part of why there are only incomplete XMI players.

TheCycoONE: Scenario 1 has historically made no sense; anyone with a midi collection has it because they have midi equipment to make it sound good, and we have always used sdl_mixer's lackluster midi support which couldn't target the equipment. But fine, it costs nothing to keep as long as sdl_mixer maintains their midi support, and while they cut it back in sdl3 (no more native support) they still do support fluidsynth.

We'll revisit this in a few more years.

CorsixTH/Lua/app.lua
TheCycoONE: @ShiroAka based on your last comment you may have missed my point. The contents of the HOSPITAL directory do not have to be a real path at all. We support directly running CorsixTH with Theme Hospital image files.

Currently that is represented by the music directory being nil. Another token would do if everywhere that tests it is updated, but not a real path.

ShiroAka: Ok, I'll remove that part. I added it mainly because I remember reading in the other PR that someone liked dropping music files in TH's sound folder :P

To avoid lookup everytime the game starts, perhaps I could add another config that determines if the algorithm should run or not (will be set to 'true' on install/first run).
Initially I would hide it from the user, but I could also add a checkbox or button to force the detection the next time the game starts.

ShiroAka:
@ARGAMX where is your config.txt file located? I suspect that error may have to do with that

ARGAMX:
> @ARGAMX where is your config.txt file located? I suspect that error may have to do with that

Default macOS CorsixTH path:
`/Users/argus/.config/CorsixTH`

ARGAMX:
Changing folder does not doing anything. 
It always stick to folder that it detected (`/Users/argus/.config/CorsixTH/Music`).
I can't play music from any another folder on PC.
So I deleted `/Users/argus/.config/CorsixTH/Music`... and now game crashes on start.

Screenshot 2026-02-05 at 14 16 21

lewri:
> I am getting this error in this PR on run:
> 
> Image
> I am not sure why it's happen, because when I run app in terminal to debug, it never fail like that and works fine.
> 
> Through trial and error way, I discovered that if I add the condition `if not (normalized_path == "") then` to line 572 in `app.lua`, the game no longer crashes.
> 
> So I suspect that `local normalized_path = stripTrailingSlash(path)` for some reason gets "" for `normalized_path`.

The TH music directory, and the TH install park I suspect have some handling for a non-existent folder? It likely needs to call the find music directory stuff again.

CorsixTH/Lua/app.lua
TheCycoONE: I think they were inventing problems. This search is fine to run every start up as far as perf. In user experience - delete or move your Music folder if you don't want it feels acceptable.

That said we should be searching again if the path doesn't exist.

ARGAMX:
Music folder delete still crashing game for me.
Other things resolved. 👌

Added:
In fact, it even crash on this scenario before this PR too.
0.69.2 also doesn't pass this test and crashes.
So this issue is not the result of this PR. 👌

TheCycoONE:
I want to test this with an ISO if no one has done that yet

ShiroAka:
> I want to test this with an ISO if no one has done that yet

I could help with that if you want.
How does the game detect an ISO (I never knew you could do that until recently) and what exactly are you planning on testing (the code doesn't look in TH's files anymore)?

TheCycoONE:
Works well for me. I want the crash when the music folder is removed fixed, but that can be outside of this PR.

CorsixTH/Lua/app.lua
```diff
@@ -514,6 +515,83 @@ function App:initScreenshotsDir()
...
 
+function App:initMusicDir()
+  -- No need to set the music dir if it's already configured
+  if self.config.audio_music then
+    return
+  end
+
+  -- Checks if the given directory name is a valid music directory name
+  local function isValidMusicDirName(name)
+    local posible_dirs = {"music", "sound"}
```
TheCycoONE: Surely we can be prescriptive here and just accept music?

CorsixTH/Lua/app.lua
```diff
@@ -514,6 +515,83 @@ function App:initScreenshotsDir()
...
+  local conf_path = self.command_line["config-file"] or nil  -- Gets the path of the config file ('nil' if it's in CorsixTH's dir)
+  conf_path = conf_path:match("^(.-)[^" .. pathsep .. "]*$") -- Removes the config file name from the above path
+
+  if(operating_system == "macOS") then
+    -- On macOS skip checking the game's directory, since it's difficult to obtain and/or open
+    corsixth_path = nil
+  end
+
+  -- Check corsixth's install directory first, then the config file directory
+  local music_dir = findMusicDir(corsixth_path) or findMusicDir(conf_path)
```
TheCycoONE: What was the reasoning for the order?

CorsixTH/Lua/audio.lua
```diff
@@ -41,6 +41,15 @@ function Audio:Audio(app)
...
+    "MID", "MIDI", "KAR", "669", "AMF", "AMS", "DBM",
+    "DSM", "FAR", "GDM", "IT", "MED", "MDL", "MOD", "MOL", "MTM", "NST", "OKT", "PTM",
+    "S3M", "STM", "ULT", "UMX", "WOW", "XM", "XMI"
```
TheCycoONE, line 51: Not all of these will work on every platform. I'm not sure there's a great way to detect what will and won't ahead of time though, maybe it doesn't matter.

CorsixTH/Lua/audio.lua
TheCycoONE: XMI works nicely.

CorsixTH/Lua/app.lua
```diff
@@ -514,6 +515,83 @@ function App:initScreenshotsDir()
...
+  -- If found, the directory's path is returned
+  local function findMusicDir(path)
+    if path == nil or not isDirectory(path) then
+      return nil
+    end
+
+    local normalized_path = stripTrailingSlashes(path)
+
+    -- Check every file and folder inside the given path
+    for entry in lfs.dir(normalized_path) do
```
TheCycoONE, line 565: Why are we doing this instead of telling the user they have to put their music directly in Music?

It feels like you were trying to support the SOUND/MIDI of the original game?

If the user puts a couple different folders of music in their Music folder we end up picking one and that seems more unexpected than only supporting files in the root. Fewer files to stat too.

TheCycoONE:
Works good for me, my preference would be to keep it simpler. Only support music (case insensitive) as the folder name, no nesting, and a small list of places the folder can be in like you have.

I'd like to keep the option to use Sound in the future for replacement sound effects.

Oh, and fixing the crash if the folder is missing would be great, this code would be a good spot to do that (to a filesystem check on the configured audio directory and nil it out / search again if missing).

CorsixTH/Lua/app.lua
ShiroAka: Yeah, I missed that when removing the code to look in TH's sound folder.
Still, I could leave that list there in case someone wants to add other directories in the future, but may not be needed at all

CorsixTH/Lua/app.lua
ShiroAka: Just because I'm used to put the files in the game's folder, but I may flip that if you think it's best to look in the config folder first (especially if we tell the user somewhere that's the preferred location)

CorsixTH/Lua/audio.lua
ShiroAka: I left it like that just to be consistent with what's in the "audio.lua" code (I could also make those lists some sort of global constants that are available to all classes, so we improve the code a little bit).

If we find a way to easily detect which formats are usual in each OS, I could add something to return an appropiate list of formats to look for based on the OS and that could also be used in the audio class.

CorsixTH/Lua/app.lua
ShiroAka: The algorithm only does a shallow search inside the Music/Sound folder, so it doesn't matter if it has nested dirs inside (I thought about allowing it to search nested folders but that would complicate things a lot)

ShiroAka:
> Works good for me, my preference would be to keep it simpler. Only support music (case insensitive) as the folder name, no nesting, and a small list of places the folder can be in like you have.
> 
> I'd like to keep the option to use Sound in the future for replacement sound effects.
> 
> Oh, and fixing the crash if the folder is missing would be great, this code would be a good spot to do that (to a filesystem check on the configured audio directory and nil it out / search again if missing).

I didn't test if deleting the folder would cause a crash, but I added some code that should help prevent that.
Will test later and upload a fix if needed

CorsixTH/Lua/audio.lua
TheCycoONE: Out of scope, this is good for now.

CorsixTH/Lua/app.lua
TheCycoONE: This is good for now.

CorsixTH/Lua/app.lua
TheCycoONE: Prefer to drop it because I think we might use sound for something different some day.

CorsixTH/Lua/app.lua
TheCycoONE: Mostly curious, but yeah if it were up to me I'd swap them because the config is in the user's folder always and the game may be in a shared system folder (/usr/share/corsixth, C:/Program Files or /Applications)

CorsixTH/Lua/utility.lua
```diff
@@ -483,3 +485,53 @@ function pause_gc_and_use_weak_keys(fn, ...)
...
+--!param path (string) The path to strip.
+--!return (string) The path without trailing slashes.
+function stripTrailingSlashes(path)
+  -- Remove one or more trailing / or \ characters
+  return path:gsub("[/\\]+$", "")
+end
+
+--! Detects the operating system the game is running on.
+--!return (string) One of "Windows", "macOS", "Linux", "Unix-like", or "Unknown".
+function detectOS()
```
TheCycoONE: The OS is already available in the app compile options structure.

Drop this function and use that.

CorsixTH/Lua/app.lua
```diff
@@ -514,6 +515,83 @@ function App:initScreenshotsDir()
...
+    corsixth_path = nil
+  end
+
+  -- Check corsixth's install directory first, then the config file directory
+  local music_dir = findMusicDir(corsixth_path) or findMusicDir(conf_path)
+  if music_dir then
+    self.config.audio_music = music_dir
+  end
+
+  self:saveConfig()
```
lewri: I'm intrigued why this init saves the config here but not the other inits.

ShiroAka: Don't know why the others don't try to set the path in the config file if it's not set.

IMO the screenshots and saves path in the config file should be set from the begining (in case it doesn't already do that, which may be the case based on the first lines in each init function)

TheCycoONE: As part of the config cleanup and simplification in #3254 I intend to call saveConfig once unconditionally at the end of app:init() so it's not necessary to call it each place the config is updated/fixed. If this lands first I'll just remove it, in the mean time no harm.

ShiroAka: I commented that line and added a comment explaining why, just in case someone wonders why the config is not saved there

CorsixTH/Lua/app.lua
```diff
@@ -516,6 +517,83 @@ function App:initScreenshotsDir()
...
+    corsixth_path = nil
+  end
+
+  -- Check corsixth's install directory first, then the config file directory
+  local music_dir = findMusicDir(conf_path) or findMusicDir(corsixth_path)
+  if music_dir then
+    self.config.audio_music = music_dir
+  end
+
+  -- self:saveConfig()  --> Don't save the config here, it'll get saved later in the app initialization process
```
TheCycoONE, line 594: Just remove the line entirely

PR #3219, opened 2026-02-01
Update contributing.md guidance for project devs

lewri:
- Documents the PR configuration process now used to properly format any incoming patch.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,9 +50,8 @@ You can also submit a **Draft Pull Request** as an alternative to the
 **[WIP]** label.
 
 ### CorsixTH Developers: Configuring Pull Requests ###
-All pull requests in CorsixTH must have an appropriate PR label before merging
-(current Humbe project excepted). This ensures it goes into the current project.
-More than one PR label can be applied, if appropriate. Those in Draft/[WIP] do not
+All pull requests in CorsixTH must have an appropriate PR label before merging to
+master. This ensures it goes into the current project. Those in Draft/[WIP] do not
 need a label until past that stage.
 
 Available labels are:
@@ -60,8 +59,8 @@ Available labels are:
 | ----- | ----------- |
 | **PR:Bugfix** | Code that fixes something broken. |
 | **PR:DevTools/Repo** | Code that mainly serves the CorsixTH repository, developer assets, or compilation and CI tools. |
-| **PR:Feature** | Implementation of something brand new to CorsixTH. |
-| **PR:Improvement** | Improvements to current functionality of the code. |
+| **PR:Feature** | Implementation of something brand new to CorsixTH that did not exist before. Includes creating a missing feature of TH. |
+| **PR:Improvement** | Improvements to a feature or function that already exists. This includes changes to achieve closer matching behaviour with TH.
 | **PR:Refactor** | Altering the structure of existing code without necessarily changing desired functionality. |
 | **PR:Translation** | Contributions to CorsixTH's translations. |
 


CONTRIBUTING.md
```diff
@@ -49,8 +49,27 @@ Please make sure you label and update your Pull Request accordingly.
 You can also submit a **Draft Pull Request** as an alternative to the
 **[WIP]** label.
 
+### CorsixTH Developers: Configuring Pull Requests ###
+All pull requests in CorsixTH must have an appropriate PR label before merging
+(current Humbe project excepted). This ensures it goes into the current project.
```
TheCycoONE: Let's keep this more timeless and just say before merging into master.

CONTRIBUTING.md
```diff
@@ -49,8 +49,27 @@ Please make sure you label and update your Pull Request accordingly.
...
+More than one PR label can be applied, if appropriate. Those in Draft/[WIP] do not
```
TheCycoONE: I don't think we need to call this out.

CONTRIBUTING.md
```diff
@@ -49,8 +49,27 @@ Please make sure you label and update your Pull Request accordingly.
...
+need a label until past that stage.
 
-#### CONTRIBUTING CODE: ####
+Available labels are:
+| Label | Description |
+| ----- | ----------- |
+| **PR:Bugfix** | Code that fixes something broken. |
+| **PR:DevTools/Repo** | Code that mainly serves the CorsixTH repository, developer assets, or compilation and CI tools. |
+| **PR:Feature** | Implementation of something brand new to CorsixTH. |
```
TheCycoONE: Could use clarification, particularly between feature and improvement.

We may also want to point out that closer matching the functionality of Theme Hospital is an improvement not a bugfix, by way of example.

CONTRIBUTING.md
```diff
@@ -49,8 +49,27 @@ Please make sure you label and update your Pull Request accordingly.
...
+| **PR:Improvement** | Improvements to current functionality of the code. |
+| **PR:Refactor** | Altering the structure of existing code without necessarily changing desired functionality. |
+| **PR:Translation** | Contributions to CorsixTH's translations. |
+
+There is also the label `PR:need-rebase`, that developers should apply to any pull
```
TheCycoONE, line 67: Doesn't github generally already flag these?

lewri: It warns of conflicts in the patch but not at a glance in the pr menu.

It also doesn't warn if the tests were updated and the pr should be rerun against the new tests.

(Why this is a should more than a must)

CONTRIBUTING.md
ARGAMX: > Could use clarification, particularly between feature and improvement.
> 
> We may also want to point out that closer matching the functionality of Theme Hospital is an improvement not a bugfix, by way of example.

This is a very interesting topic, by the way.

Why do we even need to separate changes into **features** and **improvements**?

Also, the division between bugs and other changes is not entirely clear.

Why do we even need to separate changes into **bugs** and **improvements**?
What's the purpose of this division into two groups?
Maybe the answer to this question could help us understand how to proper separate them.

I've been thinking about this problem periodically lately, and frankly, I don't understand what a **bug** is and what an **improvement** is on this particular project.

To register something as a **bug**, you need a **reference**. Usually, this is a product specification.

But we don't have one.

I think this project is a clone of an existing game that's no longer in development. Therefore, its state doesn't change. And we could use this game as a **reference**.

And if we start from this, and the proposed change brings this product closer to the **reference**, then it means it's a bug fix and it was a **bug**.

If the proposed change doesn't bring the product closer to the **reference** then it's an **improvement**. (For example, external music functionality, music folders, default paths, etc.)

I like this logic because it gives a clear boolean answer: Yes or No.

However, this logic doesn't cover the development of functionality that in original game but hasn't yet been implemented: rat functionality, competing hospitals, etc., which would seem to fall into the "bugfix" category. Perhaps the **feature** should cover this category.

And this logic contradicts your statement that approaching the  **reference** isn't a bug fix, but an **improvement**.

Then I can say for sure that I don't understand how to evaluate what's a bug fix and what's an improvement. 🙂

Until there's a standard, **reference** , hypothetically, nothing can be a bug.

`It's Not a Bug, It's a Feature!`

Alberth289346 (approved):
Looks fine to me

PR #3221, opened 2026-02-05
Remember/select the casebook disease selection

### LINKED ISSUE #3220: Casebook does not keep the last position
Image

----

### Describe the issue

When you exit the Casebook (`F5`) in the original TH, it keeps which entry you're on in Casebook.
And when you open the Casebook again, you are on the entry you were when you exited.
In CorsixTH, it doesn't keep. You always have to start over from the bottom.

----

### Steps to Reproduce

1. Start a new game.
2. Press `F5` to open Casebook.
3. Scroll up a few entries.
4. Exit from the Casebook screen.
5. Press `F5` to open Casebook again.
6. Observe that you are on the last record. And not the one you were on when you exited.

----

### Expected Behaviour

Casebook should remember which entry you're on between opens.

----

### CorsixTH Version

0.68.0, 0.69.2

----

tobylane:
*Fixes #3220*

- Save any selection of a disease
- Return to that selection on opening, if appropriate

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
@@ -86,6 +86,8 @@ function UICasebook:UICasebook(ui, disease_selection)
 
   if disease_selection then
     self:selectDisease(disease_selection)
+  elseif self.ui.casebook_selected_disease then
+    self:selectDisease(self.ui.casebook_selected_disease)
   else
     self.selected_index = #self.names_sorted
     self.selected_disease = self.names_sorted[self.selected_index]
@@ -198,6 +200,7 @@ function UICasebook:updateIcons()
 
   self.ui:updateTooltip() -- for the case that mouse is hovering over icon while player scrolls through list with keys
   self.percentage_counter = 50
+  self.ui.casebook_selected_disease = disease
 end
 
 function UICasebook:draw(canvas, x, y)
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -193,9 +193,8 @@ function UI:UI(app, minimal)
   end
 
   self:setCursor(self.default_cursor)
-
-
   self:setupGlobalKeyHandlers()
+  self.casebook_selected_disease = false
 end
 
 function UI:runDebugScript()


MandrakeHorse:
Out of curiosity, will the saved selection persist if the player loads a different save file, or closes CorsixTH?

tobylane:
It persists only in that savegame, through relaunches.

CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
```diff
@@ -86,6 +86,8 @@ function UICasebook:UICasebook(ui, disease_selection)
 
   if disease_selection then
     self:selectDisease(disease_selection)
+  elseif self.ui.casebook_selected_disease then
```
lewri, line 89: We never use the `disease_selection` parameter in the window creation. Which may be irrelevant to the PR unless you wanted to use it.

CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
```diff
@@ -198,6 +200,7 @@ function UICasebook:updateIcons()
 
   self.ui:updateTooltip() -- for the case that mouse is hovering over icon while player scrolls through list with keys
   self.percentage_counter = 50
+  self.ui.casebook_selected_disease = disease
```
lewri, line 203: Is it worth defining this value as nil in `UI:UI` on level creation? It will float in UI without definition, though it's not like we don't do that anyway in the past...

lewri (approved):

CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
ARGAMX: I like not using this `disease_selection` and leaving it as is in this PR at the moment.

It's probably good that tobylane didn't use this.

ARGAMX (approved):

CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
tobylane: It comes from the patient dialog. https://github.com/CorsixTH/CorsixTH/blob/4c2f3eb6b8b8315ffdc87b3684a2ef95a3ea7a0e/CorsixTH/Lua/dialogs/patient.lua#L334

CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
tobylane: Yes, done.

CorsixTH/Lua/dialogs/fullscreen/drug_casebook.lua
lewri: Thanks, not sure how I missed it.

PR #3223, opened 2026-02-09
In the staff management dialog, order by chosen attributes

### LINKED ISSUE #3061: Sort by Morale/Tiredness/Skill on Staff Management panel

Add the ability to sort staff members on the Staff Management panel by Morale, Tiredness, and Skill by clicking the respective header (highlighted in the below screenshot).

- The first click would sort the column by descending (highest to lowest).
- The second click would sort the column by ascending (lowest to highest).
- The third click would reset the staff back to the original order (earliest hired to most recently hired)

Staff management panel with headers selected

The player would be able to identify their most unhappy, tired, and unskilled staff more easily.
The player would not need to scroll through pages to check on their staff's happiness and tiredness.

No

0.69.1

No

### Additional Information

_No response_

tobylane:
*Fixes #3061*

- Make the morale, tiredness, and skill headers buttons which order the staff list in descending, ascending order or default (hire order).

### DIFF
--- a/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
+++ b/CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
@@ -179,13 +179,38 @@ function UIStaffManagement:updateStaffList(staff_member_removed)
     Handyman = {},
     Receptionist = {},
   }
-  for _, staff in ipairs(hosp.staff) do
+  for i, staff in ipairs(hosp.staff) do
+    staff.hire_order = i
     local list = staff_members[staff.profile.humanoid_class]
     list[#list + 1] = staff
     -- The selected staff might have been moved because someone else was removed from the list.
     if selected_staff == staff then
       self.selected_staff = #list
     end
+  end
+  -- Sort staff tables by attribute and direction
+  for _, staff_tbl in pairs(staff_members) do
+  if self.list_direction == "up" then
+    if self.list_order == "hire" then
+      table.sort(staff_tbl, function(a, b) return a.hire_order > b.hire_order end)
+    elseif self.list_order == "morale" then
+      table.sort(staff_tbl, function(a, b) return a:getAttribute("happiness") > b:getAttribute("happiness") end)
+    elseif self.list_order == "tiredness" then
+      table.sort(staff_tbl, function(a, b) return a:getAttribute("fatigue") > b:getAttribute("fatigue") end)
+    elseif self.list_order == "skill" then
+      table.sort(staff_tbl, function(a, b) return a.profile.skill > b.profile.skill end)
+    end
+  else
+    if self.list_order == "hire" then
+      table.sort(staff_tbl, function(a, b) return a.hire_order < b.hire_order end)
+    elseif self.list_order == "morale" then
+      table.sort(staff_tbl, function(a, b) return a:getAttribute("happiness") < b:getAttribute("happiness") end)
+    elseif self.list_order == "tiredness" then
+      table.sort(staff_tbl, function(a, b) return a:getAttribute("fatigue") < b:getAttribute("fatigue") end)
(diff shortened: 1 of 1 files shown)


ARGAMX (commented):
This feature is useful, and I've been missing it! 👍

I set this PR for myself and tried it out.
It seems to sort well! 👍

But overall, I think it's still a bit rough around the edges, and I'm not happy with the UI/UX:

1. The header hasn't changed visually at all; there's no indication that you can click on it now. Users know there's no sorting here. Why would someone who knows there's nothing here click here?
2. After clicking, the currently active sorting mode isn't displayed. I clicked around and realised there are 3 modes. But until you hire enough staff, it is hard to get the difference between the modes.
3. There's no any tooltip on hover either.
4. There's no click sound when you click. Matroftt recently added the missing sounds to some clicks.
5. Wouldn't it be great if the mode persisted after closing? Okay, am I being picky? 😀

I think that in its current state, hardly any users will notice this change and will use it.
We, the developers, will use it ourselves, and users who don't read the changelog very carefully will barely even know about it. And that's the majority of players.

Thank you! 👍

Screenshot 2026-02-10 at 13 41 34

CorsixTH/Lua/dialogs/fullscreen/staff_management.lua
```diff
@@ -179,14 +179,39 @@ function UIStaffManagement:updateStaffList(staff_member_removed)
...
+      table.sort(staff_tbl, function(a, b) return a.hire_order < b.hire_order end)
+    elseif self.list_order == "morale" then
+      table.sort(staff_tbl, function(a, b) return a:getAttribute("happiness") < b:getAttribute("happiness") end)
+    elseif self.list_order == "tiredness" then
+      table.sort(staff_tbl, function(a, b) return a:getAttribute("fatigue") < b:getAttribute("fatigue") end)
+    elseif self.list_order == "skill" then
+      table.sort(staff_tbl, function(a, b) return a.profile.skill < b.profile.skill end)
+    end
+  end
+  end
```
ARGAMX, line 207: I think there is no needs to copy-past block with `table.sort` calls just to change `>` to `<`.
We can just swap `a` and `b`. Like that:

```suggestion
  for _, staff_tbl in pairs(staff_members) do
    table.sort(staff_tbl, function(left, right)
      local ascending = self.list_direction == "up"
      local a = ascending and left or right
      local b = not ascending and left or right
      if self.list_order == "hire" then
        return a.hire_order > b.hire_order
      elseif self.list_order == "morale" then
        return a:getAttribute("happiness") > b:getAttribute("happiness")
      elseif self.list_order == "tiredness" then
        return a:getAttribute("fatigue") > b:getAttribute("fatigue")
      elseif self.list_order == "skill" then
        return a.profile.skill > b.profile.skill
      end
    end)
  end
```

also there is missing indent for nested `if self.list_direction == "up" then`

tobylane: That's much better, thank you.

tobylane:
1 and 2 Not changed, still wondering. If there's a suitable pair of icons to stick after the right header, how to represent hire order?
3 and 4 Added, the tooltip says the order and direction
5 As with the casebook, it's doable. I'll try it out.

lewri:
Font00 has some arrows coded into it if you were able to somehow pluck them and force them to show in bitmap...

image

TheCycoONE:
Those symbols are in the slot for greater than > and less than <; they are not provided in other fonts, and they are not arrows in the Russian version of the game, and I suspect not in the other regional versions either which I'd like to support (Swedish, Chinese, Japanese, etc.)

Much preferred that a sprite is found or a bitmap made.

TheCycoONE (approved):

TheCycoONE:
Thanks @tobylane I like it.

PR #3226, opened 2026-02-12
Setting autosave frequency

### LINKED ISSUE #1426: Question: Any easy way to shorten the autosave period?
Hi,
Is there some relatively easy way of shortening the autosave period?

Why I ask: This would make it easier when trying to reproduce a crash and to generate a save game just before. If possible I would set it to autosave every 1 or 2 minutes.

### LINKED ISSUE #3107: Autosave every game day
----

Option to make autosave every game day, not just 1 save per month.

Image

----

First of all, this function is very useful for debugging. But not only.
The current state of the project still requires a lot of debugging, observation and experimentation.
This feature would be useful not just for developers.

I've been using it for more than half a year.
This allowed me to get a lot of useful saves that allowed me to reproduce the defects.
Some of my discoveries and investigations were only possible thanks to this feature.
From the latest is 3056 issue.

At first, I used this functionality exclusively for myself.
But I gave this code to user Mickman, and during his play he provided a lot of save files with various obscure situations and bugs thanks to this. As example 3051 issue.
That's when I realized that this feature should be available for other developers and players, and not just for me.

So this feature can make players a more effective source of information for reproducing bugs.
It's probably even strange that no one has added this feature before.
Also possibly #2925 can benefit from this.

I didn't notice any significant negative effects from this feature. 
The game didn't run any slower.
The Autosave folder size increased to 500 MB.
This is most likely a consequence of the fact that the game saves are stored in an uncompressed state.

I also considered autosaving with a configurable interval, but that seemed too complicated.
So for now I've settled on two options: saving once a month (like now) and saving once a day.

----

### CorsixTH Version

0.69.1

----

ARGAMX:
*Fixes #3107*
*Fixes #1426*

- The following approach has been implemented: https://github.com/CorsixTH/CorsixTH/issues/3107#issuecomment-3810384695

Some things that deviated from that course:
- Weekly - Autosave once per week on the 1st, ~~8th, 15th, 22nd, and 29th~~ 7th, 14th, 21nd, and 28th of each month. 
- This setting hosted not in `Settings -> Customise` but in `Settings`.

Screenshot 2026-02-12 at 12 10 50

Screenshot 2026-02-12 at 12 11 04

- The rest is as planned in the message.

### DIFF
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1074,6 +1074,11 @@ function World:onEndDay()
       -- Weekly autosave
       self.autosave_next_tick = true
     end
+  elseif self.app.config.autosave_frequency == 1 then
+    if day == self.game_date:lastDayOfMonth() then
+      -- Monthly autosave
+      self.autosave_next_tick = true
+    end
   end
 end
 
@@ -1116,12 +1121,6 @@ function World:onEndMonth()
     end
   end
   self.current_tick_entity = nil
-
-  -- Autosave
-  if self.app.config.autosave_frequency == 1 then
-    -- Monthly autosave
-    self.autosave_next_tick = true
-  end
 end
 
 -- Called when a month ends. Decides on which dates patients arrive


CorsixTH/Lua/config_finder.lua
```diff
@@ -420,7 +421,14 @@ local string_01 = [=[
 -- By default subtitles are not displayed. If you would like the game to
 -- display subtitles for your hospital's announcements, turn this option on.
 --]=] .. '\n' ..
-'enable_announcer_subtitles = ' .. tostring(config_values.enable_announcer_subtitles) .. '\n' .. [=[]=]
+'enable_announcer_subtitles = ' .. tostring(config_values.enable_announcer_subtitles) .. '\n' .. [=[
+
+-------------------------------------------------------------------------------
+-- By default, the game autosaves every in-game month. If you would like to
+-- autosave more ofter, every week or every day, change this setting to 2 or 3.
+-- Set 1 for Monthly, 2 for Weekly, 3 for Daily autosaves.
```
lewri, line 431: Recommend adding the daily usage size warning in here too.

CorsixTH/Lua/languages/english.lua
```diff
@@ -490,11 +490,24 @@ options_window = {
   scrollspeed = "Scroll Speed",
   shift_scrollspeed = "Shift Scroll Speed",
   zoom_speed = "Zoom Speed",
+  autosave_periodicity = "Autosave Periodicity",
```
lewri: `Autosave Frequency` or `Autosave Period` is probably a bit more universally language friendly?

CorsixTH/Lua/languages/english.lua
```diff
@@ -490,11 +490,24 @@ options_window = {
...
+autosave_periodicity = {
+    monthly = "Monthly",
+    weekly = "Weekly",
+    daily = "Daily",
+}
+
+tooltip.autosave_periodicity = {
+    monthly = "Autosave once per month, on the first day of each month. A total of 12 saves with a depth of 1 year.",
+    weekly = "Autosave once per week on the 1st, 7th, 14th, 21nd, and 28th of each month. A total of 60 saves with a depth of 1 year.",
+    daily = "Autosave once per day at the start of each in-game day. A total of 365 saves with a depth of 1 year. Please note that a typical save can take up to 1 megabyte or even more. This way your autosaves folder can grow to 300-500 MB.",
```
lewri, line 508: I would've liked its own warning message. Only thing to bear in mind is this tooltip will be long.

CorsixTH/Lua/world.lua
```diff
@@ -1055,6 +1064,17 @@ function World:onEndDay()
       self.spawn_hours[hour] = self.spawn_hours[hour] and self.spawn_hours[hour] + 1 or 1
     end
   end
+
+  -- Autosave
+  if self.app.config.autosave_periodicity == 3 then
+    -- Daily autosave
+    self.autosave_next_tick = true
+  elseif self.app.config.autosave_periodicity == 2 then
+    if day == self.game_date:lastDayOfMonth() or day == 6 or day == 13 or day == 20 or day == 27 then
```
lewri, line 1073: Double check dates in the summary, last date for month says 25, this says 27?

CorsixTH/Lua/world.lua
```diff
@@ -1055,6 +1064,17 @@ function World:onEndDay()
       self.spawn_hours[hour] = self.spawn_hours[hour] and self.spawn_hours[hour] + 1 or 1
     end
   end
+
+  -- Autosave
+  if self.app.config.autosave_periodicity == 3 then
+    -- Daily autosave
+    self.autosave_next_tick = true
```
lewri, line 1071: I assume we don't try to fire 2 autosaves on daily for the end of the month?

CorsixTH/Lua/languages/english.lua
MandrakeHorse: I personally like "Autosave Frequency"

CorsixTH/Lua/languages/english.lua
MandrakeHorse: When the user toggles Alien Patients and Fractured Bones in the Custom Settings, a popup appears to warn the user of the effects of changing each setting. Can we add something similar when the user selects Daily?

CorsixTH/Lua/world.lua
ARGAMX: > I assume we don't try to fire 2 autosaves on daily for the end of the month?

This shouldn't happen.
At the end of the day or month, the `autosave_next_tick` flag may be set.
Even if this happened twice on the last day of the month, and the flag was set twice, the autosave would still only happen once per tick, after which the flag is reset.
Thanks for your concern about this.

CorsixTH/Lua/world.lua
ARGAMX: > Double check dates in the summary, last date for month says 25, this says 27?

Indeed, I made a typo in the summary.
25 is an incorrect value.
On the 27th day end we raise the flag so that an autosave occurs on the 28th day tick.
I corrected the summary.
Thanks for noticing.

CorsixTH/Lua/languages/english.lua
ARGAMX: I agree, `Autosave Periodicity` probably doesn't sound nice.
I renamed all feature to `Autosave Frequency`. 👍

CorsixTH/Lua/config_finder.lua
ARGAMX: > Recommend adding the daily usage size warning in here too.

Sounds reasonable.
Added warning here. 👌

CorsixTH/Lua/languages/english.lua
ARGAMX: I'd like users to enable this option to save every day, unlike the Alien Patients.
But that window will scare them too much.
Still, 300 MB isn't much in 2026. 
On macOS, CorsixTH takes up that much.

CorsixTH/Lua/config_finder.lua
```diff
@@ -420,7 +421,16 @@ local string_01 = [=[
 -- By default subtitles are not displayed. If you would like the game to
 -- display subtitles for your hospital's announcements, turn this option on.
 --]=] .. '\n' ..
-'enable_announcer_subtitles = ' .. tostring(config_values.enable_announcer_subtitles) .. '\n' .. [=[]=]
+'enable_announcer_subtitles = ' .. tostring(config_values.enable_announcer_subtitles) .. '\n' .. [=[
+
+-------------------------------------------------------------------------------
+-- By default, the game autosaves every in-game month. If you would like to
+-- autosave more ofter, every week or every day, change this setting to 2 or 3.
```
TheCycoONE: Typo: ofter -> often

CorsixTH/Lua/world.lua
```diff
@@ -948,6 +937,26 @@ function World:onTick()
   self.tick_timer = self.tick_timer - 1
 end
 
+--! Creates a save file in the "Autosaves" folder, using the "Autosave-mm-dd.sav" name template for filename.
+-- If the "Autosaves" folder does not exist, folder will be created.
+function World:executeAutosave()
```
TheCycoONE: `_` prefix for private functions. I think this wouldn't be called outside of World.

ARGAMX: > `_` prefix for private functions. I think this wouldn't be called outside of World.

👌

CorsixTH/Lua/config_finder.lua
ARGAMX: > Typo: ofter -> often

👌

CorsixTH/Lua/world.lua
TheCycoONE: It still might be cleaner to move the remaining case with the other two.

Nothing that blocks merging, just an opinion.

ARGAMX: > It still might be cleaner to move the remaining case with the other two.
> 
> Nothing that blocks merging, just an opinion.

Agree, it's a good idea. For some reason, I didn't think of that.
Moved 👌

TheCycoONE (approved):

PR #3228, opened 2026-02-13
Target start menu and desktop icons for all users

### LINKED ISSUE #3227: Windows Installer: Start menu items write to current user location only
Found this out while tinkering around with CorsixTH. The start menu shortcuts write to the Roaming AppData folder; whereas they should be ideally writing to ProgramData instead. The downside of this is that any other user who wants to play CorsixTH on a Windows device won't have the start menu entries.

CorsixTH
Image

Another application:
Image

I'm not sure about cleaning up the old roaming appdata entries though if this is changed.

lewri:
*Fixes #3227*

- Start menu items are written to `C:\ProgramData\Microsoft\Windows\Start Menu`; 
- CorsixTH AppData shortcut is now install aware and won't be made if they chose not to have their saves there.
A possible future TODO is to let this be customisable in the installer between 'Install for everyone' and 'Install for me' (Install for me is by design meant to install CorsixTH to `%APPDATA%` rather than Program Files.

I need to run checks on what this does to the installer via actions to make sure it works as intended.

### DIFF
--- a/WindowsInstaller/Win32Script.nsi
+++ b/WindowsInstaller/Win32Script.nsi
@@ -53,9 +53,6 @@ InstallDir "$PROGRAMFILES\${PRODUCT_NAME}"
 ; This is needed to be able to install into the program files directory
 RequestExecutionLevel admin
 
-; Also use the all users scope exclusively
-SetShellVarContext all
-
 ; Set installer compression
 SetCompressor lzma
 SetCompressorDictSize 16
@@ -227,6 +224,17 @@ Section "MainSection" SEC01
     File /nonfatal /r x86\*.*
   ${EndIf}
 
+  ; Remove old shortcuts (old installs wrote to APPDATA instead of ProgramData)
+  ; Non-existent files are ignored
+  SetShellVarContext all
+  Delete "$SMPROGRAMS\${PRODUCT_NAME}\*.*"
+
+  RMDir /r "$SMPROGRAMS\${PRODUCT_NAME}"
+  SetShellVarContext current
+  Delete "$SMPROGRAMS\${PRODUCT_NAME}\*.*"
+
+  RMDir /r "$SMPROGRAMS\${PRODUCT_NAME}"
+  
   ; Time to make the configuration file and Saves folder at the correct location
   ${If} $CONFIGAPPDATA == ${BST_CHECKED}
     SetOutPath "$APPDATA\CorsixTH"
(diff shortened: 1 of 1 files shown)


WindowsInstaller/Win32Script.nsi
```diff
@@ -53,6 +53,9 @@ InstallDir "$PROGRAMFILES\${PRODUCT_NAME}"
 ; This is needed to be able to install into the program files directory
 RequestExecutionLevel admin
 
+; Also use the all users scope exclusively
+SetShellVarContext all
```
lewri: @TheCycoONE I think I read this may also need to be additionally declared in the uninstall context, but from the way this file is structured it looks like declarations made this high up apply to both installer and uninstaller?

ARGAMX:
Again, the exhausted `DevTools/Repo` label that is used everywhere where others do no fit. 😁

lewri:
The installer is here for anyone who can test before I can analyse it; which won't be until Monday.
https://github.com/CorsixTH/CorsixTH/actions/runs/21991760434

(Maybe use a clean environment)

TheCycoONE:
I'm not in a good position to test this but it looks good

lewri:
> I'm not in a good position to test this but it looks good

The old install's shortcuts still remain at the moment, which leads to a bit of duplication.
Windows also hides the uninstall shortcut by default unless you change uninstall to something else. But modern windows you can right click installations anyway and choose uninstall.
So it's rename it or drop it, thoughts?

Plus, the commit actually needs updating to what the nsis-testbed branch actually does.

lewri:
Do we want the 'Uninstall CorsixTH' dropped @TheCycoONE if it will no longer show anymore with these changes? Or if you want it, we need to rename it 'Remove CorsixTH'.
Or I can leave the shortcut there, as it might still appear in some Windows setups.

TheCycoONE:
I'm curious how the mechanism works in modern windows; if it hides the uninstall and offers the right click option my intuition would be that it's using the hidden program to do it. These aren't MSI installs so it would have no way to track what is part of the install on it's own.

TheCycoONE:
I looked into it, and it's as I thought, the right click uninstall option in Windows is just hiding a call to uninstall.exe - it's still critical to have it.

TheCycoONE:
> Do we want the 'Uninstall CorsixTH' dropped @TheCycoONE if it will no longer show anymore with these changes? Or if you want it, we need to rename it 'Remove CorsixTH'. Or I can leave the shortcut there, as it might still appear in some Windows setups.

Based on what I've read, leave it, and it will be available for people that need it, e.g. windows 7.

TheCycoONE:
As for the other issue, we can notify people they should uninstall first. It wouldn't be the first time.

PR #3233, opened 2026-02-16
Ask (by default) at least 4 patients before computing an average patient value

Alberth289346:
- Require at least a handful of patients for getting an average patient property (warmth, thirst, happiness).
- When there are not enough patients in the hospital, the default value is returned.

This avoids getting adviser messages about the above properties when the hospital is practically empty.

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -1748,31 +1748,17 @@ function Hospital:updatePercentages()
   self.percentage_cured = math.round(cured)
 end
 
---! Compute average of an attribute for all patients in the hospital. There must be
---! enough patients available to get a useful value.
+--! Compute average of an attribute for all patients in the hospital.
 --!param attribute (str) Name of the attribute.
 --!param default_value Value to return if there are not enough patients.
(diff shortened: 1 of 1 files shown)


ARGAMX (approved):
👍

TheCycoONE:
This function is used for several more places than advisor messages (endconditions, progress report chart, vip rating), and in the early levels it's rare to have that many patients in a hospital.

Maybe needs to be more conservative?

lewri:
> This function is used for several more places than advisor messages (endconditions, progress report chart, vip rating), and in the early levels it's rare to have that many patients in a hospital.
> 
> Maybe needs to be more conservative?

Like 4? Though I wonder if the changes to how we allocate patients to hospitals might cause some further issues meeting allocation threshold.

Alberth289346:
The trigger for this was playing level 1, and getting reports about missing drinks machines while there was 1 patient and several placed drinks machines.

They are not used in an empty hospital.

Modified the patch:
- Reduce minimum count to 4 patients.
- Avoid large jumps in the value when the humanoid count is around the limit.
- Generalize the method for handling staff as well (without compensating for lack of staff at all).

TheCycoONE (approved):

PR #3234, opened 2026-02-17
Fix save/load issues with palettes

TheCycoONE:
When removing from load_info they needed to be otherwise blocked from saving. Also when loading an older save game it expected the palette to be returned by the permanent function (load_info)

Sorry about that.


lewri:
Not checked but I assume making the palette names case insensitive would've been the other way to go?

TheCycoONE:
Yeah, I didn't want to add the overhead of lower on every call, and got most of the way through changing everywhere to call with lowercase names before deciding it was an overly big change for the size of the issue. One afterLoad like case seemed better.

ARGAMX (approved):
👍

PR #3235, opened 2026-02-17
Add missing staff markers for handyman repair animations

Alberth289346:
Fixes handyman staff markers for the inflator, x-ray and slicer machine repair animations.

### DIFF
--- a/CorsixTH/Lua/objects/machines/x_ray.lua
+++ b/CorsixTH/Lua/objects/machines/x_ray.lua
@@ -99,7 +99,6 @@ anim_mgr:setPatientMarker(object.idle_animations.north, {-2, -2})
 
 local kf0 = {59, -4, "px"}
 local kf1 = {55, -8, "px"}
-local kf2 = {54, -8, "px"}
 local kf2 = {51, -8, "px"}
 local kf3 = {47, -11, "px"}
 local kf4 = {43, -12, "px"}


lewri:
#2794

CorsixTH/Lua/objects/machines/x_ray.lua
```diff
@@ -97,4 +97,26 @@ object.orientations = {
 local anim_mgr = TheApp.animation_manager
 anim_mgr:setPatientMarker(object.idle_animations.north, {-2, -2})
 
+local kf0 = {59, -4, "px"}
```
lewri, line 100: Was here no stop points in the motion of this animation? I'm guessing none of it is smooth, hence a definition every frame?

Alberth289346: `kf14` is the last frame with movement iirc. After that, the handyman starts actual 'repairs' while standing.

I added positions for each frame with a new position, in all animations where I added marker information. This one is not different from others in that respect.

I guess interpolation would work too, but I didn't try that.

Alberth289346:
FYI I checked all machines in level 4 of the game, and with this patch they now all seem to have handyman repair markers.

CorsixTH/Lua/objects/machines/x_ray.lua
```diff
@@ -97,4 +97,26 @@ object.orientations = {
...
+local kf1 = {55, -8, "px"}
+local kf2 = {54, -8, "px"}
```
Alberth289346:
```suggestion
```

CorsixTH/Lua/objects/machines/x_ray.lua
lewri: Mm, interpolation is handy :D Either way this can be merged imo

PR #3236, opened 2026-02-19
Support for official Russian Theme Hospital files

### LINKED ISSUE #2066: [Bug] Hans with localized LANG-0.DAT RUS
#### Steps to Reproduce
1. Use data file LANG-0.DAT of the edition localized to Russian (official).
2. Programs hangs after showing CD with doctors image.

#### Expected Behavior
Starts as usual

#### System Information
**CorsixTH Version:** 
0.65.1

**Operating System:** 
Windows 10 x64 21H1 19043.1237

**Theme Hospital Version:*
CD image of Russian localization.
Can be downloaded here: https://www.old-games.ru/game/download/get.php?fileid=26584&modal=1

#### Gamelog.txt
Gamelog not created for this issue

TheCycoONE:
Adds support for loading the LANG-0.DAT and using the font files from the official Russian Theme Hospital install.

As a consequence with this version the list of languages that can be supported without a unicode font is reduced to Russian only.

Fixes #2066 

There are some additional niceties that are not completed in this PR but would be nice to follow up with:

strings.lua has a custom tolower/toupper implementation that is based around CP437 / does not support Russian. There seems to be no reason for strings.lua to have any awareness of cp437 at all; it is always managed in c++. This isn't new to this patch, but I noticed it while working on this.

The official russian fonts with our current russian translations do not always fit the space allowed for them, and sometimes placement is odd.

Set to RFC because I finished late and want to reread/test the code myself with fresh eyes before merging.

CorsixTH-SoftClub

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -519,9 +519,17 @@ end
 --!return err (string) What to report back to the player on failure
 function App:initLanguage()
   -- Make sure that we can actually show the desired language.
-  -- If we can't, work out the most common issue and reset to English or Russian.
+  -- If we can't, work out the most common issue and reset to a revert to a
+  -- language that is likely to be supported for the version of Theme Hospital
+  -- the player has.
   local success, err = true, nil
   local language = self.config.language
+
+  -- Revert to Russian for the Russian regional release of Theme Hospital or
+  -- English otherwise.
+  --
+  -- When we support more regional versions of Theme Hospital this function
+  -- should be extended to cover those as well.
   local function revertToSupportedLanguage()
     local l
     if self.gfx:hasLanguageFont("cp437") then
@@ -529,7 +537,7 @@ function App:initLanguage()
     elseif self.gfx:hasLanguageFont("mik") then
       l = [[Russian]]
     else
-      print("Warning: No suitable font found for English or Russian. Defaulting to English, but the game may be broken.")
+      print("Warning: No suitable font found. Defaulting to English, but the game may be broken.")
       l = [[English]]
     end
     self.config.language = l


ARGAMX:
Well done, nice work 👍
Works good for me 🔥
Thanks

ARGAMX:
TH GOG Version (default) + English language (default):
Screenshot 2026-02-19 at 15 34 22

ARGAMX:
Overall, it looks great:

+Screenshot 2026-02-19 at 15 06 21
+Screenshot 2026-02-19 at 14 53 21
+Screenshot 2026-02-19 at 14 52 53
+Screenshot 2026-02-19 at 14 55 34
+Screenshot 2026-02-19 at 15 01 34
+Screenshot 2026-02-19 at 15 01 45
+Screenshot 2026-02-19 at 15 04 26
+Screenshot 2026-02-19 at 15 07 16
+Screenshot 2026-02-19 at 14 53 52

------

In the original official Rus translation the Casebook had the same weird purple color. This is OK.

TheCycoONE:
Fixed the english number offset issue.

TheCycoONE:
The ISO version (multi-5) continues to work, loading save games between versions works (as well as any time you switch languages)

TheCycoONE:
Fixed/removed some confusion I introduced with builtin vs TH assets; where builtin should mean assets included in CorsixTH e.g. the bootstrap font, not those from Theme Hospital's install.

TheCycoONE:
Ironically trying to play English on the Russian version is a very bad experience because all the original strings are in Russian so we're missing most of the translations.

Nothing to do, but I think it's funny. I'm assuming everyone with this version will want to play in Russian.

ARGAMX (approved):
chuck-norris-approves-this-pr

ARGAMX:
I have some founds and suggestions on some of the screens and functionality, but overall I'd like to merge this PR and then considering suggestions in separated issue.

CorsixTH/Lua/languages/brazilian_portuguese.lua
```diff
@@ -23,9 +23,11 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
+Font("cp437")
 Language("Português Brasileiro", "Brazilian Portuguese", "pt_br", "br")
 Inherit("English")
 Encoding(utf8)
+IsArabicNumerals(true)
```
lewri, line 30: This will require a wiki change I presume?
The cp437 denotions seem a bit odd on every language though except Russian? Couldn't we easier just assume it unless told otherwise?

CorsixTH/Lua/languages/russian.lua
```diff
@@ -18,7 +18,7 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE. --]]
 
-Font("unicode")
+Font("mik")
```
lewri, line 21: We're still able to handle the english version of the game loading russian correct?

CorsixTH/Lua/app.lua
```diff
@@ -519,36 +519,44 @@ end
...
-    language = [[English]]
-    self.config.language = language
+  local function revertToSupportedLanguage()
+    local l
+    if self.gfx:hasLanguageFont("cp437") then
+      l = [[English]]
+    elseif self.gfx:hasLanguageFont("mik") then
+      l = [[Russian]]
+    else
+      print("Warning: No suitable font found for English or Russian. Defaulting to English, but the game may be broken.")
```
lewri: This is a bit of a curveball to those not even aware there was a russian localisation of TH.

CorsixTH/Lua/app.lua
```diff
@@ -519,36 +519,44 @@ end
 --!return err (string) What to report back to the player on failure
 function App:initLanguage()
   -- Make sure that we can actually show the desired language.
-  -- If we can't, work out the most common issue and reset to English.
+  -- If we can't, work out the most common issue and reset to English or Russian.
```
lewri: `or Russian if using that localisation of the original game`?

lewri:
Probably best to wait for @Alberth289346 to review before merging.

CorsixTH/Lua/app.lua
TheCycoONE: yes

CorsixTH/Lua/app.lua
TheCycoONE: There is no expected circumstance where this error will come up, I'm not concerned. To get this error they couldn't be using the normal english version either.

CorsixTH/Lua/languages/russian.lua
TheCycoONE: Yes

CorsixTH/Lua/languages/brazilian_portuguese.lua
TheCycoONE: I like it better explicit, and there was already logic that assumes that if there is no font it matches everything, which isn't true.

Hopefully having the same headers on every language file will make things clearer when they don't hold.

TheCycoONE:
> If they're using the Russian copy of the game, do we disable the original languages in the language selection? Or maybe disable other languages entirely...

If they use a Russian copy of the game only Russian shows up unless they have a TTF font defined, same as we do for the unsupported alphabets in the multi-5 version

CorsixTH/Lua/graphics.lua
```diff
@@ -389,7 +390,7 @@ function Graphics:loadBuiltinFont()
     sheet:setPalette(palette)
     sheet:load(dernc(tab), dernc(dat), true, self.target)
     font = TH.bitmap_font()
-    font:setSheet(sheet)
+    font:setSheet(sheet, 1) -- CorsixTH only ships with a cp437 font
```
Alberth289346:
```suggestion
    font:setSheet(sheet, charsets["cp437"]) -- CorsixTH only ships with a cp437 font
```

CorsixTH/Src/th.cpp
```diff
@@ -95,6 +96,20 @@ void CopyStringCP437(const uint8_t*& sIn, uint8_t*& sOut) {
...
+  do {
+    cChar = *sIn;
+    ++sIn;
+    if (cChar < 0x80) {
+      *sOut = cChar;
+      ++sOut;
+    } else {
+      utf8encode(sOut, cpmik_to_unicode_table[cChar - 0x80]);
+    }
+  } while (cChar != 0);
```
Alberth289346, line 110: This fails with the empty string.

CorsixTH/Src/th_gfx_font.cpp
```diff
@@ -118,9 +136,10 @@ text_layout bitmap_font::draw_text_wrapped(render_target* pCanvas,
 
       for (s = sMessage; s != sMessageEnd;) {
         const char* sOld = s;
-        unsigned int iChar =
-            unicode_to_codepage_437(next_utf8_codepoint(s, sMessageEnd));
-        iNextChar = unicode_to_codepage_437(static_cast<unsigned char>(*s));
+        unsigned int iChar = unicode_to_font_character(
+            next_utf8_codepoint(s, sMessageEnd), character_set);
+        iNextChar = unicode_to_font_character(static_cast<unsigned char>(*s),
+                                              character_set);
```
Alberth289346, line 142: I wonder if the character-set should be a class.

CorsixTH/Src/th.cpp
Alberth289346: Hmm, or are you relying on encoded utf-8 also being a `0` byte?

Alberth289346:
Likely not that useful in this project at this time, but OpenTTD at some point created separate files that describe properties of a graphics set.

It has things like names and CRC sums.

CorsixTH/Src/th.cpp
TheCycoONE: I am, the end of string null byte is the same. Matches the other versions of this function.

Alberth289346: ok, let's have a unicode 0 then :)

CorsixTH/Src/th_gfx_font.cpp
TheCycoONE: Trade offs, can't add a pure virtual base class to bitmap_font directly or pass it by value, it'd have to be a wrapped in a pointer both in the class and where it's passed in. Since there's no data associated with it, I think it's kinda ugly that way - yey c++

TheCycoONE: "C.4: Make a function a member only if it needs direct access to the representation of a class" would seem to apply here too; though if Uniform Function Call is ever implemented I'd wish that I swapped the argument order. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0251r0.pdf

As part of a bigger refactor it might make sense to put the character set handling, enum, tables, and related functions in a specific namespace.

CorsixTH/Lua/app.lua
TheCycoONE: Updated the warning message, no longer mentions Russian.

CorsixTH/Lua/app.lua
TheCycoONE: Updated comment to be clearer.

PR #3238, opened 2026-02-19
Adjust overall machine decrease when earthquake

### LINKED ISSUE #3123: [Bug] Earthquakes increase overall machine usage counter
Image

### Describe the issue

Earthquakes increase the overall machine usage counter (“Number of times hardware used”), which is not connected with the machine durability, but only used for statistics.
Earthquakes should increase only "Times Used" which is connected with the machine's durability.
As in the original Theme Hospital.

----

### Steps to Reproduce

1. Load attached savegame.
2. Press on the Cardio machine (bottom right on the screen), to open Machine Window.
3. Observe the top value - “Number of times hardware used” - currently it is 4.
4. Press the “Z” key and hold until 30 Oct when a second earthquake strikes. 
5. Observe that the “Number of times hardware used” value now is not 4, but increased significantly up to 9 due strike.

----

### Expected Behaviour

“Number of times hardware used” value should not increase due to earthquakes. 
Only "Times Used".
As in the original TH.

----

### Save Game

[Earthquake.0.69.0.sav.zip](https://github.com/user-attachments/files/22413628/Earthquake.0.69.0.sav.zip)

----

### CorsixTH Version

0.69.1

----

ManoloMC2002:
*Fixes #3123*

When the function *Machine:earthquakeImpact(room)* is called, a local boolean attribut named *earthquake_activated* in machine.lua will change to *true*. And when the incrementUsageCounts is called, there is a verification to know if the reason of the increment came from a earthquake.

### DIFF
--- a/CorsixTH/Lua/entities/machine.lua
+++ b/CorsixTH/Lua/entities/machine.lua
@@ -126,11 +126,10 @@ end
 
 --! Call on machine use.
 --!param room (object) machine room
+--!param earthquake_activated (boolean) If true, machine was damaged by an active earthquake
 --!return (bool) is room exploding after this use
 function Machine:machineUsed(room, earthquake_activated)
-  if type(earthquake_activated) ~= "boolean" then
-      earthquake_activated = false
-  end
+  earthquake_activated = not not earthquake_activated
   -- Do nothing if the room has already crashed
   if room.crashed then
     return
@@ -157,7 +156,6 @@ end
 
 --! Call after use of the machine.
 function Machine:incrementUsageCounts(total_usage_only, is_earthquake)
-  total_usage_only = total_usage_only or false
   if not is_earthquake then
       self.total_usage = self.total_usage + 1
   end


CorsixTH/Lua/entities/machine.lua
```diff
@@ -121,21 +121,24 @@ end
...
-  self:machineUsed(room)
+  self:machineUsed(room, true)
 end
 
 --! Call on machine use.
 --!param room (object) machine room
 --!return (bool) is room exploding after this use
-function Machine:machineUsed(room)
+function Machine:machineUsed(room, earthquake_activated)
+  if type(earthquake_activated) ~= "boolean" then
```
TheCycoONE: This seems unnecessary, a nil is falsy, and you only use it as a boolean in the next call.

If you did want to strictly convert to a boolean here `not not` works for that.

ManoloMC2002: I want to force the use of this second parameter to be only a boolean. If I use a ```not not``` and the programmer just put a number or a string, it will return ```true``` and I don't want that.

lewri: Lua is a dynamically typed language; as such values hold the type and not the variable. However, from the point of lua, values are considered `true` except when their values are either `false` or `nil`. You can't control a variable in lua to be exclusively boolean as a type though.

See
https://www.lua.org/pil/2.html
and 
https://www.lua.org/manual/5.1/manual.html#:~:text=2%2E2%20%E2%80%93%20Values%20and%20Types (section 2.2)

The correct answer is what cyco says in light of this.
`nil` is the equivalent to `false` in this case. You don't need a check for boolean.

For a developer standard, you should also define what you expect by editing the doc above the function as below:
```lua
--! Call on machine use.
--!param room (object) machine room
--!param earthquake_activated (boolean) If true, machine was damaged by an active earthquake
--!return (bool) is room exploding after this use
```

ManoloMC2002: Oh okay I didn't know thank you

lewri (approved):
Lua 5.1 test error is unrelated.

PR #2979, opened 2025-08-24
Better error handling for non-existent languages

lewri:
*Fixes #2978*

**Describe what the proposed change does**
- Adds helper function to check player's specified language exists
- Returns language to English and shows error message on main menu (fixes crash at end of movie and elsewhere). 
- Tell the player the language value they originally set and where to go to fix (as there are now two error types).
- Also includes some refactor to accommodate new handling
- Doc, as usual.

One quirk is that as part of `App:fixConfig()` we force lowercase the language on launch. However, we still try to convert back to lowercase in other places (mostly) unnecessarily. My use of capitalised `[[English]]` on App.lua line 516 probably needs changing if there's a future PR to clean things up.

image

image

PR #3240, opened 2026-02-20
[RFC] Revert bracket support in filenames

### LINKED ISSUE #3034: [Bug] Language value in config causes language reset
Image

### Describe the issue

Regression of #2979 

Language values set by CorsixTH following this PR are often incompatible with version 0.69.1 and earlier.
The opposite is also true.

As a result, CorsixTH reads the language as `[english` 
Within the config, its value is represented as `language = [=[english]=]`

### Steps to Reproduce

Thanks @ARGAMX for reproduction!

1. Open config.txt; delete the entire language line. 

2.  Launch 0.69.1, config will add value back as `language = [=[english]=]`

3. Launch a recent master after 2979. Note that the game says the language `[english` couldn't be loaded and has been reset to English.

4. Launch 0.69.1, note that the game's reports that `[english` isn't a language known.

You can keep going back and forth between these two versions and keep getting the error.

### Expected Behaviour

The game is meant to only identify the language name, but superfluous parts of the string syntax end up in the value, this shouldn't happen.

### Save Game

### CorsixTH Version

0.69.1 and Master

### LINKED ISSUE #3224: Lost forward compatibility for paths in config
----

### Describe the issue

Seems the handler of the `config.txt` file has compatibility issues. 

If you run the latest version, say 0.69, and set the path to the screenshots folder, and then launch 0.68 or any older version, the paths in the older version are not reset. It stays as it was set in 0.69.
So forward compatibility works ok. 

But in our current master version, 0.70.0, forward and backward compatibility doesn't work. 
I'm constantly switching between either 0.70.0 and 0.69.2, and the font paths are constantly reset for me.
I didn't experience this problem when we were in the cycle between 0.68 and 0.69.

This includes saves folder, screenshots folder, soundfont and font paths.

I bisected it and for the first time that can be detected in
bd9f15d7070afe505c5d0bc81d2d0fcb1a9836d5 (Toolbar machine button (#2966))

I reverted that commit and... the problem still persists. Apparently, that commit wasn't the cause but rather the trigger, as it simply changed `config_finder.lua`, which, by the way, happens also in other commits.

But right before this commit there was another commit that was just related to `config.txt`. It is #2979.
After reverting the #2979 problem is gone.

So forward compatibility for paths in config for 0.70.0 was lost in #2979.

----

### Operating System

macOS 14.5

----

### Steps to Reproduce

1. Prepare build 0.69.2 and trunk build 0.70.0.
2. Run trunk build 0.70.0.
3. Go to `Settings` -> `Folders`.
4. Set any non-default location for `Screenshots`.
5. Close 0.70.0 game.
6. Run build 0.69.2.
7. Go to `Settings` -> `Folders`.
8. Observe that the location for `Screenshots` is now `Default location`.
9. Close 0.69.2 game.
10. Run trunk build 0.70.0 again. 
11. Go to `Settings` -> `Folders`.
12. Observe that the location for `Screenshots` is `Default location` too. So this setting was completely lost.

13. Repeat this experiment with the 0.68 and 0.69 pair. Set path in 0.69. It will stay in 0.68.

----

### Expected Behaviour

Forward compatibility for paths in config should work as before.

----

lewri:
*Fixes #3224*
*Fixes #3034*

- Revert #2391 but leaves the utility function change intact for now.

Asking @tobylane to review if he's got any alternatives (bar changing config file format) before we break support for square brackets in config values.

This is in prep in case we don't manage to agree on changes to the config file system before next release. The other reason is sometimes I've found cases where the `[=[` format is incompatible with itself.

@ARGAMX it looks like this change will convert `[=[ ]=]` formatted strings back to `[[ ]]`, but won't crash on reading them.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1084,7 +1084,7 @@ function App:saveConfig()
           handled_ids[identifier] = true
           if value ~= tostring(self.config[identifier]) then
             lines[#lines] = string.format("%s = %s", identifier,
-                serialize(self.config[identifier], { long_bracket_level_start = 1 } ))
+                serialize(self.config[identifier], { long_bracket_level_start = 0 } ))
           end
         end
       end


CorsixTH/Lua/app.lua
```diff
@@ -1083,8 +1083,13 @@ function App:saveConfig()
           -- replace the line, if needed
           handled_ids[identifier] = true
           if value ~= tostring(self.config[identifier]) then
-            lines[#lines] = string.format("%s = %s", identifier,
-                serialize(self.config[identifier], { long_bracket_level_start = 1 } ))
```
lewri, line 1103: Alternatively if we want to preserve this more, change that 1 to a 0.

TheCycoONE: This sounds like a safer change

TheCycoONE:
I missed some conversation somewhere, and it's not at all clear from the description - how is reverting the serialization change from 2024 fixing a recent problem with configs reverting?

lewri:
> I missed some conversation somewhere, and it's not at all clear from the description - how is reverting the serialization change from 2024 fixing a recent problem with configs reverting?

It's always been there since the original pr. But #2979 makes it much more obvious. It's possible to replicate this with the TH install value as well.

TheCycoONE:
And what causes the revert, that is fixed by avoiding = signs?

tobylane:
> The other reason is sometimes I've found cases where the [=[ format is incompatible with itself.

I see that too, and it's hard to argue for any option with that confusion. Ignoring that, I slightly prefer the 1 > 0 change (by removing that parameter), to continue using the serialize function. 

For fixing an errant [ once it's in, put this after revertToEnglish in App:initLanguage.

```lua
  -- Strip any leading [
  while language:sub(1,1) == "[" do language = language:sub(2) end
  if language ~= self.config.language then
    self.config.language = language
    self:saveConfig()
  end
```

lewri:
> > The other reason is sometimes I've found cases where the [=[ format is incompatible with itself.
> 
> I see that too, and it's hard to argue for any option with that confusion. Ignoring that, I slightly prefer the 1 > 0 change (by removing that parameter), to continue using the serialize function.
> 
> For fixing an errant [ once it's in, put this after revertToEnglish in App:initLanguage.
> 
> ```lua
>   -- Strip any leading [
>   while language:sub(1,1) == "[" do language = language:sub(2) end
>   if language ~= self.config.language then
>     self.config.language = language
>     self:saveConfig()
>   end
> ```

Seems hacky, though it's the best we've got for the moment (doesn't resolve backwards compatibility though, only forward)

> And what causes the revert, that is fixed by avoiding = signs?

Pretty much it seems. I'm not entirely sure why this has broken though on a deeper level.

lewri:
I have no idea what the cause is unfortunately.

This pr is mainly to keep open long term as a last resort if it's not fixed before the next release. But I'll update the code to make the smaller change for now.

lewri:
Closed in favour of #3254

PR #3241, opened 2026-02-21
Simplify dropdown item creation

lewri:
*Fixes #*

- Refactor code to drop the no longer necessary differentiation between dropdown items with and without tooltips.

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/dropdown.lua
+++ b/CorsixTH/Lua/dialogs/resizables/dropdown.lua
@@ -61,7 +61,7 @@ function UIDropdown:UIDropdown(ui, parent_window, parent_button, items, callback
       :setTooltip(item.tooltip and unpack(item.tooltip) or nil)
       :setLabel(item.text, item.font)
       :makeToggleButton(-1, -1, width, height, nil,
-          --[[persistable:dropdown_tooltip_callback]] function() self:selectItem(i) end)
+          --[[persistable:dropdown_callback]] function() self:selectItem(i) end)
       :enable(not item.disabled)
     y = y + height
   end


CorsixTH/Lua/dialogs/resizables/dropdown.lua
```diff
@@ -57,17 +57,12 @@ function UIDropdown:UIDropdown(ui, parent_window, parent_button, items, callback
 
   local y = 0
   for i, item in ipairs(items) do
-  if item.tooltip and item.tooltip[1] then
     self:addBevelPanel(1, y + 1, width - 2, height - 2, parent_window.colour)
-      :setTooltip(unpack(item.tooltip))
+      :setTooltip(item.tooltip and unpack(item.tooltip) or nil)
       :setLabel(item.text, item.font)
-      :makeToggleButton(-1, -1, width, height, nil, --[[persistable:dropdown_tooltip_callback]] function() self:selectItem(i) end)
+      :makeToggleButton(-1, -1, width, height, nil,
```
lewri, line 63: Making a dropdown item a toggle button is actually harmless.

CorsixTH/Lua/dialogs/resizables/dropdown.lua
```diff
@@ -57,17 +57,12 @@ function UIDropdown:UIDropdown(ui, parent_window, parent_button, items, callback
...
+          --[[persistable:dropdown_tooltip_callback]] function() self:selectItem(i) end)
       :enable(not item.disabled)
-  else
-    self:addBevelPanel(1, y + 1, width - 2, height - 2, parent_window.colour)
-      :setLabel(item.text, item.font)
-      :makeButton(-1, -1, width, height, nil, --[[persistable:dropdown_callback]] function() self:selectItem(i) end)
```
lewri, line 69: In a menu afaik dropping a persistable doesn't affect the game? I would rename the remaining persistable to this but not sure if that'd risk breaking more.

TheCycoONE: It depends on whether or not it can be in the object tree when the game is saved. If we have an UIDropdown use in game then removing or renaming the persistable is a breaking change. If we don't then the persistable is not needed at all.

TheCycoONE: I don't see any use of UIDropDown in the game, only in new game, options, and sound settings, so you're free to do what you want with those labels.

TheCycoONE (approved):

CorsixTH/Lua/dialogs/resizables/dropdown.lua
```diff
@@ -57,17 +57,12 @@ function UIDropdown:UIDropdown(ui, parent_window, parent_button, items, callback
...
   local y = 0
   for i, item in ipairs(items) do
-  if item.tooltip and item.tooltip[1] then
     self:addBevelPanel(1, y + 1, width - 2, height - 2, parent_window.colour)
-      :setTooltip(unpack(item.tooltip))
+      :setTooltip(item.tooltip and unpack(item.tooltip) or nil)
       :setLabel(item.text, item.font)
-      :makeToggleButton(-1, -1, width, height, nil, --[[persistable:dropdown_tooltip_callback]] function() self:selectItem(i) end)
+      :makeToggleButton(-1, -1, width, height, nil,
+          --[[persistable:dropdown_tooltip_callback]] function() self:selectItem(i) end)
```
lewri:
```suggestion
          --[[persistable:dropdown_callback]] function() self:selectItem(i) end)
```

PR #3245, opened 2026-02-23
Support predefined colours in panels

lewri:
*Fixes #*

- **The code currently is unused, it'll be used in further PRs**
- Adds colour by string support to the Window class
- If a string is given that has no definition, it will use the `default` colour.

This originally was going to include the other changes to the dialog files but there's more movement going on at the moment in those files I'll hold off until this is accepted.

### DIFF
--- a/CorsixTH/Lua/utility.lua
+++ b/CorsixTH/Lua/utility.lua
@@ -249,6 +249,55 @@ AnimationEffect.None = 0
 AnimationEffect.Glowing = 1
 AnimationEffect.Jelly = 2
 
+-- Predefined colours
+Colours = {}
+Colours.PanelDefault = { -- CorsixTH's default panel colour (pale purple)
+  red = 154, green = 146, blue = 198
+}
+Colours.Disabled = { -- Specific grey with purplish aspect
+  red = 127, green = 123, blue = 149
+}
+Colours.Setting = Colours.PanelDefault
+Colours.SettingHover = { -- Slightly darkened from a pale purple
+  red = 133, green = 128, blue = 183,
+}
+Colours.SettingActive = Colours.SettingHover
+Colours.HotkeyBox = { -- Purple
+  red = 081, green = 076, blue = 150,
+}
+Colours.HotkeyBoxActive = { -- Deep purple
+  red = 041, green = 036, blue = 090,
+}
+Colours.Textbox = { -- Black
+  red = 000, green = 000, blue = 000
+}
+Colours.Scrollbar = { -- Purple
+  red = 081, green = 076, blue = 150,
+}
+Colours.Title = { -- Very pale purple
+  red = 174, green = 166, blue = 218
+}
+Colours.Caption = { -- Murky purple
+  red = 134, green = 126, blue = 178
+}
+
+-- Include the standard colours also
+Colours.White = {
+  red = 255, green = 255, blue = 255
+}
+Colours.Black = {
+  red = 000, green = 000, blue = 000
+}
+Colours.Red = {
+  red = 255, green = 000, blue = 000
+}
+Colours.Green = {
+  red = 000, green = 255, blue = 000
+}
+Colours.Blue = {
+  red = 000, green = 000, blue = 255
+}
+
 -- Compare values of two simple (non-nested) tables
 function compare_tables(t1, t2)
   local count1 = 0
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -23,14 +23,15 @@ self = false
 globals = { -- Globals
             "_A", "_S",
             "corsixth",
-            "class", "compare_tables", "DrawFlags", "DrawingLayers",
-            "destrict", "flag_clear", "flag_isset", "flag_set", "flag_toggle",
-            "lfs", "list_to_set", "loadfile_envcall", "loadstring_envcall",
+            "class", "Colours", "compare_tables", "DrawFlags",
+            "DrawingLayers", "destrict", "flag_clear", "flag_isset",
+            "flag_set", "flag_toggle", "lfs", "list_to_set",
+            "loadfile_envcall", "loadstring_envcall",
             "pause_gc_and_use_weak_keys", "permanent",
-            "rangeMapLookup", "rnc", "strict_declare_global", "table_length",
-            "unpermanent", "values", "serialize", "array_join", "shallow_clone",
-            "staff_initials_cache", "hasBit", "bitOr", "inspect",
-            "getRandomEntryFromArray",
+            "rangeMapLookup", "rnc", "strict_declare_global",
+            "table_length", "unpermanent", "values", "serialize",
+            "array_join", "shallow_clone", "staff_initials_cache",
+            "hasBit", "bitOr", "inspect", "getRandomEntryFromArray",
 
             -- Game classes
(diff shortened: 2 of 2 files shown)


CorsixTH/Lua/window.lua
```diff
@@ -205,6 +205,53 @@ function Panel:makeHotkeyBox(...)
...
+    ["setting_active"] = { -- Deep purple
+      red = 041, green = 036, blue = 090
+    },
+    ["title"] = { -- Vey pale purple
+      red = 174, green = 166, blue = 218
+    },
+    ["caption"] = { -- Murky purple
+      red = 134, green = 126, blue = 178
+    },
+    ["textbox"] = { -- Black
```
lewri: Because this is black, I wasn't sure if to make a `black` definition that was the same as this; or remove `textbox` and have only `black`.

Alberth289346: I would add both.

CorsixTH/Lua/window.lua
```diff
@@ -487,7 +487,7 @@ size.
 ]]
 function Window:addColourPanel(x, y, w, h, r, g, b, apply_ui_scale)
   if type(r) == "string" then
-    local col = self:_getPredefinedColour(r)
+    local col = Panel:_getPredefinedColour(r)
```
TheCycoONE: Not sure what you're trying to do necessarily with this. If you don't want to pass in self, use `.` not `:` but -- since this function doesn't need self, and is private it can just be a local function in the file, doesn't need to be part of Panel at all; that would be better.

TheCycoONE:
Cleaner to have color constants on a new Color table than to overload various functions (but not all functions) to take strings.

lewri:
> Cleaner to have color constants on a new Color table than to overload various functions (but not all functions) to take strings.

You may need to elaborate on that one more as I'm not sure what's being suggested to implement

TheCycoONE:
> > Cleaner to have color constants on a new Color table than to overload various functions (but not all functions) to take strings.
> 
> You may need to elaborate on that one more as I'm not sure what's being suggested to implement

https://github.com/CorsixTH/CorsixTH/blob/89ec2328fb1df8539655226cde50bb64ee51399d/CorsixTH/Lua/utility.lua#L217-L250

So like:

```lua
Colors = {}
Colors.Red = { red = 255, blue = 0, green = 0 }
```

elsewhere:

```
mypanel.setColor(Colors.Red)
```

Then it works everywhere we take color tables, without changing every spot.

lewri:
Should be done.

CorsixTH/Lua/utility.lua
```diff
@@ -249,6 +249,46 @@ AnimationEffect.None = 0
...
+  red = 000, green = 000, blue = 000
+}
+Colours.Red = {
+  red = 255, green = 000, blue = 000
+}
+Colours.Green = {
+  red = 000, green = 255, blue = 000
+}
+Colours.Blue = {
+  red = 000, green = 000, blue = 000
```
ARGAMX:
```suggestion
  red = 000, green = 000, blue = 255
```

lewri: 🤣 forgot about that one

CorsixTH/Lua/utility.lua
```diff
@@ -249,6 +249,46 @@ AnimationEffect.None = 0
 AnimationEffect.Glowing = 1
 AnimationEffect.Jelly = 2
 
+-- Predefined colours
+Colours = {}
+Colours.Default = { -- CorsixTH's default panel colour (pale purple)
```
TheCycoONE: Maybe better as PanelBackground or similar, since these colours are now disconnected from panels.

lewri: It's not strictly for just panel backgrounds, so it'd be PanelColour at its shortest

CorsixTH/Lua/utility.lua
```diff
@@ -249,6 +249,46 @@ AnimationEffect.None = 0
...
 AnimationEffect.Jelly = 2
 
+-- Predefined colours for panels (such as backgrounds, buttons, bevels etc)
+PanelColour = {}
+PanelColour.Default = { -- CorsixTH's default panel colour (pale purple)
+  red = 154, green = 146, blue = 198
+}
+PanelColour.Disabled = { -- Grey
+  red = 127, green = 123, blue = 149
+}
```
Alberth289346: Somewhat a pity that a value doesn't fit at a single line.

Perhaps?:
``` lua
PanelColour._grey = {red = 127, green = 123, blue = 149}

PanelColour.Disabled = PanelColour._grey
```
The name of the colour is then naturally included.

lewri: They can fit on one line, it was a stylistic choice though not to.
If we're going to chain-reference can we at least stick with `PanelColour.Grey`?

Edit: The grey actually defined here is not a true neutral grey, it's an extraction of the disabled grey the code makes from Color.Default (and I need it elsewhere in other PRs)

ARGAMX: I vote for chain-reference style too.
And it makes sense that in such a style, `Pure Colors` should be private.
To avoid the temptation to use them directly in layout.

Alberth289346: > They can fit on one line, it was a stylistic choice though not to.

Oh, ok. I prefer minimal clutter. I believe a list of single line assignments is simpler to read than a list of 3-line assignments.
Making all numbers 3 digits long seemed to point in that direction as well.

But if you want it differently, that's ok.

> If we're going to chain-reference can we at least stick with `PanelColour.Grey`?

Sure, no problem. I was just preserving the set of externally usable names.

CorsixTH/Lua/utility.lua
```diff
@@ -249,6 +249,46 @@ AnimationEffect.None = 0
...
+  red = 174, green = 166, blue = 218
+}
+PanelColour.Caption = { -- Murky purple
+  red = 134, green = 126, blue = 178
+}
+PanelColour.Textbox = { -- Black
+  red = 000, green = 000, blue = 000
+}
+-- Include the standard colours also
+PanelColour.White = {
```
TheCycoONE: This ticket is becoming rather bikeshed, but I don't like the standard colors being namespaced as 'PanelColour'

CorsixTH/Lua/utility.lua
```diff
@@ -248,6 +249,55 @@ AnimationEffect.None = 0
...
+}
+Colours.Disabled = { -- Specific grey with purplish aspect
+  red = 127, green = 123, blue = 149
+}
+Colours.Setting = Colours.PanelDefault
+Colours.SettingHover = { -- Slightly darkened from a pale purple
+  red = 133, green = 128, blue = 183,
+}
+Colours.SettingActive = Colours.SettingHover
+Colours.HotkeyBox = { -- Purple
```
lewri, line 265: This table has been extended now as I had to redo it anyway (as mentioned in chats yesterday, I had added more colour sets)

TheCycoONE:
Do you want to use these in this same PR, now that we have the list sorted out?

lewri:
> Do you want to use these in this same PR, now that we have the list sorted out?

No, this can merge as is. I will transition it slowly as we go. It means some code will duplicate things between them but will be cleaned up at the end.

PR #3249, opened 2026-02-24
Truncate the name of the player on new game

### LINKED ISSUE #2124: [Bug] Player name has no length limit
#### Describe the issue
Names that are too long are accepted and used

#### Steps to Reproduce
1. In the New Game screen enter a long name (over 15 characters)

2. Start game

3. Look at name, eg Progress Report. The name isn't clipped and may continue beyond the space for it.

#### Expected Behavior
It's refused, or clipped.

#### System Information
3e7b389f current master

### Additional Info
https://github.com/CorsixTH/CorsixTH/blob/dfa5af941cf8ebc6e8f76640519f407ed95a4761/CorsixTH/Lua/dialogs/resizables/new_game.lua#L117-L118 has the character limit
https://github.com/CorsixTH/CorsixTH/blob/81db3338d89df253e85fc6b7ace9d937092dafaa/CorsixTH/Lua/window.lua#L1027 has the text input check

ManoloMC2002:
*Fixes #2124*

I truncate the maximum player name before any change of username, so the variable already have a maximum length.


CorsixTH/Lua/dialogs/resizables/new_game.lua
```diff
@@ -110,7 +110,7 @@ function UINewGame:UINewGame(ui)
       if not name:find("%S") then
         self.name_textbox:setText(self.player_name)
       else
-        self.player_name = name
+        self.player_name = name:sub(1, 8)
```
lewri: Consensus on if 8 is what we want for sure should be double checked.

CorsixTH/Lua/dialogs/resizables/new_game.lua
```diff
@@ -110,7 +110,7 @@ function UINewGame:UINewGame(ui)
       if not name:find("%S") then
         self.name_textbox:setText(self.player_name)
```
lewri, line 93: Textbox entry would still need to control for the max length (safer after submission, bonus if we show some courteous message we will trim/have trimmed)

CorsixTH/Lua/dialogs/resizables/new_game.lua
tobylane: The user is expecting to type until they can't, then that input is accepted. Ideally this limit is the same as the trimming, but this limit has no effect. The bug is in the input handling, not the dialog.

tobylane (changes requested):
This surprises the user too much.

CorsixTH/Lua/dialogs/resizables/new_game.lua
lewri: I agree, though I imagine we can still prematurely trim on Line 102 when we retrieve the OS Name or predefined value in config?

tobylane: Yes that's good.

ManoloMC2002: Sooo max length is 8 and I should control the textbox ?

CorsixTH/Lua/dialogs/resizables/new_game.lua
Alberth289346: Config file also needs that information.

[People do edit the config file manually, and thus never enter their name in the GUI.]

ManoloMC2002:
I'll make appear a message while creating the game to inform the player of max character

tobylane:
This changes two panels that don't exist. I quickly added those two panels in a proof of concept, but you may have a design in mind.
Screenshot 2026-05-13 at 18 29 41

```diff
index 58a3dc21..ef9c3e94 100644
--- a/CorsixTH/Lua/dialogs/resizables/new_game.lua
+++ b/CorsixTH/Lua/dialogs/resizables/new_game.lua
@@ -84,6 +84,7 @@ function UINewGame:UINewGame(ui)
   self.player_name = (pname and pname:len() > 0) and pname or os.getenv("USER") or os.getenv("USERNAME") or "PLAYER"
   self:addBevelPanel(20, 45, 140, 30, col.caption, col.bg, col.bg)
     :setLabel(_S.new_game_window.player_name).lowered = true
+  self.char_limit_label = self:addBevelPanel(165, 30, 140, 15, col.button)
   self.name_textbox = self:addBevelPanel(165, 45, 140, 30, col.textbox, col.bg, col.bg)
     :setTooltip(_S.tooltip.new_game_window.player_name):setAutoClip(true)
     :makeTextbox(
@@ -112,7 +113,7 @@ function UINewGame:UINewGame(ui)
     :makeToggleButton(0, 0, 135, 30, nil, self.buttonTutorial):setTooltip(_S.tooltip.new_game_window.tutorial)
 
   -- Start and Cancel
-  self:addBevelPanel(20, 165, 140, 40, col.button):setLabel(_S.new_game_window.start):makeButton(0, 0, 135, 40, nil, self.buttonStart):setTooltip(_S.tooltip.new_game_window.start)
+  self.start_button_panel = self:addBevelPanel(20, 165, 140, 40, col.button):setLabel(_S.new_game_window.start):makeButton(0, 0, 135, 40, nil, self.buttonStart):setTooltip(_S.tooltip.new_game_window.start)
   self:addBevelPanel(165, 165, 140, 40, col.button):setLabel(_S.new_game_window.cancel):makeButton(0, 0, 135, 40, nil, self.buttonCancel):setTooltip(_S.tooltip.new_game_window.cancel)
 end
```

Alberth289346:
An alternative is to end with `...`, as in `alb...`. In that way it's clear more text exists but it isn't displayed.

Ideally it would compute this on the fly rather than forbid longer text or truncate the last part. There are more places where text needs to be shortened (such as long file paths), and tomorrow we may find more space for displaying the name or path.

I can live with the current solution though.

tobylane:
This is a more complex issue than it seems, the bug isn't in this dialog. Part of that is that setting the limit to 15 in this way still lets a 16th character be shown in this dialog, but it is correctly cut off for the hospital.

PR #3250, opened 2026-02-27
Increase bevel intensity

lewri:
*Fixes #*

- Make the highlight and shadow of the bevel more extreme from base colour, so they pop more visually.
- Doesn't affect custom highlight and shadow

image


Alberth289346:
The approach is somewhat crude (you get different colours rather than different intensities of the same colour), but good enough imho.

PR #3251, opened 2026-02-27
Fix z order issue in options

### LINKED ISSUE #3242: UI Settings dialog has a weird behavior
### Describe the issue

Game malfunctions when you are trying to change settings that create a new dialog when you click on it

### Steps to Reproduce

1. Open CorsixTH and go into Settings menu

2. Start changing Zoom speed / Scroll speed / Shift scroll speed

Image

3. When trying to interact with this dialog when it is above settings menu, it actually goes to the bottom layer instead of responding to your interaction

Image

And sometimes it can result in this:
Image

### Expected Behaviour

Do not malfunction

- [ ] Right mouse panning

### Save Game

No

### CorsixTH Version

master

### Operating System

Windows 11

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml

```

### Additional Information

_No response_

lewri:
*Fixes #3242*

- The mini windows from Options will no longer get sent to back

Ping @Matroftt 

<sup>Rel notes: Don't include</sup>

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/options.lua
+++ b/CorsixTH/Lua/dialogs/resizables/options.lua
@@ -617,11 +617,11 @@ function UIResolution:ok()
   end
 end
 
-function UIResolution:onMouseDown(button, x, y)
+function UIResolution:onMouseUp(button, x, y)
   if not self:hitTest(x, y) then
     self:close(false)
   end
-  UIResizable.onMouseDown(self, button, x, y)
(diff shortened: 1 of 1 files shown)


ARGAMX:
I missed and didn't review 3215, but I noticed that if you need to handle a click and you don't know which to choose, `onMouseDown` or `onMouseUp`, then choose `onMouseUp` (or at least try it first).
I've noticed several bugs in our game related to `onMouseDown` being used where `onMouseUp` should be.

PR #3254, opened 2026-03-02
Config cleanup

### LINKED ISSUE #3034: [Bug] Language value in config causes language reset

TheCycoONE:
Clean up the config_finder and consolidate loading and saving configuration into it.

Simplify configuration management by removing all manual parsing of the configuration file. It is now always saved as a template and always loaded as a lua environment.

config_finder's return values are now named for ease of use, and the defaults are returned by function rather than as static values that could be corrupted.

We no longer try to detect when new configuration values are added to the template, instead the config is always saved at the end of app initialization. This should incorporate any changed / detected parameters during initialization so saveConfig should not be called in other init functions.

Ignore the individual commits, this PR is intended to be squashed.

Fixes #3034

### DIFF
--- a/CorsixTH/Lua/config_finder.lua
+++ b/CorsixTH/Lua/config_finder.lua
@@ -36,6 +36,10 @@ SOFTWARE. --]]
 -- function to include the parameter in the hotkeys.txt configuration file.
 -- The same params syntax applies.
 --
+-- After changing the config.txt file you need to generate a new template for
+-- the windows installer by running `lua scripts/generate_windows_config.lua`
+-- from the root of the repository.
+--
 -- The format of the configuration files should not be depended on anywhere
 -- outside of this module. It is a layering violation to attempt to parse or
 -- update the file anywhere else in the codebase e.g. app.lua.


tobylane:
This won't create a config file where the file is missing or empty. Check for those cases before `loadfile_envcall`, as that should fail on a missing file?

edit: These long lines could be neatened with a helper function? 
```lua
local function user_or_default(setting, default_value)
  if config[setting] then return serialize(config[setting])
 else return "nil -- " .. default_value
end
...
'theme_hospital_install = ' .. (user_or_default("theme_hospital_install", "[[X:\\ThemeHospital]]") .. '\n' .. [=[
```

TheCycoONE:
> This won't create a config file where the file is missing or empty. Check for those cases before `loadfile_envcall`, as that should fail on a missing file?
> 
> edit: These long lines could be neatened with a helper function?
> 
> ```lua
> local function user_or_default(setting, default_value)
>   if config[setting] then return serialize(config[setting])
>  else return "nil -- " .. default_value
> end
> ...
> 'theme_hospital_install = ' .. (user_or_default("theme_hospital_install", "[[X:\\ThemeHospital]]") .. '\n' .. [=[
> ```

Ah - yes, I lost that with my changes yesterday, it's not suppose to error in app.lua; but print something and continue with the default values.

TheCycoONE:
I went back to creating the default config file in config_finder like it use to instead of changing the app behavior; having implemented both ways, this was a smaller change with fewer side effects.

Now it does it on load, mostly to avoid being aware of the windows generation tool.

CorsixTH/Lua/config_finder.lua
```diff
@@ -443,38 +508,39 @@ local string_02 = [=[
...
+'theme_hospital_install = ' .. (config_values.theme_hospital_install and serialize(config_values.theme_hospital_install) or 'nil -- [[X:\\ThemeHospital]]') .. '\n' .. [=[
 
 -------------------------------------------------------------------------------
 -- Font file setting. Can be changed from main game menu
 -- Specify a font file here if you wish to play the game in a language not
 -- present in the original game. Examples include Russian, Chinese and Polish.
---
-unicode_font = nil -- [[X:\ThemeHospital\font.ttc]]
+--]=] .. '\n' ..
+'unicode_font = ' .. (config_values.unicode_font and serialize(config_values.unicode_font) or 'nil -- [[X:\\ThemeHospital\\font.ttc]]') .. '\n' .. [=[
```
lewri: I assume these lines added of very similar functionality can't be extracted out to call a function instead?

lewri (commented):
Please could we document somewhere appropriate the format of adding new config/hotkey values with this change?

CorsixTH/Lua/config_finder.lua
TheCycoONE: Sure they could, they would also be much simpler if we didn't try to put:
`-- [[X:\\ThemeHospital\\font.ttc]]` on the same line / conditionally.

If we were willing to reformat to:
```
-- Theme hospital install folder: original game data files are loaded from this
-- folder. Between the square braces should be the folder which contains the
-- original HOSPITAL.EXE and/or HOSP95.EXE file. This can point to a copy of
-- the Theme Hospital demo, though a full install of the original game is
-- preferred.
-- e.g. theme_hospital_install = [[X:\ThemeHospital]]
theme_hospital_install = nil
```

Then it could be:

```lua
part[5] = [=[
------------------------------- FOLDER SETTINGS -------------------------------
-- These settings can also be changed from the Folders Menu
-- Theme hospital install folder: original game data files are loaded from this
-- folder. Between the square braces should be the folder which contains the
-- original HOSPITAL.EXE and/or HOSP95.EXE file. This can point to a copy of
-- the Theme Hospital demo, though a full install of the original game is
-- preferred.
-- e.g. theme_hospital_install = [[X:\\ThemeHospital]]
--]=] .. '\n' ..
'theme_hospital_install = ' .. serialize(config_values.theme_hospital_install) .. '\n' .. [=[
```

TheCycoONE:
> Please could we document somewhere appropriate the format of adding new config/hotkey values with this change?

Like what, a wiki page that says add to config content and default values?

TheCycoONE:
> #3034 ok, resolved. I can't reproduce it anymore.
> 
> regarding #3224: If I specify the font path in `0.69.2` and then run this `0.70.0` build, the game launches with the selected font. Good. If I now run `0.69.2` again, the path resets to the default. Sad. The same applies to the path to screenshots, saves, music, and soundfont. Backward compatibility works. Forward compatibility doesn't work. Thus #3224 isn't resolved.

That doesn't sound possible to resolve, if the old version was broken whenever the file contains any new values. We just need to make sure it works going forward. Your new 0.70 version presumably isn't using `[=[` because this version doesn't unless necessary.

lewri:
I think as long as we don't crash a previous version of the game trying to load the new config changes that is probably the best we can ask for (the language crash notwithstanding, that was a bug in strings that fell over on a bad config read)


> Like what, a wiki page that says add to config content and default values?

Wiki or the file, either is fine imo

ARGAMX:
> > #3034 ok, resolved. I can't reproduce it anymore.
> > regarding #3224: If I specify the font path in `0.69.2` and then run this `0.70.0` build, the game launches with the selected font. Good. If I now run `0.69.2` again, the path resets to the default. Sad. The same applies to the path to screenshots, saves, music, and soundfont. Backward compatibility works. Forward compatibility doesn't work. Thus #3224 isn't resolved.
> 
> That doesn't sound possible to resolve, if the old version was broken whenever the file contains any new values. We just need to make sure it works going forward. Your new 0.70 version presumably isn't using `[=[` because this version doesn't unless necessary.

I see. Pity. Ok then.
Then I think we can detach 3224 from this PR and simply close 3224 manually as "not planned".
That would probably be the right approach.

PR #3255, opened 2026-03-02
Machines is being repaired and accepting patients simultaneously fix

### LINKED ISSUE #2949: [Bug] Paging Handyman to repair machine does not visibly mark it for repair

### LINKED ISSUE #2996: X-ray is being repaired and accepting patients simultaneously
X-ray is being repaired and accepting patients simultaneously

v0.68.0

Screenshot 2024-09-26 at 22 25 42

Moreover, there is second handyman who wants to fix X-Ray too. 😁

[xray_reapearing_and_using_at_the_same_time.sav.zip](https://github.com/user-attachments/files/17155095/xray_reapearing_and_using_at_the_same_time.sav.zip)

----------------

v0.69.0

[x-ray_is_repaired_and_used_at_the_same_time.sav.zip](https://github.com/user-attachments/files/21666543/x-ray_is_repaired_and_used_at_the_same_time.sav.zip)

Image

----------------

Issue separated from #2630

Possibly related #70

ARGAMX:
*Fixes #2996*
*Fixes #2949*

- Fixed bug when room was not locked against patients in case when handyman entered to repair machine on his own. (without user 'call handyman' button press)
- Some refactoring related to locking/unlocking room.

The room is now locked to patients if:
1. The "Call a handyman" button was pressed. or
2. A handyman has arrived to repair the machine and currently is in the room.

The room unlocks if:
1. The handyman grabbed from the room, but only for case when the "Call a handyman" button was not pressed. or
2. The repair is complete.
3. Room exploded or edited.

Issue can be reproduced and tested in this save on X-ray.
[x-ray_reapearing_and_using_at_the_same_time_v0.69.sav.zip](https://github.com/user-attachments/files/25688674/x-ray_reapearing_and_using_at_the_same_time_v0.69.sav.zip)

### DIFF
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -1214,7 +1214,7 @@ function Room:getRoomMachine()
   end
 end
 
---! Locks the room to prevent patient from entering due to repair in roon.
+--! Locks the room to prevent patient from entering due to repair in room.
 function Room:lockRoomOnRepair()
   self.needs_repair = true
 end


CorsixTH/Lua/entities/machine.lua
```diff
@@ -433,11 +436,9 @@ function Machine:setRepairing(handyman, is_manual_repair)
 end
 
 function Machine:setRepairingMode(lock_room, is_manual_repair)
-  if lock_room ~= nil then
-    self.repairing_lock_room = lock_room
```
ARGAMX, line 437: this `self.repairing_lock_room` was never user outside of this function.
I don't understand why it was stored in the object property.
so I removing it.

CorsixTH/Lua/room.lua
```diff
@@ -1182,6 +1203,27 @@ function Room:calculateHappinessFactor()
...
+  for object, _ in pairs(self.world:findAllObjectsNear(fx, fy)) do
+    if object.strength ~= nil then
+      return object
+    end
+  end
+end
+
+--! Locks the room to prevent patient from entering due to repair in roon.
+function Room:lockRoomOnRepair()
+  self.needs_repair = true
```
ARGAMX, line 1219: `self.needs_repair = true` preventing patient from entering the room:

```Lua
function Room:canHumanoidEnter(humanoid)
  ...
  if class.is(humanoid, Patient) and not self.needs_repair then
    return true
  end
  ...
  return false
end
```

CorsixTH/Lua/entities/machine.lua
```diff
@@ -418,11 +418,14 @@ function Machine:setRepairing(handyman, is_manual_repair)
   local anim = {icon = 4564} -- The only icon for machinery
   local room = self:getRoom()
   local should_repair = handyman or is_manual_repair
+  room.manual_repair = handyman and is_manual_repair
+  -- put/remove gear icon above machine
   self:setMoodInfo(should_repair and anim or nil)
-  room.needs_repair = should_repair
```
```Lua
local should_repair = handyman or is_manual_repair
room.needs_repair = should_repair
```
This code was supposed to lock the room.
But it doesn't work properly.

The problem here is that `handyman` can be `nil` on `function Machine:setRepairing(handyman, is_manual_repair)` call.
I haven't investigated why this happens.

So it turns out that the room was securely locked only if `is_manual_repair` is true.

CorsixTH/Lua/room.lua
```diff
@@ -341,6 +341,14 @@ function Room:onHumanoidEnter(humanoid)
     -- the handyman didn't arrive as a part of a job
     if humanoid.on_call then
       assert(humanoid.on_call.object:getRoom() == self, "Handyman arrived is on call but not arriving to the designated room")
+      local machine = self:getRoomMachine()
+      if machine then
+        local handyman = machine.repairing
+        if humanoid == handyman then
+          -- The handyman arrived to fix the machine. Should lock the room to keep patients out.
+          self:lockRoomOnRepair()
```
lewri, line 349: Does the room unlock successfully if you picked up the handyman out the room (both cases of either before or after they repair the machine)?

lewri: I can see actually how it unlocks anyway after repair. Just can't find the before yet.

NB: Repair high priority would always lock the door actually if I remember. Low priority?

ARGAMX: > Does the room unlock successfully if you picked up the handyman out the room (both cases of either before or after they repair the machine)?

Of course.
Once the machine is repaired, the room is unlocked, so grab the handyman no longer matters.

But if handyman is removed/leaved before the repair is completed, then the room will be unblocked by the code marked with this comment:
> -- The handyman arrived to fix the machine was grabbed from the room before repair finished.

ARGAMX: > NB: Repair high priority would always lock the door actually if I remember. Low priority?

The priority of repair does not affect the room locking in any way.

lewri: Check here as the code indicates it can:

https://github.com/CorsixTH/CorsixTH/blob/6108b6ffa762795d91c0f3c9b0676b29f28da7ff/CorsixTH/Lua/entities/machine.lua#L170-L180

And here:
https://github.com/CorsixTH/CorsixTH/blob/6108b6ffa762795d91c0f3c9b0676b29f28da7ff/CorsixTH/Lua/calls_dispatcher.lua#L94-L100

Manual should already set a lock_room to true
I'm not saying you're wrong, but it looks like the dispatcher also doe some stuff with locking room (though I think its locking room when the handyman isn't present); just wanted a sanity check.

ARGAMX: > but it looks like the dispatcher also doe some stuff with locking room (though I think its locking room when the handyman isn't present); just wanted a sanity check.

I double checked and
Sorry, I couldn't find any confirmation of this in the code and in the debug.
Only that comment claim this.

But the actual code doesn't do this.
No matter the priority, the room doesn't lock properly.

What's also interesting is that even if this priority system would work properly, the room would still need to be locked at a low priority!
Otherwise, the bug would be still reproducible at low priority.
And thus this mechanism is still invalid.

All that priority affects is the setting value `self.repairing_lock_room`.
But as I mentioned above, this value of `repairing_lock_room` isn't read anywhere where the room should be locked.

Perhaps somewhere along the way, the read of `repairing_lock_room` was lost, thus breaking this mechanism.

But thank you for pointing this out. 👍
This made me realise that this comment and the fragments of priorities that aren't working need to be removed too. Not just `repairing_lock_room` variable.

CorsixTH/Lua/room.lua
```diff
@@ -1182,6 +1203,27 @@ function Room:calculateHappinessFactor()
...
+function Room:getRoomMachine()
+  local fx, fy = self:getEntranceXY(true)
+  for object, _ in pairs(self.world:findAllObjectsNear(fx, fy)) do
+    if object.strength ~= nil then
+      return object
+    end
+  end
+end
+
+--! Locks the room to prevent patient from entering due to repair in roon.
```
ARGAMX:
```suggestion
--! Locks the room to prevent patient from entering due to repair in room.
```

CorsixTH/Lua/entities/machine.lua
```diff
@@ -410,34 +408,30 @@ function Machine:reduceStrengthOnRepair()
...
 end
 
---! Tells the machine to start showing the icon that it needs repair.
---!   also lock the room from patient entering
---!param handyman The handyman heading to this machine. nil if repairing is finished
---!param is_manual_repair (bool) true if the repairing mode was set manually through the dialog; nil otherwise
-function Machine:setRepairing(handyman, is_manual_repair)
-  local anim = {icon = 4564} -- The only icon for machinery
+--! Set the room's repair status manually. Lock room and show gear icon.
+function Machine:setRepairingStatus()
```
lewri, line 412: `setManualRepair` ? There is no requirement anymore for this status to be set if the repair isn't manual. But I can see why this format was chosen.

ARGAMX:
> It looks like we no longer set the gear icon when the handyman is present? Was that intended? I would argue yes, original doesn't do this.

Did we ever set it based on the presence of a handyman? I'm not quite sure what you mean here.

Maybe in CorsixTH, when a handyman came in for repairs without pressing a button, the gear wheel would also show up? I'm not sure, I don't remember, and I haven't checked. But it doesn't match the original TH.

In the original TH, the gear appears when the handyman button is pressed. 
The gear disappears when repair are completed or new equipment is purchased.

The gear could also disappear during animations—whether it was being used or repaired. As far as I remember, this defect possible in both CorsixTH and the original TH.

Handyman themselves don't affect the appearance of gear in the original TH.
But in CorsixTH version 0.68, for example, seems handyman did affect the appearance, and we tried to remove this in one of our PRs (PR 3006).

As a result of 3006 and 3255, I think the gear display should be in line with the original game.

> 
> Also, I can't remember if previously discussed but manual repair does not prevent patient entry in the original. In fact, locking room isn't even a thing. 
>
> Image

My tests on original TH indicates that manual repair does prevent patient entry.
But
If button pressed after the patient leaved the queue and started his walk toward to the room, then he will enter the room.
So looks like the original TH do not performs this check on actually door use.

CorsixTH performs this check multiple times. Before knocking on the door. Before opening the door. Before leaving the queue. So, a patient won't be able to enter CorsixTH even if they've already left the queue. This is a slight deviation from the original, of course, but it doesn't bother me too much for now.

So if you pressed the button in TH too late, I think we are ok with the result you got.

> 
> However, locking the room on handyman entry I do think makes sense. The behavioural realignment change for locking at start of manual repair needs to be its own issue.

In the original TH, I couldn't get the next patient to enter after the handyman had entered for repairs and before they'd finished the repair. Patients wouldn't enter. This was my first priority in this PR.

Regarding to manual repair, I believe that behavior for locking at start of manual repair works pretty close to the original TH in this PR. Maybe something can be improved here, but at least I haven’t understand it yet.

CorsixTH/Lua/entities/machine.lua
ARGAMX: > `setManualRepair` ? There is no requirement anymore for this status to be set if the repair isn't manual. But I can see why this format was chosen.

Yes, but then it will not be obvious that the function `Machine:removeRepairingStatus()` is the opposite of this function.
On the other hand, `Machine:removeRepairingStatus()` is always called, both after manual and automatic repairs.
And `Machine:setRepairingStatus()` is only called on manual repairs.
Honestly, I don't know which is better.
The function's comment states that this is for manual repair. I think that's sufficient.
I guess I'll leave it that way.

PR #3261, opened 2026-03-07
Various sprite viewer enhancements

TheCycoONE:
Fix crash on Russian version with missing dat files

Add the ability to switch background colours, e.g. to see black fonts.

Add the ability to scale the sprites for a better view.

image

### DIFF
--- a/CorsixTH/Lua/sprite_viewer.lua
+++ b/CorsixTH/Lua/sprite_viewer.lua
@@ -41,10 +41,11 @@ local y_off
 local old_event_handlers
 
 local background_colours = {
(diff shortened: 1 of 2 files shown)


CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -603,6 +603,15 @@ bool render_target::fill_black() {
   return true;
 }
 
+bool render_target::fill_colour(uint32_t colour) {
```
TheCycoONE, line 606: fillBlack could call this, but I thought it was worth leaving the fractionally cheaper specialization in for the common case.

PR #3263, opened 2026-03-07
Fix bottom panel bar offset

Matroftt:
image

- if machine menu button is enabled, then move progress bar to the right side of panel, as it is without button

### DIFF
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -316,6 +316,9 @@ function UIBottomPanel:drawDynamicInfo(canvas, x, y)
     return
   end
 
+  local info = self.dynamic_info
+  local font = self.white_font
+
   local dynamic_info_panel_x = 364
   local progress_bar_r_offset = 11
   local progress_bar_width = 100


Alberth289346:
Why not compute the position relative to the right edge of the window?

Then it is not dependent on the specific menu item, and will continue to work if we ever change the button list again in the future.

Alberth289346:
Note sure if `Tiredness` is at the correct spot.

Visually it seems to be part of the descriptive text at the left at first glance. It is not immediately clear that it belongs with the horizontal bar, I think.

Maybe it can be improved as well?

ARGAMX:
> Note sure if `Tiredness` is at the correct spot.
> 
> Visually it seems to be part of the descriptive text at the left at first glance. It is not immediately clear that it belongs with the horizontal bar, I think.
> 
> Maybe it can be improved as well?

The original game has exactly the same layout.
Now that you've pointed it out, I also see that it's odd.
But being accustomed to the original game, it looks expected and familiar.
And perhaps moving it somewhere else could have the opposite effect.

Matroftt:
> Note sure if `Tiredness` is at the correct spot.
> 
> Visually it seems to be part of the descriptive text at the left at first glance. It is not immediately clear that it belongs with the horizontal bar, I think.
> 
> Maybe it can be improved as well?

I agree, Tiredness label moved to right side does look nicer atleast for me
If we are moving tiredness then we should also move this: 
{77C41174-D296-4A66-AB74-5148DBECE826}

But, guess it is not getting implemented and labels are being kept as they are currently

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -318,18 +318,21 @@ function UIBottomPanel:drawDynamicInfo(canvas, x, y)
 
   local info = self.dynamic_info
   local font = self.white_font
+
+  local w = x + (self.width - 364) * s
```
lewri: One magic with another :D 
I assume you mean this is to calculate the start of the progress bar? Maybe `progress_bar_start = 364` (if that is what this is for).

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -318,18 +318,21 @@ function UIBottomPanel:drawDynamicInfo(canvas, x, y)
...
   for i, text in ipairs(info["text"]) do
     font:drawWrapped(canvas, text, x + 20 * s, y + 10 * i * s, 240 * s)
     if i == #info["text"] and info["progress"] then
       local white = canvas:mapRGB(255, 255, 255)
       local black = canvas:mapRGB(0, 0, 0)
       local orange = canvas:mapRGB(221, 83, 0)
-      canvas:drawRect(white, x + 165 * s, y + 10 * i * s, 100 * s, 10 * s)
-      canvas:drawRect(black, x + 166 * s, y + s + 10 * i * s, 98 * s, 8 * s)
-      canvas:drawRect(orange, x + 166 * s, y + s + 10 * i * s, math.floor(98 * info["progress"] * s), 8 * s)
+      canvas:drawRect(white, w - 109 * s, y + 10 * i * s, 100 * s, 10 * s)
```
lewri: The 1 pixel difference is annoying, but not you. These lines are in too deep in magic for me to consider it in scope to refactor.

CorsixTH/Lua/dialogs/bottom_panel.lua
Matroftt: there is going to be alot of local variables

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -316,20 +316,24 @@ function UIBottomPanel:drawDynamicInfo(canvas, x, y)
     return
   end
 
-  local info = self.dynamic_info
-  local font = self.white_font
```
lewri, line 320: Oops

PR #3268, opened 2026-03-10
Support mousewheel in dropdown scrollbars

lewri:
<details>
  <summary></summary>
  
<sup>dropdown_scrollbar2: electric boogaloo</sup>
  
</details>

- You can use your mousewheel on the scrollbars in dropdowns. It scrolls 1 item at a time in this format.
- Mouse must be in the dropdown to scroll, to prevent clashing with anything else (though unlikely).

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/dropdown.lua
+++ b/CorsixTH/Lua/dialogs/resizables/dropdown.lua
@@ -161,7 +161,7 @@ end
 --!param x (int) UNUSED, 1 on right scroll, -1 on left scroll
 --!param y (int) 1 on up scroll, -1 on down scroll
 function UIDropdown:onMouseWheel(x, y)
-  if x ~= 0 then return false end -- Do nothing on x scroll
+  if y == 0 then return false end -- Do nothing on x only scroll
   local bar = self.scrollbar
 
   if not self:hitTest(self.cursor_x, self.cursor_y) or not (bar and bar.enabled) then


CorsixTH/Lua/dialogs/resizables/dropdown.lua
```diff
@@ -157,6 +157,35 @@ function UIDropdown:onMouseDown(button, x, y)
   return UIResizable.onMouseDown(self, button, x, y)
 end
 
+--! Allow the dropdown to scroll through mousewheel input
+--!param x (int) UNUSED, 1 on right scroll, -1 on left scroll
+--!param y (int) 1 on up scroll, -1 on down scroll
+function UIDropdown:onMouseWheel(x, y)
+  if x ~= 0 then return false end -- Do nothing on x scroll
```
TheCycoONE: Maybe safer with if y == 0; or abs(y) < abs(x).

A mouse wheel will only do one or the other, but I understand other devices like touchpads or magic mice might change both in a scroll event.

lewri: Confirmed on my laptop. It can do a x = 1, y = -1 for example. Will update

CorsixTH/Lua/dialogs/resizables/dropdown.lua
```diff
@@ -157,6 +157,35 @@ function UIDropdown:onMouseDown(button, x, y)
   return UIResizable.onMouseDown(self, button, x, y)
 end
 
+--! Allow the dropdown to scroll through mousewheel input
+--!param x (int) UNUSED, 1 on right scroll, -1 on left scroll
+--!param y (int) 1 on up scroll, -1 on down scroll
+function UIDropdown:onMouseWheel(x, y)
+  if y == 0 then return false end -- Do nothing on x only scroll
```
lewri, line 164: It seems a limitation in potentially our implementation of SDL's mousewheel event we don't need to worry about diagonal or float based scrolling.

Edit: confirmed, we push an int and not the float value.
https://github.com/CorsixTH/CorsixTH/blob/4dc8f3a0334c3cca933f7c38a1fddf90a4aa21e8/CorsixTH/Src/sdl_core.cpp#L191-L196

PR #3271, opened 2026-03-12
Fix palette detection for sprite viewer

TheCycoONE:
There was a false assumption that PAL names matched the sprite table names. This is not true, though they generally start with the same word.

Most sprite tables that do not have a matching palette use MPalette.dat - fonts are a notable exception but there is no way to auto-detect the palette for a font.

### DIFF
--- a/CorsixTH/Lua/sprite_viewer.lua
+++ b/CorsixTH/Lua/sprite_viewer.lua
@@ -39,6 +39,7 @@ local scale = 1
 local bg_colour_index = 1
 local y_off
 local old_event_handlers
+local mpalette_index
 
 local background_colours = {
   Colours.Black,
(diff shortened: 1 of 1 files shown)


lewri:
By chance how did we do the palette for font before? Or didn't we do anything?

I wonder if some might need finding and hinting to the viewer. E.g. Font34V and Font31V seems to use Town01V going by the town map file.

ARGAMX:
This isn't related to this PR at all, just an idea about the sprite viewer:

Could we also make it possible to exit the sprite viewer by pressing `Esc`?
I don't know why it's done this way, with `Q` being the only way to exit.
It's not intuitive at all.

Also, if you have anything other than English as your input language, and it's not currently selected, then even pressing `Q` won't work. You need to change the language to English first and only then press `Q`.
And `Esc` is the `Esc` key in all keyboard layouts.
These exit difficulties remind me of Vim. 😁

TheCycoONE:
> By chance how did we do the palette for font before? Or didn't we do anything?
> 
> I wonder if some might need finding and hinting to the viewer. E.g. Font34V and Font31V seems to use Town01V going by the town map file.

From what I can see in the code this never worked, the broken exact filename check is more than 10 years old. It would have been passing a nil, and using the cached sprite table from graphics.lua or MPalette.dat for everything.

lewri:
Ah SPointer uses multiple palettes.

PR #3273, opened 2026-03-14
Prevent handyman take tasks to repair already repairing machine

### LINKED ISSUE #2630: [Bug] X-ray. Handyman take task to fix machine which is already being fixing
### Describe the issue

If one Handyman already repairing the machine, then other Handyman can take the same task to fix that machine.

Screenshot 2024-09-22 at 21 11 29

### Steps to Reproduce

2. You will see X-Ray room and two Handymans in it. First Handyman fixing the X-Ray machine. Second Handyman is heading to fix the same X-Ray machine. 
3. While the First Handyman is still fixing the X-Ray machine, take the Second Handyman that is walks around the X-Ray room.
4. While the First Handyman is still fixing the X-Ray machine, put the Second Handyman outside the room.
5. Observe, that Second Handyman outside is "Going to repair X-Ray".

This problem also reproduces without player intervention.

Bit later one of them will go to fix Blood Machine. And the same bug situation can repeat.

### Expected Behaviour

The Second Handyman should be unable to take occupied task.

### Save Game

[several_janitors_taken_task_to_fix_rentgen.sav.zip](https://github.com/user-attachments/files/17090693/several_janitors_taken_task_to_fix_rentgen.sav.zip)

### CorsixTH Version

0.68.0 Beta 1

### Operating System

MacOS Sonoma 14.5

ARGAMX:
*Fixes #2630*

- If `assignedDistance == false` make additional checks before take repair task from another handyman.

### DIFF
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -2086,6 +2086,7 @@ function Hospital:searchForHandymanTask(handyman, taskType)
           print("Warning: Orphaned handyman is still assigned a task. Removing.")
           v.assignedHandyman:unassignTask()
         else
+          -- Check if handyman are closer to the task than the assigned handyman.
           local assignedDistance = self.world:getPathDistance(v.tile_x, v.tile_y, v.assignedHandyman.tile_x, v.assignedHandyman.tile_y)
           if assignedDistance ~= false then
             if v.assignedHandyman.profile.is_consultant then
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/hospital.lua
```diff
@@ -2099,6 +2100,9 @@ function Hospital:searchForHandymanTask(handyman, taskType)
             else
               distance = distance / multiplier
             end
+          elseif taskType == "repairing" and v.assignedHandyman:getRoom() == self.world:getRoom(v.tile_x, v.tile_y) then
```
lewri, line 2103: Huh, didn't realise object wasn't aware what room it was in (likely because of litter -- which interestingly also has no call).

lewri (approved):

PR #3274, opened 2026-03-14
Move empty table check to utility.

Alberth289346:
Also, a slower function that was used for the same purpose was removed.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -29,9 +29,9 @@ globals = { -- Globals
             "loadfile_envcall", "loadstring_envcall",
             "pause_gc_and_use_weak_keys", "permanent",
             "rangeMapLookup", "rnc", "strict_declare_global",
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/app.lua
```diff
@@ -28,7 +28,7 @@ local SDL = require("sdl")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 241 -- Load palette on init
+local SAVEGAME_VERSION = 242 -- Move empty table check to utility
```
lewri: Doesn't look like version bump is needed for this one?

lewri (approved):
Just a sanity check on the version bump's necessity

Alberth289346:
Rebased, dropped the lua version increment

PR #3275, opened 2026-03-16
Update UIOptions to Colours

lewri:
*Fixes #*

- Moves colour definitions in file over to their Colours constants
- Has removed the highlight/shadow from textboxes (now pure black)
- Applies SettingActive to settings that launch a child window while it is open

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/options.lua
+++ b/CorsixTH/Lua/dialogs/resizables/options.lua
@@ -28,34 +28,17 @@ local UIOptions = _G["UIOptions"]
 local BTN_WIDTH = 135
 local BTN_HEIGHT = 20
 
-local col_bg = {
-  red = 154,
-  green = 146,
-  blue = 198,
-}
-
-local col_textbox = {
-  red = 0,
-  green = 0,
-  blue = 0,
-}
-
-local col_highlight = {
-  red = 174,
-  green = 166,
-  blue = 218,
-}
-
-local col_shadow = {
-  red = 134,
-  green = 126,
-  blue = 178,
-}
-
-local col_caption = {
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -165,56 +148,56 @@ function UIOptions:UIOptions(ui, mode)
...
     local updates_y_pos = self:_getOptionYPos()
     local updates_string = app.config.check_for_updates and
         _S.options_window.option_enabled or _S.options_window.option_disabled
-    self:addBevelPanel(20, updates_y_pos, BTN_WIDTH, BTN_HEIGHT, col_shadow, col_bg, col_bg)
+    self:addBevelPanel(20, updates_y_pos, BTN_WIDTH, BTN_HEIGHT, col.caption, col.bg, col.bg)
         :setLabel(_S.options_window.check_for_updates):setTooltip(_S.tooltip.options_window.check_for_updates).lowered = true
     self.updates_panel =
-        self:addBevelPanel(165, updates_y_pos, BTN_WIDTH, BTN_HEIGHT, col_bg):setLabel(updates_string)
+        self:addBevelPanel(165, updates_y_pos, BTN_WIDTH, BTN_HEIGHT, col.setting):setLabel(updates_string)
     self.updates_button = self.updates_panel:makeToggleButton(0, 0, 140, BTN_HEIGHT, nil, self.buttonUpdates)
```
lewri, line 163: There is a bug here (and line 173) that should use `BTN_WIDTH` instead of `140` (allows you to activate the button outside the bounds). It'll be fixed in the next UIOptions PR after this.

PR #3280, opened 2026-03-18
UIOptions Rework

lewri:
*Fixes #*

- Commit 1 fixes the bug noted in #3275 
- Commit 2 consolidates the repeated window element generation of the settings buttons in this window to a reusable helper
- Commit 3 shows unavailable resolutions as disabled with a tooltip.

image

Query: Should Custom be brought to be above the disabled items?

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/options.lua
+++ b/CorsixTH/Lua/dialogs/resizables/options.lua
@@ -94,7 +94,7 @@ local available_resolutions = function()
   local enable_list, disable_list = {}, {}
   for _, opt in ipairs(suggested_resolutions) do
     local enabled = App.MIN_WINDOW_WIDTH * s <= opt.width and
-    App.MIN_WINDOW_HEIGHT * s <= opt.height
+        App.MIN_WINDOW_HEIGHT * s <= opt.height
     opt.disabled = not enabled
     opt.tooltip = opt.disabled and { _S.tooltip.options_window.resolution_unavailable }
     if enabled then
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -81,14 +90,25 @@ local available_resolutions = function()
...
   }
 
-  local res = {}
   local s = TheApp.config.ui_scale
+  local enable_list, disable_list = {}, {}
   for _, opt in ipairs(suggested_resolutions) do
-    if App.MIN_WINDOW_WIDTH * s <= opt.width and App.MIN_WINDOW_HEIGHT * s <= opt.height then
-      res[#res + 1] = opt
+    local enabled = App.MIN_WINDOW_WIDTH * s <= opt.width and
+    App.MIN_WINDOW_HEIGHT * s <= opt.height
```
tobylane:
```suggestion
        App.MIN_WINDOW_HEIGHT * s <= opt.height
```

ARGAMX:
> > Query: Should Custom be brought to be above the disabled items?
> 
> Yes. It's most discoverable at the top of the list, but maybe everyone who wants it would scroll down the list as far as the first disabled option, so Custom only needs to be above that.

I agree with that.
I'm also in favor of it being at the very top of the list.

lewri:
Have moved to be above disabled items, to give a visual hint that there are incompatible resolutions before they enter a custom one.

Future todo if someone wants is to make the custom window give some more info/edit the error text to mention ui scaling.

PR #3290, opened 2026-03-22
Some code refactoring for handyman

ARGAMX:

Fixed currently Lua crash on handyman task search.
Also some refactoring.

Fixes in the logic:

If the handyman task is released, then release the corresponding call in dispatcher.
Previously, task releases was able to go unsync with call in dispatcher releases.

Various thoughts:

Seems that the essence of a task for a handyman is split between objects in the hospital and objects in the dispatcher.
`Hospital` class manages the `handymanTasks` array. 
`CallDispatcher` class manages the `call_queue`.
For me it looks like a duplication.
Tasks should have a single storage location, and this should also be the single source of truth regarding tasks.
I didn't do anything about it.

Also in `CallDispatcher` there is `setNextAction` calls thereby `CallDispatcher` working with the humanoid action queue directly which is not good.

### DIFF
--- a/CorsixTH/Lua/calls_dispatcher.lua
+++ b/CorsixTH/Lua/calls_dispatcher.lua
@@ -468,7 +468,7 @@ end
 --! Unassign call
 --!param call call to unassign from assigned humanoid.
 --!param answer_next_call (bool) should the humanoid assigned to this call then
--- answer for another new call or not
+-- answer for another new call
 function CallsDispatcher.unassignCall(call, answer_next_call)
   local assigned = call.assigned
   assert(assigned.on_call == call, "Unassigning call but the staff was not on call or a different call")
--- a/CorsixTH/Lua/entities/humanoids/staff/handyman.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff/handyman.lua
@@ -96,7 +96,8 @@ function Handyman:onPickup()
 end
 
 -- Deattaches the handyman from the task and forces him to search for a new one.
---!param reset_action_queue (bool) Should the worker interrupt the current action to search for a new task
+--!param reset_action_queue (bool) Should the handyman interrupt
+-- the current action to search for a new task
 function Handyman:releaseHandymanFromTask(reset_action_queue)
   self:setDynamicInfoText("")
   self:unassignTask()
@@ -167,10 +168,13 @@ function Handyman:doMeandering()
   self:queueAction(MeanderAction())
 end
 
-function Handyman:assignHandymanTask(taskIndex, taskType)
-  local task_object = self.hospital:assignHandymanToTask(self, taskIndex, taskType)
+-- Assign a handyman to a task
+--!param task_index (integer) task index
+--!param task_type (string) task type
+function Handyman:assignHandymanTask(task_index, task_type)
+  local task_object = self.hospital:assignHandymanToTask(self, task_index, task_type)
   if task_object then
-    if taskType == "cleaning" then
+    if task_type == "cleaning" then
       self:processCleaningTask(task_object)
     else
       if task_object.call.dropped then
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -2061,7 +2061,7 @@ function Hospital:searchForHandymanTask(handyman, task_type)
   local index = -1
   local priority = 0
   local multiplier = 1
-  local found_any_task = false
+  local first_attempt = true
 
   if handyman.profile.is_consultant then
     multiplier = 0.5
@@ -2075,7 +2075,7 @@ function Hospital:searchForHandymanTask(handyman, task_type)
     local v = task_candidate
     local distance = self.world:getPathDistance(v.tile_x, v.tile_y, handyman.tile_x, handyman.tile_y)
     local found_suitable_task = true
-    if found_any_task and v.priority < priority then
+    if not first_attempt and v.priority < priority then
       found_suitable_task = false
     end
     if not v.parcelId then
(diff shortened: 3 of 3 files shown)


CorsixTH/Lua/entities/humanoids/staff/handyman.lua
```diff
@@ -85,105 +82,146 @@ function Handyman:fulfillsCriterion(criterion)
   return criterion == "Handyman"
 end
 
-function Handyman:afterLoad(old, new)
-  if old < 163 then
-    self.leave_priority = AnnouncementPriority.High
-    self.leave_sounds = {"sack006.wav"}
-  end
-  Staff.afterLoad(self, old, new)
-end
```
ARGAMX, line 94: Just moved to the end of the file.

CorsixTH/Lua/calls_dispatcher.lua
```diff
@@ -27,7 +27,7 @@ class "CallsDispatcher"
 ---@type CallsDispatcher
 local CallsDispatcher = _G["CallsDispatcher"]
 
-local debug = false -- Turn on for debug message
+local debug_enabled = false -- Turn on for debug message
```
ARGAMX, line 30: Variable name "debug" have collision with the other global variable "debug". 
So on calling `TheApp.world:gameLog(debug.traceback())` in this class you get an error.

CorsixTH/Lua/entities/humanoids/staff/handyman.lua
```diff
@@ -85,105 +82,146 @@ function Handyman:fulfillsCriterion(criterion)
...
+          task_found = true
         end
       end
     end
   end
-  if assignedTask == false then
-    -- Make sure that the handyman isn't meandering already.
-    for _, action in ipairs(self.action_queue) do
-      if action.name == "meander" then
-        return false
```
lewri, line 150: Doesn't staff.lua have a function for this already? I.e. `Staff:isMeandering()`

CorsixTH/Lua/calls_dispatcher.lua
```diff
@@ -465,12 +465,15 @@ function CallsDispatcher:dropFromQueue(object, key)
   self:onChange()
 end
 
-function CallsDispatcher.unassignCall(call)
+function CallsDispatcher.unassignCall(call, answer_next_call)
+  if answer_next_call == nil then answer_next_call = true end
```
lewri: Can we add doc for this one please :)
Though, the `nil == true` is a little irksome; getting direct zoom flashbacks. I'm ok to leave it though.

Any chance we could push existing unassign calls to have true and false explicitly defined.

CorsixTH/Lua/entities/humanoids/staff/handyman.lua
```diff
@@ -85,105 +82,146 @@ function Handyman:fulfillsCriterion(criterion)
...
+function Handyman:unassignCall()
+  if self.on_call then
+    CallsDispatcher.unassignCall(self.on_call, false)
+  end
+end
+
+-- Set task type priority value
+--!param task_type (string) task type
+--!param priority (float) priority to set
+function Handyman:setPriority(task_type, priority)
```
lewri: Ideally new functions would be nicer to be nearer where they are used.

CorsixTH/Lua/hospital.lua
```diff
@@ -2018,40 +2019,41 @@ end
...
+      former_handyman:releaseHandymanFromTask(true)
     end
+    handyman.task = task_object
+    task_object.assignedHandyman = handyman
+    return task_object
   end
+  return nil
 end
 
+--!return index (int) The task index.
```
lewri: Assuming we didn't get any further in fully documenting this function. Sure, it is a horrid one.

CorsixTH/Lua/hospital.lua
```diff
@@ -2018,40 +2019,41 @@ end
...
 function Hospital:searchForHandymanTask(handyman, taskType)
   local subTable = self:findHandymanTaskSubtable(taskType)
   --if a distance is smaller than this value stop the search to
   --save performance
   local thresholdForStopping = 3
-  local first, dist, index, priority, multiplier = true, 0, -1, 0, 1
+  -- information about the best task found
+  local found_any_task, dist, index, priority = false, 0, -1, 0
```
TheCycoONE: Since you're refactoring this line anyway it can be split to multiple lines.

CorsixTH/Lua/entities/humanoids/staff/handyman.lua
ARGAMX: > Ideally new functions would be nicer to be nearer where they are used.

Yes, but
It's just a typical one-line getter/setter.

I moved it to the top of the file, but overall, I don't see much of a difference.

CorsixTH/Lua/hospital.lua
ARGAMX: > Assuming we didn't get any further in fully documenting this function. Sure, it is a horrid one.

Oops, I must have fallen asleep halfway through. 
I have added a proper description. 👍

CorsixTH/Lua/hospital.lua
ARGAMX: > Since you're refactoring this line anyway it can be split to multiple lines.

Done 👌

CorsixTH/Lua/calls_dispatcher.lua
ARGAMX: > Can we add doc for this one please :) Though, the `nil == true` is a little irksome; getting direct zoom flashbacks. I'm ok to leave it though.
> 
> Any chance we could push existing unassign calls to have true and false explicitly defined.

Okay, I added the explicit parameter everywhere. 👌

Also some doc added. 👌

CorsixTH/Lua/entities/humanoids/staff/handyman.lua
ARGAMX: > Doesn't staff.lua have a function for this already? I.e. `Staff:isMeandering()`

Indeed. 
And what's most interesting is that I wrote this function myself. 
I replaced this with this one call.

Issue #3286, opened 2026-03-21
[Bug] Fluidsynth generating warnings, jukebox slower to initialise


lewri:
### Describe the issue

In Windows, when CorsixTH loads up the mainmenu after the intro movie ends it generates warnings from Fluidsynth.

### Steps to Reproduce

1. Load game on Windows, where MIDI is set to Software.

### Expected Behaviour

No warnings should generate. I also wonder if this is what causes a delay in the Jukebox starting in master (0.69.2 is instant).

### Have you enabled any current CorsixTH Experimental Features?

- [ ] Right mouse panning
- [x] Falling actions (applies to CorsixTH versions < 0.70.0)

### Save Game

Save game attached to report: No

### CorsixTH Version

Recent master

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
fluidsynth: warning: Modulator with source 1 set to none api fluid_synth_add_default_mod mod.src1=0
```

TheCycoONE:
As I said in the channel these are harmless new warnings added in fluidsynth to catch soundfonts that are not fully spec compliant.

Don't combine these issues. Unless newer fluidsynth is also slower (maybe) they are unrelated.

TheCycoONE:
[...]

PR #3291, opened 2026-03-23
Music fixes

TheCycoONE:
A collection of fixes for various music issues, found while trying to diagnose #3286 

That issue is not resolved, but #3309 is.

Each commit is a different fix, so keep the commit messages.

### DIFF
--- a/CorsixTH/Lua/audio.lua
+++ b/CorsixTH/Lua/audio.lua
@@ -839,5 +839,6 @@ function Audio:destroy()
   self.has_bg_music = false
   self.not_loaded = not TheApp.config.audio
   self.speech_file_name = nil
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/audio.lua
```diff
@@ -193,6 +193,10 @@ function Audio:init()
     self.background_playlist = {}
     return
   end
+
+  -- Prevent lazy-init of the midi subsystem from causing delays later
+  local silent_mid = SDL.audio.loadSilentMidi()
```
lewri: If this is run in async I'm pretty sure this would 'mask' the windows issue, at least for the ones that load the movie.

PR #3293, opened 2026-03-24
Bump vcpkg and improve cmake install

TheCycoONE:
Bump vcpkg to the 2026-03 versions of packages

If only it was that simple. The new vcpkg was incompatible with the run-cmake/run-vcpkg github actions. They have an open bug report, but between that and the recent github action supply chain attack I thought it would be better to just make our own actions.

In making our own actions I found problems with our cmake --install for window targets; dlls weren't copied at all when run as it's own phase, and the paths were all wrong; so there are some cmake fixes for that.

After fixing the installer most of the post install pipeline for setting up the artifact went away, so I finished the job by adding a cmake option to fetch the soundfont.

All told this is hopefully easier to maintain.

There is a lingering vcpkg cache issue with github actions, but I traced that down to github supplying us slightly different runners with slightly different compiler versions - enough for the ABI check in vcpkg to fail/say the cache doesn't match.

It's not huge, but commit history has been pruned and are intended to be read individually.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -50,7 +50,8 @@
       "cacheVariables": {
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
         "VCPKG_TARGET_TRIPLET": "x64-windows-release",
-        "ENABLE_UNIT_TESTS": "ON"
+        "ENABLE_UNIT_TESTS": "ON",
+        "FETCH_SOUNDFONT": "ON"
       }
     },
     {
@@ -61,7 +62,8 @@
       "cacheVariables": {
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
         "VCPKG_TARGET_TRIPLET": "x86-windows",
-        "ENABLE_UNIT_TESTS": "ON"
+        "ENABLE_UNIT_TESTS": "ON",
+        "FETCH_SOUNDFONT": "ON"
       }
     },
     {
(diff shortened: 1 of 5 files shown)


CMakePresets.json
```diff
@@ -100,6 +102,11 @@
       "configurePreset": "linux-dev-vcpkg",
       "configuration": "Debug"
     },
+    {
+      "name": "win-dev",
+      "configurePreset": "win-dev",
```
lewri, line 107: Does the SDL2 stuff no longer fall over in debug? I believe we had a long standing note in the compile options about something with debug in Windows.

TheCycoONE:
Yeah will wait on that. Thanks @lewri

CMakePresets.json
TheCycoONE: I think that's solved.

CMakePresets.json
```diff
@@ -50,7 +50,8 @@
       "cacheVariables": {
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
         "VCPKG_TARGET_TRIPLET": "x64-windows-release",
-        "ENABLE_UNIT_TESTS": "ON"
+        "ENABLE_UNIT_TESTS": "ON",
+        "FETCH_SOUNDFONT": "ON"
```
lewri, line 54: FluidR3 appears to be downloaded as part of `_deps` but is not put in place when built for x64 (probably x86 too)

Image

TheCycoONE: It's put in place with cmake --install. You can see it in the artifact here.

lewri: Ok yes the Install build puts it there. Did the CorsixTH build option never put it in the directory before then?

image

CMakePresets.json
lewri: win-rel-x64 on debug does crash still, but is that expected?

CMakePresets.json
TheCycoONE: It didn't no. Only the pipeline added it at all.

PR #3295, opened 2026-03-24
Added method for getting object visibility

ARGAMX:
- Added method for getting object visibility

### DIFF
--- a/CorsixTH/Src/th_lua_anims.cpp
+++ b/CorsixTH/Src/th_lua_anims.cpp
@@ -489,9 +489,10 @@ int l_anim_make_invisible(lua_State* L) {
 }
 
 template <typename T>
-int l_anim_get_visibility(lua_State* L) {
+int l_anim_is_visible(lua_State* L) {
   T* pAnimation = luaT_testuserdata<T>(L);
-  uint32_t alpha_flags = static_cast<uint32_t>(thdf_alpha_50 | thdf_alpha_75);
+  constexpr uint32_t alpha_flags =
+      static_cast<uint32_t>(thdf_alpha_50 | thdf_alpha_75);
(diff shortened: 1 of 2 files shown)


CorsixTH/Src/th_lua_anims.cpp
```diff
@@ -488,6 +488,15 @@ int l_anim_make_invisible(lua_State* L) {
   return 1;
 }
 
+template <typename T>
+int l_anim_get_visibility(lua_State* L) {
+  T* pAnimation = luaT_testuserdata<T>(L);
+  uint32_t alpha_flags = static_cast<uint32_t>(thdf_alpha_50 | thdf_alpha_75);
```
TheCycoONE: This could be a constexpr

CorsixTH/Src/th_lua_anims.cpp
```diff
@@ -488,6 +488,15 @@ int l_anim_make_invisible(lua_State* L) {
   return 1;
 }
 
+template <typename T>
+int l_anim_get_visibility(lua_State* L) {
```
TheCycoONE: Could be clearer named is_visible

TheCycoONE:
Need to bump the api_version

CorsixTH/Src/th_lua_anims.cpp
ARGAMX: > This could be a constexpr

Done 👌

CorsixTH/Src/th_lua_anims.cpp
ARGAMX: > Could be clearer named is_visible

Done 👌

ARGAMX:
> Need to bump the api_version

Done 👌

PR #3298, opened 2026-03-25
Cleanly exit the sprite viewer

### LINKED ISSUE #3297: [Bug] Sprite Viewer exit breaks text input handler
### Describe the issue

Spriteviewer can crash CorsixTH's textinput handler on pressing `q` to exit. 
It has been this way since at least `0.66`

### Steps to Reproduce

1. Run CorsixTH. 
2. Don't press any keyboard key until these steps say to do that. Use only the mouse.
3. In the main menu press `Single Scenario` -> Select any level at left -> press `Start`.
4. Close the `Briefing` window with the mouse.
5. Open `Sprite Viewer`: Top Menu -> `Debug` -> `Sprite Viewer`. (debug should be enabled in you config to make this `Debug` section available to you)
6. Press `q` on the keyboard to close `Sprite Viewer`.
7. Observe `An error has occurred!` message in the game console.
8. Open the save window and try to type save name.
9. Observe that the `key handler` crashed and you can't type save name.

### Expected Behaviour

The crash shouldn't happen, or the key handler at least should be handled correctly on quit.

### CorsixTH Version

0.69.2, 0.68, 0.67, 0.66

### Operating System

Any

### Gamelog.txt

```yaml
Lua\utility.lua:316: bad argument #1 to 'for iterator' (table expected, got nil)
stack traceback:
        [C]: in for iterator 'for iterator'
        Lua\utility.lua:316: in global 'compare_tables'
        Lua\ui.lua:850: in function <Lua\ui.lua:834>
        (...tail calls...)
        Lua\app.lua:1205: in function <Lua\app.lua:1200>
```

### Additional Information

Might be related to #1785 

If the crash happens (as it's been observed sometimes it doesn't), text input will break in the functionality e.g. save game.

lewri:
*Fixes #3297*

- Prevents a crash when exiting sprite viewer on a new game 
- I do not understand why loading an existing save does not crash this.

Preferably check I haven't now made it crash in other cases on q where it did work (the textinput handler)

### DIFF
--- a/CorsixTH/Lua/sprite_viewer.lua
+++ b/CorsixTH/Lua/sprite_viewer.lua
@@ -149,9 +149,6 @@ local function DoKey(_, rawchar)
   elseif key == "s" then
     sdown = true
     need_draw = true
-  elseif key == "q" then
-    TheApp.eventHandlers = old_event_handlers
-    need_draw = false
   elseif key == "-" then
     if scale > 1 then
       scale = scale - 1
@@ -171,6 +168,10 @@ end
 
 local function DoKeyUp(_, rawchar)
     local key = rawchar:lower()
+    if key == "q" then -- Exit sprite viewer
+      TheApp.eventHandlers = old_event_handlers
+      return
+    end
     if key == "w" then
         wdown = false
     end


lewri:
@TheCycoONE -- might be good to verify with you if you'd rather unearth the root cause. It sounds like if textinput isn't initialised in a new game (by pausing, changing speed, saving game etc) this can cause the textinput crash. But existing saves have done textinput and it seems like there might be some persistence issue that is actually stopping this one breaking.

TheCycoONE:
Don't block any PR that leaves the code in a better state than before it - this is good for me.

PR #3303, opened 2026-03-31
Invert game loop Lua/C

TheCycoONE:
Instead of the game mostly running in a co-routine called from lua after init, the SDL main event loop is called directly and calls out to the lua for dispatching various events, in a more traditional setup.

The error handling is moved to it's own method and the old run method removed. The infinite loop counter is moved to a hook in C++, which avoids clobbering any other hook that may have been set, as it's a low priority mostly debugging feature.

The infinite loop counter, and fps counters are now static lifetime variables in C++ instead of lua. We could consider doing similar with other structures like MoviePlayer that function as if they were static lifetime singletons. This would put a little less pressure on lua memory management.

I have tested several error conditions, triggering error and divide by zero on click, in the onTick of a patient, and triggering the infinite loop hook. In all these cases the experience is the same on master and this branch. I'm not satisfied with the error handling behaviour but for this refactor I'm just going for consistency. I have tested a crash in the music load callback only on the after code where it behaves similar to other errors.

Unfortunately this change is not sufficient for good results from tracy which still thinks the mainloop callbacks are nested quite deep in the lua stack. That is because ttacy's lua hook is not capable of handling the longjumps triggered by errors and our language loading code depends on errors for signalling.

### DIFF
--- a/CorsixTH/Src/sdl_core.cpp
+++ b/CorsixTH/Src/sdl_core.cpp
@@ -397,7 +397,7 @@ void mainloop(lua_State* L) {
           if (lua_pcall(L, 1, 0, -3) != LUA_OK) {
             SDL_RemoveTimer(timer);
           }
-          lua_pop(L, 1); // Remove l_error_handler
+          lua_pop(L, 1);  // Remove l_error_handler
           nargs = 0;
           break;
         case SDL_USEREVENT_TICK:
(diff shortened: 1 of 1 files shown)


TheCycoONE:
@lewri can you see if this changes anything on windows

PR #3304, opened 2026-04-01
Cancel accidentally grabbed object

### LINKED ISSUE #2624: Cancel grabbed by accident item action.

Ingame. Sometimes you grab an item (right mouse button click) by accident and don't know where it was placed before. There should be a mechanism to cancel this action and put the item back where it was taken from. At the moment if you press 'Esc' in such a situation that item just Sells, which does not look like an obvious and expected result. The same poor UX was in the original game. Never understand why it was done in such a poor manner. From my point of view 'Esc' should cancel the action and put the item back. Item still can be sold by pressing the Red Cross in the 'placement menu' appeared after grab.

Better UX.

No

0.68.0 Beta 1

No

### Additional Information

_No response_

### LINKED ISSUE #3300: A bench facing north unable to become invisible. Multiple `animations` for single object.
Image

----

### Describe the issue

A bench facing north cannot become invisible by a `object.th:makeInvisible()` call.

At the same time, absolutely all other corridor items in all rotations become invisible after this call and function correctly. The radiators, trash cans, drink machines, fire extinguishers — everything works fine.
Benches facing west, south, and east also become invisible perfectly.

I wouldn't care at all if it weren't for a certain functionality that involves applying invisibility to objects that I'm working on.

----

### CorsixTH Version

0.69.2

----

### Steps to Reproduce

1. Download this branch: https://github.com/ARGAMX/CorsixTH/tree/makeInvisible-bench-test 
2. That branch diff can be accessed there: https://github.com/ARGAMX/CorsixTH/commit/4139aa332f5ed8920de98dc22bd8bdf16c7375bd
3. As you can see, this code makes an item invisible when it is picked up from the floor and logs its flags before and after `object.th:makeInvisible()` call.
4. Build/apply that source and run game.
5. Select `Single Scenario` and start any level.
6. Buy various corridor items and place them around the hospital in different orientations. Be sure to buy a bench and place it facing north, as shown in the image.
7. Experiment with the items and notice that the north-facing bench cannot be made invisible. The other items become invisible normally, as expected.
8. Check log/console.

----

### Expected Behaviour

The bench faced north should become invisible.

----

### Additional information:

Here is an example of a log before and after calling the method `object.th:makeInvisible()` for corridor items, except bench facing north:

```yaml
### Before. self.th:getFlag()): 0  | getVisible: +
### After.  self.th:getFlag()): 12 | getVisible: -
```

A bench facing north:

```yaml
### Before. self.th:getFlag()): 8192 | getVisible: +
### After.  self.th:getFlag()): 8204 | getVisible: -
```

Values In binary format:

```
0    = 0000000000000000
12   = 0000000000000110
8192 = 0010000000000000
8204 = 0010000000000110
```

The 14th bit is apparently this:

```C++
//! Draw using nearest pixel hinting
thdf_nearest = 1 << 14,
```

Bench footprint:

```Lua
object.orientations = {
  north = {
    render_attach_position = { {0, 0}, {-1, 0}, {0, -1} },
    footprint = { {0, 0, complete_cell = true}, {0, -1, only_passable = true, invisible = true, shareable = true} },
    use_position = "passable",
  },
  east = {
    footprint = { {0, 0, complete_cell = true}, {1, 0, only_passable = true, invisible = true, shareable = true} },
    use_position = "passable",
  },
  south = {
    footprint = { {0, 0, complete_cell = true}, {0, 1, only_passable = true, invisible = true, shareable = true} },
    use_position = "passable",
  },
  west = {
    footprint = { {0, 0, complete_cell = true}, {-1, 0, only_passable = true, invisible = true, shareable = true} },
    use_position = "passable",
  },
}
```

Bench facing north have `render_attach_position` and the other direction does not.
`render_attach_position = { {0, 0}, {-1, 0}, {0, -1} }`

If you remove `render_attach_position` for the north bench, the defect disappears and the bench becomes invisible after the call as expected.

However, removing `render_attach_position` could potentially create rendering issues with the northern bench, as nearby objects could begin to obscure it. I haven't tested this, though.

For more details, check out the Discord discussion:

https://discord.com/channels/731272375949328486/854282144101040179/1487174907895746671 

Or search by text `I'm currently working on a task that requires calling` in `#technical` channel.

ARGAMX:
*Fixes #2624*
*Fixes #3300*

- Change provide ability to place back object which was picked up.
If the user, while attempting to grab the doctor, instead accidentally picked up an object from the ground, and sometimes the original location of the object is not visible due to the walls, then with this change by pressing `Esc` user can return the object back to its place, thereby simply canceling the pickup action.

With this change the grabbed item will now not actually be lifted from its place, but will become invisible until it is either placed in a new location or removed (sold). This will prevent humanoids from building paths over a tile that has been occupied by this item, so that a collision cannot be created when the object is returned.

This change only affects "corridor items" - items from 'Furnish corridor' menu (even if they placed in the rooms).

A similar strategy currently is already used when grabbing staff.

If the user still wants to sell the grabbed item (what is needed when a user wants to remove a lot of benches from a corridor to make space for building a room), he can now do so in the following three ways:
- Click the button with the red cross in the window that appears (UIPlaceObjects)
- Press new `ingame_sellPickedItem` hotkey ('delete' by default).
- Press `global_cancel` hotkey ('q' by default).

the save file on which I conducted the tests:
[0.69.2 cancel grabbed by accident item.sav.zip](https://github.com/user-attachments/files/26416192/0.69.2.cancel.grabbed.by.accident.item.sav.zip)


CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -681,6 +710,58 @@ function UIPlaceObjects:setBlueprintCell(x, y)
       end
     end
 
+    -- flags for the case of moving a previously placed object to a new location
```
TheCycoONE: This function was already too big, it's mixing logic between local function declarations, I'm finding this hard to follow.

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -697,7 +778,8 @@ function UIPlaceObjects:setBlueprintCell(x, y)
           flag = "passable"
         end
         if tile.only_side then
-          if object.thob == 50 and direction == "east" then
+          local trash_bin_thob = 50
```
TheCycoONE: Not sure defining this in three places just before use is best. Either we micro-optimize because it's an interpreted language and leave the 50 (with a comment), or we make a list of thob_id constants somewhere in the global table and import them in each file that needs them.

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -711,18 +793,25 @@ function UIPlaceObjects:setBlueprintCell(x, y)
...
         -- ignore placed object tile if it is shareable
         if not tile.shareable and is_object_allowed then
           -- Check 4: only one object per tile allowed original TH
           -- can build on litter and unoccupied tiles and only placeable if not on another objects passable footprint unless that too is a shareable tile
           local objchk = map:getCellFlags(xpos, ypos, flags)["thob"]
-          is_object_allowed = objchk == 0 or objchk == 62 or objchk == 64 -- no object, litter/puke, ratholes
-          is_object_allowed = is_object_allowed and world:isTileExclusivelyPassable(xpos, ypos, 10)
+          local objects_do_not_interfere =
+            (objchk == 0) or (objchk == 62) or (objchk == 64) -- no other object, except litter/puke, ratholes
+          local moving_existing_object_on_same_place =
```
TheCycoONE: This only works when the footprint is 1x1?

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > Not sure defining this in three places just before use is best. Either we micro-optimize because it's an interpreted language and leave the 50 (with a comment), or we make a list of thob_id constants somewhere in the global table and import them in each file that needs them.

Ok, I changed it to a comment everywhere.

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > This function was already too big, it's mixing logic between local function declarations, I'm finding this hard to follow.

Ok, I split this function into 2 functions.

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > This only works when the footprint is 1x1?

This calculates for each tile in the footprint.
So in one iteration 1x1 tile is checked.

```Lua
-- for each tile in object_footprint check map tile availability
for i, tile in ipairs(object_footprint) do
```

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: It turns out this wasn't a good idea because now it creates a conflict with 3282.

I refactored this function locally on top of the 3282.
And moved the changes from one branch to another.

But now the linter is complaining about this PR.
If I agree to the linter's demands, I'll get even more conflicts when merging with 3282.

Therefore, to refactor this function, I need to apply 3282 first.
Reverting back that refactor for now.

TheCycoONE: Ok, maybe better to land that PR first then

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -977,6 +839,165 @@ function UIPlaceObjects:selectObjectType(object_type)
...
+          world:isFootprintTileBuildableOrPassable(xpos, ypos, tile, object_footprint, flag, player_id)
+      elseif is_object_allowed then
+        is_object_allowed = (map:getCellFlags(xpos, ypos, flags)[flag] or same_placement) and
+          (player_id == 0 or flags.owner == player_id)
+      end
+
+      -- ignore placed object tile if it is shareable
+      if not tile.shareable and is_object_allowed then
+        -- Check 4: only one object per tile allowed original TH
+        -- can build on litter and unoccupied tiles and only placeable if not on another objects passable footprint unless that too is a shareable tile
```
Alberth289346: We generally have max line length of about 80 characters

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -977,6 +839,165 @@ function UIPlaceObjects:selectObjectType(object_type)
...
+      -- Having checked if the tile is good set its blueprint appearance flag:
+      if is_object_allowed then
+        if not tile.invisible then
+          map:setCell(xpos, ypos, 4, good_tile)
+        end
+      else
+        if not tile.invisible then
+          map:setCell(xpos, ypos, 4, bad_tile)
+        end
+        setAllGood(tile)
```
Alberth289346: This looks horribly wrong, but it's just the function name that sends the wrong message.

CorsixTH/Lua/dialogs/place_objects.lua
Alberth289346: Ah, not your fault. I am also fine with leaving it as-is.

CorsixTH/Lua/languages/english.lua
```diff
@@ -748,6 +748,7 @@ hotkey_window = {
   ingame_poopLog = "Dump Log",
   ingame_poopStrings = "Dump Strings",
   ingame_patient_gohome = "Send Home",
+  ingame_sellPickedItem = "Sell Picked Item",
```
lewri: `Picked-up` / `Picked Up`

Mainly because of the wording in the tooltip of edit room.

Image

ARGAMX: > `Picked-up` / `Picked Up`
> 
> Mainly because of the wording in the tooltip of edit room.
> 
> Image

👌

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > This looks horribly wrong, but it's just the function name that sends the wrong message.

agree. is better now?

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > We generally have max line length of about 80 characters

I have split that string 👌

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > Ok, maybe better to land that PR first then

PR take off again 👌

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -641,132 +669,222 @@ function UIPlaceObjects:setBlueprintCell(x, y)
...
+    east = { x = 1, y = 0, buildable_flag =  "buildableEast", passable_flag = "travelEast", needed_side = "need_east_side"},
+    south = { x = 0, y = 1, buildable_flag = "buildableSouth", passable_flag = "travelSouth", needed_side = "need_south_side"},
+    west = { x = -1, y = 0, buildable_flag = "buildableWest", passable_flag = "travelWest", needed_side = "need_west_side"}
+    }
+
+  local room = self.room and self.room.id
+  local valid_placement, room_id, passable_flag = self:_placementAvailableHereAndOtherObjectsDoNotInterfere(
+      x, y, object, real_obj, object_footprint, direction, direction_parameters, room, map, world)
+
+  local function _setPartialFlags(x_, y_, map_, flag_altpal_, allgood)
```
TheCycoONE: Doesn't use any upvalues, no reason to be nested.

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -641,132 +669,222 @@ function UIPlaceObjects:setBlueprintCell(x, y)
...
-    local object_footprint = object.orientations[self.object_orientation].footprint
-    local map = self.map.th
-    if #object_footprint ~= #self.object_footprint then
-      self.object_footprint = {}
-      for i = 1, #object_footprint do
-        self.object_footprint[i] = {}
-      end
+
+  if not x or not y or #self.objects == 0 then
+    self.object_footprint = {}
```
TheCycoONE, line 690: Is there any reason to hit this case or should it be considered an error?

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -641,132 +669,222 @@ function UIPlaceObjects:setBlueprintCell(x, y)
...
-    self.object_footprint = {}
+  if moving_existing_object and original_footprint_flags then
+    -- we've tested the placement and now we need to re-occupy the cells.
+    self:_occupyObjectSpaceBack(original_footprint_flags, map)
+  end
+
+  return valid_placement, room_id, passable_flag
+end
+
+--! Set cell flags in accordance that the tile is available for build and placement.
```
TheCycoONE: Why do we have to do this now when we use to just call clearBlueprint?

CorsixTH/Lua/dialogs/place_objects.lua
Alberth289346, line 798: Split different checks into different functions instead of nesting independent checks into one endless single function.

Not doing that now is fine though.

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > Is there any reason to hit this case or should it be considered an error?

I don't know for what purposes this is here for. 
This check and that bypass was here before my changes. 
I didn't make any research on this.
I just inverted condition body to remove the nesting of the big block below.

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > Why do we have to do this now when we use to just call clearBlueprint?

I'm sure Blueprint, as an entity, isn't intended to change the state of tile flags.

When I was working on this PR and the area blocking PR, I was convinced that Blueprint was simply the image we see under the cursor.

So the calls to `UIPlaceObjects:clearBlueprint` and `UIPlaceObjects:setBlueprintCell` aren't intended to change tile flags themselves. Even if they do, they should return everything back to normal by the end every function for itself.

I believe the main purpose of these methods is to ensure the Blueprint correctly renders the image under the cursor.

The methods that call by `UIPlaceObjects:setBlueprintCell` are forced to change flags during the process to check for valid placement, and then they clean up their changes after themselves, without waiting for `clearBlueprint` to be called.

`UIPlaceObjects:setBlueprintCell` was changing and clearing its flags changes even before my area blocking PR. This seemed logical to me back then, and I continued this logic here too.

Honestly, I have no idea what `UIPlaceObjects:clearBlueprint` does. I never even analysed what exactly it does.

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > Doesn't use any upvalues, no reason to be nested.

I was thinking about moving out `setPartialFlags` as a separate function. But then I'd probably have to write documentation for it?
But I don't understand what it does and for what.
It's a black box for me, even though it's small function.
In both PR, I didn't need to know this.
In the PR for blocking areas, Lewri also was suggested to move this `setPartialFlags` out, but I rejected proposal with the argument that I didn't understand what the function do.
Maybe I even gave it a bad name. Possible this is where the Blueprint rendering actually happens.
Should it be taken out without a description?

lewri: In my opinion if a function is going to be doing complex things it should be documented regardless of where it lives (a local doesn't need luadoc though, it can be just normal comment lines).

If you genuinely don't know what it does, do you at least have some idea to write down? Otherwise we'll just live with it not being defined yet.

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -641,132 +669,222 @@ function UIPlaceObjects:setBlueprintCell(x, y)
...
+    valid_placement = valid_placement and self:_isNonSideObjectPlacementValid(x, y, object, self.object_orientation, room_id, world)
+    _setPartialFlags(x, y, map, flag_altpal, valid_placement)
+  elseif object.class == "SideObject" then
+    -- SideObject - an object that lives on the edge of a tile (radiator, bin, extinguisher).
+    valid_placement = valid_placement and self:_isSideObjectPlacementValid(x, y, room_id, passable_flag, map, world)
+    _setPartialFlags(x, y, map, flag_altpal, valid_placement)
+  end
+end
+
+--! Function checks whether it is available for placing the object and
```
tobylane: What is the it, a set of tiles that match the object and its use positions, that a user sees as a blueprint?

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -641,132 +669,222 @@ function UIPlaceObjects:setBlueprintCell(x, y)
...
     end
-    local optional_tiles = 0
-    for _, tile in ipairs(object_footprint) do
-      if tile.optional then
-        optional_tiles = optional_tiles + 1
-      end
+  end
+  local world = self.ui.app.world
+  local direction = self.object_orientation
+  local direction_parameters =  {
```
tobylane: These three and room_id (keep the name consistent) should be set in the functions that need them rather than passed through. direction parameters could be a file-level local variable.

ARGAMX: > These three and room_id (keep the name consistent) should be set in the functions that need them rather than passed through. direction parameters could be a file-level local variable.

moved to functions 👌

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > What is the it, a set of tiles that match the object and its use positions, that a user sees as a blueprint?

Oops, lost in translation. Rephrased. 👌

CorsixTH/Lua/dialogs/place_objects.lua
TheCycoONE: Move it out and don't document it yes, if you don't know you don't know; but it can clearly be moved regardless.

CorsixTH/Lua/dialogs/place_objects.lua
TheCycoONE: Well the answer doesn't inspire a lot of confidence :face_exhaling: we were able to sell / move / delete objects before your PR, so the code to do so should already exist. My question remains why we need new code to clear flags when we're adding a cancel option.

ARGAMX: > Well the answer doesn't inspire a lot of confidence 😮‍💨 we were able to sell / move / delete objects before your PR, so the code to do so should already exist. My question remains why we need new code to clear flags when we're adding a cancel option.

I just didn't understand why you were asking this at first.
Now I get the point.

The thing is, the object essentially remains in place while "picked up".
Just invisible.
Because of this, its flags remain in place.
And these flags prevent the object from being placed back in the same place or just slightly moved by 1 tile, manually by user cursor. 

So, you first have to remove all the flags for that object. (`_deoccupyObjectSpace`)
Run a placement check.
And then return the flags to their original state. (`_occupyObjectSpaceBack`)

So, `clearBlueprint` has nothing to do with flags.

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > Move it out and don't document it yes, if you don't know you don't know; but it can clearly be moved regardless.

Moved out 👌

CorsixTH/Lua/dialogs/place_objects.lua
lewri: It sounds like the issue is that if on pickup, you cancel and an entity has moved there within that time, the entity might get stuck on the tile? Another option could be to make the tile impassable during pickup (but then you also need to make sure the player can still put that object back in the same place manually if they want).

Cyco are you mainly just asking if the solution feels over-engineered for what we already had?

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -314,16 +329,22 @@ function UIPlaceObjects:removeObjects(object_list, refund)
   end
 end
 
-function UIPlaceObjects:close()
+--! Close menu.
+--!param move_canceled (bool) moving objects has been cancelled.
+function UIPlaceObjects:close(move_canceled)
   self.ui:tutorialStep(1, {4, 5}, 1)
-  self:removeAllObjects(true)
+  self:removeAllObjects(true, move_canceled)
```
lewri: I think you're accidentally refunding the player on a move cancel here?

lewri (commented):
I realised with this PR I occasionally used Esc to delete an object. 😵

CorsixTH/Lua/dialogs/place_objects.lua
TheCycoONE: I'm asking why we need new code to clear flags on a tile when we clearly already have some code somewhere that does that or we wouldn't have been able to move anything.

ARGAMX: As I already explained, this PR doesn't reset the flags after all.

> Another option could be to make the tile impassable during pickup.

This is almost the same as what's done here.
In fact, the flags for the item aren't removed during pick up.
Item actually continues to occupy its place until new place selected.

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > I think you're accidentally refunding the player on a move cancel here?

True 
Thanks! This is a nice find! 👍

There really was something wrong with the money management here.

I fixed it.

ARGAMX:
> I realised with this PR I occasionally used Esc to delete an object. 😵

Me too, and that's understandable.
It was a bad UX from the start.
The player adapts to it and gets used to it.

And now we're going through withdrawal.
I've been thinking about adding a flag, but it seems like a compromise to this wrong UX in the first place.
So I decided not to do it.

CorsixTH/Lua/dialogs/place_objects.lua
TheCycoONE: I've been referring to this block:
```
    local completely_availability_flags = {
      passable = true,
      buildable = true,
      buildableNorth = true,
      buildableEast = true,
      buildableSouth = true,
      buildableWest = true
    }
    map:setCellFlags(target_x, target_y, completely_availability_flags)
```
To me it looks wrong. object.lua has a lot of code for restoring the original cell flags, including accommodating for if the tile is part of the footprint of multiple objects which this code seems to ignore. I think the code in object.lua needs to be refactored out of the setTile function so that it can be reused here.

ARGAMX: > I've been referring to this block:
> 
> ```
>     local completely_availability_flags = {
>       passable = true,
>       buildable = true,
>       buildableNorth = true,
>       buildableEast = true,
>       buildableSouth = true,
>       buildableWest = true
>     }
>     map:setCellFlags(target_x, target_y, completely_availability_flags)
> ```
> 
> To me it looks wrong. object.lua has a lot of code for restoring the original cell flags, including accommodating for if the tile is part of the footprint of multiple objects which this code seems to ignore. I think the code in object.lua needs to be refactored out of the setTile function so that it can be reused here.

Ok, now I finally get what you mean. 👌

Previously I thought I understood, but no, it was a false enlightenment.

I thought you meant why this PR is doing operations that are already being done. I thought you’re simply suggesting that part of code in that PR should be removed.

But in reality you mean that the just old code needs to be reused. I didn't understand that at first.

Only from your last message did I understand what exactly you were getting at.

ARGAMX: > I've been referring to this block:
> 
> ```
>     local completely_availability_flags = {
>       passable = true,
>       buildable = true,
>       buildableNorth = true,
>       buildableEast = true,
>       buildableSouth = true,
>       buildableWest = true
>     }
>     map:setCellFlags(target_x, target_y, completely_availability_flags)
> ```
> 
> To me it looks wrong. object.lua has a lot of code for restoring the original cell flags, including accommodating for if the tile is part of the footprint of multiple objects which this code seems to ignore. I think the code in object.lua needs to be refactored out of the setTile function so that it can be reused here.

Done. ✅

`Object.setTile` refactored: 
Part of the function's content has been moved into two functions: `Object.occupyTilesByObjectFootprint` and `Object.deoccupyTilesByObjectFootprint`. ✅

The functions `UIPlaceObjects:_deoccupyObjectSpace` and `UIPlaceObjects:_occupyObjectSpaceBack` have been removed. ✅

Instead, `Object.deoccupyTilesByObjectFootprint` and `Object.occupyTilesByObjectFootprint` are reused. ✅

-----------

Actually, I didn't know the `Object.setTile` function is responsible for setting and removing cell flags until you explicitly mentioned it here. 

The thing is, `UIPlaceObjects:setBlueprintCell(x, y)` also temporarily sets and clears cell flags. It don't reuse `Object.setTile`.

While working on PR `Support for blocking off areas in rooms 3282`, I had to work a lot with `UIPlaceObjects:setBlueprintCell(x, y)` and the functions it calls - each function set and cleared cell flags independently and in its own way, if needed - this approach became standard for me.

I had never seen anything else, so with all that experience, it was perfectly logical to do exactly the same here.

At the same time, `Object.setTile` has no description at all. For some reason, I completely missed the fact that it's the one the "Main part" who sets and clears cell flags. So I thought the setting up and removal of flags was just scattered and messed up throughout the project.

After that, when I started to analyse `Object.setTile(x, y)` I found something which confusing me: if you create an object and initialise it with `Object.setTile(x, y)`, then call `Object.setTile(nil)` on that object, and then again `Object.setTile(x, y)`, you won't get a normal object in the end. It will disappear from the world and won't return normally.

Not only is `Object.setTile` overloaded with two opposite responsibilities at once: "initialisation" and then "destruction".
Moreover, both of these operations don't perform mirrored actions. Code like this is difficult to reuse. However, I managed to do so by cutting out the useful part of this function.

But `Object.setTile`, frankly, still requires more thorough refactoring. I see that you are very knowledgeable about this area. 😜

CorsixTH/Lua/dialogs/resizables/hotkey_assign.lua
```diff
@@ -338,6 +338,7 @@ function UIHotkeyAssign:UIHotkeyAssign(ui, mode)
             "ingame_restartLevel",
             "ingame_quitLevel",
             "ingame_rotateobject",
+            "ingame_sellPickedItem",
```
lewri: Another picked > picked up inconsistency.

lewri (commented):
Picking up a bench a patient is sat on duplicates the bench instead of making them get up. Also free money I guess?

Image

Also just noticed I can place it, in front of the patient? I guess a side effect.

Edit: Ok, placing it back down will cause the humanoid to be teleported to standing as the bench subsequently moves.

lewri:
Also happens with drinks machine if patient is assigned to use it (just leaving, purple one)
image

ARGAMX:
> Picking up a bench a patient is sat on duplicates the bench instead of making them get up. Also free money I guess?
> 

Thanks. 👍
Fixed. 👌

ARGAMX:
> Also happens with drinks machine if patient is assigned to use it (just leaving, purple one) 

Thanks. 👍
Fixed. 👌

ARGAMX:
Now, when object "picked up", the usage, if any, is reset and the reservation is also canceled.

ARGAMX:
Bump. For somebody to approve this.

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -240,11 +242,24 @@ function UIPlaceObjects:addObjects(object_list, pay_for)
...
+    object.existing_object.picked_up = false
+    if move_canceled then
+      -- Object placed back
+      self.ui:playSound("place_r.wav")
+    elseif not placed then
+      -- Destroy object
+      -- The player should be reimbursed for the cost of the item, i.e.
+      -- whether they were previously purchased. So do refund.
+      local build_cost = self.ui.hospital:getObjectBuildCost(object.object.id)
+      local msg = _S.transactions.sell_object .. ": " .. object.object.name
```
TheCycoONE: Out of scope / not your fault - it would be better if the translation included the placeholder rather than concatenation. Off the top of my head, french uses an extra space before a `:`, and I imagine there are languages with different punctuation.

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -782,10 +847,12 @@ end
 --! param object (object) target SideObject.
 --! param object_orientation (string) footprint orientation name.
 --! param room_id (integer) room id if we in a build mode.
---! param world (object) world class instance
+--! param world (object) world class instance.
```
TheCycoONE: World was not added as a parameter, comment doesn't match function definition.

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -840,17 +907,19 @@ end
 --! param y (integer) target tile y coordinate.
 --! param room_id (integer) room id if we in a build mode.
 --! param passable_flag (string) passable flag name. orientation of an object relative to the cardinal directions.
---! param map (object) map class instance
---! param world (object) world class instance
+--! param map (object) map class instance.
+--! param world (object) world class instance.
```
TheCycoONE: Ditto, no world parameter.

TheCycoONE:
Better architecturally. Approved when code issues / play testing are complete.

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > World was not added as a parameter, comment doesn't match function definition.

Fixed 👌

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > Ditto, no world parameter.

Fixed 👌

ARGAMX:
Bump. For somebody to approve this.

lewri:
Humanoids now get stuck in must_stand an can't move when you yoink the bench from above them. They only move when another factor gets triggered on them (e.g. toilet, called to room)
image

They need to run a search for a new bench.


Drinks machine fixed.

Receptionist now can continue to serve patients while you are holding the reception desk. They also do not move to find a new desk until you place back down.
image

lewri:
It looks like in general, you need to make all other entities behave as if the object is gone. I am guessing, introduced here, the entities are regarding it as still active and in place.

ARGAMX:
> Humanoids now get stuck in must_stand an can't move when you yoink the bench from above them. They only move when another factor gets triggered on them (e.g. toilet, called to room)
> 
> They need to run a search for a new bench.
> 
> Drinks machine fixed.
> 
> Receptionist now can continue to serve patients while you are holding the reception desk. They also do not move to find a new desk until you place back down. 

It's all fixed 👍

All other entities should behave now as if the object is gone.

CorsixTH/Lua/objects/bench.lua
```diff
@@ -172,8 +172,18 @@ function Bench:removeUser(user)
...
 end
 
---Called when the player picks up a bench
 function Bench:onDestroy()
+  self:resetUsageAndReservaton()
+  Object.onDestroy(self)
+end
+
+--! Called when the player picks up a bench
+function Bench:onPickUp()
```
TheCycoONE: Why does this method exist. Object already calls self:resetUsageAndReservation() onPickup.

CorsixTH/Lua/objects/reception_desk.lua
```diff
@@ -208,12 +208,25 @@ function ReceptionDesk:setTile(x, y)
   Object.setTile(self, x, y)
   if x and y then
     self:checkForNearbyStaff()
+    self.hospital:buildReceptionDesksCache()
   end
   return self
 end
 
 function ReceptionDesk:onDestroy()
+  self:resetUsageAndReservaton()
```
TheCycoONE: Why doesn't Object call resetUsageAndReservation onDestroy?

CorsixTH/Lua/objects/bench.lua
ARGAMX: > Why does this method exist. Object already calls self:resetUsageAndReservation() onPickup.

Yes, indeed, these methods are excessed.
After all, the Object class calls these methods, which are what trigger the whole inheritance calls.
So, I removed them. And in the other classes, too.
Thanks.

CorsixTH/Lua/objects/reception_desk.lua
ARGAMX: > Why doesn't Object call resetUsageAndReservation onDestroy?

Removed 👌

ARGAMX:
Bump. For somebody to approve this.

TheCycoONE:
I could in that I don´t have any more code nits, I haven´t play tested this.

TheCycoONE (approved):

lewri:
> Bump. For somebody to approve this.

I should probably do the play testing again shouldn't I?

lewri:
image

A picked up radiator continues outputting heat on its origin tile until placed down (did have to double check this one in an older version)

This also happens with fire extinguishers and plants (showing present on town map when picked up)


Picking up operating table has caused a crash:
```yml
An error has occurred!
Lua\world.lua:2154: attempt to perform arithmetic on a nil value (local 'y')
stack traceback:
        Lua\world.lua:2154: in method 'removeObjectFromTile'
        Lua\entities\object.lua:561: in method 'deoccupyTilesByObjectFootprintAt'
        Lua\dialogs\place_objects.lua:758: in field '_placementAvailableHereAndOtherObjectsDoNotInterfere'
        Lua\dialogs\place_objects.lua:695: in method 'setBlueprintCell'
        Lua\dialogs\place_objects.lua:1072: in field 'onCursorWorldPositionChange'
        Lua\dialogs\edit_room.lua:1528: in method 'onCursorWorldPositionChange'
        Lua\dialogs\place_objects.lua:241: in method 'addObjects'
        Lua\entities\object.lua:813: in field 'onClick'
        Lua\entities\machine.lua:486: in function <Lua\entities\machine.lua:480>
        (...tail calls...)
        Lua\ui.lua:899: in function <Lua\ui.lua:882>
        (...tail calls...)

Error in buttonup: 0
An error has occurred!
Running: The motion handler.
Lua\dialogs\place_objects.lua:661: bad argument #1 to 'setCell' (number expected, got nil)
stack traceback:
        [C]: in method 'setCell'
        Lua\dialogs\place_objects.lua:661: in method 'clearBlueprint'
        Lua\dialogs\place_objects.lua:674: in method 'setBlueprintCell'
        Lua\dialogs\place_objects.lua:1072: in field 'onCursorWorldPositionChange'
        Lua\dialogs\edit_room.lua:1528: in method 'onCursorWorldPositionChange'
        Lua\window.lua:1543: in function <Lua\window.lua:1539>
        (...tail calls...)
        Lua\game_ui.lua:581: in function <Lua\game_ui.lua:569>
        (...tail calls...)

Error in motion: 0
Warning: No event handler for motion
An error has occurred!
Running: The timer handler.
Lua\dialogs\place_objects.lua:661: bad argument #1 to 'setCell' (number expected, got nil)
stack traceback:
        [C]: in method 'setCell'
        Lua\dialogs\place_objects.lua:661: in method 'clearBlueprint'
        Lua\dialogs\place_objects.lua:674: in method 'setBlueprintCell'
        Lua\dialogs\place_objects.lua:1072: in field 'onCursorWorldPositionChange'
        Lua\dialogs\edit_room.lua:1528: in method 'onCursorWorldPositionChange'
        Lua\window.lua:1543: in function <Lua\window.lua:1539>
        (...tail calls...)
        Lua\game_ui.lua:905: in method 'onTick'
        Lua\app.lua:1277: in function <Lua\app.lua:1272>
        (...tail calls...)

Error in timer callback: 0
Warning: No event handler for timer
```

ARGAMX:
> A picked up radiator continues outputting heat on its origin tile until placed down (did have to double check this one in an older version)
> 
> This also happens with fire extinguishers and plants (showing present on town map when picked up)
> 
> Picking up operating table has caused a crash:
> 

I fixed the operating table crash. I introduced this crash in commit 798ccb7 two weeks ago.

Regarding the picked up radiator continuing to heat up the place... yes, ideally that shouldn't happen. But honestly, as a player, I don't realy care whether it continues to heat up the place or not while picked up. It very short period of time and has very little impact on gameplay. And fire extinguishers and other items are even make smaller effect.

To try to fix this, I will have to continue to make more changes, which potentialy would introduce new defects, as this with Operation Table was. I'm not sure it's worth it. So let it be.

Once the user finishes pickup, and he will do this quite quickly because the pickup made his cursor low-functional, this "issue" disappears.

The fact that you even checked this shows how passionately you checked that PR. Commendable. 👍

lewri (changes requested):
This PR now stops player being refunded on cancelling purchase (pressing  Esc) during placement. The X and sell hotkey work.

For new objects during initial placement, I would expect Esc to also refund and the object is destroyed.

NB: Also applies when building a room and buying bonus objects, press Esc when building room does not refund the unused bonus objects.
NBNB: Pressinbg the tick also doesn't refund unused bonus objects.

lewri:
Bench duplication is also back again if a patient is on the bench
image

Also vending machine, It does send the patient away though
image

They do disappear once you place down again though

lewri:
> egarding the picked up radiator continuing to heat up the place... yes, ideally that shouldn't happen. But honestly, as a player, I don't realy care whether it continues to heat up the place or not while picked up. It very short period of time and has very little impact on gameplay. And fire extinguishers and other items are even make smaller effect.

I've dug around again. This bug actually isn't you.

CorsixTH/Lua/dialogs/place_objects.lua
```diff
@@ -647,133 +685,168 @@ function UIPlaceObjects:setBlueprintCell(x, y)
...
-          is_object_allowed = is_object_allowed and world:isTileExclusivelyPassable(xpos, ypos, 10)
+    -- Check 1: Does the tile have valid map coordinates?:
+    if not world:isOnMap(xpos, ypos) then
+      updateResult(tile)
+      xpos = 0
+      ypos = 0
+    else
+      local flag = "buildable"
+      local good_tile = 24 + flag_alpha75
+      local bad_tile = 67 + flag_alpha75
```
lewri, line 785: Nit but can we label the magic numbers, I'm guessing it's an `_offset`

CorsixTH/Lua/world.lua
```diff
@@ -1742,14 +1745,14 @@ function World:newEntity(class, animation, mood_marker)
   return entity
 end
 
-function World:destroyEntity(entity)
+function World:destroyEntity(obj)
```
lewri: Can't remember if discussed before but I'm pretty sure it was something about obj not being a game object (like a desk etc); but just a code object? I would doc this changed param name if so. Otherwise someone's going to ask why is a humanoid an object (e.g. room:crashRoom and SpawnAction)?

lewri (commented):
I'm not sure @MandrakeHorse might be available to give this a good thrashing, if not, functionally I can't find anything else.

CorsixTH/Lua/dialogs/place_objects.lua
ARGAMX: > Nit but can we label the magic numbers, I'm guessing it's an `_offset`

I didn't write this code and I don't understand what it does.
You can see it was just moved during refactoring.
So you can simply offer a ready-made code version here.
I don't understand what these magic numbers mean.

ARGAMX: > Nit but can we label the magic numbers, I'm guessing it's an `_offset`

I figured it out a bit. 
These are tile flags, not the offset. Bit flags.
But I still don't want to touch it yet.

CorsixTH/Lua/world.lua
ARGAMX: > Can't remember if discussed before but I'm pretty sure it was something about obj not being a game object (like a desk etc); but just a code object? I would doc this changed param name if so. Otherwise someone's going to ask why is a humanoid an object (e.g. room:crashRoom and SpawnAction)?

Okay.
Then I rename it back to what it was.
Also, I tried to remember why I renamed `entity` to `obj` here, and I think the reasoning wasn't convincing enough.

lewri (approved):

PR #3305, opened 2026-04-02
Clean up keydown events in keyup

### LINKED ISSUE #3299: UI.modifiers_down is persisted when it shouldn't be
**A lot of this investigation mainly talks about how the sprite viewer fails to quit; for the title subject, it simply supports the section at the very end of this.**

As part of further investigation to the root cause of #3297 I have found out that the reason the sprite viewer would crash on a new game with no text in put is because `ui.modifiers_down` has never been set.

In the function below, the doc will suggest that the onKeyDown event has already been processed.
https://github.com/CorsixTH/CorsixTH/blob/32249f50da31154066a64169a089ad420afbd378/CorsixTH/Lua/ui.lua#L831-L834

This is technically true. However, in the case of the sprite viewer we use an override of this event function with doKey
https://github.com/CorsixTH/CorsixTH/blob/32249f50da31154066a64169a089ad420afbd378/CorsixTH/Lua/sprite_viewer.lua#L129

This function uses `_` to ignore self, and also doesn't take the `modifiers` parameter used in the main keyDown handler. Because of this override, we don't set the `ui.modifiers_down` value at all, as it is set in the main `onKeyDown` event.
https://github.com/CorsixTH/CorsixTH/blob/32249f50da31154066a64169a089ad420afbd378/CorsixTH/Lua/ui.lua#L770

What happens from the perspective of the sprite viewer's exit is that we called the `keyDown` event, which returns the old event handlers in place, we then call onTextInput from the main handlers with a compare that has a nil table.
https://github.com/CorsixTH/CorsixTH/blob/32249f50da31154066a64169a089ad420afbd378/CorsixTH/Lua/ui.lua#L842-L851

Additionally, the UI isn't told either the keypress is handled on `ui.key_press_handled` in the above, which also helps lead to the crash.

For the case of sprite viewer, there are two possible patch outs as alternatives to #3298 :
1. Don't ignore the `modifiers` passed to the keyDown event in the sprite viewer; and set it on q (or at every doKey event)
```lua
local function DoKey(_, rawchar, modifiers)
...
  elseif key == "q" then -- Exit sprite viewer
    -- We must send modifiers to the UI as our DoKey event overrides the normal
    -- keyDown behaviour.
    TheApp.ui.modifiers_down = modifiers
    TheApp.eventHandlers = old_event_handlers
    need_draw = false
```
2. Set `key_press_handled` explicitly each time inside doKey.

In terms of further going to the root of the issue, this also unearths that `ui.modifiers_down` is being persisted in saves when it shouldn't be. This can be seen by loading up any save before accessing the sprite viewer where it will not crash. The `modifiers_down` has something assigned to it that gets restored to the save. **Unfortunately, as I don't know why it persists itself I can't go deeper with how to stop that happening.**

lewri:
*Fixes #3299*

- Properly clear keydown values set on keyup
- Refactor key determination to be shared between keydown and keyup

The API version bump may not be necessary.


TheCycoONE:
version bump is probably needed but it's also broken at the moment I think, probably my regression again.

TheCycoONE (approved):
Makes some sense to be symmetric but I wouldn't depend on it too much, the user could release the modifier first.

CorsixTH/Src/sdl_core.cpp
```diff
@@ -304,8 +304,8 @@ void mainloop(lua_State* L) {
         case SDL_KEYUP:
           last_dispatch = dispatch_keyup;
           push_app_dispatch(L, last_dispatch);
-          lua_pushstring(L, SDL_GetKeyName(e.key.keysym.sym));
-          nargs = 2;
+          l_push_modifiers_table(L, e.key.keysym.mod);
```
lewri, line 308:
```suggestion
          lua_pushstring(L, SDL_GetKeyName(e.key.keysym.sym));
          l_push_modifiers_table(L, e.key.keysym.mod);
```

lewri:
Somehow this lost the API bump.

Issue #3277, opened 2026-03-18
CorsixTH is throttling and dropping frames


fusionstream:
### Describe the issue

Game resolution 3442x1440
Game speed 5

All of a sudden, with no updates, the game jitters by freezing for about a second every few seconds on any game speed (but with slightly reduced freezing time at slower speeds - it's still about a second).

To be clear, everything freezes except audio, even the mouse freezes in place. No new audio clips are played but whatever is playing does complete during the freeze.

### Steps to Reproduce

I first noticed this problem when there were too many people waiting in line for the reception desk. Obvious issue.

But in the 3.7 save file, you'll notice that there aren't that many people and not very many queueing at reception at all.

And in 3.8 you'll notice that there's barely any people.

### Expected Behaviour

Not to be jittery

### Save Game

[Saves.zip](https://github.com/user-attachments/files/26073054/Saves.zip)

### CorsixTH Version

0.69.2

### Operating System

Windows 10 Pro

### Hardware

- Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz   3.70 GHz
- RAM 32.0 GB (31.9 GB usable) (2x16)
- Graphics Car  NVIDIA GeForce RTX 3060 (12 GB), Intel(R) UHD Graphics 630 (128 MB)
- Storage HDD TOSHIBA HDWE140 3.64 TB
- Storage HDD TOSHIBA DT01ACA300 2.73 TB 
- Storage SSD Faspeed H5-500G, 466 GB 
- Storage SSD Samsung SSD 850 EVO M.2 500GB

### Theme Hospital Version

Origin

### Video

https://github.com/user-attachments/assets/c88a2c17-7025-4773-adcd-e17b06f57139

lewri:
Hello, that game log has come from a much older version of CorsixTH 0.62.

We can check the saves for the actual game log anyway.

ARGAMX:
Hi @fusionstream !

Thanks for posting this. 👍

Are you on CorsixTH's Discord?
If not, you can find the discord link on the description page of this repository.

Please ping me there in #general, I would like to discuss it there. @ARGAMX

lewri:
Extracted gamelog from .7 just indicates door deadlocks, and one instance of a bad queue:

<summary></summary>
<details>

```yaml
Savegame version is 222 (v0.69.2), originally it was 222 (v0.69.2).
Warning: Trying to resolve door deadlock at (106, 21)
[...]

PR #2762, opened 2025-01-18
Change reroute behaviour on trimmed paths

lewri:
*Fixes #2501*

**Describe what the proposed change does**
- Adds a new parameter to the `Walk` action where a trimmed path issued
- Instead of attempting a reroute inside the Walk, the action is finished instead
- In the case of `SeekReception`, that action calls a new `Walk` with an updated path.

@Alberth289346 and I discussed this one, and both agree currently there may not be a very nice way to do this, but as it's a regression of #1852 introduced in 0.67 it may be worth considering a fix for now. A better long term solution is recoding how walk is handled as a whole.

PR #3307, opened 2026-04-03
Cache reception desks

tobylane:
- Cache the sets of all and only staffed reception desks
- Add functions to get the necessary info (eg the existence of a staffed desk)
- Update the cache on a change in receptionists or desks

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/staff/receptionist.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff/receptionist.lua
@@ -75,6 +75,7 @@ function Receptionist:onPlaceInCorridor()
     local obj = world:getObject(x, y, "reception_desk")
     return obj and obj:occupy(self)
   end)
+  self.hospital:buildReceptionDesksCache() -- Rebuild cache of staffed desks
 end
 
 
@@ -90,6 +91,16 @@ function Receptionist:getDrawingLayer()
     return DrawingLayers.ReceptionistFacingAway
   end
   return DrawingLayers.ReceptionistFacingUser
+end
+
+function Receptionist:onPickup()
+  Staff.onPickup(self)
+  self.hospital:buildReceptionDesksCache() -- Rebuild cache of staffed desks
+end
+
+function Receptionist:fire()
+  Staff.fire(self)
+  self.hospital:buildReceptionDesksCache() -- Rebuild cache of staffed desks
 end
 
 function Receptionist:afterLoad(old, new)
--- a/CorsixTH/Lua/hospital.lua
+++ b/CorsixTH/Lua/hospital.lua
@@ -120,6 +120,7 @@ function Hospital:Hospital(world, avail_rooms, name)
     bench = 0,
     general = 0,
   }
+  self:buildReceptionDesksCache()
 
   self.num_visitors = 0 -- Counts patients only
   self.num_deaths = 0
@@ -503,6 +504,9 @@ function Hospital:afterLoad(old, new)
     self.hosp_cheats.active_cheats["spawn_rate_cheat"] = self.spawn_rate_cheat
     self.spawn_rate_cheat = nil
   end
+  if old < 243 then
+    self:buildReceptionDesksCache()
+  end
 
   -- Update other objects in the hospital (added in version 106).
   if self.epidemic then self.epidemic:afterLoad(old, new) end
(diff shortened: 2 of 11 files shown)


TheCycoONE:
I'll have to check how this affects the profile results. Thanks @tobylane

CorsixTH/Lua/hospital.lua
```diff
@@ -861,38 +865,34 @@ function Hospital:isPlayerHospital()
...
-  for _, desk in ipairs(self:findReceptionDesks()) do
-    if desk.receptionist or desk.reserved_for then return true end
-  end
-  return false
+--! Does the hospital have a reception desk
+--!param staffed (bool) Does the desk need to be staffed
+--!return (bool) Whether there is a reception desk in the hospital.
+function Hospital:hasReceptionDesk(staffed)
+  return staffed and self.staffed_reception_desks[1] or self.reception_desks[1]
 end
```
lewri, line 873: I think I preferred separate functions to determine if you wanted any desk or a staffed desk. Or the function doc/naming may need a bit of broader scope.

`is_staffed` may be enough to plainly denote it's an additional question.

Or `hasStaffedDesk` could call `hasReceptionDesk(true)`

(I'm not serious enough to block the merging of the functions though).

CorsixTH/Lua/humanoid_actions/seek_reception.lua
```diff
@@ -70,25 +70,23 @@ local function action_seek_reception_start(action, humanoid)
...
-      local distance_from_desk = humanoid.world:getPathDistance(humanoid.tile_x, humanoid.tile_y, x, y)
-      if distance_from_desk then
-        local this_score = distance_from_desk + desk:getUsageScore()
-        if not score or this_score < score then
-          -- It is better, or the first one!
-          score = this_score
-          best_desk = desk
-          best_distance_from_desk = distance_from_desk
-        end
+  for _, desk in ipairs(humanoid.hospital:getReceptionDesks(true)) do
```
lewri, line 73: May be worth checking this doesn't crash on an existing queue if you interrupt the patient's path before they join queue (I don't think it will, specifically referring to #2762 )

CorsixTH/Lua/entities/humanoids/staff/receptionist.lua
lewri, line 74: Not you but it would be nice to note placing ontop of the reception desk tile immediately assigns them to it.

CorsixTH/Lua/entities/humanoids/staff/receptionist.lua
```diff
@@ -92,6 +93,16 @@ function Receptionist:getDrawingLayer()
   return DrawingLayers.ReceptionistFacingUser
 end
 
+function Receptionist:onPickup()
+  Staff.onPickup(self)
+  self.hospital:findReceptionDesks() -- Rebuild cache of staffed desks
```
Alberth289346: `buildReceptionDesksCache()` instead ?

CorsixTH/Lua/humanoid_actions/seek_reception.lua
tobylane: I added to Object onClick to prevent a crash that may be this one.

CorsixTH/Lua/hospital.lua
tobylane: Yes, is_staffed is better.

CorsixTH/Lua/entities/humanoids/staff/receptionist.lua
tobylane: Done.

ARGAMX (commented):
Image

It's not you, but

While testing this PR, I found that
If you grab a secretary and place her miles away from the reception desk, `buildReceptionDesksCache()` counts that desk as staffed.

This way, patients will go to that desk and queue up for it, even though it's not actually being staffed yet.

0.69.2 has the same behavior, so it's not you.

I checked how this works in the original TH – there, a desk isn't considered staffed until a secretary actually stands at it.

ARGAMX (commented):
```Yaml
An error has occurred!
Running: The timer handler.
../Resources/Lua/world.lua:1007: attempt to call a nil value (method 'hasStaffedDesk')
stack traceback:
	../Resources/Lua/world.lua:1007: in method 'onEndDay'
	../Resources/Lua/world.lua:878: in method 'onTick'
	../Resources/Lua/app.lua:1305: in function <../Resources/Lua/app.lua:1302>
	(...tail calls...)
	../Resources/Lua/app.lua:1205: in function <../Resources/Lua/app.lua:1200>
```

[0.69.2.HVM3.7-Autosave-01-Oct.sav.zip](https://github.com/user-attachments/files/26487923/0.69.2.HVM3.7-Autosave-01-Oct.sav.zip)

Crash at 04 Oct.

tobylane:
Wow, that's a hospital for the profiler.

Fixed.

ARGAMX:
linked to #3277

ARGAMX (approved):
👍

TheCycoONE:
Did it speed up the frames, I haven't checked yet.

ARGAMX:
> Did it speed up the frames, I haven't checked yet.

I didn't compare them using a profiler.

But visually, I didn't notice any difference.

FPS showed a stable, small increase: in an experiment under identical conditions, it increased from 21-22 frames to 22-24.

tobylane:
`hasStaffedDesk` was called 1670 times for a total of 1.83s, with the latest calls taking 600us to 1.2ms. Average 1ms.

`hasReceptionDesk` was called 334 times for a total of 650us, with the latest calls taking 450ns to 1.60us. Average 2us.

(Is this the right way to report Tracy results?)

lewri (approved):

TheCycoONE:
Yeah that's good, thanks.

PR #3316, opened 2026-04-09
Fix flashing issue with Ubuntu 22.04

TheCycoONE:
It seems that the code got into an update loop where every time it called SDL_SetMinimumSize it detected a resolution change and tried to update again.

Work around that by only calling SDL_SetMinimumSize if the new minimum is different from the old one as returned by SDL_GetMinimumSize.

Fixes #3314


lewri (approved):
Nice

Issue #2302, opened 2023-01-30
[Bug] Tooltip not working outside window frame


lewri:
### Describe the issue

When loading up all language options in CorsixTH it extends past the height of the options window. As such, the tooltips are not displayed when hovering over languages outside this window.

I originally thought it was to do with the custom font displaying though languages, but it is not as Magyar and Polski both show tooltip.

### Steps to Reproduce

1. Set a font in settings
2. Hover over a language form the language list outside of the window's dimensions

### Expected Behaviour

Show tooltip. 

### Save Game

Save game attached to report: No

### CorsixTH Version

0.66

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Additional Information

![image](https://user-images.githubusercontent.com/20030128/215482566-338dcdc2-d74f-4225-8a62-6cf98fbef8e9.png)

lewri:
If you extend the height of `UIOptions` the tooltips will appear. So it's definitely referencing `UIOptions` instead of `UIDropdown` for showing the tooltip.

lewri:
Not sure if you can make a scrollable dropdown -- I can't see our current UI code supporting this out the box. We can support a 'menu list item', which includes a scrollbar, if a new daughter window is made instead.

Alberth289346:
OpenTTD has a dropdown child window too, should work.

PR #3318, opened 2026-04-12
Convert new_game to Colours + adjustment

lewri:
*Fixes #*

- Converts colour definitions to Colours
- Swaps Difficulty and Tutorial settings over so difficulty doesn't go outside of window (mitigation, see #2302 )

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/new_game.lua
+++ b/CorsixTH/Lua/dialogs/resizables/new_game.lua
@@ -24,38 +24,20 @@ class "UINewGame" (UIResizable)
 ---@type UINewGame
 local UINewGame = _G["UINewGame"]
 
-local col_bg = {
-  red = 154,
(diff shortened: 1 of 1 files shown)


Alberth289346:
Technically, it's not a refactor as you also change colours and modify positions.
But close enough.

I may have opted for `cols` instead of `col` as the table has more than one colour, but `col.xyz` may be more intuitive than `cols.xyz`.

Alberth289346 (approved):

lewri:
> I may have opted for `cols` instead of `col` as the table has more than one colour, but `col.xyz` may be more intuitive than `cols.xyz`.

Yep noted; I used col in UIOptions so was continuing the theme. I can make a correction PR to that file if you feel strongly about it.

TheCycoONE (approved):

PR #3320, opened 2026-04-12
Make sprite viewer reopenable

### LINKED ISSUE #1785: [Bug] Sprite Viewer breaks event handlers
#### Describe the issue
When you open the Sprite Viewer from debug menu in 0.64+ all event handlers error out.

#### Steps to Reproduce
1. Open a save/new game

2. Go to debug > Sprite Viewer. Console should output warnings as below and app may momentarily freeze.

3. Browse through spriteviewer then exit. Try to launch spriteviewer again and fails to  re-open.

#### Expected Behavior
Handle the Sprite Viewer properly and not cause event handler issues. 

#### System Information
**CorsixTH Version:** 0.64 (and Trunk 143)

**Operating System:** W10

#### Gamelog.txt

```
Warning: No event handler for motion
Warning: No event handler for active
Warning: No event handler for textediting
Warning: No event handler for textinput
Warning: No event handler for buttondown
Warning: No event handler for buttonup
```

### Additional Info
illustrate the problem.*

### LINKED ISSUE #2308: UIInformation box is loading pink

lewri:
*Fixes #1785*
*Fixes #2308*

- Make spriteviewer reopenable in the same session. It resumes where you left off (which is I assume what it already did previously anyway, you just couldn't access it).

Please could we get some additional checks this doesn't break saves or to the best of our ability any other part of the game handlers. I've tested myself and seems ok.

### DIFF
--- a/CorsixTH/Lua/dialogs/menu.lua
+++ b/CorsixTH/Lua/dialogs/menu.lua
@@ -873,7 +873,7 @@ function UIMenuBar:makeGameMenu(app)
         :appendCheckItem(_S.menu_debug_overlay.byte_7,      false, overlay(7, 7, true), "")
         :appendCheckItem(_S.menu_debug_overlay.parcel,      false, overlay("parcel"), "")
       )
-      :appendItem(_S.menu_debug.sprite_viewer, function() corsixth.require("sprite_viewer") end)
+      :appendItem(_S.menu_debug.sprite_viewer, function() corsixth.require("sprite_viewer")() end)
     )
   end
 end
--- a/CorsixTH/Lua/dialogs/menu.lua
+++ b/CorsixTH/Lua/dialogs/menu.lua
@@ -873,7 +873,7 @@ function UIMenuBar:makeGameMenu(app)
         :appendCheckItem(_S.menu_debug_overlay.byte_7,      false, overlay(7, 7, true), "")
         :appendCheckItem(_S.menu_debug_overlay.parcel,      false, overlay("parcel"), "")
       )
-      :appendItem(_S.menu_debug.sprite_viewer, corsixth.require("sprite_viewer"))
+      :appendItem(_S.menu_debug.sprite_viewer, function() corsixth.require("sprite_viewer")() end)
     )
   end
 end
--- a/CorsixTH/Lua/dialogs/menu.lua
+++ b/CorsixTH/Lua/dialogs/menu.lua
@@ -873,7 +873,7 @@ function UIMenuBar:makeGameMenu(app)
         :appendCheckItem(_S.menu_debug_overlay.byte_7,      false, overlay(7, 7, true), "")
         :appendCheckItem(_S.menu_debug_overlay.parcel,      false, overlay("parcel"), "")
       )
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/dialogs/menu.lua
```diff
@@ -873,7 +873,7 @@ function UIMenuBar:makeGameMenu(app)
         :appendCheckItem(_S.menu_debug_overlay.byte_7,      false, overlay(7, 7, true), "")
         :appendCheckItem(_S.menu_debug_overlay.parcel,      false, overlay("parcel"), "")
       )
-      :appendItem(_S.menu_debug.sprite_viewer, function() corsixth.require("sprite_viewer") end)
+      :appendItem(_S.menu_debug.sprite_viewer, function() corsixth.require("sprite_viewer")() end)
```
TheCycoONE, line 876: Should be able to simplify this by removing the lambda (wrapping function) and the parenthesis after sprite_viewer.

CorsixTH/Lua/sprite_viewer.lua
```diff
@@ -245,10 +249,23 @@ local function DoTimer()
...
+    frame = DoFrame,
+    keydown = DoKey,
+    keyup = DoKeyUp,
+    timer = DoTimer,
+  }
+
+  need_draw = true
+end
+
+return function()
```
TheCycoONE: Same, you don't need an anonymous function here, you can return SpriteViewer.open (without calling it)

TheCycoONE:
I don't think you need the SpriteViewer object at all, just return the open function?

CorsixTH/Lua/sprite_viewer.lua
```diff
@@ -245,10 +250,24 @@ local function DoTimer()
...
-  keydown = DoKey,
-  keyup = DoKeyUp,
-  timer = DoTimer,
-}
+local function open()
+  if is_open then
+    return
+  end
+  is_open = true
+  gfx.cache.tabled = {}
```
lewri: @TheCycoONE don't feel like this one should be reset every time it's opened? I can move it back...

lewri:
Wrapping function put back as discussed.

CorsixTH/Lua/sprite_viewer.lua
TheCycoONE: It actually should reset back to normal on close and to this or a new object on close.

ARGAMX (approved):
Works good for me
Issue resolved 👍

lewri:
Yep, don't merge yet until I resolve the gfx.cache.tabled thing. (Edit forgot it was on RFC :D )

lewri:
Proper cache swapping added, as a bonus fixes the pink issue.
Ready to review again.

TheCycoONE (approved):

PR #3321, opened 2026-04-12
Optimization for slow frames on large saves

TheCycoONE:
On large hospitals we spend a lot of time on
Humanoid:findObjectsInSquare. That function calls map:getRoomId repeatedly but gets the rest of the object information from the entity map. map:getRoomId was using the l_getcellflags c++ function which included a lot of cell information that was simply thrown away.

This specializes a version that just returns the roomId for this case. In my tests it improved frame time by about 50%.

Should help with #3277 - system specs matter. Looking for additional verification.

The new linux tracy profile applies compiler optimizations doesn't apply sanitizer hooks for a 'close' to release view.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -31,6 +31,20 @@
       }
     },
     {
+      "name": "linux-tracy",
+      "binaryDir": "${sourceDir}/build/tracy",
+      "generator": "Ninja",
+      "cacheVariables": {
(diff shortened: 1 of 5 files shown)


tobylane (approved):

lewri:
> Should help with #3277 - system specs matter. Looking for additional verification.

I can't verify as my pc is too powerful but will review.

lewri (approved):

PR #3323, opened 2026-04-16
Add accessibility menu to options

### LINKED ISSUE #3029: Accessibility Settings menu

There are now a number of issues which cover accessibility options, such as disabling screen shake, disabling flickering imagery, and subtitles. We should add an Accessibility Settings menu (similar to the Custom Settings menu) for these options so that they can be kept together in Settings.

All the accessibility options can be found in the same place.

No

0.69.1

Yes - see #3007, #3027, #3028

### Additional Information

_No response_

tobylane:
*Fixes #3029*

- Add an Accessibility settings dialog with settings for screenshake during earthquakes and announcer subtitles taken from the Customise dialog, and settings for scroll, shift scroll and zoom speeds taken from the Settings dialog. No new options at this time.
- This dialog is built with a new dialog builder, supported by refactored existing builder functions.
- Remove volume-casebook option
- Move strings around in all languages

When other dialogs, where possible, are moved to this dialog builder, it will be possible to make a consistent look in element widths, etc.

Screenshot 2026-04-25 at 21 51 08
Screenshot 2026-04-25 at 21 45 21
Screenshot 2026-04-25 at 21 45 09


CorsixTH/Lua/languages/english.lua
```diff
@@ -584,6 +574,35 @@ tooltip.audio_window = {
...
+  volume = "Volume Down Hotkey",
+  enable_screen_shake = "Screen Shake",
+  enable_announcer_subtitles = "Announcer Subtitles",
+  scrollspeed = "Scroll Speed",
+  shift_scrollspeed = "Shift Scroll Speed",
+  zoom_speed = "Zoom Speed",
+  back = "Back",
+}
+
+tooltip.accessibility_window = {
```
lewri, line 587: Are we updating remapping of strings of other languages too?

CorsixTH/Lua/dialogs/resizable.lua
```diff
@@ -208,6 +229,58 @@ function UIResizable:beginResize(x, y, mode)
   end
 end
 
+-- Create our setting items. This create a caption/label for the setting
+-- and the setting itself. We return both elements of the setting (panel
+-- and the button made from the panel)
+function UIResizable:createOptionsElement(option_label, option_tooltip,
```
lewri, line 285: I agree with the centralisations. While we have no fullscreen windows that use this resizable is fine to me.

CorsixTH/Lua/dialogs/resizable.lua
```diff
@@ -208,6 +229,58 @@ function UIResizable:beginResize(x, y, mode)
...
+  -- Make the setting value panel
+  local setting_panel = self:addBevelPanel(setting_x, y_pos, BTN_WIDTH,
+      BTN_HEIGHT, s_col.bg, s_col.highlight, s_col.shadow,
+      s_col.disabled, s_col.active)
+    :setLabel(setting_label)
+    :setTooltip(setting_tooltip)
+  -- Make the value panel a button
+  local setting_button = setting_panel:makeToggleButton(0, 0, BTN_WIDTH,
+      BTN_HEIGHT, nil, callback)
+    :setToggleState(toggle_state)
```
lewri, line 307: Did I saw making a button a toggle was harmless when writing this? I think I did.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -266,51 +197,38 @@ function UIOptions:UIOptions(ui, mode)
...
       { bg = col.setting, active = col.setting_active },
       self.dropdownAutosaveFrequency, false)
 
-
-  -- The right row is currently uneven with the left row, add an additional spacer
-  -- to avoid an overlap.
+  -- Odd number of settings, skip a row
   self:_getOptionYPos()
 
+  local upper_row_y_pos = self:_getOptionYPos() - 5
```
lewri: Why are we offsetting? Is it just to minimise the disruption to the other placement code?

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/accessibility.lua
```diff
@@ -0,0 +1,421 @@
...
+    self.ui:addWindow(UIMainMenu(self.ui))
+  end
+end
+
+-- Private functions
+
+--- Calculates the Y position for the dialog box in the option menu
+-- and increments along the current position for the next element
+-- @return The Y position to place the element at
+function UIAccessibility:_getOptionYPos()
```
lewri: Although we moved them, could these functions not be adapted to be centralised too since they've been moved to resizable?

lewri (commented):
Mostly seems okay, not had a look yet at the artefact.

CorsixTH/Lua/languages/english.lua
tobylane: Now done.

CorsixTH/Lua/dialogs/resizables/options.lua
tobylane: I've moved the offset but there still is one. It's a visual thing, it will likely change as more of the settings are added.

CorsixTH/Lua/dialogs/resizable.lua
Alberth289346: I am very hesitant here.

While I can see the value of more automating the construction, I fear you'll have different needs for different windows.

If you stick them all in the base class, but that may become a very big piece of code.

A different strategy is to have a builder class. It's also known as a factory class.
The only purpose of that class is to create things you need.

The advantage of a separate factory class is that it's simple to make a second or a third class, one for each "kind" of objects that you need.  
You can derive from other factory classes if it makes sense, but it is not required.

Alberth289346: To create a window, you first make a factory class, and configure it (for as far as needed).
Then you request the parts that you need from it, and you attach the parts to the window that you are creating.

CorsixTH/Lua/dialogs/resizable.lua
```diff
@@ -208,6 +229,58 @@ function UIResizable:beginResize(x, y, mode)
...
+end
+
+--- Calculates the Y position for the dialog box in the option menu
+-- and increments along the current position for the next element
+-- @return The Y position to place the element at
+function UIResizable:_getOptionYPos()
+  -- Offset from top of options box
+  local STARTING_Y_POS = 15
+  -- Y Height is 20 for panel size + 10 for spacing
+  local Y_HEIGHT = 30
```
Alberth289346: Such things should not be hard-coded in general.

ARGAMX (commented):
For me, the different rules for placing elements on different screens are still unclear.

Image

CorsixTH/Lua/config_finder.lua
```diff
@@ -510,6 +504,19 @@ param(config_values, 'remove_destroyed_rooms') .. [=[
...
+-- This way your autosaves folder can grow to 300-500 MB with daily autosaves.
+-- Set 1 for Monthly, 2 for Weekly, 3 for Daily autosaves.
+--]=] .. '\n' ..
+param(config_values, 'autosave_frequency') .. '\n'
+
+parts[5] = [=[
+--------------------------- ACCESSIBILITY SETTINGS ----------------------------
+-- These settings can also be changed from the Accessibility Menu
+-------------------------------------------------------------------------------
 -------------------------------------------------------------------------------
```
ARGAMX: I think the duplication of
```
```
line is unnecessary here

ARGAMX:
How our Settings screen looks like with this change:

Screenshot 2026-04-18 at 12 23 47

lewri:
> How our Settings screen looks like with this change:
> 
> Screenshot 2026-04-18 at 12 23 47

A 3+2 with centering might be nicer but at what point do we begin a scope creep? This is also part of why I started labelling the colours of those bottom ones as button, in case the UI visuals would benefit from those being different.

A true scope creep is tabulating the settings menu like hotkeys, but at this point might be a thing worth thinking about in future.

CorsixTH/Lua/dialogs/resizable.lua
lewri: STARTING_Y_POS probably shouldn't be defined and passed to the function (or be set on an init). Y_HEIGHT could be the same treatment. Especially where you want to centralise it.

It depends how uniform you want your window layouts I guess... The folders one I did recently has a different STARTING_Y_POS.

I get the idea though, but we're also a long way off of truly centralising window generation (and it isn't/shouldn't be this PR's scope to try and implement everything).

Maybe these centralisation changes maybe start off in their own PRs to maintain independence? We really need to come up with a window design standard too, but later.

TheCycoONE:
Did we add any options that weren't in dialogs before? Can the PR description please say where options were moved from?

Thanks @ARGAMX for pointing out the main settings screen. Would say it's in scope that we don't regress our settings dialogs while trying to improve their organization.

TheCycoONE:
Volume Down Hotkey doesn't need to exist does it? Seem redundant to the hotkeys dialog. Hardly an accessibility setting.

Tested this and neither option makes the volume down button open the casebook. Feels like a relic?

Edit:
Followed up on what this setting does, and it really is a waste of space. It just decides whether the casebook_alt (casebook 2) or the casebook_alt02 (casebook 3) hotkey is active (defaults to "c" or "shift+c")

Burn this setting, eliminate the casebook 3 / casebook_alt02 hotkey.

tobylane:
This almost exclusively only deals with Albert's request for a factory. The bare minimum is set in self.entry_list, and `UIResizable:buildDialog` makes that into the same dialog it was before. I will want to finish this last commit, and probably redo all the commits rather than two refactor commits in one PR. `UIResizable:onButtonClick` is an attempt to replace the repetitive button functions, but self needs replacing.

tobylane:
```lua
    self.labels[name], self.buttons[name] = self:createOptionsElement(
      _S.accessibility_window[name], _S.tooltip.accessibility_window[name],
      current_value, _S.tooltip.accessibility_window[name], nil,
      entry.func or self.onButtonClick(name, current_value), (not entry.raised) and self.app.config[name])
  end

  -- Back
  local back_button_y_pos = self:_getOptionYPos()
  self:addBevelPanel(20, back_button_y_pos, 280, 40, col.bg):setLabel(_S.options_window.back)
    :makeButton(0, 0, 280, 40, nil, self.buttonBack):setTooltip(_S.tooltip.options_window.back)
end

function UIResizable:onButtonClick(setting, value)
  return function()
    local app = TheApp
    app.config[setting] = not app.config[setting]
    self.buttons[setting]:toggle()
    self.labels[setting]:setLabel(get_button_label(setting, value))
    app:saveConfig()
    self:reload()
  end
end
```
Removed this onButtonClick function and its use beside entry.func that didn't work, but something like this would be useful in the future.

Addressed the other feedback, updated the first comment. Ready for review.

TheCycoONE:
I think several of the earlier placement comments still apply; the initial options menu is now very unbalanced, the title on the accessibility dialog isn't centered, the spacing on the accessibility dialog and the customize dialog are different.

The initial options menu layout is in scope because it was (mostly) balanced before this PR and isn't afterwards.

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/accessibility.lua
```diff
@@ -0,0 +1,348 @@
...
+  self.buttons.enable_announcer_subtitles:toggle()
+  self.labels.enable_announcer_subtitles:setLabel(app.config.enable_announcer_subtitles and _S.options_window.option_on or _S.options_window.option_off)
+  app:saveConfig()
+  self:reload()
+end
+
+function UIAccessibility:buttonScrollSpeed()
+  local callback = function(scrollspeed_number)
+    self.labels.scroll_speed : setLabel(tostring(scrollspeed_number))
+    self.buttons.scroll_speed : setToggleState(false)
```
lewri, line 91: These spaces don't seem intentional

CorsixTH/Lua/dialogs/resizables/menu_list_dialogs/accessibility.lua
```diff
@@ -0,0 +1,348 @@
...
+  end
+
+  self.ui:addWindow(UIShiftScrollSpeed(self.ui, callback))
+end
+
+function UIAccessibility:buttonZoomSpeed()
+  local callback = function(zoomspeed_number)
+    self.labels.zoom_speed : setLabel( tostring(zoomspeed_number) )
+    self.buttons.zoom_speed : setToggleState(false)
+  end
```
lewri, line 114: And at these points

tobylane:
I will add the dialog builder and convert an existing dialog to it in a new pr, then come back to this new dialog.

PR #3328, opened 2026-04-25
Scrap use of Font01V in loadMenuFont

### LINKED ISSUE #3278: Default text is not displayed in paths for Unicode
----

### Describe the issue

For some Unicode languages default text is not displayed in paths.

This was regressed in #3236.

v0.70.0 after

Image

.
v0.70.0 before

Image

----

### CorsixTH Version

current trunk master https://github.com/CorsixTH/CorsixTH/commit/f6931fd82ce54b2a3c9d0c1a6c991c776f209070

----

### Steps to Reproduce

1. Run game.
2. Go to Settings -> Folders.
3. Observer that default text displayed in paths correctly.
4. Press Back and change language to Russian.
5. Press Folders button (`Папки`).
6. Observe `????????? ?? ????????` text.

----

### Expected Behaviour

It should display the same as before this PR.

----

lewri:
*Fixes #3278*

- Prioritise using a defined unicode font if present when loading menu font.
- Fallback to builtin font if not set

The reason we want to prioritise unicode font is for maximum compatibility with OS file paths. The next compatible is builtin, as in `Font01V` it is missing `\` used by Windows and some other symbols.

To note: hotkey_assign has evidence it used this function but it actually doesn't. Instead it falls back to `Font01V` because that's the default set by `Window`. Arguably, it should probably use builtin in that window too.

A second note: Technically the function name is wrong by its current usage, but I don't have any ideas right now on something better fitting. I don't have immediate plans to change it but it can be done, I guess.

Checks required this doesn't break the Russian version more.

### DIFF
--- a/CorsixTH/Lua/graphics.lua
+++ b/CorsixTH/Lua/graphics.lua
@@ -483,9 +483,11 @@ end
 
 --! Utility function to return preferred font for main menu ui
 function Graphics:loadMenuFont()
-  local font
-  if self:_isLanguageSupportedByTHAssets() then
-    font = self:loadFontAndSpriteTable("QData", "Font01V", nil, nil, { apply_ui_scale = true })
+  -- Use the defined unicode font if set
+  local font = self:hasLanguageFont("unicode")
+  if font then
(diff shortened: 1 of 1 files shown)


ARGAMX (approved):
chuck-norris-approves-this-pr

PR #3333, opened 2026-05-01
Fix bundled font detection in Windows

### LINKED ISSUE #2936: Unable to detect CorsixTHUnicode font
On _new_ Windows installs, the game is unable to find the new bundled font on `C:\Program Files\CorsixTH`.

Image

lewri:
*Fixes #2936*

It turns out giving lfs an empty string is a bad idea. Giving it a blank string causes it to assume you want to look in `C:\`. However, lfs on windows seems to default to its own current directory (where `CorsixTH.lua` lives) when its path isn't overridden. Giving it a pure file or extended directory, however also works because it knows where to start from.

-  Remove scraping of font files from the CorsixTH install, instead define files to look for and lfs will find them if they exist in the CorsixTH.lua folder

Tested on Windows, briefly on Linux

You can verify on Linux by dumping a validly named font file in `/usr/share/corsix-th` (for arch based) and seeing if CorsixTH detects it (remove any font specified in the config first).
Similarly with Windows, but where CorsixTH.lua lives (with CorsixTH.exe)
I am not sure on Mac, but it seems to be `../Resources/` inside the .app from Argus' comments on discord.

### DIFF
--- a/CorsixTH/Lua/graphics.lua
+++ b/CorsixTH/Lua/graphics.lua
@@ -179,27 +179,30 @@ function Graphics:loadFontFile()
 
   local function getFontPath()
     local config_err, compile_err = "", ""
+    -- Check for font specified in config file
     if config_path then
       if check(config_path) then return config_path
       else config_err = "Configured font set but not found. Check path " .. config_path
       end
     end
+    -- Check for font specified in compile options
     if compile_path then
       if check(compile_path) then return compile_path
       else compile_err = " Compiled font path set but not found. Check path " .. compile_path
       end
     end
-    local path = self.app:getFullPath({})
-    for file in lfs.dir(path) do
-      for _, ext in pairs({"%.ttc$", "%.ttf$", "%.otc$", "%.otf$"}) do
-        if file:match(ext) then
-          return path .. file
-        end
+    -- Check the CorsixTH directory for our unicode font
+    local path = self.app:getFullPath()
+    local font_names = { "CorsixTHUnicode.ttf", "GoNotoKurrent-Regular.ttf" }
+    for _, name in ipairs(font_names) do
+      if check(path .. name) then
+        return path .. name
       end
     end
     return nil, config_err .. compile_err
   end
   local font_file, err = getFontPath()
+
   if not font_file then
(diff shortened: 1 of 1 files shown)


tobylane:
👍 
Why is it being given an empty string? You posted this in discord
```lua
function App:getCorsixTHDataDir()
  if TH.GetCompileOptions().os == "windows" then
    return lfs.currentdir() .. pathsep
  end
  return self:getFullPath(nil, true)
end
```
If this is needed, is it always needed?
```lua
function App:getFullPath(folders, trailing_slash)
  local root_dir = debug.getinfo(1, "S").source:sub(2, -12)
  if root_dir:len() == 0 then root_dir = lfs.currentdir() end
  if type(folders) ~= "table" then folders = {folders} end
  local ending = trailing_slash and pathsep or ""
  return root_dir .. table.concat(folders, pathsep) .. ending
end
```
If `getFullPath(nil, true)` is needed to get a trailing slash on Windows, that's different to Mac, where just `getFullPath()` ends in a slash. If the local path given in folders isn't a single file, always adding a slash may help.

```lua
local ending = folders[#folders] and folders[#folders]:match("%.lua$") and "" or pathsep
```

lewri:
> 👍 Why is it being given an empty string? You posted this in discord
> 
> ```lua
> function App:getCorsixTHDataDir()
>   if TH.GetCompileOptions().os == "windows" then
>     return lfs.currentdir() .. pathsep
>   end
>   return self:getFullPath(nil, true)
> end
> ```
> 
> If this is needed, is it always needed?

This isn't really needed now that I know what lfs is doing. However, it would be useful if you always want absolute path, for example.

ARGAMX (approved):
chuck-norris-approves-this-pr

TheCycoONE:
Reads ok; not planning on adding the fetch_font option in cmake in this PR?

TheCycoONE (approved):

lewri:
> Reads ok; not planning on adding the fetch_font option in cmake in this PR?

I still do not understand the magic that is cmake, I would leave that to someone else, please.

lewri:
@tobylane please merge if you're happy.

PR #3334, opened 2026-05-04
Add cmake option FETCH_UNICODE_FONT to download GoNoto unicode font

tobylane:
- With cmake switch -DFETCH_UNICODE_FONT=1, it will install the GoNoto font to the folder containing CorsixTH.lua, renamed as CorsixTHUnicode.ttf.

> -- Up-to-date: /Applications/CorsixTH.app/Contents/Resources/FluidR3.sf3
-- Installing: /Applications/CorsixTH.app/Contents/Resources/CorsixTHUnicode.ttf

Should this be included in cmake presets with the soundfont? The licence in the licence file?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@
 #   - WITH_LUAROCKS : Install required luarocks in the macOS app (requires luarocks)
 #   - FETCH_CATCH2 : Use FetchContent to obtain Catch2 if ENABLE_UNIT_TESTS=yes
 #   - FETCH_SOUNDFONT : Use FetchContent to download a soundfont to use (no)
+#   - FETCH_UNICODE_FONT: Use FetchContent to download the GoNoto Unicode font (no)
 
 # Tests and debug options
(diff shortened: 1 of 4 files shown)


lewri (commented):
Do we need to clean up `WITH_FONT` at all with this? This just I believe sets a path for the font file to use? Useful in Mac probably as it has proper Arial Unicode (which I think we do set).

It's likely you don't want `FETCH_UINICODE_FONT` and `WITH_FONT` at the same time/used to point at two different things.

tobylane:
Yes, certainly not both at once. The combined setting would have three values - yes, no, a path. That could be overstuffed. I don't think cmake has a neat way to tell a path apart from the various Boolean true representations.

TheCycoONE:
Yes it should go in the preset and license.

We don't need to do anything about WITH_FONT. Might be nice to warn about the incompatibility if both are set.

tobylane:
Added a cmake warning, "CorsixTH will attempt to use the WITH_FONT value ${WITH_FONT} before the fetched Unicode font."

TheCycoONE:
Nice! thanks @tobylane

TheCycoONE (approved):

PR #3337, opened 2026-05-07
[RFC] Treating an impossible meander

### LINKED ISSUE #3335: Inability to build a route to reception crashes Lua
Image

----
### Describe the issue
If you block the entrance to the hospital, then the patient who cannot navigate to the reception desk generates a Lua error.
In the original TH, this situation is impossible, as the game requires that there be a path between the patient respawn and the reception area.
However, the humanoid's movement algorithm should be resilient to such challenges and give valid results.

----
### Save Game
[3335-0.69.2-inability-to-build-a-route-to-reception-crash-Lua.zip](https://github.com/user-attachments/files/27410794/3335-0.69.2-inability-to-build-a-route-to-reception-crash-Lua.zip)

----
### CorsixTH Version
0.69.2

----
### Steps to Reproduce
1. Load attached save game №1.
2. Press 5.
3. Observe Lua crash.
4. Restart game.
5. Load attached save game №2.
6. Press 5.
7. Observe Lua crash.

----
### Expected Behaviour
While blocked patients should not throw errors or teleport.
Once the blockage is cleared, the patients must successfully build a route and begin moving.

----
### Gamelog.txt
```yaml
Error in timer handler:
Lua\humanoid_actions\walk.lua:31: Invalid value for parameter 'x'
stack traceback:
[C]: in global 'assert'
Lua\humanoid_actions\walk.lua:31: in local 'constructor'
Lua\class.lua:74: in global 'WalkAction'
Lua\humanoid_actions\meander.lua:114: in field '?'
Lua\entities\humanoid.lua:596: in method 'startAction'
Lua\entities\humanoid.lua:686: in method 'queueAction'
Lua\humanoid_actions\seek_reception.lua:140: in field '?'
Lua\entities\humanoid.lua:596: in method 'startAction'
Lua\entities\humanoid.lua:702: in method 'finishAction'
Lua\humanoid_actions\walk.lua:229: in local 'timer_function'
Lua\entity.lua:222: in field 'tick'
Lua\entities\humanoids\patient.lua:599: in method 'tick'
Lua\world.lua:916: in method 'onTick'
Lua\app.lua:1275: in function <Lua\app.lua:1272>
(...tail calls...)
Error in timer callback: 0
Warning: No event handler for timer
```

ARGAMX:
*Fixes #3335*

- If unable to meander, don't `finishAction()`, just Idle.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/meander.lua
+++ b/CorsixTH/Lua/humanoid_actions/meander.lua
@@ -80,7 +80,7 @@ local function meander_action_start(action, humanoid)
       humanoid.tile_y, math.random(1, 24), meander_parcel)
 
   if not x or (x == humanoid.tile_x and y == humanoid.tile_y and #humanoid.action_queue == 1) then
-    -- Somebody stuck in your hospital!
+    -- TODO: Call to show "Someone has got stuck. Plan your hospital better" adviser message should be here.
     -- Unable to find valid tiles for meandering. Queue an idle animation.
     action.can_idle = false
     humanoid:queueAction(IdleAction():setCount(5):setMustHappen(action.must_happen), 0)
--- a/CorsixTH/Lua/humanoid_actions/meander.lua
+++ b/CorsixTH/Lua/humanoid_actions/meander.lua
@@ -81,8 +81,7 @@ local function meander_action_start(action, humanoid)
 
   if not x or (x == humanoid.tile_x and y == humanoid.tile_y and #humanoid.action_queue == 1) then
     -- Somebody stuck in your hospital!
-    -- 'not x' - There weren't any available tiles for meander found.
-    -- Nowhere to walk to. We might be at a tight dead end. Let's just try Idle instead.
+    -- Unable to find valid tiles for meandering. Queue an idle animation.
     action.can_idle = false
     humanoid:queueAction(IdleAction():setCount(5):setMustHappen(action.must_happen), 0)
     return
--- a/CorsixTH/Lua/humanoid_actions/meander.lua
+++ b/CorsixTH/Lua/humanoid_actions/meander.lua
@@ -79,12 +79,12 @@ local function meander_action_start(action, humanoid)
   local x, y = humanoid.world.pathfinder:findIdleTile(humanoid.tile_x,
       humanoid.tile_y, math.random(1, 24), meander_parcel)
 
-  if x == humanoid.tile_x and y == humanoid.tile_y then
-    -- Nowhere to walk to - go idle instead, or go onto the next action
-    if #humanoid.action_queue == 1 then
-      humanoid:queueAction(IdleAction())
-    end
-    humanoid:finishAction()
+  if not x or (x == humanoid.tile_x and y == humanoid.tile_y and #humanoid.action_queue == 1) then
+    -- Somebody stuck in your hospital!
+    -- 'not x' - There weren't any available tiles for meander found.
+    -- Nowhere to walk to. We might be at a tight dead end. Let's just try Idle instead.
+    action.can_idle = false
+    humanoid:queueAction(IdleAction():setCount(5):setMustHappen(action.must_happen), 0)
     return
   end
   if action.todo_interrupt then
--- a/CorsixTH/Lua/humanoid_actions/seek_reception.lua
+++ b/CorsixTH/Lua/humanoid_actions/seek_reception.lua
@@ -124,8 +124,8 @@ local function action_seek_reception_start(action, humanoid)
     humanoid.hospital:msgMultiReceptionDesks()
 
   else
-    -- No reception desk found. One will probably be built soon, somewhere in
-    -- the hospital, so either walk to the hospital, or walk around the hospital.
+    -- No reachable reception desk found. One will probably be built soon, somewhere
+    -- in the hospital, so either walk to the hospital, or walk around the hospital.
     local procrastination
     if humanoid.hospital:isInHospital(humanoid.tile_x, humanoid.tile_y) then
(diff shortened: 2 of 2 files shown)


CorsixTH/Lua/humanoid_actions/meander.lua
```diff
@@ -79,12 +79,12 @@ local function meander_action_start(action, humanoid)
...
-  if x == humanoid.tile_x and y == humanoid.tile_y then
-    -- Nowhere to walk to - go idle instead, or go onto the next action
-    if #humanoid.action_queue == 1 then
-      humanoid:queueAction(IdleAction())
-    end
-    humanoid:finishAction()
+  if not x or (x == humanoid.tile_x and y == humanoid.tile_y and #humanoid.action_queue == 1) then
+    -- Somebody stuck in your hospital!
+    -- 'not x' - There weren't any available tiles for meander found.
+    -- Nowhere to walk to. We might be at a tight dead end. Let's just try Idle instead.
```
lewri: Condense this one into a comment block, though I feel like the comment as a whole is a bit disjointed. Is there a nice one liner to say what happened?

CorsixTH/Lua/humanoid_actions/meander.lua
```diff
@@ -79,12 +79,12 @@ local function meander_action_start(action, humanoid)
...
+    action.can_idle = false
+    humanoid:queueAction(IdleAction():setCount(5):setMustHappen(action.must_happen), 0)
```
lewri, line 86: Why have we dropped the finish action? This is usually needed to help invoke the queued one.

ARGAMX: > Why have we dropped the finish action? This is usually needed to help invoke the queued one.

Because if we can't perform an action (in this case, meandering/walking), that doesn't mean we should just discard that action. And that's exactly what it's doing here. Sometimes, that might does not lead to anything bad. But not here.

CorsixTH/Lua/humanoid_actions/meander.lua
ARGAMX: > Condense this one into a comment block, though I feel like the comment as a whole is a bit disjointed. Is there a nice one liner to say what happened?

The notes "-- Somebody stuck in your hospital!" are simply there so that in the future, Adviser can call these places to notify player.

I probably should have written it that way, but I failed miserably and have already started using that text in other places so I can later search for it and replace it with Adviser calls.

I don't understand what's wrong with the rest of the text.
Maybe you could suggest your own variant?

lewri:
```lua
-- A patient has become stuck because there are no valid tiles to move to.
-- Queue an idle animation until the route is clear.
```

ARGAMX: > ```lua
> -- A patient has become stuck because there are no valid tiles to move to.
> -- Queue an idle animation until the route is clear.
> ```

Ok, thanks.
Writing a clear and consistent comment here turned out to be quite challenging.
I've paraphrased it a bit, in part based on your suggestions. I hope it's better now.

tobylane (approved):

CorsixTH/Lua/humanoid_actions/meander.lua
```diff
@@ -79,12 +79,11 @@ local function meander_action_start(action, humanoid)
...
       humanoid.tile_y, math.random(1, 24), meander_parcel)
 
-  if x == humanoid.tile_x and y == humanoid.tile_y then
-    -- Nowhere to walk to - go idle instead, or go onto the next action
-    if #humanoid.action_queue == 1 then
-      humanoid:queueAction(IdleAction())
-    end
-    humanoid:finishAction()
+  if not x or (x == humanoid.tile_x and y == humanoid.tile_y and #humanoid.action_queue == 1) then
+    -- Somebody stuck in your hospital!
```
lewri:
```suggestion
    -- A humanoid is stuck in the hospital
```
We don't need to shout at ourselves 🙃. Also as a developer, in code there is no my hospital. My hospital would be on the graphical level when we need to tell the player.

lewri (commented):
Functionally good with me, just another comment nit.

Also found another undocumented behaviour, so yay....

CorsixTH/Lua/humanoid_actions/meander.lua
ARGAMX: > We don't need to shout at ourselves 🙃. Also as a developer, in code there is no my hospital. My hospital would be on the graphical level when we need to tell the player.

Okay, I changed it to this:
```
-- TODO: Call to show "Someone has got stuck. Plan your hospital better" advisor message should be here.
```

lewri: Sure

lewri (approved):
All fine, have gently nudged to update title for commit message to update its wider scope

ARGAMX:
Currently I'm plan to postponing this, skip 0.70.0, targeting this to 0.71.0.

ARGAMX:
This PR is primarily needed to enable humanoid blocking.
But this PR alone isn't enough, and more work needs to be done.

Furthermore, as usual, this change may have some negative effects. Or maybe not.
In any case, the functionality for blocking humanoids in hospital corridors is still in development.

I've decided that it's best to first implement blocking of hospital areas without the ability to block humanoids.
And only after that is completed, will consider blocking humanoids and return to this PR accordingly.

Otherwise, the block areas task becomes too enormous, with too many problems and things to solve at once.
I'll break down the task of blocking hospital areas and move forward gradually.
But then, at this stage, this PR isn't needed yet, and I rushed it.

I think I'll return to it later and reopen it.
For now, let it rest and gone from the list of open PRs.

PR #3339, opened 2026-05-09
Prevent message drawer overflow

### LINKED ISSUE #3327: Message drawer overflow
Image

----

### Describe the issue

The message drawer can fit only 5 messages.
But currently there can be many more of them and they all don’t fit.
For example, requests for salary increases can easily fill up the entire space.
This way you may miss a fax about the end of a level or about emergency patients.

The original TH solved this problem by not displaying more than one message of the same type in a drawer.
Messages of the same type accumulate in an invisible queue. 
Until the user freed up space from this one message, the next message of this type would not be pushed into the drawer.

This way, messages about deciding what to do with a patient or salary requests could not overflow the drawer.

----

### Save Game

[3327-0.69.2-message-drawer-overflow.sav.zip](https://github.com/user-attachments/files/27004289/3327-0.69.2-message-drawer-overflow.sav.zip)

----

### CorsixTH Version

0.69.2

----

### Steps to Reproduce

1. Load attached save game. 
2. Observe the drawer is full.

----

### Expected Behaviour

I have no better ideas than to do it like in the original game.
After this, by the way, the #1827 issue will have an even greater effect, so it is also advisable to solve it.

----

ARGAMX:
*Fixes #3327*

- Display a maximum 1 message of each type.
- If more than 1 message of the same type arrives, store all remaining messages of the same type in a separate "hidden" queue. (`message_queue_unshowed`)
- When a message is removed from the main queue, add a message from the "hidden" queue of the same type to the main queue, if there are any in the hidden queue.
- The rest is just refactoring and renaming.

Alternatively, instead of creating a new second table (`message_queue_unshowed`), we could have stored everything in a single main queue and displayed only one of each type, rather than all messages. 
However, this approach required refactoring the code related to the `factory_direction` and `factory_counter` variables, which are initialised with the magic numbers `-50`, `-1`, `9` and `22`, and which are used not only as counters, but also participate in checking conditions. I think this is some kind of ancient Aztec cipher that I don't understand (c99d04f). Only a superman can refactor it.
Unfortunately, doing this I broke my brain and fail at this approach. So I had to create a second table.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local SDL = require("sdl")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 243 -- Cache list of reception desks
+local SAVEGAME_VERSION = 244 -- Don't show same type messages in message drawer
 
 class "App"
 
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -40,8 +40,9 @@ function UIBottomPanel:UIBottomPanel(ui)
   self.show_animation = true
   self.factory_counter = 22
   self.factory_direction = 0
-  self.message_windows = {}
-  self.message_queue = {}
+  self.message_windows = {} -- fax icons in bottom inbox drawer. stores no more than 5 items in total and no more than 1 item of each message type.
+  self.message_queue = {} -- fax messages which have not yet reached the bottom inbox drawer. can store more than 5 items, but no more than 1 item of each message type.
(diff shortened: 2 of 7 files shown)


TheCycoONE:
You don't actually need to do anything with the counters (50, 9, 22) except know if there's a message to promote or not; they are to do with animations and the sliding door.

I think a one list solution is quite a bit cleaner; I'm open to trying it out myself if you don't want to.

ARGAMX:
> You don't actually need to do anything with the counters (50, 9, 22) except know if there's a message to promote or not; they are to do with animations and the sliding door.
> 
> I think a one list solution is quite a bit cleaner; I'm open to trying it out myself if you don't want to.

No, I've already gotten burned there and don't want to try again.
I didn't intend to interfere with these counters, but this approach required to interfere with `factory_direction` / `factory_counter`, otherwise the drawer damper don't close.

It's up to you. Give it a try. 👍

ARGAMX:
Closed in favour #3341 as more optimal solution

PR #3340, opened 2026-05-11
Prevent objects placing in tiles with humanoids

### LINKED ISSUE #3331: Humanoids can walk through walls
Image

----
### Describe the issue

Humanoids and VIPs in particular can walk through walls.
Crash case:
https://github.com/user-attachments/assets/ff90351b-8700-4098-a30a-980c4a6334d1
No crash case:
https://github.com/user-attachments/assets/27497e17-bb89-4821-93f0-5ac60842e179

----
### Save Game
[3331-0.69.2-VIP_in_room.zip](https://github.com/user-attachments/files/28925457/3331-0.69.2-VIP_in_room.zip)

----
### CorsixTH Version
0.69.2

----
### Steps to Reproduce
1. Load attached save game №3.
2. Observe a bench placed on VIP tile and a room placed in front of VIP.
3. Press 5 and observe how the VIP goes through the wall.
4. Load attached save game №2.
5. Build a room in front of VIP as it was in №3 save.
6. Don't place a bench.
7. Press 5 and observe how the VIP reroute and not go through the wall.
8. Load attached save game №2.
9. Build a room in front of VIP as it was in №3 save.
10. Place the bench on the same tile where VIP is located.
11. Press 5 and observe how the VIP goes through the wall.
----
### Expected Behaviour
Humanoids should not be able to walk through walls.

----
### Additional information
Thanks **Monkkero** from Discord for providing save before issue occurred.

ARGAMX:
*Fixes #3331*

- On `UIPlaceObjects:onMouseUp(button, x, y)` added additional check, that any humanoid don't occupy target cell.

I initially planned to use `EntityMap:getHumanoidsAtCoordinate(x, y)` for that purposes.
But this function turned out to be generally unsuitable for this target use case. Read the description before the function I added. I added an description to the function that explains why it wasn't suitable for this case.

After the initial idea with the `EntityMap:getHumanoidsAtCoordinate(x, y)` failed, I was sad.
Then I remembered that when building a room, a check is also made to ensure that humanoids do not occupy a certain territory.
The only minor issue with this function was that it checked by area. After all, no room could be smaller than 4x4. But I needed to check 1x1.
But tests with a function call `UIEditRoom.isHumanoidObscuringArea(humanoid, x - 1, x + 1, y - 1, y + 1)` gave a positive result and I decided to use this option, although it looks strange as if a 3x3 territory is being checked. But surprisingly, in this form it works exactly as it should for this task.

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -256,7 +256,7 @@ function UIEditRoom:confirm(force)
   end
 end
 
-local function isHumanoidObscuringArea(humanoid, x1, x2, y1, y2)
+function UIEditRoom.isHumanoidObscuringArea(humanoid, x1, x2, y1, y2)
   if humanoid.tile_x then
     if x1 <= humanoid.tile_x and humanoid.tile_x <= x2 and
         y1 <= humanoid.tile_y and humanoid.tile_y <= y2 then
@@ -294,7 +294,7 @@ function UIEditRoom:clearArea()
     local y2 = rect.y + rect.h
     for _, entity in ipairs(world.entities) do
       if class.is(entity, Humanoid) and
-          isHumanoidObscuringArea(entity, x1, x2, y1, y2) then
+          UIEditRoom.isHumanoidObscuringArea(entity, x1, x2, y1, y2) then
         humanoids_to_watch[entity] = true
 
         -- Try to make the humanoid leave the area
@@ -366,7 +366,7 @@ function UIEditRoom:onTick()
           if not humanoid.hospital then
             self.humanoids_to_watch[humanoid] = nil
           end
-        elseif not isHumanoidObscuringArea(humanoid, x1, x2, y1, y2) then
+        elseif not UIEditRoom.isHumanoidObscuringArea(humanoid, x1, x2, y1, y2) then
           self.humanoids_to_watch[humanoid] = nil
         end
       end
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -256,31 +256,6 @@ function UIEditRoom:confirm(force)
   end
 end
 
-function UIEditRoom.isHumanoidObscuringArea(humanoid, x1, x2, y1, y2)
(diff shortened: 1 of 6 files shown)


CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -256,7 +256,7 @@ function UIEditRoom:confirm(force)
   end
 end
 
-local function isHumanoidObscuringArea(humanoid, x1, x2, y1, y2)
+function UIEditRoom.isHumanoidObscuringArea(humanoid, x1, x2, y1, y2)
```
TheCycoONE: Let's move this function, calling an edit_room function from place_object is messy.

Moving to Humanoid itself seems like a good option (as a method instead)

TheCycoONE (changes requested):

CorsixTH/Lua/dialogs/edit_room.lua
ARGAMX: > Let's move this function, calling an edit_room function from place_object is messy.
> 
> Moving to Humanoid itself seems like a good option (as a method instead)

Yeah, I agree.
Moved to `humanoid`. 👍

tobylane:
Can the object be grey when not placable? Other than that, 👍.

ARGAMX:
> Can the object be grey when not placable? Other than that, 👍.

The problem is that humanoids move, sometimes creating, sometimes removing obstacles they create by themselves.
But the blueprint redrawing is currently triggered not by the external events in the simulation, but by the user moving the cursor.

So if we draw a gray cursor, then when a humanoid moves out from under the cursor tile, blueprint will still remain gray.
Which would also be "incorrect".
Since previously, the moving objects was unable to create an obstacles, blueprint redrawing by the user moving the cursor was enough.

To accordingly update the cursor, we would literally need to put this check in a continuous loop.
Or somehow create a subscription to the tiles we're interested in.
In short, I don't think this is worth it.

I checked this in the original game – Bullfrog didn't make such a hassle for themselves either. So it's fair.
There, the blueprint is also always colored on humanoids in original.

But if someone wants to implement this, they welcome. 😀

TheCycoONE (approved):

PR #3341, opened 2026-05-13
Prevent multiple faxes of the same type

### LINKED ISSUE #3327: Message drawer overflow

TheCycoONE:
Fixes #3327

- Display a maximum 1 message of each type.

This is an alternative (I think simpler) alternative to #3339

### DIFF
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -550,10 +550,10 @@ function UIBottomPanel:_showMessage()
   end
 end
 
--- If there is any message in the queue that is suitable to be shown
--- as in there is no currently shown message of the same type and
--- there is less than 5 currently shown messages, then return it's
--- index, otherwise return nil.
+-- Return the index of the first fax message in the queue that is suitable to be
+-- shown, and nil if there are none. A message is suitable to be shown if there
+-- are less than 5 currently shown messages, no message of the same type is
+-- already shown.
 function UIBottomPanel:_findMessageToShow()
   if #self.message_windows >= 5 then
     return nil
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -24,6 +24,12 @@ class "UIBottomPanel" (Window)
 ---@type UIBottomPanel
 local UIBottomPanel = _G["UIBottomPanel"]
 
+local FACTORY_DIR_OPENING = -1
+local FACTORY_DIR_CLOSING = 1
+local FACTORY_DIR_NONE = 0
+
+local FACTORY_ANIM_TICKS = 22
+
 function UIBottomPanel:UIBottomPanel(ui)
   self:Window()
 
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -516,22 +535,44 @@ end
...
-function UIBottomPanel:showMessage()
-  if self.factory_direction ~= -1 then
-    self.factory_direction = -1
+function UIBottomPanel:_showMessage()
+  if self.factory_direction ~= FACTORY_DIR_OPENING then
+    self.factory_direction = FACTORY_DIR_OPENING
     if self.factory_counter < 0 then
       -- Factory is already opened so don't wait to show the message
       self.show_animation = false
       self.factory_counter = 9
```
ARGAMX, line 544: Is this `9` the delay between the continuous output of two visible items in the bottom panel without the door closing? Can we also save this number as a named constant?

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -590,24 +633,24 @@ function UIBottomPanel:createMessageWindow(index)
     onClose, message_info.type, message_info.message, message_info.owner, message_info.timeout, message_info.default_choice, message_info.callback)
   message_windows[#message_windows + 1] = alert_window
   self:addWindow(alert_window)
-  self.factory_direction = 1
+  self.factory_direction = FACTORY_DIR_CLOSING
   self.show_animation = true
   self.factory_counter = -50                -- Delay close of message factory
```
ARGAMX, line 638: I might try opening another PR later to try and bring some more order here.
For example, these three commands are clearly linked by some kind of named action.
But the code is silent about this.
And personally, this wasn't entirely obvious to me initially. Only now do I realize it after spending time on these counters.

ARGAMX (approved):
Thanks, works smooth for me 👍

tobylane (approved):

CorsixTH/Lua/dialogs/bottom_panel.lua
TheCycoONE: yep

PR #3343, opened 2026-05-15
RFC: Modify rendering, add proxies

Alberth289346:
It took quite some time, but here is the branch that should improve rendering of the game. It does the following things:
* The existing way of rendering the game was quite complicated and still had graphic glitches. My experiments with simpler forms of rendering suggest that there is no simple way to render the game while eliminating glitches.
  The complicated way thus did not have many advantages and was dropped. It is replaced with drawing tiles in a simple nested loop. The outermost loop increments y. The innermost loop increments x. This aligns with the order of tiles in memory.
* This change makes the early list and scanline iterators obsolete, so they are removed.
* At Lua level, animations were duplicated to avoid glitching (Each animation of the same object draws a column of it). This approach eats Lua processing time, has risks of duplicated animations getting out of sync, and makes handling an animation in Lua more complicated.
  The duplication is eliminated by adding proxies. A proxy also draws a columns of an animation, but all proxies of an object use the same animation. The proxies are managed at C++ level, which simplifies animation handling at Lua level and reduces Lua computation time.
* Due to the different way of rendering tiles, the proxies are needed at different spots in the map. Humanoid walking and the objects have been adjusted.

----
As for what is next:

* There is an open question what to do with depersisting. I looked, but did not find a way to distinguish between 'old' saves and 'new' saves.
* I don't know if there is any way to rescue existing saves. It would be helpful in testing the patch.
* Likely other things should or must be done, but I am not aware of any such thing currently.


CorsixTH/Src/th_gfx.cpp
TheCycoONE, line 1486: It might be possible to write your version marker here, since there's only a limited set of valid draw flags. I believe the top half of the word is free.

CorsixTH/Src/th_map.cpp
```diff
@@ -953,293 +953,216 @@ bool level_map::layer_exists(uint16_t layer, int& height) const {
...
-    // Draw floor shadows immediately after floor tiles ensuring that all
-    // shadow pixels are drawn onto freshly drawn opaque floor tile pixels.
-    if (itrNode1->flags.shadow_full) {
-      wall_blocks->draw_sprite(pCanvas, 74, tile_x, tile_y,
-                               thdf_alpha_75 | thdf_nearest);
-    } else if (itrNode1->flags.shadow_half) {
-      wall_blocks->draw_sprite(pCanvas, 75, tile_x, tile_y,
-                               thdf_alpha_75 | thdf_nearest);
+namespace {
+const int left_safety = 3 * 64;
```
TheCycoONE: `constexpr`

TheCycoONE:
Yes it would be very helpful for testing the patch ... mmm

TheCycoONE:
Found a glitch:
image

TheCycoONE:
I like the new rendering approach. I'd love to see it benchmarked on some of the larger hospitals but that will require putting some effort into save compatibility.

I suspect without a lot of effort we're going to regress substantially on clipping issues that slowly made the previous renderer messier and messier.

CorsixTH/Src/th_gfx.cpp
Alberth289346: No simple way to solve this problem for the general case, I guess?

TheCycoONE: Maybe there is; I'll consider implementing a version on the save that is made available to all the depersist functions.

mugmuggy:
I had experimented with similar rendering loops. I did the increasing Y then X, X then Y. I also had a rendering loop down X and Y like nested chevrons from top and from bottom. Everytime it was a glitch from a humanoid walking between tiles, they weren't anchored to the map tile but moving. Or an animation that is anchored to a tile that isn't the objects location - which is the screenshot Cyco. When its drawing the object in that screenshot, its the tile where the doctor is using it from, is rendered. If you clip it, you only draw the Dr, when looping through the tiles you have no idea if an animation is there or not. to half render the cabinet or not.

If you processed by rooms, you would draw the 'back walls' where you see the interior, render objects in the room, render the 'front walls' which are the exterior. You would still need to solve for object rendering order in the room. The walking humanoids, it depends on the direction they are heading for the glitch to show up.

Alberth289346:
> When its drawing the object in that screenshot, its the tile where the doctor is using it from, is rendered. If you clip it, you only draw the Dr, when looping through the tiles you have no idea if an animation is there or not. to half render the cabinet or not.

I know.

The current implementation mostly fixes that by running multiple instances of the same animation at Lua level, each drawing a small part of the animation. In my patch here, I use the same trick but at C++ level without duplicating complete animations.  
An animation can have proxies, which are basically wrappers for their parent animation attached to different tiles. The animation and each proxy has their own cropping parameters. As the renderer visits the tiles, parts of a multi-tile animation can be drawn before other animations in front of it are drawn.

I don't understand why that is apparently not working. I did modify all objects to have proxies and checked if they were correct.

Not an issue in this case, but I also fixed the humanoids walking.

The best theory I came up with so far, is that the proxies are lost when a cabinet is rendered with a different animation. That would explain the screenshot.

ARGAMX:
I created a build on this thread and played for a bit.

I was immediately pleasantly surprised that I could load any save game.

Although after loading, you can see that some animations are broken, misaligned, and not displayed properly. But the fact that the game loads at all is pleasing.

Loaded v0.69.2 save:
Image

Then I started a new game and began building a hospital from scratch. But even building a hospital from scratch, I encountered numerous animation glitches.

The reception desk overlaps patients and doctors in front of it.
Image

When someone uses the door, the animation doesn't appear at all. The door and the patient disappear. But when the animation ends the door and the patient appears.
Image

But some doors disappear forever and never appear again.
You may also notice that the bench is also slightly cut off.
Screenshot 2026-05-20 at 21 08 43

The changing screen disappeared after use.
Screenshot 2026-05-20 at 20 39 48

The patient disappeared during treatment at the pharmacy.
Screenshot 2026-05-20 at 20 42 59

If place a surgical screen in this position in a room, it disappears after installation.
Screenshot 2026-05-20 at 20 49 27

The inflation is cut off  the way we used to love to cut off sofas.
Remember that time? Oh, how young we were...
Screenshot 2026-05-20 at 20 58 30

Projector in the lecture room.
Screenshot 2026-05-20 at 21 07 37

I played for about 40 minutes, saving the game periodically.
But then something made me want to load the save I made.
Unfortunately, saves made with this build won't load.

```Lua
<LocalisedString> Current value:Error while loading game: ../Resources/Lua/persistance.lua:289: sync fail after 367 objects (122657 bytes)
```

And that's it. The game won't load.
If an error occurs while loading a save, you're kicked out of the current level and taken to the main menu.

[Albert_proxies_1.sav.zip](https://github.com/user-attachments/files/28070210/Albert_proxies_1.sav.zip)

[Albert_proxies_2.sav.zip](https://github.com/user-attachments/files/28070215/Albert_proxies_2.sav.zip)

That's how my game ended.
My level reset, I couldn't continue playing it, and I can't load the save either.
At this point, I felt bit sad, as I had already fallen in love with the hospital I built. 
It was so young...

That was the end of my experiment for today. 🙂

Alberth289346:
Thanks for trying.

I am not surprised you can load an existing save, since that happens entirely outside our control. Pretty likely a running animation will continue to work. I am however not sure what happens next.  
There is no code that deals with a table of animations anymore. Likely the next animation overwrites the table, and that's the end of the table with animations.

As for the glitches, I think the root cause is the same as in the screenshot of thecyco; a humanoid is used to perform the animation, which means the object `render_attach_positions` are not applied. As a result, it glitches all over the place.

Loading a save may still be broken. I raised a question how to deal with depersistence (ie loading saves), but there is not yet an answer. Currently it does something such that you can at least start a new game, without guarantee you can load anything (probably).

PR #3344, opened 2026-05-15
[Draft] Prevent return to the toilet after leaving the toilet

### LINKED ISSUE #555: Patients in a loop of loo use, keep on re-entering the Toilet room
The issue is still relevant for 2026.

----

### Describe the issue

Patients in a loop of loo use, keep on re-entering the Toilet room

----

https://github.com/user-attachments/assets/ad0823d5-cc79-4ade-825f-510b399a3500

----

### Save Game

[toilet-test.zip](https://github.com/user-attachments/files/25804645/toilet-test.zip)

----

### CorsixTH Version

trunk master 0.70.0 53d8091

----

### Steps to Reproduce

1. Load attached save game. 
2. Observe that there is a new toilet in the upper, northwest part of the hospital. Its queue is empty yet. Several patients are heading to it.
3. Observe that when these patients start to use that room, after use and exit the room patients will go back to the toilet queue. So they get stuck here forever, creating a never-ending queue.

----

### Expected Behaviour

After using the toilet, patients should go on with their business, move further into diagnosis and treatment, rather than queueing for the same toilet again.

----

Earlier issue description (2014):

_From [wolfy1...@gmail.com](https://code.google.com/u/111955299291263404988/) on May 07, 2013 17:33:54_

What steps will reproduce the problem? 1.i am not sure how to reproduce this problem. 2. 3. What is the expected output? What do you see instead? A patient that just used the toilets returns even though he does not need to.

0.21 What operating system / compile settings are you using? Windows 8 Pro 64 bit
Full Game level Attach a saved game file (preferably an autosave from before the error) if you have one and it is relevant. Please provide any additional information

**Attachment:** [config.txt gamelog.txt thomas01.sav](http://code.google.com/p/corsix-th/issues/detail?id=1623)

_Original issue: http://code.google.com/p/corsix-th/issues/detail?id=1623_

ARGAMX:
*Fixes #555*

- When leaving the toilet, if the room where the patient went before the toilet is a toilet, do not return to the toilet.
- Instead of that to decide which room should be next.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/patient.lua
+++ b/CorsixTH/Lua/entities/humanoids/patient.lua
@@ -194,7 +194,7 @@ end
 
 --! Updates the patients diagnostic progress based on the doctors skill
 -- called when they are done using a diagnosis room
-function Patient:completeDiagnosticStep(room)
+function Patient:advanceDiagnosticProgress(room)
   -- Base: depending on difficulty of disease as set in sam file
   -- tiredness reduces the chance of diagnosis if staff member is above 50% tired
   local multiplier = 1
@@ -354,10 +354,6 @@ function Patient:isTreatmentEffective()
   cure_chance = cure_chance + (service_base * service_factor)
 
   return (cure_chance >= math.random(1,100))
-end
-
-function Patient:hasMoreDiagnosisRoomsAvailable()
-  return #self.available_diagnosis_rooms ~= 0
 end
 
 --! Change patient internal state to "cured".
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -1366,5 +1362,72 @@ function Patient:interruptAndRequeueAction(current_action, queue_pos, meander_be
...
+  end
+
+  if self.diagnosed then
+    self:sendPatientToTreatmentRoom()
+    return
+  end
+
+  local last_room_gp = self.treatment_history[#self.treatment_history] == _S.rooms_short.gps_office
+  if last_room_gp then
+    self:sendPatientToNextDiagnosisRoom()
```
lewri, line 1382: It is likely the real next diagnosis room was lost. As a result, we're punishing the player because the humanoid went to the toilet really. As such, they will in the grand scheme of things lose diagnostic progress.

It also means it's now possible a patient goes home after using the toilet. This is not good.

CorsixTH/Lua/entities/humanoids/patient.lua
```diff
@@ -1366,5 +1362,72 @@ function Patient:interruptAndRequeueAction(current_action, queue_pos, meander_be
...
+    self:setDiagnosed()
+  end
+end
+
+function Patient:sendPatientToGPRoom()
+  if self:agreesToPay("diag_gp") then
+    self:queueAction(SeekRoomAction("gp"))
+  else
+    self:goHome("over_priced", "diag_gp")
+  end
```
lewri, line 1404: We're now rolling the dice twice here:
1. For checking we want the gp after the previous diagnosis rooms/reception
2. For checking when we come out the toilet

The answer could be different in both scenarios.

lewri (changes requested):
This needs to go one step earlier to fix; and properly cache the previous target room of the patient to avoid any negative effects.

There are too many negatives on the player for a patient needing the toilet here to recover.

If I also remember, when they 'do it on the floor', they probably should be dropping any toilet queued action anyway.

ARGAMX:
> This needs to go one step earlier to fix; and properly cache the previous target room of the patient to avoid any negative effects.
> 
> There are too many negatives on the player for a patient needing the toilet here to recover.
> 
> If I also remember, when they 'do it on the floor', they probably should be dropping any toilet queued action anyway.

Okay. Fair enough.
I honestly hadn't thought of that.
Thanks.

But that requires completely changing the fix concept, because this approach doesn't satisfy the conditions at all.
I think I'll pass on this task for now, because I don't have an idea yet on how to fix it properly.

PR #3345, opened 2026-05-16
Close the door if fax isn't available

TheCycoONE:
Fix regression in #3341

Refactors the fax door (factory) logic to be quite a bit simpler and less prone to bad state issues.

One thing that was dropped was that there was two different delays for an open fax door, 9 ticks to show the next message from whenever it became available if the door was open (which sounds a bit broken when said out loud) and 50 ticks to begin closing the door after popping the fax. I reduced them to a single 9 tick delay for either action. In my play experience this seems fine, but if others consider it unacceptable I can introduce a separate delay counter for each.

I made a separate branch that keeps the logic at https://github.com/CorsixTH/CorsixTH/commit/25a89eff6311289b505164cfead6e55e45ff2f38 to compare.

### DIFF
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -24,9 +24,6 @@ class "UIBottomPanel" (Window)
 ---@type UIBottomPanel
 local UIBottomPanel = _G["UIBottomPanel"]
 
-local FACTORY_DIR_OPENING = -1
-local FACTORY_DIR_CLOSING = 1
-
 local FAX_DOOR_FULLY_OPEN = 0
 local FAX_DOOR_FULLY_SHUT = 22
 
@@ -608,9 +605,7 @@ function UIBottomPanel:createMessageWindow(index)
     onClose, message_info.type, message_info.message, message_info.owner, message_info.timeout, message_info.default_choice, message_info.callback)
   message_windows[#message_windows + 1] = alert_window
   self:addWindow(alert_window)
-  self.factory_direction = FACTORY_DIR_CLOSING
   self.show_animation = true
-  self.factory_counter = -50                -- Delay close of message factory
   table.remove(self.message_queue, index)   -- Delete the last element of the queue
 end
 
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -26,9 +26,9 @@ local UIBottomPanel = _G["UIBottomPanel"]
 
 local FACTORY_DIR_OPENING = -1
 local FACTORY_DIR_CLOSING = 1
-local FACTORY_DIR_NONE = 0
 
-local FACTORY_ANIM_TICKS = 22
+local FAX_DOOR_FULLY_OPEN = 0
+local FAX_DOOR_FULLY_SHUT = 22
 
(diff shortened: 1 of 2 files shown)


ARGAMX (commented):
Nice, thank you! 👍

I can no longer reproduce the bug in this form, when the drawer breaks completely.
Now messages are always delivered.

BUT

This unfortunate drawer door still won't close successfully and visually remains open.

Just load this save, set the speed to 5, soon one of the staff will ask for a raise, a message will arrive and the door will remain open.

[Inbox-Drawer-Defects-07.sav.zip](https://github.com/user-attachments/files/27852134/Inbox-Drawer-Defects-07.sav.zip)

Screenshot 2026-05-16 at 12 06 04

TheCycoONE:
Sounds like we're going to rewrite this component. I had started a larger refactor but tossed it; I guess I'll write a new one.

TheCycoONE:
@ARGAMX save link doesn't seem to work

TheCycoONE:
Ok, I got around to the larger refactor I was talking about. @ARGAMX can you see how this does on your save; I think it's much more 'stuck proof' and simpler to reason about.

ARGAMX (approved):
Great, thanks! 
This is much better, and all the logic is inside a single function, `_faxDoorTick`.
I tried it out and tried to break it on my saves, but it works smooth.
Looks pretty robust and much better that was. 👍

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -681,14 +624,32 @@ function UIBottomPanel:onTick()
...
-  -- Start moving message out of the queue if a suitable one is available
-  if self:_findMessageToShow() then
-    self:_showMessage()
-  end
+  self:_faxDoorTick()
 
   Window.onTick(self)
 end
 
+function UIBottomPanel:_faxDoorTick()
```
lewri: Strictly speaking this is a message door/drawer etc, because staff pay rise messages also come through this

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -1032,6 +993,21 @@ function UIBottomPanel:afterLoad(old, new)
   if old < 236 then
    self:_initFonts(self.ui.app.gfx)
   end
+  if old < 244 then
+    -- ignore the delay logic for save games, minor
+    if self.factory_counter < FAX_DOOR_FULLY_OPEN then
```
lewri: Risk of a compare number with nil error, unlikely however.

lewri (approved):
Apart from the nit that `fax` is the wrong terminology it seems ok.

CorsixTH/Lua/dialogs/bottom_panel.lua
TheCycoONE: Sure, but it goes beyond my change, like canQueueFax - fax naming is used all over

CorsixTH/Lua/dialogs/bottom_panel.lua
TheCycoONE: How?

CorsixTH/Lua/dialogs/bottom_panel.lua
TheCycoONE: Changed fax=>message for my changes, didn't expand scope of PR to change everywhere.

PR #3347, opened 2026-05-17
Allow to open a fax even if the message isn't in the drawer

### LINKED ISSUE #1827: [Bug] Wage requests can't be viewed if drawer is full
#### Describe the issue
If your drawer on the bottom is full of faxes then and you have a staff member unhappy, you can't view their wage request by clicking on them.

#### Steps to Reproduce
1. Get a full bottom drawer of faxes.

2. Find an unhappy staff and click. It actually brings up the staff profile instead of the wage request.

#### Save Game

[no_wage_increase.zip](https://github.com/CorsixTH/CorsixTH/files/6118748/no_wage_increase.zip)

#### Expected Behavior
Allow wage requests to be viewed regardless of drawer fullness by clicking on staff.

#### System Information
**CorsixTH Version:** *0,64*

#### Gamelog.txt

```
Paste gamelog.txt output here
```

### Additional Info
illustrate the problem.*

ARGAMX:
*Fixes #1827*

- If the incoming message had not yet get into the drawer, then it was impossible to open it in any other way, not from the drawer. This limitation has been removed.
- For example, suppose 10 doctors want a promotion, but only 5 of them will show up in the drawer. If we find one of the doctors in the hospital whose message is in the drawer and click on they, the promotion request will be displayed on screen. If we find one of the doctors in the hospital whose message isn't in the drawer and click on they, the promotion request won't be displayed. The situation is similar with patients who don’t know where to go and ask for your decision.
- This issue happens because the `UIMessage` object, which is responsible for displaying a fax or promotion request on the screen, was created too late, only when the message got in the drawer. Therefore, if the message wasn't in the drawer, it simply couldn't be opened.
- And since we implemented the restriction to prevent messages of the same type from displaying in the drawer at the same time in #3341, the situation with this issue has become even worse. Because now out of 10 simultaneous requests for salary increases, only 1 is displayed in drawer. And only that 1 can be opened.
- So now `UIMessage` are created as soon as an incoming message is received in the queue.

### DIFF
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -610,12 +610,11 @@ function UIBottomPanel:_showMessageIcon(index)
   if not message_info then
     return
   end
-  -- Create the message window, note this does not show it to the player on creation.
-  local alert_window = UIMessage(self.ui, 175, 1 + #message_windows * 30,
-    onClose, message_info.type, message_info.message, message_info.owner, message_info.timeout, message_info.default_choice, message_info.callback)
-  message_windows[#message_windows + 1] = alert_window
-  self:addWindow(alert_window)
-  table.remove(self.message_queue, index)   -- Delete the last element of the queue
+  local drawer_icon = message_info.drawer_icon
+  drawer_icon:setXLimit(1 + #message_windows * FACTORY_ICON_WIDTH)
+  message_windows[#message_windows + 1] = drawer_icon
+  self:addWindow(drawer_icon)
+  table.remove(self.message_queue, index)   -- Delete element of the queue
 end
 
 function UIBottomPanel:onTick()
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -461,7 +461,9 @@ function UIBottomPanel:queueMessage(type, message, owner, timeout, default_choic
       -- If icon position found
       if this_icon_index ~= nil then
         -- Shift all icons that are on the right to the left
-        icon:setXLimit(1 + (i - 2) * FACTORY_ICON_WIDTH)
+        -- The '_factoryIconWitdh()' call is used for compatibility with 0.69 saves and older
+        -- where FACTORY_ICON_WIDTH was not yet declared when the closure was created.
+        icon:setXLimit(1 + (i - 2) * self._factoryIconWitdh())
       elseif icon == this_icon then
         this_icon_index = i
       end
@@ -493,6 +495,10 @@ function UIBottomPanel:queueMessage(type, message, owner, timeout, default_choic
   else
     self:cancelFax(message_info.type)
   end
+end
+
+function UIBottomPanel:_factoryIconWitdh()
+  return FACTORY_ICON_WIDTH
 end
 
 --[[ A fax can be queued if the event the fax causes does not affect
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -27,8 +27,9 @@ local UIBottomPanel = _G["UIBottomPanel"]
 local FACTORY_DIR_OPENING = -1
 local FACTORY_DIR_CLOSING = 1
 local FACTORY_DIR_NONE = 0
-
+local FACTORY_PAUSE_BEFORE_NEXT_TICKS = 9
 local FACTORY_ANIM_TICKS = 22
+local FACTORY_ICON_WIDTH = 30
 
 function UIBottomPanel:UIBottomPanel(ui)
   self:Window()
@@ -453,7 +454,23 @@ function UIBottomPanel:queueMessage(type, message, owner, timeout, default_choic
     self.world.ui.adviser:say(_A.information.fax_received)
     self.ui.hospital.message_popup = true
   end
-  local fax = {
+
+  local --[[persistable:bottom_panel_message_window_close]] function onClose(this_icon)
+    local this_icon_index
+    for i, icon in ipairs(self.message_windows) do
+      -- If icon position found
+      if this_icon_index ~= nil then
+        -- Shift all icons that are on the right to the left
+        icon:setXLimit(1 + (i - 2) * FACTORY_ICON_WIDTH)
+      elseif icon == this_icon then
+        this_icon_index = i
+      end
+    end
+    table.remove(self.message_windows, this_icon_index)
+    self:deleteMessage(this_icon.owner)
+  end
+
+  local message_info = {
     type = type,
     message = message,
     owner = owner,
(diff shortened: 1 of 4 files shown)


CorsixTH/Lua/dialogs/message.lua
```diff
@@ -147,8 +147,9 @@ function UIMessage:removeMessage(choice_number)
       self.owner.message_callback = nil
     end
     if self.onClose then
-      self:onClose(false)
+      local onClose = self.onClose
```
TheCycoONE, line 150: Why this dance with the new temp variable?

TheCycoONE (approved):

CorsixTH/Lua/dialogs/message.lua
ARGAMX: > Why this dance with the new temp variable?

I agree, it somehow looks strange and out of context.
While working on this problem, I ran into an infinite loop where `removeMessage` called `onClose`, and `onClose` called `removeMessage`. 
And so on, looping, which resulted in a stack overflow.
I decided this shouldn't be possible, even with wrong decisions elsewhere in the code.
So I decided to break the loop there.
This change ensures that the callback will only be called once.
Perhaps by the end of the work on this issue, the root of the infinite loop had been removed, and now the infinite loop doesn't even start.
I haven't checked that.
But I decided to keep this change as useful.

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -604,22 +627,7 @@ end
 -- an actual message window; if no index is provided the first suitable message
 -- in the queue is popped.
 --!param index (integer|nil) the index in the message_queue to pop
-function UIBottomPanel:createMessageWindow(index)
-  local --[[persistable:bottom_panel_message_window_close]] function onClose(window)
```
lewri, line 576: Do you need to stub the persistable at all?

CorsixTH/Lua/dialogs/message.lua
lewri: I agree it looks a bit odd. Did the false passed in the old code no longer matter?

lewri (approved):

CorsixTH/Lua/dialogs/message.lua
ARGAMX: > I agree it looks a bit odd. Did the false passed in the old code no longer matter?

Yes, it no longer matter and nobody read it.
Moreover, there's actually an object expecting there.
I don't even understand how it worked before.
But I had to fix it.

CorsixTH/Lua/dialogs/bottom_panel.lua
ARGAMX: > Do you need to stub the persistable at all?

This code was simply moved to a different location. 
I suppose stab isn't needed there?

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -461,7 +461,9 @@ function UIBottomPanel:queueMessage(type, message, owner, timeout, default_choic
       -- If icon position found
       if this_icon_index ~= nil then
         -- Shift all icons that are on the right to the left
-        icon:setXLimit(1 + (i - 2) * FACTORY_ICON_WIDTH)
+        -- The '_factoryIconWitdh()' call is used for compatibility with older saves
+        -- where FACTORY_ICON_WIDTH was not yet declared when the closure was created.
+        icon:setXLimit(1 + (i - 2) * self._factoryIconWitdh())
```
ARGAMX: I was have to remove the explicit use of `FACTORY_ICON_WIDTH` from here (inside closure body).

The thing is, if you load an old save that already has messages in the drawer and open one of them, Lua crashes at this point:

```
../Resources/Lua/dialogs/bottom_panel.lua:464: ​​use of undeclared variable 'FACTORY_ICON_WIDTH'
```

I didn't expect this, but apparently the closure fetches mentioned external entities available at the time, which didn't include `FACTORY_ICON_WIDTH`.
And now, when executing the new closure body with the old scope it saved, it doesn't know what `FACTORY_ICON_WIDTH` is.
At least that's how I explain it to myself.

Here's a save where you can reproduce this. Open any message in the drawer, but not the last (right) one. The last one doesn't crash.

[SuperDuper_year_end_December_30.sav.zip](https://github.com/user-attachments/files/28009270/SuperDuper_year_end_December_30.sav.zip)

I could have set it back to `30` as it was before this PR, and it would have worked.
But then `30` would remain a magic constant.
I decided to wrap `FACTORY_ICON_WIDTH` in a function call `self._factoryIconWitdh()`.
And it worked.

By the way, thanks to Lewri for the idea to double-check this.

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -453,7 +454,25 @@ function UIBottomPanel:queueMessage(type, message, owner, timeout, default_choic
...
+
+  local --[[persistable:bottom_panel_message_window_close]] function onClose(this_icon)
+    local this_icon_index
+    for i, icon in ipairs(self.message_windows) do
+      -- If icon position found
+      if this_icon_index ~= nil then
+        -- Shift all icons that are on the right to the left
+        -- The '_factoryIconWitdh()' call is used for compatibility with older saves
+        -- where FACTORY_ICON_WIDTH was not yet declared when the closure was created.
+        icon:setXLimit(1 + (i - 2) * self._factoryIconWitdh())
```
lewri, line 451: Definitely not the most elegant, I agree.
The only other option is declaring the local again in the persistable, which seems just as slightly-less-than-elegant.

ARGAMX: > Definitely not the most elegant, I agree. The only other option is declaring the local again in the persistable, which seems just as slightly-less-than-elegant.

The problem I am afraid is that method with declaring the local again isn't really slightly-less-than-elegant, as it introduces duplication: you can change a variable in one place, but it will remain the same in another.
The function approach, while some ugly, solves the problem and avoid duplication, so the second copy isn't stored.

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -453,7 +454,25 @@ function UIBottomPanel:queueMessage(type, message, owner, timeout, default_choic
...
   end
-  local fax = {
+
+  local --[[persistable:bottom_panel_message_window_close]] function onClose(this_icon)
+    local this_icon_index
+    for i, icon in ipairs(self.message_windows) do
+      -- If icon position found
+      if this_icon_index ~= nil then
+        -- Shift all icons that are on the right to the left
+        -- The '_factoryIconWitdh()' call is used for compatibility with older saves
```
TheCycoONE: Say which save number, will help when we want to clean it up later to know if it's old enough.

ARGAMX: > Say which save number, will help when we want to clean it up later to know if it's old enough.

Added a version note 👌

CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -642,40 +610,15 @@ function UIBottomPanel:_showMessageIcon(index)
...
-  self.factory_direction = FACTORY_DIR_CLOSING
-  self.show_animation = true
-  self.factory_counter = -50                -- Delay close of message factory
-  table.remove(self.message_queue, index)   -- Delete element of the queue
+  -- Create the message window, note this does not show it to the player on creation.
+  local alert_window = UIMessage(self.ui, 175, 1 + #message_windows * 30,
+    onClose, message_info.type, message_info.message, message_info.owner, message_info.timeout, message_info.default_choice, message_info.callback)
+  message_windows[#message_windows + 1] = alert_window
+  self:addWindow(alert_window)
+  table.remove(self.message_queue, index)   -- Delete the last element of the queue
```
ARGAMX:
```suggestion
  local drawer_icon = message_info.drawer_icon
  drawer_icon:setXLimit(1 + #message_windows * FACTORY_ICON_WIDTH)
  message_windows[#message_windows + 1] = drawer_icon
  self:addWindow(drawer_icon)
  table.remove(self.message_queue, index)   -- Delete element of the queue
```

PR #3348, opened 2026-05-20
 Obey user initiated pauses

### LINKED ISSUE #2254: mustPause windows can unpause a user-initiated pause

lewri:
*Fixes #2254*

- Caches when a player pauses the game.
- Speed up now blocked during pause (this could break pauses)
- Redoccing/Refactors

### DIFF
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1083,9 +1083,9 @@ end
 -- Function for signaling that the window is added.
 -- To pause game if the necessary requirements are met.
 --!param window (object) window added
-function UI:_onAddWindow(window, previous_must_pause_state)
+function UI:_onAddWindow(window)
   if self.app.world and window:mustPause() then
-    self.app.world:mustPauseWindowAdd(previous_must_pause_state)
+    self.app.world:mustPauseWindowAdd()
   end
 end
 
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1057,8 +1057,9 @@ function UI:addWindow(window)
   if window.modal_class == "main" or window.modal_class == "fullscreen" then
     self.editing_allowed = false -- do not allow editing rooms if main windows (build, furnish, hire) are open
   end
+  local active_must_pauses = self:anyMustPauseWindowOpen()
   Window.addWindow(self, window)
-  self:_onAddWindow(window)
+  self:_onAddWindow(window, active_must_pauses)
 end
 
 --! Remove a window from the game
@@ -1083,9 +1084,9 @@ end
 -- Function for signaling that the window is added.
 -- To pause game if the necessary requirements are met.
 --!param window (object) window added
-function UI:_onAddWindow(window)
+function UI:_onAddWindow(window, previous_must_pause_state)
   if self.app.world and window:mustPause() then
-    self.app.world:mustPauseWindowAdd()
+    self.app.world:mustPauseWindowAdd(previous_must_pause_state)
   end
 end
 
@@ -1101,6 +1102,7 @@ end
 --! Function to check if we have any must pause windows open
 --!return (bool) Returns true if a must pause window is found
 function UI:anyMustPauseWindowOpen()
+  if not self.windows then return false end
   for _, window in pairs(self.windows) do
     if window:mustPause() then return true end
   end
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1057,9 +1057,8 @@ function UI:addWindow(window)
   if window.modal_class == "main" or window.modal_class == "fullscreen" then
     self.editing_allowed = false -- do not allow editing rooms if main windows (build, furnish, hire) are open
   end
-  local active_must_pauses = self:anyMustPauseWindowOpen()
   Window.addWindow(self, window)
-  self:_onAddWindow(window, active_must_pauses)
+  self:_onAddWindow(window)
 end
 
 --! Remove a window from the game
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1040,7 +1040,9 @@ function UI:onTick()
   return repaint
 end
 
-
+--! Adds a window to the game
+--! This also handles pause and permitted interaction behaviour as required
+--!param window (window) What is to be displayed
 function UI:addWindow(window)
   if window.closed then
     return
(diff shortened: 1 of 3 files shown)


CorsixTH/Lua/world.lua
```diff
@@ -762,7 +767,6 @@ function World:setSpeed(speed)
   if self:isCurrentSpeed(speed) then
     return
   end
-  tracy.Message("Changing speed to " .. speed)
```
lewri: I assume this tracy wasn't intended to be left here...

TheCycoONE: No it was intended, changing game speeds is a notable thing in a profile, it's good to have a marker.

TheCycoONE:
I wonder if the whole mustPause window logic isn't overly complicated 

It seems to me that we would just need to iterate through windows in onTick and if there were any that were mustPause we don't advance the hours/earthquake. That way we remove any clever tricks about remembering pause state.

CorsixTH/Lua/world.lua
lewri:
```suggestion
  tracy.Message("Changing speed to " .. speed)
```

lewri:
> I wonder if the whole mustPause window logic isn't overly complicated
> 
> It seems to me that we would just need to iterate through windows in onTick and if there were any that were mustPause we don't advance the hours/earthquake. That way we remove any clever tricks about remembering pause state.

That could be done, though I also assume we do want to always be mindful now of anything we add that run every tick.

On reflection, simply stopping the tick progression on mustPause windows probably isn't good enough, there are other factors that must trigger on pause/unpause, the game still has to be paused in that respect (blue tint, user actions).

TheCycoONE:
There is no reason to make the window set 'alreadyPaused', and it seems fragile/broken if it's possible for two must pause windows to open.

I think it's better to track paused because you set the game speed separately from paused due to must paused window, functions in world could hide the differences from the rest of the game; it makes the state tracking sane.

lewri:
> There is no reason to make the window set 'alreadyPaused', and it seems fragile/broken if it's possible for two must pause windows to open.

To note commit 9aa48c02d31e6a9966713c089557f7061c9f43a0 does address this.

lewri:
NB: This patch is still needed for obeying user initiated pauses but can't be updated yet

CorsixTH/Lua/world.lua
```diff
@@ -789,41 +795,59 @@ function World:setSpeed(new_speed)
...
+--!return (boolean) True if paused
 function World:isPaused()
   return self:isCurrentSpeed("Pause")
 end
 
+function World:setUserPaused(state)
+  self.user_paused = state
+end
+
+function World:isGameUserPaused()
```
lewri, line 853: Doc req'd **TODO**

lewri:
Code rebased and completely rewritten. We now always make note when a player pauses the game independent of must pause.

TheCycoONE:
Incidentally you don't need new state for this, all you need is for prev_speed to be set to Paused.

I wonder if this could be done without a new method even; if setSpeed is called and not isMustPauseActive() and the speed is not "Speed Up" then you can set prev_speed.

CorsixTH/Lua/ui.lua
```diff
@@ -1078,9 +1083,9 @@ end
 -- Function for signaling that the window is added.
 -- To pause game if the necessary requirements are met.
 --!param window (object) window added
-function UI:_onAddWindow(window)
+function UI:_onAddWindow(window, previous_must_pause_state)
```
lewri: Leftover to be removed

lewri:
> Incidentally you don't need new state for this, all you need is for prev_speed to be set to Paused.

If you set prev_speed to paused, how would you then know what the players previous speed was before that? You'd need a `prev_prev_speed` in that case, which sounds messy.

The other option I can look at, but the entire setSpeed function feels like a mess in itself, don't really want to keep adding loads of unique controls to something that already feels fragile.

TheCycoONE:
Fair, distegard.

CorsixTH/Lua/world.lua
```diff
@@ -783,58 +798,90 @@ function World:setSpeed(new_speed)
...
+--! the game
 function World:mustPauseWindowRemoved()
-  if self:isCurrentSpeed("Pause") then
-    self:pauseOrUnpause()
+  if self:isMustPauseActive() then
+    return
   end
+
+  if self:isCurrentSpeed("Pause") and not self:isGameUserPaused() then
+    self:setSpeed(self.prev_speed or "Normal")
```
lewri:
```suggestion
    self:setSpeed(self.prev_speed)
```

TheCycoONE:
I guess the original game didn't have a Pause menu option. I was thinking that if you select pause from the menu there's no reason for it to switch back to a previous speed without selecting something else explicitly, but for the p hotkey it makes sense.

I'm not sure it makes sense to wrap setSpeed in setUserSpeed, we could just as easily call togglePause explicitly and have the logic there?

lewri:
> I guess the original game didn't have a Pause menu option. I was thinking that if you select pause from the menu there's no reason for it to switch back to a previous speed without selecting something else explicitly, but for the p hotkey it makes sense.
> 
> I'm not sure it makes sense to wrap setSpeed in setUserSpeed, we could just as easily call togglePause explicitly and have the logic there?

User pause passthrough dropped. We now put all pauses from the user through pauseOrUnpause.

lewri:
> I guess the original game didn't have a Pause menu option. I was thinking that if you select pause from the menu there's no reason for it to switch back to a previous speed without selecting something else explicitly, but for the p hotkey it makes sense.
> 
> I'm not sure it makes sense to wrap setSpeed in setUserSpeed, we could just as easily call togglePause explicitly and have the logic there?

I've fallen into some more traps by actually removing the setUserSpeed stuff where state gets stuck, so I may have to put it back.

lewri:
> I'm not sure it makes sense to wrap setSpeed in setUserSpeed, we could just as easily call togglePause explicitly and have the logic there?

I managed to find that when I made this change it broke the patch. PR now reverted back to userpause because we need to update state if a player also presses a number/speed from ui and not just if they pressed P or not.

TheCycoONE (approved):

PR #3349, opened 2026-05-21
Prevent accidentally closed windows fax

### LINKED ISSUE #2638: [Bug] 'Go To Next Level' fax closes and disappears accidentally.
### Describe the issue

Screenshot 2024-09-26 at 22 31 06

If you received 'Go To Next Level' fax and pressed any of these keys:
`1 2 3 4 5 Q Z C F T G B P M`
then fax closes and disappears.

pressing these keys:
`6 7 8 9 0 W E A S D X V Y H N U J I K O L SPACE ENTER `
do not causes this problem.

At the same time all other type of faxes do not disappears on any key press.
They can be closed on Esc press, but not disappears.

In original game 'Go To Next Level' letter do not closes on any key press. Even on Esc press.

### Steps to Reproduce

2. Press 5 or hold Z to change speed to max.
3. 1 Oct you will receive 'Go To Next Level' fax which opens automatically.
4. If you was holding Z fax will closes and disappears instantly.
5. If you was not holding Z, press 1 or P to slowdown the time to calmly read fax.
6. Observe, that it closes and disappears without any selection.

### Expected Behaviour

The "Go to the next level" fax should not close until the user selects any of the options. As in original game.

### Save Game

[finish_level7_save.sav.zip](https://github.com/user-attachments/files/17155282/finish_level7_save.sav.zip)

### CorsixTH Version

0.68.0 Beta 1

### Operating System

MacOS Sonoma 14.5

ARGAMX:
*Fixes #2638*

- The "System Pause" / "Must Pause" logic has been improved. "Must Pause" now used instead "System Pause".
- A "Must Pause" is defined as a pause initiated by the application, not the user, based on specific events. For example, displaying a modal dialog or modal window that requires a response from the user.
- "Must Pause" can only be set and unset by the game. The user cannot unpause the game directly, only by responding to the modal window that initiated the pause.
- "Must Pause" controls the `world.user_actions_allowed` variable, which is determines the restrictions imposed on user actions. The logic behind `user_actions_allowed` effects has not changed. Only the reasons for setting and clearing `user_actions_allowed` have changed.
- The `world.system_pause` variable itself has been removed from the code. Instead, "Must Pause" is managed by the `ui.checkForMustPauseWindows()` function.
- The `mustPause` is set separately for each window, but this functionality was not modified here.
- The logic for set the blue filter on screen has also been improved.
- Logic behind setting/clearing the `user_actions_allowed` and `setBlueFilterActive` moved to one single place. (I hope all of it)
- This "System Pause" / "Must Pause" logic generally existed before this PR, but it had some issues with consistency. As a result, its purpose and how it was originally intended to work weren't entirely clear. In different tests, this functionality gave different and inconsistent results.
- Thus, as a result, the user will no longer be able to unpause or accidentally close a system/modal window by pressing Z, 1, 2, 3, P, or other keys that are not directly related to closing anything.
- The user will have to either make a decision in the modal window or explicitly close the modal window that is the cause.
- I also took inspiration from how it worked in the original game, where the simulation would pause while user working with a fax machine or win letter.

### DIFF
--- a/CorsixTH/Lua/dialogs/confirm_dialog.lua
+++ b/CorsixTH/Lua/dialogs/confirm_dialog.lua
@@ -79,17 +79,11 @@ function UIConfirmDialog:UIConfirmDialog(ui, must_pause, text, callback_ok, call
   self.height = last_y + 63
 
   self:registerKeyHandlers()
-  if self.must_pause then self:systemPause() end
 end
 
 -- Confirm dialogs are used for errors, if it is an error then pause the game
 function UIConfirmDialog:mustPause()
   return self.must_pause
-end
-
---! Function to tell the game a system pause is needed
-function UIConfirmDialog:systemPause()
-  TheApp.world:setSystemPause(true)
 end
 
 function UIConfirmDialog:registerKeyHandlers()
@@ -109,7 +103,6 @@ end
 --!param confirmed (boolean or nil) whether to call the confirm callback (true) or cancel callback (false/nil)
 function UIConfirmDialog:close(confirmed)
   -- NB: Window is closed before executing the callback in order to not save the confirmation dialog in a savegame
-  if self.must_pause then TheApp.world:setSystemPause(false) end -- Error dealt with
   Window.close(self)
   if confirmed then
     if self.callback_ok then
--- a/CorsixTH/Lua/dialogs/message.lua
+++ b/CorsixTH/Lua/dialogs/message.lua
@@ -105,13 +105,10 @@ end
 
 --! Displays the fax/strike message to the player when opened from the bottom_panel.
 function UIMessage:openMessage()
-  if TheApp.world:isUserActionProhibited() and not self.ui:checkForMustPauseWindows() then
+  if TheApp.world:isUserActionProhibited() then
     self.ui:playSound("wrong2.wav")
     self:adjustToggle()
     return
-  end
-  if TheApp.world:isCurrentSpeed("Speed Up") then
-    TheApp.world:previousSpeed()
   end
   if self.type == "strike" then -- strikes are special cases, as they are not faxes
     self.ui:addWindow(UIStaffRise(self.ui, self.owner, self.message))
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1052,13 +1052,10 @@ function UI:addWindow(window)
     end
     self.modal_windows[window.modal_class] = window
   end
-  if self.app.world and window:mustPause() then
-    self.app.world:setSpeed("Pause")
-    self.app.video:setBlueFilterActive(false) -- mustPause windows shouldn't cause tainting
-  end
   if window.modal_class == "main" or window.modal_class == "fullscreen" then
     self.editing_allowed = false -- do not allow editing rooms if main windows (build, furnish, hire) are open
   end
+  self:updateSystemPause(true, window)
   Window.addWindow(self, window)
 end
 
@@ -1068,18 +1065,26 @@ function UI:removeWindow(closing_window)
     if class and self.modal_windows[class] == closing_window then
       self.modal_windows[class] = nil
     end
-    if self.app.world and self.app.world:isCurrentSpeed("Pause") then
-      local pauseGame = self:checkForMustPauseWindows()
-      if not pauseGame and closing_window:mustPause() then
-        self.app.world:setSpeed(self.app.world.prev_speed)
-      end
-    end
     if closing_window.modal_class == "main" or closing_window.modal_class == "fullscreen" then
       self.editing_allowed = true -- allow editing rooms again when main window is closed
     end
+    self:updateSystemPause(false, closing_window)
     return true
   else
     return false
+  end
+end
+
+--! Toggles system pause on and off if needed
+--!param pause (bool) flag indicating whether to turn the pause state on or off
+--!param window (object) window causing the status update
+function UI:updateSystemPause(pause, window)
+  if not self.app.world then return end
+  if window:mustPause() then
+    local otherPauseWindows = self:checkForMustPauseWindows()
+    if not otherPauseWindows then
+      self.app.world:setSystemPause(pause)
+    end
   end
 end
 
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1056,7 +1056,7 @@ function UI:addWindow(window)
     self.editing_allowed = false -- do not allow editing rooms if main windows (build, furnish, hire) are open
   end
   Window.addWindow(self, window)
-  self:updateSystemPause(true, window)
+  self:_onAddWindow(window)
 end
 
 function UI:removeWindow(closing_window)
@@ -1068,30 +1068,34 @@ function UI:removeWindow(closing_window)
     if closing_window.modal_class == "main" or closing_window.modal_class == "fullscreen" then
       self.editing_allowed = true -- allow editing rooms again when main window is closed
     end
-    self:updateSystemPause(false, closing_window)
+    self:_onRemoveWindow(closing_window)
     return true
   else
     return false
   end
 end
 
---! Toggles system pause on and off if needed
---!param pause (bool) flag indicating whether to turn the pause state on or off
---!param window (object) window causing the status update
-function UI:updateSystemPause(pause, window)
-  if not self.app.world then return end
-  if window:mustPause() then
-    if pause then
-      self.app.world:systemPause()
-    else
-      self.app.world:systemUnpause()
-    end
+-- Function for signaling that the window is added.
+-- To pause game if the necessary requirements are met.
+--!param window (object) window added
+function UI:_onAddWindow(window)
+  if self.app.world and window:mustPause() then
+    self.app.world:mustPauseWindowAdd()
+  end
+end
+
+-- Function for signaling that the window is to be removed.
+-- To unpause game if the necessary requirements are met.
+--!param window (object) window to be removed
+function UI:_onRemoveWindow(window)
+  if self.app.world and window:mustPause() then
+    self.app.world:mustPauseWindowRemoved()
   end
 end
 
 --! Function to check if we have any must pause windows open
 --!return (bool) Returns true if a must pause window is found
-function UI:checkForMustPauseWindows()
+function UI:anyMustPauseWindowOpen()
   for _, window in pairs(self.windows) do
     if window:mustPause() then return true end
   end
--- a/CorsixTH/Lua/ui.lua
+++ b/CorsixTH/Lua/ui.lua
@@ -1055,8 +1055,8 @@ function UI:addWindow(window)
   if window.modal_class == "main" or window.modal_class == "fullscreen" then
     self.editing_allowed = false -- do not allow editing rooms if main windows (build, furnish, hire) are open
   end
-  self:updateSystemPause(true, window)
   Window.addWindow(self, window)
+  self:updateSystemPause(true, window)
 end
 
 function UI:removeWindow(closing_window)
@@ -1081,9 +1081,10 @@ end
 function UI:updateSystemPause(pause, window)
   if not self.app.world then return end
   if window:mustPause() then
-    local otherPauseWindows = self:checkForMustPauseWindows()
-    if not otherPauseWindows then
-      self.app.world:setSystemPause(pause)
+    if pause then
+      self.app.world:systemPause()
+    else
+      self.app.world:systemUnpause()
     end
   end
 end
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -757,7 +757,7 @@ end
 -- "Max speed", or "And then some more".
 function World:setSpeed(new_speed)
   if self:isCurrentSpeed(new_speed) then return end
-  if self:mustPause() and new_speed ~= "Pause" then return end -- System pause takes precedence
+  if self:mustPause() and new_speed ~= "Pause" then return end -- "Must pause" takes precedence
   tracy.Message("Changing speed to " .. new_speed)
 
   if new_speed == "Pause" then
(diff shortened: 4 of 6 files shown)


CorsixTH/Lua/dialogs/confirm_dialog.lua
```diff
@@ -79,19 +79,13 @@ function UIConfirmDialog:UIConfirmDialog(ui, must_pause, text, callback_ok, call
...
 -- Confirm dialogs are used for errors, if it is an error then pause the game
 function UIConfirmDialog:mustPause()
   return self.must_pause
 end
 
---! Function to tell the game a system pause is needed
-function UIConfirmDialog:systemPause()
-  TheApp.world:setSystemPause(true)
-end
-
```
ARGAMX, line 94: `UIConfirmDialog` is `Window` inheritor. 
So lets keep all "System Pause" management logic in Window.

CorsixTH/Lua/ui.lua
```diff
@@ -1052,14 +1052,11 @@ function UI:addWindow(window)
     end
     self.modal_windows[window.modal_class] = window
   end
-  if self.app.world and window:mustPause() then
-    self.app.world:setSpeed("Pause")
-    self.app.video:setBlueFilterActive(false) -- mustPause windows shouldn't cause tainting
-  end
```
ARGAMX, line 1058: Window in `addWindow` and `removeWindow` was directly manage BlueFilter and Pause state. Not smell good.
Let's just report here that a window has opened/closed.

ARGAMX:
Bump for review.

CorsixTH/Lua/ui.lua
```diff
@@ -1068,21 +1065,30 @@ function UI:removeWindow(closing_window)
...
     return true
   else
     return false
   end
 end
 
+--! Toggles system pause on and off if needed
+--!param pause (bool) flag indicating whether to turn the pause state on or off
+--!param window (object) window causing the status update
+function UI:updateSystemPause(pause, window)
```
lewri: Move to world as play state should be there. Outside of games, we don't have any concept of game speed.
Or App, as it's program related.

It should no longer belong in UI.

CorsixTH/Lua/world.lua
```diff
@@ -756,29 +752,34 @@ function World:getCurrentSpeed()
...
+-- One of: "Pause", "Slowest", "Slower", "Normal",
 -- "Max speed", or "And then some more".
-function World:setSpeed(speed)
-  if self:isCurrentSpeed(speed) then
-    return
-  end
-  tracy.Message("Changing speed to " .. speed)
-  if speed == "Pause" or self.system_pause then
+function World:setSpeed(new_speed)
+  local isSystemPause = self:isSystemPauseActive()
```
lewri: Snake case needed

CorsixTH/Lua/world.lua
```diff
@@ -756,29 +752,34 @@ function World:getCurrentSpeed()
...
-  elseif self:getCurrentSpeed() == "Pause" then
-    self.user_actions_allowed = true
   end
 
+  -- Actions are not allowed when the game is System paused.
+  self.user_actions_allowed = not isSystemPause and
+    (new_speed ~= "Pause" or TheApp.config.allow_user_actions_while_paused)
+
+  -- Set the blue filter according to whether the user can build or not.
+  TheApp.video:setBlueFilterActive(not isSystemPause and not self.user_actions_allowed)
```
lewri: While I did express support generally for this. I would like some agreement elsewhere this is ok and I haven't neglected anything. It will cause blue to appear on players who had actions while paused enabled, so I'd look for some opinion to shore it up.

lewri (dismissed):
I'm half convinced this change now just makes `must_pause == system_pause`. I'd pick one name or the other instead of confusing two differently named variables for the same purpose.

If you want to encompass all pause reasons, then `must_pause` is the one to use. System pause was originally used to imply something was wrong with the game (or a potentially destructive action e.g. restarting a level, was required), rather than stopping as a consequence of normal gameplay.

CorsixTH/Lua/ui.lua
TheCycoONE: This could be a private function (add _ prefix) to UI; it manages the intersection of window and world which is within the scope of UI.

CorsixTH/Lua/world.lua
TheCycoONE: Resolved? user_action_allowed is part of the condition.

CorsixTH/Lua/world.lua
```diff
@@ -756,29 +752,34 @@ function World:getCurrentSpeed()
...
-  end
-  tracy.Message("Changing speed to " .. speed)
-  if speed == "Pause" or self.system_pause then
+function World:setSpeed(new_speed)
+  local isSystemPause = self:isSystemPauseActive()
+  if self:isCurrentSpeed(new_speed) then return end
+  if isSystemPause and new_speed ~= "Pause" then return end -- System pause takes precedence
+  tracy.Message("Changing speed to " .. new_speed)
+
+  if isSystemPause or new_speed == "Pause" then
```
TheCycoONE: isSystemPause is redundant here, new_speed must be Pause when isSystemPause is true per 761.

TheCycoONE:
Good work. I prefer this solution over #3348 once it is cleaned up.

CorsixTH/Lua/world.lua
lewri: No, this will still turn the screen blue as it needs both conditions to be met to be not blue

TheCycoONE: Z = ~A and ~B
```
A|B|Z
1  1  0
1  0  0
0  1  0
0  0  1
```

A = isSystemPause
B = self.user_actions_allowed

Blue only happens if not a system pause and user actions are not allowed. What am I missing?

lewri: Yes, I can't read.

CorsixTH/Lua/ui.lua
ARGAMX: This has been refactored.
Now here are local function calls (_ prefix).
Their job is to notify the necessary objects (world) that a window has closed or opened.
The function names have been abstracted from the "system pause".

Moreover, the concept of "system pause" has been renamed to "mustPause". Now it's just "mustPause" everywhere, just as Lewri asked.

So the UI and the World are separated here as they should be.

CorsixTH/Lua/world.lua
ARGAMX: Resolved 👌 (removed in the end)

CorsixTH/Lua/world.lua
ARGAMX: > isSystemPause is redundant here, new_speed must be Pause when isSystemPause is true per 761.

Applied 👌

ARGAMX:
> I'm half convinced this change now just makes `must_pause == system_pause`. I'd pick one name or the other instead of confusing two differently named variables for the same purpose.
> 
> If you want to encompass all pause reasons, then `must_pause` is the one to use. System pause was originally used to imply something was wrong with the game (or a potentially destructive action e.g. restarting a level, was required), rather than stopping as a consequence of normal gameplay.

Ok, one name picked. `mustPause`. 
"system pause" naming concept removed. 👌

CorsixTH/Lua/dialogs/message.lua
```diff
@@ -105,14 +105,11 @@ end
...
 function UIMessage:openMessage()
-  if TheApp.world:isUserActionProhibited() and not self.ui:checkForMustPauseWindows() then
+  if TheApp.world:isUserActionProhibited() and not TheApp.world:mustPause() then
     self.ui:playSound("wrong2.wav")
     self:adjustToggle()
     return
   end
-  if TheApp.world:isCurrentSpeed("Speed Up") then
-    TheApp.world:previousSpeed()
-  end
```
ARGAMX, line 115: We should no longer directly manage pause state in arbitrary places in the system. This should be more centralised. This procedure is now initiated by windows open/close.

CorsixTH/Lua/world.lua
lewri, line 2815: We'll need a new afterload to have `self.system_pause = nil` to clean it from world.
I don't think it needs a savegame bump though (arguably it should but it seems a waste of a bump)

CorsixTH/Lua/world.lua
```diff
@@ -756,29 +752,32 @@ function World:getCurrentSpeed()
...
 -- "Max speed", or "And then some more".
-function World:setSpeed(speed)
-  if self:isCurrentSpeed(speed) then
-    return
-  end
-  tracy.Message("Changing speed to " .. speed)
-  if speed == "Pause" or self.system_pause then
+function World:setSpeed(new_speed)
+  if self:isCurrentSpeed(new_speed) then return end
+  if self:mustPause() and new_speed ~= "Pause" then return end -- System pause takes precedence
```
lewri: Inline comment needs word changing

CorsixTH/Lua/world.lua
```diff
@@ -790,41 +789,48 @@ function World:setSpeed(speed)
...
+
+function World:mustPauseWindowRemoved()
+  if self:isCurrentSpeed("Pause") then
+    self:pauseOrUnpause()
+  end
+end
+
+--! Check if we have any must pause windows open
+--!return (bool) returns true if we have any must pause windows open
+function World:mustPause()
```
lewri, line 804: Function name arguably doesn't match doc.

lewri: In fact, you're verbatim used the same doc as the UI call. In that case, I'd use the same function name as UI; which is how we usually do it.

lewri (commented):
Much better to me

CorsixTH/Lua/ui.lua
```diff
@@ -1068,24 +1065,37 @@ function UI:removeWindow(closing_window)
...
+function UI:_onRemoveWindow(window)
+  if self.app.world and window:mustPause() then
+    self.app.world:mustPauseWindowRemoved()
+  end
+end
+
 --! Function to check if we have any must pause windows open
 --!return (bool) Returns true if a must pause window is found
-function UI:checkForMustPauseWindows()
+function UI:anyMustPauseWindowOpen()
```
lewri, line 1098: My name was that bad eh? 🤣 (this isn't an objection)

CorsixTH/Lua/world.lua
ARGAMX: Okay, I came up with a better description that better suits this particular class in relation to "Must Pause" functionality.

CorsixTH/Lua/world.lua
ARGAMX: > Inline comment needs word changing

Resolved 👌

CorsixTH/Lua/world.lua
ARGAMX: > We'll need a new afterload to have `self.system_pause = nil` to clean it from world. I don't think it needs a savegame bump though (arguably it should but it seems a waste of a bump)

Added `self.system_pause = nil` in afterload. 👌

CorsixTH/Lua/world.lua
```diff
@@ -790,41 +789,48 @@ function World:setSpeed(speed)
...
+  return self.ui:anyMustPauseWindowOpen()
+end
+
 function World:isPaused()
   return self:isCurrentSpeed("Pause")
 end
 
+--! Set the blue filter according to whether the user can build or not.
+function World:updateScreenBlueFilter()
+  TheApp.video:setBlueFilterActive(not self:mustPause() and not self.user_actions_allowed)
```
TheCycoONE: self.user_actions_allowed is always false when self:mustPause() now right? per 768.

If that's not the case, put the cheap condition first.

TheCycoONE (approved):

CorsixTH/Lua/world.lua
```diff
@@ -2815,6 +2790,9 @@ function World:afterLoad(old, new)
       end
     end
   end
+  if old < 244 then
```
TheCycoONE: I disagree with using an old number without bumping, if another PR merges first it won't conflict and may never happen. (And I know that contradicts what @lewri said)

TheCycoONE:
I really like this PR; I think it's a prefect example of leaving the code cleaner than you found it, while also solving a defect. Nice work.

CorsixTH/Lua/world.lua
lewri: Oh, yeah don't use an existing number. Either a new one or no number at all.

ARGAMX: > Oh, yeah don't use an existing number. Either a new one or no number at all.

The thing is If we just use a next number without bumping, then if we don't bump `savegame_version` in any other PR after this one that we have to merge before the release, then the condition will always trigger and will always clear the variable. Until bump happen sometime in future.
We also can't reset `system_pause` without a number, because then if lightning strikes the same place a second time and we start using `system_pause` again, then this reset will interfere and create problems.
Overall, I consider all these problems so minor and unlikely that I'm too lazy to even consider and discuss them, so I fully understand you. Although hypothetically these things are possible.
But since Cyco brought it up, and if we're going to be meticulous here, then we have to bump.

CorsixTH/Lua/world.lua
ARGAMX: > self.user_actions_allowed is always false when self:mustPause() now right? per 768.
> 
> If that's not the case, put the cheap condition first.

Okay, I've changed the order to execute cheap first.
I also made some changes to ensure `self.user_actions_allowed` is up to date before calling `updateScreenBlueFilter`.

CorsixTH/Lua/world.lua
lewri: Strictly speaking it's a characteristic of the window rather than a mode (asking UI about said characteristics). 
It's probably return true if mustPause state is in effect.

But we're into nits here.

ARGAMX: > Strictly speaking it's a characteristic of the window rather than a mode (asking UI about said characteristics). It's probably return true if mustPause state is in effect.
> 
> But we're into nits here.

Yes, but you know this because you know what's inside `ui.lua`.
The `World` doesn't iterate through windows itself, so it certainly can't be sure about this. 
And it shouldn't.
So for him it's just some kind of mode or state.
He doesn't know what this window characteristic.
In this light, even the external name `anyMustPauseWindowOpen` violate the abstraction.
In this context, I think the description is quite fair.
It doesn't tell you more than it needs to and doesn't bog you down in details that don't pertain to `World`.

PR #3355, opened 2026-05-25
Reinsert discarded music call for Windows

lewri:
- Reinserts the discarded `playRandomBackgroundTrack` call during audio init to have Windows prepare the async music playback in the background in the movie.

As I'm not sure if cyco still wants to try to async call a midi silence instead this simply just reverts behaviour to how it was before #3286 (where the movie masks the setup call -- demo or movies off still always had this issue).

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -287,7 +287,7 @@ function App:init()
   self.audio:init()
 
   -- Hack to early initialise the audio and reduce delay on Windows
-  self:playRandomBackgroundTrack()
+  self.audio:playRandomBackgroundTrack()
   self.audio:stopBackgroundTrack()
 
   -- Load movie player
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -286,6 +286,10 @@ function App:init()
   self:initMusicDir()
   self.audio:init()
 
+  -- Hack to early initialise the audio and reduce delay on Windows
+  self:playRandomBackgroundTrack()
+  self.audio:stopBackgroundTrack()
+
   -- Load movie player
   corsixth.require("movie_player")
   self.moviePlayer = MoviePlayer(self, self.audio, self.video)
--- a/CorsixTH/Lua/audio.lua
+++ b/CorsixTH/Lua/audio.lua
@@ -193,6 +193,8 @@ function Audio:init()
     self.background_playlist = {}
     return
   end
+
+  -- Hack to early initialise the audio and reduce delay on Windows
   self:playRandomBackgroundTrack()
 end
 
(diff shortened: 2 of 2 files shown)


TheCycoONE:
Did you check if this caused issues for midi / rtmidi?

lewri:
> Did you check if this caused issues for midi / rtmidi?

For virtualmidisynth it does :(
Physical midi devices are too fast as they're not held in the async call; so UI isn't ready at the time they can play.
In theory, software audio playback could also have this problem if the async happened immediately.

Instead it should be inserted just before the intro movie plays, but it would need silence (midi devices manage to start playing before the movie can start)

```yml
An error has occurred in CorsixTH:
Lua\audio.lua:819: attempt to index a nil value (field 'ui')
stack traceback:
        Lua\audio.lua:819: in method 'notifyJukebox'
        Lua\audio.lua:713: in method 'playBackgroundTrack'
        Lua\audio.lua:565: in method 'playRandomBackgroundTrack'
        Lua\audio.lua:189: in method 'init'
        Lua\app.lua:287: in method 'init'
        CorsixTH.lua:108: in main chunk
        [C]: in ?
bootstrap.cpp:82: attempt to call a nil value (field 'mainloop')
An error has occurred!
Running: The active handler.
Lua\app.lua:1370: attempt to index a nil value (field 'ui')
stack traceback:
        Lua\app.lua:1370: in function <Lua\app.lua:1369>
        (...tail calls...)
```

TheCycoONE:
I thought so... we'll need a different fix.

lewri:
> I thought so... we'll need a different fix.

Would you be able to assist? I could guard against midi player detection and not run it if so, but these are all hacks around trying a better solution.

TheCycoONE:
Sure, still hacky, but try this.

lewri:
> Sure, still hacky, but try this.

I think this will work  fine, but it will play music briefly so might need a guard for midi devices. Will test later

TheCycoONE (approved):

lewri:
Confirmed music now plays before movie. We'll have to guard it.

lewri:
Additional commit should fix the random music now before movie inits. Please check ok @TheCycoONE

TheCycoONE:
We might put the play and the stop next to each other - I think that would be clearer as they're involved in the same hack and they run back to back anyway. I don't mind if it's in the init or in app.

lewri:
> We might put the play and the stop next to each other - I think that would be clearer as they're involved in the same hack and they run back to back anyway. I don't mind if it's in the init or in app.

Moved into app. You don't want it stopping during an audio init outside of launch (i.e. turning music back on).

TheCycoONE (approved):

CorsixTH/Lua/app.lua
```diff
@@ -286,6 +286,10 @@ function App:init()
   self:initMusicDir()
   self.audio:init()
 
+  -- Hack to early initialise the audio and reduce delay on Windows
+  self:playRandomBackgroundTrack()
```
tobylane:
```suggestion
  self.audio:playRandomBackgroundTrack()
```

tobylane (commented):
If play_music is false but audio_music is set to a folder on a sleeping external hard drive, then nothing launches until the hard drive is awake. This is a rare case, but confusing.

PR #3359, opened 2026-05-31
Move the Machine Menu button to the right of the Research button

ARGAMX:
- Move the Machine Menu button to the right of the Research button

Before:
Screenshot 2026-05-31 at 10 13 42

After:
Screenshot 2026-05-30 at 14 42 15

Justification for the change:

The downside with "Machine Menu" button as second position is that when you launch the game in 640x480, the "Machine Menu" button disappears, and "Town Map" is now the second button again. 

If you play the original game occasionally, "Town Map" is also the second button. 

When you're heating a hospital and positioning heaters, you have to constantly press the Town Map button to make sure you haven't overheated anything and are covering the maximum area.
Open the map and see if any heating adjustments need to be made. Then open the map again, and repeat the process.
So has to be pressed many times trying to optimally place the radiators.
If you do this with a mouse, then you already start clicking Town Map by muscle memory.

As a result, you're torturing your muscle memory with constant, contradictory trainings with "Town Map" button. 
When it is sometimes in second position, and sometimes in third.

And if you've played the game before, you might already be used to Town Map being in second place.

This way I always press the Machine Menu button when I want to open the map.

This problem is further exacerbated by the fact that until your cursor moves to this block, the display info appears instead, and you can't see the buttons at all.

So you press buttons from memory if you don't make pause see buttons beforehand.

Screenshot 2026-05-31 at 10 42 30

.
I've noticed that the buttons in the center are used less often and are less susceptible to muscle memory because they are too far from the edge to be oriented relative to the edge.

Therefore, a place on the right, for example after the research button but before the progress button, seems more appropriate.

Although initially placing the Machine Menu button next to the Staff button was quite sensible and consistent, and at the time it seemed like the right decision.

But after playing around with it, I came to what It's better for the new Machine button not to take up the space of those buttons that are located on the edge and are subject to good muscle memory training.

### DIFF
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -108,44 +108,25 @@ function UIBottomPanel:drawPanels()
   -- Buttons that are shown instead of the dynamic info bar when hovering over it.
   local panels = {}
   local buttons = {}
-  local app = self.ui.app
   panels[1]  = self:addPanel(15, 364, 0) -- Staff management button
   buttons[1] = panels[1]:makeToggleButton(6, 6, 35, 36, 16, self.dialogStaffManagement):setTooltip(_S.tooltip.toolbar.staff_list)
-  if self:machineMenuButtonExists() then
-    -- Sprites for machine menu button doesn't exist in original game. Let's import them from aux_ui.dat and draw
-    local aux_sprites = app.gfx:loadSpriteTable("Bitmap", "aux_ui", true)
-    self:addPanel(0, 407, 0):makeToggleButton(2, 6, 36, 36, 0, self.dialogMachineMenu)
-      :setSound() -- override
-      :setTooltip(_S.tooltip.toolbar.machine_menu)
-      .panel_for_sprite.custom_draw = --[[persistable:machine_menu_buttons]] function(panel, canvas, x, y)
-      local s = TheApp.config.ui_scale
-      x = x + panel.x * s
-      y = y + panel.y * s
-      panel.window.panel_sprites:draw(canvas, panel.sprite_index, x, y, { scaleFactor = s })
-      local btn = panel.window.active_button
-      if panels[1].visible then
-        local w = self.ui:getWindow(UIMachineMenu)
-        if w or btn and btn.panel_for_sprite == panel and btn.active then
-          aux_sprites:draw(canvas, 23, x, y, { scaleFactor = s })
-        else
-          aux_sprites:draw(canvas, 22, x, y, { scaleFactor = s })
-        end
-      end
-    end
-  end
-  panels[2]  = self:addPanel(17, 407 + self.offset, 0) -- Town map button
+  panels[2]  = self:addPanel(17, 407, 0) -- Town map button
   buttons[2] = panels[2]:makeToggleButton(1, 6, 35, 36, 18, self.dialogTownMap):setTooltip(_S.tooltip.toolbar.town_map)
-  panels[3]  = self:addPanel(19, 445 + self.offset, 0) -- Casebook button
+  panels[3]  = self:addPanel(19, 445, 0) -- Casebook button
   buttons[3] = panels[3]:makeToggleButton(1, 6, 35, 36, 20, self.dialogDrugCasebook):setTooltip(_S.tooltip.toolbar.casebook)
-  panels[4]  = self:addPanel(21, 483 + self.offset, 0) -- Research button
+  panels[4]  = self:addPanel(21, 483, 0) -- Research button
   buttons[4] = panels[4]:makeToggleButton(1, 6, 35, 36, 22, self.dialogResearch)
     :setSound():setTooltip(_S.tooltip.toolbar.research) -- Remove default sound for this button
+  if self:machineMenuButtonExists() then
+    self:_addMachineButton(panels, 521) -- Machine menu button
+  end
   panels[5]  = self:addPanel(23, 521 + self.offset, 0) -- Status button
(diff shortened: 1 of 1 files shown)


MandrakeHorse:
Although I agree that we should move the Machines Menu button from its current position, placing it in between Research and Status will still disrupt the original order of the buttons. Considering that this is a brand new button that's exclusive to CorsixTH and because it's bound to F10 by default (see our [Game Guide](https://github.com/CorsixTH/CorsixTH/wiki/Game-Guide#machines-menu)), perhaps we should place it right at the very end next to the Policy button, so that it matches the order of the hotkeys?

ARGAMX:
> Although I agree that we should move the Machines Menu button from its current position, placing it in between Research and Status will still disrupt the original order of the buttons. Considering that this is a brand new button that's exclusive to CorsixTH and because it's bound to F10 by default (see our [Game Guide](https://github.com/CorsixTH/CorsixTH/wiki/Game-Guide#machines-menu)), perhaps we should place it right at the very end next to the Policy button, so that it matches the order of the hotkeys?

Putting Machine Menu button in the last position at first glance also seems to make some sense.

But the hotkeys can be reassigned.
And It's unclear what the benefits of arranging the buttons on the panel in the same order as the default F button numbers would be. I'm not sure this approach will solve any problems. At least none of the problems I see.

The thing is that for muscle memory this order works in both directions, from left to right and from right to left.

Likewise, I always click by blind the last button, expecting it is Hospital Policy button.

Screenshot 2026-05-31 at 10 42 30_2

.
Therefore, Machine Menu button in the last position, the issue is the same as described.

But with the Machine Menu button in the middle, this problem less significant, since they are difficult to press blindly.

And placing this button in any position will disrupt the order of the original buttons, and this is simply impossible to avoid.

If we consider the problem in terms of the order of hotkeys and actually using them, then it doesn't matter where the panel button is. The problem is only relevant if player don't use hotkeys.

lewri:
I think wherever you put it, you're going to have muscle memory problems as you describe.

Saying that, I do acknowledge mandrake's point (Mandrake is also for some part a SME on these kinds of things). Putting it at the far end of existing buttons is usually the safest method to minimise that behaviour. Shifting multiple icons along could also cause a muscle memory issue with multiple buttons (vs one button by putting it at the end).

I would suggest finding a consensus on its location, then implementing that. If most people want it in the middle, sure. Left where it is, sure. On the right, etc. etc.

MandrakeHorse:
> I would suggest finding a consensus on its location, then implementing that. If most people want it in the middle, sure. Left where it is, sure. On the right, etc. etc.

Perhaps we should run a poll asking our community where they would like us to put this button, with options for all eight possible positions? Discord might be the best place to ask since that has a pretty active community, although we will welcome feedback wherever it gets posted (including this thread, of course!)

TheCycoONE:
Yep lets poll. The second position makes sense because it's similar to the staff dialog. The last position makes sense because last and hotkeys as mentioned, the middle makes less sense to me, but per @ARGAMX it may interfere least with memorized positions.

The first position might also be a reasonable option no one has proposed.

TheCycoONE:
https://github.com/CorsixTH/CorsixTH/discussions/3369

ARGAMX:
> #3369

The voting in the form in which it is done seems senseless to me.
It ignored the main problem.
Which I described here in topic #3359.

If the vote had been held a few months ago, I would have voted for the second option too.

Because at that time I haven't built any new hospitals since this button was added.

But when I tried building a new hospital during testing #3343, I encountered what #3359 described.

The vote didn't say anything about this. It simply offers a choice without context. But that's pointless. With input like that, I would have voted for the second option too.

Therefore, it would probably be important to consider how many hours or how many new hospitals the voter built. And after that, whether they were able to grasp the context of the problem or not.

If user haven't played with this new button much... either he hasn't played before at all or he played a very long time ago and he has no muscle memory left. Well, then user probably don't understand the problem. Just like I didn't a few months ago.

At least, I see that seems no one understood it yet.

To understand the problem for more people, it might make sense to leave the button in the same location and get feedback from 0.70.0 players a year from now. But then, some players who haven't played before and don't have the same habits would start to get used to the second position, and then most would likely object to the move, having already gotten used to the new one.

So it seemed like the only option here was to move it in authoritarian manner.
Players without muscle memory wouldn't really care where new button appear.
But for players with muscle memory, this is the solution to stop taping on wrong buttons on new places.
But quorum essentially brings the idea to a dead end.

It's probably easier to close this PR.
I think there is no any opportunity here. It's easier to move the button in my local game copy.

Due to being busy testing 0.70.0 beta 1, I simply didn't have opportunity to write this answer.

MandrakeHorse:
Just to clarify that the purpose of this poll is to gather feedback on what the wider community thinks and that it won't be a binding vote - the option that we decide to go with might not even receive the most votes, but knowing what options are preferred can guide us towards a decision

TheCycoONE:
The voting was 2 + @MandrakeHorse for the far right, 2 for the current position, and @ARGAMX alone for the 5th.

I don't think it's right to dismiss other people's opinions; particularly when we petitioned beta testers and developers who have had a chance to play with it.

There was also a suggestion to make the position configurable, and my follow up that maybe it would be better in that case if all button positions were configurable.

I understand the justification of putting it in a position of a button that is rarely activated to minimize the impact on players habituated to the existing layout; but in its current form we've failed to reach consensus to merge the PR as is. That may be motivation to follow through with the configurable layout task.

TheCycoONE (changes requested):
Blocking moving the button, the refactor looks good if you want to submit on it's own or with a configurable panel layout.

PR #3363, opened 2026-06-01
Bottom panel fixes

lewri:
**0.70.0 beta 1**

- Restores sounds to machine menu and research policy buttons when paused when user actions aren't allowed
- Prevents machine menu being being opened from said button in same scenario

### DIFF
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -819,6 +819,8 @@ end
 function UIBottomPanel:dialogResearch(enable)
   if not self.world.user_actions_allowed then
     self:updateButtonStates()
+    self.ui:playSound(self.ui.hospital.research_dep_built and
+        "selectx.wav" or "wrong2.wav")
     return
   end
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/bottom_panel.lua
```diff
@@ -819,6 +819,8 @@ end
 function UIBottomPanel:dialogResearch(enable)
   if not self.world.user_actions_allowed then
     self:updateButtonStates()
+    self.ui:playSound(self.ui.hospital.research_dep_built and
+        "selectx.wav" or "wrong2.wav")
```
TheCycoONE, line 823: What's the logic here, should a sound play for using_demo_files too?

lewri: In the code below it looks like the sound plays with an error message. However, in pause that message won't show if the user actions aren't allowed. Probably best not to try and throw up a window behind a blue filter.

ARGAMX (approved):
Works nice for me.
Code also fine.
Thanks 👍

PR #3364, opened 2026-06-01
Set builtin/unicode for soundfont path

lewri:
**0.70.0 beta 1**

- Use unicode font, then builtin font if not available to show soundfont path if set
- If not set, the text stays in the normal game font/language font
- Cleans up unnecessary `a and a or b`  to just `a or b`

image

Future todo: add reset button


TheCycoONE (approved):

ARGAMX (approved):
Works nice for me.
Thanks 👍

PR #3366, opened 2026-06-03
Reset pickedup on placeObject

ARGAMX:
Screenshot 2026-06-02 at 15 30 34

.
**0.70.0 beta 1**
.
- Fixes bug when `object.picked_up` stay `true` after object placed.
- Always set `object.picked_up = false` on `objectPlaced`.

Issue introduced by #3304

**How to reproduce:**

2. Start new level among Single Scenario.
3. Build Training Room. Place one chair in room. Finish build that room.
4. Enter room edit mode and edit that Training Room.
5. Pickup chair and place it in the same place.
6. Finish edit that room.
7. Hire any student and place they in that room.
8. Observe as the advisor says there are no free chairs in the room and the student leaves the room without starting the study.


lewri (approved):
Afterload shouldn't be needed if we add instructions to reset state to Beta 2 rel notes.

PR #3367, opened 2026-06-03
Disable sell hotkey in UIEditRoom

lewri:
**0.70.0 beta 1**

Pressing Delete when editing a room causes breakage by returning edit room to doors phase

- With #3304 we accidentally added a route to UIEditRoom where Delete (sell hotkey) caused the room to go backward to doors phase.
- Hotkey now has no effect when editing a room. As discussed on discord, a nice future todo is having the hotkey sell any non-core room items instead in objects phase.


ARGAMX (approved):
Ah, now I get it. 
The new "sell" hotkey was producing unexpected results in the room. 
Change make sense. 👍

PR #3370, opened 2026-06-04
Avoid duplicates in `world.entities`

### LINKED ISSUE #3376: [Bug] Objects corrupts save game
### Describe the issue

There is a problem when we place some objects (like radiators) in hospital. When we make a room, the place where was the object is unavailable. And when we erase the room, and save, we can't load again. (I guess it is that)

Image

### Steps to Reproduce

2. 

3.

### Expected Behaviour

The game needs to loaded correctly

- [ ] Right mouse panning

### Save Game

[save loikos.zip](https://github.com/user-attachments/files/28669090/save.loikos.zip)
09-01 : before corrupted
10-01 : corrupted

### CorsixTH Version

0.70 beta 1

### Operating System

win 11

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml

```

### Additional Information

_No response_

ARGAMX:
Screenshot 2026-06-02 at 15 08 12

.
**0.70.0 beta 1**
Dirty hack fix for a bug that prevented the game from loading.

Fixes #3376

- Avoid adding duplicates to `world.entities`.
- Before adding entity to , check that this object is not already there.

Issue introduced by #3304

**How to reproduce:**

1. Start new level among Single Scenario.
2. Buy and place one Radiator.
3. Pick it up from the floor and then cancel the grab by pressing `Esc`.
4. Pick it up from the floor and press the red `X` in menu, thereby selling the item.
5. Save the game.
6. Load the save you just made.
7. Observe error: 
`Error while loading game: ../Resources/Lua/entities/object.lua:827: bad argument #1 to 'eraseObjectTypes' (number expected, got nil)`

**Why is this happening?**

Every time player place an object, `World:objectPlaced(entity)` is called.
`World:objectPlaced(entity)` executes `self.entities[#self.entities + 1] = entity`.

Because before 3304 objects couldn't exist before they were placed, so it was make sense to add the object to the `World.entities` array when `objectPlaced` was called.

But now, with 3304, an object can already exist before `objectPlaced` and already be in `World.entities`. 
This is because `objectPlaced` was previously called for that object when it was placed previous time.

And now, when we move object from place to place the same object is added to `World.entities` again and again.
Because `World:objectPlaced(entity)` doesn't check whether placed object already exists in `World.entities` or not.

Therefore, when we delete an object from the world that has been added to `World.entities` multiple times, only one of its entries is deleted from `World.entities`. The others entries remain, since it was added there multiple times.

Also, since we're deleting object from the world, its coordinates are reset to `nil`.

As a result, we end up with references to this object in `World.entities` that have `nil` coordinates, since it was set to `nil` for the object.

This crashes the game in `object.lua` line 827.

Perhaps a more bold solution here would be to remove the object from `World.entities` when we pick it up object from the floor.

But due to concerns that this could lead to a new issues, I settled on a more defending solution.

### DIFF
--- a/CorsixTH/Lua/utility.lua
+++ b/CorsixTH/Lua/utility.lua
@@ -423,8 +423,8 @@ function table_merge(...)
 end
 
 --! Checks if the table contains item
-function is_contains(table, value)
-  for index, val in ipairs(table) do
+function table_contains(table, value)
+  for _, val in pairs(table) do
     if val == value then
       return true
     end
--- a/CorsixTH/Lua/utility.lua
+++ b/CorsixTH/Lua/utility.lua
@@ -422,6 +422,16 @@ function table_merge(...)
   return result
 end
 
+--! Checks if the table contains item
+function is_contains(table, value)
+  for index, val in ipairs(table) do
+    if val == value then
+      return true
+    end
+  end
+  return false
+end
+
 local function serialize_string(val, options)
   local level = options and options.long_bracket_level_start or 0
   while string.find(val, ']' .. string.rep('=', level) .. ']') do
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -30,7 +30,7 @@ globals = { -- Globals
             "pause_gc_and_use_weak_keys", "permanent",
             "rangeMapLookup", "rnc", "strict_declare_global",
             "unpermanent", "values", "serialize",
-            "array_join", "table_merge", "shallow_clone", "staff_initials_cache",
+            "array_join", "table_merge", "is_contains", "shallow_clone", "staff_initials_cache",
             "hasBit", "bitOr", "inspect", "getRandomEntryFromArray", "isTableEmpty",
             "stripTrailingSlashes", "isDirectory", "canOpenDirectory", "tracy",
 
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -30,9 +30,11 @@ globals = { -- Globals
             "pause_gc_and_use_weak_keys", "permanent",
             "rangeMapLookup", "rnc", "strict_declare_global",
             "unpermanent", "values", "serialize",
-            "array_join", "table_merge", "is_contains", "shallow_clone", "staff_initials_cache",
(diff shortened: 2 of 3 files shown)


CorsixTH/Lua/utility.lua
```diff
@@ -422,6 +422,16 @@ function table_merge(...)
   return result
 end
 
+--! Checks if the table contains item
+function is_contains(table, value)
```
lewri:
```suggestion
function table_contains(table, value)
```

CorsixTH/Lua/utility.lua
```diff
@@ -422,6 +422,16 @@ function table_merge(...)
...
+  for index, val in ipairs(table) do
```
lewri: Some tables are not interative, it's safer to use pairs if you're going to make the function universal

lewri: We can also drop `index` because we don't use it. Can be `_` instead

lewri (commented):
Afterload not possible as far as I can tell due to legitimate cases where tile_x and tile_y are nil. Though they may be only used during transition to destroying which happens in the same code block I think.

CorsixTH/Lua/utility.lua
lewri: In fact, entities has been known to have indexes of nil for some scenarios, see #1467

CorsixTH/Lua/utility.lua
ARGAMX: Applied 👌

CorsixTH/Lua/utility.lua
ARGAMX: > Some tables are not interative, it's safer to use pairs if you're going to make the function universal
> We can also drop `index` because we don't use it. Can be `_` instead

Applied 👌

ARGAMX: > In fact, entities has been known to have indexes of nil for some scenarios, see #1467

Okay, thanks, I know about that, but I think 1467 is irrelevant to what's going on in this PR.

lewri (approved):
Approved unless there's a suddenly better presenting temporary patch, or some issue with it.

Issue #1583, opened 2020-02-21
Crash to desktop when placing a room.


willmcc:
#### Steps to Reproduce
1. *Load save game attached to this ticket*
You should see that I had just placed an operating theatre.
2. Click the check icon to confirm placement of the room

3. On my system, the game is crashing without any error message.

#### Save Game
*It is often useful for us if you have a save game from shortly before the
issue occurs that can be used to recreate the problem. Sometimes a save that
shows the issue happening may be useful as well. As GitHub currently only
supports uploading of images, you will have to upload your your savegames to
an another source such as [Google Drive](https://drive.google.com),
[Dropbox](https://dropbox.com), or
[SkyDrive/OneDrive](https://onedrive.live.com). Add the link to the file to
your issue and make the file(s) public, so we can access your gamelog or
savegame. The most important thing is to not remove these files after you
uploaded them!*

#### Expected Behavior
I should be able to complete placing the operating theatre and add furniture, etc.

#### System Information
**CorsixTH Version:** *e.g. 0.50 or 7d886f35ca*
0.63 (also seeing in 0.62)
[Theatre - pre crash.zip](https://github.com/CorsixTH/CorsixTH/files/4238482/Theatre.-.pre.crash.zip)

**Operating System:** *e.g. Windows 10*
Windows 10
**Theme Hospital Version:** *CD, GOG.com, Origin, or Demo*
CD (re-release around late 1990s, early 2000s)
#### Gamelog.txt
*For information about where to find gamelog.txt see:
https://github.com/CorsixTH/CorsixTH/wiki/Frequently-Asked-Questions#where-do-i-find-the-configuration-or-the-gamelog-file*
```
Paste gamelog.txt output here
```
```
Created game with savegame version 127.
Warning: Trying to remove nonexistant room build callback (function: 000001BA6B65C8A0) from humanoid (table: 000001BA71FE5DD0).
Warning: Trying to remove nonexistant room remove callback (function: 000001BA72304E70) from humanoid (table: 000001BA71FE5DD0).
Warning: Trying to remove nonexistant room build callback (function: 000001BA04B51A90) from humanoid (table: 000001BA78F75BC0).
Warning: Trying to remove nonexistant room remove callback (function: 000001BA78031CC0) from humanoid (table: 000001BA78F75BC0).
Warning: Trying to remove nonexistant room build callback (function: 000001BA793ACF20) from humanoid (table: 000001BA7C03ABF0).
[...]

PR #3375, opened 2026-06-06
Fix regressed staff circle tooltip

lewri:
**0.70.0 beta 1**

- Fixes a regression found in the save at https://github.com/CorsixTH/CorsixTH/issues/1583#issuecomment-3368900763 where 0.68.0 saves would crash when hovering over the staff view circle.

### DIFF
--- a/CorsixTH/Lua/dialogs/staff_dialog.lua
+++ b/CorsixTH/Lua/dialogs/staff_dialog.lua
@@ -147,8 +147,8 @@ function UIStaff:UIStaff(ui, staff)
 
   self:makeTooltip(_S.tooltip.staff_window.salary, 90, 191 + offset, 204, 214 + offset)
   -- Non-rectangular tooltip has to be realized with dynamic tooltip at the moment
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/dialogs/staff_dialog.lua
```diff
@@ -147,8 +147,8 @@ function UIStaff:UIStaff(ui, staff)
 
   self:makeTooltip(_S.tooltip.staff_window.salary, 90, 191 + offset, 204, 214 + offset)
   -- Non-rectangular tooltip has to be realized with dynamic tooltip at the moment
-  self:makeDynamicTooltip(--[[persistable:staff_dialog_center_tooltip]]function(x, y)
-    if is_in_view_circle(x, y, class.is(staff, Handyman)) then
+  self:makeDynamicTooltip(--[[persistable:staff_dialog_center_tooltip]]function(x, y, target_staff)
```
lewri: For some reason this is a rare case the linter will moan if you try to use the same variable name.

ARGAMX (approved):
👍

PR #3378, opened 2026-06-07
Handle messages with `owner` `nil`

ARGAMX:
Screenshot 2026-06-07 at 10 38 03

.
**0.70.0 beta 1** 
game crash if player have to deal with message which have `owner` is `nil`.

**Why is this happening?**

If user get some message with `owner` `nil` and will ignore or handle that message, game will crash.
Not all messages have `owner` `nil`.
For example, messages about the discovery of a new disease don't have an owner.

Initially there was only 1 function for deleting messages. And it was taking `owner` as a ID for target message search and delete. `UIBottomPanel:deleteMessage(owner)`. It seemed perfectly reasonable and safe to continue using this function for broader purposes. But this was a wrong assumption.

So we have to have 2 delete functions now, not 1... one searches for the target message by `owner`, and the second searches for the target message by `drawer_icon`.

Issue introduced by #3347

**How to reproduce:**

[0.70.0_fax-message-crash.sav.zip](https://github.com/user-attachments/files/28677482/0.70.0_fax-message-crash.sav.zip)
2. Load save game and wait until 26 May. Game will crash.

```Lua
Lua\dialogs\bottom_panel.lua:591: attempt to index a nil value (local 'owner')
stack traceback:
        Lua\dialogs\bottom_panel.lua:591: in method 'deleteMessage'
        Lua\dialogs\bottom_panel.lua:461: in local 'onClose'
        Lua\dialogs\message.lua:149: in method 'removeMessage'
        Lua\dialogs/fullscreen\fax.lua:244: in method 'choice'
        Lua\dialogs\message.lua:137: in method 'removeMessage'
        Lua\dialogs\message.lua:195: in method 'onWorldTick'
        Lua\window.lua:1939: in method 'onWorldTick'
        Lua\window.lua:1939: in method 'onWorldTick'
        Lua\world.lua:972: in method 'onTick'
        Lua\app.lua:1275: in function <Lua\app.lua:1272>
        (...tail calls...)
```

- When deleting messages, take in account that some messages have a `owner` `nil`. Therefore, we cannot use `owner` as a unique identifier among messages.
- Since a "draw icon" object is now created for absolutely all messages, even for those that are not yet displayed, add a function for deleting a message using this object as an identifier as a reference.

Thanks @Lewri for detecting and report.

### DIFF
--- a/CorsixTH/Lua/dialogs/bottom_panel.lua
+++ b/CorsixTH/Lua/dialogs/bottom_panel.lua
@@ -458,7 +458,7 @@ function UIBottomPanel:queueMessage(type, message, owner, timeout, default_choic
     if this_icon_index then
       table.remove(self.message_windows, this_icon_index)
     end
-    self:_deleteMessageByIcon(this_icon)
+    self:deleteMessage(nil, this_icon)
   end
 
   local message_info = {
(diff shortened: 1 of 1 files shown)


ARGAMX:
Good. Should work.

tobylane (approved):

PR #3383, opened 2026-06-08
Don't phase through walls

### LINKED ISSUE #3382: Walls are passable during a walk in build mode
Image

Reproduction (from Argus):
1. Load 0.70.0_nurse_get_through_wall_2.sav. The game will be paused.
2. Cancel the current construction, as it is defective due to issue 345/1583.
3. Start building any new room. Place the blueprint so that it covers the nurse and is adjacent to the pharmacy.
4. In the construction window, click the ✔️ button to proceed to the 2nd or 3rd construction stage.
5. Unpause and watch as the nurse passes through the wall and enters the pharmacy.

[0.70.0_nurse_get_through_wall.zip](https://github.com/user-attachments/files/28725674/0.70.0_nurse_get_through_wall.zip)

The reason behind this is because during this stage of building, the tiles underneath a humanoid become not passable. Our walk code only checks to recalculate the route if their current tile is passable. As a result, although a route recalculation is requested, it does not and will request movement to the next time on the current path instead.

The incidence of this happening are generally quite rare, but likely have happened by accident once or twice over the past few years in games. There is usually no crash as the humanoid will leave the room.

Note: this is akin to #3350 in the same way the obstruction is not handled.

lewri:
*Fixes #3382*

- If a humanoid's next tile is blocked by a wall, they will recalculate their route
- Does not fix the side object scenario

I would note with this, some silent murmurs in my head tell me the code had a design originally to try and get a humanoid off a bad tile no matter the cost. As we start to lock these down, hopefully we don't cause a humanoid to get completely stuck.
E.g. case like this
image

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/walk.lua
+++ b/CorsixTH/Lua/humanoid_actions/walk.lua
@@ -222,8 +222,9 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
 
   -- Make sure the returned wall id of a cell flag is real
   local function hasWallId(value)
-    local offset = 1024 -- wall tiles have an early_list modifier
-    return value and value - offset > 0
+    -- Account for transparency setting
+    local transparency_offset = TheApp.ui.transparent_walls and 1024 or 0
+    return value and value - transparency_offset > 0
   end
 
   -- Helper function to check for a shared wall on adjacent tiles (here and there)
@@ -251,9 +252,8 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
       _, there_wall, _ = th:getCell(there_x, there_y + 1)
     end
 
-  return hasWallId(here_wall)
-      and hasWallId(there_wall)
-      and here_wall == there_wall
+    return hasWallId(here_wall) and hasWallId(there_wall) and
+        here_wall == there_wall
   end
 
   if recalc_route then
--- a/CorsixTH/Lua/humanoid_actions/walk.lua
+++ b/CorsixTH/Lua/humanoid_actions/walk.lua
@@ -219,6 +219,43 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
       recalc_route = action.trimmed
     end
   end
+
+  -- Make sure the returned wall id of a cell flag is real
+  local function hasWallId(value)
+    local offset = 1024 -- wall tiles have an early_list modifier
+    return value and value - offset > 0
+  end
+
+  -- Helper function to check for a shared wall on adjacent tiles (here and there)
+  local function tilesHaveConcurrentWall(here_x, here_y, there_x, there_y)
+    local th = humanoid.world.map.th
+    local dx = there_x - here_x
+    local dy = there_y - here_y
+    local _, here_wall, there_wall
+
+    if dx == 1 then
+      -- Moving east: here east wall vs there west wall.
+      _, _, here_wall = th:getCell(here_x + 1, here_y)
+      _, _, there_wall = th:getCell(there_x, there_y)
+    elseif dx == -1 then
+      -- Moving west: here west wall vs there east wall.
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/humanoid_actions/walk.lua
```diff
@@ -219,6 +219,44 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
...
+  local function hasWallId(value)
+    local offset = 1024 -- wall tiles have an early_list modifier
+    return value and value - offset > 0
+  end
+
+  -- Helper function to check for a shared wall on adjacent tiles (here and there)
+  local function tilesHaveConcurrentWall(here_x, here_y, there_x, there_y)
+    local th = humanoid.world.map.th
+    local dx = there_x - here_x
+    local dy = there_y - here_y
```
lewri, line 234: Potentially an assert would be nice, But potentially that should be all the way up at line 182/183

TheCycoONE:
Are we going to 0.71 this? The ticket is a low, and it seems risky.

lewri:
> Are we going to 0.71 this? The ticket is a low, and it seems risky.

Yes, we can do. The wall compares are harmless, it'll be the on restart.

Edit: Tagged mugmuggy if he's around to check the walk behaviour stuff as he knows it more than me.

CorsixTH/Lua/humanoid_actions/walk.lua
```diff
@@ -219,6 +219,43 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
       recalc_route = action.trimmed
     end
   end
+
+  -- Make sure the returned wall id of a cell flag is real
+  local function hasWallId(value)
+    local offset = 1024 -- wall tiles have an early_list modifier
```
lewri: Replace with `DrawFlags.EarlyList`?

ARGAMX: > Replace with `DrawFlags.EarlyList`?

I think so. 
At least it's definitely a `2^10` bit.

CorsixTH/Lua/humanoid_actions/walk.lua
```diff
@@ -219,6 +219,43 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
...
+    return value and value - offset > 0
```
ARGAMX: I'm a little unsure why `=` isn't used here?
If some high-order or low-order bit is set, will the `<` and `>` yield incorrect results for bit sequence comparisons?
Maybe we should refactor this to use `function hasBit(value, bit)`?

lewri: *I* will probably leave this bit out for the moment as I still don't have confidence understanding how bits work.

Also it's weird we have this bit stuff and the hasbit function is only used once in edit room. Other uses call the number outright for some reason. But I'm happy to at least point to DrawFlags as the point of truth for that number for now.... Whenever I clean the pr up.

lewri: Not actually a bit on this sense so much it turns out. As I was debugging on your save sent over, it had transparent walls. Transparent walls modify the value with a +1024.

Code needs to be modified, will look at it hopefully on the weekend.

lewri:
Will this one will be needed with #3414? I've not kept up with how many fixes that PR is aiming for.

ARGAMX:
> Will this one will be needed with #3414? I've not kept up with how many fixes that PR is aiming for.

Nope.
With #3414 there will be a much simpler way to detect that there's a wall (or any other obstacle) on the way. 
Literally in one line.
We won't have to detect wall in such a complicated way anymore.
I will take care of it and will open the appropriate PR for it.
Thank you for your efforts here. 👍

ARGAMX:
Time to close this PR?

PR #3391, opened 2026-06-11
Basic scale factor support for sprite lists

### LINKED ISSUE #3389: Unreadable text in the top menu
Image

----

### Describe the issue

Unreadable text in the top menu for Unicode languages at `direct_zoom = false`.

----

### CorsixTH Version

0.70.0 Beta 1

Regressed in "UI scale #3159" (bisected)

----

### Operating System

macOS

----

### Steps to Reproduce

1. Set `direct_zoom = false` in `config.txt `
2. Take **0.70.0 beta 1** build. Run game.
3. In the Main Menu select Settings.
4. In `UI Scale` Select 100%. (At 200% the defect is also visible, but at 100% it is maximal) 
5. In `Language` Select any Unicode language.
6. Back to the Main Menu (Esc).
7. Start any custom level (Button 3 from the top in the main menu).
8. Expand any item in the top menu.
9. Observe that the text in the top menu is unreadable.

----

### Expected Behaviour

As it was before "UI scale 3159"

----

TheCycoONE:
Adds support for setting a scale factor on a sprite list and using it to draw. This patch does not account for the scale factor for hit tests which may be desired (though I don't see any use in the game). The immediate use for this patch is for drawing menus.

Fixes #3389

### DIFF
--- a/CorsixTH/Lua/api_version.lua
+++ b/CorsixTH/Lua/api_version.lua
@@ -32,4 +32,4 @@ Note: This file compiles as both Lua and C++. */
 
 #endif /*]] --*/
 
-return 2711;
+return 2712;
--- a/CorsixTH/Lua/dialogs/menu.lua
+++ b/CorsixTH/Lua/dialogs/menu.lua
@@ -199,10 +199,8 @@ function UIMenuBar:drawMenu(menu, canvas)
   local panel_sprites_draw = panel_sprites.draw
   local x, y, w, h = menu.x * s, menu.y * s, menu.width * s, menu.height * s
 
-  -- It would be better if spriteList supported scaling directly
-  canvas:scale(s)
-  menu.render_list:draw(canvas, menu.x, menu.y)
-  canvas:scale(1)
+  menu.render_list:setScaleFactor(s)
+  menu.render_list:draw(canvas, x, y)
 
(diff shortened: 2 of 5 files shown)


ARGAMX (approved):
Resolves the issue for me, thanks 👍

ARGAMX:
I found another very similar regression bug.
Not in the Top Menu, but in Display Info.
Also Unicode, also `direct_zoom = false`.
Reproduction scenario is similar.
I can create a separate ticket or just add this to #3159.
As you wish?

[3056 Operating Theatre blocks resolved 4.sav.zip](https://github.com/user-attachments/files/22984109/3056.Operating.Theatre.blocks.resolved.4.sav.zip)

0.69.2
DisplayInfo_0 69 2

0.70.0
DisplayInfo_0 70 0

TheCycoONE:
Looking into the second issue, but the fix can't be the same, there's no scale added around here and no spriteList involved. It's also not as pronounced.

TheCycoONE:
Lets make a new issue for this other case. Maybe there was a common cause under it all but I'm not really sure what it is.

ARGAMX:
> Lets make a new issue for this other case. Maybe there was a common cause under it all but I'm not really sure what it is.

Ok, issue created: #3394

PR #3400, opened 2026-06-14
Russian translate improvements

ARGAMX:
- Added missing strings that appeared in the 0.70.0 cycle.
- Some file organisation and bringing some order and consistency.


CorsixTH/Lua/languages/english.lua
```diff
@@ -597,8 +597,6 @@ customise_window = {
   regular_patients = "Regular Patients",
   male_only = "Male Only",
   male_and_female = "Male and Female",
-  every_month = "Every Month",
-  every_day = "Every Day",
```
ARGAMX, line 601: These strings were added in "Setting autosave frequency" 3226.
But they're not actually used.
They simply duplicate the same strings in another place.
I forgot to remove them at the time.

TheCycoONE:
Big change, I take it you dumped the translations over again?

TheCycoONE (approved):

PR #3403, opened 2026-06-15
Adviser history delete message fix

### LINKED ISSUE #3395: [Bug] Delete buttons do not immediately remove message in Adviser History
### Describe the issue

When you click the delete button belonging to the first message in the Adviser History panel, the message will be removed immediately. However, if you click any of the other delete buttons, their corresponding message will not disappear until the player forces the panel to update (such as closing/reopening the panel, moving the scrollbar, or deleting the first message).

This can lead to a subsequent bug where if you click between the delete icons enough times, the following error will occur and the Adviser History panel will get stuck to the cursor:
```
Error in buttonup handler: 
...dfnMF/usr/bin/../share/corsix-th/Lua/dialogs/adviser.lua:102: bad argument #1 to 'remove' (position out of bounds)
stack traceback:
    [C]: in function 'table.remove'
    ...dfnMF/usr/bin/../share/corsix-th/Lua/dialogs/adviser.lua:102: in method 'removeMessageFromHistory'
    ...are/corsix-th/Lua/dialogs/resizables/adviser_history.lua:71: in method 'deleteButtonClicked'
    ...are/corsix-th/Lua/dialogs/resizables/adviser_history.lua:87: in local 'callback'
    ...t_CorsixAdfnMF/usr/bin/../share/corsix-th/Lua/window.lua:732: in method 'handleClick'
    ...t_CorsixAdfnMF/usr/bin/../share/corsix-th/Lua/window.lua:1728: in method 'onMouseUp'
    ...t_CorsixAdfnMF/usr/bin/../share/corsix-th/Lua/window.lua:1711: in field 'onMouseUp'
    ...mount_CorsixAdfnMF/usr/bin/../share/corsix-th/Lua/ui.lua:911: in function <...mount_CorsixAdfnMF/usr/bin/../share/corsix-th/Lua/ui.lua:898>
    (...tail calls...)
```

This error can be resolved by making sure that all the delete buttons correctly remove their corresponding message when clicked.

### Steps to Reproduce

2. Click any of the delete buttons (except for the one corresponding to the first message)

### Expected Behaviour

Clicking the delete button next to any message will immediately delete it, and all subsequent messages will move up one row to replace it.

- [ ] Right mouse panning

### Save Game

[adviserhistory.zip](https://github.com/user-attachments/files/28920124/adviserhistory.zip)

### CorsixTH Version

0.70.0-beta2

### Operating System

Linux Mint 23

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

### Additional Information

Image

ARGAMX:
**0.70.0 beta 2**

Screenshot 2026-06-15 at 09 39 42

*Fixes #3395*

### **Describe what the proposed change does**

- Messages deleted now match the one the user clicks on.
- After message delete list now properly updates.

### **More Details**

**Firstly, previously the messages that are deleted are not the ones user click the cross on.**

When deleting a message, the order number of that message on the screen was passed. 
This number was then used as the index of the message to be deleted in the array of all messages.
For example, you have 8 messages in the history. 
The message window displays no more than 4 messages at a time.
Suppose we want to delete message number 5. 
We open the message history window. After opening the message history window, we don't see message number 5, but only the first 1-4 messages. We scroll to the very bottom, and now we see messages 5-8.
So, if we click the cross next to message 5, the delete function will receive index 1, not 5... why 1? 
Because the current number of the message on the screen is passed, not the one in the array. This is incorrect.
This will delete the wrong message.
This also can produce `adviser.lua:102: bad argument #1 to 'remove' (position out of bounds)
stack traceback:` because this error gives opportunity to delete the number out of array bounds.
To delete the correct message, the index is now calculated as:
the number of the message on the screen + scroll offset.

**Secondly, previously the message list is not updates on delete.**

The message history list is trying to update on `UIAdviserHistory:onTick()`.
But to prevent the message history list from being updated unnecessarily, a check is placed every tick to determine whether the list has changed.
If there have been changes, it updates the list.
If there have been no updates, it doesn't update the list at that tick.
This check only remembered the topmost, most recent message. 
And each time, it checked: if the last message has changed, then it updates the list and remembers the new first message.
This logic works well if messages are deleted or added only to the very top.
But when we click the cross next to a message, we can delete not only just the first message.
Therefore, the list didn't respond or update if any message other than the first message was deleted.
I had to add variable for the message count.
So this function would check not only whether the topmost message has changed, but also whether the total number of messages has changed.
Thus, if we delete messages, the total number of messages changes, and the function updates the UI.

---------

*Additional note from lewri*
Extra commit fixes a longstanding bug where if afterLoads run through `self.windows` closed the window, it would mutate the list and cause an early termination.

### DIFF
--- a/CorsixTH/Lua/window.lua
+++ b/CorsixTH/Lua/window.lua
@@ -2151,14 +2151,23 @@ function Window:afterLoad(old, new)
     self.apply_ui_scale = true
   end
 
+  -- If a window or panel is asked to close during an afterLoad cycle we
+  -- can skip entries so use backwards iteration here instead
   if self.windows then
-    for _, w in pairs(self.windows) do
-      w:afterLoad(old, new)
+    for i = #self.windows, 1, -1 do
+      local window = self.windows[i]
+      if window then
+        window:afterLoad(old, new)
+      end
     end
   end
+
   if self.panels then
-    for _, p in pairs(self.panels) do
-      p:afterLoad(old, new)
+    for i = #self.panels, 1, -1 do
+      local panel = self.panels[i]
+      if panel then
+        panel:afterLoad(old, new)
+      end
     end
   end
 end
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local SDL = require("sdl")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 248 -- 0.70.0 beta 2
+local SAVEGAME_VERSION = 249 -- Advisor message history dialog afterLoad close
 
 class "App"
 
--- a/CorsixTH/Lua/dialogs/resizables/adviser_history.lua
+++ b/CorsixTH/Lua/dialogs/resizables/adviser_history.lua
@@ -190,3 +190,9 @@ function UIAdviserHistory:onTick()
   end
 end
 
+function UIAdviserHistory:afterLoad(old, new)
+  if old < 249 then
+    self:close()
+  end
+  Window.afterLoad(self, old, new)
+end
--- a/CorsixTH/Lua/dialogs/resizables/adviser_history.lua
+++ b/CorsixTH/Lua/dialogs/resizables/adviser_history.lua
@@ -40,6 +40,7 @@ function UIAdviserHistory:UIAdviserHistory(ui)
   self.resizable = false
 
   self.last_seen_history_top = nil  -- Stores the top message in the adviser's 'message_history' list at the time of the last update
+  self.last_history_count = nil -- Stores the messages count in the adviser's 'message_history' list at the time of the last update
   self.list_table = {}       -- List of controls in this dialog
 
   self.default_button_sound = "selectx.wav"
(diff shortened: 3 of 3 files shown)


lewri (commented):
Could we do a savegame bump and do a `self:close()` please just to be safe on editing a persistable.

lewri:
@MandrakeHorse tag for review

lewri (commented):
All seems to work functionally.

ARGAMX:
> Could we do a savegame bump and do a `self:close()` please just to be safe on editing a persistable.

I have added afterLoad but this bring some regression, I don't understand where it come from:

If the Adviser History Window was open in a saved game, after load it will close. 
However, you won't be able to open it again using hotkeys. 
Furthermore, you won't be able to open any other windows using hotkeys.
Meanwhile, the console is empty and doesn't report any problems.
But this is clearly not a healthy state.

You can reproduce it with 0.70.0 beta 2 save where Adviser History Window was open:
[0.70.0_opened_adviser_history_window.sav.zip](https://github.com/user-attachments/files/29001446/0.70.0_opened_adviser_history_window.sav.zip)

TheCycoONE (approved):

CorsixTH/Lua/window.lua
```diff
@@ -2151,13 +2151,25 @@ function Window:afterLoad(old, new)
...
     end
   end
   if self.panels then
-    for _, p in pairs(self.panels) do
+    local panels = {}
+    for _, panel in pairs(self.panels) do
+      panels[#panels + 1] = panel
+    end
+
+    for _, p in ipairs(panels) do
```
lewri: For ref this is a wider fix than this PR's scope, but it is what caused the afterLoad regression.

TheCycoONE: You could iterate over the windows and panels backwards, that's possibly cheaper if normally nothing is removed, see Window:removeWindow for an example of this done right.

lewri: As these only fire once during loading a save, do you feel strongly about backwards iteration for this specific use case?

TheCycoONE: Not strongly but I think it makes a smaller simpler PR

lewri: Changed

tobylane (approved):

PR #3411, opened 2026-06-19
Properly reset `Last room` for humanoid if he unable to return

### LINKED ISSUE #3410: Doctor is meandering and not responding to calls
Image

----

### Describe the issue

The doctor after exiting the Staff room starts meandering and not responding to calls.

----

### Save Game

[0.70.0_doctor_dont_answer_calls.sav.zip](https://github.com/user-attachments/files/29136622/0.70.0_doctor_dont_answer_calls.sav.zip)

----

### CorsixTH Version

0.70.0

----

### Steps to Reproduce

1. Load attached save.
2. Observe the doctor in the Staff room with surgery skill.
3. Press 5.
4. Observe the doctor leave the Staff room and start meandering.
5. Listen for the announcement for 2 surgeons needed.
6. Observe that the doctor doesn't answer that call. And only 1 doctor moves to the Operating Theatre.

----

### Expected Behaviour

That doctor should answer that call.

----

ARGAMX:
**0.70.0 beta 2**

*Fixes #3410*

- Properly reset `Last room` for humanoid if he unable to return to that last room.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/use_staffroom.lua
+++ b/CorsixTH/Lua/humanoid_actions/use_staffroom.lua
@@ -128,7 +128,8 @@ local function use_staffroom_action_start(action, humanoid)
         humanoid:queueAction(room:createEnterAction(humanoid))
         humanoid:setDynamicInfoText(_S.dynamic_info.staff.actions.heading_for:format(room.room_info.name))
       else
-        -- Send the staff out of the room
+        -- Forget last room and send the staff out of the room
+        humanoid.last_room = nil
         humanoid:queueAction(MeanderAction())
       end
     else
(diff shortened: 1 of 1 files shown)


lewri (commented):
This will no longer execute the meander code below with this change:
https://github.com/CorsixTH/CorsixTH/blob/020162de358ce8f080093f5208db108a5e98762f/CorsixTH/Lua/humanoid_actions/meander.lua#L51-L58

It looks like the last_room is still needed; and perhaps we should be seeing why training/research didn't clear if it fails.
Maybe on fail check if last room is training/research and unset it if so?

ARGAMX:
> This will no longer execute the meander code below with this change:
> 
> https://github.com/CorsixTH/CorsixTH/blob/020162de358ce8f080093f5208db108a5e98762f/CorsixTH/Lua/humanoid_actions/meander.lua#L51-L58
> 
> It looks like the last_room is still needed; and perhaps we should be seeing why training/research didn't clear if it fails. Maybe on fail check if last room is training/research and unset it if so?

Changes have been made 👍

lewri (approved):

PR #3412, opened 2026-06-20
[Translation] Update french translation

Ouaz:
Mainly update the Menu translation.

- Add missing menu translations (buttons and tooltips)
- Fix translations being too long to fit the UI
- Fix some typos and improve some translations

I hope this is not too late, as I see there's already a Release Candidate.

### DIFF
--- a/CorsixTH/Lua/languages/french.lua
+++ b/CorsixTH/Lua/languages/french.lua
@@ -500,12 +500,12 @@ diseases = {
 }
 
 confirmation = {
-  quit = "Vous avez choisi Quitter. Voulez-vous vraiment quitter le jeu ?",
-  return_to_blueprint = "Êtes-vous sûr de vouloir revenir au mode Tracé ?",
-  restart_level = "Êtes-vous sûr de vouloir relancer ce niveau ?",
-  overwrite_save = "Il y a déjà une partie sauvegardée ici. Êtes-vous sûr de vouloir l'écraser ?",
+  quit = "Vous avez choisi Quitter. Voulez-vous vraiment quitter la partie ?",
+  return_to_blueprint = "Voulez-vous vraiment revenir au mode Tracé ?",
+  restart_level = "Voulez-vous vraiment relancer ce niveau ?",
+  overwrite_save = "Il y a déjà une partie sauvegardée ici. Voulez-vous vraiment l'écraser ?",
   delete_room = "Voulez-vous vraiment détruire cette salle ?",
-  sack_staff = "Êtes-vous sûr de vouloir licencier ?",
+  sack_staff = "Voulez-vous vraiment licencier ce membre du personnel ?",
 }
 
 -- The originals of these strings contain one space too much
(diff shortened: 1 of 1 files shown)


CorsixTH/Lua/languages/french.lua
```diff
@@ -1256,6 +1297,21 @@ tooltip.machine_menu = {
   close = "Fermer la fenêtre de la liste des machines.",
 }
 
+adviser_history = {
+  message = "Messages",
+  close = "Fermer",
+}
+
+tooltip.adviser_history = {
+  delete_message = "cliquer pour supprimer ce message",
```
Alberth289346: AS a suggestion, starting with an uppercase letter here? The other messages seem to all do that.

lewri:
> I hope this is not too late, as I see there's already a Release Candidate.

I'm happy to roll with translations as long as everyone else is.

lewri (commented):
Your test failures below are likely that you accidentally put a space after the `"` at the end of the lines here

CorsixTH/Lua/languages/french.lua:928:77-78: (W612) line contains trailing whitespace
    CorsixTH/Lua/languages/french.lua:995:127-128: (W612) line contains trailing whitespace
    CorsixTH/Lua/languages/french.lua:999:56-57: (W612) line contains trailing whitespace

lewri (approved):

Ouaz:
Sorry for the missing uppercase letter and the trailing spaces. Should be OK now.

I have a question though:

With the default font (CorsixTHUnicode.ttf), the `Ê` and `/` glyphs are not rendered. (based on the [Localization](https://github.com/CorsixTH/CorsixTH/wiki/Localization) wiki)
`Ê` appears as `?` and `/` is replaced by a blank space.

What users of CorsixTH usually do ? Do they use another font or most of them play with the default font ?
Therefore, is it better to avoid `Ê` and `/` glyphs, or use them if most of the players change to another font ?

lewri:
> With the default font (CorsixTHUnicode.ttf), the `Ê` and `/` glyphs are not rendered. (based on the [Localization](https://github.com/CorsixTH/CorsixTH/wiki/Localization) wiki) `Ê` appears as `?` and `/` is replaced by a blank space.
> 
> What users of CorsixTH usually do ? Do they use another font or most of them play with the default font ? Therefore, is it better to avoid `Ê` and `/` glyphs, or use them if most of the players change to another font ?

No this is correct behaviour. Because french uses the original game font files it is limited to the characters in the table shown in that wiki page (some fonts don't have some characters though, because the original game needed to save space where it could). The language file can be switched to the unicode declaration to use the unicode font instead, but it doesn't look as pure as the original. 

I can't remember if there was a future enhancement to fetch the unicode character if it is missing from the font file but that is what it would need to be.

TheCycoONE:
It wasn't about saving space, code page 850 would have had better coverage in the same space as 437.  I think it's just because they were English and didn't think about it very much/went with what was popular (on English computers) in the 90s. Even by 1998 this was a bit of an archaic choice; many apps were using Windows-1252, so I suspect it was inherited from Theme Park like many of the other more dated  aspects of the engine.

PR #3414, opened 2026-06-20
Get humanoids to avoid some tiles, and unconditionally refuse to enter an unpassable tile.

### LINKED ISSUE #3404: Add a "avoid-this-tile" flag to tiles
Currently blueprint-tiles are unpassable. Path-finding has been modified to ignore unpassable if the tile of the humanoid is unpassable itself. That makes that the pathing algorithm ignores all forms blockages, including things like walls and objects.

By switching to tiles that should be avoided with a blue-print, the pathfinder can distinguish blueprint tiles from real blockages, and avoid generating paths that are unrealistic.

Also, it's useful to make tiles to avoid more expensive in paths, to make the path algorithm avoid them as much as possible.

Alberth289346:
- Cleanup the Lua map/path-finder interface functions somewhat.
- Add an 'avoid this tile' flag to the map flags and make it available from Lua.
- Disallow idling at tiles that should be avoided.
- Unconditionally refuse to enter a tile if it is not passable.
  While this prevents walking through unpassable tiles, this is going to be a major problem if a humanoid finds itself in an unpassable area.
- Add a path cost to the path finder, and make tiles to avoid 128 times more expensive than normal tiles.

closes #3404


Alberth289346:
Patch needs an API bump due to the new flag that is exposed in Lua

CorsixTH/Src/th_lua_map.cpp
```diff
@@ -709,22 +710,28 @@ int l_map_setwallflags(lua_State* L) {
...
   if (lua_gettop(L) >= 7) {
-    pNode->tile_layers[tile_layer::ground] =
-        static_cast<uint16_t>(luaL_checkinteger(L, 4));
-    pNode->tile_layers[tile_layer::north_wall] =
-        static_cast<uint16_t>(luaL_checkinteger(L, 5));
-    pNode->tile_layers[tile_layer::west_wall] =
-        static_cast<uint16_t>(luaL_checkinteger(L, 6));
-    pNode->tile_layers[tile_layer::ui] =
-        static_cast<uint16_t>(luaL_checkinteger(L, 7));
+    uint16_t value = static_cast<uint16_t>(luaL_checkinteger(L, 4));
```
TheCycoONE, line 724: Any reason for preferring the extra temporary L-value?

CorsixTH/Src/th_pathfind.cpp
```diff
@@ -663,6 +666,7 @@ void pathfinder::depersist(lua_persist_reader* pReader) {
     }
     path_node* pPrevNode = &nodes[iY * iWidth + iX];
     pNode->distance = iLength - 1 - i;
+    pNode->cost = iLength - 1 - i;  // Can be too optimistic.
```
TheCycoONE, line 669: // Heuristic: values <= actual cost are acceptable

TheCycoONE (approved):
A* changes look correct. I have mixed opinions on the various places temporary L-values were introduced; some are obviously clearer, some I feel just add noise.

Nice change.

CorsixTH/Src/th_lua_map.cpp
TheCycoONE: I see you did it throughout, this is a spot where I don't feel it adds anything.

Alberth289346: I really don't like this kind of code that changes indent every single line.
```
    pNode->tile_layers[tile_layer::ground] =
        static_cast<uint16_t>(luaL_checkinteger(L, 4));
    pNode->tile_layers[tile_layer::north_wall] =
        static_cast<uint16_t>(luaL_checkinteger(L, 5));
    pNode->tile_layers[tile_layer::west_wall] =
        static_cast<uint16_t>(luaL_checkinteger(L, 6));
    pNode->tile_layers[tile_layer::ui] =
        static_cast<uint16_t>(luaL_checkinteger(L, 7));
```
I'd really like write each assignment in a single line if that was possible. Then it's trivially readable.

TheCycoONE: Ok, but I would shrink that block like this:

```cpp
    auto lua_int16 = [&L](int arg) {
      return static_cast<uint16_t>(luaL_checkinteger(L, arg));
    };

    pNode->tile_layers[ground] = lua_int16(4);
    pNode->tile_layers[north_wall] = lua_int16(5);
    pNode->tile_layers[west_wall] = lua_int16(6);
    pNode->tile_layers[ui] = lua_int16(7);
```

The enum doesn't need the prefix here, and the repetitive lua stuff can be replaced nicely with a lambda.

Edit: I might debate whether to pass L as an argument or a lambda reference. I suspect they're identical with compiler operations, but without it's cheaper if the lambda captures nothing but then you have a repetitive `L,`

CorsixTH/Src/th_lua_map.cpp
```diff
@@ -514,6 +514,7 @@ const std::map<std::string, map_tile_flags::key> lua_tile_flag_map{
     {"buildableEast", map_tile_flags::key::buildable_e_mask},
     {"buildableSouth", map_tile_flags::key::buildable_s_mask},
     {"buildableWest", map_tile_flags::key::buildable_w_mask},
+    {"avoidTile", map_tile_flags::key::avoid_tile_mask},
```
ARGAMX, line 517: I noticed that all these tile flags don't use the word "tile" in their names, only this new flag does.
e.g. "passable", "buildable"
This seems somehow inconsistent compared to the other flags.
Why don't other flags have the word "tile" in their names then?
Maybe the new flag doesn't actually need it in its name either?
After all, these flags all are properties of tile anyway.

TheCycoONE: Sure, avoid_mask would be more consistent e.g. with do_not_idle_mask.

ARGAMX (approved):
Thank you, Albert 👍

PR #3417, opened 2026-06-21
Fix panel keys overflow

lewri:
**0.70.0 rc 1**

Thanks Mickmane for bringing to our attention!

<details>
<summary>Expand for bug screenshot</summary>

image
</details>

*Fixes Panel Keys hot key assignment overflowing due to new adviser history*

- Adds a second panel keys tab in hotkey assignment. **This fix is dirty, because we're in a string freeze, but it should be enough to get by for release**.

<details>
<summary>Expand for fix for now</summary>

image
</details>

I will make a second PR to add the proper strings, but the FIXME is in case I'm otherwise unavailable.


TheCycoONE (approved):

TheCycoONE:
Can you paste an after screenshot?

ARGAMX (approved):

PR #3423, opened 2026-06-24
Actually show patients are dying

lewri:
*Fixes regressed behaviour where patient dying moods did not show/update*

- Fixes old_dying_mood being given a table or index and new_dying_mood being given nil
- Checks if an index is returned on lookup to exit early
- Assigns values from the returned lookup properly (more cost effective than an unpack for number of uses)
- Puts rangemap outside the function to only generate once per session


ARGAMX (approved):
For me it resolves the issue. Nice 👍

PR #3432, opened 2026-06-27
Fix window hit test on scrollwheel

### LINKED ISSUE #3431: [Bug] UI Scaling: Hit test failure for mousewheel zoom
### Describe the issue

When using UI scaling, the middle of the screen is reporting a hittest with UIBottomPanel.

### Steps to Reproduce

1. Set UI scaling to 200%

2.  Load a game, attempt to mousewheel zoom around the middle areas of the screen (across the window in a small rectangle that matches UIBottomPanel's size)

### Expected Behaviour

Mousewheel zoom scroll should happen

- [ ] Right mouse panning

### Save Game

### CorsixTH Version

0.70.1 rc1

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml

```

### Additional Information

Debug check
```lua
function GameUI:onMouseWheel(x, y)
  local inside_window = false
  if self.windows then
    for _, window in ipairs(self.windows) do
      if window:hitTest(self.cursor_x - window.x, self.cursor_y - window.y) then
        print("In window?!", class.type(window))
        inside_window = true
        break
      end
    end
  end
  if not inside_window then
    print("Not in window")
    -- Apply momentum to the zoom
    if math.abs(self.current_momentum.z) < 12 then
      self.current_momentum.z = self.current_momentum.z + y
    end
  end
  return UI.onMouseWheel(self, x, y)
end
```

Resolution used for test:
1280x960 - 2x UI scale

TheCycoONE:
Fixes #3431

### DIFF
--- a/CorsixTH/Lua/game_ui.lua
+++ b/CorsixTH/Lua/game_ui.lua
@@ -784,7 +784,8 @@ function GameUI:onMouseWheel(x, y)
   local inside_window = false
   if self.windows then
     for _, window in ipairs(self.windows) do
-      if window:hitTest(self.cursor_x - window.x, self.cursor_y - window.y) then
+      local wx, wy = window:getRealXY()
+      if window:hitTest(self.cursor_x - wx, self.cursor_y - wy) then
         inside_window = true
         break
       end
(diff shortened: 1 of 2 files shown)


ARGAMX (approved):
Works for me 👍

lewri (approved):

PR #3433, opened 2026-06-27
Cap UI scales to 3x

lewri:
*Fixes unlimited UI scaling*

I.e. this was possible:
image

- #3170 made it possible for UI scales to be unlimited if the resolution was high enough. Now force caps to 3x

@TheCycoONE <del>this only fixes the UI setting. It likely it still possible to change in the settings file. File has notes on what values to use already.</del> Up to you if you want this merged now or after release.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -1132,10 +1132,10 @@ function App:fixConfig()
     elseif key == "height" and (type(value) ~= "number" or value < App.MIN_WINDOW_HEIGHT) then
       self.config[key] = App.MIN_WINDOW_HEIGHT
 
-    -- For scale, clamp to integer scale >= 1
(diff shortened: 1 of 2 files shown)


TheCycoONE:
Why? If you actually have a resolution high enough I don´t see the harm.

TheCycoONE:
Of course the game will break for other reasons as mentioned in the channel, but they are unrelated to UI scaling.

lewri:
> Why? If you actually have a resolution high enough I don´t see the harm.

If we've written explicitly in our documentation and publishing we're going up to 3x I think we should stick to that. If we're going to allow it to be unlimited then the documentation should be changed instead (I'm not supportive of saying one thing and actually doing another)

I also imagine at 4x and beyond the quality of the upscale is going to diminish more and more.

TheCycoONE:
Change the documentation.

PR #3436, opened 2026-06-28
[RDY] SDL3

TheCycoONE:
Port of CorsixTH to SDL 3.4+

This is a conservative port, with only minimal changes and an effort to preserve the SDL2 look and feel.

Some distros ship SDL 3.4 but do not yet ship SDL3_mixer - so a FETCH_SDL_MIXER option was added to make it easy to build there, alternately vcpkg works well even on older distros.

There is a warning message printed to the console at least on my system that appears to be a regression in SDL_Mixer: https://github.com/libsdl-org/SDL_mixer/issues/875 but it does not cause any problems running the game.

On Windows with sf3 soundfonts (particularly larger sf3 soundfonts) there is a noticeable delay between music tracks. In the short term using an sf2 soundfont e.g. https://musical-artifacts.com/artifacts/1229 is the recommended workaround.

It is noteworthy that the wav parser in SDL3_mixer is more strict about malformed files than SDL2_mixer was. Certain unofficial versions of Theme Hospital are known not to work, but retail versions appear fine as far as we've seen.

A few efforts are outside of the scope of this PR but should be followed up on afterwards:
* HiDPI support - A key feature of SDL3 was improved HiDPI support and we should take advantage.
* Fluidsynth - We could make a better user experience by taking more control of our fluidsynth playback rather than using SDL3_mixer.
* Further refactoring - I saw lots of opportunities to refactor in the process of making this patch series and had to resist to keep it small.
    * sdl_audio.cpp should be merged with th_sound.cpp
    * SDL_SetLogicalRenderingPresentation is used unnecessarily in the zoom code
    * some form of dependency injection would be better than globals (fps / sdl_mixer) or throwing lots of singleton classes in lua registers.


TheCycoONE:
I suspect we want to at least optionally pull in the abandoned SDL touch support before we merge for parity.

ARGAMX:
1.
I built the current version, but I started having problems with cursor clicks (on trackpad) and keyboard hotkeys.

Now, after start to click any button in the main menu (or anywhere in the interface), I have to quickly double-click it (like on the Windows desktop to launch any shortcut).
Or I have to simultaneously hold Command key (the spiritual equivalent of Ctrl in Windows) on the keyboard for a single cursor click to work.
Otherwise any single clicks are ignored.
The keyboard shortcut keys are completely non-functional.
Even keyboard arrow keys that move the map view don't work.

BUT
If I switch to another desktop window (I run in windowed mode) and then back to CorsixTH window, then a single click starts working, so trackpad starts works good and so I no longer need to hold down the Command key. Also keyboard start to working properly.

I didn't see this issue in yesterday's build.

ARGAMX:
2.
Regarding resolution on macOS.
Yesterday's issue with only seeing 1/4 of the game is gone.
But now, the entire visible area is compressed into that 1/4 area.
So, I'm running the game at 1074x768, but the actual window is 537x384 (in terms of virtual desktop pixels, not physical ones), and the entire visible area is squeezed into that virtual 537x384.

Screenshot 2026-07-16 at 11 37 05

If I choose a higher resolution, the window gets larger, but it still doesn't match the operating system's virtual pixels size.
Up to that moment this was seemed like a bug to me.
BUT

At this point, I noticed that I was essentially setting physical pixels, not virtual ones, with the game's settings.
Of course, maybe I'm just imagining things, but at least the image clarity in the game has improved significantly!

I actually see that this 537x384 window is incredibly detailed and actually represents many more pixels on a physical monitor, not 537x384.

So I launched 0.70.0 with 1074x768 100% settings.
And this branch with 2048x1536 200% settings.

This way, I got two absolutely identically sized windows with identical interface sizes. (Although I don't understand where the 100 horizontal pixels went, since 1074 multiplied by 2 is 2148, not 2048)

In the main menu, I don't see any significant differences.
But when I loaded the save, I saw the hospital with twice the detail !!!

0.70.0 in this regard It's just a low quality blurry game ; just compare for yourself:

👎 SDL2_CTH.70_Screenshot 2272 × 1816: 
SDL2_CTH 70_Screenshot

👍 SDL3_CTH.71.dev_Screenshot 2272 × 1816: 
SDL3_CTH 71 dev_Screenshot

The same piece:

👎 SDL2_CTH.70 :
SDL2_CTH 70

👍 SDL3_CTH.71.dev :
SDL3_CTH 71 dev

I want that second (SDL3) option! I don't want to play this blurry game anymore!

ARGAMX:
3.
Related to music/sound.
Music and announcements - I didn't notice any differences or changes compared to 0.70.0. The music is playing. Everything works fine.
Sounds - everything works too, except for one thing: When you press buttons in the game's main menu, there's no clicking sound. The same applies to the buttons on the bottom panel (for example, the "Build Rooms" button). Launch the game without music and you'll notice that the main menu is completely silent. No any sound.

ARGAMX:
4.

Another curious bug with `direct_zoom = false`.
Below a certain line on the screen, certain objects in a certain orientation are cutted or invisible.
Move the screen down, and they appear again!
Moreover, the disappearance line on the screen is fixed and sharply cuts off objects.

Note how the sofas and beds in the ward are cropped in the screenshot:
General Diagnosis cart it totally invisible.

Screenshot 2026-07-16 at 17 03 45

Fully below the cut line:

Screenshot 2026-07-16 at 17 22 40

The cut line is somewhere horizontal in the center:

Screenshot 2026-07-16 at 17 25 05

This problem applies to the following items:

1. Benches. (direction as on screen)
2. Sofa. (direction as on screen)
3. Ultra scanner. (direction as on screen)
4. Ward bed. (direction as on screen)
5. General Diagnosis cart. (faced to the South)

I'm not sure I've recorded every object, so don't rely on this list.
I haven't tested every object in every orientation.

The defect is reproduced only at scales when a huge part of the map falls into your field of view.

This line typically located some point below the center of the screen but above the bottom panel. Generally, objects are cut off at the bottom of the screen.

Reproduces at different resolution and UI scale values.

Also set `direct_zoom = false` of course.

[sdl3_cuted_objects_direct_zoom_false.sav.zip](https://github.com/user-attachments/files/30090934/sdl3_cuted_objects_direct_zoom_false.sav.zip)

TheCycoONE:
Thanks @ARGAMX 

The switch to physical pixels was intentional (HiDPI support) though it seems that MacOS and Wayland behaved differently here in SDL 2, because in 3 I'm now matching the 2 behaviour which is what you're seeing on 3. Good, we can call it out as a feature for MacOS.

The click one is really odd, maybe an SDL bug as there's nothing in our code I can think of that would cause it. Will look into it.

The direct zoom one is likely my fault / something else related to HiDPI. I'll see if I can reproduce with Wayland.

I don't reproduce missing clicks on the main menu; but I'll look into it more.

TheCycoONE:
@ARGAMX Unfortunately I wasn't able to reproduce any of these issues. Was the click one a one time thing or reproducible. For the missing parts of furniture does it matter whether you zoom or not?

ARGAMX:
> @ARGAMX Unfortunately I wasn't able to reproduce any of these issues. Was the click one a one time thing or reproducible. For the missing parts of furniture does it matter whether you zoom or not?

The clicking issue is reproducible 100% of the time.
Immediately after launching the game.
I pulled the latest changes and it still issue.

Missing parts of furniture is better to reproduce on "high scale" zoom - move the camera as high as possible above the hospital, as in the screenshot. 
I pulled the latest changes and it still issue.

Regarding CPP debugging, how about syncing with Discord and helping me set up CPP debugging in CLion?

Last time, you suggested installing CLion.
And I think I was able to build the project in CLion then.
But I couldn't seem to get CorxisTH running directly in debugger mode within CLion.
That is, breakpoints weren't working for me.
This required installing Ninja or something similar, I think.

But since I have little expertise in this area, I didn't get any further with it that time.
Maybe if you could guide me with this, I'll be able to configure my environment to support CPP debugging?

[Screen Recording 2026-07-19 at 12.01.22.webm](https://github.com/user-attachments/assets/fa226cc6-b819-4fca-9e90-d6eb045f08b1)

TheCycoONE:
Sounds good. I cannot reproduce those issues but I can produce another bug I need to fix - I cannot type the save game name.

TheCycoONE:
@ARGAMX I found an SDL bug that looks like it might be causing your input errors, added the suggested workaround, could you let me know if it helped?

ARGAMX:
> @ARGAMX I found an SDL bug that looks like it might be causing your input errors, added the suggested workaround, could you let me know if it helped?

Yes, nice, that helped!
Problem 1 is completely gone!
I can't reproduce it anymore.

ARGAMX:
5.

I also found another bug where some shadows are displayed through the staff/patient preview window, outside the preview borders:

Screenshot 2026-07-21 at 08 43 52

I discovered that this is this shadow inside the room:

Screenshot 2026-07-21 at 08 42 50

You can load this save and press 5.
[Argus_SuperDuper_100.sav.zip](https://github.com/user-attachments/files/30213775/Argus_SuperDuper_100.sav.zip)

You'll see these glitches appearing at the bottom of the screen.

This issue on both: on `direct_zoom = false` and on `direct_zoom = true`

TheCycoONE:
In that case 1 was https://github.com/libsdl-org/SDL/issues/13920

It impacts SDL2 as well but you probably weren't launching it the same way, or it only happens when the window is hidden at launch.

ARGAMX:
> It impacts SDL2 as well but you probably weren't launching it the same way, or it only happens when the window is hidden at launch.

Yes, that's true.
I just rechecked the build from the day before yesterday and noticed that I was launching the game from the command line. And then there were input issues.
But if I launch this old build from the File Explorer, the issue doesn't reoccur.
But now it doesn't reoccur in either case.

TheCycoONE:
To recap:

Issue 1 clicks/key presses: fixed (SDL focus bug)
Issue 2 native resolution instead of virtual: works as intended (HiDPI support)
Issue 3 missing main menu sounds: works for me even on MacOS / not yet diagnosed
Issue 4 some objects cut off in lower screen: Metal renderer only, failed to work around so far
Issue 5 shadows showing up from patient preview: fixed (removed empty clip workaround)

TheCycoONE:
@ARGAMX I fixed issue 4 (at least on my VM) - it also fixed my issue with the initial layout.

It turned out not to be an SDL bug, but a quirk of the SDL3 MacOS HiDPI support and me using the wrong type of 'pixel' in a spot.

ARGAMX:
> @ARGAMX I fixed issue 4 (at least on my VM) - it also fixed my issue with the initial layout.
> 
> It turned out not to be an SDL bug, but a quirk of the SDL3 MacOS HiDPI support and me using the wrong type of 'pixel' in a spot.

Fixed issue 4 for me too!
Good job! 👍

lewri:
From latest build 6149c54 (for Windows)

1/4 screen clip in intro movie and menu etc.
image
image

It also appears hidpi has gone wrong. I use 1280x720, due to the hidpi (I assume), it scaled down to 640x480

However, if I change resolution to something else and back (or fullscreen), it sizes itself properly for the rest of that session.


As another result a fullscreen setting window we use e.g. Hotkey assignment, will load off too and not bound
image

Occasional tile tearing, not in sdl2 but I think we had it elsewhere.
image

Enabling music from the menu causes a game hang of 1-2s. This includes stopping (then playing) or skipping a track. I imagine moving to next track will also cause it. Edit: Nope, it actually fails to move to the next song and plays the same one (without lag).


No other things of note right now, though I imagine `direct_zoom = off` testing hasn't been done yet

TheCycoONE:
@lewri window size issue is addressed for Windows, which I believe is also the reason for each of the layout issues you showed.

lewri:
Windowing fixed on my main machine. At some point in future I'll load it on a second machine.

New issue: Changing MIDI API from Default (software?) to Windows MM causes a hard crash and exit. But on reloading the game it stays as Windows MM. This also happens in the other direction.

Also the music hangs I talked about only occurs in the Default mode too. I suspect the asynch init calls we had in sdl2 aren't working here.

TheCycoONE:
@lewri can you clarify what you mean here?

"""
Enabling music from the menu causes a game hang of 1-2s. This includes stopping (then playing) or skipping a track. I imagine moving to next track will also cause it. Edit: Nope, it actually fails to move to the next song and plays the same one (without lag).
"""

Edit:
I see I had the loop count wrong (1 loop as in play twice, rather than 0 as in play once).

TheCycoONE:
I can reproduce the delays with the midi tracks on a windows VM. I don't like my chances of being able to resolve this in SDL3_mixer, and it does happen between tracks as well.

It will need to be profiled and then probably reported upstream.

ARGAMX (approved):

Alberth289346 (approved):

TheCycoONE:
@tobylane found a nullptr error accessing sdl_mixer::get_music_track. I've added guards for that but it should be retested.

lewri:
Occasionally the right edge of this screen appears to disappear briefly for a tick or few and comes back. I've only seen it twice so it's quite random and haven't captured it yet. Not sure if it sdl3 related yet
image

tobylane (approved):

lewri:
The screen issue is not sdl3 related but notably weird. 

Going to put it in the META
Seems to be triggered by a patient position
image

CorsixTH/Lua/app.lua
```diff
@@ -61,7 +61,9 @@ function App:App()
     music_over = self.onMusicOver,
     movie_over = self.onMovieOver,
     sound_over = self.onSoundOver,
-    multigesture = self.onMultiGesture
```
lewri, line 64: If a multigesture was active during an autosave would this crash?

CorsixTH/Src/th_movie.h
```diff
@@ -447,7 +446,7 @@ class movie_player {
 
   //! Indicate that playback is paused
   std::atomic<bool> paused{};
-  uint32_t pause_start_time{};  ///< The time in SDL ticks at which we paused
+  uint64_t pause_start_time{};  ///< The time in SDL ticks at which we paused
```
lewri, line 449: Is this killing off x86 support?

.github/workflows/Linux.yml
```diff
@@ -22,14 +22,14 @@ jobs:
...
-            packages: cmake libluajit2-5.1-dev # CMake 4, Openresty's luajit
+            packages: cmake libluajit-5.1-dev # CMake 4, Openresty's luajit
             cmakejit: '-DWITH_LUAJIT=ON -DLUA_LIBRARY=/usr/lib/x86_64-linux-gnu/libluajit-5.1.so -DWITH_MOVIES=OFF -DWITH_MIDI_DEVICE=OFF'
           - lua: Lua 5.1
             packages: 'libavcodec-dev libavformat-dev libavutil-dev libswresample-dev libswscale-dev librtmidi-dev'
             cmake: 1
             static_analysis: 1
             docs: 1
-    runs-on: ubuntu-24.04 # noble
+    runs-on: ubuntu-26.04 # noble
```
lewri: Racoon?

lewri (approved):
Image
This isn't an SDL3 bug either but I'm not sure if it's something worth fixing here. In direct_zoom off we get a resizing black bar on the right edge of the window in game. It resizes based on zoom.

Otherwise this seems okay enough.

lewri:
Our own tracking issue for the Windows audio hang may be useful so we can decide what to do about it later if SDL_Mixer can't resolve

CorsixTH/Src/th_movie.h
TheCycoONE: No

CorsixTH/Lua/app.lua
TheCycoONE: We don't save during events.

TheCycoONE:
> Occasionally the right edge of this screen appears to disappear briefly for a tick or few and comes back. I've only seen it twice so it's quite random and haven't captured it yet. Not sure if it sdl3 related yet

This is not related, there's a few of these clipping issues, but I could reproduce them in SDL2 when I isolated them.

TheCycoONE:
@lewri I added pre-decoding to the music, this moves the decoding off the main thread again (like sdl2) so it doesn't freeze (still delayed).

ARGAMX:
> Related to music/sound. Music and announcements - I didn't notice any differences or changes compared to 0.70.0. The music is playing. Everything works fine. Sounds - everything works too, except for one thing: When you press buttons in the game's main menu, there's no clicking sound. The same applies to the buttons on the bottom panel (for example, the "Build Rooms" button). Launch the game without music and you'll notice that the main menu is completely silent. No any sound.

Issue 3 recap:

The problem was with the resources, specifically with the fact that I used a third‑party `Sound-0.dat` in the Theme Hospital files. It was a `Sound-0.dat` with Russian voiceover.
That’s why the defect was reproduced for me, but not for Cyco.
Cyco found out that it turned out that `SDL3_mixer` couldn’t read some sounds from that `Sound-0.dat` because they had incorrect trailing tags.
SDL2 handled this OK. But SDL3 unable.
To solve the problem, Cyco wrote a script that fixes these defects in the sound file.
The file was added to the repository:
In #3470 PR (sound-#.dat repair tool)
We used this script on my `Sound-0.dat`. 
The output is fixed version of `Sound-0.dat`.
We tested that version and the problem is gone!

So if the problem repeats in future on some other `Sound-0.dat` of which there may be many, in this case it will also need to be run through this script.
In any case, the CorsixTH is best compatible with GOG version of the game.

PR #3437, opened 2026-06-29
Change UI scale documentation

lewri:
Remove references to an upper bound of UI scale

*Fixes #*

- Updates documentation to remove references to UI scale upper limit

(This is the same branch as the previous PR but history rewritten)


TheCycoONE:
Want to tell people how to calculate their upper bounds?

TheCycoONE:
Something like this should be no larger than the smaller of width / 640 and height / 480?

lewri:
Rewrote the whole setting block to be uniformly laid out. 
Also added Greek to the language list as that was missing.

TheCycoONE (approved):

PR #3439, opened 2026-07-03
0.70.0 final

lewri:
Date can be subject to moving depending how many more/less days we want to wait before cutting for release. I've set it to our usual 7, but I'm not sure who exactly we wait on for those days -- flatpak and repos?


ARGAMX (approved):

lewri:
@TheCycoONE merge if you're happy to roll

PR #3444, opened 2026-07-09
ci: Trigger website rebuild on more release events

rmnvgr:
When a release was deleted/unpublished, the website was not updated and the release would still appear there.

This PR make the website update on these release events (from https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=published#release):

- `created`: A draft was saved, or a release or pre-release was published without previously being saved as a draft.
- `deleted`: A release, pre-release, or draft release was deleted.
- `released`: A release was published, or a pre-release was changed to a release.
- `unpublished`: A release or pre-release was unpublished.

### DIFF
--- a/.github/workflows/update-website.yml
+++ b/.github/workflows/update-website.yml
@@ -3,7 +3,10 @@ name: Update website
 on:
   release:
     types:
-      - published
+      - created
+      - deleted
+      - released
+      - unpublished
 
 jobs:
   dispatch:


lewri (approved):
We might not need created, as my blog seems to trigger updating on drafts, but it also is harmless with how little resource it takes up

TheCycoONE (approved):

ARGAMX (approved):

PR #3445, opened 2026-07-10
Post 0.70.0 buffer

lewri:
*Fixes #*

- Post 0.70.0 10 savegame bump hotfix buffer


lewri:
@TheCycoONE It just occurred to me should we technically bump API version too, or do we care less about that? (I think API version might like less being jumped around)

ARGAMX (approved):

TheCycoONE:
no need except on API changes unless we think we missed one, I don't think we missed any.

PR #3427, opened 2026-06-26
[RFC] Unexpect patient from previous queue on expect to a next queue

ARGAMX:
*Fixes #3426*

**Describe what the proposed change does**
- Store in `patient.expected_queue` where patient expected.
- Unexpect patient from previous queue on patient expect to a next queue.
- Unexpect patient from current queue on die.

<del>This won't help saves that already have problems in queues, but it will help prevent issue in newly created games.</del> Afterload added by lewri

PR #3452, opened 2026-07-13
Error trapping for midi playback

### LINKED ISSUE #3448: [Bug] Setting MIDI Port to Virtual Port may crash CorsixTH
### Describe the issue

If a Linux user sets MIDI API to JACK and MIDI Port to Virtual Port in the Sound settings, CorsixTH may freeze before crashing. It will then always crash on subsequent launches until the user either fixes or resets their configuration file.

### Steps to Reproduce

1. Launch CorsixTH
2. Go to Settings
3. Go to Sound
4. Set MIDI API to JACK
5. Set Port to Virtual Port

### Expected Behaviour

CorsixTH should continue running normally and will load correctly when relaunched.

- [ ] Right mouse panning

### Save Game

### CorsixTH Version

0.70

### Operating System

Linux 22.3 - Cinnamon 64-bit

### Theme Hospital Version

CD

### Gamelog.txt

```yaml
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock

MidiOutJack::initialize: JACK server not running?

Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock

MidiOutJack::initialize: JACK server not running?

Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock

MidiOutJack::initialize: JACK server not running?

jack_port_register called with a NULL client

MidiOutJack::openVirtualPort: JACK error creating virtual port

Failed to open Virtual MIDI port: MidiOutJack::openVirtualPort: JACK error creating virtual port
No suitable MIDI output port found
No MIDI port is open for playback
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock

MidiOutJack::initialize: JACK server not running?

Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock

MidiOutJack::initialize: JACK server not running?

Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock

MidiOutJack::initialize: JACK server not running?

Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock

MidiOutJack::initialize: JACK server not running?

jack_port_register called with a NULL client

MidiOutJack::openVirtualPort: JACK error creating virtual port

terminate called after throwing an instance of 'RtMidiError'
  what():  MidiOutJack::openVirtualPort: JACK error creating virtual port
Aborted (core dumped)
```

### Additional Information

To fix this problem so that CorsixTH will launch, edit your [configuration file](https://github.com/CorsixTH/CorsixTH/wiki/Configuration-File) and ensure these lines appear:
```
midi_api = nil
midi_port = nil
```
Alternatively, you can delete the config.txt file and CorsixTH will generate a fresh configuration file with default settings the next time it gets launched.

TheCycoONE:
Adds some additional tests to determine if a MIDI API is suitable for use and some additional error trapping if problems get through that test.

Fixes #3448

Should be tested by @MandrakeHorse

### DIFF
--- a/CorsixTH/Src/midi_player.cpp
+++ b/CorsixTH/Src/midi_player.cpp
@@ -148,7 +148,7 @@ bool api_supports_virtual_port(RtMidi::Api api) {
 bool test_midi_api(RtMidi::Api api) {
   try {
     RtMidiOut test_midi(api);
-    return test_midi.getPortCount() > 0 || api_supports_virtual_port(api);
+    return test_midi.getPortCount() > 0;
   } catch (const RtMidiError&) {
     std::printf("INFO: Cannot query %s", midi_api_to_name(api).c_str());
     return false;
--- a/CorsixTH/Src/midi_player.cpp
+++ b/CorsixTH/Src/midi_player.cpp
@@ -94,6 +94,34 @@ RtMidi::Api midi_api_from_string(std::string_view apiName) {
   throw std::invalid_argument("Unknown API name");
 }
 
+std::string midi_api_to_name(RtMidi::Api api) {
+  switch (api) {
+    case RtMidi::LINUX_ALSA:
+      return {"ALSA"};
+    case RtMidi::UNIX_JACK:
+      return {"JACK"};
+    case RtMidi::MACOSX_CORE:
+      return {"CoreMIDI"};
+    case RtMidi::WINDOWS_MM:
+      return {"Windows MM"};
+    case RtMidi::WEB_MIDI_API:
+      return {"Web MIDI API"};
+#if RTMIDI_VERSION_MAJOR > 5
+    case RtMidi::WINDOWS_UWP:
+      return {"Windows UWP"};
+    case RtMidi::ANDROID_AMIDI:
+      return {"Android AMidi"};
+#endif
+    case RtMidi::RTMIDI_DUMMY:
+    case RtMidi::UNSPECIFIED:
+    case RtMidi::NUM_APIS:
+      return {""};
+  }
+
+  // Silences compiler warnings
+  return {""};
+}
+
 /**
  * Return whether the given RtMidi API supports the openVirtualPort method.
  *
@@ -107,6 +135,23 @@ bool api_supports_virtual_port(RtMidi::Api api) {
       return true;
     default:
       return false;
+  }
+}
+
+/**
+ * Test if the specified API is likely supported by the system runtime.
+ *
+ * The test currently attempts to establish a connection to the API to query the
+ * port count. If that fails, or there is no possible port for the API we
+ * return false.
+ */
+bool test_midi_api(RtMidi::Api api) {
+  try {
+    RtMidiOut test_midi(api);
+    return test_midi.getPortCount() > 0 || api_supports_virtual_port(api);
+  } catch (const RtMidiError&) {
+    std::printf("INFO: Cannot query %s", midi_api_to_name(api).c_str());
+    return false;
   }
 }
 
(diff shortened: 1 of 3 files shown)


MandrakeHorse:
I've built from midi_bad_port_fix but now the program crashes when I set MIDI API to JACK (`midi_api = [[JACK]]`), with these errors appearing in my console:

```
MidiOutJack::initialize: JACK server not running?


MidiOutJack::initialize: JACK server not running?

Error while fetching children for /lost+found: cannot open /lost+found: Permission denied

MidiOutJack::initialize: JACK server not running?


MidiOutJack::initialize: JACK server not running?


MidiOutJack::initialize: JACK server not running?


MidiOutJack::initialize: JACK server not running?


MidiOutJack::initialize: JACK server not running?

Segmentation fault (core dumped)

**Comment by TheCycoONE** [comment]
@MandrakeHorse if the config file is not set to Jack, do you still see Jack in the list? If not I think this should be merged (not sure if to 0.70 or master)

File: CorsixTH/Src/midi_player.cpp [hunk 1]
```diff
@@ -94,6 +94,34 @@ RtMidi::Api midi_api_from_string(std::string_view apiName) {
...
+#if RTMIDI_VERSION_MAJOR > 5
+    case RtMidi::WINDOWS_UWP:
+      return {"Windows UWP"};
+    case RtMidi::ANDROID_AMIDI:
+      return {"Android AMidi"};
+#endif
+    case RtMidi::RTMIDI_DUMMY:
+    case RtMidi::UNSPECIFIED:
+    case RtMidi::NUM_APIS:
+      return {""};
```
Alberth289346, line 118: This gets printed in case of an error, I am guessing. Would `"Unknown"` be better here?

Alberth289346 (approved):
Looks ok

MandrakeHorse:
> @MandrakeHorse if the config file is not set to Jack, do you still see Jack in the list? If not I think this should be merged (not sure if to 0.70 or master)

Yes, Jack still appears as an option in MIDI API, and selecting it causes CorsixTH to crash.

CorsixTH/Src/midi_player.cpp
TheCycoONE: This is dead code, it needs to exist only because I want exhaustive switch condition warnings but the if condition before calling this skips them in the loop.

There were a couple of choices:
- empty string - because that's the config value for default encoder (though default means software not rtmidi)
- a sentinel value like 'unknown'
- throw an exception.

TheCycoONE:
Hmm, that means rtmidi isn't throwing an error initializing it, which makes this tricky.

TheCycoONE:
@MandrakeHorse the new commit will exclude any midi api without a real port, even if the api supports virtual ports (like jack or alsa). I think that's reasonable? On my machine both JACK and ALSA have a midi through port by default even without any real devices connected, so I don't see any real loss of functionality. Maybe on some JACK setups the user would need to plug in their device before starting CorsixTH.

MandrakeHorse:
> @MandrakeHorse the new commit will exclude any midi api without a real port, even if the api supports virtual ports (like jack or alsa). I think that's reasonable? On my machine both JACK and ALSA have a midi through port by default even without any real devices connected, so I don't see any real loss of functionality. Maybe on some JACK setups the user would need to plug in their device before starting CorsixTH.

@TheCycoONE, I've just recompiled from the latest version of this branch and have some good news!

- The MIDI API dropdown only contains Default and ALSA. (i.e. Jack no longer appears for me and therefore cannot be selected.)
- When the MIDI API is set to ALSA, the MIDI Port dropdown contains Default, Midi Through:Midi Through Port-0 14:0, and Virtual Port. CorsixTH does not crash when I select any of these options (even though I cannot hear any music playing).

Therefore, the bug that I was experiencing on my specific computer no longer occurs.

TheCycoONE:
@lewri the remaining question is whether to cut a 0.70.1

lewri:
There is this that might be a candidate (though I've never experienced it, only that nothing happened).
There is also #3462 to prevent the entry crashes in psych room. 

Those are both some hard crashes. I think Argus wanted #3427 though I am not sure if that is a scope creep for a hotfix as it is not a hard crash, it's just messy queue data that isn't cleaned up properly.

CorsixTH/Src/midi_player.cpp
Alberth289346: `else crash();` is the standard pattern in my code everywhere, but here I can live with anything including keeping it as-is.

Alberth289346 (approved):

ARGAMX:
> There is this that might be a candidate (though I've never experienced it, only that nothing happened). There is also #3462 to prevent the entry crashes in psych room.
> 
> Those are both some hard crashes. I think Argus wanted #3427 though I am not sure if that is a scope creep for a hotfix as it is not a hard crash, it's just messy queue data that isn't cleaned up properly.

Also I have another regression crash in the bottom drawer in 0.70.0.
I haven't created issue for it yet.

TheCycoONE:
Ok, I'll merge this to 0.70 then. If we do one for just this I think it should be linux only or linux and mac only; doesn't impact windows at all.

PR #3455, opened 2026-07-14
Use `avoidTile` flag on room build

### LINKED ISSUE #3382: Walls are passable during a walk in build mode

### LINKED ISSUE #3386: [Bug] Patients fail to reroute on blueprints when original room destroyed
### Describe the issue

[Found by Argus]

If you build a room in pause mode ontop of a patient, and their original target room was removed while paused; the game crashes because of no route available (the queue action's x and y coordinates become nil).

NB: not a regression from current cycle

### Steps to Reproduce

2.  While paused, remove the middle pharmacy in view (see screenshot)

3. Remaining paused, build any room ontop of the humanoid that was next to the door of the middle pharmacy (blonde haired male, blue shirt with white jacket).
4. Reach doors phase, place a door and continue.
5. Observe crash.

### Expected Behaviour

The patient should be able to get off the blueprint tile successfully.

- [ ] Right mouse panning

### Save Game

[queuepositionchange_error.zip](https://github.com/user-attachments/files/28770801/queuepositionchange_error.zip)

### CorsixTH Version

0.70.0 Beta 1

### Operating System

Windows 10 (Argus also being on Mac)

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml
Error in buttonup handler: 
../Resources/Lua/humanoid_actions/queue.lua:236: Patient in a queue has no calculated distance from themselves to their target door/desk!
stack traceback:
    [C]: in global 'error'
    ../Resources/Lua/humanoid_actions/queue.lua:236: in method 'onChangeQueuePosition'
    ../Resources/Lua/dialogs/edit_room.lua:302: in method 'clearArea'
    ../Resources/Lua/dialogs/edit_room.lua:226: in local 'callback'
    ../Resources/Lua/window.lua:732: in method 'handleClick'
    ../Resources/Lua/window.lua:1728: in field 'onMouseUp'
    ../Resources/Lua/dialogs/place_objects.lua:492: in function <../Resources/Lua/dialogs/place_objects.lua:491>
    (...tail calls...)
    ../Resources/Lua/window.lua:1711: in field 'onMouseUp'
    ../Resources/Lua/ui.lua:911: in function <../Resources/Lua/ui.lua:898>
    (...tail calls...)
Error in buttonup: 0
```

### Additional Information

Image

Another part of this is that during a room's destruction, `Queue:rerouteAllPatients()`is called with a built in delay to ensure the room is destroyed. If the game was unpaused this works fine, but while pauses the actions hang in limbo after the queue action. The patient needs to be told immediately to get out their current queue and to a new one.

An experimental partial success is with the adjusted function below, but currently can lead to a rare occasion of an improper queue action so needs work.

```lua
--! Called when reception desk is destroyed, or when a room is destroyed from a crashed machine.
--!param room_id Id of the room to reroute to, 'nil' for reception.
function Queue:rerouteAllPatients(room_id)

  for i = #self, 1, -1 do
    local humanoid = self[i]
  --for _, humanoid in ipairs(self) do
    if class.is(humanoid, Patient) then
      -- slight delay so the desk is really destroyed before rerouting
      humanoid:setNextAction(IdleAction():setCount(1))

      local action
      if room_id then
        self:removeValue(humanoid)
        action = SeekRoomAction(room_id)
      else
        action = SeekReceptionAction()
      end
      humanoid:queueAction(action)

    elseif class.is(humanoid, Staff) then
      -- likewise believe we need action here to stop
      humanoid:setNextAction(IdleAction():setCount(1))
      humanoid:queueAction(MeanderAction())
    else
      -- other humanoids don't enter rooms
      humanoid:setNextAction(MeanderAction())
    end
  end
  for humanoid, callback in pairs(self.expected) do
    -- call the callback if registered as door is closing
    if callback then
      callback.callback()
    end
    self:unexpect(humanoid)
  end
end
```

ARGAMX:
Together with and on top of the #3414

This PR 

*Fixes #3382*
*Fixes #3386*

and forces humanoids to bypass construction sites.

- Leave the `{passable = false}` setting to ensure the room won't create blocked areas. But immediately after checking, remove the `{passable = false}` just set. This way stop use `{passable = false}` as way to kick out humanoids away from a construction site. (Albert thoughts https://github.com/CorsixTH/CorsixTH/issues/3331#issuecomment-4744284886 taken into account)
- Instead of `{passable = false}` use `{avoidTile = true}`. Place `{avoidTile = true}` and `{avoidTile = false}` in the appropriate places in the room build code.
- Some refactor and move `map:setCellFlags(x, y, flags)` code to separated `function UIEditRoom:_setCellFlagsOnBlueprint(flags)`
- Finally, in `walk.lua` check `avoidTile` while walking to prevent entering construction site and passing through walls from construction site.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/walk.lua
+++ b/CorsixTH/Lua/humanoid_actions/walk.lua
@@ -198,28 +198,33 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
   local map = humanoid.world.map.th
   map:getCellFlags(x1, y1, flags_here)
   map:getCellFlags(x2, y2, flags_there)
-  local recalc_route = not flags_there.passable and flags_here.passable
+  local avoid = (not flags_here.avoidTile) and flags_there.avoidTile
+  local not_passable = flags_here.passable and (not flags_there.passable)
+  local obstacle_on_the_way = not_passable or avoid -- approaching avoidable or impassable tile
+
   -- Also make sure that a room hasn't unexpectedly been built on top of the
   -- path since the route was calculated.
-  if not recalc_route and flags_here.roomId ~= flags_there.roomId then
+  if not obstacle_on_the_way and flags_here.roomId ~= flags_there.roomId then
     -- if going to the corridor we don't care as we are at the door
     -- as this will be false we won't bother checking for rerouting
-    recalc_route = flags_there.room
     local door = TheApp.objects.door.id
     local door2 = TheApp.objects.swing_door_right.id
     local doorcheck = {[door] = true, [door2] = true}
     -- but we should see if this is the same room id we want to go to and cancel the reroute
     -- ensure we still have a door on this route
-    if recalc_route and (humanoid.world:getObject(x1, y1, doorcheck) or
+    if flags_there.room and (humanoid.world:getObject(x1, y1, doorcheck) or
         humanoid.world:getObject(x2, y2, doorcheck)) and -- is there any door
         map:getCellFlags(path_x[#path_x], path_y[#path_y]).roomId ==
         flags_there.roomId then
       -- A walk including a trimmed path could have the last tile inside a room, and
       -- might need a new route completely (e.g. seek reception)
-      recalc_route = action.trimmed
+      obstacle_on_the_way = action.trimmed
+    else
+      obstacle_on_the_way = flags_there.room
     end
   end
-  if recalc_route then
+
+  if obstacle_on_the_way then
     if map:getCellFlags(x1, y1).passable then
       humanoid:setTilePositionSpeed(x1, y1)
       if action.on_next_tile_set then
@@ -229,6 +234,7 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
         humanoid:finishAction(action)
         return
       end
+      -- find new path
       return action:on_restart(humanoid)
     end
   end
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -198,6 +198,7 @@ function UIEditRoom:cancel()
       self.ui:setDefaultCursor(nil)
       self.check_for_clear_area_timer = nil
       self.humanoids_to_watch = nil
+      self:_setCellFlagsOnBlueprint({avoidTile = false})
     end
     self.phase = "walls"
     self:returnToWallPhase()
@@ -262,6 +263,7 @@ function UIEditRoom:clearArea()
   local world = self.ui.app.world
   world:clearCaches() -- To invalidate idle tiles in case we need to move people
   local humanoids_to_watch = {}
+  self:_setCellFlagsOnBlueprint({avoidTile = true})
   do
     local x1 = rect.x - 1
     local x2 = rect.x + rect.w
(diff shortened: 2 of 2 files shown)


CorsixTH/Lua/humanoid_actions/walk.lua
```diff
@@ -198,28 +198,33 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
   local map = humanoid.world.map.th
   map:getCellFlags(x1, y1, flags_here)
   map:getCellFlags(x2, y2, flags_there)
-  local recalc_route = not flags_there.passable and flags_here.passable
+  local avoid = (not flags_here.avoidTile) and flags_there.avoidTile
```
Alberth289346, line 201: This won't catch walking humanoids that are already in the middle of an area to avoid.

Not sure how to improve here. It may need storing the set of "avoid" tiles that exist at the time of creating the path, perhaps?
A cheaper way would be very nice.

Hmm, a flag in the path itself could do the trick. Then you can find out if path-computing has seen the flag.

Alberth289346 (approved):

CorsixTH/Lua/humanoid_actions/walk.lua
ARGAMX: > This won't catch walking humanoids that are already in the middle of an area to avoid.
> 
> Not sure how to improve here. It may need storing the set of "avoid" tiles that exist at the time of creating the path, perhaps? A cheaper way would be very nice.
> 
> Hmm, a flag in the path itself could do the trick. Then you can find out if path-computing has seen the flag.

Yes, we do have a problem with route conditions changing after the route has been built.

This isn't a new problem with `avoidTile = true`; it existed before with `passable = false`.

Currently Lua tries to compensate for this by having 'walk.lua' check every cell to detect conditions that prevent passage.

In essence, some conditions under which passage from one adjacent tile to another adjacent tile is prohibited are described in both CPP and Lua.

First, CPP checks the conditions when building the route.
Then, Lua checks the conditions when traversing the route on each subsequent cell.

If Lua detects conditions that prevent passage, `walk.lua` restarts, requesting a new route.
However, not all conditions are defined in `walk.lua`.

The problem you're referring to is that if a humanoid is already following a route and is already inside such an area, it's impossible to determine it and exit it as efficiently and correctly as possible.

After this PR, I plan to open another PR that, when applying a blueprint (and therefore aovidTile), will ask all humanoids in that area to reroute once.

But the situation where a VIP passes through a wall while standing on a bench—yes, it turns out we couldn't solve it. And we still can't.
The VIP situation is currently largely prevented by the fact that you can't place a bench on a humanoid's tile.

But it would probably be cool if we could allow placing items on humanoid tile again without the consequences of a VIP passing through a wall while standing on a bench.

But for this we need to know more properties of the route that was built.

PR #3457, opened 2026-07-16
Move "leave the area" code to a function

ARGAMX:
Only minor refactoring.

- Move "Try to make the humanoid leave the area" code to a into a separate function.

### DIFF
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -859,7 +859,7 @@ end
 
 --! Function attempts to expel the humanoid from the area.
 -- It does not require any input parameters since it expects that the area
--- to be left is already marked with tile flags.
+-- to be left is already marked with the `avoidTile` tile flag.
 function Humanoid:leaveArea()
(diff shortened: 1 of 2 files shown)


TheCycoONE (approved):

CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -857,6 +857,51 @@ function Humanoid:walkTo(tile_x, tile_y, must_happen)
       :setMustHappen(not not must_happen))
 end
 
+--! Function attempts to expel the humanoid from the area.
+-- It does not require any input parameters since it expects that the area
+-- to be left is already marked with tile flags.
```
TheCycoONE: Maybe worth indicating exactly which tile flag here.

lewri (approved):
I guess seconding the tile flag indicator suggestion too.

CorsixTH/Lua/entities/humanoid.lua
ARGAMX: > Maybe worth indicating exactly which tile flag here.

Ok, added. 👍

PR #3470, opened 2026-07-28
sound-#.dat repair tool

TheCycoONE:
Inspired by problems with a Russian sound-0.dat file which could not be read by SDL3_mixer, this tool removes all trailing tags, avoiding corruption issues.

AI Disclaimer: Written with the assistance of Google Gemini

For more context:
https://github.com/CorsixTH/CorsixTH/pull/3436#issuecomment-5523687418

### DIFF
--- a/scripts/sound-dat-repair.py
+++ b/scripts/sound-dat-repair.py
@@ -28,13 +28,14 @@ SOUND_ENTRY_POS_OFFSET = 18
 SOUND_ENTRY_LEN_OFFSET = 26
 
 
-def fix_wav_buffer(raw_wav: bytes) -> bytes:
+def fix_wav_buffer(raw_wav: bytes, clean_name: str) -> bytes:
     """Parses internal RIFF sub-chunks, trims trailing junk after 'data',
     and updates the main RIFF header length.
     """
     if len(raw_wav) < 12 or raw_wav[:4] != b'RIFF' or raw_wav[8:12] != b'WAVE':
-        print("Error: Could not find the signature of a sound-file.", file=sys.stderr)
-        sys.exit(1)
+        print("Warning: Could not find the signature of sound "
+              f"'{clean_name}', skipping this entry.", file=sys.stderr)
+        return raw_wav
 
     offset = 12
     data_found = False
@@ -54,8 +55,9 @@ def fix_wav_buffer(raw_wav: bytes) -> bytes:
         offset = chunk_data_start + chunk_size
 
     if not data_found:
-        print("Error: Could not find the data in the sound-file.", file=sys.stderr)
-        sys.exit(1)
+        print(f"Warning: Could not find the data of sound '{clean_name}', "
+              "skipping this entry.", file=sys.stderr)
+        return raw_wav
 
     # Slice off extra bytes past the data chunk payload
     cleaned_wav = bytearray(raw_wav[:true_end_offset])
@@ -69,13 +71,15 @@ def fix_wav_buffer(raw_wav: bytes) -> bytes:
 
 def process_dat(input_dat_path: str, output_dat_path: str):
     if not os.path.isfile(input_dat_path):
-        print(f"Error: Input file '{input_dat_path}' not found.", file=sys.stderr)
+        print(f"Error: Input file '{input_dat_path}' not found.",
+              file=sys.stderr)
         sys.exit(1)
 
     file_size = os.path.getsize(input_dat_path)
 
     if file_size < 4 + ARCHIVE_HEADER_SIZE:
-        print("Error: File is too small to be a valid sound archive.", file=sys.stderr)
+        print("Error: File is too small to be a valid sound archive.",
+              file=sys.stderr)
         sys.exit(1)
 
     with open(input_dat_path, 'rb') as f:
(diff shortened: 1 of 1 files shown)


lewri:
I don't mind hosting this in https://github.com/lewri/CorsixTH-Scripts if it doesn't quite belong here

scripts/sound-dat-repair.py
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env python3
+
+"""CorsixTH SOUND-0.DAT Repair Utility
+
+Some localized versions of Theme Hospital included malformed WAV data in their
```
Alberth289346:
```suggestion
Some localized versions of Theme Hospital have malformed WAV data in their
```

scripts/sound-dat-repair.py
```diff
@@ -0,0 +1,187 @@
...
+        out_f.write(raw_archive_header)
+
+        # Write final 4-byte pointer to the archive header
+        out_f.write(struct.pack('<I', new_header_position))
+
+    print(f"Successfully repacked archive to '{output_dat_path}'.")
+
+
+if __name__ == '__main__':
+    if len(sys.argv) < 3:
```
Alberth289346:
```suggestion
    if len(sys.argv) != 4:
```
Python has C-style argv.

scripts/sound-dat-repair.py
```diff
@@ -0,0 +1,187 @@
...
+SOUND_ENTRY_POS_OFFSET = 18
+SOUND_ENTRY_LEN_OFFSET = 26
+
+
+def fix_wav_buffer(raw_wav: bytes) -> bytes:
+    """Parses internal RIFF sub-chunks, trims trailing junk after 'data',
+    and updates the main RIFF header length.
+    """
+    if len(raw_wav) < 12 or raw_wav[:4] != b'RIFF' or raw_wav[8:12] != b'WAVE':
+        return raw_wav
```
Alberth289346: Could throw an exception or print an error

```
print("Input file is not a sound file.", file=sys.stderr)
sys.exit(1)
```
Returning the unmodified file as "fixed" looks confusing at best.

scripts/sound-dat-repair.py
```diff
@@ -0,0 +1,187 @@
...
+
+        if chunk_id == b'data':
+            data_found = True
+            true_end_offset = chunk_data_start + chunk_size
+            break
+
+        offset = chunk_data_start + chunk_size
+
+    if not data_found:
+        return raw_wav
```
Alberth289346: Print here a message or error as well, I guess.

scripts/sound-dat-repair.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # Correct the RIFF header length field at byte offset 4
+    new_riff_size = len(cleaned_wav) - 8
+    cleaned_wav[4:8] = struct.pack('<I', new_riff_size)
+
+    return bytes(cleaned_wav)
+
+
+def process_dat(input_dat_path: str, output_dat_path: str):
+    if not os.path.exists(input_dat_path):
```
Alberth289346:
```suggestion
    if not os.path.isfile(input_dat_path):
```
Slightly stricter check, may not work for soft-links though iirc.

scripts/sound-dat-repair.py
```diff
@@ -0,0 +1,187 @@
...
+    if file_size < 4 + ARCHIVE_HEADER_SIZE:
+        print("Error: File is too small to be a valid sound archive.")
+        sys.exit(1)
+
+    with open(input_dat_path, 'rb') as f:
+        # 1. Read last 4 bytes -> headerPosition
+        f.seek(-4, os.SEEK_END)
+        header_position = struct.unpack('<I', f.read(4))[0]
+
+        # CorsixTH safety check
```
Alberth289346: Euhm, are we doing something special to this file? (ie, just 'Safety check' ?)

scripts/sound-dat-repair.py
Alberth289346: Sorry, it has 2 arguments, so `!= 3` here rather than `4`,

scripts/sound-dat-repair.py
```diff
@@ -0,0 +1,187 @@
...
+
+    print(f"Successfully repacked archive to '{output_dat_path}'.")
+
+
+if __name__ == '__main__':
+    if len(sys.argv) < 3:
+        print("Usage: python sound-dat-repair.py <input.dat> <output.dat>")
+        sys.exit(1)
+
+    process_dat(sys.argv[1], sys.argv[2])
```
Alberth289346, line 196: Could add a `sys.exit(0)` but probably not needed.

scripts/sound-dat-repair.py
Alberth289346: There are more error messages with `sys.exit(1)`, could add `, file=stderrr` there to print the text at `stderr`.

Alberth289346 (commented):
Could use a bit of love

TheCycoONE:
Much more love than I intend to give it. Closing the PR unless someone else wants to take it over.

lewri:
> Much more love than I intend to give it. Closing the PR unless someone else wants to take it over.

If you want to push the commit to my scripts repo (lewri/CorsixTH-Scripts) you can, saves binning it outright.

Alberth289346:
Updated the script.

TheCycoONE:
I thought it was better to keep a corrupted or non-wav file entry as is and continue with the rest. CorsixTH will simply not play a bad file so as long as all the indexes are kept the same and the user it warned that seemed more useful.

So instead of exit(1), I'd just print a warning to stderr that the track wasn't processed and return the raw so it can be injected back. That's what I was doing before except that I wasn't informing the user.

TheCycoONE (commented):
These changes all look good to me. Thanks @Alberth289346

ARGAMX (approved):
Thank you guys! 👍

PR #3474, opened 2026-07-30
Health Inspector must give epidemic results

### LINKED ISSUE #3336: [Bug] Health Inspector meander endlessly
### Describe the issue

When the health inspector arrives after an epidemic, if you move the reception desk the health isnspector walks around in a few steps but never walks to the reception desk or leave your hospital

### Steps to Reproduce

1. Health inspector arrives after an epedemic

2. Move the reception before he reaches it

3.

### Expected Behaviour

Reroutes to new location

### CorsixTH Version

0.69.2

### Operating System

Windows 11

### Gamelog.txt

```yaml
Gamelog generated on 2024-04-07 21:35:23
Compiled with api_version: 2684, arch: x64, audio: true, os: windows
SDL renderer: direct3d
Lua 5.4 run with api version: 2684, game version: v0.67, savegame version: 180
Created game with savegame version 180.

Error in timer handler: 
Lua\entities\humanoid.lua:772: bad argument #1 to 'getCellTemperature' (number expected, got nil)
stack traceback:
	[C]: in method 'getCellTemperature'
	Lua\entities\humanoid.lua:772: in field 'tickDay'
	Lua\entities\humanoids\patient.lua:584: in method 'tickDay'
	Lua\world.lua:1137: in method 'onEndDay'
	Lua\world.lua:1037: in method 'onTick'
	Lua\app.lua:1202: in function <Lua\app.lua:1199>
	(...tail calls...)
	Lua\app.lua:1102: in function <Lua\app.lua:1097>
```

tobylane:
*Fixes #3336*

- When a health inspector is checking reports of an epidemic, they will wait five days in a hospital without a reception desk before sending the report in anyway and leaving. This wait is arbitrary, but similar to when a VIP visits a hospital without any rooms, they leave after five days.

Original behaviour found by Argamx
> when manipulating by movement and deleting the reception desk, there was a loophole that allowed undeserved epidemic victories. Essentially, it was cheating.
I think this was a bug in the original, and I'm not sure we need to reproduce it.

Previous Corsix behaviour was for the Inspector to stay at the hospital forever and never deliver the report.

### DIFF
--- a/CorsixTH/Lua/entities/humanoids/inspector.lua
+++ b/CorsixTH/Lua/entities/humanoids/inspector.lua
@@ -66,6 +66,26 @@ function Inspector:announce()
   self.world.ui:playAnnouncement("vip008.wav", AnnouncementPriority.High)
 end
 
+function Inspector:tick()
+  Humanoid.tick(self)
+  self:checkForReceptionDesk()
+end
+
+--! React to reception desk status
+function Inspector:checkForReceptionDesk()
+  if self.going_home then return end
+  -- When the inspector realises the hospital does not have a reception desk or
+  -- the desk they were aiming for is gone, they will either:
+  if self:isMeandering() then
+    if self.hospital:hasReceptionDesk(true) then
+      self:setNextAction(SeekReceptionAction()) -- Head towards an existing desk
+    else
+      self.hospital.epidemic:handleInspectorArrival() -- Deliver results anyway
+      self:goHome()
+    end
+  end
+end
+
 function Inspector:afterLoad(old, new)
   if old < 213 then
     self.mood_marker = 2
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -1196,3 +1196,18 @@ function Humanoid:isObscuringArea(x1, x2, y1, y2)
   end
   return false
 end
+
+--! Check whether humanoids (staff and Inspector) are meandering
+--!return true if they currently has a meander action
+function Humanoid:isMeandering()
+  if #self.action_queue < 2 then return false end
+
+  -- "meander" action always insert "move" or "idle" action before itself.
+  -- so when humanoid "meandering" his action queue usually looks like:
+  -- [1 idle, 2 meander] or [1 walk, 2 meander].
+  local idle_is_first = self.action_queue[1].name == "idle"
+  local walk_is_first = self.action_queue[1].name == "walk"
+  local meander_is_second = self.action_queue[2].name == "meander"
+
+  return (idle_is_first or walk_is_first) and meander_is_second
+end
--- a/CorsixTH/Lua/entities/humanoids/staff.lua
+++ b/CorsixTH/Lua/entities/humanoids/staff.lua
@@ -509,21 +509,6 @@ function Staff:adviseWrongPersonForThisRoom()
   end
 end
 
---! Check whether staff are meandering
---!return true if staff currently has a meander action
-function Staff:isMeandering()
-  if #self.action_queue < 2 then return false end
-
-  -- "meander" action always insert "move" or "idle" action before itself.
-  -- so when humanoid "meandering" his action queue usually looks like:
-  -- [1 idle, 2 meander] or [1 walk, 2 meander].
-  local idle_is_first = self.action_queue[1].name == "idle"
-  local walk_is_first = self.action_queue[1].name == "walk"
-  local meander_is_second = self.action_queue[2].name == "meander"
-
-  return (idle_is_first or walk_is_first) and meander_is_second
-end
-
 -- Function to decide if staff currently has nothing to do and can be called to a room where they're needed
 function Staff:isIdle()
   -- Make sure we're not in an undesired state
--- a/CorsixTH/Lua/epidemic.lua
+++ b/CorsixTH/Lua/epidemic.lua
@@ -419,8 +419,8 @@ function Epidemic:finishCoverUp()
   self:turnOffVaccinationMode()
 end
 
---[[ Inspector had arrived at reception desk. Check if
-any infected still in hospital and determine final verdict.]]
+--[[ Inspector had arrived at reception desk, or a hospital without a desk.
+Check if any infected still in hospital and determine final verdict.]]
 function Epidemic:handleInspectorArrival()
   local still_infected = self:countInfectedPatients()
   self:determineFaxAndFines(still_infected)


lewri (approved):

ARGAMX:
I tried these changes, but defect 3336 still reproduces for me according to the Steps to Reproduce I described. The inspector wanders endlessly.

tobylane:
Good catch. This now deals with Inspectors who were never procrastinating in the first place.

CorsixTH/Lua/entities/humanoids/inspector.lua
```diff
@@ -66,6 +66,21 @@ function Inspector:announce()
...
 
+function Inspector:tickDay()
+  if not self.going_home then
+    if self.waiting then
+      self.waiting = self.waiting - 1
+      if self.waiting == 0 then -- There was not a reception desk for five days
+        self.hospital.epidemic:handleInspectorArrival() -- Deliver results anyway
+        self:goHome()
+      end
+    elseif self.hospital:isInHospital(self.tile_x, self.tile_y) then
```
lewri: Isn't there a chance this could fire prematurely? E.g. they walk through another building first to get to the reception desk; or the reception desk is > 5 days away from the entrance?

tobylane: Yes, now they must head towards the intended desk, find it's not there and meander, then try to find a desk. If there still isn't one, then they count down five days.

ARGAMX:
> Good catch. This now deals with Inspectors who were never procrastinating in the first place.

Now the situation is as follows:

If I grab the reception desk and continue holding it, the Inspector wanders around endlessly. Although the task didn't mention anything about this. But this indicates that the "waiting" part essentially does nothing. That is, "self.hospital.epidemic:handleInspectorArrival() -- Deliver results anyway" is never called. At least, I couldn't get that to work.

If I grab the reception desk and place it to a different location, Inspector wanders around for a while and then builds a path to the new reception desk. That's better.

I wonder if we can make Inspector not to wander around and then immediately go to the reception desk? I just don't understand why it needs to wander aimlessly for a while in this case? Is this can be improved?

tobylane:
The five days wait is to allow for moving the desk, or even briefly extending the time to cure patients. Once the desk is placed, the Inspector's desk search is in the next tickDay, it could be the next tick if for a quicker reaction.

ARGAMX:
> The five days wait is to allow for moving the desk, or even briefly extending the time to cure patients. Once the desk is placed, the Inspector's desk search is in the next tickDay, it could be the next tick if for a quicker reaction.

How to make the added line `self.hospital.epidemic:handleInspectorArrival()` to fire?

CorsixTH/Lua/entities/humanoids/inspector.lua
```diff
@@ -66,6 +66,24 @@ function Inspector:announce()
   self.world.ui:playAnnouncement("vip008.wav", AnnouncementPriority.High)
 end
 
+function Inspector:tickDay()
+  if self.going_home then return end
+  if self.waiting then -- Aware that there is no desk where it used to be
+    self:setNextAction(SeekReceptionAction())
```
lewri, line 72: Does this not endlessly spam a SeekReceptionAction each day even if we are actively seeking?

Alberth289346: For 5 days, that may be acceptable.

tobylane: The walk is to a specific place, but the Reception action isn't, so they will stick to the first desk they choose to walk to. As part of picking the best desk, waiting is set to nil.

I tried out some further changes, like putting this setNextAction in Inspector:tick, but it doesn't seem better.

ARGAMX (dismissed):
I have a several notes on this PR.
The very first of them is to figure out how `self.waiting` is used and under what conditions `self.hospital.epidemic:handleInspectorArrival()` might be triggered.
I couldn't get this to work.

I also believe that the `tickDay()` function is very general and should call other functions with clearly names rather than doing other functions' specialised work.

tobylane:
> The very first of them is to figure out how self.waiting is used and under what conditions

It is set in seek reception action, after a desk has not been found. https://github.com/CorsixTH/CorsixTH/blob/f558f43f5ce34b3635f3bd269bd22e4b0fb14275/CorsixTH/Lua/humanoid_actions/seek_reception.lua#L126
This specific value is used in a few places, I took inspiration from `VIP:tickDay`, where it's used to end a visit to an unbuilt hospital. https://github.com/CorsixTH/CorsixTH/blob/f558f43f5ce34b3635f3bd269bd22e4b0fb14275/CorsixTH/Lua/entities/humanoids/vip.lua#L96

> under what conditions self.hospital.epidemic:handleInspectorArrival() might be triggered.

In `VIP:tickDay`, they react to entering a hospital with no desk and waiting five days. The Inspector goes beyond this and tries to find the new desk repeatedly on their way to any desk. I tried causing problems of the Inspector changing which desk they go to, but it's always the best, a consistent choice.

> I also believe that the tickDay() function is very general and should call other functions

 `Humanoid:tickDay` is what might be called, but as that deals with temperature and mood, Inspector doesn't. I don't know why VIP does, that line has been in since VIPs were added so it may have had a purpose.

> I couldn't get this to work.

The Inspector overstayed?

ARGAMX:
> > The very first of them is to figure out how self.waiting is used and under what conditions
> 
> It is set in seek reception action, after a desk has not been found.
> 
> https://github.com/CorsixTH/CorsixTH/blob/f558f43f5ce34b3635f3bd269bd22e4b0fb14275/CorsixTH/Lua/humanoid_actions/seek_reception.lua#L126
> 
> 
> This specific value is used in a few places, I took inspiration from `VIP:tickDay`, where it's used to end a visit to an unbuilt hospital.
> https://github.com/CorsixTH/CorsixTH/blob/f558f43f5ce34b3635f3bd269bd22e4b0fb14275/CorsixTH/Lua/entities/humanoids/vip.lua#L96
> 
> > under what conditions self.hospital.epidemic:handleInspectorArrival() might be triggered.
> 
> In `VIP:tickDay`, they react to entering a hospital with no desk and waiting five days. The Inspector goes beyond this and tries to find the new desk repeatedly on their way to any desk. I tried causing problems of the Inspector changing which desk they go to, but it's always the best, a consistent choice.
> 
> > I also believe that the tickDay() function is very general and should call other functions
> 
> `Humanoid:tickDay` is what might be called, but as that deals with temperature and mood, Inspector doesn't. I don't know why VIP does, that line has been in since VIPs were added so it may have had a purpose.

Okay, I see, thank you.

> The Inspector overstayed?

Yes.

Steps to Reproduce:

1. Inside
`function Inspector:tickDay()`
I added debug code:
```Lua
print("Inspector:tickDay() | waiting: ", self.waiting, (self.going_home and "going_home" or "not going_home"), ((self.action_queue[2] and self.action_queue[2].name == "meander") and "meander" or "not meander") )
```
2. I loaded save `3336-0.69.2-epidemic health inspector near reception.sav` from 3336.
3. I picked up the reception desk and press 5 to remove pause.
4. Here are the logs I got from 8 in-game days:
```Lua
Inspector:tickDay() | waiting: 4 not going_home meander
Inspector:tickDay() | waiting: 4 not going_home meander
Inspector:tickDay() | waiting: 4 not going_home meander
Inspector:tickDay() | waiting: 4 not going_home meander
Inspector:tickDay() | waiting: 4 not going_home meander
Inspector:tickDay() | waiting: 4 not going_home meander
Inspector:tickDay() | waiting: 4 not going_home meander
Inspector:tickDay() | waiting: 4 not going_home meander
```
5. Thus `self.waiting == 0` never occurs and `self.hospital.epidemic:handleInspectorArrival()` is never called.

But even if it worked, I still don't like the idea.
It means the player can move the reception desk many times and get extra time an infinite number of times.

Can we just call `self.hospital.epidemic:handleInspectorArrival()` right away if the inspector can't find the table?

tobylane:
> It means the player can move the reception desk many times and get extra time an infinite number of times.
Can we just call self.hospital.epidemic:handleInspectorArrival() right away if the inspector can't find the table?

Calling that as soon as they start meandering looks a bit harsh, as they don't even pause to look for a desk moved one block. Calling that as soon as they are meandering in a hospital without desks is what I've chosen, though it doesn't give any leeway for moving a desk while unpaused.

ARGAMX:
> > It means the player can move the reception desk many times and get extra time an infinite number of times.
> > Can we just call self.hospital.epidemic:handleInspectorArrival() right away if the inspector can't find the table?
> 
> Calling that as soon as they start meandering looks a bit harsh, as they don't even pause to look for a desk moved one block. Calling that as soon as they are meandering in a hospital without desks is what I've chosen, though it doesn't give any leeway for moving a desk while unpaused.

Great, it works much better now! Thanks

If I remove the last reception desk, `self.hospital.epidemic:handleInspectorArrival()` is called. This didn't happen before, and the epidemic would hang forever.
If I move a desk to a new place while paused, the inspector reroutes to the new location. This didn't happen before, and the epidemic would hang forever.

Although, if I move the reception desk while paused multiple times, I can still make VIP walk indefinitely. But frankly, that's beyond the scope of this PR and Issue.

Overall, I'm happy with the functionality and I believe this is the best approach to resolve this issue.
However, I do have a few more comments regarding the code organisation.

CorsixTH/Lua/entities/humanoids/inspector.lua
```diff
@@ -66,6 +66,20 @@ function Inspector:announce()
   self.world.ui:playAnnouncement("vip008.wav", AnnouncementPriority.High)
 end
 
+function Inspector:tick()
+  Humanoid.tick(self)
+  if self.going_home then return end
+  -- When the inspector realises the desk they were aiming for is gone, they will
+  if self.action_queue[2] and self.action_queue[2].name == "meander" then
```
ARGAMX: We already have an `isMeandering()` function.
And I think we just can reuse it here.

But there's a catch: `isMeandering()` defined in the `Staff` class: `function Staff:isMeandering()`
So, why don't we move this function from `Staff` to `Humanoid` class, and for the `Inspector`, we'll just do `self.isMeandering()`?

CorsixTH/Lua/entities/humanoids/inspector.lua
```diff
@@ -66,6 +66,20 @@ function Inspector:announce()
...
+    if self.hospital:hasReceptionDesk(true) then
+      self:setNextAction(SeekReceptionAction()) -- Head to an existing desk
+    else
+      self.hospital.epidemic:handleInspectorArrival() -- Deliver results anyway
+      self:goHome()
+    end
+  end
+end
```
ARGAMX, line 87: The `tick()`, `tickDay()`, and `onSomeAction()` functions are fairly "general".

This is clear from their names. The name tells you when the function is called. This is very convenient for understanding when a function is called.
However, the downside is that the name doesn't tell you what the function does.
And digging inside a function, reading the code, and figuring out what it does can be quite mentally taxing.
But to solve this problem, it's recommended to use general functions only as calls to other, more specialised functions.
The result is a very convenient design:
We know when the general function is called, and it's very easy to read what it does.

Check `function Epidemic:tick()` for reference:

```Lua
function Epidemic:tick()
  self:infectOtherPatients()
  self:checkIfReadyToReveal()
  self:showAppropriateAdviceMessages()
  self:tryAnnounceInspector()
  self:markedPatientsCallForVaccination()
  self:checkInfectedLeftHospital()
  self:checkNoInfectedPatients()
  self:checkPatientsForRemoval()
end
```

Isn't that beautiful?
Now imagine that instead of these calls, the code for all these functions, each with its own separate goal and meaning, would be inside this `tick()` function?

So, can we move all this code into a separate function with an appropriate name?
And in `Inspector:tick()` we'll keep only `Humanoid.tick(self)` and the call to this new function?
I believe Readability will increase exponentially.

tobylane: That's good, done both.

CorsixTH/Lua/entities/humanoids/inspector.lua
```diff
@@ -66,6 +66,26 @@ function Inspector:announce()
   self.world.ui:playAnnouncement("vip008.wav", AnnouncementPriority.High)
 end
 
+function Inspector:tick()
+  Humanoid.tick(self)
+  self:checkForRecepDesk()
+end
+
+-- React to reception desk status
+function Inspector:checkForRecepDesk()
```
lewri: I'd recommend using the full name, not sure what the abbreviated version is saving us here

CorsixTH/Lua/entities/humanoids/inspector.lua
```diff
@@ -66,6 +66,26 @@ function Inspector:announce()
...
+  if self.going_home then return end
+  -- When the inspector realises the hospital does not have a reception desk or
+  -- the desk they were aiming for is gone, they will
```
lewri: A nit of nits but that end should really be `will:` or `will either:`

CorsixTH/Lua/entities/humanoids/inspector.lua
```diff
@@ -66,6 +66,26 @@ function Inspector:announce()
   self.world.ui:playAnnouncement("vip008.wav", AnnouncementPriority.High)
 end
 
+function Inspector:tick()
+  Humanoid.tick(self)
+  self:checkForRecepDesk()
+end
+
+-- React to reception desk status
```
lewri:
```suggestion
--! React to reception desk status
```

lewri (commented):
Nit territory 

PR description needs an update too.

ARGAMX (approved):
Good to me 👍
Thanks

PR #3476, opened 2026-07-31
Add a settings dialog builder

tobylane:
- Add a dialog builder to make the settings menus with more preset sizes of elements. As much as possible depends on the name of the setting.
- Use this on the customise dialog (minus the volume casebook setting)
- Update language files with the new string names based on the setting name

Before and after.
before
after


TheCycoONE:
@tobylane are you able to add a second dialog to prove the generalization?

tobylane:
folders before


Before
folders after


After


The folders dialog, with some additions to resizable.lua that work for the customise dialog. I removed the current path from the tooltip, not sure on that change.
https://github.com/CorsixTH/CorsixTH/compare/master...tobylane:CorsixTH:refs/heads/foldersmenu

ARGAMX:
> folders before
> Before folders after
> 
> After
> 

Why did the blocks in which the path is displayed have a different shadow now?
Before the change, they were convex up, like buttons, and it is logical that this meant that they could be pressed as buttons.
Now they are bulging down, as if they are already pressed.

tobylane:
That's easily done, just `raised = true`. But I will try to do the tooltips better, maybe bring the current path back.

Screenshot 2026-08-04 at 21 20 15

TheCycoONE:
Ok. I see this growing into a lua table based declarative UI language (like a greatly simplified QML or XAML) as more dialogs get consumed (e.g. the front options menu would require several more primitives).

It's certainly interesting.

TheCycoONE (approved):
Interesting, curious where this goes. Would like the ability to add scrolling lists of options; maybe we unify the look and feel of our settings windows.

tobylane:
https://github.com/CorsixTH/CorsixTH/compare/master...tobylane:CorsixTH:refs/heads/optionsmenu

This is the options menu converted, with a minor fix to the dropdowns so they close on a second click. All the resizable.lua changes are in the first commit, the other languages commit doesn't cover folders or options menus yet.

A scrolling list instead of two columns?

TheCycoONE:
@tobylane are you about ready to merge this one? I'm waiting for it before any attempt at splitting off a new Display options dialog.

tobylane:
I had got on with the folders and audio menu but waited in case your redesign was in progress. 

https://github.com/tobylane/CorsixTH/tree/refs/heads/audiomenu Sound settings on top of the other menus, but not your most recent SDL3 PR, and the other languages have only been adjusted for the Customise dialog.

This push is the same three commits, plus all of the changes to resizable.lua from all the menus, rebased on master.

tobylane:
Added the options menu here for progress.

PR #3480, opened 2026-08-04
Updating PT-BR Translation

rdfortega:
The PT-BR translation had some accentuation issues from the character ~ being replaced by ¨ on characters so ã became ä.


tobylane:
The characters you use aren't in the original Theme Hospital font, so they come out as ?. A previous translator may have chosen from the reduced range. You can switch to unicode, but it doesn't look as good. Before, with your change, and with Unicode in screenshots, the third word of the third line is one you change.

Screenshot 2026-08-04 at 21 23 43
Screenshot 2026-08-04 at 21 27 25
Screenshot 2026-08-04 at 21 28 26

ARGAMX (commented):
I think we need to write a large, noticeable warning about this in the language file.
And not only in this language file.
This is not the first time when guys get to work and add or change accentuation, letter accents and other unsupportable diacritical signs.
And then it turns out that the work was done in vain.

lewri:
@rdfortega for some of our languages if a certain accented character is missing the translator will use the regular letter instead e.g. `e` instead of `ë`. I'm guessing the previous translator chose the closest approximate accent instead. 

Are any of your changes to the translation still relevant outside the accents?

lewri:
To be reopened if any translation changes apply not related to accents (or this languages wants to transition to unicode).

PR #3481, opened 2026-08-05
Separate window size from render resolution

TheCycoONE:
This PR prepares for SDL3 HiDPI support for supporting operating systems and window managers where the logical window size does not match actual pixel resolution, such as MacOS and Wayland.

The biggest user visible change here is that the resolution setting no longer dictates the full screen resolution - I didn't see any way of aligning the logical/physical pixel concepts and intentionally selecting a lower resolution. The aspect ratio is respected in this patch; but I'm left wondering if our config is really ideal.

Reading the patches:
The first three commits are simple cleanup/refactor that could be applied without any intention to support HiDPI; removing unnecessary rendering complexity.

The forth commit is by far the biggest, that is where I separate the idea of window size from render pixels in anticipation of HiDPI support. It has one user visible change which is to always apply full screen at (logical) desktop resolution.

The fifth commit is to address some minor flicker and more importantly an SDL issue running on MacOS where a window may not be given focus.

The final commit is to ensure that everything remains in sync on updates as it would have with SDL2. It may not be necessary, but felt safer.

Out of scope:
1. The constraints around the ui_scale option. We went pretty far (too far?) preventing the user from selecting values that didn't make sense for their window size; but now that the window size isn't the actual render size the settings are over constrained. I'm not exactly sure yet how we are going to get out of that one. It either involves automatically/always applying a hidden ui scale factor on top of the user selected scale factor when physical pixels are not the same as window dimensions OR allowing the user the freedom to pick window sizes and ui_scales that would not work together and clamping the size at rendering time instead.

2. A thought I had is to have a 4:3 checkbox option visible in both full screen mode which adds the letter boxing on wide screen monitors. I would then hide the window size option and show the aspect ratio option when full screen was checked. There is NO reason anyone would want an aspect ratio other than their native or 4:3 (original game).

3. Enabling HiDPI. I did not want to regress the experience on MacOS (or Wayland) until we sorted out the ui_scale options.

### DIFF
--- a/CorsixTH/Lua/dialogs/resizables/file_browser.lua
+++ b/CorsixTH/Lua/dialogs/resizables/file_browser.lua
@@ -78,9 +78,10 @@ function FilteredTreeControl:FilteredTreeControl(root, x, y, width, height, col_
   self.num_rows = (self.tree_rect.h - self.y_offset) / self.row_height
 
   -- Magic numbers used to find a static position across different screen resolutions.
-  local button1x = math.floor(TheApp.ui.app.config.width / 2 - 90)
+  local scr_w, scr_h = TheApp.video:getRenderSize()
+  local button1x = math.floor(scr_w / 2 - 90)
   local button2x = button1x + 210
-  local buttony = math.floor(TheApp.ui.app.config.height / 4 - 95)
+  local buttony = math.floor(scr_h / 4 - 95)
   -- Add the two column headers and make buttons on them.
   if show_dates then
     self:addBevelPanel(1, 1, width - 170, 13, col_bg):setLabel(_S.menu_list_window.name)
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -371,8 +371,6 @@ render_target::scoped_target_texture::scoped_target_texture(
   }
 
   // Clear the new texture to transparent/black.
-  SDL_SetRenderLogicalPresentation(target->renderer, rect.w, rect.h,
-                                   SDL_LOGICAL_PRESENTATION_LETTERBOX);
   SDL_SetRenderDrawColor(target->renderer, 0, 0, 0, SDL_ALPHA_TRANSPARENT);
   SDL_RenderClear(target->renderer);
   target->current_target = this;
@@ -399,11 +397,6 @@ render_target::scoped_target_texture::~scoped_target_texture() {
     std::fprintf(stderr, "scoped_target_texture destructor error: %s",
                  SDL_GetError());
   }
-  SDL_SetRenderLogicalPresentation(
-      target->renderer,
-      previous_target ? previous_target->rect.w : target->width,
-      previous_target ? previous_target->rect.h : target->height,
-      SDL_LOGICAL_PRESENTATION_LETTERBOX);
   SDL_SetTextureBlendMode(texture, SDL_BLENDMODE_BLEND);
   target->current_target = previous_target;
   if (scale) {
--- a/CorsixTH/Src/th_gfx_sdl.cpp
+++ b/CorsixTH/Src/th_gfx_sdl.cpp
@@ -188,6 +188,66 @@ class scoped_color_mod {
   SDL_Texture* texture;
 };
 
+/**
+ * Applies or removes letterboxing while preserving 1:1 pixel output
+ * with the renderer for crisp output.
+ *
+ * \param renderer - The SDL_Renderer to apply to
+ * \param fullscreen - Whether the window is fullscreen. In window mode it is
+ *                     assumed that the aspect ratio matches the window and no
+ *                     letterbox is needed.
+ * \param ar_w - The aspect ratio width, e.g. 16 in 16:9, or 4 in 4:3. Can use
+ *               the target resolution width.
+ * \param ar_h - The aspect ratio height, e.g. 9 in 16:9, or 3 in 4:3. Can use
+ *               the target resolution height.
+ */
+void apply_letterbox(SDL_Renderer* renderer, bool fullscreen, int ar_w,
+                     int ar_h) {
+  // If not fullscreen we can assume a match and exit early
+  SDL_SetRenderViewport(renderer, nullptr);
+  if (!fullscreen) {
+    return;
+  }
+
+  int w;
+  int h;
+  // Get the true render size (without any letterbox/logical voodoo)
+  SDL_GetRenderOutputSize(renderer, &w, &h);
+
+  float real_ar = static_cast<float>(w) / static_cast<float>(h);
+  float target_ar = static_cast<float>(ar_w) / static_cast<float>(ar_h);
+
+  // Common target aspect ratios are 4:3, 16:10, and 16:9 which work out to
+  // 1.333, 1.6, and 1.777
+
+  // If we are closer than 0.01 to the target aspect ratio, assume we
+  // meant to match.
+  if (std::abs(real_ar - target_ar) < 0.01) {
+    return;
+  }
+
+  int target_w;
+  int target_h;
+
+  // Otherwise determine if we are letterboxing vertically or horizontally to
+  // preserve the target aspect ratio.
+
+  // If the real aspect ratio is greater than the target aspect ratio it means
+  // the screen is wider than the target - we want vertical bars.
+  // Otherwise, we want horizontal bars.
+  if (real_ar > target_ar) {
+    target_w = static_cast<int>(h * target_ar);
+    target_h = h;
+  } else {
+    target_w = w;
+    target_h = static_cast<int>(w / target_ar);
+  }
+
+  SDL_Rect viewport = {(w - target_w) / 2, (h - target_h) / 2, target_w,
+                       target_h};
+  SDL_SetRenderViewport(renderer, &viewport);
+}
+
 }  // namespace
 
 palette::palette(const uint8_t* pData, size_t iDataLength, bool is8bit) {
@@ -415,16 +475,16 @@ render_target::scoped_target_texture::~scoped_target_texture() {
 }
 
 render_target::render_target(const render_target_creation_params& params)
-    : width{params.width},
-      height{params.height},
-      direct_zoom{params.direct_zoom} {
+    : direct_zoom{params.direct_zoom} {
   pixel_format = SDL_GetPixelFormatDetails(SDL_PIXELFORMAT_ABGR8888);
 
   SDL_PropertiesID winProps = SDL_CreateProperties();
   SDL_SetStringProperty(winProps, SDL_PROP_WINDOW_CREATE_TITLE_STRING,
                         "CorsixTH");
-  SDL_SetNumberProperty(winProps, SDL_PROP_WINDOW_CREATE_WIDTH_NUMBER, width);
-  SDL_SetNumberProperty(winProps, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, height);
+  SDL_SetNumberProperty(winProps, SDL_PROP_WINDOW_CREATE_WIDTH_NUMBER,
+                        params.width);
+  SDL_SetNumberProperty(winProps, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER,
+                        params.height);
   SDL_SetNumberProperty(winProps, SDL_PROP_WINDOW_CREATE_FLAGS_NUMBER,
                         SDL_WINDOW_RESIZABLE);
   SDL_SetBooleanProperty(winProps, SDL_PROP_WINDOW_CREATE_FULLSCREEN_BOOLEAN,
@@ -459,10 +519,7 @@ render_target::render_target(const render_target_creation_params& params)
 
   SDL_SetWindowMinimumSize(window, params.min_width, params.min_height);
 
-  SDL_RendererLogicalPresentation rlp = params.fullscreen
-                                            ? SDL_LOGICAL_PRESENTATION_LETTERBOX
-                                            : SDL_LOGICAL_PRESENTATION_DISABLED;
-  SDL_SetRenderLogicalPresentation(renderer, width, height, rlp);
+  apply_letterbox(renderer, params.fullscreen, params.width, params.height);
 
   // Workaround for https://github.com/libsdl-org/SDL/issues/13920 on MacOS
   SDL_Event evt;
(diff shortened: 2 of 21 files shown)


CorsixTH/Src/th_gfx_sdl.cpp
```diff
@@ -188,6 +188,68 @@ class scoped_color_mod {
...
+ * \param renderer - The SDL_Renderer to apply to
+ * \param fullscreen - Whether the window is fullscreen. In window mode it is
+ *                     assumed that the aspect ratio matches the window and no
+ *                     letterbox is needed.
+ * \param ar_w - The aspect ratio width, e.g. 16 in 16:9, or 4 in 4:3. Can use
+ *               the target resolution width.
+ * \param ar_h - The aspect ratio height, e.g. 9 in 16:9, or 3 in 4:3. Can use
+ *               the target resolution height.
+ */
+void apply_logical_presentation(SDL_Renderer* renderer, bool fullscreen,
```
TheCycoONE: There is an argument to be made that we should be using https://wiki.libsdl.org/SDL3/SDL_SetRenderViewport instead of https://wiki.libsdl.org/SDL3/SDL_SetRenderLogicalPresentation. The benefit would be that we would be guaranteed no scaling, the down side is that we would need to do a little more work (blank the background, properly place the viewport, and constrain the width/height where https://wiki.libsdl.org/SDL3/SDL_GetCurrentRenderOutputSize is being used.

TheCycoONE:
Moving back to WIP, as I see some clipping issues.

ARGAMX:
> 2. A thought I had is to have a 4:3 checkbox option visible in both full screen mode which adds the letter boxing on wide screen monitors. I would then hide the window size option and show the aspect ratio option when full screen was checked. There is NO reason anyone would want an aspect ratio other than their native or 4:3 (original game).

Wisely. It makes sense. Agree.

> 1. The constraints around the ui_scale option. We went pretty far (too far?) preventing the user from selecting values that didn't make sense for their window size; but now that the window size isn't the actual render size the settings are over constrained. I'm not exactly sure yet how we are going to get out of that one. It either involves automatically/always applying a hidden ui scale factor on top of the user selected scale factor when physical pixels are not the same as window dimensions OR allowing the user the freedom to pick window sizes and ui_scales that would not work together and clamping the size at rendering time instead.

In my dreams, I imagined that it would work according to this logic:

Automatically/always applying a hidden ui scale factor on top of the user selected scale factor when physical pixels are not the same as window dimensions.

ARGAMX:
1. 
I had run game in 800x600 100%. Then I set 1280x960 and 200%. And window fail to keep center.

Screenshot 2026-08-05 at 13 24 39

ARGAMX:
2.

When I run game from Terminal clicks don't work.

This issue come back:
https://github.com/CorsixTH/CorsixTH/pull/3436#issuecomment-4989326901

I have to hold Command key all the way.

ARGAMX:
3.

I am glad that 800x600 in 0.70.0 has the same window size as 800x600 in this PR.
BUT the size of the interface has become 2 times smaller.
I want to increase it, but the 200% option is not available.
That is, the interface has simply become 2x smaller compared to 0.70.0.

800x600 100% 0.70.0:

Screenshot 2026-08-05 at 13 41 16

------------------------

800x600 100% this PR:

Screenshot 2026-08-05 at 13 41 25

ARGAMX:
4.

Maybe it's time to scale the cursor.
Since everything is 100% shallow and nothing is visible, I had to turn on 200%.
But to turn on 200%, I need to set a minimum of 1280x960.
But now the cursor is very small.
To check this out try to find cursor on the screen on the following screenshot...

Screenshot 2026-08-05 at 13 55 16

TheCycoONE:
1. Ok I'll look into this
2. The workaround is still in place. Obviously there is more to it. Love to have this fixed upstream.
3. I mentioned in the description. Adjusting the ui_scale will be another PR. Was trying to get discussed earlier but I wasn't getting any feedback, and auto adjusting will be a lot of changes that will make this one harder to review.
4. Agreed but out of scope of this PR. If we go with autoscale on the UI we would do the same.

TheCycoONE:
@ARGAMX the MacOS focus issue is an upstream race condition. It's pretty tricky, but I threw in some more processing before raising the window, so that might help. At least our users will generally not hit it.

TheCycoONE:
For 1, do you mean the window isn't displayed in the center of the screen or the content isn't centered in the Window? I'm not seeing either issue...

TheCycoONE:
```
width = 1980
height = 960
ui_scale = 2
```

image

TheCycoONE:
Disregard, I was able to reproduce. Doesn't happen on a new game, just when changing the ui scale in the settings menu.

1. Set config as follows:
```
width=800
height=600
ui_scale=1
```

2. Launch CorsixTH and go to Options
3. Change resolution to something higher like 1920x1080 (so far ok)
4. Set UI Scale to 200%
(observe options window off center)

5. Close game
6. Open game (observe everything centered again)

TheCycoONE:
@ARGAMX Fixed 1, maybe 2 again?, 3 and 4 are out of scope for this PR, but in scope for 0.71.

ARGAMX:
1. Window fail to keep center - thanks, resolved 👍
2. Clicks don't work. - thanks, resolved 👍

ARGAMX:
5. 
Immediately after launching the game, one song starts playing, then after about 0.1-0.2 seconds, it switches to another track. Original in-game music.
This doesn't happen every time I launch the game, but it happens more than 50% of the time.

ARGAMX:
6.

Defect when switching resolution:

1. Set config as follows:
```
width=1280
height=960
ui_scale=2
```
2. Launch CorsixTH and go to Options
3. Set UI Scale to 100%
4. Change resolution to 800x600

Got:
Game window still 1280x960.

Expected:
800x600 window resolution.

Select 800x600 again - now it changed size to 800x600 - ok.

TheCycoONE:
> Immediately after launching the game, one song starts playing, then after about 0.1-0.2 seconds, it switches to another track. Original in-game music. This doesn't happen every time I launch the game, but it happens more than 50% of the time.

Yes, this was an SDL2 workaround for a delay (the same delay that is now between every track) on Windows. Not tied to this HiDPI work, worth ticketing, or maybe adding to the delay tracking ticket. Obviously the workaround is useless now and causing problems.

TheCycoONE:
6 is caused by this PR, will look into it.

TheCycoONE:
Fixed 6 (which I broke in my fix for 1)

ARGAMX:
6. Defect when switching resolution - thanks, resolved 👍

ARGAMX (approved):
I am still missing point 3.
Without it, it works not so good.
But since you say it will be in another PR, then we'll wait for that PR.
I have no further requests about this PR specifically.
Thanks, Well done 👍

TheCycoONE:
One thing I can do is revert the actual HiDPI part of this patch a268e41e2c31943661982cb2f1ab237a6fe4c030 now that it's been tested.

Then this PR is preparation.

TheCycoONE:
Removed HiDPI from the PR - this is now strictly a 'prep' PR for HiDPI.

tobylane:
One thing lost since SDL2 is that if you have a higher display resolution and 200% scale, then lower the display resolution and double click the window title, it will resize to the available space **and not allow you to lower the scale**.  
I went from 1920x1080 200%, where I could switch to 100%, to 1600x900 200%, where I can't switch to 100% because if I was on 100% I couldn't switch to 200%. On restarting the game I can switch to 100%.

Is there a purpose in passing params to `onWindowPixelSizeChanged` that aren't used?

CorsixTH/Lua/dialogs/resizables/file_browser.lua
```diff
@@ -78,9 +78,10 @@ function FilteredTreeControl:FilteredTreeControl(root, x, y, width, height, col_
   self.num_rows = (self.tree_rect.h - self.y_offset) / self.row_height
 
   -- Magic numbers used to find a static position across different screen resolutions.
-  local button1x = math.floor(TheApp.ui.app.config.width / 2 - 90)
+  local scr_w, scr_h = TheApp.video:getRenderDimensions()
```
lewri: Can we afford `screen_width` etc where we used `scr_w`/`sw` for transparency?

CorsixTH/Src/th_lua_gfx.cpp
```diff
@@ -1195,6 +1202,7 @@ void lua_register_gfx(const lua_register_state* pState) {
     lcb.add_function(l_surface_pop_clip, "popClip");
     lcb.add_function(l_surface_get_width, "getWidth");
     lcb.add_function(l_surface_get_height, "getHeight");
+    lcb.add_function(l_surface_get_render_dimensions, "getRenderDimensions");
```
lewri: On lua's side this feels slightly less intuitive/obvious than before.  I'd of preferred a pass through function that then called the C++ function to have it documented in lua's side. Either that or the function name just throws me.

lewri (commented):
In fullscreen mode the resolution box is now obsolete as you say, but it becomes a lie, which isn't nice. In this case it is likely useful to disable this setting really (maybe override its text)... or hide it.
Edit to above: I just wanted to check what another game does. In Barotrauma for borderless windowed (i.e. our fullscreen implementation), if you change the resolution it snaps it right back to what it is (your monitor's resolution).

However, I like that exiting fullscreen now snaps back to the previous resolution.

Some bugs:
- Also of note is that in fullscreen UI scale is still disabled. As I'm now at 1080p, this shouldn't be the case. In addition, If you change the resolution to 4k you do get 4x UI scale available.

- Aspect ratio also goes a bit funky if you fullscreen from having the game maximised (I would expect 16:9 in this case, instead its just a best guess aspect

Image

<p>

- In maximise mode (not fullscreen), resolution changes also have no effect. Coming out of maximise with your changed resolution also has no effect (stays the same as the maximise). You have to change resolution again.

- If you exit the game in maximise, then open the game again, it opens non-maximised. As a result the window's title bar can be pushed out of view (partially usually). Updating resolution in this mode also keeps the title bar out of view until you move it (e.g. a lower resolution).

!! Still need to test this with the laptop

CorsixTH/Lua/dialogs/resizables/file_browser.lua
TheCycoONE: scr_w as a convention predates this PR in world.lua and game_ui.lua, I was just continuing it. Otherwise I don't like the name, I'd use render_width or similar.

CorsixTH/Src/th_lua_gfx.cpp
TheCycoONE: Less obvious in what way? width and height are at best ambiguous.

TheCycoONE:
> In fullscreen mode the resolution box is now obsolete as you say, but it becomes a lie, which isn't nice. In this case it is likely useful to disable this setting really (maybe override its text)... or hide it. Edit to above: I just wanted to check what another game does. In Barotrauma for borderless windowed (i.e. our fullscreen implementation), if you change the resolution it snaps it right back to what it is (your monitor's resolution).
> 
Did you see my out of scope notes, because I think I addressed this.

> However, I like that exiting fullscreen now snaps back to the previous resolution.
> 
> Some bugs:
> 
>     * Also of note is that in fullscreen UI scale is still disabled. As I'm now at 1080p, this shouldn't be the case. In addition, If you change the resolution to 4k you do get 4x UI scale available.
> 
I don't know how to address this, at least not in this PR. We can consider it for the ui_scale PR.

>     * Aspect ratio also goes a bit funky if you fullscreen from having the game maximised (I would expect 16:9 in this case, instead its just a best guess aspect
> 
Because without a separate aspect ratio option I'm matching the aspect ratio of what the window was, and a maximized window is typically not 16:9, except maybe on a 16:10 monitor. If you have an idea I'll entertain it.
> 
> Image
> 
>     * In maximise mode (not fullscreen), resolution changes also have no effect. Coming out of maximise with your changed resolution also has no effect (stays the same as the maximise). You have to change resolution again.
> 
I'll look at this

>     * If you exit the game in maximise, then open the game again, it opens non-maximised. As a result the window's title bar can be pushed out of view (partially usually). Updating resolution in this mode also keeps the title bar out of view until you move it (e.g. a lower resolution).

We've never tracked whether the game was maximized, that sounds like a request for a new feature?

CorsixTH/Src/th_lua_gfx.cpp
TheCycoONE: A pass through function is the wrong solution to this, I think you're looking for a type definition file, similar to d.ts in typescript.

It's a good idea, but pretty out of scope of this PR, open an issue.

TheCycoONE:
> Is there a purpose in passing params to `onWindowPixelSizeChanged` that aren't used?

Not really, an earlier version of this PR used the size from that event instead of fetching it, and I have some desire to go back to that - passing the size through onChangeResolution because I think it would be more efficient, but I didn't get around to it.

Calling C functions from Lua (and vice versa) is expensive.

TheCycoONE:
>One thing lost since SDL2 is that if you have a higher display resolution and 200% scale, then lower the display resolution and double click the window title, it will resize to the available space and not allow you to lower the scale.
I went from 1920x1080 200%, where I could switch to 100%, to 1600x900 200%, where I can't switch to 100% because if I was on 100% I couldn't switch to 200%. On restarting the game I can switch to 100%.

@tobylane I'm trying to follow what you said here, can you make it step by step instructions for me?

CorsixTH/Src/th_lua_gfx.cpp
TheCycoONE: Renamed to getRenderSize(), which I hope is clearer.

tobylane:
Start in the System Settings app with a display resolution that will support 200% scale, like 1900x1080. Open CorsixTH and maximise the window, which (makes it 1920x1027 for me) unlocks 200% scale. Select 200% scale. Lower the display resolution to 1600x900 and maximise the CorsixTH window (which makes it 1600x847 for me). The scale is still 200%, the box says 200%, but it's greyed out as 100% is the only valid option, and I can't change it.

TheCycoONE:
> Start in the System Settings app with a display resolution that will support 200% scale, like 1900x1080. Open CorsixTH and maximise the window, which (makes it 1920x1027 for me) unlocks 200% scale. Select 200% scale. Lower the display resolution to 1600x900 and maximise the CorsixTH window (which makes it 1600x847 for me). The scale is still 200%, the box says 200%, but it's greyed out as 100% is the only valid option, and I can't change it.

Ok, I can reproduce this, but I can reproduce it without the changes from this PR too. I'd like to address it in a follow up that deals with scale.

Issue #3465, opened 2026-07-24
Cache path meta-information while creating the path


Alberth289346:
Currently, code is reverse-engineering the state of tiles at a path (in the hospital) at the time of creating a path. Obviously, this is complicated and probably not 100% correct.

As an alternative, meta-information about the state of tiles can be stored in the delivered path while computing the path. While the entity is using the path information, walk-code can then simply compare current state of a tile against the meta-information from the path, and reliably detect changes in tiles.

The data structure to store the information can be
- The computed path as a list of nodes, where a node contains an xy pair, a direction, and the tile-data at the time of creating the path.
- A 2D table to jump into the correct node of the path from an xy pair.

Open points:
- I am not sure what meta-information should be stored alongside the path. That needs to be discussed/decided.
- There are a number of different path computing functions. A discussion/decision is needed which functions should return meta-information alongside the path.
- I'd suggest to aim for a C++ object with the information rather than creating a new Lua result. (The current 2 separate lists output is not simple to extend at first sight.)

ARGAMX:
👍

regarding the data structures:
- list of nodes sounds transparent and simple.
- If it's a 2D table, how to store the next tile in a route? In the previous table cell (which store previous tile)? How is this different from the list variant then? I don't quite understand the advantages and disadvantages of a 2D table versus a list.

Which option uses less memory?

----------------

- which meta-information to store:
I think we need to store the XY and Tile Flags for that XY used when constructing the route: passable, avoidTile, travelWest, travelNorth, travelEast, travelSouth. For each tile, respectively.
I think that's all we need?

- which functions should return meta-information alongside the path:
obviously `pathfinder:findPath`? only it.

- output result:
I don't like the way the route is currently returned from CPP to Lua: two separate arrays. First contains x coordinates. The second contains y coordinates. so probably any other way would be better. is it possible?

----------------

Perhaps it would be a good idea to even create a CPP function to check each step of the route?
So, let's imagine the CPP outputs a route with meta information for each tile.
When humanoid+walk traversing the route, Lua calls this method at each tile and passes to the CPP: the meta information of two nodes/tiles: the one from which the humanoid is moving, and the one to which the humanoid is moving. The CPP receives this, also takes the current meta information for these two nodes/tiles, and compares the current meta information with the meta information it received as a params (from route).
If the information is identical, everything is great; we continue along the route! We return to Lua that everything is OK.
[...]

PR #3489, opened 2026-08-09
Recheck the path step is valid at each step

### LINKED ISSUE #3331: Humanoids can walk through walls

### LINKED ISSUE #3350: [Bug] New Sideobjects are ignored in a constructed walk
### Describe the issue

If a humanoid has a constructed walk and a sideobject (radiator/bin) blocks one of their walks in the path, it will ignore the object and walk through it.

This does not happen with new walk construction, as the path is made with awareness of the side objects.

### Steps to Reproduce

1. Identify a path a humanoid will walk though (they tend to follow a pattern in hospital layouts)

2.  Place a radiator 2 tiles ahead of the humanoid, on a tile they will walk on

3. Observe humanoid walks through the side object, even if the edge they would enter is blocked.

### Expected Behaviour

Humanoids should properly check for blocked edges on a tile while walking, with side objects.

- [ ] Right mouse panning

### Save Game

### CorsixTH Version

0.69.2

### Operating System

Windows 10

### Theme Hospital Version

GOG.com

### Gamelog.txt

```yaml

```

### Additional Information

Image

Image

Image

ARGAMX:
*Fixes #3350*
*Fixes #3331*

- When humanoid traversing a previously constructed path, check at each step that the next tile in the path is still directly connected to the current tile.

### DIFF
--- a/CorsixTH/Lua/humanoid_actions/walk.lua
+++ b/CorsixTH/Lua/humanoid_actions/walk.lua
@@ -198,9 +198,10 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
   local map = humanoid.world.map.th
   map:getCellFlags(x1, y1, flags_here)
   map:getCellFlags(x2, y2, flags_there)
-  local avoid = (not flags_here.avoidTile) and flags_there.avoidTile
-  local not_passable = flags_here.passable and (not flags_there.passable)
-  local obstacle_on_the_way = not_passable or avoid -- approaching avoidable or impassable tile
+  local avoid = (not flags_here.avoidTile) and flags_there.avoidTile -- approaching avoidable tile
+  local not_passable = flags_here.passable and (not flags_there.passable) -- approaching impassable tile
+  local obstacle_on_the_way = not_passable or avoid or
+      (humanoid.world:getPathDistance(x1, y1, x2, y2) ~= 1) -- path not valid anymore
 
   -- Also make sure that a room hasn't unexpectedly been built on top of the
   -- path since the route was calculated.


CorsixTH/Lua/humanoid_actions/walk.lua
```diff
@@ -198,9 +198,10 @@ local action_walk_tick; action_walk_tick = permanent"action_walk_tick"( function
   local map = humanoid.world.map.th
   map:getCellFlags(x1, y1, flags_here)
   map:getCellFlags(x2, y2, flags_there)
-  local avoid = (not flags_here.avoidTile) and flags_there.avoidTile
-  local not_passable = flags_here.passable and (not flags_there.passable)
-  local obstacle_on_the_way = not_passable or avoid -- approaching avoidable or impassable tile
+  local avoid = (not flags_here.avoidTile) and flags_there.avoidTile -- approaching avoidable tile
+  local not_passable = flags_here.passable and (not flags_there.passable) -- approaching impassable tile
+  local obstacle_on_the_way = not_passable or avoid or
+      (humanoid.world:getPathDistance(x1, y1, x2, y2) ~= 1) -- path not valid anymore
```
TheCycoONE, line 204: This is valid, it's also inefficient. If we find that's a problem we can make a C++ function to test travel to the neighbour without setting up full pathfinding to do it.

ARGAMX: > This is valid, it's also inefficient. If we find that's a problem we can make a C++ function to test travel to the neighbour without setting up full pathfinding to do it.

It would be nice if we had such a C++ function.

Alberth289346: Like #3465? That's in-progress.

I am not sure why it runs a path-finder here, while in #3465 we found that it only needs checking passable tile and edge flags.

ARGAMX: > Like #3465? That's in-progress.
> 
> I am not sure why it runs a path-finder here, while in #3465 we found that it only needs checking passable tile and edge flags.

Yes, as in #3465. Ok, then let's wait for #3465.

The pathfinding call didn't require that the rules for traversability between adjacent tiles be programmed in both Lua and C++.
Currently, such rules don't exist in Lua, and I don't want to add them here.
As they already exist in C++ part.

TheCycoONE: I would also hope that it would be pure c++, without the separate calls to getcellflags.

ARGAMX:
Blocked by #3465

PR #3490, opened 2026-08-09
HiDPI support: hidpi flag, unlocked/automatic UI scaling, aspect ratio option, window size limiting

TheCycoONE:
This PR covers everything I feel was needed for true HiDPI support in SDL3:
* Actual UI scale is automatically determined at runtime rather than fixed in config.
* Config options for scale are unlocked up to 400%.
* New 'Auto' option for UI scale is added, which currently scales as large as will fit the window.
* 4:3 aspect ratio option added that applies pillar boxing or letter boxing as needed.
* Resolution renamed to Window Size to better indicate the what it's used for.
* Window Size greyed out in Fullscreen mode.
* Some refactoring, particularly around video modes to make them easier to toggle.
* Request HiDPI windows from the windowing system (flag, true by default).
* Automatically adjust zoom, ui scale, and cursor scale to the windowing system's requested scale (flag, true by default).
* Experimental support for fractional scaling from the windowing system (flag, false by default).
* Maximized state is now preserved in game and in config between games.

We no longer limit the minimum window size based on the selected UI scale, or the UI scale options based on the window size. There were too many factors at play to make that practical, especially when HiDPI is considered.

In an earlier version of these changes I put these resolution dependent helper functions in ui.lua, but ui.lua is replaced by game_ui.lua and for awhile when loading there is neither. It is also persisted with save games and instantiated late, so I think graphics.lua is the next best place.

fractional_scaling crashes almost immediately, but the flag is there for future testing.

I realize this PR is on the large side - too large probably. I think my previous attempts at constraining the scope have caused more confusion though as people didn't understand where I was going. With this PR everything with respect to HiDPI, and scaling is in scope except true support for fractional scaling.


CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -90,11 +90,9 @@ local available_resolutions = function()
     {text = "1920x1200 (16:10)", width = 1920, height = 1200 },
   }
 
-  local s = TheApp.config.ui_scale
   local enable_list, disable_list = {}, {}
   for _, opt in ipairs(suggested_resolutions) do
-    local enabled = App.MIN_WINDOW_WIDTH * s <= opt.width and
-        App.MIN_WINDOW_HEIGHT * s <= opt.height
+    local enabled = true -- It might be nice to filter sizes that don't fit the display
```
lewri, line 95: This feels a bit hacky just to bypass code that now gets unused. Would it not make more sense to just remove the enable list + disable list entirely, if you're intending to go down that path?

I assume you're planning to work on the approach in a future pr in your comment, which might be why we're just using true too.

TheCycoONE: Yeah I don't plan to leave it like this. Didn't add the function to get the display size yet and haven't decided if we want it.

CorsixTH/Lua/ui.lua
```diff
@@ -991,8 +991,9 @@ end
 --! Window has been resized by the user
 --!param width (integer) New window width
 --!param height (integer) New window height
-function UI:onWindowResized(width, height)
-  if not self.app.config.fullscreen then
+--!param state (integer) Window state: 0 - window, 1 - fullscreen, 2 - maximized, 3 - minimized
```
lewri, line 973: Would it not make sense to go smallest to largest in the integers, unless we're being bound by sdl3 here?
0 min
1 window 
2 maximise 
3 fullscreen

TheCycoONE: I don't think so. I particularly prefer ordinary window mode to be 0.

TheCycoONE: for what it's worth the sdl values are:
0 - no flags (window)
1 - full-screen
64 - minimized
128 - maximized
(and lots of other states that can be combined)

I intentionally don't use theirs. In part because they use an unsigned 64bit int and that can't be represented in lua.

tobylane:
When turning off original aspect, should the menus be recentered?

In fullscreen or maximised, the window button gives the last windowed window size, but is giving the current resolution what the player expects? I keep the game maximised, but that choice is only remembered if I enter it as a custom resolution or drag the window to that size.

Changing the ui scale forces a change from maximised to windowed.

On starting the game with ui scale auto the button will show 0 not auto.

original_aspect_ratio and aspect_ratio_4_3 are naming the same thing.

In config, ui_scale says whole number, cursor_scale says natural number.

In the error for changing mode in UI it may be possible to give a translated name for the mode, eg copying the label name from options.

The rest of it plays well and looks good. :+1:

ARGAMX:
1. 

Screenshot 2026-08-16 at 14 07 05

If I enable Fullscreen mode, the Window Size tile becomes grayed out, and the Window Size value shows the resolution I had before I enabled Fullscreen.
But why it should show it to user now? 
I can't change it right now. And this "640x480" value is completely irrelevant right now.

All this makes me think that the Fullscreen and Window Size blocks are simply mutually exclusive.
If so, then it would be logical to place Fullscreen as one of the selections in the Window Size list? 
For example, as the first item?
Then, if I select Fullscreen in Window Size, I'll see "Window Size" = "Fullscreen", not the meaningless 640x480. 
That makes sense, right?
Because the current UI/UX is a bit weird.

ARGAMX:
Screenshot 2026-08-16 at 14 17 32

I need bigger
I need 800% cursor scale!
(kidding)

cursor scaling nice 👍

TheCycoONE:
We had an irc conversation about remembering maximized. I thought we decided not to but I can definitely add it. Maximized is different than a window that happens to take the full size, it would be a new config state. Mutually exclusive with fullscreen.

I'd probably replace fullscreen with window_mode. That will help avoid the ui_scale resetting it.

TheCycoONE:
> If I enable Fullscreen mode, the Window Size tile becomes grayed out, and the Window Size value shows the resolution I had before I enabled Fullscreen. But why it should show it to user now? I can't change it right now. And this "640x480" value is completely irrelevant right now.
> 
> All this makes me think that the Fullscreen and Window Size blocks are simply mutually exclusive. If so, then it would be logical to place Fullscreen as one of the selections in the Window Size list? For example, as the first item? Then, if I select Fullscreen in Window Size, I'll see "Window Size" = "Fullscreen", not the meaningless 640x480. That makes sense, right? Because the current UI/UX is a bit weird.

Never seen a UI that does this - I copied Wesnoth, though it looks like they show the monitor resolution in the box until you uncheck fullscreen, then it goes back to how it was. I can tweak that.

image

ARGAMX:
> > If I enable Fullscreen mode, the Window Size tile becomes grayed out, and the Window Size value shows the resolution I had before I enabled Fullscreen. But why it should show it to user now? I can't change it right now. And this "640x480" value is completely irrelevant right now.
> > All this makes me think that the Fullscreen and Window Size blocks are simply mutually exclusive. If so, then it would be logical to place Fullscreen as one of the selections in the Window Size list? For example, as the first item? Then, if I select Fullscreen in Window Size, I'll see "Window Size" = "Fullscreen", not the meaningless 640x480. That makes sense, right? Because the current UI/UX is a bit weird.
> 
> Never seen a UI that does this - I copied Wesnoth, though it looks like they show the monitor resolution in the box until you uncheck fullscreen, then it goes back to how it was. I can tweak that.
> 
> image

Yes. 
Or instead of what I wrote to make the resolution showed match the real one.

TheCycoONE:
> When turning off original aspect, should the menus be recentered?

Yes - I even fixed this at one point but it clearly regressed before I finished. (DONE)
> 
> In fullscreen or maximised, the window button gives the last windowed window size, but is giving the current resolution what the player expects? I keep the game maximised, but that choice is only remembered if I enter it as a custom resolution or drag the window to that size.

That was by design, but I think two tweaks would make it better:
1. The window size should show the maximized/full-screened window size, but should NOT save that setting to the config. This allows restore / unchecking full-screen to restore their old window size which is what people expect. (DONE)
2. We should save whether the game was maximized on close and restore it the same way. (DONE)
 
> Changing the ui scale forces a change from maximised to windowed.

Ops, unexpected - will fix. (DONE)
 
> On starting the game with ui scale auto the button will show 0 not auto.

Fixed (DONE)

> original_aspect_ratio and aspect_ratio_4_3 are naming the same thing.

Yeah, original_aspect_ratio in a player facing context, and aspect_ratio_4_3 in an engine facing context was my justification but it's a valid criticism. Do you think aspect_ratio_4_3 is OK for the config file? (Discuss)

> In config, ui_scale says whole number, cursor_scale says natural number.

Whole numbers are 0,1,2,3...; natural numbers are 1,2,3... The description in the config is right, though perhaps too technical (high school level math.) (Discuss)

> In the error for changing mode in UI it may be possible to give a translated name for the mode, eg copying the label name from options.

You mean the "Warning: could not toggle..." message? That's printed to the console, we don't tend to translate those. Perhaps there should be a UI error, but there wasn't before and I've never seen it come up.

> The rest of it plays well and looks good. 👍

Great! Thanks for testing.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -90,13 +90,19 @@ local available_resolutions = function()
...
   }
 
-  local s = TheApp.config.ui_scale
   local enable_list, disable_list = {}, {}
+  local max_window_width, max_window_height = TheApp.video:getMaxWindowSize()
+  if max_window_width == 0 then
+    max_window_width = 3000
+  end
+  if max_window_height == 0 then
+    max_window_height = 2000
```
lewri: Magic? I'd like to know why these are chosen and in what scenario we need it. Either in the C++ or here.
I'm not convinced 3000x2000 is a sane maximum either.

TheCycoONE: Truly magic / arbitrary.

The SDL function we're calling to get the maximum window size that can fit on the display has an error mode of returning 0. I haven't seen it, I have no idea under what scenario it would happen - probably some strange windowing system that refused to tell the program the display size.

I picked numbers bigger than any resolution we show so that it wouldn't disable any options. Figured that was better than not allowing any.

I'll add a comment.

TheCycoONE:
I think I have addressed every issue that didn't require further discussion. If we're lucky I didn't regress anything along the way.

ARGAMX:
2.

Screenshot 2026-08-17 at 11 35 22

Should we make `UI Scale` default value `Auto`?
Currently the default value is `100%`.
But I can’t find what advantages `100%` has over `Auto`?

Moreover, when user launch the game for the first time and the interface for selecting the path to the original game folder appears, it's already awkward and quite small.
And without Auto, in `100%` it's completely hard to use.

ARGAMX:
3. 
The camera zoom level shifts after load your old save game. Zoom is not compatible with old saves.

How to reproduce:
1. Start any level in version 0.70.0.
2. Save your game.
3. Load your save.
4. Press 0 (to reset the zoom level to default).
5. Observe that the zoom level hasn't changed, as the zoom level in your save game was already the default.
6. Close 0.70.0 and run the build from this PR.
7. Load your save.
8. Press 0 (to reset the zoom level to default).
9. Observe that the zoom level has changed. However, after loading, you should have the same zoom level as when you saved.

ARGAMX:
Screenshot 2026-08-17 at 12 03 32

It's a small thing, but I'd place `Fullscreen` and `Aspect 4:3` side by side. 
Either together at the top. Or together at the bottom.
Probably even better at the bottom.

I like how it on the screenshot you posted from Wesnoth.
There's a resolution dropdown first, and then three grouped checkboxes.
Although we don't have checkboxes, I think it's somewhat common to group Boolean switches together within a single section.

image

It might even be convenient to do it in this order:

Window Size (Dropdown)
UI Scale (Dropdown)
Cursor Scale (Dropdown)
Fullscreen (Bool)
Aspect 4:3 (Bool)
Check for Updateds (Bool)

What do you think?

Although this could be done in another PR, the question is rather whether to do it or not.

TheCycoONE:
> Screenshot 2026-08-17 at 11 35 22
> 
> Should we make `UI Scale` default value `Auto`? Currently the default value is `100%`. But I can’t find what advantages `100%` has over `Auto`?
> 
> Moreover, when user launch the game for the first time and the interface for selecting the path to the original game folder appears, it's already awkward and quite small. And without Auto, in `100%` it's completely hard to use.

It is auto, but you'll need a fresh config.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -90,13 +90,23 @@ local available_resolutions = function()
...
+  local max_window_width, max_window_height = TheApp.video:getMaxWindowSize()
+
+  -- It is possible that getMaxWindowSize could fail to detect the actual
+  -- available display space, in which case it will return 0, 0.
+  -- In that case we don't have any information to hide a resolution from the
+  -- user, so set the maximum to large enough values that nothing in the list
+  -- will be disabled. The numbers below are arbitrarily large.
+  if max_window_width == 0 or max_window_height == 0 then
+    max_window_width = 4000
+    max_window_height = 3000
```
lewri, line 103: Oh you went bigger... I was going to suggest a more realistic smallest maximum size...
From the Steam Hardware Survey 1920x1080 is still the leading primary monitor resolution. With a notable portion in 1440p.

Image

ARGAMX: > Oh you went bigger... I was going to suggest a more realistic smallest maximum size... From the Steam Hardware Survey 1920x1080 is still the leading primary monitor resolution. With a notable portion in 1440p.
> 
> Image

Interesting.
But 1920x1080 has been around for over 20 years.
I think this format is pretty outdated.

I gave away all my devices that couldn't display higher than 1920x1080 to my parents and neighbors' kids.
It's strange that it's the most popular format on Steam.
Probably my neighbors' kids with my old 1920x1080 monitors spend more time on Steam than me.

I think we should consider something around 4k.
In this regard, 4000 and 3000 seem more appropriate.

Maybe we should use values ​​like
3840 + 1
2160 + 1
?
they seem less magical than 4000 and 3000.

lewri: Just because something is dated doesn't mean it can't be the correct answer. Only 5% have a 4k screen; which means we're offering unsuitable sizes to potentially 95% of people with smaller screen.

If you want to cut the difference than target something with low utilisation then use 2560x1440 imo

lewri: Actually I misread cyco's comments added into the code; this is only to enable all resolutions on failure. In which case I'm less offput by such values.


Though you could have less magic by just modifying how the `enabled` stuff works.

```lua
local enabled = max_window_width == 0 or max_window_height == 0
if not enabled then
  enabled = opt.width < max_window_width and opt.height < max_window_height
end
```

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -90,13 +90,23 @@ local available_resolutions = function()
...
+  end
+
   for _, opt in ipairs(suggested_resolutions) do
-    local enabled = App.MIN_WINDOW_WIDTH * s <= opt.width and
-        App.MIN_WINDOW_HEIGHT * s <= opt.height
+    local enabled = opt.width < max_window_width and opt.height < max_window_height
```
lewri: This should be `<=` inequality no?

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE: Leaving as is.

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE: Yes, yes it should.

TheCycoONE:
> It might even be convenient to do it in this order:
> 
> Window Size (Dropdown) UI Scale (Dropdown) Cursor Scale (Dropdown) Fullscreen (Bool) Aspect 4:3 (Bool) Check for Updateds (Bool)
> 
> What do you think?
> 
> Although this could be done in another PR, the question is rather whether to do it or not.

Makes sense to me, the current order is pretty haphazard.

TheCycoONE:
> The camera zoom level shifts after load your old save game. Zoom is not compatible with old saves.
> 
> How to reproduce:
> 
>     1. Start any level in version 0.70.0.
> 
>     2. Save your game.
> 
>     3. Load your save.
> 
>     4. Press 0 (to reset the zoom level to default).
> 
>     5. Observe that the zoom level hasn't changed, as the zoom level in your save game was already the default.
> 
>     6. Close 0.70.0 and run the build from this PR.
> 
>     7. Load your save.
> 
>     8. Press 0 (to reset the zoom level to default).
> 
>     9. Observe that the zoom level has changed. However, after loading, you should have the same zoom level as when you saved.

Default zoom level is not part of a save game, this patch is the first time it's set to anything other than 1. As far as I understand what you're saying it's intentional. The save loads with the zoom level of the save which is not the new default (unless you happened to zoom in equivalent to your display scale).

TheCycoONE:
Can we discuss original_aspect_ratio and aspect_ratio_4_3 please?

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE: Fixed.

lewri:
I don't think I'll be able to review this properly now until next week as I'll be away from today. Unless I get some time this morning.

ARGAMX:
> Can we discuss original_aspect_ratio and aspect_ratio_4_3 please?

> Yeah, original_aspect_ratio in a player facing context, and aspect_ratio_4_3 in an engine facing context was my justification but it's a valid criticism. Do you think aspect_ratio_4_3 is OK for the config file? (Discuss)

The logic you followed is quite consistent and clear.
My opinion is that, overall, it doesn't really matter whether we name the variable in the config file `original_aspect_ratio` or `aspect_ratio_4_3`.
Without a comment, both names aren't very informative.
But if I had to choose, I'd better choose `original_aspect_ratio` for config.
Perhaps we could even use the name `original_aspect_ratio_4_3`.
Overall, I don't think this is important.

ARGAMX:
> Default zoom level is not part of a save game, this patch is the first time it's set to anything other than 1. As far as I understand what you're saying it's intentional. The save loads with the zoom level of the save which is not the new default (unless you happened to zoom in equivalent to your display scale).

Yes, I understand that the default zoom level is not part of a save game.
Nor should it be.
But the current zoom level is a part of a save game.

The problem isn't even the default zoom level.
The problem is the lack of compatibility.
I simply used the default zoom level as a clear reference point to keep things simple.
You can use any other zoom level and repeat the experiment.

The results of the experiment indicate that the same zoom level in the save is interpreted differently before this PR (or may be not even before this PR? I haven't checked that for sure. I used 0.70 as a 'before this PR') and after this PR. Thus, there is no zoom level compatibility between 0.70 and the hypothetical 0.71.
This may only be true on macOS and Wayland.

I see that the zoom level differs by about 2 times.

The same save:

in 0.70.0:

Screenshot 2026-08-18 at 11 24 56

in This PR:

Screenshot 2026-08-18 at 11 25 12

.
In general, the loss of this compatibility isn't a big deal, and I could live with it.
I just want us to clearly understand that this loss exists and that we consciously accept it. Or not accept.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -543,12 +561,23 @@ function UIOptions:buttonUpdates()
...
       local err = {_S.errors.unavailable_screen_size}
       self.ui:addWindow(UIInformation(self.ui, err))
       self.fullscreen_button:toggle()
   end
-  self.fullscreen_panel:setLabel(self.ui.app.fullscreen and _S.options_window.option_on or _S.options_window.option_off)
+  self.fullscreen_panel:setLabel(self.ui.app.modes.fullscreen and _S.options_window.option_on or _S.options_window.option_off)
+  self.resolution_button:enable(not self.ui.app.modes.fullscreen)
+end
+
+function UIOptions:buttonOriginalAspectRatio()
```
lewri, line 575: Image
We're a bit tight on the caption box width now. Translations will possibly overflow more easily.

Though I'm also not a fan of the chosen string. It seems detached currently and I'm not immediately associating what it is for.

At this point you might have better luck making a child window called 'Display Settings'.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -119,11 +129,9 @@ end
 
 local available_ui_scales = function()
   local res = {}
-  local s = 1
-  while s * App.MIN_WINDOW_WIDTH <= TheApp.config.width and
-      s * App.MIN_WINDOW_HEIGHT <= TheApp.config.height do
+  res[1] = { text = _S.options_window.scale_ui_auto, scale = 0 }
+  for s = 1, 4 do
     res[#res + 1] = { text = tostring(s * 100) .. '%', scale = s }
```
lewri, line 134: Showing me scales I can't use but can select is poor UX. I get why, but I would rather the shown UI scale value updated its setting based on the size change than leaving it unchanged.

lewri: Please can the Auto option also be referenced in a tooltip to what will happen.

CorsixTH/Lua/dialogs/resizables/options.lua
lewri: I feel like Original Aspect should only apply in-game too. There is no need to add bars out of the game.

WindowsInstaller/config_template.txt
```diff
@@ -423,6 +434,28 @@ blocking_off_areas = 2
...
+
+-- Fractional Scaling: Allows ui_scale and cursor scale to be a fractional
+-- number. With fractional_scaling enabled and apply_window_display_scale
+-- enabled CorsixTH will use the exact display scale provided by the
+-- windowing system instead of rounding down to the nearest whole number
+-- greater than zero.
+--
+-- THIS SETTING IS HIGHLY EXPERIMENTAL AND WILL CRASH YOUR GAME!!!
+--
+fractional_scaling = false
```
lewri: Can you change this to debug_fractional_scaling so we don't end up in a direct_zoom debacle again?

lewri (commented):
I would like to see a cursor scale auto option too. We do have a current state of 'auto' UI scale on resize, but no 'auto' for the cursor on resize.

<details>
<summary>This one isn't a you problem though in maximise (1920x1017) and UI scale 2 can make the 'fullscreen' windows panic around a bit as they hit the boundary edge.</summary>

Image
</details>

Also I don't even normally use hidpi in Windows. So I need to still set up testing that; so my review probably isn't done or as thorough yet I'd like to see.

ARGAMX:
> I would like to see a cursor scale auto option too. We do have a current state of 'auto' UI scale on resize, but no 'auto' for the cursor on resize.

+1 for that

CorsixTH/Lua/dialogs/resizables/options.lua
ARGAMX: > I feel like Original Aspect should only apply in-game too. There is no need to add bars out of the game.

Generally speaking, yes.
But then, when user switch this option in the menu, it's unclear what it actually does, because nothing happens.
User can only change it in the menu.

lewri: Improving the text probably solves this, as I don't think the current string is good as is.

We do a similar thing with capture mouse, it doesn't take effect until in game.

tobylane:
> Can we discuss original_aspect_ratio and aspect_ratio_4_3 please?

Weak preference for the first (informs new people of its intent), strong preference for the choice being used throughout, and original or 4:3 only mentioned in config comments and the dialog strings.

There's a noticeable size change with maximised = true, right after the window is made at the smaller size. It's not inherently bad, but it's different to considering the maximised resolution a real choice.

TheCycoONE:
What's the logic for cursor_scale auto?

TheCycoONE:
I think your zoom thing is not as clear cut as you might think it is.

1. I think it only impacts macOS
2. I think it depends on the hidpi flag setting
3. I think one would have to adapt every time on save and load to be consistent across hidpi on / off.
4. I think if display_scale is off you would see the same issue with the default zoom level.
5. It 'might' be more consistent if I apply the display_scale on top of the configured zoom instead of just setting a new default. This is worth trying in this PR.

WindowsInstaller/config_template.txt
TheCycoONE: Sure.

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE: Provide wording please.

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE: I feel strongly that this is a bad idea.

TheCycoONE:
> > Can we discuss original_aspect_ratio and aspect_ratio_4_3 please?
> 
> Weak preference for the first (informs new people of its intent), strong preference for the choice being used throughout, and original or 4:3 only mentioned in config comments and the dialog strings.

Ok
> 
> There's a noticeable size change with maximised = true, right after the window is made at the smaller size. It's not inherently bad, but it's different to considering the maximised resolution a real choice.

I might be able to hide that. I don't recall if I tried and it didn't work or not. The window needs to be created at the size it will restore down to.

TheCycoONE:
> > There's a noticeable size change with maximised = true, right after the window is made at the smaller size. It's not inherently bad, but it's different to considering the maximised resolution a real choice.
> 
> I might be able to hide that. I don't recall if I tried and it didn't work or not. The window needs to be created at the size it will restore down to.

looks like I can't hide that.

ARGAMX:
> What's the logic for cursor_scale auto?

Can `Auto` value for `Cursor scale` choose applied value the same way as `UI scale` `Auto` selects applied value?
So 
For 200% `UI scale` user will get 200% `Cursor scale`,
For 300% `UI scale` user will get 300% `Cursor scale`,
and etc.

TheCycoONE:
> > What's the logic for cursor_scale auto?
> 
> Can `Auto` value for `Cursor scale` choose applied value the same way as `UI scale` `Auto` selects applied value? So For 200% `UI scale` user will get 200% `Cursor scale`, For 300% `UI scale` user will get 300% `Cursor scale`, and etc.

Yes, I can add an Auto option to cursor scale that defers to the current actual UI scale.

TheCycoONE:
@ARGAMX see what you think of this zoom implementation. If anything the amount of time I sunk into it makes it worth it :/

CorsixTH/Lua/game_ui.lua
```diff
@@ -69,19 +69,27 @@ function GameUI:GameUI(app, local_hospital, map_editor)
...
-  self.visible_diamond = self:makeVisibleDiamond(scr_w, scr_h)
+
+  -- With the zoom the game screen is the pixel size of the screen divided by
+  -- the effective zoom level (zoom_factor * display scale)
+  local wwz, hwz = scr_w / efz, scr_h / efz
+  self.visible_diamond = self:makeVisibleDiamond(wwz, hwz)
   if self.visible_diamond.w <= 0 or self.visible_diamond.h <= 0 then
     -- For a standard 128x128 map, screen size would have to be in the
     -- region of 3276x2457 in order to be too large.
     if not self.map_editor then
```
lewri, line 85: I realise these changes now likely effect the text and errors here. We now have UI scale and will autozoom with it too? I'm not entirely sure on the impacts but the messaging might now be wrong.

TheCycoONE: We don't auto zoom with UI scale, we auto zoom with the windowing system's display scale. On MacOS in particular users seem to be use to thinking of window coordinates (with display scale factored in) as their resolution already so they would expect this messaging. On other platforms I'm not sure what they'd expect.

Say you had an 8k monitor, before you couldn't play CorsixTH, you'd get this error. But now if you had a display_scale of 200% in your operating system, CorsixTH would treat it as effectively a 4k monitor and you could play.

I'm having a hard time thinking of a realistic scenario where they would still get this message and don't have a 100% display scale.

I've wondered for awhile why we stop people from playing at all instead of just zooming in until the game fits - but that's out of scope.

lewri: I think you've managed to confuse me more. If you have UI scale set to 1x, but you have your OS scaling to 2x are you saying the zoom will be 2x rather than 1x?

TheCycoONE: Yes, and so will UI scaling. Perhaps check the comment on apply_window_display_scale config.

TheCycoONE:
I think the current action items on this are:

@ARGAMX determine if zoom changes improve your experience.
@lewri provide updated wording for anything you think is misleading.

And someone can open a ticket about a desired options menu layout, because that can be hashed out in a design ahead of any implementation and independent of this PR.

ARGAMX (approved):
Zoom changes improve my experience 👍
I am happy, thanks
All my other comments have also been resolved.
Great job 👍

TheCycoONE:
At least for me picking up staff isn't working on this branch; looking into it. Regression with the zoom changes.

TheCycoONE:
Putting back to WIP while I play test more heavily since there were some rather large changes to accommodate feedback.

TheCycoONE:
Add PRs for any label / error wording changes, I don't think we need to hold this for that.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -239,24 +248,38 @@ function UIOptions:UIOptions(ui, mode)
       "", _S.tooltip.options_window.select_resolution,
       { bg = col.setting, active = col.setting_active },
       self.dropdownResolution, false)
+  self.resolution_button:enable(not app.modes.fullscreen)
 
   -- UI Scale
-  local scale_label = TheApp.config.ui_scale * 100 .. "%"
+  local scale_label = TheApp.config.ui_scale == 0 and
+      _S.options_window.scale_auto or TheApp.config.ui_scale * 100 .. "%"
```
lewri, line 255: A nice to have would be `Auto (current%)` but not something I'd lose sleep over.

CorsixTH/Lua/dialogs/resizables/options.lua
lewri, line 523: For some reason this setting isn't saving and resets back to 100% on relaunch

CorsixTH/Lua/languages/english.lua
```diff
@@ -484,8 +484,10 @@ options_window = {
   option_enabled = "Enabled",
   option_disabled = "Disabled",
   fullscreen = "Fullscreen",
-  resolution = "Resolution",
+  resolution = "Window Size",
+  original_aspect_ratio = "Original Aspect (4:3)",
```
lewri, line 488: It's likely dropping the word 'Ratio' is what was irksome here that I came across last week.

CorsixTH/Lua/languages/english.lua
```diff
@@ -484,8 +484,10 @@ options_window = {
   option_enabled = "Enabled",
   option_disabled = "Disabled",
   fullscreen = "Fullscreen",
-  resolution = "Resolution",
+  resolution = "Window Size",
```
lewri: Nit but the variable name should also be updated

CorsixTH/Lua/languages/english.lua
```diff
@@ -524,20 +526,20 @@ tooltip.autosave_frequency = {
 tooltip.options_window = {
   fullscreen = "Whether the game should run in fullscreen or windowed mode",
   fullscreen_button = "Click to toggle fullscreen mode",
-  resolution = "The resolution the game should run in",
-  select_resolution = "Select a new resolution",
-  resolution_unavailable = "Resolution unavailable at current UI Scale",
-  scale_ui = "Scale the user interface. Only scale options that fit the display are shown, for more increase the resolution.",
+  resolution = "The window size the game should run in",
+  select_resolution = "Select a new window size",
```
lewri: Also here as part of that nit.

CorsixTH/Lua/languages/english.lua
```diff
@@ -524,20 +526,20 @@ tooltip.autosave_frequency = {
...
+  original_aspect_ratio = "Apply black bars if necessary to preserve the original game's 4:3 aspect ratio",
+  resolution_does_not_fit = "The window size does not fit the current display",
```
lewri: And here

CorsixTH/Lua/dialogs/resizables/options.lua
lewri: Scaling will adjust based on the chosen window size

--

I do find it odd still though that we will show all UI scales as available even when not possible at a given window size, but we will grey out window sizes not suitable.

CorsixTH/Lua/app.lua
```diff
@@ -1144,10 +1142,17 @@ function App:fixConfig()
...
       if type(value) == "number" then
-        self.config[key] = math.max(math.floor(value), 1)
+        self.config[key] = math.max(value, 0)
+      else
+        self.config[key] = 0
+      end
+
+    elseif key == "cursor_scale" then
+      if type(value) == "number" then
+        self.config[key] = math.max(value, 1)
```
lewri: Is this where the cursor scale keeps getting reset by for Auto?

CorsixTH/Lua/config_finder.lua
```diff
@@ -720,6 +737,28 @@ param(config_values, 'blocking_off_areas') .. [=[
...
 param(config_values, 'direct_zoom') .. [=[
 
+-------------------------------------------------------------------------------
+-- HiDPI: Inform the windowing system that the game is HiDPI aware. This will
+-- allow the game to be rendered at the native resolution of the display.
+-- The exact effect depends on the windowing system and the display.
+--
+-- In HiDPI mode the windowing system supplies the application with a display
+-- scale value which if applied will scale the cursor, ui, and default zoom in
+-- addition to the configured value for those fields.
```
lewri, line 747: I'm not entirely sure where the code bit is for this but I'm also not sure if it is working. Some tests today on a 3840 x 2160 display appear that changing the Windows scaling from 150% to 100% still keep UI scale at 2x even on auto.

Interestingly this is greyed out on that display regardless of Windows scaling: 

Image

Interestingly, if you do have Windows scaling to 200%; the game's UI does scale to 200% but will still say 100%. If you do use Auto, then switch, it treats 100% as 200%; 200% as 300% etc.

lewri (commented):
(This is from a setup I can properly do hidpi testing in)

- When changing UI sizes, tip of the day gets refreshed, even if the size does not change.
- Dragging a window from a larger display to a smaller one does not update its own window size. However, the dropdown down update.
- For your fractional scaling I assume this is the expected error:
```sh
Lua\graphics.lua:264: bad extra argument #-1 to 'draw' (number has no integer representation)
stack traceback:
        [C]: in method 'draw'
        Lua\graphics.lua:264: in field 'draw'
        Lua\ui.lua:318: in method 'draw'
        Lua\app.lua:1337: in function <Lua\app.lua:1330>
        (...tail calls...)

Error in frame callback: 0
Warning: No event handler for frame
```

CorsixTH/Lua/config_finder.lua
ARGAMX: > Interestingly, if you do have Windows scaling to 200%; the game's UI does scale to 200% but will still say 100%. If you do use Auto, then switch, it treats 100% as 200%; 200% as 300% etc.

It looks right to me. on macOS work the same way. and that’s how it should be I think.
The operating system has its own scaling, and the game has its own scaling.

CorsixTH/Lua/app.lua
TheCycoONE: Yes it is, will fix.

CorsixTH/Lua/config_finder.lua
TheCycoONE: 150% is ignored, we only support 100%, 200%, 300% etc. The reason for that was discussed on the channel, has to do with fractional scaling.

Basically since we can't do fractional scaling, if we did support 150% it would go something like

1 x 1.5 = 1
2 x 1.5 = 3
3 x 1.5 = 4
4 x 1.5 = 6

And that would be really confusing.

200% window scaling with display_scale on should treat 100% as 200%, 200% as 400%, 300% as 600% and so on. That part is expected. You mentioned needing to select auto and then go back? That wouldn't be expected, it should take effect immediately. Can you tell if we got a display scale changed event?

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE: fixed.

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE: Not hard to do, but possibly misleading? Possibly also not worth the additional getCurrentUIScaleWithoutDisplayScale or whatever it would be called.

TheCycoONE:
3840 x 2160 is probably greyed out because the usable window area is less than that unless you are in full-screen; the windows taskbar and any window chrome is subtracted from that.

CorsixTH/Lua/languages/english.lua
TheCycoONE: These were intentional, but maybe not for a good reason. I thought the usage was close enough that we'd rather existing translations were used instead of reverting to English.

Maybe that is a bad assumption.

CorsixTH/Lua/languages/english.lua
TheCycoONE: What bothers me here is that I spent time trying to come up with something that fit. I know it's not ideal, and nothing that's been said about it is helpful. That's why I'm asking for suggested wording, otherwise we leave it for some future play with the settings dialogs PR.

CorsixTH/Lua/dialogs/resizables/options.lua
lewri: It depends I guess... I usually like knowing the real UI scale being presented to me. I think you can also override hidpi's scale by setting 100% again in game?

CorsixTH/Lua/config_finder.lua
lewri: I imagine the new idea of this is what is more jarring than it actually being a common feature in modern apps...

lewri: I assume if you want to go smaller, you have to turn off hidpi?

CorsixTH/Lua/languages/english.lua
lewri: I suppose other languages will have an inconsistency now where they use resolution instead of window size. A new variable would force a change though it will also present in english until a translator comes along. Or you rename every language's `resolution` to `window_size` to at least have something

ARGAMX: I also think that the variable name should be changed in such cases. We have already translated it everywhere as “Resolution”.

CorsixTH/Lua/config_finder.lua
TheCycoONE:
```
-- HiDPI: Inform the windowing system that the game is HiDPI aware. This will
-- allow the game to be rendered at the native resolution of the display.
-- The exact effect depends on the windowing system and the display.
--
-- In HiDPI mode the windowing system supplies the application with a display
-- scale value which if applied will scale the cursor, ui, and default zoom in
-- addition to the configured value for those fields.
--
hidpi = true
apply_window_display_scale = true
```

TheCycoONE: To ignore the scale setting in the OS and smaller, you would need to disable that field  (apply_window_display_scale). When we support fractional we could consider e.g. 50% as an option.

Setting hidpi = false does practically nothing on windows I think? on MacOS setting it to false goes back to the 0.70 / SDL 2 behaviour of everything being big and blurry.

The effect of setting both to true on MacOS is everything is the same size as it was in SDL2, but much sharper, particularly unicode fonts.

CorsixTH/Lua/languages/english.lua
TheCycoONE: fine, done.

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE: I don't want to do this, but I won't block it if someone else wants to make a PR; I'll guide them through it if desired.

TheCycoONE:
> (This is from a setup I can properly do hidpi testing in)
> 
>     * When changing UI sizes, tip of the day gets refreshed, even if the size does not change.

It gets refreshed on every call to onChangeResolution, it probably shouldn't, but that's not a regression. We call onChangeResolution any time the size might change.  master and 0.70 are the same.

>     * Dragging a window from a larger display to a smaller one does not update its own window size. However, the dropdown down update.
> 

Not on Windows, does on MacOS (depending on the display scale setting in the OS.) This is operating system behaviour.

>     * For your fractional scaling I assume this is the expected error:
> 
> 
> ```shell
> Lua\graphics.lua:264: bad extra argument #-1 to 'draw' (number has no integer representation)
> stack traceback:
>         [C]: in method 'draw'
>         Lua\graphics.lua:264: in field 'draw'
>         Lua\ui.lua:318: in method 'draw'
>         Lua\app.lua:1337: in function <Lua\app.lua:1330>
>         (...tail calls...)
> 
> Error in frame callback: 0
> Warning: No event handler for frame
> ```

Yes.

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -514,7 +514,7 @@ function UIOptions:selectUIScale(number)
   TheApp:saveConfig()
   self.scale_ui_panel:setLabel(res.text)
   TheApp.gfx:onChangeUIScale()
-  self.ui:onChangeResolution()
+  self.ui:onChangeWindowSize()
```
tobylane:
```suggestion
  self:onChangeWindowSize()
```

CorsixTH/Lua/dialogs/resizables/options.lua
```diff
@@ -579,7 +579,7 @@ function UIOptions:buttonOriginalAspectRatio()
     self.aspect_button:toggle()
   end
   self.aspect_panel:setLabel(self.ui.app.modes.aspect_ratio_4_3 and _S.options_window.option_on or _S.options_window.option_off)
-  self.ui:onChangeResolution()
+  self.ui:onChangeWindowSize()
```
tobylane:
```suggestion
  self:onChangeWindowSize()
```

tobylane (commented):
After these two, all good.

edit: Or did you intend to stick to `UI:onChangeResolution`?

CorsixTH/Lua/dialogs/resizables/options.lua
TheCycoONE: These are intended to stay as self.ui:onChangeResolution(); to handle any potential screen resolution changes as a result.

TheCycoONE: Fixed.

TheCycoONE:
> After these two, all good.
> 
> edit: Or did you intend to stick to `UI:onChangeResolution`?

Intended to stick with UI:onChangeResolution

ARGAMX:
Sorry, I just noticed that the Sprite Viewer has become twice as small.
And increasing the zoom doesn’t affect it.
Today I needed to work with it, and I noticed that it’s now incredibly small.

Screenshot 2026-09-09 at 16 36 44

PR #3494, opened 2026-08-12
Generate per-file pages so the docs file hierarchy links resolve

### LINKED ISSUE #1793: Lua Documentation Links broken in github-pages
Following error occurs on all files/class links in GitHub Pages docs

```
404

File not found

The site configured at this address does not contain the requested file.

If this is your site, make sure that the filename case matches the URL.
For root URLs (like http://example.com/) you must provide an index.html file.

Read the full documentation for more information about using GitHub Pages
```

BrunosGits:
Closes #1793

The "Files" tab in the generated Lua docs links every file and directory to a page that does not exist, so all those links 404 on GitHub Pages. LDocGen only ever wrote class pages and index pages, never a page per source file, and the file tree built the hrefs from `LuaFile:getId()` which bakes the directory path and the `.lua` extension into the name.

Changes:
- `LDocGen/main.lua` now writes one page per source file. Each file page lists the classes and functions declared in that file, linking classes to their existing class pages
- `LDocGen/templates/file.htlua` is the new file page template
- `LDocGen/templates/lua_file_tree.htlua` renders directories as plain text (no directory pages exist) and keeps file entries as links to their new pages

Verification: built the docs locally (`cmake --build build --target doc_corsixth_lua`) and checked every local href across all generated pages resolves, 20465 links checked with no broken ones.

### DIFF
--- a/LDocGen/main.lua
+++ b/LDocGen/main.lua
@@ -109,6 +109,44 @@ for name, var in globals:pairs() do
     })
   end
 end
+local file_globals = {}
+for _, var in globals:pairs() do
+  local file_name = var:getFile()
+  if file_name then
+    local list = file_globals[file_name]
+    if not list then
+      list = {}
+      file_globals[file_name] = list
+    end
+    list[#list + 1] = var
+  end
+end
+
+local function collect_files(directory, prefix, files)
+  for _, item in ipairs(directory.children) do
+    local path = prefix == "" and item.name or prefix .. "/" .. item.name
+    if class.is(item, LuaDirectory) then
+      collect_files(item, path, files)
+    else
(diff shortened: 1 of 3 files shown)


TheCycoONE:
Welcome @BrunosGits 
Thanks for the contribution!

TheCycoONE:
The change looks good and works well for me. It would be nicer if directories were sorted to the top or bottom of the lists, but that is out of scope of the bug and would mess up the code a little so we can leave it.

I'll let @tobylane take a look, but I'm happy to merge.

TheCycoONE (approved):

tobylane (approved):
Thank you! You've rescued my hope in AI coding too.

PR #3500, opened 2026-08-16
Fix missing permissions block on workflows

TheCycoONE:
Flagged by github's security codescan. Without specifying permissions the workflows default to having fairly broad access to read and write the repo, issues, PRs, etc. Applying least-privileged principal it's better to define only what they need.

### DIFF
--- a/.github/workflows/update-website.yml
+++ b/.github/workflows/update-website.yml
@@ -1,5 +1,7 @@
 name: Update website
 
+permissions: {}
+
 on:
   release:
     types:
--- a/.github/workflows/Linux.yml
+++ b/.github/workflows/Linux.yml
@@ -1,6 +1,10 @@
 ---
 name: Linux and Tests
 
+permissions:
+  contents: read
+  pull-requests: read
+
 on:
   push:
     branches-ignore:
--- a/.github/workflows/Windows.yml
+++ b/.github/workflows/Windows.yml
@@ -1,6 +1,10 @@
 ---
 name: Windows
 
+permissions:
+  contents: read
+  pull-requests: read
+
 on:
   push:
     branches-ignore:
--- a/.github/workflows/WindowsInstaller.yml
+++ b/.github/workflows/WindowsInstaller.yml
@@ -1,6 +1,9 @@
 ---
 name: Windows installer
 
+permissions:
+  contents: read
+
 on:
   workflow_dispatch:
     inputs:


TheCycoONE:
It might be good to remove the pr input from workflows. I've never used it and it's considered unsafe; unlike a normal pull_request action run_workflow runs with our $GITHUB_TOKEN.

tobylane (approved):
Yes, remove the pr bit too. I may try to find out how to do it in a secure way.

tobylane:
> I may try to find out how to do it in a secure way.

```
      - name: Checkout
        uses: actions/checkout@v6
        with:
          repository: ${{ inputs.pr != '' && 'CorsixTH/CorsixTH' || github.repository }}
          ref: ${{ inputs.pr != '' && format('refs/pull/{0}/head', inputs.pr) || '' }}
```
CodeQL had no complaints about this replacing the PR bit of Define Job.

TheCycoONE:
> > I may try to find out how to do it in a secure way.
> 
> ```
>       - name: Checkout
>         uses: actions/checkout@v6
>         with:
>           repository: ${{ inputs.pr != '' && 'CorsixTH/CorsixTH' || github.repository }}
>           ref: ${{ inputs.pr != '' && format('refs/pull/{0}/head', inputs.pr) || '' }}
> ```
> 
> CodeQL had no complaints about this replacing the PR bit of Define Job.

How do you interpret the v7 notes here https://github.com/actions/checkout

lewri:
> It might be good to remove the pr input from workflows. I've never used it and it's considered unsafe; unlike a normal pull_request action run_workflow runs with our $GITHUB_TOKEN.

I occasionally need to use the PR function to build Windows artifacts I may have missed before they expired, though it is not often. In the AppImage one I use an option to build against someone else's repo + branch, but I can't see that being needed here.

TheCycoONE:
> > It might be good to remove the pr input from workflows. I've never used it and it's considered unsafe; unlike a normal pull_request action run_workflow runs with our $GITHUB_TOKEN.
> 
> I occasionally need to use the PR function to build Windows artifacts I may have missed before they expired, though it is not often. In the AppImage one I use an option to build against someone else's repo + branch, but I can't see that being needed here.

Sounds like that could be a tag option. But we can probably move all this to a discussion, and merge the PR as is (without removing any functionality)?

.github/workflows/update-website.yml
```diff
@@ -1,5 +1,7 @@
 name: Update website
 
+permissions: {}
```
lewri, line 3: I assume it's right this one needs no perms?

lewri (approved):

.github/workflows/update-website.yml
TheCycoONE: That's what I gathered.

TheCycoONE: Because it doesn't use our GITHUB_TOKEN but the token for the other repo for everything it does.

PR #3504, opened 2026-08-18
Fix #1467: Deferred entity destruction to prevent iteration skip

### LINKED ISSUE #1467: entities table is modified inside ipairs loop over entities

BrunosGits:
## Fix for #1467: Deferred Entity Destruction

**Problem:** Iterating `world.entities` with `ipairs` while handlers call `destroyEntity` shifts later entries into already-visited slots, skipping their tick calls.

**Solution:** Defer removal until after the iteration. When `current_tick_entity` is set (inside an iteration), `destroyEntity` queues the entity with `to_destroy=true`, calls `onDestroy()` immediately, and removes it in `_flushDestroyedEntities()` after the loop. Outside iterations, destroy immediately.

**Files changed:**
- `CorsixTH/Lua/world.lua` — Core fix + `_flushDestroyedEntities()` + lazy `entities_to_destroy` for old savegames
- `CorsixTH/Luatest/spec/world_spec.lua` — 23 unit tests (immediate/deferred destroy, no-skip, self-destruct, cascading, old-save compat, plant tickDay, checkForDeadlock, nested iterations)
- `smoketest.lua` — Smoke test improvements (heartbeat, load-only, intro movie fix, auto-difficulty)

**Validation:**
- ✅ 86/86 unit tests pass (busted)
- ✅ luacheck clean (297 files)
- ✅ Negative control: disabling the fix reproduces the exact bug ("dummy C was skipped")
- ✅ Full game smoke test passes on full Theme Hospital data:
  - Offscreen (`SDL_VIDEODRIVER=offscreen`): 3/3 green
  - xvfb (`xvfb-run -a`): 3/3 green  
  - Demo data control: 2/2 green
- ✅ Save/load round-trip works
- ✅ Old-save compatibility verified (lazy queue creation)

**Key discovery during testing:** Full game data autoplays the intro movie (`moviePlayer.playing=true`), which blocks `World:onTick` entirely — this was the real cause of the 500s timeout. Fixed by adding `TheApp.moviePlayer:stop()` in the smoketest.

The fix is minimal, targeted, and guarded only when iteration is actually happening.

### DIFF
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local SDL = require("sdl")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 264 -- SDL 3
+local SAVEGAME_VERSION = 265 -- SDL 3
 
 class "App"
 
--- a/CorsixTH/Lua/app.lua
+++ b/CorsixTH/Lua/app.lua
@@ -28,7 +28,7 @@ local SDL = require("sdl")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 265 -- SDL 3
+local SAVEGAME_VERSION = 265 -- Deferred entity destruction fix
 
 class "App"
 
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -1851,14 +1851,7 @@ function World:destroyEntity(entity)
     -- entity from that table to avoid skipping entities in the loop.
     -- The entity is removed as soon as the iteration has finished.
     entity.to_destroy = true
-    -- Loaded games created before the deferred destruction existed will
-    -- not have an entities_to_destroy table, so create it lazily.
-    local queue = self.entities_to_destroy
-    if not queue then
-      queue = {}
-      self.entities_to_destroy = queue
-    end
-    queue[#queue + 1] = entity
+    self.entities_to_destroy[#self.entities_to_destroy + 1] = entity
     entity:onDestroy()
   else
     for i, e in ipairs(self.entities) do
@@ -1876,9 +1869,7 @@ end
 --! self.entities, so this must only be called once that iteration finished.
 function World:_flushDestroyedEntities()
   local queue = self.entities_to_destroy
-  if not queue or #queue == 0 then
-    -- A missing queue means a game saved before deferred destruction was
-    -- added was loaded, in which case there is nothing to flush.
+  if #queue == 0 then
     return
   end
   self.entities_to_destroy = {}
@@ -2947,6 +2938,11 @@ function World:afterLoad(old, new)
   self:previousSpeed()
 
   self.earthquake:afterLoad(old, new)
+
+  if old < 265 then
+    -- Deferred entity destruction queue (added in #1467 fix)
+    self.entities_to_destroy = {}
+  end
 
   -- Savegame version housekeeping
   if not self.original_savegame_version then
--- a/CorsixTH/Lua/world.lua
+++ b/CorsixTH/Lua/world.lua
@@ -60,6 +60,7 @@ function World:World(app, free_build_mode)
   self.pathfinder = TH.pathfinder()
   self.pathfinder:setMap(app.map.th)
   self.entities = {} -- List of entities in the world.
+  self.entities_to_destroy = {} -- Entities queued for destruction while iterating self.entities.
   self.dispatcher = CallsDispatcher(self)
   self.objects = {}
   self.objects_notify_occupants = {}
@@ -968,12 +969,13 @@ function World:onTick()
         end
       end
       for _, entity in ipairs(self.entities) do
-        if entity.ticks then
+        if entity.ticks and not entity.to_destroy then
           self.current_tick_entity = entity
           entity:tick()
         end
       end
       self.current_tick_entity = nil
+      self:_flushDestroyedEntities()
       self.map:onTick()
       self.map.th:updateTemperatures(outside_temperatures[self.game_date:monthOfYear()],
           0.25 + self.hospitals[1].heating.radiator_heat * 0.3)
@@ -1051,14 +1053,16 @@ end
 function World:onEndDay()
   local local_hospital = self:getLocalPlayerHospital()
   for _, entity in ipairs(self.entities) do
-    if entity.ticks and class.is(entity, Humanoid) then
+    if entity.ticks and class.is(entity, Humanoid) and not entity.to_destroy then
       self.current_tick_entity = entity
       entity:tickDay()
-    elseif class.is(entity, Plant) then
+    elseif class.is(entity, Plant) and not entity.to_destroy then
+      self.current_tick_entity = entity
       entity:tickDay()
     end
   end
   self.current_tick_entity = nil
+  self:_flushDestroyedEntities()
 
   --check if it's time for a VIP visit
   if self.game_date:isSameDay(self.next_vip_date) then
@@ -1173,12 +1177,13 @@ function World:onEndMonth()
 
   self:makeAvailableStaff(self.game_date:monthOfGame())
   for _, entity in ipairs(self.entities) do
-    if entity.checkForDeadlock then
+    if entity.checkForDeadlock and not entity.to_destroy then
       self.current_tick_entity = entity
       entity:checkForDeadlock()
     end
   end
   self.current_tick_entity = nil
+  self:_flushDestroyedEntities()
 end
 
 -- Called when a month ends. Decides on which dates patients arrive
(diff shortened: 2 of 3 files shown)


BrunosGits:
@lewri  @TheCycoONE 

I created a clean pull, the old one was too dirty with a lot of weird files.

Love TH, grew up playing it and I'd like to contribute to this project so others can continue enjoying this game and have as much fun as I had and can hear pearls like: 

_Patients are asked not to die in the corridors_

CorsixTH/Lua/world.lua
```diff
@@ -1837,13 +1842,54 @@ function World:newEntity(class, animation, mood_marker)
...
+  end
+  if self.current_tick_entity then
+    -- The caller is iterating over self.entities, so defer removing the
+    -- entity from that table to avoid skipping entities in the loop.
+    -- The entity is removed as soon as the iteration has finished.
+    entity.to_destroy = true
+    -- Loaded games created before the deferred destruction existed will
+    -- not have an entities_to_destroy table, so create it lazily.
+    local queue = self.entities_to_destroy
+    if not queue then
```
TheCycoONE: This guard seems unnecessary, the queue is an invariant of the world class (or would be if it was also created in afterLoad)

CorsixTH/Lua/world.lua
```diff
@@ -1837,13 +1842,54 @@ function World:newEntity(class, animation, mood_marker)
...
+function World:_flushDestroyedEntities()
+  local queue = self.entities_to_destroy
+  if not queue or #queue == 0 then
+    -- A missing queue means a game saved before deferred destruction was
+    -- added was loaded, in which case there is nothing to flush.
+    return
+  end
+  self.entities_to_destroy = {}
+  for i = #self.entities, 1, -1 do
+    if self.entities[i].to_destroy then
```
TheCycoONE, line 1877: Seems that you lost the break, entities only appear in the list once.

TheCycoONE:
smoketest.lua is a new sort of thing, I'm not entirely sure it belongs in the repo, if it does it needs a different name and instructions on running it and what it does.

It also looks very brittle which might be a reason not to include it for now. I am interested in exploring better automated end to end testing which this was an attempt to do; but less slop and more framework.

BrunosGits:
Hi, wanted to clarify my thinking on the break. I left it out because I assumed _flushDestroyedEntities needs to remove all queued entities in one pass, not just the last one.

Say during a single tick e2.tick() destroys both e1 and e4, like a room crashing and taking its objects with it. Both would have to_destroy = true. Would a break here remove only e4 and leave e1 stuck in the list?

I iterated backwards so that table.remove doesn't shift indices of elements I haven't scanned yet. Does that hold, or am I off on something?

lewri:
Iterating backwards is a good way to go to prevent skipping items if we're removing things from the reference table of the loop E.g.
https://github.com/CorsixTH/CorsixTH/pull/3403/changes/88c635392d94f7d07f89f1eedc341b579a712241

CorsixTH/Lua/world.lua
BrunosGits: For the current status it's unnecessary, but if we remove the guard it would break loading old savegames.

lewri: I see.
Instead you need to add the value in an afterload and bump the savegame version.

BrunosGits:
I studied the codebase and the save/load migration practices (following the patterns in `Lua/world.lua:World:afterLoad`, `Lua/app.lua:App:afterLoad`, and `Lua/persistance.lua`), and implemented the cleaner pattern you suggested (thanks @lewri for the guidance on `afterLoad` + version bump).

## Changes in this update:

**`CorsixTH/Lua/app.lua`**
- Bumped `SAVEGAME_VERSION` from 264 \u2192 265

**`CorsixTH/Lua/world.lua`**
- Added `entities_to_destroy = {}` initialization in `World:afterLoad(old, new)` for `old < 265` (following the established migration pattern after the bench count at v47 in `Lua/world.lua`)
- Removed lazy queue creation in `destroyEntity()` \u2014 the queue is now guaranteed to exist
- Removed lazy guard in `_flushDestroyedEntities()` \u2014 the invariant is now maintained by constructor + `afterLoad`

**`CorsixTH/Luatest/spec/world_spec.lua`**
- Removed 2 tests that expected the old lazy behavior (queue missing / lazy creation)
- All 84 tests pass

## Validation

All checks pass:
- 84/84 unit tests (busted)
- luacheck clean (297 files)
- Build succeeds
- Offscreen, xvfb, and demo mode all load without crash
- Whitespace check clean (425 files)

The queue is now a true invariant: created in `World:World()` for new games, initialized in `afterLoad` for old saves. No more lazy guards in hot paths.

TheCycoONE (approved):

CorsixTH/Lua/app.lua
```diff
@@ -28,7 +28,7 @@ local SDL = require("sdl")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 264 -- SDL 3
+local SAVEGAME_VERSION = 265 -- SDL 3
```
lewri: Please update the comment on the line to reference the bump (e.g. Prevent skipping entities during destruction)

CorsixTH/Lua/world.lua
```diff
@@ -2902,6 +2939,11 @@ function World:afterLoad(old, new)
 
   self.earthquake:afterLoad(old, new)
 
+  if old < 265 then
+    -- Deferred entity destruction queue (added in #1467 fix)
```
lewri, line 2943: The issue number isn't critical here but there's no harm in keeping it.

CorsixTH/Lua/world.lua
```diff
@@ -1837,13 +1842,45 @@ function World:newEntity(class, animation, mood_marker)
...
 
 function World:destroyEntity(entity)
-  for i, e in ipairs(self.entities) do
-    if e == entity then
+  if entity.to_destroy then
+    -- Already queued for destruction, so onDestroy has already run.
+    return
+  end
+  if self.current_tick_entity then
+    -- The caller is iterating over self.entities, so defer removing the
```
lewri, line 1850: This could potentially be shorter. Or it might be useful to look at if the function definition at the top could absorb any of this, as it's not currently documented.

CorsixTH/Lua/world.lua
```diff
@@ -1837,13 +1842,45 @@ function World:newEntity(class, animation, mood_marker)
 end
 
 function World:destroyEntity(entity)
-  for i, e in ipairs(self.entities) do
-    if e == entity then
+  if entity.to_destroy then
+    -- Already queued for destruction, so onDestroy has already run.
```
lewri, line 1846: Just the first half of this sentence is needed I think.

lewri (dismissed):
Requested change only relates to updating the savegame bump line in `app.lua`. The rest are up to you.

CorsixTH/Lua/world.lua
```diff
@@ -1173,12 +1177,13 @@ function World:onEndMonth()
...
   self:makeAvailableStaff(self.game_date:monthOfGame())
   for _, entity in ipairs(self.entities) do
-    if entity.checkForDeadlock then
+    if entity.checkForDeadlock and not entity.to_destroy then
       self.current_tick_entity = entity
       entity:checkForDeadlock()
     end
   end
   self.current_tick_entity = nil
+  self:_flushDestroyedEntities()
```
lewri, line 1186: We only flush once a month? I'm not sure if we have a benefit waiting so long. Did you have a reason for this interval level?

lewri (commented):
Cool adding some luatests btw

CorsixTH/Lua/world.lua
BrunosGits: I understand why it looks like we only flush monthly at first glance. But I see that the monthly flush specifically handles entities destroyed during the deadlock-check loop in onEndMonth(). When I iterate self.entities for deadlock checks, current_tick_entity gets set, so destroyEntity() defers removal via the to_destroy flag. I need that flush right after the loop to clean them up before month-end systems run. 

The tick- and day-level flushes cover regular gameplay. This one ensures no deferred destructions linger into monthly stats or VIP checks. If I take it out, deadlock-test scenarios break. 

I've validated this across all test modes. Happy to show you a test case if it'd help. It's not redundant.

TheCycoONE (approved):
Still approved. It's clear to me. I'll be away so lewri can merge.

CorsixTH/Lua/world.lua
lewri: Yep sorry missed the other one

PR #3506, opened 2026-08-20
[RFC] Object happiness factor calculated from room contents

tobylane:
- Add to the room happiness factor the effect from the room objects. Rely on this factor to add to happiness when in a room. This means plants and litter outside the room don't count, a major change from before. The aim is to reduce the number of object searches. Is this acceptable? Also clarify that happiness factor only applies to staff.
- Add a check for indoors before calculating patient happiness, as plants, litter, etc are only indoors.

### DIFF
--- a/CorsixTH/Lua/room.lua
+++ b/CorsixTH/Lua/room.lua
@@ -1084,9 +1084,6 @@ function Room:afterLoad(old, new)
     -- no longer using this so empty it
     self.humanoids_enroute = {}
   end
-  if old < 186 then
-    self:calculateHappinessFactor()
-  end
   if old < 233 then
     if self.waiting_staff_member then
       -- Cancel delayed replace existing staff member in room
@@ -1104,6 +1101,9 @@ function Room:afterLoad(old, new)
         room_name ~= "research" then
       self.staff_member_set = nil
     end
+  end
+  if old < 265 then
+    self:calculateHappinessFactor()
   end
 end
 
@@ -1180,10 +1180,11 @@ function Room:calculateRemovalCost()
   return cost
 end
 
---! Calculate the effect the room has on humanoid happiness
+--! Calculate the effect the room has on staff happiness
 function Room:calculateHappinessFactor()
   -- The number of windows affects happiness
-  local window_factor, space_factor, window_count = 0, 0, self:countWindows()
+  local window_factor, space_factor, object_factor = 0, 0, 0
+  local window_count = self:countWindows()
   if window_count > 0 then
     if self.room_info.id == "staff_room" then
       -- Staff are pleased to rest in a staff room with windows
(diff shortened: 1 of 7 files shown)


CorsixTH/Lua/room.lua
```diff
@@ -1200,7 +1201,25 @@ function Room:calculateHappinessFactor()
...
 
-  self.happiness_factor = window_factor + space_factor
+  -- Various optional room items affect happiness
+  local room_objects = {
+    extinguisher = 0.002,
+    bin          = 0.001,
+    bookcase     = 0.003,
+    skeleton     = 0.002,
+    tv           = 0.0005,
+    plant  = function(plant) return -0.003 + (plant:isPleasingFactor() * 0.001) end,
```
ARGAMX, line 1211: Pure magic.
Although this is not you code, you just moved this code, and so this is not related to your PR, I realised that there are magic in this code.

First, there are some other numbers in the `Plant` class that, along with `0.003`, provide context and understanding of what `0.003` is and why exactly it's `0.003`.
Without them, `0.003` seems like a magic number.

I looked at
`Plant:isPleasingFactor()`
and
`Plant.phases = 5`
and understood why it's `0.003`.
(1 - is a wilted plant, 3 - is the middle state, 5 - is a watered, fragrant flower)

Let's start with the fact that this is actually not `0.003`, but `3 * 0.001`.

Next, I think the number `3` should be in the `Plant` class.
Otherwise, right now we have with the `Room` class knowing how many states a plant has and what the middle state is.
The `Room` class shouldn't know about it.

ARGAMX: I think we need something like this:

```Lua
function Plant:pleasingFactor2()
  local middle_phase = 3
  return self:isPleasingFactor() - middle_phase
end
```

```Lua
  plant  = function(plant) return plant:pleasingFactor2() * 0.001 end,
```

(I haven't come up with a name for the `pleasingFactor2` function. Of course, `pleasingFactor2` is not a suitable name, don't use it please)

ARGAMX:
For now, patients will still be searching for objects in the room the old-fashioned way using findObjectsInSquare, right?

ARGAMX (commented):
I think the PR is a bit lacking context. 🙂
You've described the changes you're making.

But the most important question is:
Why are you doing this?
What's your motivation?

Without understanding this, it's difficult to analyze this PR.
You know this. But the reader doesn't.

I myself have some general context regarding happiness factor and `findObjectsInSquare`.
But I didn't immediately grasp the idea behind it.

After thinking about it, I've come to the conclusion that your main motivation is probably optimisation?

So, to prevent such things from happening, I see two good solutions:
1. Create an Issue and describe all your pain points, ideas, and proposed solutions. Describe the current behavior. Describe the desired behavior. Moreover, at this stage, the proposed solution can be refined collaboratively. Otherwise, if we write the code and create a PR right away, it might turn out later, during the PR stage, that not everyone likes this approach.
2. Or if we don't want to create an issue, we can place this entire description here in the PR. The description that follows the template we have for Issues (how to reproduce, what version, how it works now, how it should work, what worked before). I remember Lewri wasn't particularly fond of this. But I think it's at least better than providing no context at all, or insufficient context.

Now I think it can be format it a little better. 🙂

It's especially difficult when you're studying the history—old changes, old PRs. I often dig through old ones now. And sometimes nothing is written at all about why the change was made, what the motivation was. They just did it and that's it! Especially if it is a deviation from the original.

And sometimes, when problems with this behavior arise, you have to decide whether to remove it or change it somehow.
But to remove it, you need to understand why it was added. And this might be missing...

ARGAMX:
I've just reread that  again, and I think it's correct, and that's the main goal, as you mentioned:

> The aim is to reduce the number of object searches

tobylane:
I hadn't thought through the bigger picture, and this was an attempt to pick off a smaller piece of it. But the first commit is wrong to deviate from TH97. The idea behind the second one may be useful in another form. I'll plan better and write more about it.

Patients can't have any change in happiness when they are in a room, I only learned this in writing the first commit.

The very small numbers are magic numbers based on old interpretations of what TH97 did. They could be put in the class, some of them I hope to replace with level config values and defaults.

PR #3521, opened 2026-08-31
Add rats

### LINKED ISSUE #3520: Implement rats: shootable vermin from ratholes

Add rats + shooting rats.

Later: Dead rats + patient vomit, year-end rat trophies, and the bonus rat level.

It restores a fun piece of the original game and rewards players for keeping a clean hospital.

Yes

master 99cb578b (v0.70.1)

Picks up the old #328 (WIP Rats, closed 2016) and touches on #1604 (rathole placement).

### Additional Information

Lua-only for the first steps.

gojomoso1:
Refs #3520

- Adds shootable rats that travel between real ratholes along corridors.
- Supports hidden east/south holes, clean-hospital spawning, and activity scaling by hospital size and litter.
- Debug action creates two visible holes and spawns from a newly created hole.
- Crosshair follows the rendered rat position; street kills leave no splat.

Builds on @TheCycoONE's work.

Trophies, diagonal pathing, trigger tuning, and the bonus level remain follow-up work.


CorsixTH/Lua/entities/rat.lua
```diff
@@ -0,0 +1,302 @@
+--[[ Copyright (c) 2021 Stephen "TheCycoONE" Baker
```
lewri, line 1: Review to come later but as this is a new file and one of your commits added changes to it it's appropriate on this file to also add your authorship here :)

ARGAMX:
Looks nice

[Screen Recording 2026-08-31 at 13.57.33.webm](https://github.com/user-attachments/assets/a37322ab-fa56-47f1-addb-76f25ab6b23d)

ARGAMX:
1.
I am not sure if this is by design or not, but when press “make debug rat”, the rat can appear in a place where it shouldn’t, simply out of nowhere. A hole appears, rat comes out, and then the hole disappears. Probably, the hole is just part of the animation. But the point is that the mouse spawns not from a wall, but from thin air.

Screenshot 2026-08-31 at 14 10 40

.
Updated: Resolved.

ARGAMX:
2.
The algorithm for determining whether to display a crosshair or not generally works, but it can probably be improved.
Sometimes the rat is very close to the cursor, but it doesn’t react in any way and there is no crosshair. And sometimes the rat is very far from the cursor, but the cursor already reacts and turns into a crosshair.
That is, visually, the algorithm for changing the cursor doesn’t take into account the pixel where the rat’s animation is actually displayed.

Screenshot 2026-08-31 at 14 16 23

.
Updated: Resolved.

CorsixTH/Lua/entities/rat.lua
```diff
@@ -0,0 +1,302 @@
...
+-- must equal 8 or the movement glitches (see humanoid_actions/walk.lua, where the
+-- same 2x is the "fast" preset).
+local walk_factor = 2
+local walk_quantity = 4
+
+-- Rats don't head straight home; they dart to a few random nearby tiles first
+-- before finally running into a hole.
+local min_wander_legs = 3
+local max_wander_legs = 7
+local wander_radius = 5 -- tiles from the current position for a random dart
```
ARGAMX: Why rats don't head straight home?
As far as I remember from the original game, I believe they move along the shortest valid path between the two holes. No strolls.

CorsixTH/Lua/entities/rat.lua
```diff
@@ -0,0 +1,302 @@
...
+
+local rat_bounty = 5 -- cash awarded for shooting a rat
```
ARGAMX: Why and for what reason is a monetary reward given for killing a rat?

CorsixTH/Lua/hospital.lua
```diff
@@ -693,47 +693,139 @@ function Hospital:_fixBoiler()
...
+  local active = 0
+  for _, entity in ipairs(self.world.entities) do
+    if class.is(entity, Rat) then active = active + 1 end
+  end
+  if active >= max_active_rats then return end
+
+  local litter = self.world.map.th:getLitterFraction(self:getPlayerIndex())
+  if math.random() < litter * rat_spawn_chance then
+    self:spawnRat()
+  end
```
ARGAMX, line 837: In the original game, rats run around quite often in a completely clean hospital.
With this condition in CorsixTH, we won’t see rats in a completely clean hospital? If so, that’s sad.

Although it would be fun if rats spawn chance would depended on the amount of trash. But it would be nice if there was some minimum threshold that would give rats a chance to appear even in a clean hospital.

ARGAMX:
3. 
In this PR If a rat is shot on the street, it leaves a bloody mark on the street.
Although in the original, the mouse left nothing on the street after being killed.

Screenshot 2026-08-31 at 21 38 50

.
Updated: Resolved.

ARGAMX:
Zoom certainly has an advantage in shooting at rats.

Screenshot 2026-08-31 at 21 45 55

CorsixTH/Lua/hospital.lua
```diff
@@ -693,47 +693,139 @@ function Hospital:_fixBoiler()
...
+--!return (Rat or nil) The spawned rat, or nil if there is no hole to use.
+function Hospital:spawnRat(hole)
+  if not hole then
+    -- Emerge from a random north-wall hole (only north has a leaving anim);
+    -- fall back to any hole if there are no north ones.
+    local north_holes = {}
+    for _, h in ipairs(self.ratholes) do
+      if h.wall == "north" then north_holes[#north_holes + 1] = h end
+    end
+    local pool = #north_holes > 0 and north_holes or self.ratholes
```
ARGAMX: Do rats emerge only from the northern holes?
And do they only enter the northern and western holes?

I’m sure that in the original game, rats emerged and entered, among also the eastern and southern holes. But it was impossible for player to see them in the original game. And, accordingly, the animations weren’t visible either. But such holes did exist.

ARGAMX: Screenshots from original game:

Image

.
Image

.
Image

ARGAMX: Since the player can’t see whether there’s a hole in the southern and eastern walls, I’m wondering if the algorithm could treat absolutely any southern or eastern wall as a potential hole?

CorsixTH/Lua/hospital.lua
```diff
@@ -693,47 +693,139 @@ function Hospital:_fixBoiler()
...
+    end
+  end
+  self:spawnRat()
+end
+
+local function dailyUpdateRatholes(self)
+  local th = self.world.map.th
+
+  local wanted_holes = math.round(th:getLitterFraction(self:getPlayerIndex()) * 200)
+  if #self.ratholes < wanted_holes then -- Not enough holes, find a new spot
```
ARGAMX, line 792: I also doubt that the number of holes depends on the level of pollution.
I can provide a save from the original, where there are plenty of holes but almost no litter.
And if you remove the old holes, you’ll see new ones appearing.
I believe the number of max visible holes depends on the number of available surfaces. 
Not on the amount of litter.

CorsixTH/Lua/hospital.lua
```diff
@@ -693,47 +693,139 @@ function Hospital:_fixBoiler()
...
+    for _ = 1, math.min(12, 6 + 2 * (wanted_holes - #self.ratholes)) do
+      if self:tryAddRathole() then break end
+    end
+  end
+end
+
+-- Chance per day of a rat emerging, scaled by how littered the hospital is, up
+-- to a small cap on simultaneously active rats.
+local rat_spawn_chance = 0.5
+local max_active_rats = 4
```
ARGAMX: At the tiny level and at the huge level, in both cases, can a maximum of 4 rats be present at the same time?
Why shouldn’t the number depend on the size of the hospital?
I don’t know how it works in the original, but maybe it would make sense here to take the hospital’s pollution as one of the factors?

CorsixTH/Lua/world.lua
```diff
@@ -2719,6 +2720,11 @@ function World:afterLoad(old, new)
       obj:afterLoad(old, new)
     end
   end
+  -- update entity map if it exists, newly created one for saves older than 88
+  -- will be up to date
+  if self.entity_map then
+    self.entity_map:afterLoad(old, new)
+  end
```
ARGAMX: I’m curious—what does this code do, and why did it appear at this particular point in the `afterLoad` method? Why is it saying 88? Why isn’t it wrapped in a version number then?
It looks mysterious.

CorsixTH/Lua/entities/rat.lua
gojomoso1: @ARGAMX - for some reason I remembered it doing that. But after reloading the original game I realized it doesn't. Will remove.

CorsixTH/Lua/hospital.lua
```diff
@@ -693,47 +693,139 @@ function Hospital:_fixBoiler()
...
+  local result = {}
+  for i = 1, math.min(count, #candidates) do result[i] = candidates[i] end
+  return result
+end
+
+--! Debug helper: generate two ratholes (A and B) near the given point and spawn
+-- a rat that emerges from A and runs to B.
+--!param near_x (integer) X coordinate near which to place the holes.
+--!param near_y (integer) Y coordinate near which to place the holes.
+function Hospital:makeDebugRat(near_x, near_y)
```
TheCycoONE, line 788: Function appears to differ from it's description, with respect to the number of holes and the behaviour when holes already exist in the hospital.

CorsixTH/Lua/hospital.lua
TheCycoONE: This isn't new in this PR, just moved. Can be considered out of scope to change.

TheCycoONE:
Thanks, I'm glad to see my earlier work cleaned up and completed to a usable / value add state.

There are a number of things beyond the scope of this PR that I am content to leave that way:
* In the original rats have their own path finder or path simplification algorithm that isn't restricted to 4 directional movement
* Trophies
* Endless tweaking to the trigger conditions
* Bonus level

I'm just back home so I'll provide a more detailed review in the coming days.

CorsixTH/Lua/hospital.lua
TheCycoONE: After testing, I'd prefer this did what it said, 2 holes not 3, and create them on the screen even if there are existing holes off screen. Maybe we don't need to create additional holes if there are already 2 on the screen 🤷

CorsixTH/Lua/hospital.lua
```diff
@@ -693,47 +693,148 @@ function Hospital:_fixBoiler()
...
+  -- Ensure a few holes exist near the view so rats emerge from varied places.
+  if #self.ratholes < 3 then
+    for _, t in ipairs(self:findRatholeTiles(3, near_x, near_y)) do
+      self:addRathole(t.x, t.y, t.wall, t.parcel)
+    end
+  end
+  self:spawnRat()
```
TheCycoONE, line 785: You want to pass in one of the new holes if you created any.

TheCycoONE:
Why did you refactor the rathole creation code?

CorsixTH/Lua/hospital.lua
TheCycoONE: The original game had defined holes on those walls; it's not clear to me that it ever used them to cheat a missing hole. This is important as the paths were predictable if you paid attention. The PR seems to implement it that way except with a cheat to use any in some situations that I hope the author can better describe.

TheCycoONE:
> The algorithm for determining whether to display a crosshair or not generally works, but it can probably be improved. Sometimes the rat is very close to the cursor, but it doesn’t react in any way and there is no crosshair. And sometimes the rat is very far from the cursor, but the cursor already reacts and turns into a crosshair. That is, visually, the algorithm for changing the cursor doesn’t take into account the pixel where the rat’s animation is actually displayed.

This was a problem in my implementation. The empty crosshair is based on the tile rather than centered on the point of the rat. I don't remember if I confirmed the original was the same or not. This check is a lot more efficient, otherwise we have to pull 9 or at least 4 tiles around the mouse cursor and calculate the distance to any rats within the tick. Worth profiling.

ARGAMX:
> > The algorithm for determining whether to display a crosshair or not generally works, but it can probably be improved. Sometimes the rat is very close to the cursor, but it doesn’t react in any way and there is no crosshair. And sometimes the rat is very far from the cursor, but the cursor already reacts and turns into a crosshair. That is, visually, the algorithm for changing the cursor doesn’t take into account the pixel where the rat’s animation is actually displayed.
> 
> This was a problem in my implementation. The empty crosshair is based on the tile rather than centered on the point of the rat. I don't remember if I confirmed the original was the same or not. This check is a lot more efficient, otherwise we have to pull 9 or at least 4 tiles around the mouse cursor and calculate the distance to any rats within the tick. Worth profiling.

Apparently, the original does not rely on tiles. Since the crosshair sometimes is displayed on neighboring tiles too.
https://www.youtube.com/watch?v=r-gXWUnn7eg

Of course, I’m not insisting that we do this in this PR.
Just to understand what can still be improved in future.
I think this PR is already quite functional.

TheCycoONE:
> Apparently, the original does not rely on tiles. Since the crosshair sometimes is displayed on neighboring tiles too. https://www.youtube.com/watch?v=r-gXWUnn7eg
> 
> Of course, I’m not insisting that we do this in this PR. Just to understand what can still be improved in future. I think this PR is already quite functional.

I had the opposite interpretation from that video, the distance of the cursor turning to a crosshair from a rat seems very inconsistent and it's hard to tell from a video which tile the rat is associated with.

ARGAMX:
> > Apparently, the original does not rely on tiles. Since the crosshair sometimes is displayed on neighboring tiles too. https://www.youtube.com/watch?v=r-gXWUnn7eg
> > Of course, I’m not insisting that we do this in this PR. Just to understand what can still be improved in future. I think this PR is already quite functional.
> 
> I had the opposite interpretation from that video, the distance of the cursor turning to a crosshair from a rat seems very inconsistent and it's hard to tell from a video which tile the rat is associated with.

However, I can’t reproduce that behavior in the original game — when I move the cursor right in front of the mouse, but there’s no crosshair.

Screenshot 2026-08-31 at 14 16 23

CorsixTH/Lua/hospital.lua
gojomoso1: Removed the fallback. Rats now spawn from and target only defined ratholes, including hidden east/south holes. If no distinct destination hole or valid corridor path exists, the rat despawns.

CorsixTH/Lua/hospital.lua
gojomoso1: Done.

CorsixTH/Lua/hospital.lua
gojomoso1: Done.

gojomoso1:
Crosshair: Fixed. Detection now uses the rat's rendered pixel position. A 5x5 candidate scan is required to cover the full 24-pixel radius while a rat crosses between tiles; a local synthetic profile measured about 1.76 microseconds per call.

Street splat: Fixed.

Normal rathole creation refactor: Reverted.

gojomoso1:
> > > Apparently, the original does not rely on tiles. Since the crosshair sometimes is displayed on neighboring tiles too. https://www.youtube.com/watch?v=r-gXWUnn7eg
> > > Of course, I’m not insisting that we do this in this PR. Just to understand what can still be improved in future. I think this PR is already quite functional.
> > 
> > 
> > I had the opposite interpretation from that video, the distance of the cursor turning to a crosshair from a rat seems very inconsistent and it's hard to tell from a video which tile the rat is associated with.
> 
> However, I can’t reproduce that behavior in the original game — when I move the cursor right in front of the mouse, but there’s no crosshair.
> 
> Screenshot 2026-08-31 at 14 16 23

I reverted the unrelated normal rathole-creation refactor.

Crosshair detection now uses the rat’s rendered pixel position instead of its tile. Went with a 5x5 scan.

CorsixTH/Lua/hospital.lua
```diff
@@ -693,13 +693,116 @@ function Hospital:_fixBoiler()
...
+  local emerge_hole
+  local needed = math.max(0, 2 - #visible_holes)
+  for _, tile in ipairs(self:findRatholeTiles(needed, left, top, right, bottom)) do
+    self:addRathole(tile.x, tile.y, tile.wall, tile.parcel)
+    local hole = self.ratholes[#self.ratholes]
+    visible_holes[#visible_holes + 1] = hole
+    emerge_hole = emerge_hole or hole
+  end
+  if not emerge_hole and #visible_holes > 0 then
+    emerge_hole = visible_holes[math.random(1, #visible_holes)]
```
TheCycoONE, line 793: Assumes that there is at least one candidate hole?

ARGAMX (approved):
Works nice for me 
Good job 👍

TheCycoONE:
To be clear I'm waiting for some response to my last observation; it looks like a bug.

CorsixTH/Lua/hospital.lua
gojomoso1: good catch - fixed.

TheCycoONE (approved):

ARGAMX:
The rats functionality is a very important feature that was present in the original game but has been missing from CorsixTH for so long — people have probably been waiting for more than 15 years.

Of course, we still have to add a rat level to the game.
But just the appearance of rats in the master is a big event. 🎉
I’ll open a bottle of champagne 🍾 today to celebrate the fact that we’ve merged this. 
I am seriously!

Thank you @gojomoso1 and of course credits to @TheCycoONE! 🙏

chuck-norris-approves-this-pr

PR #3522, opened 2026-08-31
Expel walk humanoids from construction site

### LINKED ISSUE #3505: Expel humanoids walking in the construction site
----

### Describe the issue

When player build a room, and at that moment there is a humanoid with the "walk" action on the construction site, then this humanoid will continue to walk on its way, and the player will have to patiently wait until the humanoid leaves the construction site.
In the original game, interrupts were sent out by humanoids and they were expelled from the construction site along the shortest possible route.

With #3414 that is possible now.

----

### CorsixTH Version

0.70.0

----

### Steps to Reproduce

1. Try building a large room so that one of the humanoids is close to the blueprint wall and has its back turned to it.
2. Try to move to the next construction stage.
3. Notice that the humanoid, instead of turning around and leaving the blueprint via the shortest path, continues on its way.

----

### Expected Behaviour

Humanoids should be expelled from the construction site along the shortest possible route as in the original game.

----

CorsixTH behaviour:

https://github.com/user-attachments/assets/27eed6b0-ccfd-4b75-84d3-43378d6d9b27

----

Vanilla TH behaviour:

https://github.com/user-attachments/assets/5335e3a0-6fde-4282-9497-643c2f5c8011

----

ARGAMX:
*Fixes #3505*

- Interrupt walking humanoids along construction site and expel them along the shortest possible route (Thanks to 3414 PR).

### DIFF
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -883,6 +883,9 @@ function Humanoid:leaveArea()
     else
       meander.done_walk = false
     end
+  elseif current_action.name == "walk" then
+    -- Make the humanoid to reroute urgently to leave the blueprint
+    self:queueAction(MeanderAction():setCount(1):setMustHappen(true), 0)
   else
     -- Look for a queue action and re-arrange the people in it, which
     -- should cause anyone queueing within the blueprint to move


CorsixTH/Lua/entities/humanoid.lua
```diff
@@ -883,6 +883,9 @@ function Humanoid:leaveArea()
     else
       meander.done_walk = false
     end
+  elseif current_action.name == "walk" then
+    -- Make the humanoid to reroute urgently to leave the blueprint
+    self:queueAction(MeanderAction():setCount(1):setMustHappen(true), 0)
```
lewri, line 888: I think I also concluded this was the answer before the more sweeping tile changes got made. Mine just didn't work without those changes :D

TheCycoONE (approved):

Issue #2759, opened 2025-01-14
[Bug] Fail on savegame load `C stack overflow`.


ARGAMX:
### Describe the issue

Fail on savegame load.

<img width="331" alt="Screenshot 2025-01-12 at 17 34 55" src="https://github.com/user-attachments/assets/8498204e-8029-4870-9acf-d624f65038eb" />

How 'C stack overflow' was obtained:
Make a lot vomit at one title.

<img width="425" alt="C_stack_overflow" src="https://github.com/user-attachments/assets/d9747886-448b-48f7-9041-f8fc06b56d79" />

Currently, the code has a condition that each patient can only vomit once (`numVomit > 0`). I removed this condition, so each patient could vomit many times. Then I loaded the game with completed hospital, deleted 4 of the 5 office GPs, thereby creating a large queue, fired all Handyman from that block with GP, so there is no one to clean up the vomit, set the maximum speed and only periodically closed the annual report.
After a two years of game time, maybe more that two, I tried to load autosave and that's when I got this error.

https://github.com/CorsixTH/CorsixTH/blob/32bdd626b7d63949d01f5a8b357b11b79108e9b4/CorsixTH/Lua/entities/humanoids/patient.lua#L702-L706

### Steps to Reproduce

1. *Load save game attached to this ticket*
2. Observe that it can't be loaded.

### Expected Behaviour

This shouldn't have happened.
I think that the same type of pollution should not be superimposed on top of the same title.

### Have you enabled any current CorsixTH Experimental Features?

- [ ] Right mouse panning
- [ ] Falling actions

### Save Game

[Loading_Error_C_stack_overflow.sav.zip](https://github.com/user-attachments/files/18389770/Loading_Error_C_stack_overflow.sav.zip)

### CorsixTH Version

0.68.0

### Operating System

MacOS Sonoma 14.5

### Theme Hospital Version

CD

TheCycoONE:
There isn't really a fix for this with the current save system except to constrain the number of objects allowed in the world or per tile as mentioned. I looked into it with the lua 5.4 update when initially the stack size was much lower, and while I made progress with saving I wasn't able to make the same progress with loading due to the order of the file.

FWIW the original game had a hard limit on the number of objects per tile.

Flavsditz:
Although I set a limit on objects per tile on the PR linked above (doesn't fix this issue... read on) this of course didn't fix the old savegame that was done before the fix was made. 
[...]

PR #3523, opened 2026-08-31
Make litter NOT stackable

### LINKED ISSUE #581: Suggestion: A litter replaces existing one on same square.
It would be more logic, that litter of same type will replace another one on same square. Vomit over vomit makes one bigger vomit, but not two layers of vomits.

It's weird that handyman cleans one place again and again.

And I think it will solve #388.

Flavsditz:
*Fixes #581*
This implements the proposal of #581 to disallow litter to be stacked.

- When a litter gets dropped it will check for the litter already in the tile
  - if the new litter is of higher precedence it will substitute the one already on the ground and remove the old one
  - else the old litter will remain and the new one will be removed  
- This way no stacking is happening. (Added tests to validate it)
- Precedence can be seen on `litter.lua` but: soot is highest, then bio-harzard (puke, dead_rat, pee - in that order) then common litter.
  - Soot can't be cleaned by janitors and is mostly around damaged machines so I figured the precedence made sense. 
  
Also added a check that would cleanup old saved games, i.e. removing the stacking there.


lewri:
@Flavsditz are you sure this pr isn't also closing/fixing #581?

Flavsditz:
> @Flavsditz are you sure this pr isn't also closing/fixing #581?

Not sure if this was a typo as the first line says that this would solve #581, so I am going to assume you meant #2759 (disregard the following if that is not the case):


Yes 100%.
The changes proposed here would remove any stacked objects on old savegames **that load** and of course not contain stacked items on new savegames. I also tested with the linked savegame on the issue.

The problem of #2759 is that when loading the animations of each tile the C++ algorithm loads them recursively. Because on the savegame linked on the PR there are so many litter instances on the same tile this leads to a long recursion line as in order to render the first, the current algorithm follows the object linked to that, and so on until it reaches the final linked object and then it starts to work its way back. 
It seems lua-c bindings set a limit on the stack size of about 200 and that recursion explodes that limit.

ARGAMX:
> > @Flavsditz are you sure this pr isn't also closing/fixing #581?
> 
> Not sure if this was a typo as the first line says that this would solve #581, so I am going to assume you meant #2759 (disregard the following if that is not the case):
> 
> Yes 100%. The changes proposed here would remove any stacked objects on old savegames **that load** and of course not contain stacked items on new savegames. I also tested with the linked savegame on the issue.
> 
> The problem of #2759 is that when loading the animations of each tile the C++ algorithm loads them recursively. Because on the savegame linked on the PR there are so many litter instances on the same tile this leads to a long recursion line as in order to render the first, the current algorithm follows the object linked to that, and so on until it reaches the final linked object and then it starts to work its way back. It seems lua-c bindings set a limit on the stack size of about 200 and that recursion explodes that limit.

Nice, thanks 👍
In that case, you can fill in topic which tasks this PR fixes (closes).
For example, this is done in tasks 3237, 3323, and 3462.
GitHub parses this and links the PR to the task.
.
But you left “Fixes #” empty.
Therefore, task 581 is not linked to this PR.
I think that’s the reason why Lewri asked you about it.
So, when we close this PR, task 581 won’t be automatically closed and will remain open.
But I think you just didn’t know about it. It is ok, not a big problem.
.
I think you can edit the message in the topic and add that issue 581 will be fixed. 
Then everything will align with what you wrote in last message. 👍

Flavsditz:
Perfect! Done!

CorsixTH/Lua/objects/litter.lua
```diff
@@ -42,12 +42,39 @@ litter_types["soot_floor"] = 3416
...
 litter_types[3] = 1898
 litter_types[4] = 1900
 
+-- Litter precedence to decide what is shown in tile
+local litter_precedence = {}
+
+-- Bio-hazard (highest)
+litter_precedence["puke"] = 4
+litter_precedence["dead_rat"] = 3
+litter_precedence["pee"] = 2
```
ARGAMX: The original game features 2 types of litter in the corridors:

Group 1: `soda_can`, `banana`, `paper`, `bottle`
Group 2: `puke`, `pee`, `dead_rat`

Litter from Group 1 can only be placed on a completely clean floor.

Litter from Group 2 can be placed on a previously littered floor and, when placed, replace any litter from Group 1 or Group 2.

This was mentioned in 581.
But the algorithm you implemented doesn’t seem to follow this.

The soot from the explosion can’t get into the corridor, so I’m not considering it.

Flavsditz: > Group 1 can only be placed on a clean tile

The algorithm would allow that as Group1 items have low precedence so it would not even show if it was placed on an unclean tile . (more on the precedence rules below)
The algorithm however does not fix any logic on the Patient side about where they will litter. If that change should be encompassed that this needs to be a wider change.

> Litter from Group 2 can be placed on a previously littered floor and, when placed, replace any litter from Group 1 or Group 2.

This is true for the algorithm as is. (more on the precedence rules below)

On the precedence: 
I did made all Group 1 with the same precedence with the idea that even if the AI for littering wasn't accounting for the "only on clean tiles" rule, this would still work as it wouldn't replace the original litter. (it is never of higher precedence).
For Group 2 however with the same precedence we would be able to account for the following on the ticket:
> Killing a mouse always replaces any other litter of all types in that title with blood. (Tested)
> Vomit always replaces any other litter in that square, including blood. (Tested)

So I introduced the ranking there. What indeed is an issue is that this, as is, doesn't support blood <> puke to be interchengeable as this would require different treatment for both. 
I can put that in but it does make the algortihm a bit quirkier. In my head if you had blod over puke you'd still see mostly puke... just redish :D

ARGAMX: > I can put that in but it does make the algortihm a bit quirkier. In my head if you had blod over puke you'd still see mostly puke... just redish :D

Well, ok, this makes sense.
Although this doesn’t match the original game, I’m having doubts it should match.
This option with high precedence for puke is probably acceptable.

ARGAMX: > The algorithm would allow that as Group1 items have low precedence so it would not even show if it was placed on an unclean tile . (more on the precedence rules below)
> The algorithm however does not fix any logic on the Patient side about where they will litter. If that change should be encompassed that this needs to be a wider change.

We don’t need to change anything at the patient level.
I meant that if a tile is already occupied by trash, then trash of type 1 won’t appear on it.
We don’t need to make the patient look for another tile to dump trash in.

CorsixTH/Lua/objects/litter.lua
```diff
@@ -78,8 +106,42 @@ function Litter:getWalkableTiles()
   return tiles
 end
 
+function Litter:getPrecedence(item)
+    local litter_type = litter_anim_to_type[self.animation_idx]
+    return litter_precedence[litter_type]
+end
```
ARGAMX: This function takes param `item` which never use?

CorsixTH/Lua/world.lua
```diff
@@ -2886,6 +2886,34 @@ function World:afterLoad(old, new)
     self.system_pause = nil
   end
 
+  if old < 264 then
```
ARGAMX: To use a number version in `afterLoad` first we need to bump that number in `app.lua` and then use that new value here.

CorsixTH/Lua/objects/litter.lua
```diff
@@ -78,8 +106,42 @@ function Litter:getWalkableTiles()
   return tiles
 end
 
+function Litter:getPrecedence(item)
+    local litter_type = litter_anim_to_type[self.animation_idx]
+    return litter_precedence[litter_type]
```
ARGAMX: 4 spaced indent instead of 2 spaces?

CorsixTH/Lua/objects/litter.lua
```diff
@@ -78,8 +106,42 @@ function Litter:getWalkableTiles()
...
+      -- List will contain this object as it was already registered so check for it
+      if tileObject.object_type.id == "litter" and tileObject ~= self then
+        local existing_type = litter_anim_to_type[tileObject.animation_idx]
+        local existing_precedence = existing_type and litter_precedence[existing_type] or 0
+        if existing_precedence >= litterToSetPrecedence then
+          -- If existing litter on the ground has higher precedence cleanup self and return early
+          self.world:removeObjectFromTile(self, self.tile_x, self.tile_y)
+          self.world:destroyEntity(self)
+          return
+        else
```
ARGAMX: After reading this code and looking into the `Patient` class, I realised that we first place the litter on a tile and only then check whether it would even be possible to do so.
We even place it on top of litter that is already there.

And only then, in the `setLitterType` function, we execute checks and remove litter if it turns out that we shouldn’t have placed it there.
The checks happen too late.

But this is clearly not the responsibility of the `setLitterType` function.
`setLitterType` should not check the tile for other litter and remove its own litter.

It’s obvious that if this litter can’t be placed here, then it shouldn’t be placed here at all.
Now I understand what you meant.
I think you should make changes to the `Patient` class.

We need a separate function that the patient will call — it will perform these checks before placing the litter.
And if litter can be placed, then the litter will be placed.
`setLitterType` shouldn’t delete anything.

I really don’t know exactly where this new function should be. Most likely in `world` or in `hospital`.

CorsixTH/Lua/world.lua
Flavsditz: The bump in the number is supposed to be made if it breaks a savegame. While my change doesn't break I've opted to re-use the number as it allows new saves to avoid going through this logic.
It was purely in terms of efficiency for future savegames

CorsixTH/Lua/objects/litter.lua
Flavsditz: I am new to the codebase so I might have been myopic on my choice. 
I will take a look of whats possible. 

I think I went there since it converged for all kinds of litter, thus covering Patient actions of small litter and puke/pee as well as rat blood.

CorsixTH/Lua/world.lua
lewri: If you are inserting an afterload you must bump the savegame version, please. The only exception is if your afterload runs every time.

CorsixTH/Lua/objects/litter.lua
Flavsditz: I did some diggin as well. 
Inside `patient.lua` there is already logic for random litter that prevents littering of small items to happen if the tile is not empty. In this case my code (as is) just adds a secondary guard.

For `humanoid_actions/vomit.lua` and `humanoid_actions/pee.lua` there is no such guard as this action is indiscriminate (when you have to go you have to go). I believe it should stay like this.

I do agree though that `setLitterType` feels weird for this, so I would propose to create the following:
- A new function like ` World:placeLitter(litter_type, x, y)` (not sure on the signature yet) that:
  - creates the object, 
  - runs precedence vs existing tile litter (destroy new / remove old)
  - adds handyman task. 

`setLitterType`  shrinks back to pure anim-setting, or precedence moves out entirely.

wdyt?

CorsixTH/Lua/world.lua
Flavsditz: Done

CorsixTH/Lua/objects/litter.lua
ARGAMX: > * A new function like ` World:placeLitter(litter_type, x, y)` (not sure on the signature yet) that:
>   
>   * creates the object,
>   * runs precedence vs existing tile litter (destroy new / remove old)
>   * adds handyman task.
> 
> `setLitterType` shrinks back to pure anim-setting, or precedence moves out entirely.
> 
> wdyt?

Yes, that sounds much better. 👍
The only thing please to check is that the precedence checks be performed before a new object is created.
So if new litter shouldn’t appear, then there’s no need to create an object and then delete it. (it works this weird way in this PR now)
Thanks

CorsixTH/Lua/app.lua
```diff
@@ -28,7 +28,7 @@ local SDL = require("sdl")
 -- and add compatibility code in afterLoad functions
 -- Recommended: Also replace/Update the summary comment
 
-local SAVEGAME_VERSION = 264 -- SDL 3
+local SAVEGAME_VERSION = 265 -- SDL 3
```
lewri: You'll also need to update the summary as advised on line 29 :)

CorsixTH/Lua/world.lua
```diff
@@ -2886,6 +2886,34 @@ function World:afterLoad(old, new)
     self.system_pause = nil
   end
 
+  if old < 265 then
+    -- Avoid Stacking of Litter on the ground. Attached to a version to avoid re-work for games saved after
+    -- however this is a non-breaking change for savegames
```
lewri:
```suggestion
    -- Multiple litter can no longer exist on a tile. Remove all but the highest order of litter.
```

order/precedence.

CorsixTH/Lua/objects/litter.lua
```diff
@@ -42,12 +42,39 @@ litter_types["soot_floor"] = 3416
...
+-- Random Trash (lowest)
+litter_precedence["soda_can"] = 1
+litter_precedence["banana"] = 1
+litter_precedence["paper"] = 1
+litter_precedence["bottle"] = 1
+
+-- Explosion Damage (can't be cleaned by handyman so it can't be displaced by others)
+litter_precedence["soot_floor"] = 99
+litter_precedence["soot_wall"] = 99
+litter_precedence["soot_window"] = 99
```
lewri: It is worth noting these litter can never exist with other types, as they appear within rooms only.

CorsixTH/Lua/objects/litter.lua
```diff
@@ -78,8 +106,42 @@ function Litter:getWalkableTiles()
...
+function Litter:getPrecedence()
+  local litter_type = litter_anim_to_type[self.animation_idx]
+  return litter_precedence[litter_type]
+end
+
 function Litter:setLitterType(anim_type, mirrorFlag)
   if anim_type then
+    local objectsInTile = self.world:getObjects(self.tile_x, self.tile_y)
+
+    --This should ideally be a single item at most but older saves might still contain more than one item
```
lewri: Your afterload in world should make this irrelevant. Write with the expectation your afterload works.

CorsixTH/Lua/world.lua
```diff
@@ -2886,6 +2886,34 @@ function World:afterLoad(old, new)
...
+
+    -- Group litter by tile, keeping the highest-precedence item per tile. The
+    -- losers are collected and removed only afterwards otherwise it would lead to items being skipped.
+    local litter_by_tile = {}
+    local litter_to_remove = {}
+    for _, entity in ipairs(self.entities) do
+      if class.is(entity, Litter) and entity.tile_x then
+        local key = entity.tile_x * 10000 + entity.tile_y -- unique int key per tile
```
lewri, line 2955: There are probably two methods to achieve this afterload.

One is the way laid out here by going through litter; the other is by going through tiles.

The tile method might be simpler, though there is almost certainly likely less litter than there are tiles (16000). For this I think you're right to target your loops as litter.

However, we do previously known class.is (and other functions) can be slower over large sets. You get better optimisation doing a check of `if entity.object_type and entity.object_type.id == "litter" then`

I would note though your keying method is a little... unique. But ok

CorsixTH/Lua/world.lua
```diff
@@ -2886,6 +2886,34 @@ function World:afterLoad(old, new)
...
+          litter_to_remove[#litter_to_remove + 1] = current_litter
+          litter_by_tile[key] = entity
+        else
+          litter_to_remove[#litter_to_remove + 1] = entity
+        end
+      end
+    end
+    for _, entity in ipairs(litter_to_remove) do
+      if entity:isCleanable() then entity:remove() end
+    end
```
lewri, line 2970: Noting again about how soot can't exist in any place litter would (soot is only in a room).

CorsixTH/Luatest/spec/entities/litter_spec.lua
```diff
@@ -0,0 +1,296 @@
...
+
+  -- helper: place existing litter on tile with a given type already set
+  local function place_existing(world, hospital, litter_type)
+    local existing = make_litter(world, hospital)
+    world.objects_on_tile[#world.objects_on_tile + 1] = existing
+    existing:setLitterType(litter_type, 0)
+    return existing
+  end
+
+  -- ── isCleanable ───────────────────────────────────────────────────────────
```
lewri: Please use hyphens

CorsixTH/Lua/objects/litter.lua
```diff
@@ -78,8 +106,42 @@ function Litter:getWalkableTiles()
...
+    local litterToSetPrecedence = litter_precedence[anim_type]
```
lewri: Variables should use snake_case

CorsixTH/Lua/objects/litter.lua
```diff
@@ -78,8 +106,42 @@ function Litter:getWalkableTiles()
...
 end
 
+function Litter:getPrecedence()
+  local litter_type = litter_anim_to_type[self.animation_idx]
+  return litter_precedence[litter_type]
+end
+
 function Litter:setLitterType(anim_type, mirrorFlag)
   if anim_type then
+    local objectsInTile = self.world:getObjects(self.tile_x, self.tile_y)
```
lewri: Another one for snake_case

lewri (commented):
For biohazard the previously common naming is 'vomit inducing'; which might fit in the broader scheme on the class.

I also feel generally `priority` might fit better than `precedence`.

ARGAMX:
By the way, your source branch is called `master`.
`Flavsditz wants to merge 9 commits into CorsixTH:master from Flavsditz:master`
I think this is the most inconvenient name. 🙂
Right now, I think it’s no longer possible to rename your branch and give it a different name, so possible it’s better to leave it as is.
At least I tried to do it at some point, and to be honest, nothing good came of it.
But in the future, it would be convenient to give branches names that correspond to the changes they contain. 😉

CorsixTH/Lua/world.lua
Flavsditz: I was trying to find a proper way of doing this in Lua but I am new to the language (literally learned to be able to contribute here). Can you offer a better approach for the keying? I don't mind a bit of critique if it means I can do something better :)

CorsixTH/Lua/world.lua
Flavsditz: I'm going to comment here but it is the same point to the other soot-related comment:

Is it still worth to validate this even so it shouldn't happen? In my opinion I wanted to be careful and make sure that even if something slips by that we don't do something incorrectly and thus introduce a bug. 
So I did opt to do the checks and add soot to the hierarchy list. 

I can of course remove it all if you think this is not worth having as if a problem happen the bug would already have happened elsewhere (which I can understand and be fine with it)

CorsixTH/Lua/objects/litter.lua
Flavsditz: see comment below

CorsixTH/Lua/objects/litter.lua
Flavsditz: I think the comment is a bit misleading. 
Considering the afterload works: While common litter does have a check for only happening if a tile is clear, vomit, rat blood and pee don't. 
So that loop is still relevant as it might have to go through and not touch benches or other objects that might be on the square at the same time it already contains the new litter. 

It would mean the second loop is a short 1 item one. I am not sure it is preferrable to have it as a single variable being erased after the fact only or if this is way slower in Lua. Would someone know this?

Flavsditz:
>  For biohazard the previously common naming is 'vomit inducing'; which might fit in the broader scheme on the class.
> I also feel generally priority might fit better than precedence.

@lewri I went with the word `rank`.
 `priority` would conflict with other parts of the code such as "priority" of Handyman tasks (see handyman.lua).

Made some changes that I also think makes litter creation more elegant as the world class now takes care of both creation and setting the type. This was previously happening on the litter classes themselves.

CorsixTH/Lua/objects/litter.lua
Flavsditz: Made the changes! Ptal

CorsixTH/Lua/world.lua
lewri: There might be nothing wrong with you method. It's just one I've not seen before. There are likely other ways, yes, though none jump up mind right now.

CorsixTH/Lua/world.lua
lewri: Arguably the litter code (not this afterload) *could* `assert` that the new litter is not inside a room unless it's soot. But this is out of scope really, handling it without a prejudice is fine on this PR.

CorsixTH/Lua/objects/litter.lua
```diff
@@ -78,6 +106,45 @@ function Litter:getWalkableTiles()
...
+  local litter_type = litter_anim_to_type[self.animation_idx]
+  return litter_rank[litter_type]
+end
+
+--! Rank for a litter type before any Litter object exists.
+--! Accepts a type name ("puke", "soda_can", ...) or one of the numeric
+--! indices 1-4 used to pick a random piece of trash.
+--!param litter_type (string or int) The litter type to look up.
+--!return (int) Rank of that type, or 0 if unknown.
+function object.getRankForType(litter_type)
```
TheCycoONE: Don't define methods on the object global in litter.lua

TheCycoONE (changes requested):

CorsixTH/Lua/world.lua
```diff
@@ -2925,6 +2943,33 @@ function World:afterLoad(old, new)
...
+  if old < 265 then
+    -- Multiple litter can no longer exist on a tile. Remove all but the highest order of litter.
+
+    -- Group litter by tile, keeping the highest-ranked item per tile. The
+    -- losers are collected and removed only afterwards otherwise it would lead to items being skipped.
+    local litter_by_tile = {}
+    local litter_to_remove = {}
+    for _, entity in ipairs(self.entities) do
+      if class.is(entity, Litter) and entity.tile_x then
+        local key = entity.tile_x * 10000 + entity.tile_y -- unique int key per tile
```
ARGAMX, line 2955:
```suggestion
        local key = entity.tile_x * self.map.height + entity.tile_y -- unique int key per tile
```

TheCycoONE (approved):

CorsixTH/Lua/world.lua
TheCycoONE: @Flavsditz ARGAMX's suggestion here is better.

PR #3524, opened 2026-09-02
When building room expel humanoids on door phase

### LINKED ISSUE #1454: Missing vanilla mechanics

ARGAMX:
*Fixes #1454*

- When building room expel humanoids from blueprint on entering to the door phase, not after leaving it.

### DIFF
--- a/CorsixTH/Lua/dialogs/edit_room.lua
+++ b/CorsixTH/Lua/dialogs/edit_room.lua
@@ -198,9 +198,7 @@ function UIEditRoom:cancel()
       self.ui:setDefaultCursor(nil)
       self.check_for_clear_area_timer = nil
       self.humanoids_to_watch = nil
-      self:_setCellFlagsOnBlueprint({avoidTile = false})
     end
-    self.phase = "walls"
     self:returnToWallPhase()
   end
 end
@@ -217,19 +215,12 @@ function UIEditRoom:confirm(force)
     self.mouse_down_y = false
     self.move_rect_x = false
     self.move_rect_y = false
-    self.phase = "door"
     self:enterDoorPhase()
   elseif self.phase == "door" then
-    self.phase = "windows"
     self:enterWindowsPhase()
   elseif self.phase == "windows" then
-    self.phase = "clear_area"
-    self:clearArea()
+    self:enterWaitingClearAreaPhase()
   elseif self.phase == "clear_area" then
-    self.ui:setDefaultCursor(nil)
-    self.phase = "objects"
-    self:finishRoom()
-    self.world:resetSideObjects()
     self:enterObjectsPhase()
   else
     -- Pay for room (subtract cost of needed objects, which were already paid for)
@@ -257,8 +248,7 @@ function UIEditRoom:confirm(force)
   end
 end
 
-function UIEditRoom:clearArea()
-  self.confirm_button:enable(false)
+function UIEditRoom:clearAreaFromHumanoids()
   local rect = self.blueprint_rect
   local world = self.ui.app.world
   world:clearCaches() -- To invalidate idle tiles in case we need to move people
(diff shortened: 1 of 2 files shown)


CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -257,8 +248,7 @@ function UIEditRoom:confirm(force)
   end
 end
 
-function UIEditRoom:clearArea()
-  self.confirm_button:enable(false)
+function UIEditRoom:clearAreaFromHumanoids()
```
TheCycoONE: s/From/Of

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -880,63 +866,96 @@ function UIEditRoom:checkReachability()
   return true
 end
 
+function UIEditRoom:enterWaitingClearAreaPhase()
+  self.phase = "clear_area"
+  self.confirm_button:enable(false)
+  if next(self.humanoids_to_watch) == nil then
```
TheCycoONE, line 888: Is it possible for a humanoid to enter this avoidTile=true area after we set the flag. I don´t recall if there was a condition that allowed that or we only needed the flag to allow people to travel out of the area safely.

If it is possible then we would want to regenerate the humanoids_to_watch list in the case it was nil here, and only confirm if the list was still empty.

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -880,63 +866,96 @@ function UIEditRoom:checkReachability()
...
+    self:returnToWallPhase(true)
+    self.ui:playSound("wrong2.wav")
+    self.ui.adviser:say(_A.room_forbidden_non_reachable_parts)
+  end
+end
+
+--! Checks that the room will not block part of the hospital.
+--!return (bool) true if will not block part of the hospital.
+function UIEditRoom:checkCanBuildRoomHere()
+  local build_allowed
```
TheCycoONE: Set it to false here and drop the inner else condition.

Not every path through this function sets build_allowed with the current implementation.

CorsixTH/Lua/dialogs/edit_room.lua
ARGAMX: > s/From/Of

`clearAreaOfHumanoids` ?

CorsixTH/Lua/dialogs/edit_room.lua
```diff
@@ -880,63 +866,96 @@ function UIEditRoom:checkReachability()
...
+
   -- make tiles impassable
   self:_setCellFlagsOnBlueprint({passable = false})
 
   -- check if all adjacent tiles of the rooms are still connected
-  if not self:checkReachability() then
+  if self:checkReachability() then
+    build_allowed = true
+  else
```
TheCycoONE: There's no reason for the additional nesting, if this were an else if. That would reduce the complexity slightly.

CorsixTH/Lua/dialogs/edit_room.lua
ARGAMX: > s/From/Of

👌

CorsixTH/Lua/dialogs/edit_room.lua
ARGAMX: > Set it to false here and drop the inner else condition.
> 
> Not every path through this function sets build_allowed with the current implementation.

👌

CorsixTH/Lua/dialogs/edit_room.lua
ARGAMX: > There's no reason for the additional nesting, if this were an else if. That would reduce the complexity slightly.

👌

CorsixTH/Lua/dialogs/edit_room.lua
ARGAMX: > Is it possible for a humanoid to enter this avoidTile=true area after we set the flag. I don´t recall if there was a condition that allowed that or we only needed the flag to allow people to travel out of the area safely.
> 
> If it is possible then we would want to regenerate the humanoids_to_watch list in the case it was nil here, and only confirm if the list was still empty.

Good observation.
I’ve added handling for such a case. 👌

TheCycoONE (approved):

tobylane:
It doesn't work on patients vomiting at the same time as finishing the door phase, and the app freezes on, I think, patients entering that stalemate of those vomiters not leaving.

lewri:
> It doesn't work on patients vomiting at the same time as finishing the door phase, and the app freezes on, I think, patients entering that stalemate of those vomiters not leaving.

In my previous attempt I also found this. You have to schedule a repeat call on those that failed to leave.

ARGAMX:
> It doesn't work on patients vomiting at the same time as finishing the door phase, and the app freezes on, I think, patients entering that stalemate of those vomiters not leaving.


> > It doesn't work on patients vomiting at the same time as finishing the door phase, and the app freezes on, I think, patients entering that stalemate of those vomiters not leaving.
> 
> In my previous attempt I also found this. You have to schedule a repeat call on those that failed to leave.

There is such a problem in general.

The thing is that this all worked poorly even before this PR, with glitches and sometimes failing to expel humanoids.
There are several reasons for this, and my goal was not to solve these problems.
And to solve only what was the goal of the task.

But in order for us to understand whether certain issue you detected are regress from this PR or not, we need to reproduce issues before and with the PR.
And to do that, we need to somehow reproduce it.

Well, to fix it we still need to be able to reproduce it.

PR #3528, opened 2026-09-07
Use cheaper search for objects in world

tobylane:
- When searching for objects near a humanoid, use the world.objects table instead of entitymap for a nearly 2x speedup. 

What is entitymap for? I see it as a two-table deep version of the one-table deep index used here.

### DIFF
--- a/CorsixTH/Lua/entities/humanoid.lua
+++ b/CorsixTH/Lua/entities/humanoid.lua
@@ -557,16 +557,19 @@ function Humanoid:findObjectsInSquare(size, object_spec)
 
   -- Search the rectangle for as far as it is within the world.
   local width, height = world_map.th:size()
-  local entity_map = self.world.entity_map
 
   size = (size >= 0) and size or 0
   for x = self.tile_x - size, self.tile_x + size do
     if x >= 1 and x <= width then
       for y = self.tile_y - size, self.tile_y + size do
         if y >= 1 and y <= height and th_map:getRoomId(x, y) == self_room_id then
-          for _, obj in ipairs(entity_map:getObjectsAtCoordinate(x, y)) do
-            local entry = objs_table[obj.id]
-            if entry then entry[#entry + 1] = obj end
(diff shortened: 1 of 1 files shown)


lewri:
#3308

TheCycoONE:
I suppose you profiled. Is the speedup lost if there was a helper function for getting the objects at a tile in world? I know the abstraction won't be free in lua.

TheCycoONE:
To answer your earlier question, the entityMap was meant to be an efficient way to index objects and humanoids; it has a nice clean interface and we lacked a profiler back then. If there's ways to improve performance by cleaning up the internal representation then we can drop the world.objects list because we don't need both.

ARGAMX (approved):
👍

tobylane:
The end number is the profile measurement, the mean in microseconds.

1. Before this change, 34
2. Without the assert in EntityMap:getObjectsAtCoordinate, 30
3. With entity_map[x][y]["objects"] used in Humanoid:findObjectsInSquare, 26
4. Using world:getObjects, 25
5. Using entity_map:getObjectsAtCoordinate(x, y) with assert and only looking at the first result, 19
6. With world.objects[index] used in Humanoid:findObjectsInSquare (current PR), 18
7. Using world:getObjects and only looking at the first result, 17

5 and 7 may be possible once litter does not stack? At that point, adjust World:getObject, World:getObjects, and other places where multiple items on one tile are assumed, to take out an unnecessary loop.
```lua
          local objects = self.world:getObjects(x, y) -- or entity_map:getObjectsAtCoordinate(x, y) 
          if objects and objects[1] then
            local entry = objs_table[objects[1].object_type.id]
            if entry then entry[#entry + 1] = objects[1] end
          end
```

edit: Radiation shield's two parts are the only thing sharing a tile in the big hospital I'm testing with.

TheCycoONE:
Merging but recognizing that entity_maps whole reason for existence is to be efficient, so maybe it needs to be rewritten.

PR #3529, opened 2026-09-07
Fix/xmi2mid note ordering

### LINKED ISSUE #3510: [Bug] XMI to MIDI conversion silences notes and truncates overlapping ones
### Describe the issue

XMI stores an explicit duration per note rather than note off events, so `xmi_to_midi_token_list` synthesizes a note off for each note on, appends it out of order, then sorts the whole list. The comparator at CorsixTH/Src/xmi2mid.cpp:41 compares only `time`, and the sort at :323 is `std::sort`, which is not stable - so events sharing a timestamp come out in an unspecified order.

Three consequences:

- Notes silenced. Where a note ends on the same tick the next note of that pitch begins, the note off can be ordered after the note on and silence it instantly. Built with MSVC 19.44, 12 notes across the eight tracks are lost this way. Because the ordering is unspecified rather than merely wrong, which notes are affected varies by compiler and standard library - which is likely why this has gone unnoticed.

- Overlapping notes truncated. A MIDI note off applies to a whole (channel, note) pair, so where a pitch is re-struck before its previous instance ends, the earlier note's release cuts the later one short. 144 notes are affected.

- A phantom held note. FORTUNE.XMI contains one zero-duration note. Both releases sort ahead of the note on, leaving a ~7 second piano note at 0:03.475 that is not in the source.

This affects both playback paths, since `xmi_to_midi_token_list` feeds SDL_mixer via `transcodeXmiToMid` and also the MIDI device player in `midi_player.cpp`.

Measuring actual sounding time rather than counting events, the audible impact is 11 windows of 40 ms or more across the whole soundtrack. While this is overall minor, the most easily noticed issue of the held and missing notes in FORTUNE does change the chord structure compared to the original.

### Steps to Reproduce

Convert FORTUNE.XMI with CorsixTH's converter and compare channel 1 against the same file converted by foobar2000's foo_midi in a DAW or MIDI editor. Two piano notes of ~3.5 s at 0:27.817 and 1:13.033 are absent, and a ~7 s note is present at 0:03.475 that should not be.

CANDY.XMI shows the same class of defect on channel 4 (five lead notes) and channel 10 (ride cymbal at 0:57.267 and 1:11.450).

I have prepared a brief screen recording illustrating the difference in the note programming and to provide a short audio sample from each song. Original, then a fixed implementation of CorsixTH's converter, then XMI -> MIDI as converted by foo_midi. The cymbal issue present in CANDY is not demonstrated as it is not easily audible, but examining the MIDI programming between CorsixTH's conversion and the original XMI shows the difference.

FORTUNE: 

https://github.com/user-attachments/assets/5495ddca-5aaf-44d6-bbfd-43fdffc0b265

CANDY:

https://github.com/user-attachments/assets/08afeb41-c8da-4387-94aa-52682dad3871

### Expected Behaviour

Converted MIDI should preserve the note content of the source: a note should never be silenced by its own release, and overlapping notes of the same pitch should not truncate each other.

- [ ] Right mouse panning

### Save Game

### CorsixTH Version

0.7.0 (release version)

### Operating System

Windows 11

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

### Additional Information

I am finalizing a fix for CorsixTH's XMI to MIDI converter and will open a pull request once ready.

Ligushka:
*Fixes #3510 XMI to MIDI conversion silences notes and truncates overlapping ones*

- Adjusts ordering of note-off events to occur before note-ons when they share a timestamp and makes the sort stable, so the output no longer depends on the standard library's handling of equal keys
- Resolves note-offs once every note is known, merging overlapping notes of the same pitch into a single run, so an earlier note's release no longer cuts a later note of the same pitch short
- Adds unit tests covering both defects; game data is not required for these tests as XMI is generated in memory
- Across the eight original Theme Hospital tracks, this eliminates 12 silenced notes and 144 truncated ones; the issue thread illustrates an example of the original behavior vs fixed behavior

Transparency notice: Claude Opus 5 was used to co-author this fix. I have verified in-game that the tracks now sound faithful to the original.


lewri:
@ravenPenfold if you have ability to test you are welcome to feed back too

CorsixTH/CppTest/test_xmi2mid.cpp
```diff
@@ -0,0 +1,330 @@
+/*
+Copyright (c) 2026 CorsixTH contributors
```
lewri: We don't need a copyright on this file, but the author is you (Ligushka, or your name, or both) if you wanted to keep it.

lewri (commented):
Thank you :) Cyco/Albert will be best reviewing C++ code for you.

Looks like you have code test errors though, you can review these by clicking the failed test.

CorsixTH/CppTest/test_xmi2mid.cpp
Ligushka: Thank you, I'll eliminate that header & resolve the code test errors and ensure I can cover the other code tests locally for any future commits. I'll submit a corrected PR a bit later today.

TheCycoONE:
This is largely the same as the algorithm I proposed in the ticket, I have no issues with it. I believe that the state table could store less data. The last_on_time is probably not needed (the file is processed sequentially, it's implied that the last_on_time is earlier or the same as the current note on, and regardless the new note_off_time is the greater of the current and what's in the state table. I suppose it makes a small difference if you intend to turn the previous note off and the file has two of the same note same channel start at the same time - which might be enough justification to keep it.

state.sounding is more definitively useless and misleading; any note_state that was ever set to true is never set to false again, so it's equivalent of `end_time != 0`.

TheCycoONE:
@Ligushka did you have a chance to review my comment?

PR #3531, opened 2026-09-08
General user soundfont

### LINKED ISSUE #3509: [Bug] FluidR3 SoundFont silently drops certain notes of the original music
### Describe the issue

FluidR3's "Baritone Sax" (GM program 67) provides no coverage above MIDI note 72. "Night Shift" writes 262 notes on that channel spanning MIDI notes 51-83; the 44 above the limit (17%) are silent.

This is not an issue with CorsixTH's MIDI playback functionality, as the same issue occurs in REAPER as demonstrated below, plus foobar2000's "foo_midi" plugin. This is not a defect in FluidR3, which deliberately models real instrument ranges. It is a mismatch between that design and the original score, which was authored for banks whose presets span the full keyboard.

I have not checked any of the other original music scores for similar issues; it is possible that other portions of songs could be similarly affected.

As a replacement SoundFont, you may wish to consider the highly compatible SoundFont ["GeneralUser GS"](https://schristiancollins.com/generaluser.php). I have verified that this SoundFont loads in CorsixTH 0.7.0 and that the output is musically complete. GeneralUser GS also explicitly mentions full compatibility with FluidSynth 2.3 or later. CorsixTH 0.7.0 bundles FluidSynth 2.6.

Excerpt from GeneralUser GS's license from `documentation\LICENSE.txt` bundled with `GeneralUser GS 2.0.3 w/ documentation rev. 6 (2026-2-23)`:

```
** License of the complete work **
You may use GeneralUser GS without restriction for your own music creation,
private or commercial. This SoundFont bank is provided to the community free of
charge. Please feel free to use it in your software projects, and to modify the
SoundFont bank or its packaging to suit your needs.
```

### Steps to Reproduce

Listen to "Night Shift" using the "Default (Software)" MIDI API and the default SoundFont bundled with CorsixTH, then listen again using another SoundFont such as the suggested GeneralUser GS or even an old Creative SoundFont.

Furthermore, here is a screen recording of playback of just channel 5 in NIGHTSH.XMI (converted to MIDI) played back in REAPER using the "juicysfplugin" VST3 set to use the FluidR3.sf3 bundled with CorsixTH 0.7.0 using the original "67 Baritone Sax" specified by the MIDI file, then immediately with "68 Oboe" to demonstrate how much music is missing.

https://github.com/user-attachments/assets/ba6581a2-cb0f-41da-ba4c-6cb2a1f5e24c

### Expected Behaviour

Output of the originally composed Bullfrog music should be complete, not missing notes.

- [ ] Right mouse panning

### Save Game

### CorsixTH Version

0.7.0 (release version)

### Operating System

Windows 11

### Theme Hospital Version

CD

### Gamelog.txt

```yaml

```

### Additional Information

If it is decided to replace FluidR3, which is fetched at build time, these definitions will need to be updated:

- `CorsixTH/CMakeLists.txt:265-275` — the `FetchContent_Declare` with a hardcoded URL and `URL_HASH`
- `CorsixTH/CMakeLists.txt:385` — the `install(FILES ${FLUIDR3_FILE} ...)` rule
- `LICENSE.txt:732` — the bundled FluidR3 license text
- `CorsixTH/Lua/config_finder.lua:613` and `WindowsInstaller/config_template.txt:317` — user-facing docs naming FluidR3

TheCycoONE:
*Fixes #3509*

I switched from FetchContent to file(DOWNLOAD) as it's more purpose built to adding a single file to the install. FetchContent is really around adding libraries (as in the case of SDL3_MIXER).

### DIFF
--- a/CorsixTH/CMakeLists.txt
+++ b/CorsixTH/CMakeLists.txt
@@ -276,15 +276,20 @@ if(WITH_MIDI_DEVICE)
 endif()
 
 if(FETCH_SOUNDFONT)
-  FetchContent_Declare(
-    FluidR3
-    URL https://raw.githubusercontent.com/Jacalz/fluid-soundfont/master/SF3/FluidR3.sf3
-    URL_HASH SHA256=32039e039c2f708467a6f171fbfd9fecdcadfe40a2327837c391490c8de70021
-    DOWNLOAD_NO_EXTRACT TRUE
+  set(SOUNDFONT_FILE "${CMAKE_BINARY_DIR}/GeneralUser-GS.sf2")
+  file(DOWNLOAD
+    https://raw.githubusercontent.com/mrbumpy409/GeneralUser-GS/684543d5e5efaef08d02be50dcda8d552478fa60/GeneralUser-GS.sf2
+    "${SOUNDFONT_FILE}"
+    EXPECTED_HASH SHA256=9575028c7a1f589f5770fccc8cff2734566af40cd26ed836944e9a5152688cfe
+    STATUS SOUNDFONT_STATUS
   )
 
-  FetchContent_MakeAvailable(FluidR3)
-  set(FLUIDR3_FILE ${fluidr3_SOURCE_DIR}/FluidR3.sf3)
+  list(GET SOUNDFONT_STATUS 0 SOUNDFONT_STATUS_CODE)
+  if(NOT SOUNDFONT_STATUS_CODE EQUAL 0)
+    message(FATAL_ERROR "Failed to download soundfont: ${SOUNDFONT_STATUS}")
+  else()
+    message("  Soundfont found")
+  endif()
 endif()
 
 if(FETCH_UNICODE_FONT)
@@ -395,7 +400,7 @@ if(NOT USE_SOURCE_DATADIRS)
   install(FILES CorsixTH.lua ../LICENSE.txt DESTINATION ${CORSIX_TH_DATADIR})
 
   if(FETCH_SOUNDFONT)
-    install(FILES ${FLUIDR3_FILE} DESTINATION ${CORSIX_TH_DATADIR})
+    install(FILES ${SOUNDFONT_FILE} DESTINATION ${CORSIX_TH_DATADIR})
   endif()
   if(FETCH_UNICODE_FONT)
     install(FILES ${GONOTO_FILE} DESTINATION ${CORSIX_TH_DATADIR})
--- a/CorsixTH/CMakeLists.txt
+++ b/CorsixTH/CMakeLists.txt
@@ -293,16 +293,20 @@ if(FETCH_SOUNDFONT)
 endif()
 
 if(FETCH_UNICODE_FONT)
-  FetchContent_Declare(
-    GoNoto
-    URL https://github.com/satbyy/go-noto-universal/releases/download/v7.0/GoNotoKurrent-Regular.ttf
-    URL_HASH SHA256=2f2cee5fbb2403df352ca2005247f6c4faa70f3086ebd31b6c62308b5f2f9865
-    DOWNLOAD_NO_EXTRACT TRUE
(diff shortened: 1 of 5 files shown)


lewri (commented):
Maybe to appease Debian we should explicitly leave in some switcher between the two files (to show at least some level of intent we compensate for those who query the authenticity of GeneralUserGS.

They might call that having the above soundfont default (in cmake) not good enough.

TheCycoONE:
> Maybe to appease Debian we should explicitly leave in some switcher between the two files (to show at least some level of intent we compensate for those who query the authenticity of GeneralUserGS.
> 
> They might call that having the above soundfont default (in cmake) not good enough.

I left in auto-picking FluidR3 and Debian doesn't use FETCH_SOUNDFONT so they are completely unaffected.

CorsixTH/CMakeLists.txt
```diff
@@ -276,28 +276,37 @@ if(WITH_MIDI_DEVICE)
...
-    FluidR3
-    URL https://raw.githubusercontent.com/Jacalz/fluid-soundfont/master/SF3/FluidR3.sf3
-    URL_HASH SHA256=32039e039c2f708467a6f171fbfd9fecdcadfe40a2327837c391490c8de70021
-    DOWNLOAD_NO_EXTRACT TRUE
+  set(SOUNDFONT_FILE "${CMAKE_BINARY_DIR}/GeneralUser-GS.sf2")
+  file(DOWNLOAD
+    https://raw.githubusercontent.com/mrbumpy409/GeneralUser-GS/684543d5e5efaef08d02be50dcda8d552478fa60/GeneralUser-GS.sf2
+    "${SOUNDFONT_FILE}"
+    EXPECTED_HASH SHA256=9575028c7a1f589f5770fccc8cff2734566af40cd26ed836944e9a5152688cfe
+    STATUS SOUNDFONT_STATUS
```
lewri, line 284: I suspect the soundfont change may leave FluidR3 behind in the installation directory on Windows. That isn't ideal as first launch of the game it may unreliably pick one of the soundfonts at random.

Do you have the knacks still to update the Win32Script.nsi -- I think we just need to set a remove on any .sf2/,sf3 in the root INSTALLDIR?

LICENSE.txt
```diff
@@ -729,30 +729,55 @@ Apache License 2.0
...
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
-EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
-OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
-NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
-HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
-WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
-FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
-OTHER DEALINGS IN THE SOFTWARE.
+Uses GeneralUser-GS Soundfont which is available under the following terms:
```
lewri, line 732: Is this 'Optionally uses'? Though Windows doesn't/won't get a choice.

TheCycoONE: No change from Fluid in that respect, we don't say Optionally in this file. It's the license for our binary releases.

CorsixTH/CMakeLists.txt
TheCycoONE: It won't be randomly, it goes through the list in order so GeneralUser will always be picked first, but yes better to clean it up.

TheCycoONE:
Didn't mean to push that last commit, undone.

tobylane (approved):

PR #3536, opened 2026-09-09
Remove pr checkout from github actions for security

tobylane:
- Following #3500, upgrade to [checkout@v7](https://github.com/actions/checkout#checkout-v7) and meet the higher security standard by taking out the PR checkout option.


TheCycoONE:
Assume pipelines should be passing before we review?

.github/workflows/Linux.yml
```diff
@@ -36,12 +33,7 @@ jobs:
     runs-on: ubuntu-26.04 # resolute raccoon
     name: Linux apt-get ${{matrix.lua}} ${{inputs.animview == 'true' && 'with AnimView' || ''}}
     steps:
-      - uses: actions/checkout@v6  # Keeps the git OAuth token after checkout
-      - name: Checkout selected PR
-        env:
-          GH_TOKEN: ${{github.token}}
-        if: inputs.PR
-        run: gh pr checkout --repo CorsixTH/CorsixTH ${{inputs.pr}}
+      - uses: actions/checkout@v7  # Keeps the git OAuth token after checkout
```
TheCycoONE, line 36: What is this comment informing, or was it meant to be temporary?

.github/workflows/Linux.yml
TheCycoONE, line 6: Should this PR let us remove this permission?

tobylane: We don't access the PR (the discussion), only the contents of the branch, so it can go.

.github/workflows/Linux.yml
tobylane: It is used in the commit to update the docs, so I've now moved it to that checkout only.

TheCycoONE (approved):

PR #3537, opened 2026-09-10
[POC] Add WebAssembly support via Emscripten and SDL3

pippy360:
This PR adds support for building and running CorsixTH in web browsers via WebAssembly using Emscripten and SDL3.

### Key Changes:
- **Non-blocking Event Loop (`CorsixTH/Src/sdl_core.cpp`)**: Implemented an asynchronous `emscripten_step` callback driven by `emscripten_set_main_loop_arg` to process SDL events, execute 18 ms simulation ticks, and render frames without blocking the browser main thread.
- **Lua State Lifetime (`CorsixTH/SrcUnshared/main.cpp`)**: Released ownership of `lua_State` with `L.release()` when targeting Emscripten so the Lua state remains valid across asynchronous browser frame callbacks.
- **Filesystem & Path Resolution (`libs/whereami/whereami.c`, `CorsixTH/Src/main.cpp`)**: Added Emscripten support in `whereami` returning `/corsix-th` and added `corsix-th/` to default search directories.
- **Root Directory & Path Normalization (`CorsixTH/Lua/utility.lua`, `CorsixTH/Lua/app.lua`)**: Fixed `stripTrailingSlashes` preserving root `"/"` (preventing an empty string return that caused `lfs.dir` errors in virtual root environments), and hardened `findMusicDir` / `initMusicDir` with `pcall` against directory traversal errors.
- **CMake Configuration (`CMakeLists.txt`, `CorsixTH/CMakeLists.txt`)**: Added Emscripten target properties, SDL3/FreeType/PNG/ZLIB link options, static Lua module linking (`lfs`, `lpeg`), and added optional cache variables `CORSIX_TH_PRELOAD_CORSIX_DIR` and `CORSIX_TH_PRELOAD_TH_DIR` for asset preloading.
- **Ignored Directories (`.gitignore`)**: Added WebAssembly build and staging folders (`corsix-wasm/`, `th_data/`, `run_web.py`) to `.gitignore`.

### Testing:
- Built with Emscripten (`emcc` 6.0.5) and SDL3 port.
- Tested and verified in Google Chrome: title screen, animations, interactive UI, and in-game Level 1 hospital simulation run without errors.


pippy360:
this is an AI generate wasm port, I'm using this myself so I thought I'd create a pull request in case it helps anyone else

TheCycoONE:
Are you interested in upstreaming some or all of these changes, or simply sharing? I wouldn't mind taking the trailing slash change for example, but I'd prefer using CorsixTH as the directory rather than add another search path if it makes no difference.

pippy360:
>Are you interested in upstreaming some or all of these changes, or simply sharing?

Whatever you guys want, I'm easy, if it's useful to you guys then please upstream it or edit it however you want

tobylane:
What does this create, js or also the wrapper html? Are there any required external files? Ideally we want a way for browser users to bring their own assets, not relying on the host to have preloaded copyright content. Are you familiar enough with emscripten to make manual changes like this?

TheCycoONE:
The way I'm reading this is take what we want, but don't expect ongoing or even current work because it satisfies your itch. If that's what you mean then thanks for being up front.

I'm impressed with how much cleaner it is than the prior emscripten effort.

TheCycoONE:
SDL3 supports using callbacks instead of a game loop; as outlined here: https://wiki.libsdl.org/SDL3/README-main-functions

I suspect that would make much of the sdl-core differences go away, and might be a motivation to pivot eventually.

pippy360:
>The way I'm reading this is take what we want, but don't expect ongoing or even current work because it satisfies your itch. If that's what you mean then thanks for being up front.

yeah exactly. I'm not trying to dump AI slop on you guys.

>I'm impressed with how much cleaner it is than the prior emscripten effort.

It actually looks like it might be missing some files, like the wrapper html. it does work though when all files are present (I've played it). I'm a bit tight on time to do any actual work on it at the moment.

TheCycoONE:
Thanks for sharing @pippy360.

I've marked the PR as POC, in that we do not intend to merge; but people can read it and be inspired. I think there are some good ideas here.

PR #3541, opened 2026-09-11
Add faithful wording

lewri:
*Fixes #*

- Adds the worth `faithful` before `reimplementation` in documented files to set expectations CorsixTH does not hold itself to be a true replica of Theme Hospital, including not being bug for bug compatible.

### DIFF
--- a/CorsixTH/corsix-th.6
+++ b/CorsixTH/corsix-th.6
@@ -37,7 +37,7 @@
 .Op Fl -load Ns = Ns Ar savegame
 .Op Fl -lua-dir Ns = Ns Ar path
 .Sh DESCRIPTION
-CorsixTH is a reimplementation of the 1997 Bullfrog business simulation game Theme Hospital.
+CorsixTH is a faithful reimplementation of the 1997 Bullfrog business simulation game Theme Hospital.
 As well as faithfully recreating the original, CorsixTH adds support for modern operating systems, high resolution displays, and much more.
 A purchased copy of Theme Hospital, or the Theme Hospital demo is required to use CorsixTH.
 .Pp
--- a/CorsixTH/com.corsixth.corsixth.metainfo.xml
+++ b/CorsixTH/com.corsixth.corsixth.metainfo.xml
@@ -7,12 +7,12 @@
   <name>CorsixTH</name>
   <summary>Open source clone of Theme Hospital</summary>
   <description>
-    <p>CorsixTH aims to reimplement the game engine of Theme Hospital, and be
-       able to load the original game data files. This means that you will need
-       a purchased copy of Theme Hospital, or a copy of the demo, in order to
-       use CorsixTH. After most of the original engine has been reimplemented in
-       open source code, the project will serve as a base from which extensions
-       and improvements to the original game can be made.</p>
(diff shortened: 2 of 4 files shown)


TheCycoONE:
Is it a British-ism to emphasize something to imply the opposite? Faithful to me is very much the wrong word. Enhanced would be more appropriate.

lewri:
> Is it a British-ism to emphasize something to imply the opposite? Faithful to me is very much the wrong word. Enhanced would be more appropriate.

Faithful as in with best intentions here. Enhanced could be open to an interpretation too I imagine. 

I only went with faithful as you and Albert agreed it fit better in discord as a description of the project (and was originally suggested by someone independently a while back)

TheCycoONE:
Sorry I don't recall.

Alberth289346:
Faithful can be interpreted as aiming for a 1:1 implementation wrt the original. Not sure that completely captures the spirit.

Perhaps rephrase to something in the direction of "with all the features of the original game" or so? To me that seems bit more open to adding other features.

----
"being able to load original save files" can probably be deleted as goal of this project, as I don't see how you'd manage that within this project, due to the clean-room policy.

ARGAMX:
Maybe we should just write that CorsixTH aims to implement a game that feels similar to the Theme Hospital?
In my opinion, these are less flashy and more honest words. The game is indeed just feels similar.

ARGAMX:
The parts about 

> and be able to load the original game data files

and 

> After most of the original engine has been reimplemented in open source code, the project will serve as a base from which extensions and improvements to the original game can be made"

I also don't like much.
After reading this, I imagine a system of extensions similar to those in OpenTTD and OpenXcom.
Considering the progress made in this area over the past 15 years I think these are overly grand promises.
We need to be realistic; we don’t have the resources to do this.
If someone does this in the future — well, that’s great. But why talk about it without doing anything...
I would take this out.

MandrakeHorse:
I think we need to change the wording of "_and be able to load the original game data files_" to something like "_and requires the original game files_". The former is easily misinterpreted as "you will someday be able to import your save files from Theme Hospital and continue playing in CorsixTH", rather than "you will need a digital or physical copy of Theme Hospital in order to play CorsixTH".

lewri:
I think we're creeping out of scope a bit here after Alberth's comment?

Looping back to that, taking a leaf from how people compare Two Point Hospital to TH, the word 'spiritual' does come to mind too.

Alberth289346:
spiritual seems fine to me.

CorsixTH/corsix-th.6
```diff
@@ -37,7 +37,7 @@
 .Op Fl -load Ns = Ns Ar savegame
 .Op Fl -lua-dir Ns = Ns Ar path
 .Sh DESCRIPTION
-CorsixTH is a reimplementation of the 1997 Bullfrog business simulation game Theme Hospital.
+CorsixTH is a faithful reimplementation of the 1997 Bullfrog business simulation game Theme Hospital.
```
lewri:
```suggestion
CorsixTH is a spiritual reimplementation of the 1997 Bullfrog business simulation game Theme Hospital.
```

PR #3544, opened 2026-09-15
Add support for compressed saves

### LINKED ISSUE #3534: Compress save games
----

### Describe the idea

CorsixTH saves seem to be quite large for a typical game save.
If you just start a campaign and immediately save the game without doing anything at all, the save size will already be 550 kilobytes.
If you start playing on some large map, the size of your save can reach several megabytes.
This isn’t a huge size on its own, but a large group of saves can result in a more impressive volume.
For example, my Saves folder for CorsixTH currently has a size of 1.5 gigabytes.
Among them, there are saves that are 8 megabytes in size.

I tried compressing this 8 MB save using different algorithms, and here’s what I got:

zip — 1500 KB
BZ2 — 632 KB
zstd — 506 KB
LZMA2 — 419 KB

I tried compressing that initial save file of 550 kilobytes and got 26 KB output for LZMA2.

You can try it online yourself.
https://www.h-schmidt.net/Zstd/ 
https://www.ezyzip.com/create-7z-file-online.html#

Overall, the save file can be compressed in size by 10–20 times. This is a good result.

Why not integrate packing and unpacking into the game during saving and loading?

----

### CorsixTH Version

0.70.1

----

TheCycoONE:
Compression and decompression is performed with zstd.

This version is a simple implementation with fairly low hanging optimization opportunities left on the table.

Fixes #3534


ARGAMX:
Works smooth 👍
The old saves load perfectly.

------

For testing purposes, I thought about what could be done differently to make it break.
I tried loading this new save in builds where this functionality isn’t available yet.
It doesn’t load, of course. But I’m curious what it will say.
Let's take our current master.

In the file interface, it’s basically not obvious at all that these are incompatible saves. 
Just like it always was with us.

Screenshot 2026-09-15 at 08 45 31

Ok. Lets press it:
After attempt to load it gives `persistance.lua` error:

Image

Overall, we used to encounter exactly the same error when loading incompatible saves from newer versions.

So, it seems that everything looks pretty good in this strange experiment. 👍

ARGAMX:
It also occurred to me to unpack the new save manually, try opening it in Notepad, and also try loading it in the current master.

I unpacked the new save here:
https://www.h-schmidt.net/Zstd/

Then I opened unpack result in Notepad — everything’s great! I can extract the log manually!

I ran the build from the master and tried loading unpack result — it loads!!!

This is very good. 👍

CorsixTH/CMakeLists.txt
```diff
@@ -331,6 +331,11 @@ if(WITH_TRACY)
   target_link_libraries(CorsixTH_lib PUBLIC Tracy::TracyClient)
 endif()
 
+if(WITH_SAVE_COMPRESSION)
```
lewri: I'd push we make this mandatory if possible. We probably don't want to to have versions with and without save compression support.

TheCycoONE: I was thinking it requires an extra dependency and it isn't a core/required feature. Most people are only sharing saves with us. On by default but optional felt right to me. I can be pursuaded - zstd isn't exactly rare.

TheCycoONE: I accept that I either have to make it mandatory or add a suitably descriptive error if the user tries to open a zstd compressed save without support; and of the two mandatory is simpler.

lewri: Yea. The end user doesn't have to be involved at all if it's mandatory.  Likely for the greater good.

If you did for some reason have a case where two friends shared saves between a compression enabled and disabled build of the game you'd need an external decompressor (probably not hard to find online but it adds some friction).

TheCycoONE:
I guess we don't come even close to making our frame time with saves, even without compression:

image

Otherwise I'd be a bit concerned that the compression time (3ms in this case) was a sizable chunk.

TheCycoONE:
This is based off the command line utility not the application, but it was a quicker way to compare compression levels (with some overhead for the cli)

```
-rw-r--r-- 1 stephen stephen 979K Sep 16 21:49 new2026battenberg.sav

[stephen@drink ztest]$ ./test2.sh new2026battenberg.sav 
Level    Compressed     Ratio        Time (ms)      Verified  
1        257K           26.20%       4.440          yes       
2        242K           24.66%       4.462          yes       
3        227K           23.11%       5.069          yes       
4        225K           22.99%       5.549          yes       
5        218K           22.25%       7.730          yes       
6        206K           20.99%       9.005          yes       
7        204K           20.81%       8.974          yes       
8        203K           20.75%       11.197         yes       
9        203K           20.74%       11.232         yes       
10       202K           20.61%       11.580         yes       
11       201K           20.54%       13.682         yes       
12       201K           20.54%       14.135         yes       
13       202K           20.56%       20.748         yes       
14       201K           20.47%       24.512         yes       
15       201K           20.45%       29.308         yes       
16       193K           19.67%       58.494         yes       
17       188K           19.15%       64.422         yes       
18       177K           18.03%       91.084         yes       
19       176K           17.96%       138.290        yes       
20       176K           17.96%       137.762        yes       
21       176K           17.96%       138.289        yes       
22       176K           17.96%       137.251        yes  

-rw-r--r-- 1 stephen stephen 2.7M Sep 16 21:40 Argus_SuperDuper_100.sav
[stephen@drink ztest]$ ./test2.sh Argus_SuperDuper_100.sav 
Level    Compressed     Ratio        Time (ms)      Verified  
1        974K           36.00%       7.441          yes       
2        811K           29.98%       8.689          yes       
3        741K           27.39%       10.716         yes       
4        727K           26.87%       12.185         yes       
5        689K           25.44%       16.748         yes       
6        659K           24.35%       21.921         yes       
7        648K           23.95%       23.966         yes       
8        645K           23.82%       28.589         yes       
9        642K           23.72%       27.752         yes       
10       636K           23.51%       32.175         yes       
11       633K           23.39%       38.837         yes       
12       633K           23.38%       42.282         yes       
13       630K           23.29%       73.621         yes       
14       627K           23.18%       87.963         yes       
15       626K           23.14%       104.556        yes       
16       598K           22.08%       198.154        yes       
17       585K           21.63%       226.306        yes       
18       549K           20.29%       326.441        yes       
19       547K           20.22%       434.567        yes       
20       547K           20.22%       437.200        yes       
21       547K           20.22%       443.738        yes       
22       547K           20.22%       432.672        yes
```

TheCycoONE:
Based on the above results I bumped the compression level to 4. I think this is ready now.

ARGAMX (approved):
Works nice for me
Good job 👍

PR #3546, opened 2026-09-17
[RFC] Add end conditions advice

tobylane:
- Generate advice on the progress of the hospital towards the world's win and lose goals. General advice is given on the second of the month, important (nearly losing) advice is given then and on the 16th, arbitrary free days in the advice schedule. All advice is considered equal, but I'm wondering about including less of the generic advice.
In ThemeHospital the financial criteria goal advice can be given as soon as the money is spent, eg a room is completed, but this change will ignore this and generate advice on the twice monthly schedule.
The judgement scale is -3 (worse than 75% towards failing) to +4 (goal met) as there's no need to say the goal is lost. This won't be the original scale, but we can move strings around within it from TH97 observations, or write new strings. I had improve_reputation at 0 at first, but half of the campaign starts off only matching that negative message, so maybe nothing should be 0.
https://github.com/CorsixTH/CorsixTH/blob/master/CorsixTH/Lua/languages/original_strings.lua#L517 for warnings, some further to use?


CorsixTH/Lua/endconditions.lua
```diff
@@ -249,3 +252,134 @@ function EndConditions:getAttribute(hospital, attribute)
...
+  "patients_unhappy", "patients_annoyed",
+  "hospital_is_rubbish",
+})
+
+-- Map strings to the statistic they are formatted with
+local format_map = {
+  target = list_to_set({ "reputation_good_enough", "financial_criteria_met", "hospital_value_enough" }),
+  gap = list_to_set({ "improve_reputation", "close_to_win_increase_value", "financial_trouble",
+    "financial_trouble2", "financial_trouble3" }),
+  -- measure = list_to_set({""})
```
TheCycoONE, line 321: leftover?

tobylane: For the future. There can be a string like "Well done for getting your reputation to 710, that'll meet targets".

lewri: You might convert this to a `-- maybe TODO` comment? Or gauge usefulness in a separate enhancement ticket.

lewri (commented):
I'd ideally like to see we don't accidentally spit out the same advice message on the 2nd and 16th. E.g. you don't need to be told you're 3/4 way to losing the level twice a month.

tobylane:
Added a check to giving advice, and todo comments.

